PR #6185, opened 2016-04-14
Type check variables between modules

jen20:
This patch makes the tests pass - we expect an error if a string value is assigned to a variable which should be a map. This will need extending for lists/maps etc but is still relevant for the next point release, hence targeting master for now.

### DIFF
--- a/terraform/eval_variable.go
+++ b/terraform/eval_variable.go
@@ -2,11 +2,74 @@ package terraform
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/config"
+	"github.com/hashicorp/terraform/config/module"
 	"github.com/mitchellh/mapstructure"
 )
+
+// EvalTypeCheckVariable is an EvalNode which ensures that the variable
+// values which are assigned as inputs to a module (including the root)
+// match the types which are either declared for the variables explicitly
+// or inferred from the default values.
+//
+// In order to achieve this three things are required:
+//     - a map of the proposed variable values
+//     - the configuration tree of the module in which the variable is
+//       declared
+//     - the path to the module (so we know which part of the tree to
+//       compare the values against).
+//
+// Currently since the type system is simple, we currently do not make
+// use of the values since it is only valid to pass string values. The
+// structure is in place for extension of the type system, however.
+type EvalTypeCheckVariable struct {
+	Variables  map[string]string
+	ModulePath []string
+	ModuleTree *module.Tree
+}
+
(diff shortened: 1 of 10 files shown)


terraform/eval_variable.go
```diff
@@ -2,12 +2,58 @@ package terraform
...
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/config"
+	"github.com/hashicorp/terraform/config/module"
 	"github.com/mitchellh/mapstructure"
 )
 
+type EvalTypeCheckVariable struct {
```
phinze: Would you mind doing a prose comment block summarizing the intended use case and context of the EvalNode? I've found these to be helpful when navigating unfamiliar EvalTrees, and the name is often not enough to fully get what the node is supposed to do.

terraform/eval_variable.go
```diff
@@ -2,12 +2,58 @@ package terraform
...
+	Variables  map[string]string
+	ModulePath []string
+	ModuleTree *module.Tree
+	DepPrefix  string
```
phinze: I believe `DepPrefix` is vestigal?

terraform/eval_variable.go
```diff
@@ -2,12 +2,58 @@ package terraform
...
+	targetConfig := currentTree.Config()
+
+	prototypes := make(map[string]config.VariableType)
+	for _, variable := range targetConfig.Variables {
+		prototypes[variable.Name] = variable.Type()
+	}
+
+	for name, declaredType := range prototypes {
+		// This is only necessary when we _actually_ check
+		// proposedValue := n.Variables[name]
```
phinze: Didn't end up using `proposedValue` it looks like.

terraform/eval_variable.go
jen20: Ah yes, correct - I missed it from this one.

terraform/eval_variable.go
jen20: Yes that seems like a worthwhile effort - I think several others can be documented based on the research that went into this PR too!

terraform/eval_variable.go
```diff
@@ -2,12 +2,75 @@ package terraform
...
+// In order to achieve this three things are required:
+//     - a map of the proposed variable values
+//     - the configuration tree of the module in which the variable is
+//       declared
+//     - the path to the module (so we know which part of the tree to
+//       compare the values against).
+//
+// Currently since the type system is simple, we currently do not make
+// use of the values since it is only valid to pass string values. The
+// structure is in place for extension of the type system, however.
```
phinze: <3

phinze:
LGTM!

PR #6188, opened 2016-04-15
provider/aws: Better randomize the CloudTrail tests

catsby:
Introduce better randomization to our CloudTrail tests. This should help avoid some collisions I believe we are hitting in our night acceptance tests


phinze:
LGTM!

PR #5221, opened 2016-02-19
CloudFront distribution and origin access identity support

vancluever:
Reference: #3330 

I'm opening this new PR just to try and move this along. This branch is pretty much good to merge.

As mentioned in #3330, this is feature-complete CloudFront distribution resource, actually slightly more feature complete than the equivalent in CloudFormation (it doesn't seem like GZIP compression is supported yet, for example). Supports multiple origins, cache behaviours, and has a config layout like you would expect to have when sending the distribution configuration to API or CloudFormation.

I have also added `aws_cloudfront_origin_access_identity`, which allows origin access identities to be generated from Terraform rather than the console, definitely pushing this feature set past what CloudFormation has.

One other thing I'm waiting on is #5218, which will allow me to enforce a max instance of 1 on some of the complex types that need it (example: `default_cache_behavior`, `viewer_certificate`, etc). But if this is merged before that is I will just add these at a later time.

Again, if anyone wants to give it ago in a custom build, if you find anything that's an issue, let me know, otherwise the acceptance tests give a good spread of use cases.

PR #6196, opened 2016-04-15
provider/aws: CloudFront post-merge review updates

vancluever:
Hey guys,

Here are the updates to address the code review issues in #5221. Each commit has been pretty well annotated and each commit is kind of an issue in its own, but I can squash if need be.

The updates:

 * Fixed the hashing for ACM and IAM certificate parameters.
 * Added error handling to complex structure setting
 * Fixed primitives to pass pointers in instead of dereferenced values
 * Doc updates so that examples are properly indented
 * Removed delete retention to ensure that tests play well with Travis

### DIFF
--- a/builtin/providers/aws/cloudfront_distribution_configuration_structure.go
+++ b/builtin/providers/aws/cloudfront_distribution_configuration_structure.go
@@ -82,12 +82,20 @@ func expandDistributionConfig(d *schema.ResourceData) *cloudfront.DistributionCo
 // aws_cloudfront_distribution resource.
 //
 // Used by the aws_cloudfront_distribution Read function.
-func flattenDistributionConfig(d *schema.ResourceData, distributionConfig *cloudfront.DistributionConfig) {
-	d.Set("origins", flattenOrigins(distributionConfig.Origins))
+func flattenDistributionConfig(d *schema.ResourceData, distributionConfig *cloudfront.DistributionConfig) error {
+	var err error
+
 	d.Set("enabled", *distributionConfig.Enabled)
-	d.Set("default_cache_behavior", flattenDefaultCacheBehavior(distributionConfig.DefaultCacheBehavior))
-	d.Set("viewer_certificate", flattenViewerCertificate(distributionConfig.ViewerCertificate))
 	d.Set("price_class", *distributionConfig.PriceClass)
+
+	err = d.Set("default_cache_behavior", flattenDefaultCacheBehavior(distributionConfig.DefaultCacheBehavior))
+	if err != nil {
+		return err
+	}
+	err = d.Set("viewer_certificate", flattenViewerCertificate(distributionConfig.ViewerCertificate))
+	if err != nil {
+		return err
+	}
 
 	if distributionConfig.CallerReference != nil {
 		d.Set("caller_reference", *distributionConfig.CallerReference)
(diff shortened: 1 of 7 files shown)


builtin/providers/aws/cloudfront_distribution_configuration_structure.go
```diff
@@ -940,3 +974,28 @@ func flattenActiveTrustedSigners(ats *cloudfront.ActiveTrustedSigners) flatmap.M
 	m["items"] = s
 	return flatmap.Flatten(m)
 }
+
+// fullOK does a proper type-switch on an interface value and returns true
+// if the underlying type contains a zero value. float is ommitted as it is
```
catsby: > and returns true if the underlying type contains a zero value

It looks like it's returning `false` for default values

builtin/providers/aws/cloudfront_distribution_configuration_structure.go
```diff
@@ -940,3 +974,28 @@ func flattenActiveTrustedSigners(ats *cloudfront.ActiveTrustedSigners) flatmap.M
...
+// currently not a data type in Terraform and definitely not in CloudFront.
+func fullOK(v interface{}) bool {
```
catsby: This function seems pretty heavy handed... it only seems to be used with attributes that are strings, is there a situation where a simple check like so is insufficient? 

```go
if v, ok := m["attribute"]; ok && v.(string) != "" {
  [...]
}
```

I get that it checks types, but those attributes should all be `string`s right?

catsby:
Looks good! One question on the `fullOK` function and docs, other than that great work 😄

catsby:
Ah, looks like `TestAccAWSCloudFrontDistribution_S3Origin` gave me a panic/nil pointer dereference, line 94 of the test file

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x8 pc=0x311a93]

goroutine 20 [running]:
panic(0x1066c20, 0xc82000e140)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
testing.tRunner.func1(0xc820098a20)
	/usr/local/go/src/testing/testing.go:467 +0x192
panic(0x1066c20, 0xc82000e140)
	/usr/local/go/src/runtime/panic.go:426 +0x4e9
github.com/hashicorp/terraform/builtin/providers/aws.testAccCheckCloudFrontDistributionDestroy(0xc820358f30, 0x0, 0x0)
	/Users/clint/Projects/Go/src/github.com/hashicorp/terraform/builtin/providers/aws/resource_aws_cloudfront_distribution_test.go:94 +0x133
```

```go
-->		if *dist.DistributionConfig.Enabled != false {
			return fmt.Errorf("CloudFront distribution should be disabled")
		}
```

vancluever:
@catsby, checking and fixing now. I'll just remove `fullOK`, you are right, it's too heavy-handed for this job as this is currently the only function using it. If other things prove to be an issue in hashing, I might look at how to do this better in general. It'd be neat if `schema.ResourceData.GetOk()` worked for general maps, for example.

vancluever:
OK, all done. Also added a way to retain the disable-only functionality for tests via setting a `TF_TEST_CLOUDFRONT_RETAIN` environment variable - let me know your thoughts!

Also, found one more bug while I was at it - the delete waiter was not trapping error messages properly, so I've fixed that now (looks like it's taking abnormally long to disable distributions today).

catsby:
Thanks! Looks good here and the acc tests are passing, gonna pull this in!

vancluever:
Thanks @catsby!

PR #6197, opened 2016-04-15
provider/fastly: Add support for Request Headers

catsby:
Request Headers let you conditionally manipulate the Headers for various types of requests from users, backends, and Fastly's caches. Here we add `header` to the config, and allow users to specify headers to manipulate and how.

Example header block that delete the `http.x-amz-request-id` header that comes from a request served by AWS S3:

```hcl
  header {
    destination = "http.x-amz-request-id"
    type        = "cache"
    action      = "delete"
    name        = "remove x-amz-request-id"
  }
```

Reference:

- Fastly docs on [Adding or modifying headers on HTTP requests and responses][headers]
- Fastly API docs - https://docs.fastly.com/api/config#header

[headers]: https://docs.fastly.com/guides/basic-configuration/adding-or-modifying-headers-on-http-requests-and-responses

### DIFF
--- a/builtin/providers/fastly/resource_fastly_service_v1.go
+++ b/builtin/providers/fastly/resource_fastly_service_v1.go
@@ -494,7 +494,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 			remove := ohs.Difference(nhs).List()
 			add := nhs.Difference(ohs).List()
 
-			// Delete removed domains
+			// Delete removed headers
 			for _, dRaw := range remove {
 				df := dRaw.(map[string]interface{})
 				opts := gofastly.DeleteHeaderInput{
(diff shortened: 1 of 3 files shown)


builtin/providers/fastly/resource_fastly_service_v1.go
```diff
@@ -369,6 +474,60 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 			}
 		}
 
+		if d.HasChange("header") {
+			// Note: we don't utilize the PUT endpoint to update a Header, we simply
+			// destroy it and create a new one. This is how Terraform works with nested
+			// sub resources, we only get the full diff not a partial set item diff.
+			// Because this is done on a new version of the configuration, this is
+			// considered safe
```
phinze: Good note here! I think there are other APIs where we (still) have race conditions because of this behavior. Nice to have it explained out front here.

phinze:
Solid. LGTM!

Issue #5913, opened 2016-03-29
vSphere provider fails to find specified resource pool


trodemaster:
I'm attempting to use the vSphere provider to deploy VMs on a vSphere 6.0 cluster. However the VSPHERE_VIRTUAL_MACHINE argument resource_pool is not working as expected. Whenever I specify a resource pool the result is this error. 

Error applying plan:

1 error(s) occurred:

* vsphere_virtual_machine.centos6-vmware: resource pool 'Image_Lab' not found

Log file
https://gist.github.com/trodemaster/843ce9b16c6790e54ec9eee6e86b965a

.tf config file specifying resource pool
https://gist.github.com/trodemaster/4664509aed64919cba7f0009cf7853f0

Not sure what other details are going to be helpful in debugging this. Please advise...

trodemaster:
Terraform version..
Terraform v0.6.15-dev (d1ac7d3b2e1be2f32092275340f33a2d73f7b31a)

xantheran:
The vSphere edition would be helpful as well
Is there more than one datacenter in your vCenter server?  Also does is the account you login with have at least read access all the way down to the root?  I have seen similar problems when the vCenter server is shared and different groups have been assigned to sub folders under the datacenter

trodemaster:
This environment is 6.0u2. I have one datacenter and have named that in the config as well. The permissions config is very simple as this is a test lab. The user account is "administrator" from the root propagated down to all objects. Without specifying the resource pool name terraform is able to deploy to the cluster which is the parent object to the resource pool.

chrislovecnm:
Is this related?? https://github.com/hashicorp/terraform/pull/4100

trodemaster:
Yes does seem related to #4100 ! From my quick read seems like resource_pool does not currently work with clusters. If that's the case clearly a big omission. Outside of a lab I would say 99% of vSphere environments are a cluster.

tommymalmqvist:
<cluster_name>/Resources/<resource_pool_name>

chrislovecnm:
@tommymalmqvist thanks!! I assume that we should add that to the docs?

tommymalmqvist:
Good idea, i'll add it as an example.

chrislovecnm:
@tommymalmqvist cookie for u!! One thing I was thinking about is mentioning the command line VMware tool that the API provides. You want me to file an issue for doc improvements?

tommymalmqvist:
Sure, assign it to me.

[...]

PR #6199, opened 2016-04-15
fixes-#5913-vSphere-provider-fails-to-find-specified-resource-pool-do…

tommymalmqvist:
…cs: example extended with more arguments.

There is some confusion about the arguments and how to use them. Hopefully this will bring some clarity.


tommymalmqvist:
I'm closing this and I'll refer to another issue

PR #6203, opened 2016-04-16
Don't read back opsworks stack cookbooks source password

### LINKED ISSUE #6192: aws_opsworks_stack custom cookbook password keeps re-applying
### Terraform Version
v0.6.14

### Affected Resource(s)
- aws_opsworks_stacks

```hcl
##########################
## create opsworks stack
##########################

resource "aws_opsworks_stack" "test-stack" {
    name = "Test Stack"
    region = "us-east-1"
	color = "rgb(57, 131, 94)"
	configuration_manager_name = "Chef"
	configuration_manager_version = "11.10"
	service_role_arn = "${aws_iam_role.some_role.arn}"
    default_instance_profile_arn = "${aws_iam_instance_profile.some_profile.arn}"
	vpc_id = "12345"
	default_subnet_id = "12345"
	use_custom_cookbooks = true
	custom_cookbooks_source = {
		# please note I am using the S3 type not git
		type = "s3"
		url = "https://s3.amazonaws.com/some-bucket/some-file.tar.gz"
		username = "IAM User Key"
		password = "IAM User Secret"
	}
}
```

### Expected Behavior
Terraform should have created the stack and set a value to the password field. On subsequent plan runs Terraform should report
`No changes. Infrastructure is up-to-date.`.

### Actual Behavior
Terraform correctly creates the stack and sets the password. However on subsequent plan runs terraform reports
```

~ aws_opsworks_stack.test-stack
    custom_cookbooks_source.0.password: "*****FILTERED*****" => "IAM User Secret"


```
If I run an apply Terraform will correctly set the password. Terraform will continue to report that the resource needs to be updated.
After Terraform successfully sets the password If I try to comment out the password line in the TF file Terraform will try to set the password to an empty string.

```
~ aws_opsworks_stack.test-stack
    custom_cookbooks_source.0.password: "*****FILTERED*****" => ""
```

This issue is more than just a minor annoyance. Every time Terraform unnecessarily updates the stack s3 password it removes the custom JSON associated with that stack.
Which in turn causes any instances associated with the stack to fail.

### Steps to reproduce
1. Define a stack in a TF file
2. Run a plan/apply
3. Verify the stack was created and the s3 password was set
4. Run `terraform plan` and you should receive the unexpected output.

### References
- #3635 (same issue but for SSH keys maybe the same solution?)

apparentlymart:
As with several other sensitive values in Opsworks, the API returns a placeholder value rather than a nil. To avoid writing the placeholder value into the state we just skip updating the password on read, letting whatever value was in the state persist.

This means that Terraform can't detect configuration drift where someone has changed the password via some other means, but Terraform will still be able to recognize changes to the password made within Terraform itself due to the "last-written" value in the state.

This fixes #6192.

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_stack.go
+++ b/builtin/providers/aws/resource_aws_opsworks_stack.go
@@ -225,12 +225,12 @@ func resourceAwsOpsworksSetStackCustomCookbooksSource(d *schema.ResourceData, v 
 		if v.Username != nil {
 			m["username"] = *v.Username
 		}
-		if v.Password != nil {
-			m["password"] = *v.Password
-		}
 		if v.Revision != nil {
 			m["revision"] = *v.Revision
 		}
+		// v.Password will, on read, contain the placeholder string
+		// "*****FILTERED*****", so we ignore it on read and let persist
+		// the value already in the state.
 		nv = append(nv, m)
 	}
 
--- a/builtin/providers/aws/resource_aws_opsworks_stack_test.go
+++ b/builtin/providers/aws/resource_aws_opsworks_stack_test.go
@@ -329,6 +329,8 @@ resource "aws_opsworks_stack" "tf-acc" {
     type = "git"
     revision = "master"
     url = "https://github.com/aws/opsworks-example-cookbooks.git"
+    username = "example"
+    password = "example"
   }
 resource "aws_iam_role" "opsworks_service" {
     name = "%s_opsworks_service"


builtin/providers/aws/resource_aws_opsworks_stack.go
```diff
@@ -225,12 +225,12 @@ func resourceAwsOpsworksSetStackCustomCookbooksSource(d *schema.ResourceData, v
 		if v.Username != nil {
 			m["username"] = *v.Username
 		}
-		if v.Password != nil {
-			m["password"] = *v.Password
-		}
 		if v.Revision != nil {
 			m["revision"] = *v.Revision
 		}
+		// v.Password will, on read, contain the placeholder string
```
stack72: where does it have `*****FILTERED*****`?

stack72: AH! It comes back from the API :)

stack72:
LGTM! thanks @apparentlymart

eedwardsdisco:
@apparentlymart This is still happening in 0.6.16

```
will be destroyed.

~ aws_opsworks_stack.elasticsearch
    custom_cookbooks_source.0.password: "" => "<MY_IAM_PASSWORD (REDACTED)>"
```

apparentlymart:
@eedwardsdisco thanks for pointing that out. Looks like my fix here wasn't correct. I've opened a new issue #6826 to track the new bug.

PR #6205, opened 2016-04-16
provider/openstack: Notes for using OpenStack with Rackspace

### LINKED ISSUE #5695: Openstack/Rackspace: Can not retrieve per-network IPs when manually specified
Terraform version: 0.6.13

I'm using the Openstack provider on Rackspace with the following resources:

```
resource "openstack_compute_instance_v2" "explicit_network" {
  name = "explicit"
  region = "DFW"
  image_id = "fabe045f-43f8-4991-9e6c-5cabd617538c"
  flavor_id = "general1-4"
  key_pair = "provisioning_key"

  network {
    uuid = "11111111-1111-1111-1111-111111111111"
    access_network = true
  }

  network {
    uuid = "00000000-0000-0000-0000-000000000000"
  }
}
```

```
resource "openstack_compute_instance_v2" "implicit" {
  name = "implicit"
  region = "DFW"
  image_id = "fabe045f-43f8-4991-9e6c-5cabd617538c"
  flavor_id = "general1-4"
  key_pair = "provisioning_key"
}
```

As a little bit of background, the 2 explicit networks above are special networks in rackspace (0000... gives you a publicly routable IP, 1111... gives you a RFC1918 address for access to internal services).

When using the explicitly defined networks, access_ip_v4 is set to some bogus IP that I can't connect to and doesn't appear in the Rackspace interface anywhere.  network.0.fixed_ip_v4 and network.1.fixed_ip_v4 are both not set to anything, so there is no way to inspect the real returned public or private IP addresses.

When using the implicitly defined networks, terraform.tfstate only shows one network, and access_ip_v4 matches network.0.fixed_ip_v4.  The network that the entry corresponds to is non-deterministically either the internal or external network though.  If it returns the external network, there is no way to find the internal network, or vice versa.

There are IMO 2 acceptable ways to solve this problem:

* When using the networks implicitly, ensure that terraform returns both networks in a deterministic order.
* When using the networks explicitly, ensure that network.#.fixed_ip_v4 is populated for both networks and that access_network is honored.

I realize that Rackspace isn't directly compatible with Openstack, but work on the Rackspace-specific provider has halted and the solution to this seems like it might be feasible to work into the Openstack provider.

It also might be related to https://github.com/hashicorp/terraform/issues/5358

jtopjian:
Fixes #5695

### DIFF
--- a/website/source/docs/providers/openstack/index.html.markdown
+++ b/website/source/docs/providers/openstack/index.html.markdown
@@ -66,6 +66,48 @@ The following arguments are supported:
     service catalog. It can be set using the OS_ENDPOINT_TYPE environment
     variable. If not set, public endpoints is used.
 
+## Rackspace Compatibility
+
+Using this OpenStack provider with Rackspace is not supported and not
+guaranteed to work; however, users have reported success with the
+following notes in mind:
+
+* Interacting with instances has been seen to work. Interacting with
+all other resources is either untested or known to not work.
+
+* Use your _password_ instead of your Rackspace API KEY.
+
+* Explicitly define the public and private networks in your
+instances as shown below:
+
+```
+resource "openstack_compute_instance_v2" "my_instance" {
+  name = "my_instance"
+  region = "DFW"
+  image_id = "fabe045f-43f8-4991-9e6c-5cabd617538c"
+  flavor_id = "general1-4"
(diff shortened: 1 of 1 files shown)


stack72:
Nice addition @jtopjian 

:)

P.

PR #3480, opened 2015-10-11
[WIP] Grafana Provider, with Data Source and Dashboard resources

apparentlymart:
A provider for Grafana, including data source and dashboard resources.

* [x] Provider
* [x] `grafana_data_source` resource.
* [x] `grafana_dashboard` resource.
* [x] `grafana_dashboard_config` logical resource.
* [ ] `grafana_graph_panel_config` logical resource.
* [ ] `grafana_single_stat_panel_config` logical resource.
* [x] `grafana_text_panel_config` logical resource.
* [x] `grafana_dashboard_list_panel_config` logical resource.
* [ ] Documentation

Grafana uses a documented JSON-based format for importing and exporting dashboards and panels. It is expected that many users of this provider will design dashboards/panels interactively in the Grafana UI, export them as JSON, and then include them in a Terraform config via the `file(...)` function, and so the `grafana_dashboard` resource just accepts a raw JSON string for the dashboard configuration.

However, the provider also includes a number of logical resources to help in dynamically constructing dashboards and panels within the Terraform config itself, allowing the configuration to be more easily edited within the Terraform config and to contain interpolations so that, for example, the range of a particular graph can depend on the number of instances that were created for a particular resource.

PR #6206, opened 2016-04-17
[WIP] Grafana Provider, with Data Source and Dashboard resources

apparentlymart:
A provider for Grafana, including data source and dashboard resources.

* [x] Provider
* [x] `grafana_data_source` resource.
* [x] `grafana_dashboard` resource.
* [ ] Documentation

Grafana uses a documented JSON-based format for importing and exporting dashboards and panels. It is expected that many users of this provider will design dashboards/panels interactively in the Grafana UI, export them as JSON, and then include them in a Terraform config via the `file(...)` function, and so the `grafana_dashboard` resource just accepts a raw JSON string for the dashboard configuration.

This was originally submitted as #3480, including some additional resources to generate dashboard JSON from within Terraform configuration. However, experience trying to implement and use those suggested they could do with some more thinking, so this is a simplified PR that includes just the physical data source and dashboard resources.

### DIFF
--- a/builtin/bins/provider-grafana/main.go
+++ b/builtin/bins/provider-grafana/main.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"github.com/hashicorp/terraform/builtin/providers/grafana"
+	"github.com/hashicorp/terraform/plugin"
+)
+
+func main() {
+	plugin.Serve(&plugin.ServeOpts{
+		ProviderFunc: grafana.Provider,
+	})
+}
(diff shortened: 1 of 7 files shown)


apparentlymart:
This thing is so old that it predates the use of godep/vendoring. I need to do some more work here to get the vendored deps in there. My first attempt just made a huge mess, so I'll have another look at this later with fresh eyes. :grinning:

bigkraig:
:+1:

stack72:
Hi @apparentlymart 

so i have been messing with this provider this morning and it looks good :) I think we are in a situation where we can merge this. I have pinged the Grafana team to try and get them to test it out as well :)

Nice work!

P.

PR #6207, opened 2016-04-17
provider/openstack: static routing entries for routers

drebes:
Hi,

This is an attempt at adding support to static routes for OpenStack router resources. The trick here is that the static route entries are not first class citizens in OpenStack, but instead are created as an update to the routers once they are already attached to the subnets with the next hops.

I worked around that by creating a ``openstack_networking_router_table_v2`` resource with a ``router_id`` argument, which specifies which router to apply the rules to. Here's the example I'm using for the acceptance tests:

```
        resource "openstack_networking_router_v2" "router_1" {
                name = "router_1"
                admin_state_up = "true"
        }

        resource "openstack_networking_network_v2" "network_1" {
                        name = "network_1"
                        admin_state_up = "true"
        }

        resource "openstack_networking_subnet_v2" "subnet_1" {
                        network_id = "${openstack_networking_network_v2.network_1.id}"
                        cidr = "192.168.199.0/24"
                        ip_version = 4
        }

        resource "openstack_networking_port_v2" "port_1" {
                name = "port_1"
                network_id = "${openstack_networking_network_v2.network_1.id}"
                admin_state_up = "true"
                fixed_ip {
                        subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
                        ip_address = "192.168.199.1"
                }
        }

        resource "openstack_networking_router_interface_v2" "int_1" {
                        router_id = "${openstack_networking_router_v2.router_1.id}"
                        port_id = "${openstack_networking_port_v2.port_1.id}"
        }

        resource "openstack_networking_network_v2" "network_2" {
                        name = "network_2"
                        admin_state_up = "true"
        }

        resource "openstack_networking_subnet_v2" "subnet_2" {
                        network_id = "${openstack_networking_network_v2.network_2.id}"
                        cidr = "192.168.200.0/24"
                        ip_version = 4
        }

        resource "openstack_networking_port_v2" "port_2" {
                name = "port_2"
                network_id = "${openstack_networking_network_v2.network_2.id}"
                admin_state_up = "true"
                fixed_ip {
                        subnet_id = "${openstack_networking_subnet_v2.subnet_2.id}"
                        ip_address = "192.168.200.1"
                }
        }

        resource "openstack_networking_router_interface_v2" "int_2" {
                        router_id = "${openstack_networking_router_v2.router_1.id}"
                        port_id = "${openstack_networking_port_v2.port_2.id}"
        }

        resource "openstack_networking_router_table_v2" "router_table_1" {
                        depends_on = ["openstack_networking_router_interface_v2.int_1", "openstack_networking_router_interface_v2.int_2"]
                        router_id = "${openstack_networking_router_v2.router_1.id}"

                        route {
                                destination_cidr = "10.0.1.0/24"
                                next_hop = "192.168.199.254"
                        }

                        route {
                                destination_cidr = "10.0.2.0/24"
                                next_hop = "192.168.200.254"
                        }
        }
```

That seems to work, except for the case in which two ``openstack_networking_router_table_v2`` resources try to configure the same router (through the same ``router_id``). I'd assume that setting resource IDs consistently related to the router would allow the detection of conflicting resources (as they would have the same ID), but it doesn't seem to be the case.

The problem I'm having with the acceptance tasks is that the CheckDestroy function is called only once the whole configuration is destroyed, which includes the router to which the route entries are attached, which is more than what I want to check the deletion for. Is there a way I can limit that the CheckDestroy function is called after the destroy of the ``openstack_networking_router_table_v2`` resource, but still before the destruction of the other resources?

I'm also open to implement it in a different way, but I'm not sure how to do it, due to the way the lifecycle of the routes need to be managed. For that reason, I still haven't written any documentation, but plan to do so.

### DIFF
--- a/builtin/providers/openstack/provider.go
+++ b/builtin/providers/openstack/provider.go
@@ -1,9 +1,13 @@
 package openstack
 
 import (
+	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
+
+// This is a global MutexKV for use within this plugin.
+var osMutexKV = mutexkv.NewMutexKV()
 
 // Provider returns a schema.Provider for OpenStack.
 func Provider() terraform.ResourceProvider {
@@ -96,6 +100,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_networking_port_v2":             resourceNetworkingPortV2(),
 			"openstack_networking_router_v2":           resourceNetworkingRouterV2(),
 			"openstack_networking_router_interface_v2": resourceNetworkingRouterInterfaceV2(),
+			"openstack_networking_router_route_v2":     resourceNetworkingRouterRouteV2(),
 			"openstack_objectstorage_container_v1":     resourceObjectStorageContainerV1(),
 		},
 
--- a/builtin/providers/openstack/resource_openstack_networking_router_route_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_router_route_v2.go
@@ -1,0 +1,214 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/openstack/networking/v2/extensions/layer3/routers"
+)
+
+func resourceNetworkingRouterRouteV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNetworkingRouterRouteV2Create,
+		Read:   resourceNetworkingRouterRouteV2Read,
+		Delete: resourceNetworkingRouterRouteV2Delete,
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				DefaultFunc: schema.EnvDefaultFunc("OS_REGION_NAME", ""),
+			},
+			"router_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"destination_cidr": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"next_hop": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceNetworkingRouterRouteV2Create(d *schema.ResourceData, meta interface{}) error {
+
+	routerId := d.Get("router_id").(string)
+	osMutexKV.Lock(routerId)
+	defer osMutexKV.Unlock(routerId)
+
+	var destCidr string = d.Get("destination_cidr").(string)
+	var nextHop string = d.Get("next_hop").(string)
+
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(d.Get("region").(string))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	n, err := routers.Get(networkingClient, routerId).Extract()
+	if err != nil {
+		httpError, ok := err.(*gophercloud.UnexpectedResponseCodeError)
+		if !ok {
+			return fmt.Errorf("Error retrieving OpenStack Neutron Router: %s", err)
+		}
+
+		if httpError.Actual == 404 {
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error retrieving OpenStack Neutron Router: %s", err)
+	}
+
+	var updateOpts routers.UpdateOpts
+	var routeExists bool = false
+
+	var rts []routers.Route = n.Routes
+	for _, r := range rts {
+
+		if r.DestinationCIDR == destCidr && r.NextHop == nextHop {
+			routeExists = true
+			break
+		}
+	}
+
+	if !routeExists {
+
+		if destCidr != "" && nextHop != "" {
+			r := routers.Route{DestinationCIDR: destCidr, NextHop: nextHop}
+			log.Printf(
+				"[INFO] Adding route %s", r)
+			rts = append(rts, r)
+		}
+
+		updateOpts.Routes = rts
+
+		log.Printf("[DEBUG] Updating Router %s with options: %+v", routerId, updateOpts)
+
+		_, err = routers.Update(networkingClient, routerId, updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating OpenStack Neutron Router: %s", err)
+		}
+		d.SetId(fmt.Sprintf("%s-route-%s-%s", routerId, destCidr, nextHop))
+
+	} else {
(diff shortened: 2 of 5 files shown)


jtopjian:
@drebes This is really cool - thank you :smile: 

A couple questions / thoughts:

1) With regard to the acceptance test failures, I was able to fix it by making the `testAccCheckNetworkingV2RouterTableDestroy` function look like this:

```go
func testAccCheckNetworkingV2RouterTableDestroy(s *terraform.State) error {
  config := testAccProvider.Meta().(*Config)
  networkingClient, err := config.networkingV2Client(OS_REGION_NAME)
  if err != nil {
    return fmt.Errorf("(testAccCheckNetworkingV2RouterTableDestroy) Error creating OpenStack networking client: %s", err)
  }

  for _, rs := range s.RootModule().Resources {
    if rs.Type != "openstack_networking_router_table_v2" {
      continue
    }

    router, err := routers.Get(networkingClient, rs.Primary.Attributes["router_id"]).Extract()
    if err != nil {
      httpError, ok := err.(*gophercloud.UnexpectedResponseCodeError)
      if !ok {
        return fmt.Errorf("Error retrieving OpenStack Neutron Router: %s", err)
      }

      if httpError.Actual == 404 {
        // The router doesn't exist,
        // therefore the routing table is inherently destroyed
        return nil
      }
    }

    if router.ID != rs.Primary.Attributes["router_id"] {
      return fmt.Errorf("Router for table not found")
    }

    if len(router.Routes) != 0 {
      return fmt.Errorf("Invalid number of route entries: %d", len(router.Routes))
    }
  }

  return nil
}
```

2) With regard to multiple routing table resources that point to the same router... that sounds like a messy situation. Right now when you're deleting, you're setting the (shared) routing table's routes to nil. In the case where only one routing table is deleting, the shared routing table will be clobbered and erased. Upon the next read/refresh, the table should be updated, but there's probably going to be a window where the table was erased.

With that in mind, what are the use-cases of having multiple tables pointing to the same router? Is it safe to say that there should never be?

3) Will there ever be a time someone wants to interpolate or do any type of dynamic route configuration in a routing table? If so, consider breaking the `route` argument out into its own resource. `route` will then be a single route that points to an upstream router or routing table. For a historical example, look at the `openstack_lb_pool_v1` resource. It used to have a `member` argument until it was discovered that this does not work well when you create a dynamic number of instances and want to tie them back to a load balancer. To remedy, a `openstack_lb_member_v1` resource was created:

* https://www.terraform.io/docs/providers/openstack/r/lb_pool_v1.html#member
* https://www.terraform.io/docs/providers/openstack/r/lb_member_v1.html

4) With the above in mind, would it be easier to either add a new TypeSet argument of `route` to the router resource _or_ create a `openstack_networking_router_route_v2` resource that ties back to an upstream router? The CRUD on that resource would then only manage a single route entry rather than a whole table which might make management easier.

Thoughts?

drebes:
Thanks for the feedback.

1) This is OK to pass the test, but I don't think it would be actually testing the destroy functionality. Another approach I tested was to remove the CheckDestroy hook and replace it by a second ``resource.TestStep``, which would first apply an updated configuration without the route table entry, then run the RouterTableDestroy test. But for some reason the second step doesn't seem to run. I can update the PR with that for you to take a look, if you're interested.

2) I don't think there is any need for one to set two routing table resources on the same router, as on one resource one can already have all the route entries one need. It's more an issue of not allowing the user to do something stupid. Maybe just adding an explicit note on the documentation is enough, saying "don't do this".

3 &4) This actually makes sense, and would solve the problem with point 2 above. I'll take a look on how to do it that way.

Thanks

drebes:
Hi, regarding creating separate ``openstack_networking_router_route_v2`` resources for each routing entry, I'll need to query the existing routes from the router before creating or updating a specific entry, as the router update API only allows me to send them all at once. Is there some example on how I can set a mutex when talking to the OpenStack client so that the operations on two route resources on the same router will not overlap?

jtopjian:
Good catch. I recently implemented the following in the a Cobbler resource. It causes all resources of the same type to be processed one at a time rather than in parallel:

https://github.com/hashicorp/terraform/blob/master/builtin/providers/cobbler/resource_cobbler_system.go#L422-L423

Off the top of my head, this should also work here.

drebes:
I did the refactoring to split the route entries in their own resources: now multiple route resources shouldn't be a problem on the same router. I also fixed the acceptance tests to better test adding/deleting just the route entries. Let me know if it looks good this way and I'll update the documentation.

builtin/providers/openstack/provider.go
```diff
@@ -1,10 +1,14 @@
 package openstack
 
 import (
+	"github.com/hashicorp/terraform/helper/mutexkv"
```
jtopjian: Cool!

builtin/providers/openstack/resource_openstack_networking_router_route_v2.go
```diff
@@ -0,0 +1,216 @@
...
+func resourceNetworkingRouterRouteV2Create(d *schema.ResourceData, meta interface{}) error {
+
+	routerId := d.Get("router_id").(string)
+	osMutexKV.Lock(routerId)
+	defer osMutexKV.Unlock(routerId)
+
+	var destCidr string = d.Get("destination_cidr").(string)
+	var nextHop string = d.Get("next_hop").(string)
+
+	d.SetId(fmt.Sprintf("%s-route-%s-%s", routerId, destCidr, nextHop))
```
jtopjian: It's best to wait to `SetId` until the resource has been confirmed to be created.

builtin/providers/openstack/resource_openstack_networking_router_route_v2.go
```diff
@@ -0,0 +1,216 @@
...
+
+		_, err = routers.Update(networkingClient, routerId, updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating OpenStack Neutron Router: %s", err)
+		}
+	} else {
+		return fmt.Errorf("Route did not exist already")
+	}
+
+	d.SetId("")
```
jtopjian: I recently learned that it's no longer required to `SetId("")` any more. A successful return of the Delete function will do this automatically. A lot of resources still have it laying around, but newer ones are/can omitting it.

jtopjian:
This looks _really_ cool! I left a few comments for some small changes. Those + docs and this is good to go!

builtin/providers/openstack/resource_openstack_networking_router_route_v2.go
drebes: Would this apply in any case where the function returns a nil, or only on Delete? If so, I can remote those from the Read and Create error handling as well...

jtopjian:
@drebes Looks like the inline comments got clobbered.

> Would this apply in any case where the function returns a nil, or only on Delete? If so, I can remote those from the Read and Create error handling as well...

I think this only applies to Delete. If you'd like to play it safe, I'm totally fine leaving all `SetId("")` in place and it can be reviewed later on.

drebes:
I've removed the SetId("")s from delete, kept them in the other cases, and added the resource documentation. Should now be good to go.

jtopjian:
Looks good! Can you do one more thing and squash the commits into a single commit? This would prevent the original resource and refactor from being in the git history.

jtopjian:
@drebes I forgot to leave a comment when I merged. Thank you very much for this -- it's a great addition!

PR #6208, opened 2016-04-17
[WIP] Plugin patterns documentation

apparentlymart:
This new documentation section is intended to capture some patterns and best practices for the UX design of features provided by plugins, to give guidance to new plugin developers who may not yet have a lot of Terraform experience, and help maintain consistency between plugins.

The goal is to keep technical implementation discussion out of this section, and focus entirely on user experience.
- [x] Navigation structure
- [x] Index page with introduction and general advice
- [x] Provider patterns
- [x] Resource patterns
- [ ] Provisioner patterns

My motivation for writing this was that I see a lot of new providers and resource implementations being submitted and they often get caught up in "design review hell" due to back and forth about UX issues. UX discussions are very important and should not be eliminated entirely, but it seems like several ideas are often repeated during these reviews and so I wanted to try to capture some best practices in the documentation.

This stuff is of course naturally rather subjective. I've tried my best to cover things I've seen broad consensus on, but I have my own biases and I'm sure they've crept in here. It seems inevitable that this will also evolve over time as we learn more lessons about building plugins, so I figured it wasn't worth obsessing about getting every detail right for a first draft.

I'm also well aware of my limitations as a writer... having an editor look over this and tighten up my rambling blather would be nice, but I wanted to just get some of this down as a starting point to build on.


stack72:
ping @apparentlymart - thoughts on the progression of this? :)

apparentlymart:
@stack72 if the content already here seems good we could just drop the "Provisioner Patterns" link from the sidebar and merge it as-is.

From quickly skimming this again it looks like I didn't revise this to make the resource patterns talk about data resources after we merged that, so that's probably the main thing that could do with some more editing before calling this done.

I can finish this up if the team thinks this is a valuable thing and there's some good content here.

apparentlymart:
Not planning to work on this in the near future, so closing it out. Might pick it up again later.

Issue #6072, opened 2016-04-07
aws_route53_zone doesn't properly refresh state


maxrothman:
## Steps to reproduce

- Create an `aws_route53_hosted_zone` resource and define the `comment` argument
- `terraform apply`
- Via the AWS web console or the API, alter the comment of the zone
- `terraform plan`

## Expected result

The state file is updated with the altered comment and the plan indicates that it will update the comment

## Actual result

The state file is not updated with the altered comment and the plan indicates that there are no changes

apparentlymart:
The resource implementation is certainly *trying* to make this work, as evidenced by [the branch to update the Comment when set](https://github.com/hashicorp/terraform/blob/d750d4d1cccb08f194cdbc9484825f548be286d2/builtin/providers/aws/resource_aws_route53_zone.go#L174-176), so it seems that something more subtle is going on here than just a missing feature.

One possible bug I see there is that it only updates the comment when it's non-nil. Each AWS API handles "empty" a little differently, and if Route53 is one of those APIs that returns `nil` for an empty value then it looks like this code would fail to update the comment if it were set to *the empty string* in the AWS console, but that shouldn't prevent it from refreshing when it's non-empty.

Seems like this one is going to require some digging to figure out.

stack72:
Hi @maxrothman 

Apologies that we haven't gotten back to you sooner on this. I am not sure this is an issue anymore - I certainly cannot recreate it i'm afraid. I created the following config - using Terraform 0.7.4 (latest!)

```
provider "aws" {
    region = "us-west-2"
}

resource "aws_route53_zone" "main" {
       	name = "hashicorp.com."
       	comment = "Custom comment"

       	tags {
       		foo = "bar"
       		Name = "tf-route53-tag-test"
       	}
}
```

A simple Terraform apply works as follows:

```
% terraform apply                                                                                                                                           ✭
aws_route53_zone.main: Creating...
  comment:        "" => "Custom comment"
  force_destroy:  "" => "false"
  name:           "" => "hashicorp.com."
  name_servers.#: "" => "<computed>"
  tags.%:         "" => "2"
  tags.Name:      "" => "tf-route53-tag-test"
  tags.foo:       "" => "bar"
  vpc_region:     "" => "<computed>"
  zone_id:        "" => "<computed>"
aws_route53_zone.main: Still creating... (10s elapsed)
aws_route53_zone.main: Still creating... (20s elapsed)
[...]

PR #6209, opened 2016-04-17
aws_route53_zone doesn't properly refresh state

sailendhran:
Issue #6072

### DIFF
--- a/builtin/providers/aws/resource_aws_route53_zone.go
+++ b/builtin/providers/aws/resource_aws_route53_zone.go
@@ -33,6 +33,7 @@ func resourceAwsRoute53Zone() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "Managed by Terraform",
+				ForceNew: true,
 			},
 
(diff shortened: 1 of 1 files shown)


apparentlymart:
@sailendhran, thanks for this!

Based on my understanding of the issue described in #6072, I don't think this is the right fix. `ForceNew` controls whether we update an existing instance or replace it with a new one when a change is detected, but it seems that for this bug the change is not being detected in the first place, and so we don't get far enough for `ForceNew` to be considered.

For bugs where external changes don't get reflected in the Terraform state, the bug will most often be in the `Read` implementation of the resource, since it is this function that is responsible for retrieving the latest data from the external service and updating the state to match it.

If this change appeared to fix things for you, it'd be helpful to see the output from your `terraform plan` (sanitized to avoid including any secrets) to compare it with what @maxrothman reported, since he apparently got a plan that was empty altogether.

catsby:
Hey Friends – I'm going to agree with @apparentlymart here, I don't believe recreating the resource is a good idea here. Thanks for contributing though!

PR #6218, opened 2016-04-18
Remove empty software dependencies list in README

julienvey:
In d21b0897a9d8d45c7c178ffd0c8713a6e1babae0, the list of needed software packages was removed, but the previous sentence was not, which doesn't make sense anymore.


phinze:
Oop good catch - thanks!

PR #3061, opened 2015-08-23
Various ECS bugfixes (IAM, destroy timeout)

radeksimko:
This fixes couple bugs:

 - #2869 
 - #2722 
 - #2902 - it's more of a work around, but I currently can't think of any easy way to fix it

PR #6225, opened 2016-04-18
Use resource.Retry for route creation and deletion

RevCBH:
Retry logic for creation/deletion of `aws_route` resources. The retry structure is cribbed from https://github.com/hashicorp/terraform/pull/3061/files. Not sure if there's a good way to test this, but it appears to correct the intermittent failure I was seeing when creating routes across peered VPCs.

### DIFF
--- a/builtin/providers/aws/resource_aws_route.go
+++ b/builtin/providers/aws/resource_aws_route.go
@@ -155,11 +155,10 @@ func resourceAwsRouteCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Route create config: %s", createOpts)
 
 	// Create the route
-	var out *ec2.CreateRouteOutput
 	var err error
 
 	err = resource.Retry(2*time.Minute, func() *resource.RetryError {
-		out, err = conn.CreateRoute(createOpts)
+		_, err = conn.CreateRoute(createOpts)
 
 		if err != nil {
 			ec2err, ok := err.(awserr.Error)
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_route.go
```diff
@@ -152,7 +155,27 @@ func resourceAwsRouteCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Route create config: %s", createOpts)
 
 	// Create the route
-	_, err := conn.CreateRoute(createOpts)
+	var out *ec2.CreateRouteOutput
```
catsby: it doesn't look like we use `out` here; it's used in `conn.CreateRoute` but then never after that, safe to remove?

catsby:
Looks solid! Can you give me an example of an `InvalidParameterException` type error that would happen in CREATE that would resolve with retries? 

Thanks!

builtin/providers/aws/resource_aws_route.go
RevCBH: Yeah, good catch, was just an artifact of copy/pasting.

RevCBH:
I believe the specific case I was running into was due to eventual consistency with route table creation (so RouteTableId) was invalid. The weird part was that when I went to `terraform apply` a second time, it would fail because the route already existed (i.e. it got created even though the API reported a failure).

I'm not 100% sure that this is what was happening, but it looked very much like a race-condition/eventual-consistency issue. I was seeing failures about 80% of the time without this change. Sorry for the voodoo-style analysis, but I wasn't able to get a lot of debugging information about the error. Since running with the change locally, I haven't seen the failure over the course of 10s of runs.

catsby:
That's cool, I was just curious if you knew off hand what kind of things would happen there

catsby:
acc tests pass, merging, thanks!

PR #6231, opened 2016-04-18
deps: Update call sites of hil.Eval from update

jen20:
hil.Eval() now returns (hil.EvaluationResult, error) instead of (value, type, error). This commit updates the call sites, but retains all previous behaviour. Tests are also updated.


jen20:
This is effectively a wrapper - so assuming Travis is green I think we a GTG with merging with minimal risk.

PR #5330, opened 2016-02-25
Make use of ComputedWhen in aws_s3_bucket_object

radeksimko:
At the moment if one tries to use `aws_s3_bucket_object` with versioning enabled on that bucket, it won't work quite as expected, because the `version_id` is updated after `apply`.

This limitation makes `aws_s3_bucket_object` not very handy when used in combination with `aws_lambda_function.s3_object_version` (mapped to `aws_s3_bucket_object.version_id`).

The full Lambda Function update cycle when used w/ `aws_s3_bucket_object` therefore requires two `apply` steps.

PR #4846, opened 2016-01-27
[WIP] helper/schema: implement ComputedWhen

phinze:
ComputedWhen is a long `TODO`d schema feature that allows any field to
be marked as computed when some target field changes.

Right now Terraform has no way of invalidating a field until it sees
_that_ field changes, but there are plenty of examples of scenarios
where one field changing implies that another field is going to be
assigned a new value.

I'd like to do a bit more testing and refactoring before this is fully ready. But I wanted to push what I had for now since it seems to be working for the cases I've tried it.

This should improve the experience of using consul_keys and atlas_artifacts (for scenarios where two plan/apply cycles would be required for a change to propagate), as well as any other resources where we find similar scenarios that we'd like to use `ComputedWhen`.

PR #6233, opened 2016-04-19
[WIP]: Basic support for aws_lambda_function publish attribute

kmamykin:
Looking for review.

Not ready to merge yet. 

* [ ] test with publish = false for backward compatibility
* [ ] test with aliases
* [ ] test with permissions
* [ ] test with mappings

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -63,6 +63,11 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Optional: true,
 				Default:  128,
 			},
+			"publish": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 			"role": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
@@ -106,6 +111,10 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Computed: true,
 			},
 			"last_modified": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"version": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
 			},
@@ -159,6 +168,7 @@ func resourceAwsLambdaFunctionCreate(d *schema.ResourceData, meta interface{}) e
 		FunctionName: aws.String(functionName),
 		Handler:      aws.String(d.Get("handler").(string)),
 		MemorySize:   aws.Int64(int64(d.Get("memory_size").(int))),
+		Publish:      aws.Bool(d.Get("publish").(bool)),
 		Role:         aws.String(iamRole),
 		Runtime:      aws.String(d.Get("runtime").(string)),
 		Timeout:      aws.Int64(int64(d.Get("timeout").(int))),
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_lambda_function.go
```diff
@@ -356,6 +342,45 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
...
+	}
+
+	log.Printf("[DEBUG] Send Update Lambda Function Code request: %#v", codeReq)
+	if codeUpdate {
+		updateFunctionCodeOutput, err := conn.UpdateFunctionCode(codeReq)
+		if err != nil {
+			return fmt.Errorf("Error modifying Lambda Function Code %s: %s", d.Id(), err)
+		}
+
+		d.Set("version", updateFunctionCodeOutput.Version)
```
kmamykin: Do I need to SetPartial for version here as well?

catsby: Looks right to me

catsby:
Thanks @kmamykin – I edited the title to note the WIP status, just remove that and ping here when you're ready! Let us know if you have any questions

sinangedik:
Could this be merged in?

kmamykin:
Well, the publish=true is working. However there is still a problem: When a new function is deployed with publish=true, the resource is updated with a new version and ARN (calculated attributes), but any dependent resources that depend on function's ARN are not updated. `terraform plan` shows that the function itself with new source_code will be deployed, but no updates calculated for dependent resource.

That looks like its a more general problem with terraform rather then an issue with this PR. But I think merging this PR as will confuse people. 

Any idea how to control dependencies on calculated properties?

When I get a bit more time, I will post cmd output.

radeksimko:
> That looks like its a more general problem with terraform rather then an issue with this PR.

Agreed.

> But I think merging this PR as is will confuse people.

Agreed.

> Any idea how to control dependencies on calculated attributes?

@kmamykin Unfortunately I don't think there's any way to do that until `CalculatedWhen` becomes a real thing.

See https://github.com/hashicorp/terraform/pull/4846 and https://github.com/hashicorp/terraform/pull/5330

radeksimko:
> But I think merging this PR as is will confuse people.

It would if we merged the current implementation, but I think there's a chance of using local file instead of S3 for the test purposes - that way we can get around the issue and it shouldn't confuse people too much to prevent this PR from merging.

That said the confusing behaviour w/ S3 should probably be documented.

radeksimko:
Hi @kmamykin 
I almost forgot about your PR 😱  and submitted https://github.com/hashicorp/terraform/pull/8653 Sorry!

---

As you mentioned in the code all published Lambda versions are immutable. Changes will always be applied to the latest version.

You're setting `Qualifier` in your changes to `GetFunctionInput` which (I assume) is because you want to get & reset the ARN to a qualified one - otherwise the response is pretty much same as response from `GetFunctionInput` without `Qualifier`.

^ The (potentially negative) side effect is that if someone deployed a new version outside of terraform it would remain unnoticed which is contrary to how most other TF resources behave today.

In addition to that qualified ARNs and unqualified ones aren't interchangeable even if they point to the same version/qualifier. Unqualified ARNs in policies are only valid for invokes via unqualified ARNs. See http://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html

> If the qualifier is not specified, the permission is valid only when requests is made using unqualified function ARN.

This leads me to a conclusion that we should output both qualified and unqualified ARN and note that these are mutually exclusive - i.e. user should use either one or the other.

kmamykin:
@radeksimko totally fine, thank you for picking it up and bringing it home. My PR was more like an experiment to see if I would make it work in my CI workflow.
Cheers.

radeksimko:
#8653 merged and will be included in the next release.

@kmamykin Thanks for the contribution.

PR #6013, opened 2016-04-05
Internal Plugins

cbednarski:
This is a port of a similar feature in packer: https://github.com/mitchellh/packer/pull/2608

Instead of compiling a separate binary for each provider / provisioner, all of the plugins are now compiled into the main `terraform` binary and invoked via the `internal-plugin` command. This is invisible to the user and happens automatically.

The change to the plugin system is designed to search for traditional, separate plugin binaries first, which means you can still build custom plugins or even override the internal plugin behavior.

There is also a code generator to automatically compile in new plugins as they are added. I project, though I have not tested, that it also should be faster to dev/test by commenting out parts of `command/internal_plugin_list.go` that you aren't working on. When I was developing this branch I used a subset of plugins and my `go build` times were around 23 seconds.

### Why do this?

There is a lot of code duplicated between the individual plugin binaries. Compiling a single binary which includes everything saves space and takes less time to compile. Here are some numbers:

#### On the current 0.7 dev branch

```
(dev-0.7) 0$ time make dev
...
real	3m14.365s
user	20m28.031s
sys	1m57.843s

(dev-0.7) 0$ du -hs bin/
751M	bin/
```

#### On the new internal-plugins branch

```
(f-internal-plugins) 0$ time make dev
...
real	0m59.019s
user	3m32.701s
sys	0m18.518s

(f-internal-plugins) 0$ du -hs bin/
130M	bin/
```

#### Summary

In summary there is approximately 70% reduction in time to run `make dev` and approximately 80% reduction in the size of the binary/ies. These gains should also carry into the cross-compilation / release process.

I also compared times for `make`, which runs tests. There is no noticeable difference in time here, though there is a slight difference in behavior. Please try out your normal dev/test workflows and let me know if you run into any slowdowns.

Also since this is mostly toolchain changes I'm not sure of the appropriate way to test this; I'm open to suggestions for this.

PR #6238, opened 2016-04-19
Reduce the size of the binaries

sorenmat:
Introducing an `TF_RELEASE` variable to the build process.
When `TF_RELEASE` is set, the LD_FLAGS `-s -w` is applied. This will strip debug information from
the binaries.
The complete size of the binaries before change was 543 mb
After the change the complete size is 395
This is a decrease of __148 mb__
Numbers are taken from OSX.

This has not been tested on other platform then OSX, but since it's standard go build flags I don't foresee any issues with it.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -35,12 +35,18 @@ if ! which gox > /dev/null; then
     go get -u github.com/mitchellh/gox
 fi
 
+LD_FLAGS="-X main.GitCommit=${GIT_COMMIT}${GIT_DIRTY}"
+# In relase mode we don't want debug information in the binary
+if [[ -n "${TF_RELEASE}" ]]; then
+    LD_FLAGS="-X main.GitCommit=${GIT_COMMIT}${GIT_DIRTY} -s -w"
+fi
+
 # Build!
 echo "==> Building..."
 gox \
     -os="${XC_OS}" \
     -arch="${XC_ARCH}" \
-    -ldflags "-X main.GitCommit=${GIT_COMMIT}${GIT_DIRTY}" \
+    -ldflags "${LD_FLAGS}" \
     -output "pkg/{{.OS}}_{{.Arch}}/terraform-{{.Dir}}" \
     $(go list ./... | grep -v /vendor/)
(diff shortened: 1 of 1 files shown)


apparentlymart:
@sorenmat, thanks for this!

I tried it on my Linux machine, building binaries for all supported targets, and it seemed to work across all of them, so I merged it. I also tweaked the `Makefile` so that this will be done by default when making distribution packages.

I am wondering whether this will mean that when users submit bug reports about crashes they'll be less useful, because we'll lose the source information. My gut is to try it and see what happens, since the savings are quite compelling, but I'm pinging @phinze in case he disagrees and wants to roll back my follow up change a1530855a3a762661dd1d78ca28fdfddc32e797f and leave this as an optional flag for people making custom distros.

#6013 should also make a considerable impact on this, by avoiding duplicating the code (and debug info!) for some core bits of Terraform repeatedly in each binary.

radeksimko:
It would be good to see the difference between crash log from standard binary vs stripped version of binary before we actually ship any stripped binaries off to the wild.

Also see https://github.com/hashicorp/terraform/pull/6013 which is reducing the size probably even more radically without introducing the tradeoffs mentioned (`715 - 130 = 585 MB` saved), so such stripping may not be necessary.

Having spent quite some time trying to reproduce weird bugs in different AWS environments I'm absolutely 👎  for any attempts of making this more difficult - i.e. I'm 👍  for rolling back this change, but I'll leave the decision on @phinze

sorenmat:
The output of a panic will be the same. I don't think there currently is a way to strip that. Not that I think we should.
So this basically means you can't debug the binary with gdb or similar, and I guess the use case for that is quite limited. If you have that usecase then you would have to rebuild terraform with debug information, I would personally be fine with that.

radeksimko:
> So this basically means you can't debug the binary with gdb or similar

right, I see. That makes me feel less worried. I was mostly worried about the crash logs.

I guess the ability to debug released binaries via gdb would be nice to have, but it's not a common task we would ask user to perform.

PR #6243, opened 2016-04-19
[WIP] aws/provider Codify S3 region support

u2mejc:
Adds region support to s3 resource. 

Presently AWS/S3 responds with a 307 for this request. I'd like some feedback on how best to use the helper to connect to a different region than specified in the shell environment, without changing it for the rest of the apply run for other resources.

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket.go
@@ -197,11 +197,14 @@ func resourceAwsS3Bucket() *schema.Resource {
 func resourceAwsS3BucketCreate(d *schema.ResourceData, meta interface{}) error {
 	s3conn := meta.(*AWSClient).s3conn
 	awsRegion := meta.(*AWSClient).region
+	// check if region is specified
+	if v, ok := d.GetOk("region"); ok {
+		awsRegion = v.(string)
+	}
 
 	// Get the bucket and acl
 	bucket := d.Get("bucket").(string)
 	acl := d.Get("acl").(string)
-	region := d.Get("region").(string)
 
 	log.Printf("[DEBUG] S3 bucket create: %s, ACL: %s", bucket, acl)
 
@@ -209,12 +212,8 @@ func resourceAwsS3BucketCreate(d *schema.ResourceData, meta interface{}) error {
 		Bucket: aws.String(bucket),
 		ACL:    aws.String(acl),
 	}
-	if *aws.String(region) != "" {
-		req.CreateBucketConfiguration = &s3.CreateBucketConfiguration{
-			LocationConstraint: aws.String(region),
-		}
-	} else if awsRegion != "us-east-1" {
-		// Special case us-east-1 region and do not set the LocationConstraint.
+	if awsRegion != "us-east-1" {
+		// If us-east-1 region, do not set the LocationConstraint.
 		// See "Request Elements: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html
 		req.CreateBucketConfiguration = &s3.CreateBucketConfiguration{
 			LocationConstraint: aws.String(awsRegion),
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_s3_bucket.go
```diff
@@ -201,17 +201,21 @@ func resourceAwsS3BucketCreate(d *schema.ResourceData, meta interface{}) error {
...
-	// Special case us-east-1 region and do not set the LocationConstraint.
-	// See "Request Elements: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html
-	if awsRegion != "us-east-1" {
+	if *aws.String(region) != "" {
+		req.CreateBucketConfiguration = &s3.CreateBucketConfiguration{
+			LocationConstraint: aws.String(region),
+		}
+	} else if awsRegion != "us-east-1" {
+		// Special case us-east-1 region and do not set the LocationConstraint.
+		// See "Request Elements: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html
```
catsby: I'm not sure this logic is doing what we'd like... if I set my region to `us-east-1` in the configuration of my S3 Bucket, won't `if *aws.String(region) != "" {` evaluate to `true` and then we'll add a the `LocationConstraint` that we don't want to add?

It does make sense though to read the `region` attribute from the configuration and not just use the default connection's region though, if you'd like to make a change like so:

```go
awsRegion := meta.(*AWSClient).region
// check if region is specified
if v,ok := d.GetOk("region"); ok {
	awsRegion = v.(string)
}

// Special case us-east-1 region and do not set the LocationConstraint.
// See "Request Elements: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html
if awsRegion != "us-east-1" {
	[...]
}
```

Would that accomplish what you're expecting?

u2mejc: Hi Clint, thank you for replying here too! Your block looks good too and I'll update to this PR to use it.

The last peace to the puzzle before I remove [wip] to merge:
[`Regions outside of us-east-1 require the appropriate LocationConstraint to be specified in order to create the bucket in the desired region`](http://docs.aws.amazon.com/cli/latest/reference/s3api/create-bucket.html)

To clarify that, the default region is us-east-1. In the api debug, you can see the PUT URL doesn't need to include the region:
```http
PUT http://my-tf-test-bucket.s3.amazonaws.com/ HTTP/1.1
*empty body*
```
Compared to other regions:
```http
PUT http://my-tf-test-bucket.s3-us-west-1.amazonaws.com/ HTTP/1.1
*snip*
<CreateBucketConfiguration><LocationConstraint>us-west-1</LocationConstraint></CreateBucketConfiguration>
```

Adding a LocationConstraint of us-east-1 to `<bucket>.s3.amazonaws.com` (the implicit us-east-1), returns a 400. (that's the "Special case" seen in the code comments.) 
Specifying a region in the codified resource, other than what is set by `AWS_DEFAULT_REGION`, returns a `HTTP/1.1 307 Temporary Redirect`.

I believe the issue is the API URL is being set in the AWS golang SDK object by being initialized, `client.s3conn = s3.New(sess)` with the region from `AWS_DEFAULT_REGION`. This would break the ability to change the LocationConstraint to anything else.

Does that make sense? If so, what is the best way to restart the client session if we run into a resource region outside that specified by `AWS_DEFAULT_REGION`?

u2mejc:
I looked at trying to restart the s3 client connection session to handle the 307 responce, but I couldn't find a good example in the existing codebase that I thought the team would merge (I could have missed it). 

Since I'm not feeling good traction on wrapping my head around the issue, I want to make sure that anyone who wants to work on the issue isn't turned away by seeing my open WIP, I'm going to close this for now. Please feel free to still ask me any questions about the issue, I'm happy to help.

PR #6247, opened 2016-04-19
provider/fastly: Add Gzip rule support

catsby:
Provides new attribute `gzip`, a set of gzip rules to apply to file types or extensions, or both. 

I've opened a support case with Fastly regarding their handling of empty `extensions` and `content_types` parameters. Right now, if you send only one, the other will come back with a set of default values. Ex, sending just extension `css` will return a rule with just the `css` extension, but a default list of `content_types`... 

For now we work around this while I wait to hear what they say about that, e.g. is it a bug that will be fixed or just how things work, and we can work around that when we hear back.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1173,7 +1173,7 @@
 		},
 		{
 			"ImportPath": "github.com/sethvargo/go-fastly",
-			"Rev": "382fee1e5e1adf3cc112fadf4f0a8a98e269ea3c"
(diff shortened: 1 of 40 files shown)


stack72:
I reviewed only the commit that adds support to the resource - LGTM! I am presuming the new test case runs work as expected? I don't have a fastly account :) Would it be worth adding a test for flattenGzips?

catsby:
@stack72 yah good call, added in aefd839

stack72:
@catsby - in that case - this looks good :) Thanks for the test addition!

PR #6255, opened 2016-04-20
provider/triton: fixed triton key_material documentation

### LINKED ISSUE #6211: triton provider key_material does not accept a file as the docs say
Version: current stable

As you can see in this [PR](https://github.com/hashicorp/terraform/pull/6090/files) `key_file` was changed in favor of `key_material`; however, it was does not work with a file. The private key value is passed to create a new auth object [here](https://github.com/hashicorp/terraform/blob/d1ac7d3b2e1be2f32092275340f33a2d73f7b31a/vendor/github.com/joyent/gosign/auth/auth.go#L54), but since it's a file name it breaks.

jhorwit2:
Completely fixes #6211 

[d72bb52](https://github.com/hashicorp/terraform/commit/d72bb52ad4d70908378a5da0dc7d945a9d0c8a88) partially fixed the ticket but missed the explanation for key_material.

### DIFF
--- a/website/source/docs/providers/triton/index.html.markdown
+++ b/website/source/docs/providers/triton/index.html.markdown
@@ -30,6 +30,6 @@ provider "triton" {
 The following arguments are supported in the `provider` block:
 
 * `account` - (Required) This is the name of the Triton account. It can also be provided via the `SDC_ACCOUNT` environment variable.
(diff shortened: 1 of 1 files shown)


phinze:
Whoops - nice catch, and thanks for the fix! 👍

PR #6261, opened 2016-04-20
provider/fastly: fix the description for "weight" on fastly_service_v1.

mattmoyer:
This description is essentially copied from the Fastly dashboard.


stack72:
Hi @mattmoyer

Thanks so much for the PR - just been looking at the [fastly docs](https://docs.fastly.com/api/config#backend) and agree that this is a better representation than what we had. I think that may have been a copy / paste error

Thanks

Paul

PR #6262, opened 2016-04-20
ID-only tests for all resources

mitchellh:
This modifies the acceptance test framework to do what is called an **ID-only refresh test** on resources. This test modifies the state to only have the ID of the resource, performs a refresh, and verifies it results in equivalent attribute state.

This is going to result in _a lot_ of test failures. Therefore, I've guarded it with an env var for now while I run it manually and fix things up over time. We can open the floodgates hopefully shortly.

No tests have to change to enable this, it'll automatically be available for all existing acceptance tests. There isn't a flag to disable yet per-resource but I have a feeling that'll be needed. I'll do that when it comes since for now its feature flagged off anyways.

**Why?** This verifies that `Refresh` on a resource is complete. This is a pre-req to us supporting importing at some point in the future.

Example of a failure:

2016-04-20 at 9 33 am


mitchellh:
Thanks! Fixing some test failures then will merge.

apparentlymart:
What are we to do about write-only attributes on resources? There are plenty of these sprinkled across various providers. I don't see any way this could ever be automatically supported... user intervention will always be required in order to support import for these.

mitchellh:
@apparentlymart I'll add a flag to disable the test for those.

PR #6275, opened 2016-04-21
Updated notify_no_data default to true

mootpt:
This value should be true, based on: https://github.com/hashicorp/terraform/blob/master/builtin/providers/datadog/resource_datadog_monitor.go#L77


stack72:
Hi @mootpt 

nice find here :) Thanks

P.

PR #6277, opened 2016-04-21
Replace flatmaps with HIL-native maps

jen20:


jen20:
Ignore this, it's broken.

PR #6278, opened 2016-04-21
urlencode interpolation function

protomouse:
This PR adds `urlencode(string)`, which calls to `url.QueryEscape`.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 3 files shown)


stack72:
Hi @protomouse 

Thanks so much for the PR here. Just curious as to what your usecase on this is? Is it for use with a private cloud provider or something? :)

Paul

protomouse:
@stack72 Without going into the specifics, I needed to configure a third-party proxy with an unsafe value generated by Terraform. I ended up doing things a bit differently in the end.

Issue #3966, opened 2015-11-18
vsphere provider does not delete vm if machine is powered off


pandrew:
Im trying to create a vm using the following information 

```
# Configure the vSphere Provider
provider "vsphere" {
  user           = "root"
  password       = "password"
  vcenter_server = "ip"
}

# Create a virtual machine
resource "vsphere_virtual_machine" "web" {
  name   = "terraform_web"
  vcpu   = 2
  memory = 4096

  network_interface {
    label = "VM Network"
  }

  disk {
    size = 1
    iops = 500
  }
}
```

using ESXI 6.0.0 2809111
terraform build: 0.6.7-dev 24ee56326d1b7eccda046524fe7129302d6556fd

The vm is created but does not powered on and upon deleting with terraform plan -destroy, terraform plan apply it is not being deleted. It fails by informing me that it cant be destroyed while being powered off and looking at the code it seems as it assumes it is powered on and therefore cant shutdown while already powered off.

chrislovecnm:
@tkak / @pietern can we destroy a powered off vm with govomi?  I can test when I have bandwidth.  

@apparentlymart  this is as designed, and would be a new feature.  A feature that we should have eventually, but it is based on how the govomi api will react.

pietern:
@chrislovecnm Yes, no problem. This must be something where it tries to power off first, which fails if it is already powered off. You have to check for this if you're using the bare govmomi API.

chrislovecnm:
@pietern THANKS!  Yah this is the way it is coded, and it would be good to have feedback from @tkak on what he thinks

chrislovecnm:
@apparentlymart can you change this to a feature request?  It is functioning as designed...

pandrew:
@chrislovecnm What do i need to do to make it boot in the first place? There´s no logic for to handle both powered on and off states when destroying the plan.

chrislovecnm:
@pandrew 'make it boot in the first place' - what do you mean?

pandrew:
@chrislovecnm in my case the example above does not power on the vm in esxi. I assume terraform would do that after it had created the vm. The only option i found in the docs was a boot_delay option.

chrislovecnm:
@pandrew we are calling poweron https://github.com/hashicorp/terraform/blob/master/builtin/providers/vsphere/resource_vsphere_virtual_machine.go#L1138 when you create vm from a template.

[...]

PR #6283, opened 2016-04-21
Vsphere - Checking for powered off state before deletion

aheeren:
Fixes https://github.com/hashicorp/terraform/issues/3966


stack72:
Hi @aheeren 

this looks like a great addition - nice work!

Paul

PR #6284, opened 2016-04-21
adding a * catchall for var interpolation

mzupan:
This adds the following functionality for #2312

```
variable "instance_type" {
  default = {
    prod = "m4.xlarge"
    "*" = "m4.large"
  }
}
```

It allows a catch all. The `*` needs to be quotes since terraform thinks its an invalid character. 

I tried to edit the test but my knowledge of test writing in go is bad at best. This passes all my tests that I tried.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -472,9 +472,14 @@ func interpolationFuncLookup(vs map[string]ast.Variable) ast.Function {
 			k := fmt.Sprintf("var.%s.%s", args[0].(string), args[1].(string))
 			v, ok := vs[k]
 			if !ok {
-				return "", fmt.Errorf(
-					"lookup in '%s' failed to find '%s'",
-					args[0].(string), args[1].(string))
+				k = fmt.Sprintf("var.%s.*", args[0].(string))
+				v, ok = vs[k]
+				
+				if !ok {
+					return "", fmt.Errorf(
+						"lookup in '%s' failed to find '%s'",
+						args[0].(string), args[1].(string))
+				}
 			}
 			if v.Type != ast.TypeString {
 				return "", fmt.Errorf(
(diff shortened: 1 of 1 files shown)


jen20:
Hi @mzupan! Thanks for the effort involved in a PR here. I don't personally have an opinion here on whether a catch all lookup is a good idea (perhaps @mitchellh or @phinze do), but the implementation will need some rework because of changes to how maps work on the `dev-0.7` branch. If we go ahead with this I'm happy to port your work over to that branch for 0.7. Thanks!

mitchellh:
My initial reaction is: good idea! 😄 

I didn't look at the PR (I'll leave that to @jen20 since he is most familiar with these code paths right now), so I'll just talk about high levels, UX, any pitfalls I see, etc. But since I didn't look at the PR, I apologize if your code already addresses this (likely does), I just want to be complete for the text history.

I think the functionality of this from a UX perspective is simple. When looking up "foo" in map "m":

  * If "foo" exists as a key, use the value of "foo"

  * If "*" exists as a key, use the value of "*"

  * Else, runtime error.

:+1:

jen20:
@mitchellh this needs re-implementing for 0.7 so I am inclined to defer it until then (the implementation for the current tree looks good, however). As part of this, do you think we should modify HCL to allow `*` as an object key without quotation?

mitchellh:
@jen20 Agreed, 0.7

mzupan:
@mitchellh yes correct it will look for a key first and if not found it'll look for the `*` and if its not there it'll throw the not found error linking the key name used for the lookup

mzupan:
@jen20 I'm fine trying to get it into 0.7 if someone lets me know how to try to go about doing it if not I can wait and use this patch till it gets released

manojlds:
Can we get this merged in both 0.6.x and 0.7.x so that we can leverage this quicker? I hope there is 0.6.15 coming up at the least (looking at master's CHANGELOG.md)?

mitchellh:
0.7 isn't too far away (0.6.15 will come out sooner), and we've done significant changes there. We'll likely just push this into there to avoid any instability in 0.6.x. Trust me that 0.7 is close.

mzupan:
I created a new PR for 0.7 #6351

so this can be closed since it won't go into 0.6

PR #5844, opened 2016-03-24
Renumber original binary state as V0

jen20:
This commit rectifies the fact that the original binary state is referred to as V1 in the source code, but the first version of the JSON state uses StateVersion: 1. We instead make the code refer to V0 as the binary state, and V1 as the first version of JSON state.

PR #5841, opened 2016-03-24
state: Add support for outputs of multiple types

jen20:
This commit adds the groundwork for supporting module outputs of types other than string. In order to do so, the state version is increased from 1 to 2 (though the "public-facing" state version is actually as the
first state file was binary).

Tests are added to ensure that V2 (1) state is upgraded to V3 (2) state, though no separate read path is required since the V2 JSON will unmarshal correctly into the V3 structure.

Outputs in a ModuleState are now of type map[string]interface{}, and a test covers round-tripping string, []string and map[string]string, which should cover all of the types in question.

Type switches have been added where necessary to deal with the interface{} value, but they currently default to panicking when the input is not a string.

PR #6286, opened 2016-04-21
Replace flatmaps with HIL-native maps

jen20:
This PR brings the following sections of work (some of which have already been reviewed) to the `dev-0.7` branch following some rebasing. The work is:

- Upgrade `github.com/hashicorp/hil/...`

- Renumber the original binary state of Terraform as v0 from v1 (already reviewed in #5844)

- Add support in a new V2 state for outputs of types other than strings (maps, lists) (already reviewed in #5841)

- Replace usage of the flatmap structure for map variables with native HIL maps. This also includes the work necessary to allow map variables to pass between modules.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -768,11 +768,11 @@
 		},
 		{
 			"ImportPath": "github.com/hashicorp/hil",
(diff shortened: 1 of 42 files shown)


mitchellh:
Since it is all already reviewed lets go for it.

PR #6287, opened 2016-04-21
fix local-exec by using get instead of accessing Config directly

### LINKED ISSUE #3254: local-exec provisioner missing 'command'
Having strange issue while passing vars from one module to another. First module creates VPC incl. subnets, routing tables and SGs. Second module supposed to get vpc_name, region and run local-exec.
```
module "vpc" {
  source = "../../modules/vpc-terraform"
  region = "${var.region}"
  vpc_name = "${var.vpc_name}"
  vpc_network_prefix = "${var.vpc_network_prefix}"
}

module "consul_cluster_build" {
  source = "../../modules/cluster_build"
  vpc_name = "${module.vpc.vpc_name}"
  region = "${module.vpc.vpc_region}"
  service_name = "consul"
}

```
Outputs are in place.

cluster_build module is very simple. Get parameters and run chef-provisioning.

```
variable vpc_name {}
variable vpc_region {}
variable service_name {}
variable environment {}

resource "null_resource" "cluster_provisioning" {
   provisioner "local-exec" {
    command = "EC2_VPC=${var.vpc_name} EC2_REGION=${var.vpc_region} CHEF_ENVIRONMENT=${var.environment} chef-client -z provisioning/drivers/aws-fog.rb provisioning/cluster_${var.service_name}.rb"
  }
}
```
The reason we are passing vars from vpc module to cluster_build is to make a dependancy working, as module resource has no depends_on.
The apply fails, after successfully creating all vpc module resources, with 

```
Error applying plan:

1 error(s) occurred:

* local-exec provisioner missing 'command'

```

If I'm passing vars to second module from configuration itself, and not from first module, everything runs. However run fails in chef-provisioning, because we must have vpc before using SGs and subnets in chef-provisioner.

Any Idea?

Thanks in advance!

mikedanese:
For some reason the ResourceConfig looked like:
```
&terraform.ResourceConfig{
    ComputedKeys:[]string{"command"},
    Raw:map[string]interface {}{
        "command":"echo ${k-0-master-ip} >> hello.ip.txt",
    },
    Config:map[string]interface {}{},
    raw:(*config.RawConfig)(nil),
}
```
So "command" field existed in Raw but not in Config. It is however accessible through Get.

Fixes #3254 but I'm not familar with the deserialization internals so I'm not sure why.

### DIFF
--- a/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/builtin/provisioners/local-exec/resource_provisioner.go
@@ -27,7 +27,7 @@ func (p *ResourceProvisioner) Apply(
 	c *terraform.ResourceConfig) error {
 
 	// Get the command
-	commandRaw, ok := c.Config["command"]
+	commandRaw, ok := c.Get("command")
 	if !ok {
 		return fmt.Errorf("local-exec provisioner missing 'command'")
 	}


mikedanese:
In fact this might be a case of user error and a bad error message...

mikedanese:
Actually this was caused by user error, but the error message is pretty bad: "local-exec provisioner missing 'command'" if the string failed to interpolate.

hugoduncan:
Thank you @mikedanese for your explanation of the cause of the `local-exec provisioner missing 'command` message.

PR #6184, opened 2016-04-14
core: change taint semantics

svanharmelen:
Change tainted resources to act as a normal (first class citizens) resource. This means it’s shown correctly in a plan/diff and takes into account any actions that are dependant on the tainted resource and, vice verse, any actions that the tainted resource depends on.

So this changes the behaviour from saying this resource is tainted so just forget about it and make sure it gets deleted in the background, to saying I want that resource to be recreated (taking into account the
existing resource and it’s place in the graph).

An example of what this fixes is if you taint a resource (say an instance in a network) and then call `terraform destroy`. Before this would not work properly because both the instance and the network would be destroyed in parallel (so no deps taking into account). Now the graph will be correct and so the destroy will succeed properly.

Also it brings a little better UX:

Before:
image

After:
image

This change alters the state file layout, but I don't think a migration is needed. I noticed there is another version of the state being prepared, so I think this change should tag along on that change.

PR #6288, opened 2016-04-21
core: change taint semantics

svanharmelen:
Change tainted resources to act as a normal (first class citizens) resource. This means it’s shown correctly in a plan/diff and takes into account any actions that are dependant on the tainted resource and, vice verse, any actions that the tainted resource depends on.

So this changes the behaviour from saying this resource is tainted so just forget about it and make sure it gets deleted in the background, to saying I want that resource to be recreated (taking into account the
existing resource and it’s place in the graph).

An example of what this fixes is if you taint a resource (say an instance in a network) and then call `terraform destroy`. Before this would not work properly because both the instance and the network would be destroyed in parallel (so no deps taking into account). Now the graph will be correct and so the destroy will succeed properly.

Also it brings a little better UX:

Before:
image

After:
image

This change alters the state file layout, but I don't think a migration is needed. I noticed there is another version of the state being prepared, so I think this change should tag along on that change.


svanharmelen:
This is a follow up of PR #6184 The main comments there were these:

@mitchellh:
>This is interesting. I'll color some history on here too, as well some skepticism and questions.

>A bit of history...

>The concept of "tainted" has existed since Terraform 0.1. It was originally implemented as you have done here. We ultimately added Tainted as a separate slice of instance states to support "create before destroy." This led to all sorts of other pains and bugs, and one of the first things @phinze did as a HC employee years back was introduce the formal concept of "deposed" vs. "tainted" since they are semantically a bit different.

>Which leads us to today: we have both tainted and deposed as slices of instance states. I can no longer think of a reason that there would be more than one tainted but also when @phinze was doing the deposed work we didn't want to do too much madness.

>And now PR questions...

>My main feedback: I don't see any tests for functionality this fixed/improved, I only see existing tests that are updated to the new API. For this to make sense, I think we need to prove that this fixes prior/other problems without introducing new ones. So, we need at least one new test in context_apply_test.go. Without it, the refactor is nice, but uncertain on its utility (and I believe both of you that there is one).

>I'd think very, very hard about whether we can have multiple taints in the current model. I don't think its possible any longer (since even CBD goes to deposed, not tainted).

>Further, I question: any graph ordering bugs that exist with tainted probably exist with deposed. Does this cause problems? That shouldn't affect this PR probably but we should think that those bugs probably still exist.

@apparentlymart:
>The graph ordering issues aside, I've definitely experienced (and seen others on my team experience) the shock and panic of terraform apply tfplan saying it's deleting something that was never mentioned in the plan because someone happened to leave something tainted in the state for some reason.

>So in terms of user confidence that Terraform is behaving correctly, I personally think it's important that tainting be something that's explicitly included in the plan. I think this applies equally to deposed things that weren't successfully deleted on a previous run, for the same reasons; just last week some operational issues caused us to accumulate a bunch of broken resources in the "deposed" state, and I was only able to explain Terraform's resulting behavior by delving into the state.

>I really like the UX change in this PR, where it's explicitly marked as "tainted" in the plan to help the user understand what's going on. How exactly that would be incorporated into the tests I'm not sure. Part of it would be to test that tainted and deposed stuff correctly shows up in the plan, presumably.

@mitchellh:
>Agreed with @apparentlymart that tainteds need to show up in the plan. :)

svanharmelen:
@mitchellh I hear what you're saying and agree that tests should be added to show what this PR fixes. So I looked around in the code base today searching for the right location to add 2 new tests. But after writing a few half tests is different locations, I'm still not sure what the right location for these new tests should be. 

You said to expect at least one new test in `context_apply_test.go`, but I don't think that is the correct location for these tests (I could be wrong of course). So let me explain the 2 use cases this PR fixes and after that hopefully you can point me into the right direction for where and how to add test for them? 

**1. Run `terraform destroy` on a config/state that contains a tainted resource**

Say you have a config like this:
```
provider "cloudstack" {
  ...
}

resource "cloudstack_vpc" "foo" {
  ...
}

resource "cloudstack_network" "foo" {
  ...
  vpc_id = "${cloudstack_vpc.foo.id}"
}

resource "cloudstack_instance" "foobar1" {
  ...
  network_id = "${cloudstack_network.foo.id}"
}
```

As you can see the instance depends on the network, and the network depends on the VPC. So when we want to destroy everything, we cannot delete the network until there are now more instances in that network. And w cannot delete the VPC until there are no more network in that VPC.

So a normal `destroy` on this config works just fine, destroying all the resources in the right order. But when I first `taint` the instance and then run a `terraform destroy`, the ordering is not correct anymore. It will start destroying the instance and the network at the same time as the tainted instance isn't placed in the graph correctly.

See two outputs or the resulting graph for a `terraform plan -destroy`:

v0.6.14
image
image
As you can see the network has no dependency on the instance (and the instance is not shown in the plan), yet it cannot be destroyed by the API as long as the instance exists.

master (with cherry-picked commits from this PR)
image
image
Here the dependancy is correct (and the instance is also shown in the plan) and the `destroy` will succeed without any errors.

**2. Run `terraform apply` on a config/state that contains two dependent tainted resources**

Consider the exact same config as a before, but now we `taint` both the instance and the VPC. If we do that without this PR, it will again not build a correct graph with will break the execution. With this PR the ordering will be as expected.

Again two outputs that show the different graphs:

v0.6.14
image
image

master (with cherry-picked commits from this PR)
image
image

Again this case also the plan output is more inline/correct with the actual situation and what you would expect.

So... This is a whole lot of text and images, but I think it should be clear what I was trying to fix/improve here. So any guidance on where and how to add proper testing for this would be great, so we can (hopefully) move on with this PR.

Cheers!

mitchellh:
@svanharmelen I think both cases are testable. Can you create a `null_resource` example set that exercises each (with `depends_on` and the `taint` command as a way to emulate the actual dependencies)? I can write a test for each.

svanharmelen:
Sure thing... Will whip something up after dinner. Thanks for the help!

svanharmelen:
@mitchellh here you go...

The example is really simple and basic, but still shows the problem perfectly:

```
resource "null_resource" "network" {}

resource "null_resource" "instance" {
  depends_on = ["null_resource.network"]
}
```

With this config, these commands will demonstrate the first case:

```
terraform apply
terraform taint null_resource.instance
terraform plan -destroy
```

The trace output from the last command will show the same problem (and plan output) as in case 1 from my earlier reply.

For the second case wipe your state and execute the following commands:

```
terraform apply
terraform taint null_resource.instance
terraform taint null_resource.network
terraform plan
```

Unfortunately this time the trace output from the last command will now show the same problem as it seems the tainting is ignored for the null_resource:

image

So TF says there is nothing to do and the resources remain tainted in the state. I guess this is related to them being `null_resource`'s as I cannot reproduce that with actual resources. But easy enough, just change the `null_resource`'s in the config to any other actual resource and you'll be able to reproduce the second case as well.

Was this the kind of examples that you were expecting? If not, please let me know what else you need.

svanharmelen:
@mitchellh any progress on this one? Would still very much like to get this in, so let me know if you expect anything from me... Thx!

mitchellh:
@svanharmelen I just wrapped up a bunch of work today, I'll take a look at this next.

svanharmelen:
Sounds good... Thx!

svanharmelen:
@mitchellh just noticed dev-0.7 is being merged into master today... So how do we proceed with this one? I guess I need to rebase it against master, but I radar do that only once :wink:

So let me know your thoughts...

jen20:
This was closed by accident as a result of dev-0.7 getting merged. If you can retarget it at `master` that would be great, @svanharmelen!

PR #6290, opened 2016-04-21
provider/azurerm: Error on bad creds and speed++

jen20:
This commit uses Riviera to register the Microsoft.Compute provider as a canary for whether or not the Azure account credentials are set up. It used to use the MS client, but that appeared to panic internally if the credentials were bad. It's possible that we were using it wrong, but there are no docs so `¯\_(ツ)_/¯`.

As part of this, we parellelise the registration of the other providers. This shaves the latency of each provider request times the number of providers minus 1 off the "startup" time of the AzureRM provider. The result is quite noticeable.


jen20:
cc @stack72 - this closes a ton of issues too.

stack72:
LGTM!

PR #6294, opened 2016-04-22
provider/aws: Refresh state on Directory Service not found

### LINKED ISSUE #6292: "aws_directory_service_directory" plan causes a panic if the resource is deleted manually
Hi there,

### Terraform Version
Terraform version: 0.6.14

### Affected Resource(s)
- aws_directory_service_directory

```hcl
resource "aws_directory_service_directory" "microsoft-ad" {
  name = "${var.ad-fqdn}"
  password = "whatever"
  size = "Small"
  type = "MicrosoftAD"

  vpc_settings {
    vpc_id = "${var.vpc-id}"
    subnet_ids = [ "${element(split(",", var.public-mgmt-subnet-ids), 0)}", "${element(split(",", var.public-mgmt-subnet-ids), 1)}" ]
  }
}
```

### Debug Output
https://gist.github.com/asvoboda/0f1fb8bbc7267aa12d5c59b37f249e6d

### Expected Behavior
Terraform continues merrily on its way, just as it does for any resource that was manually deleted.

### Actual Behavior
Terraform panics and crashes.

### Steps to Reproduce
1. `terraform apply` and creates a ds resource
2. Delete that resource manually and wait until it is fully removed. 
3. `terraform plan`, and terraform panics

stack72:
Fixes #6292

When a directory service was not found, Terraform was panicking due to
`dir := out.DirectoryDescriptions[0]`. The AWS API doesn't throw an
Error in this case. IT just return s0 results. Therefore, we should
check for 0 results in the return and remove the directory from the
state

### DIFF
--- a/builtin/providers/aws/resource_aws_directory_service_directory.go
+++ b/builtin/providers/aws/resource_aws_directory_service_directory.go
@@ -401,6 +401,13 @@ func resourceAwsDirectoryServiceDirectoryRead(d *schema.ResourceData, meta inter
 	out, err := dsconn.DescribeDirectories(&input)
 	if err != nil {
 		return err
+
+	}
+
+	if len(out.DirectoryDescriptions) == 0 {
+		log.Printf("[WARN] Directory %s not found", d.Id())
+		d.SetId("")
+		return nil
 	}
 
 	dir := out.DirectoryDescriptions[0]


jen20:
LGTM.

PR #6297, opened 2016-04-22
JSON render format option to output command

coen-hyde:
Hi,

This PR adds a format option to the output command. Valid formats are "display" and "json", with the default being "display"; the current renderer. Is this your preferred way to add a json format renderer? An alternative could be just to specify a --json flag or similar.

### DIFF
--- a/command/output.go
+++ b/command/output.go
@@ -1,6 +1,7 @@
 package command
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"sort"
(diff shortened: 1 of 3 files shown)


jen20:
Hi @coen-hyde! Thanks for opening up a pull request and for your work. I think this might be best targeted at the `dev-0.7` branch which has work modifying the structure of outputs in order to support lists and maps. I am happy to look over this over then next few days, and agree machine readable outputs is a great idea.

coen-hyde:
Closing in favor of #6299

PR #6298, opened 2016-04-22
core: Do not type check unset variables

### LINKED ISSUE #6230: Issue with new map check code
Having issues with a specific commit right now - d7d39702c04b3ae5e2df0b50c6a7aa9b479d19b3.

Given a configuration snippet like this:
```hcl
variable invertbool {
  type = "map"
  default = {
    "1" = false
    "0" = true
  }
}
```

Which is then called off of something like:

```hcl
${lookup(var.invertbool, var.bool)}
```

I'm getting (paraphrased):

```
variable invertbool should be type map, got type string
```

Reverting the commit fixes the issue.

It should be noted that I am not strictly working off strings here, as these are older modules that are taking "boolean" values, that ultimately get interpolated in modules into `1` and `0`. However, in messing around with the module, I haven't been able to get past this error, even when I change everything (the inner module key variable, and the outer module parameter) to be a specific string.

Also, a bit more context as to what this is specifically doing - basically, the main application here is to create a matching ephemeral rule for `aws_network_acl_rule` resources. It's designed to flip the supplied `egress` parameter.

jen20:
A consequnce of the work done in #6185 was that variables which were in a module but not set explicitly (i.e. the default value was relied upon) were marked as type errors. This was reported in #6230.

This commit adds a test case for this and a patch which fixes the issue.

Fixes #6230.

### DIFF
--- a/terraform/context_plan_test.go
+++ b/terraform/context_plan_test.go
@@ -660,6 +660,23 @@ func TestContext2Plan_moduleVarWrongTypeNested(t *testing.T) {
 	}
 }
 
+func TestContext2Plan_moduleVarWithDefaultValue(t *testing.T) {
+	m := testModule(t, "plan-module-var-with-default-value")
+	p := testProvider("null")
+	p.DiffFn = testDiffFn
+	ctx := testContext2(t, &ContextOpts{
+		Module: m,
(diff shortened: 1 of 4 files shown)


mitchellh:
LGTM!

Issue #2460, opened 2015-06-24
Allow JSON output where it makes sense


radeksimko:
As discussed previously elsewhere, it would be useful to have JSON output coming from certain terraform commands, especially when people use it as an abstraction layer in another project/tool instead of a standalone tool in CLI.

The assumption is we'd have **`-format=json`** argument accepted for the commands below.

At some point _all_ the commands could have JSON output, but I do aim to cover at least the basic use-cases **first**:
### `apply`
- [ ] `apply` implying `-input=false` & `-no-color`
- [ ] `apply` - list of errors
- [ ] `apply` - refresh events log (e.g. `aws_vpc.main: Refreshing state... (ID: vpc-1a131c7f)`) -> I'd begin with (potentially long) string 1:1 buffered output of what we have already
- [ ] `apply` - execution events log (e.g. `aws_vpc.main: Creation complete`) -> I'd begin with (potentially long) string 1:1 buffered output of what we have already
- [ ] `apply` - sums of modified resources (e.g. `{"created": 1, "changed": 0, "removed": 0}`)
- [ ] `apply` - list of `output`s
### `destroy`
- [ ] `destroy` implying `-force` & `-no-color`
- [ ] `destroy` - list of errors
- [ ] `destroy` - refresh events log (e.g. `aws_vpc.main: Refreshing state... (ID: vpc-1a131c7f)`) -> I'd begin with (potentially long) string 1:1 buffered output of what we have already
- [ ] `destroy` - execution events log (e.g. `aws_vpc.main: Destruction complete`) -> I'd begin with (potentially long) string 1:1 buffered output of what we have already
- [ ] `destroy` - sums of modified resources (e.g. `{"created": 0, "changed": 0, "removed": 3}`)
### `plan`
- [ ] `plan` implying `-no-color`
- [ ] `plan` - refresh events log (e.g. `aws_vpc.main: Refreshing state... (ID: vpc-1a131c7f)`) -> I'd begin with (potentially long) string 1:1 buffered output of what we have already
- [ ] `plan` - readable version of `terraform plan -out`
### `output`

This will likely require https://github.com/hashicorp/terraform/issues/1209 / https://github.com/hashicorp/terraform/issues/2335 / https://github.com/hashicorp/terraform/issues/2390 to make the output actually useful.
- [ ] `output` implies `-no-color`
- [ ] `output` - list of outputs
### `version`

Just because any 3rd party tool likes to know that and because it's IMO super easy to do this one.
[...]

PR #6299, opened 2016-04-22
JSON render format option to output command - dev-0.7

coen-hyde:
This is a reimplementation of #6297 against dev-0.7. 

This PR adds a format option to the output command. Valid formats are "display" and "json", with the default being "display"; the current renderer.


radeksimko:
Related: https://github.com/hashicorp/terraform/issues/2460

coen-hyde:
@radeksimko Ah, good that I implemented the format export interface as "-format=json". Does this PR's implementation fit into the wider json export vision? It might make sense to have a renderDataAs function that accepts the format and data to render, rather than a one off format switch and direct json serialization. Thoughts?

radeksimko:
> Does this PR's implementation fit into the wider json export vision?

Probably the cleanest way to implement this would be via a global flag which would be changing `cli.Ui` implementation as Mitchell suggested in the linked issue.

Whether or not to accept this PR as a short term solution that may evolve into the "global" solution is a question mark.

### Global flag

I think it would make sense to have such global `-format` flag that would match the current behaviour (`BasicUi`). e.g.

```
terraform -format=json any-subcommand
```

User should be well aware that they need to expect _stream_ of JSON objects, not a raw JSON output, hence `terraform -format=json cmd | jq .` and similar should be treated as anti-patterns.

[This piece of code](https://github.com/hashicorp/terraform/blob/master/main.go#L104-L110) also makes me think we should make some changes in the CLI library so we can have 1st class global flags, like [`codegangsta/cli`](https://github.com/codegangsta/cli/blob/master/context.go#L98), but it may not be a blocker for this feature.

### `cli.JsonUi`

Based on the global flag we could be just swapping the [`cli.BasicUi`](https://github.com/hashicorp/terraform/blob/master/commands.go#L28) with something like `cli.JsonUi`.

The significant benefits here are that all existing logging (including errors) would automatically turn into JSON too. Whether the output would be actually useful is another thing. It would be something like `{"message":"content-of\nlog-message"}` in most cases.

Some commands would naturally benefit from being able to output more complicated data-structures instead of plain string - like `output`, so we could expand the [current `cli.Ui` interface](https://github.com/mitchellh/cli/blob/master/ui.go#L19) like this:

```go
type Ui interface {
...
	OutputMap(map[string]interface{})
}
```

Then the diff for `command/output.go` would become just `-1/+1`:

```diff
@@ -81,7 +81,7 @@ func (c *OutputCommand) Run(args []string) int {
        }

        if name == "" {
-               c.Ui.Output(outputsAsString(state))
+               c.Ui.OutputMap(state.RootModule().Outputs)
                return 0
        }
```
and it would be up to the chosen `cli.Ui` to decide what to do with a map.

@jen20 I know you were recently working on map/list outputs, so you may want to add something here too.

coen-hyde:
Thanks @radeksimko for such detailed thoughts. @jen20 with the closure of this ticket. Does this mean that this has been implemented?

PR #6304, opened 2016-04-22
providers/aws: Opsworks permission resource

janschumann:
The code in this PR works, but unfortunately the tests fail, complaining that `aws_opsworks_permission.tf-acc-perm` could not be found.

Also I am not sure if I have implemented it correctly. E.g. I had to implement a `Delete` function although this resource cannot be deleted. I left the delete function empty ...

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_permission.go
+++ b/builtin/providers/aws/resource_aws_opsworks_permission.go
@@ -1,0 +1,152 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	//"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/opsworks"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsOpsworksPermission() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsOpsworksPermissionCreate,
+		Update: resourceAwsOpsworksPermissionCreate,
+		Delete: resourceAwsOpsworksPermissionDelete,
+		Read:   resourceAwsOpsworksPermissionRead,
+
+		Schema: map[string]*schema.Schema{
+			"id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"allow_ssh": &schema.Schema{
+				Type:     schema.TypeBool,
+				Computed: true,
+				Optional: true,
+			},
+			"allow_sudo": &schema.Schema{
+				Type:     schema.TypeBool,
+				Computed: true,
+				Optional: true,
+			},
+			"user_arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			// one of deny, show, deploy, manage, iam_only
+			"level": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+
+					expected := [5]string{"deny", "show", "deploy", "manage", "iam_only"}
+
+					found := false
+					for _, b := range expected {
+						if b == value {
+							found = true
+						}
+					}
+					if !found {
+						errors = append(errors, fmt.Errorf(
+							"%q has to be one of [deny, show, deploy, manage, iam_only]", k))
+					}
+					return
+				},
+			},
+			"stack_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceAwsOpsworksPermissionDelete(d *schema.ResourceData, meta interface{}) error {
+	return nil
+}
+
+func resourceAwsOpsworksPermissionRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*AWSClient).opsworksconn
+
+	req := &opsworks.DescribePermissionsInput{
+		IamUserArn: aws.String(d.Get("user_arn").(string)),
+		StackId:    aws.String(d.Get("stack_id").(string)),
+	}
+
+	log.Printf("[DEBUG] Reading OpsWorks prermissions for: %s on stack: %s", d.Get("user_arn"), d.Get("stack_id"))
+
+	resp, err := client.DescribePermissions(req)
+	if err != nil {
+		if awserr, ok := err.(awserr.Error); ok {
+			if awserr.Code() == "ResourceNotFoundException" {
+				log.Printf("[INFO] Permission not found")
+				return nil
+			}
+		}
+		return err
+	}
+
+	permission := resp.Permissions[0]
+
+	d.Set("allow_ssh", permission.AllowSsh)
+	d.Set("allow_sodo", permission.AllowSudo)
+	d.Set("user_arn", permission.IamUserArn)
+	d.Set("stack_id", permission.StackId)
+	id := *permission.IamUserArn + *permission.StackId
+	d.SetId(id)
+	d.Set("id", id)
+
+	return nil
+}
+
+func resourceAwsOpsworksPermissionValidate(d *schema.ResourceData) error {
+	return nil
+}
+
+func resourceAwsOpsworksPermissionCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*AWSClient).opsworksconn
+
+	err := resourceAwsOpsworksPermissionValidate(d)
+	if err != nil {
+		return err
+	}
+
+	req := &opsworks.SetPermissionInput{
+		AllowSudo:  aws.Bool(d.Get("allow_sudo").(bool)),
+		AllowSsh:   aws.Bool(d.Get("allow_ssh").(bool)),
+		IamUserArn: aws.String(d.Get("user_arn").(string)),
+		StackId:    aws.String(d.Get("stack_id").(string)),
+	}
+
+	var resp *opsworks.SetPermissionOutput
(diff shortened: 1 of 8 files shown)


builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -0,0 +1,152 @@
...
+			"stack_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceAwsOpsworksPermissionDelete(d *schema.ResourceData, meta interface{}) error {
```
stack72: is there no delete functionality here?

builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	log.Printf("[DEBUG] Reading OpsWorks prermissions for: %s on stack: %s", d.Get("user_arn"), d.Get("stack_id"))
+
+	resp, err := client.DescribePermissions(req)
+	if err != nil {
+		if awserr, ok := err.(awserr.Error); ok {
+			if awserr.Code() == "ResourceNotFoundException" {
+				log.Printf("[INFO] Permission not found")
+				return nil
```
stack72: if the permission is not found then we should remove it from state `d.SetId("")`

builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -0,0 +1,152 @@
...
+	d.Set("stack_id", permission.StackId)
+	id := *permission.IamUserArn + *permission.StackId
+	d.SetId(id)
+	d.Set("id", id)
+
+	return nil
+}
+
+func resourceAwsOpsworksPermissionValidate(d *schema.ResourceData) error {
+	return nil
```
stack72: why a nil validate func?

website/source/docs/providers/aws/r/opsworks_permission.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+* `allow_ssh` - (Optional) Whethe the user is allowed to use SSH to communicate with the instance
+* `allow_sudo` - (Optional) Whethe the user is allowed to use sudo to elevate privileges
+* `user_arn` - (Required) The user's IAM ARN to set permissions for
+* `level` - (Optional) The users permission level. Mus be one of `deny`, `show`, `deploy`, `manage`, `iam_only` 
+* `stack_id` - (Required) The stack to set the permissions for 
+
+## Attributes Reference
+
+The following attributes are exported:
+
```
stack72: We need to add `id` as an exported attribute

builtin/providers/aws/resource_aws_opsworks_permission_test.go
```diff
@@ -0,0 +1,41 @@
...
+					),
+				),
+			},
+		},
+	})
+}
+
+var testAccAwsOpsworksPermissionCreate = testAccAWSUserConfig + testAccAwsOpsworksStackConfigNoVpcCreate("tf-ops-acc-permission") + `
+resource "aws_opsworks_permission" "tf-acc-perm" {
+  stack_id = "${aws_opsworks_stack.tf-acc.id}"
```
stack72: where does this resource get referenced from? It's not from within this test

builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -0,0 +1,152 @@
...
+		if awserr, ok := err.(awserr.Error); ok {
+			if awserr.Code() == "ResourceNotFoundException" {
+				log.Printf("[INFO] Permission not found")
+				return nil
+			}
+		}
+		return err
+	}
+
+	permission := resp.Permissions[0]
```
stack72: I think we need to check here that `Permissions[0]` is the permission we expect. We may need to range over the list to make sure we get the correct one

builtin/providers/aws/resource_aws_opsworks_permission.go
janschumann: Yes. There is no delete. Also create is not really there, as permissions objects are created on creation of the stack. These are crated with default values an can than be changed.

builtin/providers/aws/resource_aws_opsworks_permission_test.go
janschumann: Is it not references by 
```golang 
        "aws_opsworks_permission.tf-acc-perm", "level", "iam_only",
    )
````

File: website/source/docs/providers/aws/r/opsworks_permission.html.markdown (hunk 4 above)
Comment (janschumann) [inline website/source/docs/providers/aws/r/opsworks_permission.html.markdown]: fixed in dbc3b23f895ef5a19596270f9dee0c09d45ac838

**Comment by u2mejc** [comment]
Hi guys! Happy friday! What's the status of this? @stack72 Anything else you would like to see changed before we can have it merged?

:shipit:

**Comment by u2mejc** [comment]
Morning guys! @stack72 are you still able to take a look at this, or should we see if @apparentlymart has time?

File: builtin/providers/aws/resource_aws_opsworks_permission.go [hunk 7]
```diff
@@ -0,0 +1,158 @@
...
+
+	log.Printf("[DEBUG] Reading OpsWorks prermissions for: %s on stack: %s", d.Get("user_arn"), d.Get("stack_id"))
+
+	resp, err := client.DescribePermissions(req)
+	if err != nil {
+		if awserr, ok := err.(awserr.Error); ok {
+			if awserr.Code() == "ResourceNotFoundException" {
+				log.Printf("[INFO] Permission not found")
+				d.SetId("")
+				d.Set("id", "")
```
stack72: We don't need this - the SetId("") takes care of it :)

builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		return err
+	}
+
+	found := false
+	id := ""
+	for _, permission := range resp.Permissions {
+		id = *permission.IamUserArn + *permission.StackId
+
+		if d.Get("user_arn").(string)+d.Get("stack_id").(string) == id {
```
stack72: after we set found = true, do we need to break the loop? This looks like it may continue to loop unnecessarily?

builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -0,0 +1,158 @@
...
+			found = true
+			d.SetId(id)
+			d.Set("id", id)
```
stack72: We don't need this - d.SetId(id) takes care of it for us :)

stack72:
Hi @janschumann 

Apologies for this slipping through the cracks (thanks @u2mejc for the reminder!)

I have just ran the tests (after leaving a few more small comments). The test fails I am afraid

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSOpsworksPermission'
2016/06/02 01:10:13 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSOpsworksPermission -timeout 120m
=== RUN   TestAccAWSOpsworksPermission
--- FAIL: TestAccAWSOpsworksPermission (142.00s)
	testing.go:239: Step 0 error: Error applying: 1 error(s) occurred:

		* aws_opsworks_permission.tf-acc-perm: ResourceNotFoundException: Unable to find user with ARN arn:aws:iam::881237884953:user/test-user
			status code: 400, request id: b94c910a-2856-11e6-9260-019c62f4ef41
FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/aws	142.019s
make: *** [testacc] Error 1
```
this is because of this line in the test code:

```
user_arn = "${aws_iam_user.user.arn}"
```

We are not creating that test user as part of the test - once this gets created, they should pass. So close to being ready for merge here!

Thanks

Paul

stack72:
Hi @janschumann 

Any thoughts on how we get the tests for this resource working as expected?

Paul

janschumann:
Jepp sorry, I have them working already. Hope I will find time for the finish during the next few days.

Best, Jan

janschumann:
Now I was able to interpret what the failing test tells us: The user which is referenced by `${aws_iam_user.user.arn}` is created. But that user has to be imported to OpsWorks via another resource wich is referred to as `UserProfile` within the aws sdk.

Sorry for the confusion and the delay here!

I will replace `resourceAwsOpsworksPermission` with `resourceAwsOpsworksUserProfile`. A permission is not a resource, but a configuration for the user profile.

janschumann:
I will add doc for user_profile later today

stack72:
Hi @janschumann 

Testing this right now :)

Paul

builtin/providers/aws/resource_aws_opsworks_user_profile.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return nil
+}
+
+func resourceAwsOpsworksUserProfileCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*AWSClient).opsworksconn
+
+	req := &opsworks.CreateUserProfileInput{
+		AllowSelfManagement: aws.Bool(d.Get("allow_self_management").(bool)),
```
stack72: As this is an optional param, with no default, what will happen here if there is no value set?

janschumann: @stack72 unfortunately the AWS documentation does not mention a default value, but as you can see in the test, the AWS default is `false`. Should I add `Computed: true` to that parameter?

stack72: I personally would skip the Computed and add `Default: false` to the schema. That way we can guarantee that we have a value and d.GetOk will not fail to get a value

janschumann: check. default value added.

stack72:
thanks @janschumann 

will run these tests really soon - just testing something else first!

P.

PR #6307, opened 2016-04-22
provider/aws: Fix issue updating ElasticBeanstalk Configuraiton Templates

### LINKED ISSUE #6301: aws_elastic_beanstalk_configuration_template.options changes not updated
Updates/changes to aws_elastic_beanstalk_configuration_template.options are not reflected in AWS. Testing has shows that changes to options delete them from the configuration template. 
There appears to be a clear typo bug on https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_elastic_beanstalk_configuration_template.go#L174 - but patching that line does not fix the issue.

This is probably related to related: #6035 @catsby

catsby:
Fixes #6301 by correctly calculating the set difference, and then pruning any settings that are being updated from the removable list, as Elastic Beanstalk expects the additions/updates to be mutually exclusive from the removals, since it's in a single API call.

Also fix an error when a configuration template is not found, correctly removing it from state and prompting to recreate it.

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_configuration_template.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_configuration_template.go
@@ -3,10 +3,12 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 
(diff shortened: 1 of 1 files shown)


stack72:
LGTM!

Issue #6172, opened 2016-04-13
aws_s3_bucket_object uses kms_key_id attribute inconsistently


beeerd:
### Terraform Version
0.6.14

### Affected Resource(s)
- aws_s3_bucket_object
- aws_kms_key

### Terraform Configuration Files
```
resource "aws_kms_key" "main" {}

resource "aws_s3_bucket" "ca_cert" {
  bucket = "ca.cert"
  acl = "private"
}

resource "aws_s3_bucket" "ca_key" {
  bucket = "ca.key"
  acl = "private"
}

resource "aws_s3_bucket_object" "ca_cert" {
  key = "ca.crt"
  bucket = "ca.cert"
  source = "/tmp/certs/ca/ca.crt"
  kms_key_id = "${aws_kms_key.main.id}"
  depends_on = ["aws_s3_bucket.ca_cert"]
}

resource "aws_s3_bucket_object" "ca_key" {
  key = "ca.key"
  bucket = "ca.key"
  source = "/tmp/certs/ca/ca.key"
  kms_key_id = "${aws_kms_key.main.id}"
  depends_on = ["aws_s3_bucket.ca_key"]
}
```

### Debug Output
First apply: https://gist.github.com/beeerd/45da50c41eee72ea1a9d4d6aa894026b
Second apply: https://gist.github.com/beeerd/dc2f9ff9f79c61ba491e4fe21318e231

### Expected Behavior
Second apply should be no-op, since no changes are made between runs of `terraform apply`

### Actual Behavior
On the first run of `terraform apply` the `aws_kms_key.arn` resource is incorrectly interpolated as the attribute for each s3 bucket object's `kms_key_id` resource reference. This can be observed by looking at the terraform.tfstate file after the first run. Stdout does not reveal this behavior.

On the second run the `aws_kms_key.main.id` resource is applied instead of `aws_kms_key.main.arn`. This triggers an additional error about etag being used in conjunction with `kms_key_id`.

### Steps to Reproduce
Using the HCL above, run:
1. `terraform apply`
2. `terraform apply`

### Workaround
Explicitly using `aws_kms_key.*.arn` in place of `aws_kms_key.*.id` on the `kms_key_id` attribute of a `aws_s3_bucket_object` resource will avoid this behavior. This is confusing considering the attribute name implies that the kms key id should be used rather than the arn.

### References
Possibly related:
- GH-5689

catsby:
[...]

PR #6309, opened 2016-04-22
provider/aws: Update S3 Bucket Object docs for KMS Key

catsby:
Fixes https://github.com/hashicorp/terraform/issues/6172 with the power of documentation!

Currently our docs for S3 Bucket Object KMS support say this:

    * `kms_key_id` - (Optional) Specifies the AWS KMS key ID to use for object encryption

In _Terraform_ terms, this is typically the _id_: `${aws_kms_key.foo.id}`. 
In _AWS_ terms, this is... really the _arn_: `${aws_kms_key.foo.arn}`

For creating a bucket object AWS will accept both the unique/random bits at the end of the arn, or the full arn itself as the "ID". Unfortunately when we do a READ on the bucket object, the KMS Key _ID_ will be the full ARN. So users that specify the _id_ will get a reoccurring plan after, as it wants to swap out the ARN it read for the ID is has. 

This PR updates the documentation to clarify that you need to use the exported ARN attribute of a KMS Key resource.

### DIFF
--- a/website/source/docs/providers/aws/r/s3_bucket_object.html.markdown
+++ b/website/source/docs/providers/aws/r/s3_bucket_object.html.markdown
@@ -23,6 +23,27 @@ resource "aws_s3_bucket_object" "object" {
 }
 ```
 
+### Encrypting with KMS Key
+
+```
+resource "aws_kms_key" "examplekms" {
+  description             = "KMS key 1"
+  deletion_window_in_days = 7
+}
+
+resource "aws_s3_bucket" "examplebucket" {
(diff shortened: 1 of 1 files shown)


radeksimko:
As discussed via Slack, this LGTM 👍 , although long-term we could do better (i.e. make it work with `id`)

I reckon someone _will_ come back here to submit another bug arguing they used `id` and expected it to just work 😃 and honestly I wouldn't blame them for not reading the note in docs as the field is called `kms_key_`**ID**

PR #6319, opened 2016-04-23
Increasing ARM template deployment timeout to 20m

Felivel:


Felivel:
Please move this PR to the right brach. Thanks.

jen20:
Hi @Felivel! Can you describe the rationale around 20 minutes vs 10? I understand that ARM templates can take a while to apply - I am wondering whether we should actually increase the timeout further. For now I am happy to merge this - but could you rebase onto `master` to avoid having the merge commit from your other repository? Thanks!

jen20:
Hi @Felivel! I've made the change to 40 minutes and rebased this - it hit as ae38ef2. Thanks for your work!

Felivel:
Thanks @jen20.

PR #6321, opened 2016-04-24
provider/aws: Add support for api_gateway_account

radeksimko:
<del>I've reached out to Amazon to get a confirmation whether it's really impossible to reset CloudWatch Role ARN to empty string. That's why I marked this as WIP, otherwise it works ok (within the limitations of the API).</del> see below

### Test plan

```sh
make testacc TEST=./builtin/providers/aws TESTARGS='-run=AWSAPIGatewayAccount_basic'
```
```
TF_ACC=1 go test ./builtin/providers/aws -v -run=AWSAPIGatewayAccount_basic -timeout 120m
=== RUN   TestAccAWSAPIGatewayAccount_basic
--- PASS: TestAccAWSAPIGatewayAccount_basic (61.86s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	61.880s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_api_gateway_account.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_account.go
@@ -1,0 +1,124 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/apigateway"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsApiGatewayAccount() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsApiGatewayAccountUpdate,
+		Read:   resourceAwsApiGatewayAccountRead,
+		Update: resourceAwsApiGatewayAccountUpdate,
+		Delete: resourceAwsApiGatewayAccountDelete,
+
+		Schema: map[string]*schema.Schema{
+			"cloudwatch_role_arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"throttle_settings": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"burst_limit": &schema.Schema{
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+						"rate_limit": &schema.Schema{
+							Type:     schema.TypeFloat,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceAwsApiGatewayAccountRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	log.Printf("[INFO] Reading API Gateway Account %s", d.Id())
+	account, err := conn.GetAccount(&apigateway.GetAccountInput{})
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Received API Gateway Account: %s", account)
+
+	if _, ok := d.GetOk("cloudwatch_role_arn"); ok {
+		// CloudwatchRoleArn cannot be empty nor made empty via API
+		// This resource can however be useful w/out defining cloudwatch_role_arn
+		// (e.g. for referencing throttle_settings)
+		d.Set("cloudwatch_role_arn", account.CloudwatchRoleArn)
+	}
+	d.Set("throttle_settings", flattenApiGatewayThrottleSettings(account.ThrottleSettings))
+
+	return nil
+}
+
+func resourceAwsApiGatewayAccountUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	input := apigateway.UpdateAccountInput{}
+	operations := make([]*apigateway.PatchOperation, 0)
+
+	if d.HasChange("cloudwatch_role_arn") {
+		arn := d.Get("cloudwatch_role_arn").(string)
+		if len(arn) > 0 {
+			// Unfortunately AWS API doesn't allow empty ARNs,
+			// even though that's default settings for new AWS accounts
+			// BadRequestException: The role ARN is not well formed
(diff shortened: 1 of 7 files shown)


radeksimko:
I received response from AWS support confirming my findings - i.e. it's not possible to remove an ARN or make it blank. 😢 

> Hi,

> From my testing it doesn't look like you can remove this. While this shouldn't be a huge security issue as it's only logging events within that particular APIGW region, I will forward this to our APIGW team to look at as a feature request moving forward.

This is therefore ready for review/merge.

builtin/providers/aws/resource_aws_api_gateway_account.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+}
+
+func resourceAwsApiGatewayAccountRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	log.Printf("[INFO] Reading API Gateway Account %s", d.Id())
+	account, err := conn.GetAccount(&apigateway.GetAccountInput{})
+	if err != nil {
```
stack72: Is there a chance that someone deleting an ApiGatewayAccount (using the AWS Console) could cause an error here? Should we check for a 404 and remove from state if so?

builtin/providers/aws/resource_aws_api_gateway_account.go
```diff
@@ -0,0 +1,119 @@
...
+			// even though that's default settings for new AWS accounts
+			// BadRequestException: The role ARN is not well formed
+			operations = append(operations, &apigateway.PatchOperation{
+				Op:    aws.String("replace"),
+				Path:  aws.String("/cloudwatchRoleArn"),
+				Value: aws.String(arn),
+			})
+		}
+	}
+	input.PatchOperations = operations
```
stack72: If operations is empty, what will happen to the update request?

builtin/providers/aws/resource_aws_api_gateway_account.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Updating API Gateway Account failed: %s", err)
+	}
+	log.Printf("[DEBUG] API Gateway Account updated: %s", out)
+
+	d.SetId("api-gateway-account")
```
stack72: Why does the SetId only happen as part of Update? Does this mean we have no Id before the update?

builtin/providers/aws/resource_aws_api_gateway_account.go
```diff
@@ -0,0 +1,119 @@
...
+	return resourceAwsApiGatewayAccountRead(d, meta)
+}
+
+func resourceAwsApiGatewayAccountDelete(d *schema.ResourceData, meta interface{}) error {
+	// There is no API for "deleting" account or resetting it to "default" settings
```
stack72: Will this leak resources?

builtin/providers/aws/structure.go
```diff
@@ -1097,3 +1097,22 @@ func sortInterfaceSlice(in []interface{}) []interface{} {
 
 	return b
 }
+
+func flattenApiGatewayThrottleSettings(settings *apigateway.ThrottleSettings) []map[string]interface{} {
```
stack72: Would love to see a test for this

stack72:
Few small questions about the PR and 1 nit pick (about adding a test) but that can be fixed later if needed :)

builtin/providers/aws/resource_aws_api_gateway_account.go
radeksimko: There is no way to delete the "account" even through the AWS Console... so no.

builtin/providers/aws/resource_aws_api_gateway_account.go
radeksimko: Technically I only `SetId` to "create" the resource inside tfstate... there cannot be multiple API Gateway "accounts" or anything like that.

Otherwise there can be existing settings, which would be region-wide and this API call overrides that.

Also note that this method is used for both Create & Update.

builtin/providers/aws/resource_aws_api_gateway_account.go
radeksimko: Well technically yes, but there is no way to prevent this on the API level nor in the AWS Console. I can double check but I think that CloudFormation must be doing exactly the same thing with http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-account.html

builtin/providers/aws/structure.go
radeksimko: Good point, I can definitely add one.

builtin/providers/aws/resource_aws_api_gateway_account.go
radeksimko: It will send an empty update API call, which seems to be valid (or at least no error is being returned in such case). I can add a test for it.

builtin/providers/aws/resource_aws_api_gateway_account.go
stack72: This is something we can figure out later. I am just not sure how the resources will affect us - like if there are limits, costs etc.

stack72:
I guess we can merge this for now and then figure out the leaked resources part later? The tests pass (As below) so this LGTM!

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=AWSAPIGatewayAccount_basic'
TF_ACC=1 go test ./builtin/providers/aws -v -run=AWSAPIGatewayAccount_basic -timeout 120m
=== RUN   TestAccAWSAPIGatewayAccount_basic
--- PASS: TestAccAWSAPIGatewayAccount_basic (57.28s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	57.298s
```

builtin/providers/aws/resource_aws_api_gateway_account.go
radeksimko: I believe there are no costs belonging specifically to this resource, but I share your concerns - that's why I raised this with AWS. Until they implement "remove/delete" API method, there isn't much we can do on our side though.

radeksimko:
Let me just add two tests (empty resource with "empty" update API call + `flattenApiGatewayThrottleSettings`) and then you can give it final review & merge.

Hopefully I'll get to it tomorrow.

stack72:
👍

radeksimko:
Both tests added per comments. This is now ready for final review @stack72

stack72:
@radeksimko this is awesome :) Thanks!

PR #6325, opened 2016-04-25
provider/aws: Support eventually consistent aws_security_group_rule 

### LINKED ISSUE #6256: Unable to find matching ingress Security Group Rule
Terraform version: **0.6.14** (latest as of now)
Affected resource: aws_security_group_rule

AWS security group rules are not added to the tfstate file, after correctly being created because they don't appear immediately in the security group check (`DescribeSecurityGroups` request in https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_security_group_rule.go#L160):

Without looking at the debug log level, this is what the user sees (which is rather misleading):

```
* aws_security_group_rule.company_executor_yourkit_in: [WARN] A duplicate Security Group rule was found. This may be
a side effect of a now-fixed Terraform issue causing two security groups with
identical attributes but different source_security_group_ids to overwrite each
other in the state. See https://github.com/hashicorp/terraform/pull/2376 for more
2016/04/20 01:19:26 [DEBUG] /opt/terraform-0.6.14/terraform-provisioner-file: plugin process exited
information and instructions for recovery. Error message: the specified rule "peer: 0.0.0.0/0, TCP, from port: 10001, to port: 10002, ALLOW" already exists
* aws_security_group_rule.company_general_icmp_in: [WARN] A duplicate Security Group rule was found. This may be
a side effect of a now-fixed Terraform issue causing two security groups with
identical attributes but different source_security_group_ids to overwrite each
other in the state. See https://github.com/hashicorp/terraform/pull/2376 for more
information and instructions for recovery. Error message: the specified rule "peer: 0.0.0.0/0, ICMP, type: ALL, code: ALL, ALLOW" already exists
```

The work-around of including all rules in the security group declaration is not workable for us because there doesn't seem to be a smooth migration path - on apply, the security group loses all old independent rules, while the new ones are not added.

Unfortunately our setup is extremely complex and I was unable to reproduce this issue with a simpler setup. My guess is that AWS needs more time or has a race condition so the request should be retried.

Attaching:

* Initial apply log (debug level) where 2 security groups rules are omitted (grule-751804134 and sgrule-2971013440): [security.group.first.apply.log.txt](https://github.com/hashicorp/terraform/files/227139/security.group.first.apply.log.txt)
* The initial tfstate file which does not contain the 2 security groups: [security.group.tfstate.txt](https://github.com/hashicorp/terraform/files/227138/security.group.tfstate.txt)
* Second apply log (debug level) where it bombs with the error above: [security.group.second.apply.log.txt](https://github.com/hashicorp/terraform/files/227133/security.group.second.apply.log.txt)

jrnt30:
Found that there are several instances when AWS API is busy that the SG describe will not return the appropriate set of permissions.  We have observed this taking over 2 minutes in the past. 

- Refactor the IP Permission rule matching logic to be reused 
- Create now retries for up to 5 minutes to get a matching rule set appropriately
- Minor, but added the SG ID into the error message to make it easier to track down conflicting SG rules

NOTE: There are a few things that I could use another set of eyes on or clarification around: 

* The `TestAccAWSSecurityGroupRule_Egress` test fails for me, but it also fails on the master branch without any of my modifications as well. 
* The initial implementation that reused the `resourceAwsSecurityGroupRuleRead` to lookup the item after creation does quite a bit there.  I was unclear why the state file was mapped to a ipPermission with the expandIPPerm, an explicit match was found and then those same values were reset on the state file with the values from the EC2 ip permission.  If this is significant I could not understand why

Fixes #6256


catsby:
Checked this out locally, looks good! All the acceptance tests passed for me, so we'll pull this in. Nice work!

jrnt30:
Great, thanks!

cmlad:
Great!

PR #6328, opened 2016-04-25
provider/aws: Use ID in lookup for AWS KMS Aliases

catsby:
Fixes a bug when using `name_prefix` in the AWS KMS Alias resource; we set the name generated as the ID, but then use the value from `d.Get("name")` for lookup. If we're using a prefix, this value is empty.

Fixes `TestAccAWSKmsAlias_basic` [here](https://travis-ci.org/hashicorp/terraform/jobs/125476425)


stack72:
Nice fix - LGTM!

PR #6329, opened 2016-04-25
Azure does not allow CDN ports to be 0

buzztroll:
The azure tests relating to cdn endpoints (TestAccAzureRMCdnEndpoint_basic etc) are breaking because ARM is not accepting port values of 0.  The following error is received:

statusCode:BadRequest
statusMessage:{"error":{"code":"BadRequest","message":"Invalid port \"0\". Port value must be a number between 1 and 65535."}}

This patch sets the ports for example.com to 443 and 80.


stack72:
@buzztroll thanks so much for the PR here - will merge it as soon as the build goes green!

Paul

PR #6330, opened 2016-04-25
provider-aws: validate cluster_id length for aws_elasticache_cluster

lian:
validate length as described in http://docs.aws.amazon.com/cli/latest/reference/elasticache/create-cache-cluster.html#options


stack72:
Hi @lian 
Thanks so much for the PR - this looks good :)

Paul

PR #6332, opened 2016-04-25
provider/aws: Extends the `aws_elasticache_cluster` validation

stack72:
The validation as part of #6330 was only for length. This PR adds the
rules for alphanumeric, not having -- within, not ending with a - and
that the id must start with a letter.

The PR also adds tests for these rules


jen20:
LGTM pending Travis.

PR #4343, opened 2015-12-16
provider/azure: added load balancer support.

aznashwan:

PR #6335, opened 2016-04-25
[WIP] Azure Load Balancer

buzztroll:
This PR adds an acceptance test to #4343 and ports it to use the new changes in the azure-sdk-for-go.  @aznashwan please check this out and see if it is going in the right direction.  We can add more tests as well.


andreimc:
Do we have an eta on this ? I would love to use terraform for Azure but I am waiting for load balancer support.

buzztroll:
It would be great to get feedback and an example HCL from @aznashwan for this.  Currently I am trying to tease apart the initial implementation to figure out how to properly use it but I am not having much luck.  I may start a new implementation soon.

PR #6336, opened 2016-04-25
Set a log prefix for each plugin and remove go-dynect global log prefix

cbednarski:
For internal plugins we will now see a log entry like this:

    2016/04/25 15:28:29 [DEBUG] plugin: terraform: aws-provider (internal)

Also, logs will no longer be erroneously prefixed with go-dynect.


phinze:
Nice! LGTM

PR #6339, opened 2016-04-25
Warn when an internal plugin is overridden

cbednarski:
When a plugin is found on disk but we _probably_ want to use an internal plugin instead, show a warning. Example (it's pretty spammy):

```
[WARN] /usr/local/bin/terraform-provider-postgresql overrides an internal plugin for postgresql-provider.
...
[WARN] /usr/local/bin/terraform-provisioner-file overrides an internal plugin for file-provisioner.
Error configuring: 2 error(s) occurred:

* Unrecognized remote plugin message: 2|unix|/var/folders/pj/66q7ztvd17v_vgfg8c99gm1m0000gn/T/tf-plugin604337945

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.
* Unrecognized remote plugin message: 2|unix|/var/folders/pj/66q7ztvd17v_vgfg8c99gm1m0000gn/T/tf-plugin647987867

This usually means that the plugin is either invalid or simply
needs to be recompiled to support the latest protocol.
```

- [x] Add documentation page to match

/cc @phinze

### DIFF
--- a/website/source/docs/internals/internal-plugins.html.md
+++ b/website/source/docs/internals/internal-plugins.html.md
@@ -1,0 +1,75 @@
+---
+layout: "docs"
+page_title: "Internal Plugins"
+sidebar_current: "docs-internals-plugins"
+description: |-
+  Terraform includes many popular plugins compiled into the main binary.
+---
+
+# Internal Plugins
+
+Terraform providers and provisioners are provided via plugins. Each plugin provides an implementation for a specific service, such as AWS, or provisioner, such as bash. Plugins are executed as a separate process and communicate with the main Terraform binary over an RPC interface.
+
+# Upgrading From 0.7 and Earlier
+
+In versions of Terraform prior to 0.7, each plugin shipped as a separate binary. In versions of Terraform >= 0.7, all of the official plugins are shipped as a single binary. This saves a lot of disk space and makes downloads faster for you!
+
+However, when you upgrade you will need to manually delete old plugins from disk. You can do this via something like this, depending on where you installed `terraform`:
+
+	rm /usr/local/bin/terraform-*
+
+If you don't do this you will see an error message like the following:
+
+```
+[WARN] /usr/local/bin/terraform-provisioner-file overrides an internal plugin for file-provisioner.
+  If you did not expect to see this message you will need to remove the old plugin.
+  See https://www.terraform.io/docs/internals/plugins.html
+Error configuring: 2 error(s) occurred:
+
+* Unrecognized remote plugin message: 2|unix|/var/folders/pj/66q7ztvd17v_vgfg8c99gm1m0000gn/T/tf-plugin604337945
+
+This usually means that the plugin is either invalid or simply
+needs to be recompiled to support the latest protocol.
+* Unrecognized remote plugin message: 2|unix|/var/folders/pj/66q7ztvd17v_vgfg8c99gm1m0000gn/T/tf-plugin647987867
+
+This usually means that the plugin is either invalid or simply
+needs to be recompiled to support the latest protocol.
+```
+
+## Why Does This Happen?
+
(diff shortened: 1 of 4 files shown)


website/source/docs/internals/internal-plugins.html.md
```diff
@@ -0,0 +1,75 @@
...
+
+This usually means that the plugin is either invalid or simply
+needs to be recompiled to support the latest protocol.
+```
+
+## Why Does This Happen?
+
+In previous versions of Terraform all of the plugins were included in a zip file. For example, when you upgraded from 0.6.12 to 0.6.15, the new version each plugin file would have replaced the old one on disk, and you have ended up with the latest version of each plugin.
+
+Going forward there is only one file in the distrubtion so you will need to perform a one-time cleanup when upgrading from Terraform < 0.7 to Terraform 0.7 or higher.
```
phinze: s/distrubtion/distribution/

website/source/docs/internals/internal-plugins.html.md
```diff
@@ -0,0 +1,75 @@
...
+
+Another property of Go is that it does not support dynamic libraries. It _only_ supports static binaries. This is part of Go's overall design and is the reason why it produces statically-compiled binaries in the first place -- once you have a Go binary for your platform it should _Just Work_.
+
+In other languages, plugins are built using dynamic libraries. Since this is not an option for us in Go we use a network RPC interface instead. This means that each plugin is an idependent program, and instead of communicating via shared memory, the main process communicates with the plugin process over HTTP. When you start Terraform, it identifies the plugin you want to use, finds it on disk, runs the other binary, and does some handshaking to make sure they can talk to eachother (the error you may see after upgrading is a handshake failure in the RPC code).
```
phinze: s/idependent/independent/

phinze: s/eachother/each other/

cbednarski: Derp. Thanks for pointing that out. I got smart and pressed spellcheck in Sublime. :-X

phinze:
Wow great write-up! I think people upgrading will really appreciate this.

One functionality Q: do you think we should allow users to set an env var to silence the warning?

phinze:
Ah yeah I see what you're saying about the spamminess. What about building up a slice of shadowed plugins and spitting out the warning at the end with the list if it's non-empty?

cbednarski:
> One functionality Q: do you think we should allow users to set an env var to silence the warning?
 
Since this is only triggered for a plugin that overrides an internal plugin, I think people will only see this when they are working on Terraform. So if it bothers _you_ we can add an ENV var.

phinze:
> So if it bothers _you_ we can add an ENV var.

Ha! You're probably right. The one legit use case we've seen is users recompiling the AWS provider plugin to pick up an as-yet-unmerged PR on Atlas.

Feel free to punt on this and I can follow up with it if I'm bothered. 😛

cbednarski:
> Ha! You're probably right. The one legit use case we've seen is users recompiling the AWS provider plugin to pick up an as-yet-unmerged PR on Atlas.

If they recompile only that plugin they will see this. I would expect going forward they will just recompile the entirety of Terraform and will not see this. Otherwise, hopefully it serves as a reminder to upgrade when the new version comes out.

I'm happy to add the ENV if there's value but also wary of scope creep. :)

> What about building up a slice of shadowed plugins and spitting out the warning at the end with the list if it's non-empty?

The context is kinda weird here because Discover is called per-plugin, so we would need to track the state in e.g. Config and then pass it back up to the calling context. Imo this is nice to have but I think after the initial fix (removing old plugins) this will not show up very often / ever for most users.

phinze:
> Imo this is nice to have but I think after the initial fix (removing old plugins) this will not show up very often / ever for most users.

This makes sense to me!

LGTM - thanks for picking this up! 💟

PR #6302, opened 2016-04-22
Initial role documentation

mixacha:
This is a work in progress, I wanted to kick off documenting appropriate permissions to run terraform as non-administrative user. This along with datastore cluster support can help development teams within enterprises to build machines without granting administrative privileges.

PR #6340, opened 2016-04-25
provider/vsphere: Document role requirements

jen20:
This moves the information presented as part of #6302 into the provider
documentation which gets published as part of the website.


stack72:
LGTM! pending Travis run

PR #6350, opened 2016-04-26
adding the price classes from the AWS cloud formation docs

d4v3y0rk:
adding the price classes from the AWS cloud formation docs


stack72:
Hi @d4v3y0rk 

thanks for the PR here - I will merge it as soon as the build goes green!

Paul

PR #6352, opened 2016-04-26
fix some more docs issues

d4v3y0rk:
spacing and newlines were messing with the docs rendering. (I think)

and


stack72:
Hi @d4v3y0rk 

thank you again! Loving the fix ups to the docs - will merge when green!

Paul

Issue #5797, opened 2016-03-22
Missing RDS Event Subscription


iceycake:
Looks like we are missing RDS Event Subscription.  Looks like potential feature enhancement?

stack72:
http://docs.aws.amazon.com/cli/latest/reference/rds/create-event-subscription.html

radeksimko:
Totally agreed, if nobody else steps in and I don't have any more important resources to work on, I will pick this one up.

iceycake:
I was planning to work on this but I am pretty busy these few weeks and I still have an open PR needs some more works.  Feel free to take on it. 😃

stack72:
Merged in #6367

PR #4348, opened 2015-12-16
helper/schema: resource level validation functions

phinze:
Provider implementers can now add a ValidateFunc to their resources,
which is yielded an interface that allows access to the config.

This enables validations which require access to more than one attribute
to be created.

Included here is client code implementing validation errors for users
who are mixing VPC / Security Groups improperly on aws_instance.

PR #6367, opened 2016-04-26
provider/aws: Added RDS event subscriptions

bigkraig:
Addresses #5797, I still need to add more test cases and look further into the tags. You can definitely tag these subscriptions based on the AWS Console, you can see them in there, but I haven't seen how to retrieve them from the API.

### DIFF
--- a/builtin/providers/aws/resource_aws_db_event_subscription.go
+++ b/builtin/providers/aws/resource_aws_db_event_subscription.go
@@ -1,0 +1,333 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsDbEventSubscription() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsDbEventSubscriptionCreate,
+		Read:   resourceAwsDbEventSubscriptionRead,
+		Update: resourceAwsDbEventSubscriptionUpdate,
+		Delete: resourceAwsDbEventSubscriptionDelete,
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateDbEventSubscriptionName,
+			},
+			"sns_topic": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"event_categories": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+			"source_ids": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+				// ValidateFunc: validateDbEventSubscriptionSourceIds,
+				// requires source_type to be set, does not seem to be a way to validate this
+			},
+			"source_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"customer_aws_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"tags": tagsSchema(),
+		},
+	}
+}
+
+func resourceAwsDbEventSubscriptionCreate(d *schema.ResourceData, meta interface{}) error {
+	rdsconn := meta.(*AWSClient).rdsconn
+	name := d.Get("name").(string)
+	tags := tagsFromMapRDS(d.Get("tags").(map[string]interface{}))
+
+	sourceIdsSet := d.Get("source_ids").(*schema.Set)
+	sourceIds := make([]*string, sourceIdsSet.Len())
+	for i, sourceId := range sourceIdsSet.List() {
+		sourceIds[i] = aws.String(sourceId.(string))
+	}
+
+	eventCategoriesSet := d.Get("event_categories").(*schema.Set)
+	eventCategories := make([]*string, eventCategoriesSet.Len())
+	for i, eventCategory := range eventCategoriesSet.List() {
+		eventCategories[i] = aws.String(eventCategory.(string))
+	}
+
+	request := &rds.CreateEventSubscriptionInput{
+		SubscriptionName: aws.String(name),
+		SnsTopicArn:      aws.String(d.Get("sns_topic").(string)),
+		Enabled:          aws.Bool(d.Get("enabled").(bool)),
+		SourceIds:        sourceIds,
+		SourceType:       aws.String(d.Get("source_type").(string)),
+		EventCategories:  eventCategories,
+		Tags:             tags,
+	}
+
+	log.Println("[DEBUG] Create RDS Event Subscription:", request)
+
+	_, err := rdsconn.CreateEventSubscription(request)
+	if err != nil {
+		return fmt.Errorf("Error creating RDS Event Subscription %s: %s", name, err)
+	}
+
+	log.Println(
+		"[INFO] Waiting for RDS Event Subscription to be ready")
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"creating"},
+		Target:     []string{"active"},
+		Refresh:    resourceAwsDbEventSubscriptionRefreshFunc(d, meta.(*AWSClient).rdsconn),
+		Timeout:    40 * time.Minute,
+		MinTimeout: 10 * time.Second,
+		Delay:      30 * time.Second, // Wait 30 secs before starting
+	}
+
+	// Wait, catching any errors
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Creating RDS Event Subscription %s failed: %s", d.Id(), err)
+	}
+
+	return resourceAwsDbEventSubscriptionRead(d, meta)
+}
+
+func resourceAwsDbEventSubscriptionRead(d *schema.ResourceData, meta interface{}) error {
+	sub, err := resourceAwsDbEventSubscriptionRetrieve(d.Get("name").(string), meta.(*AWSClient).rdsconn)
+	if err != nil {
+		return fmt.Errorf("Error retrieving RDS Event Subscription %s: %s", d.Id(), err)
+	}
+	if sub == nil {
+		d.SetId("")
+		return nil
+	}
+
+	d.SetId(*sub.CustSubscriptionId)
+	if err := d.Set("name", sub.CustSubscriptionId); err != nil {
+		return err
+	}
+	if err := d.Set("sns_topic", sub.SnsTopicArn); err != nil {
+		return err
+	}
+	if err := d.Set("source_type", sub.SourceType); err != nil {
+		return err
+	}
+	if err := d.Set("enabled", sub.Enabled); err != nil {
+		return err
+	}
+	if err := d.Set("source_ids", flattenStringList(sub.SourceIdsList)); err != nil {
+		return err
+	}
+	if err := d.Set("event_categories", flattenStringList(sub.EventCategoriesList)); err != nil {
+		return err
+	}
+	if err := d.Set("customer_aws_id", sub.CustomerAwsId); err != nil {
+		return err
+	}
+
+	// list tags for resource
+	// set tags
+	conn := meta.(*AWSClient).rdsconn
+	arn := buildRDSEventSubscriptionARN(d.Get("customer_aws_id").(string), d.Id(), meta.(*AWSClient).region)
+	resp, err := conn.ListTagsForResource(&rds.ListTagsForResourceInput{
+		ResourceName: aws.String(arn),
+	})
+
+	if err != nil {
+		log.Printf("[DEBUG] Error retrieving tags for ARN: %s", arn)
+	}
+
+	var dt []*rds.Tag
+	if len(resp.TagList) > 0 {
+		dt = resp.TagList
+	}
+	d.Set("tags", tagsToMapRDS(dt))
+
+	return nil
+}
+
+func resourceAwsDbEventSubscriptionRetrieve(
+	name string, rdsconn *rds.RDS) (*rds.EventSubscription, error) {
+
+	request := &rds.DescribeEventSubscriptionsInput{
+		SubscriptionName: aws.String(name),
+	}
+
+	describeResp, err := rdsconn.DescribeEventSubscriptions(request)
+	if err != nil {
+		if rdserr, ok := err.(awserr.Error); ok && rdserr.Code() == "SubscriptionNotFound" {
+			log.Printf("[WARN] No RDS Event Subscription by name (%s) found", name)
+			return nil, nil
+		}
+		return nil, fmt.Errorf("Error reading RDS Event Subscription %s: %s", name, err)
+	}
+
+	if len(describeResp.EventSubscriptionsList) != 1 {
+		return nil, fmt.Errorf("Unable to find RDS Event Subscription: %#v", describeResp.EventSubscriptionsList)
+	}
+
+	return describeResp.EventSubscriptionsList[0], nil
+}
+
+func resourceAwsDbEventSubscriptionUpdate(d *schema.ResourceData, meta interface{}) error {
+	rdsconn := meta.(*AWSClient).rdsconn
+
+	d.Partial(true)
+	requestUpdate := false
+
+	req := &rds.ModifyEventSubscriptionInput{
+		SubscriptionName: aws.String(d.Id()),
+	}
+
+	if d.HasChange("event_categories") {
+		eventCategoriesSet := d.Get("event_categories").(*schema.Set)
+		req.EventCategories = make([]*string, eventCategoriesSet.Len())
+		for i, eventCategory := range eventCategoriesSet.List() {
+			req.EventCategories[i] = aws.String(eventCategory.(string))
+		}
+		requestUpdate = true
+	}
+
+	if d.HasChange("enabled") {
+		req.Enabled = aws.Bool(d.Get("enabled").(bool))
+		requestUpdate = true
+	}
+
+	if d.HasChange("sns_topic") {
+		req.SnsTopicArn = aws.String(d.Get("sns_topic").(string))
+		requestUpdate = true
+	}
+
+	if d.HasChange("source_type") {
+		req.SourceType = aws.String(d.Get("source_type").(string))
+		requestUpdate = true
+	}
+
+	log.Printf("[DEBUG] Send RDS Event Subscription modification request: %#v", requestUpdate)
+	if requestUpdate {
+		log.Printf("[DEBUG] RDS Event Subscription modification request: %#v", req)
+		_, err := rdsconn.ModifyEventSubscription(req)
+		if err != nil {
+			return fmt.Errorf("Modifying RDS Event Subscription %s failed: %s", d.Id(), err)
+		}
+
+		log.Println(
+			"[INFO] Waiting for RDS Event Subscription modification to finish")
+
+		stateConf := &resource.StateChangeConf{
+			Pending:    []string{"modifying"},
+			Target:     []string{"active"},
+			Refresh:    resourceAwsDbEventSubscriptionRefreshFunc(d, meta.(*AWSClient).rdsconn),
+			Timeout:    40 * time.Minute,
+			MinTimeout: 10 * time.Second,
+			Delay:      30 * time.Second, // Wait 30 secs before starting
+		}
+
+		// Wait, catching any errors
+		_, err = stateConf.WaitForState()
+		if err != nil {
+			return fmt.Errorf("Modifying RDS Event Subscription %s failed: %s", d.Id(), err)
+		}
+		d.SetPartial("event_categories")
+		d.SetPartial("enabled")
+		d.SetPartial("sns_topic")
+		d.SetPartial("source_type")
+	}
+
+	arn := buildRDSEventSubscriptionARN(d.Get("customer_aws_id").(string), d.Id(), meta.(*AWSClient).region)
+	if err := setTagsRDS(rdsconn, d, arn); err != nil {
+		return err
+	} else {
+		d.SetPartial("tags")
+	}
+	d.Partial(false)
+
+	return nil
+}
+
+func resourceAwsDbEventSubscriptionDelete(d *schema.ResourceData, meta interface{}) error {
+	rdsconn := meta.(*AWSClient).rdsconn
+	deleteOpts := rds.DeleteEventSubscriptionInput{
+		SubscriptionName: aws.String(d.Id()),
+	}
+
+	if _, err := rdsconn.DeleteEventSubscription(&deleteOpts); err != nil {
+		rdserr, ok := err.(awserr.Error)
+		if !ok {
+			return fmt.Errorf("Error deleting RDS Event Subscription %s: %s", d.Id(), err)
+		}
+
+		if rdserr.Code() != "DBEventSubscriptionNotFoundFault" {
+			log.Printf("[WARN] RDS Event Subscription %s missing during delete", d.Id())
+			return fmt.Errorf("Error deleting RDS Event Subscription %s: %s", d.Id(), err)
+		}
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"deleting"},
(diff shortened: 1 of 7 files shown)


builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+		return err
+	}
+	if err := d.Set("source_ids", flattenStringList(sub.SourceIdsList)); err != nil {
+		return err
+	}
+	if err := d.Set("event_categories", flattenStringList(sub.EventCategoriesList)); err != nil {
+		return err
+	}
+
+	// Set tags, but theyre not in the response
```
radeksimko: Tags can be queried via separate API, see how we do it for DB instances:
https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_instance.go#L650-L675
The same API seems to be used for many resources, see http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Tagging.html#USER_Tagging.ARN

Also keep in mind we always try to avoid extra API requests for account ID like you can see in https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_instance.go#L975 where it wasn't easily unavoidable. **It is avoidable** for RDS Event Subscriptions though since you have `sub.CustomerAwsId` available. 😺

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return nil
+}
+
+func resourceAwsDbEventSubscriptionRetrieve(
+	d *schema.ResourceData, meta interface{}) (*rds.EventSubscription, error) {
```
radeksimko: Do you think you can make the interface of this function less greedy? 😃 i.e. 
turn this
```
d *schema.ResourceData, meta interface{}
```
into
```
name string, rdsconn *rds.RDS
```

Clean code FTW 🚿 😉

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+	if err := d.Set("name", sub.CustSubscriptionId); err != nil {
+		return err
+	}
+	if err := d.Set("sns_topic", sub.SnsTopicArn); err != nil {
+		return err
+	}
+	if err := d.Set("source_type", sub.SourceType); err != nil {
+		return err
+	}
+	if err := d.Set("enabled", sub.Enabled); err != nil {
```
radeksimko: Typically we ignore errors from most `d.Set` calls unless these are setting non-primitive data structures (i.e. anything except `TypeInt`, `TypeString`, `TypeFloat` and `TypeBool`), but these checks do no harm. 👍

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+	return resourceAwsDbEventSubscriptionRead(d, meta)
+}
+
+func resourceAwsDbEventSubscriptionRead(d *schema.ResourceData, meta interface{}) error {
+	sub, err := resourceAwsDbEventSubscriptionRetrieve(d, meta)
+	if err != nil {
+		return err
+	}
+	if sub == nil {
+		d.SetId("")
```
radeksimko: Would you mind adding `WARN` log message here, like you did below? So we're aware of such things when reading the log.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+		return nil, fmt.Errorf("Unable to find RDS Event Subscription: %#v", describeResp.EventSubscriptionsList)
+	}
+
+	return describeResp.EventSubscriptionsList[0], nil
+}
+
+func resourceAwsDbEventSubscriptionUpdate(d *schema.ResourceData, meta interface{}) error {
+	rdsconn := meta.(*AWSClient).rdsconn
+
+	// check/update tags, missing from go-aws-sdk?
```
radeksimko: Have a look at how we do this in `aws_db_instance`.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+	if d.HasChange("enabled") || d.HasChange("event_categories") ||
+		d.HasChange("sns_topic") || d.HasChange("source_type") {
+		eventCategoriesSet := d.Get("event_categories").(*schema.Set)
+		eventCategories := make([]*string, eventCategoriesSet.Len())
+		for i, eventCategory := range eventCategoriesSet.List() {
+			eventCategories[i] = aws.String(eventCategory.(string))
```
radeksimko: Really a nitpick, but #golang idiomatic version would be something like
```go
for _, eventCategory := range eventCategoriesSet.List() {
	eventCategories = append(eventCategories, aws.String(eventCategory.(string)))
```

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+		// Do we want to wait while it is in "modifying" state?
```
radeksimko: Absolutely ⌛

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"deleting"},
+		Target:     []string{},
+		Refresh:    resourceAwsDbEventSubscriptionRefreshFunc(d, meta),
+		Timeout:    40 * time.Minute,
+		MinTimeout: 10 * time.Second,
+		Delay:      30 * time.Second, // Wait 30 secs before starting
+	}
+	_, err := stateConf.WaitForState()
+	return err
```
radeksimko: Can you make this error more specific like you did the one for Create/Describe? e.g. `fmt.Errorf("Deleting RDS Event Subscription %s failed: %s", d.Id(), err)`.

The errors coming from the API in these cases don't tend to be very descriptive.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func resourceAwsDbEventSubscriptionRefreshFunc(
+	d *schema.ResourceData,
+	meta interface{}) resource.StateRefreshFunc {
```
radeksimko: The function interface here can IMO also be reduced - similar to `resourceAwsDbEventSubscriptionRetrieve`.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return func() (interface{}, string, error) {
+		sub, err := resourceAwsDbEventSubscriptionRetrieve(d, meta)
+
+		if err != nil {
+			log.Printf("Error on retrieving DB Event Subscription when waiting: %s", err)
+			return nil, "", err
+		}
+
+		if sub == nil {
```
radeksimko: Have you actually experienced this edge case? Is it only caused by "eventual consistency"? Should we be actually treating the subscription like it's gone or should we retry in such case?

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+
+		if sub.Status != nil {
+			log.Printf("[DEBUG] DB Event Subscription status for %s: %s", d.Id(), *sub.Status)
+		}
+
+		return sub, *sub.Status, nil
+	}
+}
+
+func validateDbEventSubscriptionName(v interface{}, k string) (ws []string, errors []error) {
```
radeksimko: Nice one! 👍 

We usually keep these in [`validators.go`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/validators.go) and have tests for these in `validators_test.go`.

builtin/providers/aws/resource_aws_db_event_subscription_test.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+
+	return nil
+}
+
+var testAccAWSDBEventSubscriptionConfig = `
+resource "aws_db_event_subscription" "bar" {
+  name = "az"
+  sns_topic = "arn:aws:sns:us-east-1:123456789012:RDS-Event"
```
radeksimko: I'm assuming this is still WIP and it will be a reference to `aws_sns_topic.arn`, correct?

builtin/providers/aws/resource_aws_db_event_subscription_test.go
```diff
@@ -0,0 +1,100 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+var testAccAWSDBEventSubscriptionConfig = `
+resource "aws_db_event_subscription" "bar" {
+  name = "az"
```
radeksimko: This should be more specific, ideally suggesting that this resource comes from terraform acceptance test - e.g. `name = "tf-acc-test-rds-event-subs"`.

It will be helpful when we end up looking for leaked resources if those tests happen to leak for any reason.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	request := &rds.CreateEventSubscriptionInput{
+		SubscriptionName: aws.String(name),
+		SnsTopicArn:      aws.String(d.Get("sns_topic").(string)),
+		Enabled:          aws.Bool(d.Get("enabled").(bool)),
+		SourceIds:        sourceIds,
+		SourceType:       aws.String(d.Get("source_type").(string)),
+		EventCategories:  eventCategories,
+		Tags:             tags,
```
radeksimko: Some fields here are optional, have you tried testing it when these fields are not set? Does it work as expected? Are you sure `d.GetOk` isn't necessary here?

radeksimko:
@bigkraig I left you some comments there, hopefully it helps you in moving this from _WIP_ to _"ready for final review & merge"_ 😉 

The most important ones are regarding tags & retry logic for updates + acceptance tests.
It would be nice to have at least 1 more test which would be testing resource updates.

Also in terms of convenience I think that user would expect the subscription to be `enabled = true` by default, rather than disabled by default.

bigkraig:
@radeksimko thanks for all of the work! I'll get all of these looked at asap, a lot of helpful information in here.

builtin/providers/aws/resource_aws_db_event_subscription.go
bigkraig: I haven't experienced it, and I am not sure what the best thing to do would be here.

builtin/providers/aws/resource_aws_db_event_subscription.go
bigkraig: I'm not familiar with GetOk, does Get fail if the value is undefined?

bigkraig:
@radeksimko We use federated accounts here which causes iam.GetUser to fail and that means I am unable to generate ARNs for RDS resources. This becomes a problem for tag testing, which requires the ARN for making changes.

bigkraig:
I was able to fix the ARN problem by updating the aws-sdk-go library

radeksimko:
Thanks for the other PRs, but I don't think any of them are blockers for this PR, specifically you don't need to be querying API for account ID for this resource, as I mentioned in one of my comments:

> Also keep in mind we always try to avoid extra API requests for account ID like you can see in https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_instance.go#L975 where it wasn't avoidable. **It is avoidable** for RDS Event Subscriptions though since you have `sub.CustomerAwsId` available. 😺

bigkraig:
@radeksimko I don't think its avoidable in resourceAwsDbEventSubscriptionUpdate. I can get the account id from ModifyEventSubscription, but that only happens if something has changed. If only the tags are different I wouldn't have made that call.

radeksimko:
> I can get the account id from ModifyEventSubscription

You can get it in `resourceAwsDbEventSubscriptionRead` and save it to the schema, just create a new field in the schema, e.g. `customer_aws_id` and make it `Computed: true`. Then you can leverage that account ID anywhere else, in other methods.

bigkraig:
@radeksimko done! is this a wrap?

builtin/providers/aws/resource_aws_db_event_subscription.go
bigkraig: Everything works as expected with a minimally defined resource

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,326 @@
...
+			dt = resp.TagList
+		}
+		d.Set("tags", tagsToMapRDS(dt))
+	}
+
+	return nil
+}
+
+func resourceAwsDbEventSubscriptionRetrieve(
+	d *schema.ResourceData, rdsconn *rds.RDS) (*rds.EventSubscription, error) {
```
radeksimko: The only tiny part of `d *schema.ResourceData` this function really needs is the `name` field, so I don't think it's worth passing the whole object. Would you mind replacing `d *schema.ResourceData` with `name string`?

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,326 @@
...
+			Timeout:    40 * time.Minute,
+			MinTimeout: 10 * time.Second,
+			Delay:      30 * time.Second, // Wait 30 secs before starting
+		}
+
+		// Wait, catching any errors
+		_, err = stateConf.WaitForState()
+		if err != nil {
+			return fmt.Errorf("Modifying RDS Event Subscription %s failed: %s", d.Id(), err)
+		}
```
radeksimko: We should also be calling `setPartial` for all fields that have been successfully modified here - i.e. `enabled`, `event_categories`, `sns_topic` and `source_type`, otherwise if an error occurs when setting tags below, these changes might be lost, or rather treated as unsaved.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,326 @@
...
+	if _, err := rdsconn.DeleteEventSubscription(&deleteOpts); err != nil {
+		rdserr, ok := err.(awserr.Error)
+		if !ok {
+			return fmt.Errorf("Error deleting RDS Event Subscription %s: %s", d.Id(), err)
+		}
+
+		if rdserr.Code() != "DBEventSubscriptionNotFoundFault" {
+			return fmt.Errorf("Error deleting RDS Event Subscription %s: %s", d.Id(), err)
+		}
+	}
```
radeksimko: It's probably ok to have the check for resource which was deleted outside of Terraform in two places (Read + Delete), although in vast majority of other resources we only catch this in `Read`. In either case it should be accompanied w/ a WARN log message as it's a non-standard situation.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,326 @@
...
+			"event_categories": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+			"source_ids": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
```
radeksimko: Just FYI it looks like we might be able to make this update-able in the future via http://docs.aws.amazon.com/cli/latest/reference/rds/add-source-identifier-to-subscription.html (or rather the underlying API that command uses). It's absolutely 👌 to keep it `ForceNew` for the initial implementation though. Just FYI.

builtin/providers/aws/resource_aws_db_event_subscription.go
radeksimko: The current behaviour is probably reasonable here, I was more asking about the motivation.

We have [`NotFoundChecks` in `resource.StateChangeConf`](https://github.com/hashicorp/terraform/blob/master/helper/resource/state.go#L57-L60) set to `20` by default - i.e. if this happens more than 20 times, an error is returned, otherwise this function is recalled.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,326 @@
...
+
+		if sub.Status != nil {
+			log.Printf("[DEBUG] DB Event Subscription status for %s: %s", d.Id(), *sub.Status)
+		}
+
+		return sub, *sub.Status, nil
+	}
+}
+
+func buildRDSEventSubscriptionARN(d *schema.ResourceData, meta interface{}) (string, error) {
```
radeksimko: Similar to other functions I called out earlier, would you mind reducing the interface to the only necessary arguments? i.e. 

```go
func buildRDSEventSubscriptionARN(customerAwsId, subscriptionId string) (string) {
```

Reducing these interfaces is not just nitpicking. It makes such functions a lot easier to test (test-ability is important even if we aren't currently testing this function anywhere). It also makes it easier for readers to understand how such functions operate - i.e. if a function accepts 2 strings and returns a string, it is quite obvious that it's going to be some kind of concatenation not involving any API calls or other heavy operations and I can _fairly safely_ assume this without having to look at the actual implementation of it.

website/source/docs/providers/aws/r/db_event_subscription.html.markdown
```diff
@@ -0,0 +1,34 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the DB event subscription.
+* `sns_topic` - (Required) The SNS topic to send events to.
+* `source_ids` - (Optional) A list of identifiers of the event sources for which events will be returned. If not specified, then all sources are included in the response. 
+* `source_type` - (Optional) The type of source that will be generating the events.
+* `event_categories` - (Optional) A list of event categories for a SourceType that you want to subscribe to.
+* `enabled` - (Optional) A boolean flag to enable/disable the subscription.
```
radeksimko: Can we also document here that this option now defaults to `true`?

builtin/providers/aws/resource_aws_db_event_subscription_test.go
```diff
@@ -0,0 +1,179 @@
...
+  event_categories = [
+    "availability",
+    "backup",
+    "creation",
+    "deletion",
+    "maintenance"
+  ]
+  tags {
+    Name = "new-name"
+  }
```
radeksimko: We did have many cases in the past, in other resources where some fields were not updated properly and tests didn't catch this.

If we have a single acceptance test for testing update I think we should try to test as many fields as we can in that single test (not just `tags`) - i.e. can you try to make all fields which are not `ForceNew` different from `testAccAWSDBEventSubscriptionConfig` so we can verify that updates for all/most fields are working as expected?

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,326 @@
...
+				Set:      schema.HashString,
+			},
+			"source_ids": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+				// ValidateFunc: validateDbEventSubscriptionSourceIds,
+				// requires source_type to be set, does not seem to be a way to validate this
```
radeksimko: I think you're right, this is not possible to easily validate yet:
https://github.com/hashicorp/terraform/pull/4348

builtin/providers/aws/resource_aws_db_event_subscription.go
radeksimko: As long as "empty" values work the same way as _undefined_, then it's ok.

`GetOk` is typically used like this:

```go
request := &rds.CreateEventSubscriptionInput{
  RequiredField: aws.String("yada"),
}
...
if v, ok := d.GetOk("optional_field"); ok {
  request.OptionalField = aws.String(v.(string))
}
```

radeksimko:
I left you some (hopefully last) comments there, otherwise this is looking pretty good.

Also I appreciate you squashed the changes into 1 commit 👍

bigkraig:
@radeksimko Got everything addressed. I have had some problems running the update acceptance test, it deletes before it runs the update. I'm not sure what I am doing wrong here, I looked at TestAccAWSRDSCluster_backupsUpdate to compare and I am missing whatever is different.

I was able to verify by running terraform against my own configurations, so its just a matter of fixing the test at this point.

builtin/providers/aws/resource_aws_db_event_subscription.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	if err := d.Set("customer_aws_id", sub.CustomerAwsId); err != nil {
+		return err
+	}
+
+	// list tags for resource
+	// set tags
+	conn := meta.(*AWSClient).rdsconn
+	arn := buildRDSEventSubscriptionARN(d.Get("customer_aws_id").(string), d.Id(), meta.(*AWSClient).region)
+	d.Set("arn", arn)
```
radeksimko: There is no `arn` field defined in the schema. I don't think it necessarily needs to be, but in such case we shouldn't be trying to save the ARN to state.

radeksimko:
Except the ARN, there's just one last thing - probably misunderstanding of my previous comment:

```go
if d.HasChange("enabled") {
	d.SetPartial("enabled")
```

the `SetPartial` function is (in the current version if this PR) _confirming_ the value _should_ be changed at a stage where you cannot be sure whether the update operation is actually going to be successful - i.e. too early.

Better use of that function would be something like this:

```go
if requestUpdate {
	log.Printf("[DEBUG] RDS Event Subscription modification request: %#v", req)
...
	// Wait, catching any errors
	_, err = stateConf.WaitForState()
	if err != nil {
		return fmt.Errorf("Modifying RDS Event Subscription %s failed: %s", d.Id(), err)
	}
	d.SetPartial("event_categories")
	d.SetPartial("enabled")
	d.SetPartial("sns_topic")
	d.SetPartial("...another field...")
}
...
```

i.e. only call `SetPartial` when you're sure that the field was actually updated successfully.

See how we do this in [`aws_elb` for example](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_elb.go#L281-L286).

I will try to run the acceptance tests to see what the issue is there.

radeksimko:
So the acceptance tests need some tuning - possibly a copy-paste error:
```diff
diff --git a/builtin/providers/aws/resource_aws_db_event_subscription_test.go b/builtin/providers/aws/resource_aws_db_event_subscription_test.go
index 5aa5f51..26de043 100644
--- a/builtin/providers/aws/resource_aws_db_event_subscription_test.go
+++ b/builtin/providers/aws/resource_aws_db_event_subscription_test.go
@@ -24,13 +24,13 @@ func TestAccAWSDBEventSubscription_basicUpdate(t *testing.T) {
                                Check: resource.ComposeTestCheckFunc(
                                        testAccCheckAWSDBEventSubscriptionExists("aws_db_event_subscription.bar", &v),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "enabled", "true"),
+                                               "aws_db_event_subscription.bar", "enabled", "true"),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "source_type", "db-instance"),
+                                               "aws_db_event_subscription.bar", "source_type", "db-instance"),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "name", "baz"),
+                                               "aws_db_event_subscription.bar", "name", "tf-acc-test-rds-event-subs"),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "tags.Name", "name"),
+                                               "aws_db_event_subscription.bar", "tags.Name", "name"),
                                ),
                        },
                        resource.TestStep{
@@ -38,11 +38,11 @@ func TestAccAWSDBEventSubscription_basicUpdate(t *testing.T) {
                                Check: resource.ComposeTestCheckFunc(
                                        testAccCheckAWSDBEventSubscriptionExists("aws_db_event_subscription.bar", &v),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "enabled", "false"),
+                                               "aws_db_event_subscription.bar", "enabled", "false"),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "source_type", "db-parameter-group"),
+                                               "aws_db_event_subscription.bar", "source_type", "db-parameter-group"),
                                        resource.TestCheckResourceAttr(
-                                               "aws_db_instance.bar", "tags.Name", "new-name"),
+                                               "aws_db_event_subscription.bar", "tags.Name", "new-name"),
                                ),
                        },
                },
```

then it will work as expected:

```
TF_ACC=1 go test ./builtin/providers/aws -v -run=AWSDBEventSubscription -timeout 120m
=== RUN   TestAccAWSDBEventSubscription_basicUpdate
--- PASS: TestAccAWSDBEventSubscription_basicUpdate (1455.29s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	1455.318s
```

bigkraig:
@radeksimko Done!!! Thanks for all of the help with this.

radeksimko:
Thanks @bigkraig for finishing this. I reran acceptance tests and re-checked the code again and it LGTM.

PR #6370, opened 2016-04-27
AWS ELB Listener: Clarify that ssl_certificate_id must be ARN

bruun:
Just saying `id` is ambiguous, it could be interpreted as the resource ID which will fail with the follow error: `CertificateNotFound: Server Certificate not found for the key: <id>`. The AWS documentation states that the ssl certificate id parameter must be the ARN.


catsby:
Indeed ARN is correct here, thanks!

catsby:
Site is updated, thanks again!

PR #6373, opened 2016-04-27
comment grammar/spelling fix

mdb:


stack72:
Hi @mdb 

Thanks so much for the fix up here - will merge when it is green!

Paul

PR #6374, opened 2016-04-27
Switch from t1.micro to t2.micro

bartoszcisek:
t1.micro has been superseded by t2.micro instance. For some reason
 examples with t1.micro don't work any more. Both types are available
 in free tier plan.


catsby:
Hey @bartoszcisek thanks for the PR. Unfortunately the fix here is a bit more involved, for instance the AMI `ami-408c7f28` doesn't use `hvm` virtualization, so it won't work with `t2.micros`. Also, the examples use `t1`'s because we still wanted to support EC2 Classic, so we'd need a note or call out section that provides the right AMI / instance type for Classic users. 

The acceptance tests don't need to change here either. We have them setup to run nightly in Travis and some of them are specifically designed to be tested in EC2 Classic. Most run in `us-west-2` without classic as is as well, so I'd rather leave them untouched now.

If you're willing to make the suggested changes above let me know, and thanks!

eugene-dounar:
Hi,
I followed "Getting started" but changed ami image to Debian, and it didn't work since it's HVM-only. A note on this issue in the docs would certainly help.

catsby:
Closing for now; I'll revisit the docs and make them t2 compatible, along with a substitute block for EC2 Classic users in the near future

PR #6378, opened 2016-04-27
provider/fastly: Add S3 Log Streaming to Fastly Service

catsby:
Adds streaming logs to an S3 bucket to Fastly Service V1

A go-fastly update in https://github.com/sethvargo/go-fastly/pull/9 is a prerequisite for this, and the commit hash in godeps _may_ change once it's merged, which I'll address after it's committed. Until then the rest should be OK to review


jen20:
LGTM!

PR #6381, opened 2016-04-27
provider/aws: refresh state on SQS Queue not found

### LINKED ISSUE #5578: refresh does not destroy local resources it doesn't find
I think I have noticed this with `terraform refresh` in a variety of ways, but one specific occurrence happens if you delete an SQS queue through the console and then `terraform refresh`. You will get an error like:

    * aws_sqs_queue.queue: AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist for this wsdl version.

I would expect `refresh` to mark this queue as deleted rather than fail because it cannot find the queue. There may be reasons why `refresh` acts the way it does, but in that case there could be a `-force` flag or the like to delete services locally that are known to have been destroyed in the infrastructure itself.

Seemingly `taint` does this, but it also seems like you would have to know the missing service ahead of time and explicitly `taint` each individual service that is found to be missing before refreshing/applying. It would be nice if there were an easier way to destroy local resources.

stack72:
Fixes #5578 

When an SQS queue was deleted from the AWS Console, an error was thrown
to say that the Queue could not be found. This is now fixed to remove
the queue from the state on a specific not found exception

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSSQSQueue' 2>~/tf.log
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSSQSQueue -timeout 120m
=== RUN   TestAccAWSSQSQueue_basic
--- PASS: TestAccAWSSQSQueue_basic (20.52s)
=== RUN   TestAccAWSSQSQueue_redrivePolicy
--- PASS: TestAccAWSSQSQueue_redrivePolicy (18.80s)
=== RUN   TestAccAWSSQSQueue_Policybasic
--- PASS: TestAccAWSSQSQueue_Policybasic (24.17s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	63.516s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_sqs_queue.go
+++ b/builtin/providers/aws/resource_aws_sqs_queue.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/sqs"
 )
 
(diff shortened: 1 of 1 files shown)


jen20:
LGTM

bobbydeveaux:
was about to fix this bug @stack72  - encountering this issue so checked out terraform and was going to fix! Decided to see if someone else had beat me to it, given I'm running 0.6.15..!

Time to upgrade! :)

Thanks for the fix.

bobbydeveaux:
Updated - perfect, thanks @stack72 !! :)

stack72:
:+1:

Issue #5019, opened 2016-02-05
acctests: Allow using profiles for acceptance tests


radeksimko:
This was raised in https://github.com/hashicorp/terraform/pull/4939#issuecomment-180476010 but I was generally thinking about this myself for a while.

I absolutely don't want acceptance tests to be picking up the `default` profile ( :exclamation: ), but I think it would be practical if we use `terraform-acceptance-tests` or something similar for this.

This might affect the current acceptance test setup we have in Travis, but probably shouldn't as we use ChainProvider and if the profile isn't present, ENV vars should be just picked up.

This may also fix https://github.com/hashicorp/terraform/issues/4409

stack72:
@radeksimko I believe this is now solved thanks to @bigkraig and #6383

bigkraig:
i've been running the acceptance tests using AWS_PROFILE

stack72:
Me too :)

PR #6383, opened 2016-04-27
Allow running tests when AWS_PROFILE is set, needed for STS users

bigkraig:
I don't have a great understanding of how the testing system is set up, I just now learned how to run the acceptance tests with `testacc` and `TEST=./builtin/providers/aws` :)

A full test without `TESTARGS` will start to throw errors of no valid credentials, but with this patch you can at least test individual resources.

### DIFF
--- a/builtin/providers/aws/provider_test.go
+++ b/builtin/providers/aws/provider_test.go
@@ -30,11 +30,13 @@ func TestProvider_impl(t *testing.T) {
 }
 
 func testAccPreCheck(t *testing.T) {
-	if v := os.Getenv("AWS_ACCESS_KEY_ID"); v == "" {
-		t.Fatal("AWS_ACCESS_KEY_ID must be set for acceptance tests")
-	}
-	if v := os.Getenv("AWS_SECRET_ACCESS_KEY"); v == "" {
-		t.Fatal("AWS_SECRET_ACCESS_KEY must be set for acceptance tests")
+	if v := os.Getenv("AWS_PROFILE"); v == "" {
+		if v := os.Getenv("AWS_ACCESS_KEY_ID"); v == "" {
+			t.Fatal("AWS_ACCESS_KEY_ID must be set for acceptance tests")
+		}
+		if v := os.Getenv("AWS_SECRET_ACCESS_KEY"); v == "" {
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/provider_test.go
```diff
@@ -30,11 +30,13 @@ func TestProvider_impl(t *testing.T) {
 }
 
 func testAccPreCheck(t *testing.T) {
-	if v := os.Getenv("AWS_ACCESS_KEY_ID"); v == "" {
-		t.Fatal("AWS_ACCESS_KEY_ID must be set for acceptance tests")
-	}
-	if v := os.Getenv("AWS_SECRET_ACCESS_KEY"); v == "" {
-		t.Fatal("AWS_SECRET_ACCESS_KEY must be set for acceptance tests")
+	if v := os.Getenv("AWS_PROFILE"); v == "" {
```
jen20: Is this the normal precedence of profile vs keys set explicitly? If so, this LGTM - if not we should probably reverse it to work the same way as the AWS CLI?

jen20:
Hi @bigkraig! I've left a question inline in the code, but this looks good to me at least in principle!

builtin/providers/aws/provider_test.go
bigkraig: Tested this and it goes with the keys first. I'll make some adjustments

radeksimko:
Related: https://github.com/hashicorp/terraform/issues/5019

bigkraig:
@jen20 I started to work on this and realized it may not be necessary. The client is still going to use the same variables it was using before.

Case 1:
You have all three variables set, no error and client works as normal

Case 2:
Nothing is set, you get unset AWS_ACCESS_KEY_ID error.

Case 3:
You have AWS_PROFILE set only, profile is used

Case 4:
You have AWS_PROFILE unset, normal keys are used

Case 5:
You have AWS_PROFILE set, and only one of the other two variables set, you still get the same unset error

jen20:
Thanks for the clarification @bigkraig! About to merge this - personally I'll find it helpful, and it looks like @radeksimko may also! Thanks for your work!

PR #6384, opened 2016-04-27
provider/aws: Fix Read of AWS CloudWatch Log when Update was called

### LINKED ISSUE #6169: aws_cloudwatch_log_group w/ retention_in_days set results in no arn
When using retention_in_days set on a cloudwatch log group, the arn is not added to the state. If no retention is set this works as expected. This causes problems (failures like above) if trying to use the resulting log group:

```
* Resource 'aws_cloudwatch_log_group.test' does not have attribute 'arn' for variable 'aws_cloudwatch_log_group.test.arn'
```

### Terraform Version
Terraform v0.6.14

### Affected Resource(s)
- aws_cloudwatch_log_group

```hcl
resource "aws_cloudwatch_log_group" "test" {
  name = "test"
  retention_in_days = 7
}
```

### tfstate
```
"resources": {
                "aws_cloudwatch_log_group.cloudtrail": {
                    "type": "aws_cloudwatch_log_group",
                    "primary": {
                        "id": "test",
                        "attributes": {
                            "id": "test",
                            "name": "test",
                            "retention_in_days": "7"
                        }
                    }
                }
            }
```

The state is missing the arn and assuming that is referenced by any other resource it will obviously fail. On the second run (refresh) it will update it and add the arn to the state and work successfully. Verified and without the retention_in_days defined this woks fine.

The arn looks like: `"arn": "arn:aws:logs:us-east-1:XXX:log-group:test:*",`

stack72:
Fixes #6169

The Update func was hitting a return err and when the err was empty, it
was skipping over the subsequent read func

```
ake testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCloudWatchLogGroup' 2>~/tf.log
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCloudWatchLogGroup -timeout 120m
=== RUN   TestAccAWSCloudWatchLogGroup_basic
--- PASS: TestAccAWSCloudWatchLogGroup_basic (14.64s)
=== RUN   TestAccAWSCloudWatchLogGroup_retentionPolicy
--- PASS: TestAccAWSCloudWatchLogGroup_retentionPolicy (23.51s)
=== RUN   TestAccAWSCloudWatchLogGroup_multiple
--- PASS: TestAccAWSCloudWatchLogGroup_multiple (18.58s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	56.747s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudwatch_log_group.go
+++ b/builtin/providers/aws/resource_aws_cloudwatch_log_group.go
@@ -117,14 +117,18 @@ func resourceAwsCloudWatchLogGroupUpdate(d *schema.ResourceData, meta interface{
 			}
 			log.Printf("[DEBUG] Setting retention for CloudWatch Log Group: %q: %s", name, input)
 			_, err = conn.PutRetentionPolicy(&input)
+			if err != nil {
+				return err
+			}
 		} else {
 			log.Printf("[DEBUG] Deleting retention for CloudWatch Log Group: %q", name)
 			_, err = conn.DeleteRetentionPolicy(&cloudwatchlogs.DeleteRetentionPolicyInput{
 				LogGroupName: aws.String(name),
 			})
+			if err != nil {
+				return err
+			}
 		}
-
-		return err
 	}
 
 	return resourceAwsCloudWatchLogGroupRead(d, meta)


jen20:
LGTM!

builtin/providers/aws/resource_aws_cloudwatch_log_group.go
```diff
@@ -117,14 +117,18 @@ func resourceAwsCloudWatchLogGroupUpdate(d *schema.ResourceData, meta interface{
...
+				return err
+			}
 		} else {
 			log.Printf("[DEBUG] Deleting retention for CloudWatch Log Group: %q", name)
 			_, err = conn.DeleteRetentionPolicy(&cloudwatchlogs.DeleteRetentionPolicyInput{
 				LogGroupName: aws.String(name),
 			})
+			if err != nil {
+				return err
+			}
```
radeksimko: I understand what happened here, but the duplication shouldn't be necessary - i.e. this `if err` block can be below **once** instead of twice in each if branch. 😉

PR #5270, opened 2016-02-23
Support standard AWS config in the S3 remote backend.

johnrengelman:
Synchronizes the AWS Credential configuration code with the S3 Remote Backend. This allows the S3 remote backend to understand shared credential files and credential profiles.

PR #6385, opened 2016-04-28
provider/aws: Added sts:GetCallerIdentity to GetAccountId for federated logins

bigkraig:
The STS GetCallerIdentity function works for fetching the account id no matter how you're authenticated.

There appears to be a complicated GetAccountId function in auth_helpers.go that accomplishes the same thing. Since I don't know why that was created I didn't update that use, but I think that may be able to use the same method.


radeksimko:
Thanks for discovering `sts:GetCallerIdentity`! 👍 It looks like AWS is listening to our requests, that's great! 😃 

We could possibly use this to replace `iam:ListRoles` in `auth_helpers.go`, but the caveat of STS is that users can enable/disable STS per regions, so I'm rather thinking we'll make it _one of_ the ways to discover account ID and probably put it in front of `iam:ListRoles` as hacks should always go last. That way we also don't need to introduce any breaking changes and make it work for people that have sts disabled for any reasons.

> There appears to be a complicated `GetAccountId` function in `auth_helpers.go` that accomplishes the same thing.

That method is there to get around differences in environments where Terraform can run:

 - EC2 instance (IAM instance profile)
 - federated IAM roles
 - standard IAM Users

I'm happy to bump the `aws-sdk-go` version but not inclined to use `sts:GetCallerIdentity` as the only method nor inclined to use it _directly_ in each resource. It should be part of `auth_helpers.go`.

In addition to all of the above resources _should not_ be calling `GetAccountId` function directly. The only reason I exported this function was https://github.com/hashicorp/terraform/pull/5270

We should call that function once and save the account ID somewhere (possibly to [`AWSClient`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/config.go#L75) which is passed to each resource via `meta` argument), so that it's cached during the terraform run and resources don't ask API again for what we know already.

bigkraig:
@radeksimko let me know what you think. I added accountid to the AWSClient.

It looks good so far, I ran this against an existing deployment and it added tags to my aws_db_subnet_group.

builtin/providers/aws/auth_helpers_test.go
```diff
@@ -586,9 +587,9 @@ func invalidAwsEnv(t *testing.T) func() {
 	return ts.Close
 }
 
-// getMockedAwsIamApi establishes a httptest server to simulate behaviour
-// of a real AWS' IAM server
-func getMockedAwsIamApi(endpoints []*iamEndpoint) (func(), *iam.IAM) {
+// getMockedAwsIamStsApi establishes a httptest server to simulate behaviour
+// of a real AWS' IAM & STS server
+func getMockedAwsIamStsApi(endpoints []*iamEndpoint) (func(), *iam.IAM, *sts.STS) {
```
radeksimko: Would you mind also adding the actual STS mock responses (for success + error) and two tests testing STS?

1. `iam:GetUser` - 403 (failure) `=>` `sts:GetCallerIdentity` - 200 (success)
2. `iam:GetUser` - 403 (failure) `=>` `sts:GetCallerIdentity` - [any failure](http://docs.aws.amazon.com/STS/latest/APIReference/CommonErrors.html) `=>` `iam:ListRoles` - 200 (successful fallback to ListRoles)

You can use examples of responses from the API reference as mocks: http://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

builtin/providers/aws/config.go
```diff
@@ -185,6 +191,11 @@ func (c *Config) Client() (interface{}, error) {
 		// http://docs.aws.amazon.com/general/latest/gr/sigv4_changes.html
 		usEast1Sess := sess.Copy(&aws.Config{Region: aws.String("us-east-1")})
 
+		account_id, err := GetAccountId(client.iamconn, client.stsconn, cp.ProviderName)
+		if err == nil {
+			client.accountid = account_id
```
radeksimko: Just a nitpick but the convention in this context is camelCase - i.e. `accountId`.

builtin/providers/aws/config.go
```diff
@@ -345,35 +349,31 @@ func (c *Config) ValidateCredentials(iamconn *iam.IAM) error {
 
 // ValidateAccountId returns a context-specific error if the configured account
 // id is explicitly forbidden or not authorised; and nil if it is authorised.
-func (c *Config) ValidateAccountId(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (string, error) {
+func (c *Config) ValidateAccountId(account_id string) error {
```
radeksimko: Again - a nitpick but the convention in this context is also camelCase - i.e. `accountId`.

radeksimko: Again - a nitpick but the convention in this context is also camelCase - i.e. `accountId`.

radeksimko:
I left you some comments there, but overall this looks pretty good, I think it's quite close to to be merged. 👍 😉 

Would you mind separating this into two PRs - one for adding the `sts:GetCallerIdentity` into `GetAccountId` etc. and another one for modifying all the resources that would leverage this new functionality?

bigkraig:
@radeksimko this one is updated to be the PR for sts:GetCallerIdentity only

radeksimko:
The second commit 7f9db15a3ad9884e6c7887694db4148fbd9a71c7 LGTM, but I've got two reservations about the first one with dependencies:

1. There are some conflicts which need resolving
2. I think we'd prefer pinning to tagged versions rather than specific hashes if possible - i.e. `v1.1.21` instead of `v1.1.21-3-ge906984`.

bigkraig:
@radeksimko hows this?

radeksimko:
@bigkraig just one last thing - can you update your `godep` to avoid these two lines in the diff?

```diff
-	"GodepVersion": "v63",
 +	"GodepVersion": "v62",
```

Try `go get -u github.com/tools/godep`.

bigkraig:
@radeksimko done!

PR #4231, opened 2015-12-09
[WIP] Provide ENI for AWS Instance.

johnrengelman:
This is start for support for things like https://github.com/hashicorp/terraform/issues/3105.
Specifically, this address the use case of wanting to provide a known ENI to an AWS instance as it's primary network interface (for example, create an ENI, use it in the route table and boot an instance onto that ENI).

PR #6387, opened 2016-04-28
provider/aws: EC2 instance - multiple private ips

Djuke:
**Feature**: Adds support for multiple private IPs to the instance's first network interface.

**Test**: Passed all `TestAccAWSInstance` acceptance tests.

I wanted to add the `ConflictsWith` attribute to `private_ip` and `private_ips` but it conflicts with the `computed` argument which is still required for `private_ip` and should for `private_ips`.

### DIFF
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -501,6 +501,7 @@ func resourceAwsInstanceRead(d *schema.ResourceData, meta interface{}) error {
 		for _, ni := range instance.NetworkInterfaces {
 			if *ni.Attachment.DeviceIndex == 0 {
 				d.Set("subnet_id", ni.SubnetId)
+				d.Set("private_ips", flattenInstancePrivateIpAddresses(ni.PrivateIpAddresses))
 			}
 		}
 	} else {
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -79,10 +79,20 @@ func resourceAwsInstance() *schema.Resource {
 			},
 
 			"private_ip": &schema.Schema{
-				Type:     schema.TypeString,
+				Type:       schema.TypeString,
+				Optional:   true,
+				ForceNew:   true,
+				Computed:   true,
+				Deprecated: "Please use the `private_ips` field instead",
+			},
+
+			"private_ips": &schema.Schema{
+				Type:     schema.TypeSet,
 				Optional: true,
 				ForceNew: true,
 				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
 			},
 
 			"source_dest_check": &schema.Schema{
@@ -1049,7 +1059,24 @@ func buildAwsInstanceOpts(
 		}
 	}
 
-	if hasSubnet && associatePublicIPAddress {
+	private_ips := d.Get("private_ips").(*schema.Set).List()
+	if len(private_ips) != 0 {
+		ni := &ec2.InstanceNetworkInterfaceSpecification{
+			AssociatePublicIpAddress: aws.Bool(associatePublicIPAddress),
+			DeviceIndex:              aws.Int64(int64(0)),
+			SubnetId:                 aws.String(subnetID),
+			Groups:                   groups,
+			PrivateIpAddresses:       expandPrivateIPAddresses(private_ips),
+		}
+
+		if v := d.Get("vpc_security_group_ids").(*schema.Set); v.Len() > 0 {
+			for _, v := range v.List() {
+				ni.Groups = append(ni.Groups, aws.String(v.(string)))
+			}
+		}
+
+		opts.NetworkInterfaces = []*ec2.InstanceNetworkInterfaceSpecification{ni}
+	} else if hasSubnet && associatePublicIPAddress {
 		// If we have a non-default VPC / Subnet specified, we can flag
 		// AssociatePublicIpAddress to get a Public IP assigned. By default these are not provided.
 		// You cannot specify both SubnetId and the NetworkInterface.0.* parameters though, otherwise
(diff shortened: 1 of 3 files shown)


radeksimko:
Thanks for the PR,
this may conflict with https://github.com/hashicorp/terraform/pull/4231 - at least in git-sense, so I'll just reference it here for any future readers, especially for someone who may later pick that linked PR up, but based on the last comments in that thread it doesn't look like this PR should be waiting for it.

I'll get to the actual code review hopefully soon.

Djuke:
this my conflict with https://github.com/hashicorp/terraform/pull/6761, but it is the same fix

radeksimko:
I just merged your other PR #6761 as that one was easier to review - would you mind rebasing this one & resolving conflicts?

Thanks.

Mil0dV:
Ping. (we'd very much like to see this make 0.7 :))

builtin/providers/aws/resource_aws_instance.go
```diff
@@ -1049,7 +1060,24 @@ func buildAwsInstanceOpts(
 		}
 	}
 
-	if hasSubnet && associatePublicIPAddress {
+	private_ips := d.Get("private_ips").(*schema.Set).List()
+	if len(private_ips) != 0 {
```
radeksimko: I think we'll need to refactor the whole conditional block to keep handling EC2 Classic vs VPC correctly.
I'm still ok with deprecating `private_ip` in favour of `private_ips`, but we will need to add some checks for EC2 Classic && `len(private_ips) > 1` cases. I think we should error out as early as we can in such cases.

See https://github.com/hashicorp/terraform/pull/7568 which may help in making decisions. We need to be prepared for scenarios where we don't have the list of supported platforms though - I'd suggest we 1st try assigning multiple IPs and error out with an informative error message (e.g. `("Assigning %d IPs failed: %s (hint: EC2 Classic accounts don't support multiple private IPs for EC2 instances)", len(ips), err)`).

I'm aware that it may be difficult to test on EC2 Classic and many/most of us run on VPC-only accounts, but we can at least stick to expectations described in AWS docs: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html#differences-ec2-classic-vpc

I recommend carefully reading specifically the following two sections:

 - Multiple private IP addresses
 - Accessing the Internet

website/source/docs/providers/aws/r/instance.html.markdown
```diff
@@ -136,13 +138,14 @@ The following attributes are exported:
...
-* `public_dns` - The public DNS name assigned to the instance. For EC2-VPC, this 
+* `public_dns` - The public DNS name assigned to the instance. For EC2-VPC, this
   is only available if you've enabled DNS hostnames for your VPC
 * `public_ip` - The public IP address assigned to the instance, if applicable. **NOTE**: If you are using an [`aws_eip`](/docs/providers/aws/r/eip.html) with your instance, you should refer to the EIP's address directly and not use `public_ip`, as this field will change after the EIP is attached.
-* `private_dns` - The private DNS name assigned to the instance. Can only be 
-  used inside the Amazon EC2, and only available if you've enabled DNS hostnames 
+* `private_dns` - The private DNS name assigned to the instance. Can only be
+  used inside the Amazon EC2, and only available if you've enabled DNS hostnames
   for your VPC
 * `private_ip` - The private IP address assigned to the instance
```
radeksimko: I'd turn this line into:

```
`private_ip` - Deprecated, use `private_ips[0]` instead
```

website/source/docs/providers/aws/r/instance.html.markdown
```diff
@@ -52,9 +52,11 @@ instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/Use
    If you are within a non-default VPC, you'll need to use `vpc_security_group_ids` instead.
 * `vpc_security_group_ids` - (Optional) A list of security group IDs to associate with.
 * `subnet_id` - (Optional) The VPC Subnet ID to launch in.
-* `associate_public_ip_address` - (Optional) Associate a public ip address with an instance in a VPC.  Boolean value. 
-* `private_ip` - (Optional) Private IP address to associate with the
-     instance in a VPC.
+* `associate_public_ip_address` - (Optional) Associate a public ip address with an instance in a VPC.  Boolean value.
+* `private_ip` - (Optional, Deprecated) Private IP address to associate with the
+     instance in a VPC. This
+     attribute is deprecated, please use the `private_ips` attribute instead.
```
radeksimko: I would omit any description for deprecated arguments - it _should_ be mentioned in docs so people aren't confused why things still work, but we should make it more obvious that things just _won't_ work in the near future.

i.e.

```
* `private_ip` - Deprecated, use `private_ips` instead.
```

radeksimko:
Firstly sorry for the delay in providing feedback.

-----

We don't currently have a deprecation mechanisms for Computed fields unfortunately which may need addressing before we _actually_ remove the old field, otherwise some users may get seriously confused.

The following

```hcl
"${aws_instance.example.private_ip}"
```
would **not** print any warning nor error as `Deprecated` field neither if the field was **missing** in the schema. The second behaviour/bug has something to do with internal schema validation, I'm pretty sure it's logged somewhere under a different issue.

TL;DR What you did with `Deprecated` attribute is the best you can do today. 👍 

------

I left you some comments above. The most important one is EC2 Classic compatibility.

website/source/docs/providers/aws/r/instance.html.markdown
```diff
@@ -52,9 +52,11 @@ instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/Use
...
+* `private_ips` - (Optional) A list of private IP addresses to accociate with the instance's first network interface in a VPC.
```
stack72: `associate` not `accociate`

frankh:
Is there any update on this?

paddycarver:
Hi! This PR has been open for quite some time and waiting response from the community without any activity, so I'm going to close it out. We do this to try to keep our attention on PRs that are actively in-progress, it's not a comment on the code itself. If there is still interest in this PR, please feel free to reply, and we'll re-open this. Thanks!

hiroru:
Is there any update on that, I would like to have this feature available in Terraform.

Issue #6343, opened 2016-04-26
`terraform fmt` always updates source file


joshuaspence:
I am trying to write a `terraform fmt` binding for [arcanist](https://github.com/phacility/arcanist). It seems that running `terraform fmt some_file.tf` will always update the source file. This seems inconsistent with the documentation, which seems to imply that the source file will only be modified with the `-write` flag. Specifically, I am trying to have `terraform fmt some_file.tf` output the modified file to STDOUT rather than overwriting the input file.

mdxp:
I assumed that using -diff like this: `terraform fmt -diff` will only display the diff just like the man page is describing: "Display diffs instead of rewriting files" but in order to get this functionality you have to add the `-write=false` flag that is not very intuitive.

jrnt30:
That's a good point.  It appears they are all mutually exclusive.  The explanation of diff is misleading

dcarley:
The description was copied straight from [gofmt](https://golang.org/src/cmd/gofmt/gofmt.go#L31). I agree, it's confusing - documentation pull requests incoming.

radeksimko:
Can this be closed now, when #6398 was merged or is there anything left to implement/fix?

teamterraform:
Hi all! Sorry for the long silence here.

It looks like the confusing and incorrect documentation that led to this issue have been corrected (and, by now, rewritten again a few times!), so we're going to close this out.

PR #6398, opened 2016-04-28
command/fmt: Improve documentation for -diff and defaults

dcarley:
#### Document -diff doesn't disable -write

As noted in hashicorp/terraform#6343, this description misleadingly suggested that the `-diff` option disables the `-write` option.

This isn't the case and because of the default options (described in c753390) the behaviour of `terraform fmt -diff` is actually the same as `terraform fmt -write -list -diff`.

Replace the "instead of rewriting" description to clarify that.

Documentation in hcl/fmtcmd is corrected in hashicorp/hcl#117 but it's not really necessary to bump the dependency version.

#### Show flag defaults in help text

These were documented on the website but not in the `-help` text. This should help to clarify that you need to pass `-list=false -write=false -diff` if you only want to see diffs.

Accordingly I've replaced the word "disabled" with "always false" in the STDIN special cases so that it matches the terminology used in the defaults and better indicates that it is overridden.

NB: The 3x duplicated defaults and documentation makes me feel uneasy once again. I'm not sure how to solve that, though

### DIFF
--- a/command/fmt.go
+++ b/command/fmt.go
@@ -83,7 +83,7 @@ Options:
 
   -write           Write result to source file instead of STDOUT (disabled if using STDIN)
 
-  -diff            Display diffs instead of rewriting files
+  -diff            Display diffs of formatting changes
 
 `
 	return strings.TrimSpace(helpText)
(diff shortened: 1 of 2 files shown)


stack72:
Hi @dcarley 

this looks good :) Nice one!

P.

PR #6400, opened 2016-04-28
vSphere Disk image name

dkalleg:
We were passing in a disk path of `[datastore] `, which the createDisk
call would create a file named `.vmdk` on the datastore, which is not
the expected behavior.  This make sure that if the user did not pass in
a vmdk path, that we call CreateDisk with an empty string like it
expects.


jen20:
@chrislovecnm this LGTM - any opinions? /cc: @markpeek, @frapposelli?

chrislovecnm:
@jen20 LGTM

chrislovecnm:
@stack72 thanks!

Issue #6366, opened 2016-04-26
Feature:  File resource for vSphere provider


dagnello:
File resource. Pushing a file to a datastore. Not the same as the provisioner "file" command to 
push a file to a vm.

stack72:
This has been implemented in #6401

PR #6401, opened 2016-04-28
Adding File Resource for vSphere provider

dagnello:
Allows for file upload to vSphere at specified location.  This also
includes update for moving or renaming of file resources (in vSphere).
Ref issue #6366.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_file.go
+++ b/builtin/providers/vsphere/resource_vsphere_file.go
@@ -1,0 +1,253 @@
+package vsphere
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/soap"
+	"golang.org/x/net/context"
+)
+
+type file struct {
+	datacenter      string
+	datastore       string
+	sourceFile      string
+	destinationFile string
+}
+
+func resourceVSphereFile() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceVSphereFileCreate,
+		Read:   resourceVSphereFileRead,
+		Update: resourceVSphereFileUpdate,
+		Delete: resourceVSphereFileDelete,
+
+		Schema: map[string]*schema.Schema{
+			"datacenter": {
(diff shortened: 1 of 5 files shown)


builtin/providers/vsphere/resource_vsphere_file.go
```diff
@@ -0,0 +1,254 @@
...
+	datastore       string
+	sourceFile      string
+	destinationFile string
+}
+
+func resourceVSphereFile() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceVSphereFileCreate,
+		Read:   resourceVSphereFileRead,
+		Update: resourceVSphereFileUpdate,
```
stack72: As all pieces of the schema are `ForceNew: true` will this ever be called?

dagnello: hello @stack72 it will be called when "destination_file" is changed.  ForceNew: false is false for that one

stack72: got it - in that case you can exclude `ForceNew: false` from that parameter - that is the default :)

dagnello: ok, will remove it

dagnello:
@stack72 updated

builtin/providers/vsphere/resource_vsphere_file.go
```diff
@@ -0,0 +1,253 @@
...
+		},
+	}
+}
+
+func resourceVSphereFileCreate(d *schema.ResourceData, meta interface{}) error {
+
+	log.Printf("[DEBUG] creating file: %#v", d)
+	client := meta.(*govmomi.Client)
+
+	f := file{
```
stack72: What will happen if any of these are omitted? Will vsphere handle empty/null values?

I think this will need to be handled in the same way as datacenter

stack72:
@dagnello apologies for not picking this last point up. Once fixed I believe we are good to merge

builtin/providers/vsphere/resource_vsphere_file.go
dagnello: that's right these are mandatory, will update to error out if they are not provided.

dagnello:
np, just updated!

chrislovecnm:
@stack72 / @jen20 does this look good to go?

stack72:
This LGTM!

PR #6404, opened 2016-04-28
Make sure references to sub-arguments are unique

bruun:
When the HTML docs are generated, the top-level arguments are automatically wrapped in an `<a>` tag with the argument name as a `name` property. This collides with the manually created `name` tags for sub-arguments in this document. By giving them a unique ID this issue should be resolved. I couldn't get the website to run locally by just running `make dev`, so I haven't verified that the links work as intended with this change.


bruun:
This is probably not needed if https://github.com/hashicorp/terraform/pull/6405 is accepted

alexpbrown:
Haha whoops, we probably should have done some coordination on this :)

stack72:
sorry both! I just merged #6405!

thanks so much for the PR here @bruun - my fault for not checking 

P.

bruun:
Haha this is bound to happen every now and then, better with two patches than none :-)

Good work @alexpbrown, I think your changes are more future-proof and look forward to seeing them deployed

PR #6407, opened 2016-04-28
provider/aws: Updating CloudFront distribution to correct LoggingConfig

### LINKED ISSUE #6390: Failed CloudFront logging_config change stored to state and not refreshed
I had an existing `aws_cloudfront_distribution` resource without `logging_config` configuration.
The when adding a new `aws_s3_bucket` and a referencing `logging_config` block, the Bucket was created but not ready when Terraform tried to configure it for the CloudFront distribution:

```
aws_s3_bucket.logs: Creating...
  arn:              "" => "<computed>"
  bucket:           "" => "foo-logs"
  force_destroy:    "" => "0"
aws_s3_bucket.logs: Creation complete
aws_cloudfront_distribution.app: Modifying...
  logging_config.#:                          "0" => "1"
  logging_config.1955940039.bucket:          "" => "foo-logs"
  logging_config.1955940039.include_cookies: "" => "0"
  logging_config.1955940039.prefix:          "" => "cloudfront"
Error applying plan:

1 error(s) occurred:

* aws_cloudfront_distribution.app: InvalidArgument: The parameter Logging Bucket does not refer to a valid S3 bucket.
        status code: 400, request id: <xxx>

```

This is a familiar case with AWS and it's caching, but the issue is that Terraform anyway **stored the new but failed configuration to the state file** and does not see any changes in the future `plan` or `apply` calls. So it does not refresh the state from AWS.

jrnt30:
- Addresses the issue when local state file has logging_config populated and the user
  disables the configuration via the UI (or in this case an
  application of the TF config).  This will now properly set the
  logging_config during the read operation and identify the state as
  diverging

Fixes #6390

### DIFF
--- a/builtin/providers/aws/cloudfront_distribution_configuration_structure.go
+++ b/builtin/providers/aws/cloudfront_distribution_configuration_structure.go
@@ -124,12 +124,16 @@ func flattenDistributionConfig(d *schema.ResourceData, distributionConfig *cloud
 			return err
 		}
 	}
+
 	if distributionConfig.Logging != nil && *distributionConfig.Logging.Enabled {
 		err = d.Set("logging_config", flattenLoggingConfig(distributionConfig.Logging))
-		if err != nil {
-			return err
-		}
+	} else {
+		err = d.Set("logging_config", schema.NewSet(loggingConfigHash, []interface{}{}))
 	}
+	if err != nil {
+		return err
+	}
+
 	if distributionConfig.Aliases != nil {
 		err = d.Set("aliases", flattenAliases(distributionConfig.Aliases))
 		if err != nil {


builtin/providers/aws/cloudfront_distribution_configuration_structure.go
```diff
@@ -124,12 +124,16 @@ func flattenDistributionConfig(d *schema.ResourceData, distributionConfig *cloud
...
 		}
 	}
+
 	if distributionConfig.Logging != nil && *distributionConfig.Logging.Enabled {
 		err = d.Set("logging_config", flattenLoggingConfig(distributionConfig.Logging))
-		if err != nil {
-			return err
-		}
+	} else {
+		err = d.Set("logging_config", []interface{}{})
```
vancluever: Hey @jrnt30, I think right here, you'd want to do the following instead:

```
err = d.Set("logging_config", schema.NewSet(loggingConfigHash, []interface{}{}))
```

The idea here being that even though it's an empty set you still probably want to pass a `*schema.Set` and the correct hash function.

catsby:
Excellent, thanks!

PR #6410, opened 2016-04-29
provider/openstack: implement neutron security groups and rules

cristicalin:
This WIP patch adds support for neutron security groups and
security group rules. This requires a newer version of the
github.com/rackspace/gophercloud (i built and tested against
current master)

Unlike the nova security group, the neutron security group
is a separate entity from the security group rules, they
are created separately and have different identifiers.

An added benefit is that the neutron implementation allows
an admin to target a specific tenant. In this case you can
have a stack that also provisions the security side of a tenant.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
@@ -1,0 +1,209 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/openstack/networking/v2/extensions/security/rules"
+)
+
+func resourceNetworkingSecGroupRuleV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNetworkingSecGroupRuleV2Create,
+		Read:   resourceNetworkingSecGroupRuleV2Read,
+		Delete: resourceNetworkingSecGroupRuleV2Delete,
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				DefaultFunc: schema.EnvDefaultFunc("OS_REGION_NAME", ""),
+			},
+			"direction": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ethertype": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"port_range_min": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"port_range_max": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"protocol": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"remote_group_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"remote_ip_prefix": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"security_group_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"tenant_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceNetworkingSecGroupRuleV2Create(d *schema.ResourceData, meta interface{}) error {
+
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(d.Get("region").(string))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	portRangeMin := d.Get("port_range_min").(int)
+	portRangeMax := d.Get("port_range_max").(int)
+	protocol := d.Get("protocol").(string)
+
+	if protocol == "" {
+		if portRangeMin != 0 || portRangeMax != 0 {
+			return fmt.Errorf("A protocol must be specified when using port_range_min and port_range_max")
+		}
+	}
+
+	opts := rules.CreateOpts{
+		Direction:      d.Get("direction").(string),
+		EtherType:      d.Get("ethertype").(string),
+		SecGroupID:     d.Get("security_group_id").(string),
+		PortRangeMin:   d.Get("port_range_min").(int),
+		PortRangeMax:   d.Get("port_range_max").(int),
+		Protocol:       d.Get("protocol").(string),
+		RemoteGroupID:  d.Get("remote_group_id").(string),
+		RemoteIPPrefix: d.Get("remote_ip_prefix").(string),
+		TenantID:       d.Get("tenant_id").(string),
+	}
+
+	log.Printf("[DEBUG] Create OpenStack Neutron security group: %#v", opts)
+
+	security_group_rule, err := rules.Create(networkingClient, opts).Extract()
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] OpenStack Neutron Security Group Rule created: %#v", security_group_rule)
+
+	d.SetId(security_group_rule.ID)
+
+	return resourceNetworkingSecGroupRuleV2Read(d, meta)
+}
+
+func resourceNetworkingSecGroupRuleV2Read(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG] Retrieve information about security group rule: %s", d.Id())
+
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(d.Get("region").(string))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	security_group_rule, err := rules.Get(networkingClient, d.Id()).Extract()
+
+	if err != nil {
+		return CheckDeleted(d, err, "OpenStack Security Group Rule")
+	}
+
+	d.Set("protocol", security_group_rule.Protocol)
+	d.Set("port_range_min", security_group_rule.PortRangeMin)
+	d.Set("port_range_max", security_group_rule.PortRangeMax)
+	d.Set("remote_group_id", security_group_rule.RemoteGroupID)
+	d.Set("remote_ip_prefix", security_group_rule.RemoteIPPrefix)
+	d.Set("tenant_id", security_group_rule.TenantID)
+	return nil
+}
+
+func resourceNetworkingSecGroupRuleV2Delete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG] Destroy security group rule: %s", d.Id())
+
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(d.Get("region").(string))
+	if err != nil {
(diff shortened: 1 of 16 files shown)


cristicalin:
The build fails because of an older version of github.com/rackspace/gophercloud in the vendor folder.
I haven't included the updated dependency in this initial pull request as I only submit it for initial comments.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -0,0 +1,162 @@
...
+	d.Set("direction", security_group_rule.Direction)
+	d.Set("ethertype", security_group_rule.EtherType)
+	d.Set("security_group_rule_id", security_group_rule.SecGroupID)
+	d.Set("port_range_min", security_group_rule.PortRangeMin)
+	d.Set("port_range_max", security_group_rule.PortRangeMax)
+	d.Set("protocol", security_group_rule.Protocol)
+	d.Set("remote_group_id", security_group_rule.RemoteGroupID)
+	d.Set("remote_ip_prefix", security_group_rule.RemoteIPPrefix)
+	d.Set("tenant_id", security_group_rule.TenantID)
+	return nil
```
jtopjian: For the above attributes to be Set, they have to be marked as "computed" in the argument/schema definition. Arguments that are required don't have to be set.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -0,0 +1,162 @@
...
+			break
+		}
+
+		httpError, ok := err.(*gophercloud.UnexpectedResponseCodeError)
+		if !ok || httpError.Actual != 409 {
+			return err
+		}
+
+		time.Sleep(time.Second * 2)
+	}
```
jtopjian: I recommend using `StateChangeConf` for these changes. Check out another Neutron resource such as `openstack_networking_router_v2` and search for `waitForRouterActive` and `waitForRouterDelete`. Let me know if you need help implementing that :)

jtopjian:
@cristicalin This is a welcomed addition! 

I've left two comments about some areas to fix up, but in general, this is off to a great start. Let me know when you open a PR to Gophercloud and I can review it and (hopefully) help push it along.

Nice work! 😄

cristicalin:
I've made the suggested changes.

PR to Gophercloud for what ?
The security/groups and security/rules are already in place in master we just need to pull them into terraform. 
I will open a separate pull request to terraform for the refresh, from my tests the current master doesn't appear to cause any issues but I want do some more extensive tests next week to make sure nothing gets horribly or subtly broken.

I'll also work next week on proper unit/acceptance tests and add relevant info to the docs.

jtopjian:
oh! I misunderstood. I thought the tests were failing because you were also patching Gophercloud and using a local dev version. The vendoring is only pulling in packages that are in use, which is why the Neutron security group stuff isn't included yet... Sorry about that!

All sounds good!

cristicalin:
I managed to find the time today to figure out how to properly use godep to satisfy the dependencies.
I won't open a separate pull request for the group and rule deps as they can easily fit into the current PR.

cristicalin:
@jtopjian this should be now in a good shape to merge, please have a look when you can. I'll be happy to make any modifications you suggest.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -0,0 +1,195 @@
...
+				DefaultFunc: schema.EnvDefaultFunc("OS_REGION_NAME", ""),
+			},
+			"direction": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ethertype": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
jtopjian: Ethertype is required in Gophercloud.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -0,0 +1,195 @@
...
+				ForceNew: true,
+			},
+			"port_range_max": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+			"protocol": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
jtopjian: When omitting `protocol`, but specifying a port, I get the following error:

```{"NeutronError": {"message": "Must also specify protocol if port range is given.", "type": "SecurityGroupProtocolRequiredWithPorts", "detail": ""}}```

In a way, this makes `protocol` semi-required.

It might be best to check for this condition in the `Create` function before passing it onto Gophercloud/Neutron. Something like (off the top of my head, may contain errors):

```
portRangeMin := d.Get("port_min_range").(string)
portRangeMax := d.Get("port_min_range").(string)
protocol := d.Get("protocol").(string)

if portRangeMin == "" && portRangeMax == "" && protocol == "" {
  return fmt.Errorf("A protocol must be specified when using port_range_min and port_range_max")
}
```

Comment (jtopjian) [inline builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go]: In addition, maybe the `protocol` should be checked for valid values: `tcp`, `udp`, `icmp`? This isn't done in the `openstack_compute_secgroup_v2` resource, but it's something that should also be added there. Would help to error early before building the resource and causing a mid-apply failure.

File: builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go [hunk 5]
```diff
@@ -0,0 +1,195 @@
...
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	security_group_rule, err := rules.Get(networkingClient, d.Id()).Extract()
+
+	if err != nil {
+		return CheckDeleted(d, err, "OpenStack Security Group Rule")
+	}
+
+	d.Set("ethertype", security_group_rule.EtherType)
```
jtopjian: Since `ethertype` is required, this can be removed.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -0,0 +1,195 @@
...
+	d.Set("protocol", security_group_rule.Protocol)
+	d.Set("port_range_min", security_group_rule.PortRangeMin)
+	d.Set("port_range_max", security_group_rule.PortRangeMax)
+	d.Set("remote_group_id", security_group_rule.RemoteGroupID)
+	d.Set("remote_ip_prefix", security_group_rule.RemoteIPPrefix)
+	d.Set("tenant_id", security_group_rule.TenantID)
```
jtopjian: For all others above this, they need `Computed: true` in their schema in order to be set correctly.

builtin/providers/openstack/resource_openstack_networking_secgroup_v2.go
```diff
@@ -0,0 +1,155 @@
...
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				DefaultFunc: schema.EnvDefaultFunc("OS_REGION_NAME", ""),
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: false,
```
jtopjian: This line can be removed since `Required: true`.

builtin/providers/openstack/resource_openstack_networking_secgroup_v2.go
```diff
@@ -0,0 +1,155 @@
...
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	security_group, err := groups.Get(networkingClient, d.Id()).Extract()
+
+	if err != nil {
+		return CheckDeleted(d, err, "OpenStack Neutron Security group")
+	}
+
+	d.Set("description", security_group.Description)
```
jtopjian: `description` needs `Computed: true` in its schema for this to work.

jtopjian:
@cristicalin Looking really good! I left a few comments, but I'd say they're minor compared to the rest of the work you've done 😄 

One other comment is to not forget to modify `website/source/layouts/openstack.erb` to add the resources to the sidebar. 

Let me know if you have any questions or if something sounds off -- I only quickly reviewed this on a Sunday morning.

jtopjian:
Also, in preparation for merging into Terraform, can you squash everything into two commits:

1. `provider/openstack: Neutron security group resources`
2. `vendor: Updating Gophercloud dependencies`

These can wait until all reviewing is done -- just wanted to give you a heads up on the final step before this gets merged.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
cristicalin: According to http://developer.openstack.org/api-ref-networking-v2-ext.html it is not.
But looking at https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/extensions/security/rules/requests.go#L113 It looks like the gophercloud code checks for a valid ethertype which cannot be void so I'll mark it as required.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
cristicalin: Checking for a valid protocol type is done in the gophercloud code https://github.com/rackspace/gophercloud/blob/master/openstack/networking/v2/extensions/security/rules/requests.go#L125
I don't see a point in doing it again in the terraform code.
I'll add the dependency check though as that one makes sense and is not accounted for in the gophercloud code, though I think a better fix would be in gophercloud.

cristicalin:
I've made the modifications you suggested except for checking the protocol values as they are already checked in the gophercloud code Create function (see my comment on that).

I'll squash the commits as requested when you give a green light for the code.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -83,6 +88,16 @@ func resourceNetworkingSecGroupRuleV2Create(d *schema.ResourceData, meta interfa
 		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
 	}
 
+	portRangeMin := d.Get("port_min_range").(string)
```
jtopjian: It looks like there were a few issues with my rough sketch of this. This should be `d.Get("port_range_min").(int)`

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -83,6 +88,16 @@ func resourceNetworkingSecGroupRuleV2Create(d *schema.ResourceData, meta interfa
...
+	portRangeMax := d.Get("port_min_range").(string)
```
jtopjian: This should be This should be `d.Get("port_range_max").(int)`

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
```diff
@@ -83,6 +88,16 @@ func resourceNetworkingSecGroupRuleV2Create(d *schema.ResourceData, meta interfa
...
+	protocol := d.Get("protocol").(string)
+
+	if protocol == "" {
+		if portRangeMin != "" || portRangeMax != "" {
```
jtopjian: This should be:

```if portRangeMin != 0 || portRangeMax != 0 {```

Comment (jtopjian) [inline builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go]: Github is being a little funny, so I'll repeat the last comments... It looks like the rough sketch I made had some errors. After testing actual code, the following should work:

```
portRangeMin := d.Get("port_range_min").(int)
portRangeMax := d.Get("port_range_max").(int)
protocol := d.Get("protocol").(string)

if protocol == "" {
	if portRangeMin != 0 || portRangeMax != 0 {
```

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2.go
jtopjian: Yes, you're correct. I originally thought that checking within the resource would cause Terraform to error and quit during a pre-apply stage, but I was wrong. All this would do is mask the error from Gophercloud which, as you say, doesn't really have a point.

jtopjian:
@cristicalin Just the one change I noted and we're good to go 😄

cristicalin:
I made the last modifications (also updated the docs about the relevant requirements) and squashed the commits into two.

jtopjian:
Thank you for your time and hard work with this!

PR #6411, opened 2016-04-29
provider/openstack: Instance Doc Examples

jtopjian:
This commit adds several example uses of the
openstack_compute_instance_v2 resource. It also makes a clarification
about booting from volumes and image ids/names.


stack72:
<3 this! Nice addition :)

vmorris:
Thanks @jtopjian 
I'm looking forward to reviewing the content.

Issue #3292, opened 2015-09-21
var-file argument gives invalid format


noahjl:
Using terraform 0.6.3 (on Fedora 22 x64) to deploy concurrent environments to AWS using -state and -var-file arguments.

The setup uses separate file for variables (variables.tf), outputs (outputs.tf) and resources ($resource.tf). 

When deploying without the -var-file argument there is no issues. If I use the -var-file=variables.tf it throws an error that I suspect is a bug or user error. Different file names (i.e. demo.tfvars, demo_variables.tf) don't seem to make a difference.

```
$ terraform plan -var-file=variables.tf
...
2015/09/21 10:23:42 waiting for all plugin processes to complete...
invalid value "variables.tf" for flag -var-file: Error decoding Terraform vars file: root.variable: unknown type to string: ValueTypeObject

The vars file should be in the format of `key = "value"`.
Decoding errors are usually caused by an invalid format.
Usage: terraform plan [options] [dir]

  Generates an execution plan for Terraform.
...
```

Please let me know if any additional info would be helpful.
Thank you
Reviewed Issues #2659, #1084 as they seemed possibly related.

ringods:
@noahjl You can't pass a `*.tf` file as a `var-file` argument.

In Terraform, you have two file types: 
* configuration: `*.tf`
* variables: `*.tfvars`.

You must first declare your variables in e.g. `test.tf` file:
```
variable "test" {}
```

then you can provide a value via a variables file `test.tfvars`:
```
test = "value"
```

and run plan or push with the `var-file` argument:
```
$ terraform plan -var-file=test.tfvars
$ terraform push -var-file=test.tfvars
```

phinze:
@ringods you gave a perfect explanation here - thank you!

mohamaa:
Awesome explanation !!! worked like a charm.

PR #6415, opened 2016-04-29
Tweak docs to properly document .tfvars format

chids:
As reported and described in hashicorp/terraform#3292


stack72:
Hi @chids 

thanks so much for the work here :)

Paul

PR #6418, opened 2016-04-29
Triton NICs

BrianHicks:
Replace networks with NICs. NICs don't have the odd behaviour of having to change networks on every apply.

Tests pass for me, both manually and automated. But I've seen them fail randomly, so if someone could verify that the test strategy looks good before merging I'd appreciate it.

### DIFF
--- a/builtin/providers/triton/resource_machine.go
+++ b/builtin/providers/triton/resource_machine.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/joyent/gosdc/cloudapi"
 )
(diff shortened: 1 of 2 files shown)


builtin/providers/triton/resource_machine.go
```diff
@@ -108,17 +109,53 @@ func resourceMachine() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
-			"networks": {
```
stack72: Hi @BrianHicks 

Dropping `networks` will remove backward compatibility. Therefore we need to either:

1. Have a migration in place (e.g. https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_instance_migrate.go)
or 
2. Leave the field in and add the deprecated tag (https://github.com/hashicorp/terraform/blob/master/builtin/providers/cloudstack/resource_cloudstack_nic.go)

If you go for #2 you need to be able to handle the code either way

BrianHicks: Yes, see line 193. The code handles both but prints a warning because the field is marked as deprecated. I just added `nic` above `network`, which I guess shows up weird in this diff.

jen20: Yup just a weird diff artefact. Well done git!

jen20:
LGTM - thanks @BrianHicks! As an aside we now have the Triton tests running every night now, so this will get a few runs before making it into 0.6.16.

PR #6421, opened 2016-04-29
provider/aws: Fix issue replacing Network ACL Relationship

catsby:
It's possible that during an attempt to replace this
association, the Subnet in question has already been moved to
another ACL. This can happen if you're destroying a network acl
and simultaneously re-associating it's subnet(s) with another
ACL; Terraform may have already re-associated the subnet(s) by
the time we attempt to destroy them, even between the time we
list them and then try to destroy them. In this case, the
association we're trying to replace will no longer exist and
this call will fail. Here we trap that error and fail
gracefully; the association we tried to replace gone, we trust
someone else has taken ownership.

Fixes `TestAccAWSNetworkAcl_SubnetChange` that's been failing for some time


jen20:
LGTM!

PR #6424, opened 2016-04-29
provider/docker: don't crash with empty commands

### LINKED ISSUE #6409: Docker *.tf unexpected EOF crash
### Terraform Version
Terraform v0.6.15

### Affected Resource(s)
- docker_container

```hcl
# Configure the Docker provider
provider "docker" {
    host = "tcp://127.0.0.1:4243/"
}

# Create a container
resource "docker_container" "terra" {
    image = "${docker_image.terrabox.latest}"
    name = "terra"
    command =[""]
    must_run = true
    restart ="always"
}

resource "docker_image" "terrabox" {
    name = "busybox:latest"
}
```

### Panic Output
gist.github.com/JulianSza/57484554cd5a5d83324ec59e14094058

### Expected Behavior
Error message saying `command =[""]` is invalid

### Actual Behavior
Crash (unexpected EOF)

### Steps to Reproduce
`terraform apply`

jen20:
If any of the entries in `commands` on `docker_container` resources was empty, the assertion to string panic'd. Since we can't use ValidateFunc on list elements, we can only really check this at apply time. If any value is nil (resolves to empty string during conversion), we fail with an error prior to creating the container.

Fixes #6409.

### DIFF
--- a/builtin/providers/docker/resource_docker_container_funcs.go
+++ b/builtin/providers/docker/resource_docker_container_funcs.go
@@ -51,6 +51,11 @@ func resourceDockerContainerCreate(d *schema.ResourceData, meta interface{}) err
 
 	if v, ok := d.GetOk("command"); ok {
 		createOpts.Config.Cmd = stringListToStringSlice(v.([]interface{}))
+		for _, v := range createOpts.Config.Cmd {
+			if v == "" {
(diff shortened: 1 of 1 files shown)


phinze:
LGTM

PR #6428, opened 2016-04-30
Encryption should be set to nil when snapshot_id is used for EBS device in AWS launch config 

### LINKED ISSUE #5360: ebs_block_device parameter 'encrypted' evaluated as 'true' even when not set
When attempting to include an ebs_block_device with an encrypted snapshot in an aws_launch_configuration, the following error is received no matter how the ebs_block_device is defined:

_Parameter encrypted is invalid. You cannot specify the encrypted flag if specifying a snapshot id in a block device mapping. Launching EC2 instance failed._

The ebs_block_device gave the same error whether configured any of the following three  ways:

```
 ebs_block_device {
    encrypted = 0
    device_name = "/dev/xvdg"
    volume_size = "50"
    snapshot_id = "snap-1234567"
    delete_on_termination = "true"
  }
```

or

```
 ebs_block_device {
    encrypted = 1
    device_name = "/dev/xvdg"
    volume_size = "50"
    snapshot_id = "snap-1234567"
    delete_on_termination = "true"
  }
```

or 

```
 ebs_block_device {
    device_name = "/dev/xvdg"
    volume_size = "50"
    snapshot_id = "snap-1234567"
    delete_on_termination = "true"
  }
```

When set to '0' tfstate reads the following way:

  ebs_block_device.#:                                "" => "1"
  ebs_block_device.2403880458.delete_on_termination: "" => "1"
  ebs_block_device.2403880458.device_name:           "" => "/dev/xvdg"
  **ebs_block_device.2403880458.encrypted:             "" => "0"**
  ebs_block_device.2403880458.iops:                  "" => "<computed>"
  ebs_block_device.2403880458.snapshot_id:           "" => "snap-1234567"
  ebs_block_device.2403880458.volume_size:           "" => "50"
  ebs_block_device.2403880458.volume_type:           "" => "<computed>"
  ebs_optimized:                                     "" => "<computed>"

When set to '1' or *undefined tfstate shows the encrypted value as '1'.

The AWS API appears to not accept any encrypted value if a snapshot used for the ebs_block_device.

If a snapshot is specified in an ebs_block_device, terraform should ignore the encrypted value and not expose it to the API, which would hopefully allow the autoscaling group to spin up instances successfully.

sarguru:
Encryption should be set to nil when snapshot_id is used for EBS device in AWS launch config  and not false. Fixes #5360

### DIFF
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -240,6 +240,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_security_group":                           resourceAwsSecurityGroup(),
 			"aws_security_group_rule":                      resourceAwsSecurityGroupRule(),
 			"aws_spot_instance_request":                    resourceAwsSpotInstanceRequest(),
+			"aws_spot_fleet_request":                       resourceAwsSpotFleetRequest(),
 			"aws_sqs_queue":                                resourceAwsSqsQueue(),
 			"aws_sns_topic":                                resourceAwsSnsTopic(),
 			"aws_sns_topic_subscription":                   resourceAwsSnsTopicSubscription(),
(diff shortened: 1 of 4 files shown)


bobtfish:
👍

PR #4961, opened 2016-02-02
[WIP] Data-driven Terraform Configuration

apparentlymart:
This is where I'm working on the implementation of the proposal from #4169.

* [x] `DataSource` type in the config module
* [x] Loading of `data` blocks from configuration files
* [ ] Validation of `data` blocks
* Interpolation variables of the form `data.TYPE.NAME.ATTR`, with the same multi-capabilities as resources
  * [x] Parsing (in `config` package)
  * [ ] Resolving (in `terraform` package)
* [x] Introduction of data sources in the main `terraform` module
* [x] New `RefreshData` method on the `ResourceProvider` interface
* [x] `RefreshData` support in `helper/schema`
* Graph node and eval implementations for data sources:
 * [ ] Refresh
 * [ ] Plan
 * [ ] Apply
* [ ] Change resource `count` implementation to allow any non-computed variable, which will then include attributes of any data instances that do not depend on resource creation
* [x] Backward-compatibility shim in `helper/schema`, allowing specific data source implementations to be aliased as resource implementations
  * [x] `terraform_remote_state` data source based on the existing resource, as an initial example
  * [ ] Deprecation warning when using shimmed data sources as resources
* [x] User-oriented documentation

Issue #4169, opened 2015-12-04
Data-driven Terraform Configuration


apparentlymart:
A common design pattern has emerged of using read-only resources like `consul_keys` (when used without explicit `value` arguments), `terraform_remote_state` and `atlas_artifact` to produce Terraform configurations from dynamically-loaded data.

Relatedly, there are also now several "logical resources" that don't even *fetch* data: instead, they just transform a set of inputs into a set of outputs entirely within Terraform. These include `template_file`, `tls_private_key` and `tls_cert_request`.

The current resource workflow requires that a resource must always be "created" before it can be used as a dependency, even though "create" is not a meaningful action for either read-only resources or logical resources. This creates a few different issues, which I will enumerate in the sections that follow.

This proposal is to make data-driven configuration a first-class feature of Terraform, with the goal of making Terraform's behavior more intuitive, predicable and explicit.

----

Problems with Data-as-Resources
==========================

Incomplete Diffs
-------

Consider the following configuration:

```js
resource "consul_keys" "deploy_config" {
    key {
        name = "instance_type"
        path = "myapp/ec2_instance_type"
    }
    key {
        name = "instance_ami"
        path = "myapp/ec2_source_ami"
    }
}

resource "aws_instance" "app_server" {
    instance_type = "${consul_keys.deploy_config.var.instance_type}"
    ami = "${consul_keys.deploy_config.var.instance_ami}"
}
```

Here we use values from Consul to populate some arguments of an `aws_instance`. This results in the following plan:

```
+ aws_instance.app_server
    ami:                      "" => "${consul_keys.deploy_config.var.instance_ami}"
    availability_zone:        "" => "<computed>"
    ebs_block_device.#:       "" => "<computed>"
    ephemeral_block_device.#: "" => "<computed>"
    instance_type:            "" => "${consul_keys.deploy_config.var.instance_type}"
    key_name:                 "" => "<computed>"
    placement_group:          "" => "<computed>"
    private_dns:              "" => "<computed>"
    private_ip:               "" => "<computed>"
    public_dns:               "" => "<computed>"
    public_ip:                "" => "<computed>"
    root_block_device.#:      "" => "<computed>"
    security_groups.#:        "" => "<computed>"
    source_dest_check:        "" => "1"
    subnet_id:                "" => "<computed>"
    tenancy:                  "" => "<computed>"
    vpc_security_group_ids.#: "" => "<computed>"

+ consul_keys.deploy_config
[...]

PR #6430, opened 2016-04-30
Data-driven Terraform Configuration

apparentlymart:
This is where I'm working on the implementation of the proposal from #4169.

(This PR supersedes #4961, and has been rebased onto `dev-0.7` rather than `master` so it can build on the reworked plugin bits and type system changes.)

Since this change spans multiple Terraform layers, the sections that follow summarize the changes in each layer, in the hope of making this changeset easier to review. The PR is broken into a sequence of commits which, as far as possible, change only one layer at a time so that each change can be understood in isolation.

# Configuration (`config` package)

In the `config` layer, data sources are introduced by expanding the existing `Resource` concept with a new field `Mode`, which represents which operations/lifecycle this resource follows:

* `ManagedResourceMode`: previously the only mode; Terraform creates and "owns" this resource, updating its configuration and eventually destroying it.
* `DataResourceMode`: Terraform only reads from this resource

In the configuration language, `resource` blocks map to `ManagedResourceMode` resources and `data` blocks map to `DataResourceMode` resources.

`data` blocks don't permit `provisioner` or `lifecycle` sub-blocks because these concepts do not make sense for a resource that only has a `Refresh` action. Internally, data resources always have an empty `Provisioners` slice and a zero-value `ResourceLifecycle` instance.

A similar extension has been made to `ResourceVariable`, which can now represent both the existing `TYPE.NAME.ATTR` variables *and* the new `data.TYPE.NAME.ATTR` variables, again using a `Mode` field as the discriminator.

Since both traditional resources and data resources are both kinds of resources, they both appear in the `Resources` slice within the configuration struct. The `Resource.Id()` implementation keeps them distinct by adding a `data.` prefix to data resource ids, which is a convention that will continue through to the core layer.

* [x] `ResourceMode` enumeration and `Mode` attribute on `config.Resource`
* [x] Parsing of `data` blocks from configuration files
* [x] Parsing of `data.TYPE.NAME.ATTR` variables and `Mode` attribute on `config.ResourceVariable`

# Core changes

Within core is where we find the biggest divergence of codepaths for managed vs. data resources, since data resources have a simpler lifecycle.

The `ResourceProvider` interface has a new method `DataSources`, which is analogous to `Resources`. The Validate phase is consistent between the two, except that the provider abstraction distinguishes between `ValidateResource` and `ValidateDataSource`, both of which are supported by `EvalValidate` depending on mode.

The remainder of the workflow is completely distinct and handled by two different codepaths, switching on the resource mode inside `terraform/transform_resource.go`.

Even though ultimately data resources support only a "read" operation, the standard plan/apply model is supported by splitting a read into two steps in the `ResourceProvider` interface:

* `ReadDataDiff`: takes the config and returns a diff as if the data resource were being "created", allowing core to know about the data source's computed attributes without actually reading any data.
* `ReadDataApply`: takes the diff, uses it to obtain the configuration attributes, actually loads the data and returns a state.

The important special behavior for data resources is that during the "refresh" walk they will check to see if their config contains computed values, and if it doesn't then the diff/apply steps are run immediately, rather than waiting until the real plan and apply phases. This ensures that non-computed data source attributes can be safely used inside provider configurations, bypassing the chicken-and-egg problems that are caused by computed provider arguments.

A significant difference compared to managed resources is that a data source "read" does not get access to any previous state; we always create an entirely new instance on each refresh. The intended user-facing mental model for data resources is that they are not stateful at all, and we persist them in the on-disk state file only so that `-refresh=false` can act as expected without breaking the rest of the workflow.

* [x] `ResourceProvider` interface changes
* [x] Provider plugin stubs for these new methods
* [x] `EvalValidate` calls appropriate provider validate method based on resource mode.
* [x] `ResourceStateKey` understands how to deal with "orphan" data resources in the state.
* [x] `graphNodeExpandedResource` branches in `EvalTree` to support the different lifecycle for data resources.
  * [x] On refresh, do the diff/apply early if configuration is not computed and update the state
  * [x] On plan, record a creation diff if there is no state already recorded from the previous step
  * [x] On apply, if creation diff is present then produce a populated instance state
* [x] `graphNodeOrphanResource` branches in `EvalTree` to support the different lifecycle for data resources.
  * [x] On both refresh and apply, remove the resource from the state
* [x] Clean up data resource instances from state during `terraform destroy` (or applying a `plan -destroy`)

# `helper/schema` support for data sources

In the `helper/schema` layer, the new map of supported data sources is kept separate from the existing map of supported resources. Data sources use the familiar `schema.Resource` type but with only a `Read` implementation required and `Create`, `Update`, and `Delete` functions forbidden.

The `Read` implementation works in essentially the same way as it does for managed resources, getting access to its configuration attributes via `d.Get(...)` and setting computed attributes with `d.Set(...)`. The only notable differences are that `d.Get(...)` won't return values of computed attributes set on previous runs, and calling `d.SetId(...)` is optional.

To help us migrate existing "logical resources" to instead be data sources, a helper is provided to wrap a data source implementation and shim it to work as a resource implementation. In this case, the `Read` implementation *must* call `d.SetId(...)` in order to meet the expectations of a managed resource implementation.

* [x] `DataSourcesMap` within `helper.Provider`
* [x] Implementations of `DataSources`, `ValidateData`, `ReadDataDiff` and `ReadDataApply`
* [x] Backward-compatibility shim for using data sources as logical resources
  * [x] Deprecation warning when using a resource-shimmed data source

# `provider/terraform`: example remote state data source

As an example to show things working end-to-end, the `terraform_remote_state` resource is transformed into a data source, and the backward-compatibility shim is used to maintain the now-deprecated resource.

* [x] `terraform_remote_state` data source

## Targeting Data Resources

`ResourceAddress` is extended with a `ResourceMode` to handle the distinct managed and data resource namespaces. `data.TYPE.NAME` can be used to target data resources, for consistency with how data resources are referenced elsewhere.

* [x] `ResourceAddress` support for `data.TYPE.NAME` syntax and `ResourceMode`.

## UI Changes

When data resource reads appear in plan output, we show them using a distinct presentation to make it clear that no real infrastructure will be altered by this operation:

terraform-datasources-plan

Since a data resource read is internally just a "create" diff for the resource, this is just some sleight of hand in the UI layer to present it differently.

A "read" diff will appear only if the read operation cannot be completed during the "refresh" phase due to computed configuration.

* [x] Change diff output show "reads" differently
* [x] Hide the "(ID: ...)" suffix when refreshing data sources, since a data source doesn't get an id until after it is "refreshed".

# Other stuff

* [x] User-oriented documentation
* [x] Prevent data resources from being tainted explicitly with `terraform taint` ("tainting" is not meaningful for data resources because they are not created/destroyed.)


apparentlymart:
@jen20 I'm not going to be able to work on this again until at least next weekend so I tried to document where I got to as best I can in case you wanted to take the reins this week.

If you all leave review comments on it then I'm happy to address them when I next get a suitable amount of time, but I'm equally happy for one of you to just make changes directly to this code and finish it up if that helps this get done faster. This branch is in the main repo, so you can feel free to add more commits to it.

Happy to explain what I've done here so far if anything isn't clear.

apparentlymart:
I checked off all the items on my list, so this is now feature-complete according to my original plan.

I have done some ad-hoc manual testing to exercise the various different combinations of computed/non-computed configs, dependent resources, dependent providers, etc.

Unfortunately the one situation that still doesn't seem to work is the very case that this feature was intended to solve:

```hcl
data "null_data_source" "test" {
    # On read, this map gets copied into the computed map attribute "outputs"
    inputs = {
        aws_region = "us-west-2"
    }
}

provider "aws" {
    region = "${data.null_data_source.test.outputs.aws_region}"
}

    instance_type = "t1.foo"
    ami = "ami-abc123"
}
```

In the above configuration, the `aws` provider configuration depends on the data resource. The intended behavior is that the data instance gets created before the `aws` provider is configured. However, this fails with the following error on `terraform plan`:

```

* no error reported by variable "test" is nil
```

I suspect that this is happening in the "validate" walk, where the data instance still wouldn't yet be available, but I can't understand why that would cause this problem for a data resource but fail at a later phase (and with a different error message) with a managed resource.

apparentlymart:
Once again I will be taking a break from this at least until next weekend, so feel free to work on it more in the mean time; I don't want to slow down progress towards getting this in and cutting 0.7.

PR #6434, opened 2016-05-01
Add support for the Base URL endpoint for GitHub Enterprise Support

BetaKevin:
Added logic to support changing the base API URL for GitHub.  Coded so that the value is only used when provided.  Updated documentation.

BaseURL is taken from https://github.com/google/go-github/blob/576fece59c05680a3862785c1f6522eefb6cb0aa/github/github.go#L81

### DIFF
--- a/builtin/providers/github/config.go
+++ b/builtin/providers/github/config.go
@@ -1,6 +1,8 @@
 package github
 
 import (
+	"net/url"
+
 	"github.com/google/go-github/github"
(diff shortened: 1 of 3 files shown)


stack72:
Hi @BetaKevin 

Is this still a WIP?

P.

stack72:
Hi @BetaKevin 

This looks like a nice addition - thanks so much for the PR here

Paul

PR #6437, opened 2016-05-02
[docs] updating t1 -> t2

MansM:
as t1 is kinda ancient


catsby:
Hey @MansM – we've left it as t1 for now because we still support EC2 Classic (as ancient as that may be 😄 )

Do the AMIs in these examples still work on `t2`'s?

catsby:
Closing for now; I'll revisit the docs and make them t2 compatible, along with a substitute block for EC2 Classic users in the near future

Issue #6408, opened 2016-04-28
Previous AWS autoscaling policy definition with min_adjustment_step does not migrate properly to terraform 0.6.15


boboli:
I was using version 0.6.14 with an `aws_autoscaling_policy` defined with a `min_adjustment_step` value of 1. When I upgraded to 0.6.15, I saw the deprecation warning, so I replaced `min_adjustment_step` with `min_adjustment_magnitude` as I understood them to do the same thing, just one name is deprecated.

I then ran `terraform plan` and `terraform apply` to get it to a clean state. Everything succeeds and all is well, but to be safe, I run `terraform plan` again and I see:
```
~ aws_autoscaling_policy.shrink-ten-percent
    min_adjustment_step: "1" => "0"
```

I have tried applying the above plan, and it also succeeds:
```
aws_autoscaling_policy.shrink-ten-percent: Modifying...
  min_adjustment_step: "1" => "0"
aws_autoscaling_policy.shrink-ten-percent: Modifications complete

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

But I can never get a `plan` or `apply` to tell me nothing need to change, even though the .tf config hasn't been modified since.

### Terraform Version
v0.6.15

### Affected Resource(s)
- aws_autoscaling_policy

### Terraform Configuration Files
```hcl
resource "aws_autoscaling_policy" "shrink-ten-percent" {
    name = "shrink-ten-percent"
    scaling_adjustment = -10
    adjustment_type = "PercentChangeInCapacity"
    min_adjustment_magnitude = 1
    autoscaling_group_name = "${aws_autoscaling_group.webservers.name}"
}
```

### Debug Output
I have far too many other resources in a `terraform plan` run to show the debug output.

### Expected Behavior
I expected `terraform plan` to not show any changes, and `terraform apply` to not try to do anything.

### Actual Behavior
`terraform plan` always shows the min_adjustment_step going to 0, and `terraform apply` always successfully applies the change.

### Steps to Reproduce
1. Create aws_autoscaling_policy with terraform v0.6.14, and set a min_adjustment_step. Plan and apply.
2. Upgrade terraform to v0.6.15.
3. Update .tf file to say min_adjustment_magnitude instead of min_adjustment_step.
4. `terraform plan`
5. `terraform apply`
6. Repeat steps 4-5 without ever getting a clean `plan` or `apply` with no changes.

### Important Factoids
[...]

PR #4277, opened 2015-12-12
Add support for "StepScaling" autoscaling policies.

ctiwald:
Amazon added this policy type last spring, but terraform doesn't support it. This PR would implement "StepScaling" policies for autoscaling groups.

This took far more doing than I anticipated. AWS doesn't provide a secondary API endpoint for step policy management. It simply overloads the current policy API, and then switches certain attributes on and off depending on the policy type. I added support for every existing policy attribute, and a fair amount of validation within resource_aws_autoscaling_policy to confirm user input is correct.

The resource is further complicated by a nested Schema which doesn't exist outside the context of these API calls. In all this was surprisingly complicated, but I think I've got something that works. I wouldn't mind a little 3rd party testing -- the edge cases mired me for a while during development.

PR #6440, opened 2016-05-02
provider/aws: Fix issue in upgrading AutoScaling Policy

catsby:
Fixes an issue introduced with the addition of `min_adjustment_magnitude ` and  https://github.com/hashicorp/terraform/pull/4277 deprecation of `min_adjustment_step`. The upgrade path of swapping one for the other results in a reoccurring plan due to how we were handling zero values with `GetOk` (explained https://github.com/hashicorp/terraform/issues/6408#issuecomment-215887212). 

This PR addresses the issue by setting `min_adjustment_step` to `0` if `min_adjustment_magnitude ` is found. The upgrade plan should be silent now; using `min_adjustment_step` still results in the depreciation warning, `min_adjustment_magnitude` corrects the plan and returns no diffs.

Unfortunately the API will set the `MinAdjustmentMagnitude` and `MinAdjustmentStep` to be equal, so setting `MinAdjustmentStep` will result in a `MinAdjustmentMagnitude` value. Users will need to make this config swap, but with this PR the plan will be silent afterwards. 

Fixes https://github.com/hashicorp/terraform/issues/6408

### DIFF
--- a/builtin/providers/aws/resource_aws_autoscaling_policy.go
+++ b/builtin/providers/aws/resource_aws_autoscaling_policy.go
@@ -126,7 +126,7 @@ func resourceAwsAutoscalingPolicyRead(d *schema.ResourceData, meta interface{}) 
 		return nil
 	}
 
-	log.Printf("[DEBUG] Read Scaling Policy: ASG: %s, SP: %s, Obj: %#v", d.Get("autoscaling_group_name"), d.Get("name"), p)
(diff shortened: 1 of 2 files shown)


stack72:
Nice fix - LGTM! I presume we need to put this in the release notes?

catsby:
@stack72 we'll add it to the release notes yeah... unfortunately it's "deprecated so you have to use this new thing _now_" instead of "hey you should use this new thing eventually"

phinze:
LGTM!

PR #6442, opened 2016-05-02
provider/aws: Improve error handling in IAM Server Certificates

catsby:
This PR does 3 things:

- removes an IAM Server Cert from state if it's not found on `READ`, instead of throwing an error. This will correctly prompt to re-create if in a `plan`
- removes the cert from state if it's not found in a `DELETE` (API returns a `NoSuchEntity` error), instead of throwing an error locally and blocking progress
- Extends the timeout for retrying the delete if it's found to be in use. The cert can retain this "in use" status even after and ELB has stopped using it for a period greater than 1 minute. I was unable to reproduce it with a 3 minute timeout

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_server_certificate_test.go
+++ b/builtin/providers/aws/resource_aws_iam_server_certificate_test.go
@@ -51,6 +51,41 @@ func TestAccAWSIAMServerCertificate_name_prefix(t *testing.T) {
 	})
 }
 
+func TestAccAWSIAMServerCertificate_recreate(t *testing.T) {
+	var cert iam.ServerCertificate
+
+	testDestroyCert := func(*terraform.State) error {
+		// reach out and DELETE the Cert
+		conn := testAccProvider.Meta().(*AWSClient).iamconn
+		_, err := conn.DeleteServerCertificate(&iam.DeleteServerCertificateInput{
+			ServerCertificateName: cert.ServerCertificateMetadata.ServerCertificateName,
+		})
+
+		if err != nil {
+			return fmt.Errorf("Error destorying cert in test: %s", err)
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_iam_server_certificate_test.go
```diff
@@ -51,6 +51,41 @@ func TestAccAWSIAMServerCertificate_name_prefix(t *testing.T) {
 	})
 }
 
+func TestAccAWSIAMServerCertificate_recreate(t *testing.T) {
```
phinze: Conventionally I've been calling these `_disappears` tests.

builtin/providers/aws/resource_aws_iam_server_certificate_test.go
```diff
@@ -51,6 +51,41 @@ func TestAccAWSIAMServerCertificate_name_prefix(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccIAMServerCertConfig_random,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCertExists("aws_iam_server_certificate.test_cert", &cert),
+					testAccCheckAWSServerCertAttributes(&cert),
+					testDestroyCert,
+				),
+				ExpectNonEmptyPlan: true,
+			},
```
phinze: In other `_disappears` tests I add a second step w/ a blank config and expect an empty plan, which is a nice extra check that the state is actually removed.

phinze:
LGTM! Couple of optional notes inline on the test.

catsby:
Applied feedback in 4efb067, going to go ahead and pull this in

PR #6429, opened 2016-04-30
A simplifed load balancer resource for AzureRM

buzztroll:
This commit adds a resource for the ARM load balancer.  The ARM LB is
quite complicated and has many different configuration options.  Instead
of trying to implement them all in a potential confusing user experience
this resources exposes the most common uses.  With it a load balancer can
be configured on a single public IP or subnet which can route to a single
backend pool according to one load balancing rule and 1 probe.

This could be expanded to support more features of the LB or live along side of a more sophisticated load balancer implementation as an easy to use option.

PR #6445, opened 2016-05-03
Log message body for ARM API calls

buzztroll:
This PR adds a decorator to the to the Sender object so that the payload
of the REST calls can be logged.  In order to enable this the environment
variable must be set AZURERM_LOG_BODY.

### DIFF
--- a/builtin/providers/azurerm/config.go
+++ b/builtin/providers/azurerm/config.go
@@ -17,6 +17,7 @@ import (
 	mainStorage "github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/hashicorp/terraform/terraform"
 	riviera "github.com/jen20/riviera/azure"
+	"io"
 )
 
 // ArmClient contains the handles to all the specific Azure Resource Manager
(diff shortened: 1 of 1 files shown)


jen20:
Hi @buzztroll! This LGTM - but note we are using two different Azure clients (the official one and my own Riviera), so this will not currently apply to all resources. During development we have been using Charles to proxy traffic to Azure, which may be a useful trick.

jen20:
On second thoughts it might be better to log at the TRACE level and not be conditional on an environment variable - this would match the AWS provider.

buzztroll:
@jen20 makes sense.  I'll change it to trace.

jen20:
Thanks @buzztroll! I'll try to ensure this gets into 0.6.16 as it has definite debugging benefits.

buzztroll:
From what I can tell it seems that the intent of logging in the providers is to have all logs at the debug level and prepend the provider name before it.  Can you help me figure out how to switch this to TRACE?

jen20:
@buzztroll We can use a pattern similar to that used in the AWS provider: https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/config.go#L155-L158. I'll look at doing this before the next release.

stack72:
Hi @buzztroll 

Is this still needed? I didn't realise this PR was here and merged something else that does a similar thing https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/config.go#L71

Paul

buzztroll:
It is fine to close this one but I would great if you could take another look at #6429 :-)

PR #6449, opened 2016-05-03
providers/cloudflare: record can manage apex records

mitchellh:
Apex records must be created by specifying "@" as the name, but this caused Refresh and Delete errors since after create the "@" is stripped from the API results. This modifies the resource to automatically strip this when looking for the record.

Test included.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -926,6 +926,10 @@
 			"Rev": "cb6853d606ea4a12a15ac83cc43503df99fd28fb"
 		},
 		{
+			"ImportPath": "github.com/mitchellh/cloudflare-go",
+			"Rev": "84c7a0993a06d555dbfddd2b32f5fa9b92fa1dc1"
+		},
+		{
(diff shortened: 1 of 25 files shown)


catsby:
LGTM, think we may have a test account record conflict with our nightly acceptance tests to sort out though

mitchellh:
In the course of testing this, @clint found a huge issue with the provider. Due to the way it looks up records, it is capable of finding (and deleting!) incorrect records. This is caused by the fact that it looks up records by name, rather than by unique ID that Cloudflare returns. Upon further investigation, the library that we use and the official Cloudflare Go library don't support getting that ID. I've forked and made a PR to the official Cloudflare repo to do this.

This PR is now updated to use a new lib (my fork of the official Cloudflare Go lib) so that we properly use record IDs. This cleans up the code a lot. All prior tests pass:

2016-05-03 at 9 46 am

catsby:
Checks out 👍

PR #6454, opened 2016-05-03
provider/openstack allow floating ip to target a specific tenant

cristicalin:
Gophercloud library supports specifying a tenant_id in the Create function.

This patch extends the floatingip schema to allow a cloud admin to specify a tenant_id.
Major use cases:
* Q&A for neutron
* managed service where the tenants are not allowed to allocate floating IP addresses themselves due to scarcity or resources or organisational security constraints

The patch also adds relevant documentation information.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_floatingip_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_floatingip_v2.go
@@ -43,6 +43,12 @@ func resourceNetworkingFloatingIPV2() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"tenant_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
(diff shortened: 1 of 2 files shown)


cristicalin:
The CI failure appears unrelated to this patch, is there a way to trigger the CI job again ?

jtopjian:
@cristicalin Sorry for the delay on this. 

Doing a forceful push will retrigger the Travis checks:

```
git push -f -u cristicalin neutron_floatingip_tenant_id
```

If git doesn't want to accept a blank push, something like `git commit --amend` should make it go.

cristicalin:
Sorry for the late reaction to this.
I pushed again and now its validated cleanly.

jtopjian:
LGTM! would it be possible to squash the commits into one? It's good to merge after that. :smile:

cristicalin:
squashed and now it should be good to go

jtopjian:
:+1:

PR #6456, opened 2016-05-03
Recommend bucket versioning for S3 remote state.

clstokes:


stack72:
YAY! Totally agree - great addition :) LGTM pending travis

PR #6467, opened 2016-05-03
[WIP] : Azure Load Balancer with separated resource

buzztroll:
This patch shows what an implementation of the azure load balancer looks like with the resources separated.  Frontend IP, probe, and rule have all been broken out into separate resources thus making 4 resources needed total with the load balancer itself.

Note that doing this causes some potentially strange events.  As an example we cannot delete all of the frontend IPs from the load balancer or azure will send an error.  If the user is trying to change the frontend ip set this will likely result in a delete then an add.  It will be difficult to get that right.

See #6429 #6335 for alternative approaches.  #6335 could be reworked to be simpler but still have nested resources.

@jen20 @phinze any thoughts?


andreimc:
any idea when this would be looked at ?

buzztroll:
@andreimc if you are looking for something usable checkout #6429.  I think that is the go but I wanted to offer this up as an option for breaking apart the resources if terraform proper prefers it that way.

edevil:
Would these resources allow a configuration like this one? https://github.com/hashicorp/terraform/pull/6429#issuecomment-222650832

jen20:
@stack72 and I have now started working on this and will see it through to completion - WIP is at #9199, and we will update that pull request as we go. Thanks for all the work so far, @buzztroll!

PR #6469, opened 2016-05-04
add a height to svg in bnr to fix IE bug

captainill:
A bug was found on consul's website (https://github.com/hashicorp/consul/issues/2017) that applied to the announcment bnr. Terraform.io uses the same markup so I'm submitting the same fix. 

Here's a screen cap of the issue on terraform.io. Nasty bug renders the header nav useless in IE.

screen shot 2016-05-03 at 6 05 18 pm

Verified the fix on the VM as well.


jen20:
Thanks @captainill!

PR #6472, opened 2016-05-04
vendor: Update dependency vmware/govmomi

catsby:
Updates vmware/govmomi to release 0.6.1

cc @chrislovecnm


catsby:
Looks like this introduces some problems with our provider here, the updated lib has changed some things, notably some `int`'s are now `int32`'s and we're getting errors: 

dep_errors

We'll need those fixed before we can merge this... @chrislovecnm now that I've given you a path forward with updating the deps, should I just close this and let you handle the update(s)?

chrislovecnm:
WIP :)

catsby:
Thanks @chrislovecnm – be aware that we just updated our vendors in https://github.com/hashicorp/terraform/pull/6475 due to missing things, so you'll want to build off of or rebase off of master

PR #6483, opened 2016-05-04
vSphere Bumping govmomi version to v0.6.1

chrislovecnm:
This is a version bump for dependency.  Integration tests are passing, but I would appreciate someone else executing the tests as well.  We need more tests.

* updated govmomi with `godep`
* fixed multiple `int32` challenges - pls double check my syntax
* updated `newVM.RemoveDevice` it is removing the files, may want to parameterize this in the future
* updated `CreateDisk(controller, datastore.Reference(), newDiskPath)` requires `datastore.Reference()` now

No updates needed for docs or unit tests.  Did make a comment in the unit tests as we are not checking that needed env vars are set.

Last set of integration tests are running, but they have passed once already.


chrislovecnm:
someone beat me too it :)

Issue #2098, opened 2015-05-27
Add Amazon Elastic MapReduce (AWS) support


cs-dieter-kling:
In terms of provisioning and managing the AWS PaaS infrastructure, it would be very helpful to have EMR support in Terraform. 

As Andrew Langhorn pointed out, AWS GO SDK (https://github.com/awslabs/aws-sdk-go/tree/master/service/emr) already has support for EMR -
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/terraform-tool/uBMcgr9Azgw

scrossan:
+1 to this!

meylor:
Is this on the roadmap for 0.3 or 0.4? Are there thoughts as to how this should be implemented since jobs terminate outside of a ```terraform apply``` run?

jagatveer:
@harshicorp How far is this thing on your road map? +1 on this one from my end too.

sarguru:
+1

AndreaCatalucci:
+1

apparentlymart:
Looking at [the godoc for the EMR client](https://godoc.org/github.com/aws/aws-sdk-go/service/emr) I see a few different concepts here:

* Cluster (`DescribeCluster`, `ListClusters`)
* Instance Group (`AddInstanceGroups`, `ListInstanceGroups`, `ModifyInstanceGroups`, `ListInstanceGroups`)
* Instance (`ListInstances`, `ListInstancesPages`)
* Job Flow (`DescribeJobFlows`, `RunJobFlow`)
* Step (`DescribeStep`, `ListSteps`)

This API is a little opaque to me because most of the objects above don't have the full set of CRUD operations I usually expect in an AWS API (`Create...`, `Describe...`, `Modify...`, `Delete...`). It seems like there's a special workflow here that isn't obvous from the API functions alone, and that perhaps the modifications to some of these objects are implied by modifications to other objects.

Could someone who is familiar with EMR describe the typical usage patterns so we can understand better what would be an appropriate design for a Terraform integration? By this I mean a use-case and then the set of steps you'd take to achieve that use-case.

Usually Terraform support for a new service starts with the most central one or two resources in that service and then builds from there, so as well as understanding the flows, notes on use-case would also be very helpful to understand the priority of the various concepts above.

serainville:
+1

kevin-at-reflexion:
Greetings @apparentlymart.

There definitely are two sort of three EMR use cases:

[...]

PR #6492, opened 2016-05-05
provider/aws: Add AWS EMR dependency (#2098)

blakesmith:
Hey Terraform! I've begun working on a patch set for adding EMR support to the AWS provider (See: https://github.com/hashicorp/terraform/issues/2098), and @phinze suggested I get the `Godeps` change submitted before continuing with work. This change adds the `emr` service from the aws sdk and vendors it. I had some initial trouble with `godeps` but I think I got everything sorted out.

Thanks, we're loving Terraform so far!

Blake


stack72:
Hi @blakesmith 

Thanks so much for the PR here - nice to see the Godep changes added first :)

This LGTM!

P.

PR #6493, opened 2016-05-05
provider/aws: Add agent_version argument to AWS_OPSWORKS_STACK

u2mejc:
This PR adds the omitted OpsWorks Agent version, used when creating Opsworks stacks.

`agent_version` is supported in AWS_OPSWORKS_INSTANCE, but AWS's api limits you to "INHERIT" or a static version number. The API default is to statically set the current version. Adding the `agent_version` argument to `aws_opsworks_stack` allows specifying "LATEST" so that the client can be automatically updated to the latest version.

### DIFF
--- a/website/source/docs/providers/aws/r/opsworks_stack.html.markdown
+++ b/website/source/docs/providers/aws/r/opsworks_stack.html.markdown
@@ -37,6 +37,7 @@ The following arguments are supported:
 * `service_role_arn` - (Required) The ARN of an IAM role that the OpsWorks service will act as.
 * `default_instance_profile_arn` - (Required) The ARN of an IAM Instance Profile that created instances
   will have by default.
+* `agent_version` - (Optional) If set to `"LATEST"`, OpsWorks will automatically install the latest version.
 * `berkshelf_version` - (Optional) If `manage_berkshelf` is enabled, the version of Berkshelf to use.
 * `color` - (Optional) Color to paint next to the stack's resources in the OpsWorks console.
 * `default_availability_zone` - (Optional) Name of the availability zone where instances will be created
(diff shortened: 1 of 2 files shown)


website/source/docs/providers/aws/r/opsworks_stack.html.markdown
```diff
@@ -37,6 +37,7 @@ The following arguments are supported:
 * `service_role_arn` - (Required) The ARN of an IAM role that the OpsWorks service will act as.
 * `default_instance_profile_arn` - (Required) The ARN of an IAM Instance Profile that created instances
   will have by default.
+* `agent_version` - (Optional) If set to `"LATEST"`, OpsWorks will automatically install the latest version.
```
stack72: Is `Latest` the only accepted option?

stack72:
Hi @u2mejc 

Just left a small question about the options of `agent_version`

P.

website/source/docs/providers/aws/r/opsworks_stack.html.markdown
u2mejc: Hi Paul!

The SDK also allows you to pass specific version numbers there too, but that feature is already exposed at the instance level. The stack being the parent of the instance, will set the instance version by default. The reason for the PR is that _automatic updates_ are not allowed at the instance level (and it's not a default).

You can find supported versions for your stack from the cli:

```json
$ aws opsworks describe-agent-versions --stack-id ***************************************

{
    "AgentVersions": [
        {
            "Version": "3421-20150611173115",
            "ConfigurationManager": {
                "Version": "11.10",
                "Name": "Chef"
            }
        },
        {
            "Version": "3422-20150629124612",
            "ConfigurationManager": {
                "Version": "11.10",
                "Name": "Chef"
            }
        },
        {
            "Version": "3424-20150709092700",
            "ConfigurationManager": {
                "Version": "11.10",
                "Name": "Chef"
            }
        },

```

~ Justin 🍻

stack72:
What you've said makes sense :) this LGTM!

PR #6500, opened 2016-05-05
aws: Update docs after #6385 (account ID via sts)

radeksimko:
We may decide to deprecate both `iam:ListRoles` and `iam:GetUser` if `sts` turns out to be reliable, but I would not mention it anywhere in the docs yet.

cc @bigkraig


radeksimko:
Since this is just a tiny documentation change I will merge this myself and won't wait for review, but feel free to call out any specific word/sentence later on.

PR #6501, opened 2016-05-05
provider/aws: AWS API Gateway request parameters json

gthole:
Add the ability to specify request parameters on API Gateway Method and Integration resources, which achieves parity with MethodResponse and IntegrationResponse resources and allows incoming headers and query string parameters to be passed to the backend controller.

This change is a straight extension of the work done in #6344.

### DIFF
--- a/builtin/providers/aws/resource_aws_api_gateway_method.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_method.go
@@ -141,7 +141,7 @@ func resourceAwsApiGatewayMethodUpdate(d *schema.ResourceData, meta interface{})
 	}
 
 	if d.HasChange("request_parameters_in_json") {
-		ops, err := expandApiGatewayMethodResponseParametersJSONOperations(d, "request_parameters_in_json", "requestParameters")
+		ops, err := expandApiGatewayMethodParametersJSONOperations(d, "request_parameters_in_json", "requestParameters")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


fnouama:
👍 🎆  I neeeed it

mgribouski:
👍

builtin/providers/aws/resource_aws_api_gateway_method.go
```diff
@@ -134,6 +140,14 @@ func resourceAwsApiGatewayMethodUpdate(d *schema.ResourceData, meta interface{})
 		operations = append(operations, expandApiGatewayRequestResponseModelOperations(d, "request_models", "requestModels")...)
 	}
 
+	if d.HasChange("request_parameters_in_json") {
+		ops, err := expandApiGatewayMethodResponseParametersJSONOperations(d, "request_parameters_in_json", "requestParameters")
```
radeksimko: I think we should rename the function to make the purpose more accurate:
`expandApiGatewayMethodResponseParametersJSONOperations` => `expandApiGatewayMethodParametersJSONOperations`

(i.e. it is now being used for both _response_ and _request_ parameters so it should be called accordingly)

gthole: :+1:  Good call.

radeksimko: and also there are two error messages inside the body of that function that should be changed accordingly too, so they're reflecting the reality.

gthole: Done :smiley:

radeksimko:
Currently running acceptance tests for API Gateway, but the code is looking good. ⏳

PR #6503, opened 2016-05-05
Use the new AWSClient.accountid when generating ARNs

bigkraig:
With the merge of #6385 we are now able to leverage the one-time collection of the AWS account id rather than trying to fetch it from multiple places in the code.

### DIFF
--- a/builtin/providers/aws/resource_aws_db_instance.go
+++ b/builtin/providers/aws/resource_aws_db_instance.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/rds"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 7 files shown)


radeksimko:
Two notes I have in my mind currently:

1. Resource-by-resource we'll need to double check if there isn't any way around - i.e. is the API _really_ the only way to compose ARN? In some cases the API returns ARN back when raw name is passed into creation/update API call. We may be leverage that instead. I can find examples. 
2. We do need to think about cases where getting account ID may fail - i.e. the `meta.accountid` string will be empty - the `build*` functions should in such cases probably return an error and creation/update of such resource should also fail. Hopefully there will be near zero of such cases, but we should expect that to happen.

I will dive deeper into the code later on and do an actual review, I just wanted to share these thoughts here.

radeksimko:
@bigkraig I have just checked the API reference of RDS and ElastiCache and I can confirm (unfortunately) that there's no way to get the ARNs from those APIs - i.e. I must admit @stack72's original intention was totally reasonable.

That said I will submit a support ticket to AWS and let them pass the feedback down to RDS & ElastiCache teams. We may be able to get rid of all the logic at some point in the future at least.

----

The only thing that's preventing this PR from merging are missing conditions for cases when there's no account id provided. The helper function(s) for building ARN(s) should be returning error in such case(s) and that error should cause creation/update/deletion of a resource to fail with that error message - e.g. `Failed to construct ARN due to a missing AWS Account ID`.

I would consider merging all RDS helper functions into a single one (e.g. `func buildRDSARN(identifier, resourceType, accountid, region string) string`) + have wrapper functions like `buildRDSDbInstanceARN` that would be just calling `buildRDSARN` with the right `resourceType` [per documentation](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Tagging.html#USER_Tagging.ARN).

Having less functions that generate ARNs would also help when fixing https://github.com/hashicorp/terraform/issues/7309 at some point.

Otherwise this is looking good, thank you for taking the time to do such refactoring. 👍

radeksimko:
> The only thing that's preventing this PR from merging are missing conditions for cases when there's no account id provided.

This was actually implemented & merged in #7151 (available in the next release) - I am therefore closing this PR.

Thanks for the interest and sorry it took me initially so long to get back to you.

PR #6504, opened 2016-05-05
Removed port_id from Supported Arguments

DeanPoulin:
When I attempted this:

```hcl
#
# Create the Floating IP for the VM
#
resource "openstack_compute_floatingip_v2" "lnhc-affiliation-search-ws-dev-fip" {
    region = ""
    pool = "risk-internal"
    port_id = "032df78d-3ad6-46e4-aa84-c2d4fdc35307"
}
```

I receive this error:

```bash
Errors:

  * openstack_compute_floatingip_v2.lnhc-affiliation-search-ws-dev-fip: : invalid or unknown key: port_id
```

Based on the code here https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_compute_floatingip_v2.go#L18 it looks like port_id is not an argument allowed to be passed in.

### DIFF
--- a/website/source/docs/providers/openstack/r/networking_floatingip_v2.html.markdown
+++ b/website/source/docs/providers/openstack/r/networking_floatingip_v2.html.markdown
@@ -35,9 +35,6 @@ The following arguments are supported:
 * `pool` - (Required) The name of the pool from which to obtain the floating
     IP. Changing this creates a new floating IP.
 
-* `port_id` - ID of an existing port with at least one IP address to associate with
-this floating IP.
-
 ## Attributes Reference
 
 The following attributes are exported:


jtopjian:
@DeanPoulin Keep in mind there are two floating IP resources:

* `openstack_compute_floatingip_v2`: Uses the Nova API
* `openstack_networking_floatingip_v2`: Uses the Neutron API

While the "compute" floating ip resource does not accept a `port_id`, the "networking" floating ip resource [does](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_floatingip_v2.go#L41).

Does that help?

DeanPoulin:
@jtopjian - You're a life saver. I didn't notice that other Neutron based API.

I'm new to terraform, as of a couple days ago but I do gotta say it's Awesome!!

Basically, I was trying to create a VIP attached to a lb pool and every time I'd try to do that the VIP was getting created with the port in a disconnected state instead of being attached to the Neutron load balancer private IP. Perhaps I'm doing something wrong there. If I have terraform do everything except create the VIP I can go into the OpenStack Horizon UI and manually create the VIP with the same parameters I'm giving to the openstack_lb_vip_v1. Perhaps I need to use openstack_networking_port_v2 to create the port first, give that to the vip create call in the address field.

So I was digging and digging if there was another way to attach the floating IP to the VIP and tried this.

If this works though I will be one happy camper and owe you for pointing me in the right direction! 😀

This PR is accurate though right?

jtopjian:
@DeanPoulin sorry for the late reply. I'm glad to hear you're enjoying Terraform and I hope we can sort out the issue you're seeing. What you're trying to do should certainly be possible. If you're running into issues, it's most likely a bug. 

Have you checked out the example [Complete Load Balancing Stack](https://www.terraform.io/docs/providers/openstack/r/lb_pool_v1.html)? If that's not working for you, could you open an Issue with a sanitized version of the Terraform config you're using?

As for this PR, I hate to be the bearer of bad news, but the doc you're editing is for the [openstack_networking_floatingip_v2](https://www.terraform.io/docs/providers/openstack/r/networking_floatingip_v2.html) resource which most certainly does have a [port_id argument](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_floatingip_v2.go#L41). 

The `openstack_compute_floatingip_v2` resource, which uses the Nova API, does not have a `port_id` argument, and that omission is reflected correctly in the [docs](https://www.terraform.io/docs/providers/openstack/r/compute_floatingip_v2.html) as well.

I'm going to go ahead and close this PR. If you're having difficulties getting a load balancing stackup and running, definitely open an Issue and we can dig into it. 

Let me know if you have any questions :smile:

Issue #6425, opened 2016-04-30
cookies in aws_cloudfront_distribution not optional


ruippeixotog:
I have the following minimal `aws_cloudfront_distribution` resource config:

```
provider "aws" {
  access_key = "<INSERT_VALID_ACCESS_KEY>"
  secret_key = "<INSERT_VALID_SECRET_KEY>"
  region = "us-east-1"
}

resource "aws_cloudfront_distribution" "s3_distribution" {
  origin {
    domain_name = "mybucket.s3.amazonaws.com"
    origin_id = "S3-mybucket"

    s3_origin_config {
      origin_access_identity = ""
    }
  }

  enabled = true

  default_cache_behavior {
    allowed_methods = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
    cached_methods = ["GET", "HEAD"]
    target_origin_id = "S3-mybucket"

    forwarded_values {
      query_string = false
    }

    viewer_protocol_policy = "allow-all"
    min_ttl = 0
    default_ttl = 3600
    max_ttl = 86400
  }

  restrictions {
    geo_restriction {
      restriction_type = "none"
    }
  }

  viewer_certificate {
    cloudfront_default_certificate = true
  }
}
```

Running `terraform plan` (Terraform v0.6.15) with this config causes a panic (full stack trace [here](https://gist.github.com/ruippeixotog/dcb01808a0ad8f4d87857c77cae61152)). I found out that by adding a value for `default_cache_behavior.forwarded_values.cookies` everything worked as expected.

`cookies` is therefore not in accordance in [the Terraform documentation](https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html#cookies), which says the argument is optional. Even if `cookies` should be required, a human-readable error message should be printed in any case.

stack72:
Just a note here that the error occurs in the creation of the Hash for the forwardedValues

```
func forwardedValuesHash(v interface{}) int {
	var buf bytes.Buffer
	m := v.(map[string]interface{})
	buf.WriteString(fmt.Sprintf("%t-", m["query_string"].(bool)))
	if d, ok := m["cookies"]; ok {
		buf.WriteString(fmt.Sprintf("%d-", cookiePreferenceHash(d.(*schema.Set).List()[0].(map[string]interface{}))))
	}
	if d, ok := m["headers"]; ok {
		for _, e := range sortInterfaceSlice(d.([]interface{})) {
[...]

PR #6505, opened 2016-05-05
provider/aws: Require cookies for Cloudfront Distributions

catsby:
A `cookies` value must be sent for CloudFront distributions, even if `none`. The docs are hard to parse but there's a request example that shows the cookies in there and other fields mentioned as not optional, plus you get an error message if you send nothing... 

- http://docs.aws.amazon.com/AmazonCloudFront/latest/APIReference/CreateDistribution.html

This PR fixes the panic reported in https://github.com/hashicorp/terraform/issues/6425 as well, by guarding against a `*schema.Set` that contains just `nil`, in the event that they did not specify a `cookies` block. 

I think there's a core bug there as well, I can report that when this is merged, in that a required field inside a required field inside an optional field of a Resource will not error on plan if omitted


jen20:
LGTM!

Issue #4934, opened 2016-02-01
Cannot pull images from private registry with docker provider


jmprusi:
Hi! 

I've been trying to pull an image from a private registry using resource "docker_image" without success..
- Running `docker pull private/image:latest` from the instance works 
- Running `DOCKER_HOST=host docker pull private/image:latest` from the client works

But it fails when using the docker provider of terraform with a 403

there's way to specify the registry user/password? 

thanks!

superseb:
Used library should be using $HOME/.dockercfg or $HOME/.docker/config.json. Can you provide some more info on the used resources (dummy out the sensitive stuff) and some debug logging?

jmprusi:
I've been looking at the Terraform code.. I see you are using go-dockerclient, but just using an empty instance of "AuthConfiguration" 

https://github.com/hashicorp/terraform/blob/master/builtin/providers/docker/resource_docker_image_funcs.go#L110

There's this go-dockerclient function called "NewAuthConfigurationsFromDockerCfg()" (https://godoc.org/github.com/fsouza/go-dockerclient#NewAuthConfigurationsFromDockerCfg) so I've modified the previous file.. and replaced line 110, with this: 

	splitImage := strings.Split(image, "/")
	auths, _ := dc.NewAuthConfigurationsFromDockerCfg()
	auth := auths.Configs[splitImage[0]]
	if err := client.PullImage(pullOpts, auth); err != nil {
		return fmt.Errorf("Error pulling image %s: %s\n", splitImage[0], err)
	}


And it works fine.. 

There's this issue hashicorp/nomad/issues/660 , that seem to be related.

My example tf file: 

    provider "docker" {
      host = "tcp://${aws_instance.dockertesting.public_dns}:2375/"
    }

    resource "docker_container" "dockerImage" {
      image = "${docker_image.dockerImage.latest}"
      name = "echoapi"
    }

    resource "docker_image" "dockerImage" {
      name = "quay.io/Company/dockerImage"
    }

The provider just returns a 403 when trying to pull the image.. 

    2016/02/02 15:02:25 [ERROR] root: eval: *terraform.EvalRefresh, err: docker_image.DockerImage: Unable to read Docker image into resource: Unable to pull image quay.io/Company/DockerImage: Error pulling image quay.io/Company/DockerImage: Error: Status 403 trying to pull repository Company/DockerImage: "{\"error\": \"Permission Denied\"}"
[...]

PR #6507, opened 2016-05-05
providers/docker Add docker_registry

xsellier:
To fix this issue: https://github.com/hashicorp/terraform/issues/4934

Commit message:
- Add docker_registry resource
- Add tests for docker_registry
- Add documentation for docker_registry
- Add link between docker_image and docker_registry
- Add test for docker_image
- Add documentation for docker_image
- Fix issue with docker_image (repository parsing)
- TODO: Manage more than one manual docker_registry configuration

### DIFF
--- a/website/source/docs/providers/docker/r/image.html.markdown
+++ b/website/source/docs/providers/docker/r/image.html.markdown
@@ -16,9 +16,19 @@ them.
 ## Example Usage
 
 ```
+# Log to a custom docker registry
+resource "docker_registry" "ubuntu_private" {
+    auth = {
+        username = "John"
+        password = "Doe"
+        server_address = "https://hub.docker.com/"
+    }
+}
+
 # Find the latest Ubuntu precise image.
 resource "docker_image" "ubuntu" {
     name = "ubuntu:precise"
+    registry = "${docker_registry.ubuntu_private.configurations}"
 }
 
 # Access it somewhere else with ${docker_image.ubuntu.latest}
@@ -36,6 +46,7 @@ The following arguments are supported:
 * `keep_locally` - (Optional, boolean) If true, then the Docker image won't be
   deleted on destroy operation. If this is false, it will delete the image from
   the docker local storage on destroy operation.
+* `registry` - (Optional, string) The only way to get this value is to use the `docker_registry` resource as is shown in the example above.
 
 ## Attributes Reference
 
--- a/website/source/docs/providers/docker/r/image.html.markdown
+++ b/website/source/docs/providers/docker/r/image.html.markdown
@@ -18,7 +18,7 @@ them.
 ```
 # Log to a custom docker registry
 resource "docker_registry" "ubuntu_private" {
-    auth = {
+    auths = {
         username = "John"
         password = "Doe"
         server_address = "https://hub.docker.com/"
--- a/website/source/docs/providers/docker/r/image.html.markdown
+++ b/website/source/docs/providers/docker/r/image.html.markdown
@@ -18,7 +18,7 @@ them.
 ```
 # Log to a custom docker registry
 resource "docker_registry" "ubuntu_private" {
-    auths = {
+    auth = {
         username = "John"
         password = "Doe"
         server_address = "https://hub.docker.com/"
--- a/website/source/docs/providers/docker/r/registry.html.mdown
+++ b/website/source/docs/providers/docker/r/registry.html.mdown
@@ -1,0 +1,55 @@
+---
+layout: "docker"
+page_title: "Docker: docker_registry"
+sidebar_current: "docs-docker-resource-registry"
+description: |-
+  Logs to a Docker registry, downloads and exports a Docker image.
+---
+
+# docker\_registry
+
+Used to declare a private docker registry.
+This can be used alongside [docker\_image](/docs/providers/docker/r/image.html)
+to programmatically use the credentials without having to repeat them.
+
+## Example Usage
+
+```
+# Log to a custom docker registry
+resource "docker_registry" "ubuntu_private" {
+    settings_file = "~/.docker/config.json"
+}
+
+# Find the latest Ubuntu precise image.
+resource "docker_image" "ubuntu" {
+    name = "ubuntu:precise"
+    registry = "${docker_registry.ubuntu_private.configurations}"
+}
+
+# Access it somewhere else with ${docker_image.ubuntu.latest}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `settings_file` - (Optional, string) Path to the docker configuration file.
+* `auth` - (Optional, block) See [Auth](#auth) below for details.
+
+<a id="auth"></a>
+### Auth
+
+`auth` is a block within the configuration that can be repeated to specify
+the port mappings of the container. Each `ports` block supports
+the following:
+
+* `server_address` - (Required, string) Docker registry url.
+* `username` - (Required, string) Username used to login.
+* `password` - (Required, string) Password used to login.
+* `email` - (Optional, string) Email used to login.
+
+## Attributes Reference
+
+The following attributes are exported in addition to the above configuration:
+
+* `configurations` (string) - Serialized docker configurations.
--- a/website/source/docs/providers/docker/r/registry.html.mdown
+++ b/website/source/docs/providers/docker/r/registry.html.mdown
@@ -3,7 +3,7 @@ layout: "docker"
 page_title: "Docker: docker_registry"
 sidebar_current: "docs-docker-resource-registry"
 description: |-
-  Logs to a Docker registry, downloads and exports a Docker image.
+  Manages credentials for one or many docker registries.
 ---
 
 # docker\_registry
@@ -34,13 +34,13 @@ resource "docker_image" "ubuntu" {
 The following arguments are supported:
 
 * `settings_file` - (Optional, string) Path to the docker configuration file.
-* `auths` - (Optional, block) See [Auths](#auths) below for details.
+* `auth` - (Optional, block) See [Auth](#auth) below for details.
 
-<a id="auths"></a>
-### Auths
+<a id="auth"></a>
+### Auth
 
-`auths` is a block within the configuration that can be repeated to specify
-the auth to use with an image. Each `auths` block supports
+`auth` is a block within the configuration that can be repeated to specify
+the auth to use with an image. Each `auth` block supports
 the following:
 
 * `server_address` - (Required, string) Docker registry url.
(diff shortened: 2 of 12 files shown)


xsellier:
- Set docker_registry as 'ForceNew' in order to read the file at each deployment (token expired)
- fix docker_registry documentation
- Replace 'auth' by 'auths' in order to be able to manage more than one authentication per docker_registry resource

xsellier:
Tested many times during this week, this Pull Request is finally ready (Sorry, I forget the [WIP] tag).
Everything seems to work fine.

website/source/docs/providers/docker/r/registry.html.mdown
```diff
@@ -0,0 +1,55 @@
+---
+layout: "docker"
+page_title: "Docker: docker_registry"
+sidebar_current: "docs-docker-resource-registry"
+description: |-
+  Logs to a Docker registry, downloads and exports a Docker image.
```
apparentlymart: Is this description accurate? It seems like maybe this is the description from `docker_image`.

website/source/docs/providers/docker/r/registry.html.mdown
```diff
@@ -0,0 +1,55 @@
...
+
+The following arguments are supported:
+
+* `settings_file` - (Optional, string) Path to the docker configuration file.
+* `auths` - (Optional, block) See [Auths](#auths) below for details.
+
+<a id="auths"></a>
+### Auths
+
+`auths` is a block within the configuration that can be repeated to specify
```
apparentlymart: If each of these blocks represents "one auth" then the usual style is to name it in the singular so that each block appears with a singular heading in the config:

```hcl
auth {
    server_address = "..."
    username = "..."
    password = "..."
}
auth {
    server_address = "..."
    username = "..."
    password = "..."
}
```

I'd agree that it feels a bit weird in the implementation to have a list/set field with a singular name, but it makes sense given how this gets written in the configuration.

xsellier:
Any news on this ?

xsellier:
Sorry for the double post, but is there any news on this pull request ?

website/source/docs/providers/docker/r/image.html.markdown
```diff
@@ -16,9 +16,19 @@ them.
 ## Example Usage
 
 ```
+# Log to a custom docker registry
+resource "docker_registry" "ubuntu_private" {
+    auths = {
```
matthewvalimaki: This should be `auth`. Currently it gives `invalid or unknown key: auths`.

xsellier:
Nice catch !
I've amended the commit with this fix.

matthewvalimaki:
I'm getting `Unable to create container: API error (500): No command specified` when I'm trying it with 
```
resource "docker_image" "registry" {
    name = "alpine:3.1"
    registry = "${docker_registry.registry.configurations}"

}
```
Also trying to use my own hub.docker.com private repository leads to `Unable to read Docker image into resource: Unable to pull image matthewvalimaki/repo:latest`. I do not spot any mistakes in my configuration.

Tested with Docker version 1.10.3.

xsellier:
I've tested with docker:
`Docker version 1.11.1, build 5604cbe`

xsellier:
Issue is not from docker itself, because it work with 1.10.3.
So far I've tested it without specifying an image tag, tomorrow I will to fix that issue.

Good catch @matthewvalimaki

xsellier:
Hey @matthewvalimaki ,

I think I've fixed the issue. It was about the image string parsing. I've replaced it with a regular expression, it should work like a charm now.

Hope you can test it by the end of the week.

matthewvalimaki:
@xsellier it does seem to work but not as I expected.
I have to use `server_address = "https://index.docker.io/v1/"` for it to work with Docker Hub. If I try what is in readme `https://hub.docker.com/` it gives `Error pulling image, registry: 'index.docker.io/v1/', repository:`.

xsellier:
@matthewvalimaki Oh sure, I'll fix that also tomorrow :)

xsellier:
@matthewvalimaki now it should work properly as expected.

Since we can read the credentials from a file, we can have more than one credential loaded. That being said I guess you understand we have to match the `docker_registry` with the `docker_image`. So I added an exception, if there is only one credential loaded from the file, we will use it anyway.

xsellier:
any news on this ?

alexraju:
This is a much awaited essential. @xsellier Thank you.

lpmaurice:
+1

xsellier:
Hey folks,

this patch works pretty well, anything new about it ?

radeksimko:
Hi @xsellier 
thank you for taking the time to submit this PR.

Although I understand that the only reason user usually authenticates in Docker is to pull images from private docker registry, authentication & connection configuration is usually handled in `provider` rather than `resource` in Terraform.

It may also be helpful for you if you read my comment about the separation of Docker data sources/resources [here](https://github.com/hashicorp/terraform/pull/7000#issuecomment-231596846).
Admittedly it cannot be _completely_ separated due the fact that to pull image from private registry via Docker Remote API one still has to authenticate against the Docker Registry API - but in reality the Docker Remote API only requires the base64 encoded auth string, so technically we're not calling the registry API directly.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/create-an-image

screen shot 2016-07-10 at 17 31 24


I think that the problem described in https://github.com/hashicorp/terraform/issues/4934 is caused by the fact that we don't make authentication explicit -> that is basically what you're fixing here as part of this PR 😃  - except that the logic is inside a new resource rather than part of existing provider.

What do you think about the following?

```hcl
provider "docker" {
  host = "tcp://127.0.0.1:1234/"
  registry_auth = {
    username = "foo"
    password = "bar"
    server_address = "docker.io"
  }
}
```

xsellier:
Thank you for this answer @radeksimko .

Although this seems to be a good idea, I think we miss a point.
I use a registry independantly of the provider, my provider allows me to connect to a docker instance and to deploy a docker container. And, I use a `docker_registry` to pull an image Because I want to be able to use many `docker_registry` for a unique `docker_provider` I think we need a `docker_registry` tag.

Let me explain, in my specific case, I use a `docker_provider` to deploy containers. But some of the containers are on different private `docker_registry`. Because I have a _common_ registry, for common containers, and _specific_ registries for specific containers, but everything is deployed on the very same machine (the same `docker_provider`)

Although we can set more than one `registry_auth` per provider, but I also want to factorize the code, so I have a unique terraform file with `docker_registry` credentials, and I use it into others terraform files. And because I don't want to duplicate any `docker_repository`, I though it was a smart move to define a `docker_registry`.

radeksimko:
@xsellier I see, thank you for the explanation. Would defining multiple `provider`s help in such cases?

```hcl
variable "docker_host" {
  default = "/my/lovely/docker/unix/socket"
}

provider "docker_registry" {
  alias = "dockerhub"
  host = "https://registry.hub.docker.com"
  username = ""
  password = ""
}

provider "docker_registry" {
  alias = "quay"
  host = "https://quay.io"
  username = ""
  password = ""
}

data "docker_registry_image" "nomad" {
  provider = "docker_registry.dockerhub"
  name = "nomad"
  tag = "latest"
}

data "docker_registry_image" "nomad" {
  provider = "docker_registry.quay"
  name = "coreos"
  tag = "latest"
}

provider "docker" {
  alias = "dockerhub"
  host = "${var.docker_host}"
  auth {
    host = "https://registry.hub.docker.com"
    username = ""
    password = ""
  }
}

provider "docker" {
  alias = "quay"
  host = "${var.docker_host}"
  auth {
    host = "https://quay.io"
    username = ""
    password = ""
  }
}

# First use case
resource "docker_image" "nomad" {
  provider = "docker.dockerhub"
  name = "${data.docker_registry_image.nomad.name}:${data.docker_registry_image.nomad.digest}"
}

# Second use case
resource "docker_image" "nomad" {
  provider = "docker.dockerhub"
  name = "my_special_app:latest"
  trigger_pull = ["${data.docker_registry_image.nomad}"]
}

# Third use case
resource "aws_ecs_task_definition" "nomad" {
  provider = "docker.dockerhub"
  family = "nomad"
  container_definitions = <<DEFINITION
[
  {
    "name": "jenkins",
    "image": "${data.docker_registry_image.nomad.name}:${data.docker_registry_image.nomad.digest}",
    "cpu": 10,
    "memory": 500,
  }
]
DEFINITION
}
```

I wouldn't mind keeping the data source under existing `docker` provider after all, but I still believe that auth should be handled as part of the provider, not as a separate resource. Also if we want the data source to be useful w/out direct docker connection I think we would need to make `host` field in `docker` provider Optional. There are people using container schedulers and these don't talk over Docker Remote API.

Resources have CRUD, hence they Create, Read, Update and Delete things, they typically don't provide authentication - the lifecycle would also prevent us from reusing such _resource_ as authentication source for the data source.

Maybe we should have 1st class support for secrets/credentials, that would be AOK 👌 

```hcl
auth "docker_registry" "hub" {
  host = ""
  user = ""
  pass = ""
}

resource "docker_image" {
  auth = "${auth.docker_registry.hub}"
  name = ""
}
```

Obviously that would require more significant changes in Terraform's core, but it would IMO make sense to support it, especially once encryption of tfstate + Vault integration becomes a reality.

kyhavlov:
Potentially dumb question - is there any use in passing the registry creds through the provider directly, like this? For my own use, I put the registry creds in ~/.docker/config.json and use the patch [here](https://github.com/hashicorp/terraform/pull/6892) to load them. The docker config.json already supports creds for multiple registries, and `NewAuthConfigurationsFromDockerCfg()` is already built into the go-dockerclient. I thought letting secrets end up in the terraform state was an anti-pattern, and I do the same thing for the AWS provider (having it pick up the creds via a file instead of passing it into the provider block).

xsellier:
@radeksimko I like this syntax, I will implement it soon. Need to re-write some part of the code for that.

Thank you for this review.

radeksimko:
> Potentially dumb question - is there any use in passing the registry creds through the provider directly, like this?

I personally wouldn't hard-code username+password, especially because tfstate is currently stored in plaintext, but I can understand why a user would want to specify custom location of `~/.docker/config` and/or custom registry hostname. I'm not saying "let's implement this now" - I'm happy to accept a PR that **doesn't** support `username+password` based auth, but I would not prevent anyone from implementing it & merging it.

Admittedly the registry hostname could be part of `docker_image.name`... which makes me think that `name` may not be the right name for this field... but I'm really nitpicking now... 😄

radeksimko:
@xsellier In case it wasn't clear from my example above - the `provider` alias functionality is part of Terraform's core - i.e. should require no extra provider/resource/data-source logic. Just make sure that connections/settings can be shared correctly between a provider and resource/data-source via `meta`.

xsellier:
@radeksimko I got you, but I have already spent much time on this feature since few month. It has already been reviewed, so I won't spend more time on this right now, and since I don't know when it's gonna be merged after this review. So I would appreciate it very much if you make the change yourself to be compliant with the very new naming-convention.

Issue #6436, opened 2016-05-02
EIP reallocation failure


cstruct:
Hi,

Since the v0.6.15 release terraform cannot reassociate elastic IPs.

### Terraform Version
Fails on `Terraform v0.6.15`
Succeeds on `Terraform v0.6.14`

### Affected Resource(s)
- aws_eip

### Terraform Configuration Files
Changing config from:
```hcl
resource "aws_eip" "default" {
  vpc = true
  instance = "${aws_instance.instance_a.id}"
}
```
to:
```hcl
resource "aws_eip" "default" {
  vpc = true
  instance = "${aws_instance.instance_b.id}"
}
```

### Debug Output
[TRACE log](https://gist.github.com/cstruct/16b25a70711a7554b95fb0b52e23600b)

### Expected Behavior
The EIP should be reassociated with the new instance

### Actual Behavior
Terraform fails with:
```
aws_eip.default: Failure associating EIP: InvalidParameterValue: 10.x.x.x is not mapped to the interface eni-xxxxxxxx
```

### Steps to Reproduce
1. set existing `aws_eip.instance` to a new value
1. `terraform apply`

WishCow:
Also affected by this

dbrandt:
Ran in to this as well.

@cstruct Thanks for taking the time to write up a thorough bug report. I wasn't certain if I was doing something stupid with AWS or if this was an issue...

catsby:
Hey @cstruct thanks for opening this!

#6521 should fix this. Note that if you're using `private_ip` in your EIP resource, you'll need to use the new `associate_with_private_ip` attribute that's added in that pull request

cstruct:
Thank you @catsby

PR #6070, opened 2016-04-07
provider/aws: Allow multiple EIPs to associate to single ENI

phinze:
When calling AssociateAddress, the PrivateIpAddress parameter must be
used to select which private IP the EIP should associate with, otherwise
the EIP always associates with the _first_ private IP.

Without this parameter, multiple EIPs couldn't be assigned to a single
ENI. Includes covering test and docs update.

Fixes #2997

PR #6521, opened 2016-05-06
provider/aws: Fix issue changing EIP Association

catsby:
Fix issue when moving an EIP from one Instance to another. 
Currently, we provide the `PrivateIpAddress` in the `AssociateAddress` call, but we're sending the private IP of the current instance, not the instance we're targeting. Since `private_ip` is computed, on association we get the PrivateIp of the first Instance here, which incorrectly get's thrown in with the request. 

Fixes https://github.com/hashicorp/terraform/issues/6436 and failing test `TestAccAWSEIP_instance`

**Update**

Chatted with @phinze about the possible/likely conflict with https://github.com/hashicorp/terraform/pull/6070 and we've got a plan for going forward. I will re-work this PR and update soon

**More Update:**

I've implemented what @phinze and I discussed:

- remove `private_ip` as a user specified value: this value is once again computed and reflects the current private ip
- added `associate_with_private_ip`: this is a private IP you designate to associate the EIP with. 

Private IP is a computed attribute, meaning it will pick up it's value from the Instance or Network Interface that the EIP is associated with. The `associate_with_private_ip` allows the users to specify what value the EIP should take. Without distinguishing between the two cases, we're unable to move the EIP from Instance to Instance (see https://github.com/hashicorp/terraform/issues/6436 and `TestAccAWSEIP_instance` test)

### DIFF
--- a/builtin/providers/aws/resource_aws_eip.go
+++ b/builtin/providers/aws/resource_aws_eip.go
@@ -61,8 +61,12 @@ func resourceAwsEip() *schema.Resource {
 
 			"private_ip": &schema.Schema{
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
+			},
+
+			"associate_with_private_ip": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


catsby:
Tests are green:

```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEIP_ -timeout 120m
=== RUN   TestAccAWSEIP_basic
--- PASS: TestAccAWSEIP_basic (10.44s)
--- PASS: TestAccAWSEIP_instance (156.02s)
--- PASS: TestAccAWSEIP_network_interface (32.28s)
--- PASS: TestAccAWSEIP_twoEIPsOneNetworkInterface (32.19s)
--- PASS: TestAccAWSEIP_associated_user_private_ip (143.78s)
PASS
ok  	github.com/has
```

phinze:
LGTM!

PR #6522, opened 2016-05-06
provider/vsphere: read gateway information

thetuxkeeper:
The gateway was not read from the VM before and could have forced redeploying the VM when setting a static gateway.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -799,13 +799,21 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 		for _, v := range mvm.Guest.IpStack {
 			if v.IpRouteConfig != nil && v.IpRouteConfig.IpRoute != nil {
 				for _, route := range v.IpRouteConfig.IpRoute {
-					if route.Network == "0.0.0.0" && route.Gateway.Device != "" {
-						deviceID, err := strconv.Atoi(route.Gateway.Device)
-						if err != nil {
-							log.Printf("[WARN] error at processing device id %#v: %#v", route.Gateway.Device, err)
-						} else {
-							log.Printf("[DEBUG] gateway of device id %d: %s", deviceID, route.Gateway.IpAddress)
-							networkInterfaces[deviceID]["ipv4_gateway"] = route.Gateway.IpAddress
+					if route.Gateway.Device != "" {
+						gatewaySetting := ""
+						if route.Network == "::" {
+							gatewaySetting = "ipv6_gateway"
+						} else if route.Network == "0.0.0.0" {
+							gatewaySetting = "ipv4_gateway"
+						}
+						if gatewaySetting != "" {
+							deviceID, err := strconv.Atoi(route.Gateway.Device)
+							if err != nil {
+								log.Printf("[WARN] error at processing %s of device id %#v: %#v", gatewaySetting, route.Gateway.Device, err)
+							} else {
+								log.Printf("[DEBUG] %s of device id %d: %s", gatewaySetting, deviceID, route.Gateway.IpAddress)
+								networkInterfaces[deviceID][gatewaySetting] = route.Gateway.IpAddress
+							}
 						}
 					}
 				}
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -802,7 +802,7 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 					if route.Network == "0.0.0.0" && route.Gateway.Device != nil {
 						deviceID, err := strconv.Atoi(route.Gateway.Device)
 						if err != nil {
-							log.Printf("[DEBUG] error at processing device id %#v: %#v", route.Gateway.Device, err)
+							log.Printf("[WARN] error at processing device id %#v: %#v", route.Gateway.Device, err)
 						} else {
 							log.Printf("[DEBUG] gateway of device id %d: %s", deviceID, route.Gateway.IpAddress)
 							networkInterfaces[deviceID]["ipv4_gateway"] = route.Gateway.IpAddress
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -795,15 +795,19 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 			networkInterfaces = append(networkInterfaces, networkInterface)
 		}
 	}
-	for _, v := range mvm.Guest.IpStack {
-		for _, route := range v.IpRouteConfig.IpRoute {
-			if route.Network == "0.0.0.0" {
-				deviceID, err := strconv.Atoi(route.Gateway.Device)
-				if err != nil {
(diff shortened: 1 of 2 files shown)


chrislovecnm:
Should one of our test caught this? We don't need another test, but I am thinking we should improve the test.

thetuxkeeper:
The `TestAccVSphereVirtualMachine_ipv4Andipv6` test was failing for me
```
=== RUN   TestAccVSphereVirtualMachine_ipv4Andipv6
--- FAIL: TestAccVSphereVirtualMachine_ipv4Andipv6 (96.15s)
        testing.go:172: Step 0 error: Check failed: Check 10/12 error: vsphere_virtual_machine.ipv4ipv6: Attribute 'network_interface.0.ipv4_gateway' expected "10.30.8.1", got ""
FAIL
```

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -795,6 +795,19 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 			networkInterfaces = append(networkInterfaces, networkInterface)
 		}
 	}
+	for _, v := range mvm.Guest.IpStack {
+		for _, route := range v.IpRouteConfig.IpRoute {
```
chrislovecnm: do we need to check for nil on these?? Can `IpRouteConfig == nil`?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -795,6 +795,19 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
+			if route.Network == "0.0.0.0" {
+				deviceID, err := strconv.Atoi(route.Gateway.Device)
```
chrislovecnm: Another one, can `Gateway` be nil, and cause a panic?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -795,6 +795,19 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
+				if err != nil {
+					log.Printf("[DEBUG] error at processing device id %#v: %#v", route.Gateway.Device, err)
+				} else {
+					log.Printf("[DEBUG] gateway of device id %d: %s", deviceID, route.Gateway.IpAddress)
```
chrislovecnm: Remind me should these have a Debug log level?  Dumb question, but don't have time to dig :)

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
thetuxkeeper: Yes. I think we need some more checks here. `Guest`, `IpStack`, `IpRouteConfig` and `IpRoute` could be `nil`

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
thetuxkeeper: The `IpRoute` properties are always set, but not the properties of `Gateway` => yes, more `nil` checks

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
thetuxkeeper: I think you're right about the first one. This should be a Info or something like that. Have to check how it's done at other places in TF.

thetuxkeeper:
@chrislovecnm Fixed. Please review again :)

chrislovecnm:
@thetuxkeeper travis is not happy:

builtin/providers/vsphere/resource_vsphere_virtual_machine.go:802: invalid operation: route.Gateway.Device != nil (mismatched types string and nil)

thetuxkeeper:
oh, I missed that. I thought I checked it. go is not nice here ;)
Will check it later.

thetuxkeeper:
@chrislovecnm Fixed the nil/empty check
I pushed the fixes for the tests, though you said to put it in an additional/WIP PR. But there are not much changes. I can remove it again if you want (and put it as an additional PR).
before 0fac39c
```
=== RUN   TestAccVSphereVirtualMachine_basic
--- FAIL: TestAccVSphereVirtualMachine_basic (112.18s)
        testing.go:172: Step 0 error: After applying this step, the plan was not empty:

                DIFF:

                UPDATE: vsphere_virtual_machine.foo
                  network_interface.0.ipv4_gateway: "10.30.8.1" => ""

                STATE:

                vsphere_virtual_machine.foo:
                  ID = terraform-test
                  cluster = dc1.boerse-go.de
                  datacenter = dc1
                  disk.# = 2
                  disk.0.bootable = false
                  disk.0.datastore = DS1
                  disk.0.iops = 500
                  disk.0.size = 0
                  disk.0.template = terraform-test-folder/centos-terraform-template
                  disk.0.type = eager_zeroed
                  disk.0.vmdk =
                  disk.1.bootable = false
                  disk.1.datastore =
                  disk.1.iops = 500
                  disk.1.size = 1
                  disk.1.template =
                  disk.1.type = eager_zeroed
                  disk.1.vmdk =
                  domain = vsphere.local
                  gateway = 10.30.8.1
                  linked_clone = false
                  memory = 4096
                  memory_reservation = 4096
                  name = terraform-test
                  network_interface.# = 1
                  network_interface.0.adapter_type =
                  network_interface.0.ip_address =
                  network_interface.0.ipv4_address = 10.30.8.250
                  network_interface.0.ipv4_gateway = 10.30.8.1
                  network_interface.0.ipv4_prefix_length = 24
                  network_interface.0.ipv6_address = fe80::250:56ff:fe8e:3807
                  network_interface.0.ipv6_gateway =
                  network_interface.0.ipv6_prefix_length = 64
                  network_interface.0.label = cld_tst1_access
                  network_interface.0.subnet_mask =
                  resource_pool = dc1.boerse-go.de/Resources/test_server
                  skip_customization = false
                  time_zone = Etc/UTC
                  vcpu = 2
FAIL
exit status 1
FAIL    github.com/hashicorp/terraform/builtin/providers/vsphere        112.191s
```
after 0fac39c:
```
TF_ACC=1 go test ./builtin/providers/vsphere -v -run=TestAccVSphereVirtualMachine_basic -timeout 120m
=== RUN   TestAccVSphereVirtualMachine_basic
--- PASS: TestAccVSphereVirtualMachine_basic (105.38s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/vsphere        105.394s
```

chrislovecnm:
@stack72 @jen20 @phinze ~ can someone take a find review and get the code in??

Thanks!!!!

thetuxkeeper:
Like I discovered in #6590 the `ipv6_gateway` is not covered by this PR yet. I'll try to push an additional patch. It should be quite the same as with ipv4, so it shouldn't add too much more to review.

chrislovecnm:
@thetuxkeeper you want to fix ipv6?

thetuxkeeper:
@chrislovecnm : yes. just pushed it.

jen20:
This seems good to me - I have no way of verifying whether the acceptance test is passing but I am happy to take @thetuxkeeper's word for it! Thanks all!

chrislovecnm:
@jen20 much thanks!!

thetuxkeeper:
@jen20 : Thanks! I hope I can prove that I'm worth your trust. :)
But I think I missed something. I broke the DHCP tests. Since the gateway is read from vSphere, but not configured.
I think we have to change the `ipv*_gateway` settings to computed?

PR #6529, opened 2016-05-06
provider/triton: Recognize 404 for Firewall Rules

jen20:
This prevents a firewall rule deleted from underneath Terraform from causing a refresh to stall.


stack72:
Now that I got schooled on the usage of Exists, this LGTM! :)

Issue #6489, opened 2016-05-04
cloudfront distribution should export route53 zone_id


ryanking:
### Terraform Version

```
$ tf version
Terraform v0.6.15
```

### Affected Resource(s)
Please list the resources as a list, for example:
- cloudfront_distribution

### Feature request

In order to create a route53 alias you need the zone id. So to create an alias to a cloudfront distribution, we need to have the cloudfront_distribution return its zone id.

vancluever:
Hey @ryanking, there was a conscious decision to leave this out of the resource, as it is a hardcoded value that is not exposed through the CloudFront API. See https://github.com/hashicorp/terraform/pull/5221#issuecomment-192049919 for the original discussion.

The zone ID is actually global for all CloudFront distributions - `Z2FDTNDATAQYW2` - which is documented [here](http://docs.aws.amazon.com/Route53/latest/APIReference/CreateAliasRRSAPI.html).

I'm torn on this - I'm kind of weary of exposing arbitrary constants that are not exposed through the API, but also, at the same time, there seems to be a bit of an expectation that it will be there, from what I've seen from a couple of people here and on IRC, so maybe it's a good idea to have it in as a quality of life feature. I did something similar in `cloudfront_origin_access_identity` as well.

radeksimko:
To be fair we do this already for S3 buckets: https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_s3_bucket.go#L674

I think it's still better to have this hardcoded once in Terraform rather than sprinkled all over user's Terraform configs.

Also we should allow users to build the dependency chain where DNS records will only be created once the CF distribution is created - not simultaneously (which is the case when user hard-codes this IMO).

apparentlymart:
I would second @radeksimko's opinion here. Having this exposed is expected and intuitive in spite of the fact that we'd be working around a limitation in the native AWS API.

One of the big benefits Terraform provides, in my opinion, is making explicit what would normally be implicit: the relationships between different things are evident to those reading the Terraform configuration, even if they are not 100% familiar with the resource types being described... the reader doesn't have to chase down implicit relationships by noticing that an IP address over here matches with one over there, or whatever.

[...]

PR #6530, opened 2016-05-07
provider/aws: Add CloudFront hosted_zone_id attribute

vancluever:
Reference: #6489 

Back by popular demand ;)

This adds the `hosted_zone_id` attribute to the `aws_cloudfront_distribution` resource, which, at the time, is currently just an alias for `Z2FDTNDATAQYW2`. This ultimately makes using the resource more semantic, allowing an easy place to reference the zone ID from and also a natural dependency path.

Docs included, and test hooked into the first acceptance test.


stack72:
Hi @vancluever 

Thanks so much - this looks like a simple addition :)

P.

Issue #6523, opened 2016-05-06
Terraform fails to authenticate using STS temporary credentials - InvalidClientTokenId


cabarria:
I'm running Terraform v0.6.15. I have exported the following keys following an STS call for credentials:
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_SESSION_TOKEN (Can use AWS_SECURITY_TOKEN as well)

the aws-cli works, but Terraform complains with the following:

Refreshing Terraform state prior to plan...

Error refreshing state: 1 error(s) occurred:

1 error(s) occurred:

InvalidClientTokenId: The security token included in the request is invalid
status code: 403, request id:

No matter what I do... I've tried AWS_TOKEN as the key as well, Terraform fails to use temporary credentials. If I use a Permanent Access Key and Secret Access Key it works. I thought this issue had been solved.

cabarria:
Looks like this is failing because the temporary credentials are being requested using the GetFederationToken API. http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html

Those temporary credentials are not allowed to make IAM API or STS calls. There must be some code in Terraform checking for that regardless of whether the user intends to work with IAM or not.

I think this should be fixed... Terraform shouldn't presume to know that the credentials provided are not adequate to perform whatever actions the user needs.

radeksimko:
Hi @cabarria 
I was able to reproduce this. You're right in a way that Terraform currently uses `iam:GetUser` to validate credentials.

We may be able to swap this with a fairly new API endpoint `sts:GetCallerIdentity` which wasn't available before. It seems that this API method is always available, even in regions that have explicitly disabled STS support.

I need to confirm this is expected and correct behaviour w/ AWS support and then we should be good to make that change.

This also makes me think that our recent changes for `allowed_account_ids` & `forbidden_account_ids` won't work yet for federated accounts due to this validation we have in place.

radeksimko:
I just confirmed that `sts:GetCallerIdentity` works on EC2 instances.

I also approached the AWS support in regards to disabled STS regions and the representative told me to wait until they hear confirmation back from the service team. I think I will prepare a PR in the meantime though.

radeksimko:
See PR that is fixing this at #6536

cabarria:
Thanks for the quick response!!!

riponbanik:
is it being fixed on terraform v0.7.13. Stil having issue 

[...]

PR #6533, opened 2016-05-08
add `validate_credentials` option to provider-aws

iamatypeofwalrus:
My dev team is planning to use Terraform to manage our AWS resources in all of our environments (Dev, Test, and Prod).

We develop 100% locally (fake_sqs and DynamoDB Local) and do not have access to AWS credentials. We'd like to use Terraform to stand up tables against DynamoDB Local (or any other local AWS service).

I added a `validate_credentials` option to the `provider-aws` Schema. It defaults to `true`. When set to `false` the credentials are not validated at all against IAM or the white / black list of account ids.

Since DynamoDB Local does not check the validity of the credentials we can create tables against it.

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -71,7 +71,9 @@ type Config struct {
 	Ec2Endpoint      string
 	IamEndpoint      string
 	ElbEndpoint      string
-	Insecure         bool
+
+	Insecure                        bool
+	ValidateCredentialsAndAccountID bool
 }
 
 type AWSClient struct {
@@ -180,10 +182,26 @@ func (c *Config) Client() (interface{}, error) {
 		log.Println("[INFO] Initializing STS connection")
 		client.stsconn = sts.New(sess)
 
-		err = c.ValidateCredentials(client.iamconn)
-		if err != nil {
-			errs = append(errs, err)
-			return nil, &multierror.Error{Errors: errs}
+		if c.ValidateCredentialsAndAccountID {
+			log.Println("[INFO] Validating Credentials")
+			err = c.ValidateCredentials(client.iamconn)
+			if err != nil {
+				errs = append(errs, err)
+				return nil, &multierror.Error{Errors: errs}
+			}
+
+			accountId, err := GetAccountId(client.iamconn, client.stsconn, cp.ProviderName)
+			if err == nil {
+				client.accountid = accountId
+			}
+
+			log.Println("[INFO] Validating Account ID")
+			authErr := c.ValidateAccountId(client.accountid)
+			if authErr != nil {
+				errs = append(errs, authErr)
+			}
+		} else {
+			log.Println("[INFO] skipping IAM credential and Account ID validation")
 		}
 
 		// Some services exist only in us-east-1, e.g. because they manage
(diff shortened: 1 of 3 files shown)


radeksimko:
Hi @iamatypeofwalrus 
What you want to achieve most likely makes sense. You made me think about this problem in wider context, see https://github.com/hashicorp/terraform/issues/6535

I'm wondering what's going to be the user experience for other **non-DynamoDB resources** which don't have custom endpoints defined and Terraform won't be able to create these. Will user see any meaningful error message in such cases?

Does your infrastructure only consist of DynamoDB + SQS + compute? or how do you deal with other services which are not fake-able? Are you somehow communicating to developers which AWS services and resources are "supported" in local environments?

------

I would personally hold off on reviewing & merging this until we actually make the credentials validation work for STS credentials - see https://github.com/hashicorp/terraform/issues/6523#issuecomment-217717241

I'd like to get that ^ done as soon as possible as it seems to be the only blocker for STS users, then I'm happy to refocus on the use case with alternative AWS service providers and revisit this PR more thoroughly.

iamatypeofwalrus:
Thanks for the reply!

> Will user see any meaningful error message in such cases?

If users disable `validate_credentials` and they attempt to CRUD resources with fake credentials against real AWS they will see authentication or authorization errors. The errors that I receive when I attempt to create a DynamDB table with fake credentials is:

```
* InvalidClientTokenId: The security token included in the request is invalid.
	status code: 403, request id: ...
```

I think this error is pretty clear in conjunction with a user having explicitly set `validate_credentials` to false. That said the provider could always be more helpful in the case that a CRUD operation fails by providing a hint like: 

```
Is validate_credentials set to false?
```

> Does your infrastructure only consist of DynamoDB + SQS + compute?

Yes, the application itself is that simple. However, in order to be considered production ready we need to manage metrics, alarms, and paging. I created an internal terraform provider to CRUD those. We will need real credentials for request signing in order to manage those resources.

> Are you somehow communicating to developers which AWS services and resources are "supported" in local environments?

My thoughts for managing those resources is to structure an `infrastructure` directory like so:
```
infrastructure/
  README.md
  development.tfvars
  dynamodb/
    dynamodb.tf
  sqs/
    sqs.tf
  monitors/
    monitors.tf
```
The `README` would document which resources are needed in which environment. `monitors/` are Prod only.

A developer wishing to create or update dynamo tables in their sandbox would do:
```
terraform plan -var-file=development.tfvars dynamodb/
```
For managing production resources we would need to grab credentials from our credential management system.

Ultimately, there may be a set of simple shell scripts for grabbing the correct credentials, setting the Terraform vars, and calling terraform.

iamatypeofwalrus:
@radeksimko, has there been any movement on this PR? Happy to push this design forward and search for an alternative with some more feedback.

radeksimko:
@iamatypeofwalrus Thanks again for bringing this use case to my attention.

I think I'm most likely going to merge https://github.com/hashicorp/terraform/pull/7874 in favour of this PR as it has more granular options (separate option for metadata API) which in fact may/should be [even more granular](https://github.com/hashicorp/terraform/pull/7874#discussion_r74049084). I'm keeping it open just until the mentioned PR is _actually_ merged.

I'd be happy to merge a separate PR for the typo in docs (`(Optional) Optional)`) straight away. 😉

iamatypeofwalrus:
@radeksimko just happy to see progress on this :-)

can't wait for #7874 to get merged! it solve a real pain point for me.

radeksimko:
Closed in favour of #7874

PR #6534, opened 2016-05-08
Change in CLI behavior

arsdehnel:
Just wanted to call out that the CLI prompts for values for unset variables instead of an error.  Guessing that was an enhancement somewhere along the line and just didn't get updated in the docs.


jen20:
Hi @arsdehnel! I'm going to modify the wording very slightly to document that it will only prompt for string variables - maps are not prompted in this way. Thanks for opening a PR!

PR #6536, opened 2016-05-08
[WIP] aws: Use new STS endpoint to validate creds

### LINKED ISSUE #6523: Terraform fails to authenticate using STS temporary credentials - InvalidClientTokenId

radeksimko:
We were historically using `iam:GetUser` API for validating credentials, which however doesn't work on EC2 instances with IAM instance profiles (assumed roles) nor with other assumed roles for humans (via  SAML/OpenID). In addition to that any credentials generated via `sts:GetFederationToken` cannot call any STS or IAM endpoints.

We originally avoided the mentioned limitations just by silently ignoring certain error codes coming from `iam:GetUser`. We don't have to do this anymore since the new STS endpoint was introduced (`sts:GetCallerIdentity`).

I was originally afraid that it will be yet another half-working method for validation with dozens of exceptions across different environments - e.g. STS can be disabled per regions. It however looks like the `sts:GetCallerIdentity` can be used even when explicitly denied via IAM policy (`Deny` for `sts:*`) and in region which has STS disabled. It can also be used (surprisingly) on EC2 instances. 🎉 

**WIP:** I approached AWS support as I want them to confirm that the inability to "disable" this API endpoint is intention and not a bug which would be fixed later on. I'm currently awaiting response from them.

Fixes #6523

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -180,7 +180,7 @@ func (c *Config) Client() (interface{}, error) {
 		log.Println("[INFO] Initializing STS connection")
 		client.stsconn = sts.New(sess)
 
-		err = c.ValidateCredentials(client.iamconn)
+		err = c.ValidateCredentials(client.stsconn)
 		if err != nil {
 			errs = append(errs, err)
 			return nil, &multierror.Error{Errors: errs}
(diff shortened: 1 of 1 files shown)


radeksimko:
I just received confirmation from AWS that `sts:GetCallerIdentity` will work no matter if STS is disabled in that region or IAM policy explicitly denies that action. This PR is therefore ready for review.

brianantonelli:
Has this fix been verified? I just built TF from master and it still errors out:

Get a token and then set env vars and call `terraform plan`
```
aws sts get-session-token > ./foo

export AWS_ACCESS_KEY_ID=$(cat foo | jq -r ".Credentials.AccessKeyId")
export AWS_SECRET_ACCESS_KEY=$(cat foo | jq -r ".Credentials.SecretAccessKey")
export AWS_SESSION_TOKEN=$(cat foo | jq -r ".Credentials.SessionToken")

terraform plan

```

Errors:
```

* 1 error(s) occurred:

* InvalidClientTokenId: The security token included in the request is invalid
    status code: 403, request id: a45de91c-3188-11e6-8867-21f1b377912d
```

Here's my aws provider: 
```
provider "aws" {
    region     = "${var.region}"
}
```

brianantonelli:
Ahh, looks like you have to pass the session token as `AWS_SECURITY_TOKEN `

catsby:
LGTM! Thanks @radeksimko

waltervargas:
@radeksimko Thank you for this PR; it enables Okta credentials and remote state in S3.

PR #6537, opened 2016-05-08
provider/aws: Redshift Cluster not allowing 0 for automated_snapshot_retention_period

stack72:
The default value for `automated_snapshot_retention_period` is 1.
Therefore, it can be included in the `CreateClusterInput` without
needing to check that it is set.

This was actually stopping people from setting the value to 0 (disabling
the snapshots) as there is an issue in `d.GetOk()` evaluating 0 for int

Also, documented that  the Default is actually 1 here


stack72:
FYI, this wasn't noticed before as the tests for Redshift are not currently running nightly in Travis (there is a PR to fix that in place too)

jen20:
LGTM.

PR #6538, opened 2016-05-08
Adding private ip address reference to azurerm network interface

Felivel:
* Adding private ip address reference
* Updating the docs.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_network_interface_card.go
+++ b/builtin/providers/azurerm/resource_arm_network_interface_card.go
@@ -55,7 +55,6 @@ func resourceArmNetworkInterface() *schema.Resource {
 
 			"private_ip_address": &schema.Schema{
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 
--- a/builtin/providers/azurerm/resource_arm_network_interface_card.go
+++ b/builtin/providers/azurerm/resource_arm_network_interface_card.go
@@ -259,8 +259,9 @@ func resourceArmNetworkInterfaceRead(d *schema.ResourceData, meta interface{}) e
 
 	if iface.IPConfigurations != nil && len(*iface.IPConfigurations) > 0 {
 		var privateIPAddress *string
-		for _, ipConfig := range *iface.IPConfigurations {
-			privateIPAddress = ipConfig.Properties.PrivateIPAddress
+		///TODO: Change this to a loop when https://github.com/Azure/azure-sdk-for-go/issues/259 is fixed
+		if (*iface.IPConfigurations)[0].Properties != nil {
+			privateIPAddress = (*iface.IPConfigurations)[0].Properties.PrivateIPAddress
 		}
 
 		if *privateIPAddress != "" {
--- a/builtin/providers/azurerm/resource_arm_network_interface_card.go
+++ b/builtin/providers/azurerm/resource_arm_network_interface_card.go
@@ -53,6 +53,12 @@ func resourceArmNetworkInterface() *schema.Resource {
 				Computed: true,
 			},
 
+			"private_ip_address": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
 			"virtual_machine_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


Felivel:
I am working on creating the test for PR.

stack72:
Hi @Felivel 

Please let me know when you are happy for this to go on review. We hope to release 0.6.16 tomorrow (Monday) and this would be a nice addition to it :)

Paul

Felivel:
Hello @stack72, Please review my change.

I have added an extra attribute called `private_ip_address` to the root of the `azurerm_network_interface` schema, in order to be able to access the ip directly from the terraform configuration. 
E.g.: `${azurerm_network_interface.domain_controller_network_interface.private_ip_address}"`

Is there way to access an attribute of an attribute? or are we limited to `TYPE.NAME.ATTRIBUTE`?

Thanks.

Felivel:
I have done testing with a private build. I have already integrated the changes to my workflows. Everything seems to be working fine.

builtin/providers/azurerm/resource_arm_network_interface_card.go
```diff
@@ -252,6 +258,17 @@ func resourceArmNetworkInterfaceRead(d *schema.ResourceData, meta interface{}) e
 		}
 	}
 
+	if iface.IPConfigurations != nil && len(*iface.IPConfigurations) > 0 {
+		var privateIPAddress *string
+		for _, ipConfig := range *iface.IPConfigurations {
+			privateIPAddress = ipConfig.Properties.PrivateIPAddress
```
stack72: we are ranging over a collection here and adding each private ip address in an IPConfig to the same string variable

Should the schema have `private_ip_address` as a List? Or can we guarantee that only 1 private IP will be in place?

stack72:
Hi @Felivel 

I have a few questions inline. Once  answered, we should be able to merge this.

P.

builtin/providers/azurerm/resource_arm_network_interface_card.go
```diff
@@ -53,6 +53,12 @@ func resourceArmNetworkInterface() *schema.Resource {
 				Computed: true,
 			},
 
+			"private_ip_address": &schema.Schema{
```
stack72: We are only using this part of the schema in the Read func. Is this not allowed to be set specifically? If not, then we can just make it Computed, and remove optional

builtin/providers/azurerm/resource_arm_network_interface_card.go
Felivel: At the moment we can only specify one ip configuration.
https://github.com/Azure/azure-sdk-for-go/issues/259

Maybe we can pick the primary ip configuration once #259 is done.

builtin/providers/azurerm/resource_arm_network_interface_card.go
Felivel: Done.

builtin/providers/azurerm/resource_arm_network_interface_card.go
stack72: Ok, if there is only 1, then we shouldn't range over right now

We should do something like this:

```
if iface.IPConfigurations != nil && len(*iface.IPConfigurations) > 0 {
 		var privateIPAddress *string
                 if iface.IPConfigurations[0] != nil {
                     privateIPAddress = iface.IPConfigurations[0].Properties.PrivateIPAddress
                 }
}
```

Then leave a note in place referencing that Azure SDK issue where only 1 NIC is allowed - thoughts?

Felivel: Looks good. I tried doing that but I was getting some errors. I will try it again.

Felivel: I hit the same problem as last time.

```
builtin/providers/azurerm/resource_arm_network_interface_card.go:264: 
invalid operation: iface.IPConfigurations[0] (type *[]network.InterfaceIPConfiguration does not support indexing)
```

Should I assign the pointer value to a variable and try like that?

stack72: Hi @Felivel 

The issue here is that you cannot index when a pointer is in play. So it has to be dereferrenced like this:

```
(*iface.IPConfigurations)[0].Properties.PrivateIPAddress
```

Felivel: Ok, I will push the change in a moment.

stack72:
Hi @Felivel 

this looks great :) 

At the end of a Terraform run, I can now get the following:

```
output "private_ip" {
    value = "${azurerm_network_interface.test.private_ip_address}"
}

Outputs:

  private_ip = 10.0.2.4
```

Issue #5036, opened 2016-02-07
remote: Introduce locking mechanism into remote backend interface


radeksimko:
The current remote backend is expected to have the following methods:
- `Get() (*Payload, error)`
- `Put(data []byte) error`
- `Delete() error`

I think the following methods can be introduced to allow implementation of locking mechanism per backend:
- `Lock() error`
- `Unlock() error`
- `IsLocked() (bool, error)`
- _`GetLockMetadata() (*LockMetadata, error)`_ - This might return some metadata, like **who** and **when** locked the state.

After having these methods in place, we could also introduce new commands:
- `terraform remote lock`
- `terraform remote unlock [-force]` - this could remove lock, the `-force` flag may be required if user is trying to unlock state which has been locked by someone else
- `terraform plan -lock` - Lock before running plan
- `terraform apply/destroy [-no-lock]` - By default I'd assume that this command would check the lock and then lock the state
- `terraform remote get-lock-metadata` - See who and when locked the state

Obviously this won't prevent issues when someone outside of Terraform changes the resources, but at least it could prevent people in teams working on the same codebase stamping on each others feet.

Also I'm fully aware that many of the current backends won't guarantee absolutely atomic lock as locking often requires two or three requests (1st check lock existence, 2nd write lockfile, 3rd check lock existence). I'm not sure how to best address this if the backend just doesn't support this - whether to not implement locking for that backend at all or whether to accept the downside of such locking.

FYI @bobtfish this is what we discussed at CfgMgmt Camp, thanks for some tips about implementation.

Shruti29:
+1 for this feature.

As a temporary workaround, to address the problem mentioned above I have implemented the lock/unlock of tfstate file on S3 with simpleDB based on this answer from stack overflow: http://stackoverflow.com/questions/3431418/locking-with-s3

I have added the working code here. 
https://github.com/Shruti29/terraform-remote-state

apparentlymart:
@radeksimko I was thinking about this the other day while looking at some code in this area. (for #6540)

Specifically, I was thinking that with my changes in that PR it could be safe to eliminate the manual locking commands from your proposal and instead just make Terraform automatically acquire the lock for the duration of each individual operation that affects state (refresh, apply).

[...]

PR #4618, opened 2016-01-11
[WIP] Tracking Remote State Lineage to Reduce Mistakes

apparentlymart:
Each new state will have a uuid as its "lineage" value, and that value will never be changed when updating the state in future. This allows Terraform to detect whether two different states are different points in time along a particular line of states, or if they are unrelated.

In the situation where a state is already present locally (either in `terraform.tfstate` or `.terraform/terraform.tfstate`, depending on whether remote state is enabled), one can run `terraform remote config` to change the remote state configuration.

When this is run, Terraform writes the current local state to the newly-selected remote state storage backend, unconditionally overwriting whatever was there before. This can prove troublesome if, due to user error, the given new configuration points at a location where an unrelated state is already stored: the state is then overwritten, and if that backend doesn't provide any sort of versioning capability (e.g. for Consul) then the state is completely lost.

This implements the feature request in #4389

* [x] Add "lineage" to the state file
* [ ] Check lineage during `terraform remote config`

PR #4616, opened 2016-01-11
[WIP] Refuse to apply a stale plan

apparentlymart:
A "stale" plan, for the sake of this change, is one that was made against a different state than the current one. This occurs if someone tries to apply the same plan twice, for example.

There is no reason to re-apply the same plan, or to otherwise apply a plan against an earlier state, so now we'll detect this scenario and fail early, before making any changes to infrastructure.

This is part of a series of contributions I'm planning for [real-world issues I've encountered with Terraform state](https://gist.github.com/apparentlymart/657885e730d1e5abc6ea). In this case, accidentally re-applying a creation diff, causing duplicate resources to be created.

This appears to work based on my manual testing, but I've not yet figured out how the command tests work well enough to be able to write a unit test for this. I plan to circle back and do that when I have some time to study that test setup a bit more closely.

This fixes #4055.

* [x] Implementation
* [ ] Unit Tests

PR #6540, opened 2016-05-08
[WIP] Safety features for Terraform State

apparentlymart:
Through real-world use of Terraform in production I have encountered [a variety of "gotchas" relating to state management](https://gist.github.com/apparentlymart/657885e730d1e5abc6ea).

This PR is an attempt to tackle a couple of these where I think I have a reasonable path forward. These changes are grouped into a single PR because they share a common building-block:

* [x] "State lineage" concept, which allows us to determine when one state is an earlier or later version of another state vs. when the two states are entirely unrelated. Comparing `Serial` is only expected to produce meaningful results when `Lineage` matches.
  * [x] Some existing tests need work here because they need to meet the lineage-matching requirements that didn't exist before.

The "features" themselves are:

* [x] Detect when a plan applies to a different state than the current state and refuse to apply that plan. (formerly the separate PR #4616) Without locking this is not a 100% guarantee, but it makes things safer than they were before and I expect it to be pretty effective at human timescales. (Something like #5036 in addition could make this even safer; see some discussion over there for more details.)
* [x] When syncing local and remote state, fail with an error if the two states are of different lineage. This includes running `terraform remote config` to enable or re-configure remote state when a local state is already present. (formerly the separate PR #4618)
  * [x] As a special case, if local cache has different lineage but it is entirely empty of resources then just silently drop it and replace it with the remote state, since the loss of an *empty* state is trivial to recover from and this will reduce friction when users are bootstrapping a new config.

The primary goal here is to return an error when the user seems to be accidentally doing something dangerous, with as little impact as possible to "legitimate" workflows. In future version of Terraform we may make more fundamental changes to these features to help the user not make these mistakes in the first place, but this is intended as a short-term fix to reduce the risk of a state-related catastrophe.

* [ ] Rethink how the "no stale plans" change can work in light of #6811

### DIFF
--- a/command/apply_test.go
+++ b/command/apply_test.go
@@ -417,8 +417,35 @@ func TestApply_plan(t *testing.T) {
 
 	planPath := testPlanFile(t, &terraform.Plan{
 		Module: testModule(t, "apply"),
+		State: &terraform.State{
+			Version: 2,
+			Lineage: "TestApply_plan",
+			Serial:  1,
+			Modules: []*terraform.ModuleState{
+				&terraform.ModuleState{
+					Path:      []string{"root"},
+					Resources: map[string]*terraform.ResourceState{},
+				},
+			},
+		},
 	})
-	statePath := testTempFile(t)
+
+	// Current state has the same lineage and serial as the plan state,
+	// so the plan is current and should apply successfully.
+	statePath := testStateFile(t, &terraform.State{
+		Version: 2,
+		Lineage: "TestApply_plan",
+		Serial:  1,
+		Modules: []*terraform.ModuleState{
+			&terraform.ModuleState{
+				Path:      []string{"root"},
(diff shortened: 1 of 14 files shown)


apparentlymart:
@phinze I think this is ready for review now. This is the initial set of "state safety" changes we discussed a couple weeks ago for possible inclusion in 0.7.

gkze:
This is amazing! Would love to see this merged soon 🎉

apparentlymart:
This is now broken by the change from #6811. An entirely different approach will be needed for preventing the application of stale plans now that planning doesn't actually update the persistent state.

apparentlymart:
In light of #6811 I'm closing this in favor of #7026, which has the same introduction of lineage and the second feature from this PR, but skips the "stale plans" feature that now no longer makes sense.

PR #6543, opened 2016-05-08
provider/azurerm: Error creating `azurerm_virtual_machine` data_disk

stack72:
The `storage_data_disk` was trying to use vhd_url rather than vhd_uri. This was causing an error on creating a new data_disk as part of a VM

Also added validation as data_disks can only be 1 - 1023 GB in size


jen20:
LGTM!

PR #6545, opened 2016-05-09
Digital Ocean Example.

roachmd:
I did not see an example for Digital Ocean. So I worked on one today for myself today and thought I would share.


stack72:
Thanks for the work here :) this looks like a good addition to the examples

Issue #6174, opened 2016-04-14
vSphere VM is always partially tainted (missing network interface in local state)


benlangfeld:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version

Terraform v0.6.14

### Affected Resource(s)

- vsphere_virtual_machine

### Terraform Configuration Files

```hcl
provider "vsphere" {
  vsphere_server = "172.18.0.137"
  allow_unverified_ssl = true
  user = "blangfeld"
  password = "*****"
}

# Create a folder
resource "vsphere_folder" "NitroDemoPath" {
  datacenter = "PACV-DC"
  path       = "Nitro/Nitro - Demo10"
}

resource "vsphere_virtual_machine" "nitro-webserver" {
  name         = "demo10-nitro-webserver"
  folder       = "${vsphere_folder.NitroDemoPath.path}"
  datacenter   = "PACV-DC"
  vcpu         = 1
  memory       = 4096
  domain       = "mydatainmotion.com"
  time_zone    = "US/Eastern"
  gateway      = "10.0.5.254"
  dns_servers  = ["10.1.1.36", "10.1.1.37", "172.18.0.40"]
  dns_suffixes = ["powerhrg.com", "mydatainmotion.com"]
  boot_delay   = 300

  network_interface {
    label              = "Demo-VLAN4"
    ipv4_address       = "${cidrhost("10.0.5.32/27", 1)}"
    ipv4_prefix_length = "23"
  }

  disk {
    template  = "Templates/Ubuntu-14.04(10GB)"
    datastore = "gotham_vm"
    type      = "thin"
  }
}
```

### Debug Output

https://gist.github.com/benlangfeld/ee8af09e52820d8b8aa4580102c4b229

### Panic Output

N/A

### Expected Behavior

Initial `apply` run should save complete resource state.

### Actual Behavior

Initial `apply` run saved state with zero network interfaces. Later `plan` run refreshed state to include interfaces. A delay between the two is required; even after `apply` terminates, an immediate `plan` run will not correctly refresh state, will see a diff in network interfaces, and will attempt to rebuild the nodes.

### Steps to Reproduce

1. `terraform apply`
2. `sleep 120`
3. `terraform plan`

### Important Factoids

Nothing I'm currently aware of.

### References

None

chrislovecnm:
[...]

Issue #6520, opened 2016-05-06
provider/vsphere: network interface order changes


thetuxkeeper:
Config:

```
  network_interface {
    label = "cld_tst1_access"
    ipv4_address = "10.30.8.240"
    ipv4_prefix_length = 23
    ipv4_gateway = "10.30.8.1"
  }

  network_interface {
    label = "cld_tst1_storage"
  }

```

Results in (removed ipv6 parts):

```
    network_interface.0.ip_address:         "" => "<computed>"
    network_interface.0.ipv4_address:       "10.30.0.36" => "10.30.8.240"
    network_interface.0.ipv4_gateway:       "" => "10.30.8.1" (forces new resource)
    network_interface.0.ipv4_prefix_length: "23" => "23"
    network_interface.0.label:              "cld_tst1_storage" => "cld_tst1_access" (forces new resource)
    network_interface.0.subnet_mask:        "" => "<computed>"
    network_interface.1.ip_address:         "" => "<computed>"
    network_interface.1.ipv4_address:       "10.30.8.240" => "<computed>"
    network_interface.1.ipv4_gateway:       "10.30.8.1" => ""
    network_interface.1.ipv4_prefix_length: "23" => "<computed>"
    network_interface.1.label:              "cld_tst1_access" => "cld_tst1_storage" (forces new resource)
    network_interface.1.subnet_mask:        "" => "<computed>"
```

As you can see, interface `0` in the config becomes interface `1` in vmware (confirmed over webinterface of vsphere API `<vsphere-server/mob/?moid=vm-<id>&doPath=guest.net)

Perhaps a different approach for the index of the interfaces would be good (perhaps `deviceConfigId`?)?

chrislovecnm:
i understand that this is a design issue, my question is how is it a problem? What is it breaking, how is it impacting us?

thetuxkeeper:
TF would change/redeploy the affected VMs each time you trigger an `apply` (ForceNew for those parameters). Since there is no update implemented. The tests would also fail (non-empty plan) if they used multiple network interfaces.
With an update the issue wouldn't be as big as it is now. But still need a second `apply` to get it right (tests would also fail), if it wouldn't recreate the interfaces resulting in new vmware indices (which are possibly in a different order).

chrislovecnm:
What about ignoring the order and keying on mac addresses?  Does the interface have a vmware guid?  Sounds like we need some api level advice as well.

> Since there is no update implemented.

[...]

PR #6479, opened 2016-05-04
provider/vsphere: govmomi v0.6.1 update and fixes

thetuxkeeper:
I used PR #6472 (thanks @catsby :) ), rebased it to master and started to fix some issues:
- datatypes
- `RemoveDevice`: "keepFiles" boolean was added => false for network devices (should be relevant for disks only)
- `CreateDisk`: not additional parameter for the datastore. `newDiskPath` logic removed ("empty" check is in govmomi now)

cc @chrislovecnm

PR #6377, opened 2016-04-27
provider/vsphere: wait for network enhanced

thetuxkeeper:
I did some changes to the "wait for network" logic to make it more flexible and add the feature to wait for `Guest.Net` (takes ~30s longer than `Summary.Guest.IpAddress` in our environment)
* `boot_delay` with default value
* added `network_wait_timeout` which was previously hard coded to 600s. Changed the check for `boot_delay` to `network_wait_timeout > 0`
* added `wait_for_network_interfaces` to make it possible to wait for `Guest.Net` (the network interfaces)
* added the necessary logic to `waitForNetworkingActive`

I tried not to change the behavior, so `wait_for_network_interfaces` is set to `false` and `boot_delay` to `0` by default.

PR #6547, opened 2016-05-09
[WIP] rough changes that fix vsphere vm's on master

knuckolls:
I'm not sure if this should be merged in it's current form but I wanted to put this out there so people could see what I had to do to get vSphere vms working today.

Currently master pulls the network information too soon thus if anything wants to subsequently use the IP of the instance it fails. I believe this is documented in #6174 and this PR is a really rough fix for the problem. I solved this by adding in a loop that spins until the network information is not null. This probably needs a circuit breaker after some amount of time but it's effective.

I also am not sure if this would work at all with DHCP. It does work when all the addressing information is passed in statically. Furthermore, this fix (#6522) for pulling in the gateway routing information flat out didn't work for me (and I debugged it for hours) so I statically set the gateway based upon what's passed in from the configuration.

Hopefully someone finds this helpful. I'm going to use it as a stopgap for my purposes.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -738,6 +738,13 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 	}
 
 	d.SetId(vm.Path())
+
+	if n, ok := d.GetOk("network_interface"); ok {
+		d.Set("network_interface", n)
+	} else {
+		log.Printf("[DEBUG] Could not set network_interface")
+	}
+
 	log.Printf("[INFO] Created virtual machine: %s", d.Id())
 
 	return resourceVSphereVirtualMachineRead(d, meta)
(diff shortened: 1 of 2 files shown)


knuckolls:
cc: @chrislovecnm. You seem to be commenting on all of the vSphere issues / PR's surrounding getting master to work. Thought you might find this useful as a starting point for a more robust solution.

thetuxkeeper:
Hi @knuckolls,
can you try #6479 + #6377 that should fix the problem, but it is blocked by some other issues. We try to fix the acceptance tests first (that we can be sure it doesn't break other stuff) and ran into further issues (like #6522 and #6520).

chrislovecnm:
@knuckolls thanks for the work. So we believe that there is a way to do this in the API - I am @ the airport so I cannot go into details.

I think we may have a better way to do this ;) 

Again thanks for your work. We appreciate you committing and if you are interested in helping more, we definitely can use you help. Let me know if you want to coordinate more.

chrislovecnm:
@knuckolls does https://github.com/hashicorp/terraform/pull/6377 address these problems?  If so please close out this PR.  BTW thanks for contributing, if you have an opportunity to help out, we have stuff to do :)

stack72:
Hi @knuckolls / @chrislovecnm 

Any thoughts on the status of this PR is this still needed or has #6377 got this covered?

Paul

mitchellh:
Hello! I'm going through `waiting-response` labeled issues and closing issues that haven't been responded to with a meaningful response in at least 2+ weeks. If this is still an issue, please open a new issue so we can start anew. Feel free to reference this existing issue to make a link between the two. Thanks!

PR #6548, opened 2016-05-09
[WIP] Initial Work to create AzureRM Extensions

stack72:
Work to allow AzureRM Extensions (for things like joining a domain)

* [ ] schema
* [ ] CRUD
* [ ] acceptance tests
* [ ] documentation

### DIFF
--- a/builtin/providers/azurerm/provider.go
+++ b/builtin/providers/azurerm/provider.go
@@ -7,6 +7,8 @@ import (
 	"reflect"
 	"strings"
 
+	"sync"
+
 	"github.com/Azure/azure-sdk-for-go/Godeps/_workspace/src/github.com/Azure/go-autorest/autorest"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform/helper/mutexkv"
@@ -14,7 +16,6 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	riviera "github.com/jen20/riviera/azure"
-	"sync"
 )
 
 // Provider returns a terraform.ResourceProvider.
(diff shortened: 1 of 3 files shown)


clintonm9:
@stack72 This feature would be perfect for our use case. If you need anyone to test, just let me know.

clintonm9:
@stack72 I got to spend some time playing with this feature today. Seems very useful to us AzureRM people. 

I ended up using version 1.4 of the customScriptForLinux (type_handler_version = "1.4")

I got the commandToExecute to work well. Did you get to the point where you could set an array of file_uris?

Thanks

stack72:
oh you have tried this branch?

clintonm9:
I did, hope that is ok.

stack72:
that's brilliant that it sort of worked for you :)

I will get the fileUrls working - I have some issues to fix first due to their SDK api for that part :(

stack72:
@clintonm9 

So I am not that familiar with Azure VMs. What sources would you use for a file uri here? Can I just put a simple bash script into a github gist and use that?

P.

clintonm9:
@stack72 

You are correct. That would work.

We currently upload to a storage account in azure then link from there with our powershell scripts.

stack72:
and for that scenario you would need private_settings declared as well right?

clintonm9:
Currently we are not using the SA name and key (if this is what private_setting is).

We just set the files to Blob (which makes them accessible from a public url)

clintonm9:
@stack72 Any updates on when this PR will be completed?

Thanks!

stack72:
ok, after looking into this more, there is no way that this is easily achieved with Terraform. The Extension schema is **very specific** to each extension type. The work here is only for CustomExtensions. We can continue down that path if we want to but this will then only work for that type

There are many different types: https://acom-swtest-2.azurewebsites.net/en-gb/documentation/articles/virtual-machines-extensions-configuration-samples-linux/

@jen20 would love your thoughts here

P

jen20:
I guess we should support some of the ones that seem useful (on a case-by-case basis) on that page and the custom one that is part of this PR. I don't know offhand what is useful though!

gambrose:
@stack72
What is the status of vm extensions? You seemed to imply that it was too hard to do. So I'm thinking this feature my have been put on ice. But I'm hoping you have some master plan as I think this is a really important feature to support.

I have been adding extensions to my vms using azurerm_template_deployment after the vm has been created but this seems pretty hacky.

You linked to VM extensions from Linux. For what is worth these extensions are really useful for windows vms as well. I mostly use the DSC one myself to bootstrap the vm configuration.

Also if there is a plan for vm extensions would this include adding extensions to vm scale sets?

stack72:
Hi all

I am going to close this out. I am not getting time right now to circle back to it but i know that @tombuildsstuff is looking at this :)

P.

PR #5236, opened 2016-02-21
provider/aws: New top level aws_eip_association resource

jkinred:
This is to implement the functionality requested in #2680, allowing pre-existing Elastic IP's to be associated with an instance.

As per the contribution guidelines I'm raising this as WIP.

At the moment it can create and destroy based on allocation id and instance id and I'll continue to work on complete functionality, tests and documentation.

PR #6552, opened 2016-05-09
provider/aws: Change `aws_elastic_ip_association` to have computed parameters

stack72:
This PR takes the awesome work by @jkinred in #5236 and changes the parameters to be computed. This is being done so we can get this work into 0.6.16 :)

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEIPAssociation' 2>~/tf.log
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEIPAssociation -timeout 120m
=== RUN   TestAccAWSEIPAssociation_basic
--- PASS: TestAccAWSEIPAssociation_basic (181.41s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	181.432s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_eip_association.go
+++ b/builtin/providers/aws/resource_aws_eip_association.go
@@ -1,0 +1,175 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsEipAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsEipAssociationCreate,
+		Read:   resourceAwsEipAssociationRead,
+		Delete: resourceAwsEipAssociationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"allocation_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"allow_reassociation": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"instance_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"network_interface_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"private_ip_address": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"public_ip": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceAwsEipAssociationCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	request := &ec2.AssociateAddressInput{}
+
+	if v, ok := d.GetOk("allocation_id"); ok {
+		request.AllocationId = aws.String(v.(string))
+	}
+	if v, ok := d.GetOk("allow_reassociation"); ok {
+		request.AllowReassociation = aws.Bool(v.(bool))
+	}
+	if v, ok := d.GetOk("instance_id"); ok {
+		request.InstanceId = aws.String(v.(string))
+	}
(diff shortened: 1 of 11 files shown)


builtin/providers/aws/resource_aws_eip_association.go
```diff
@@ -0,0 +1,157 @@
...
+	return resourceAwsEipAssociationRead(d, meta)
+}
+
+func resourceAwsEipAssociationRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	request := &ec2.DescribeAddressesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name:   aws.String("association-id"),
```
catsby: Should this be `allocation-id` ?

builtin/providers/aws/resource_aws_eip_association.go
```diff
@@ -0,0 +1,157 @@
...
+			},
+		},
+	}
+
+	response, err := conn.DescribeAddresses(request)
+	if err != nil {
+		return fmt.Errorf("Error reading EC2 Elastic IP %s: %#v", d.Get("allocation_id").(string), err)
+	}
+
+	return readAwsEipAssociation(d, response.Addresses[0])
```
catsby: we should check if `response.Addresses[0]` is non-nil before we pass it on, and likely error our if it is

builtin/providers/aws/resource_aws_eip_association.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	return nil
+}
+
+func readAwsEipAssociation(d *schema.ResourceData, address *ec2.Address) error {
+	if err := d.Set("allocation_id", *address.AllocationId); err != nil {
+		return err
+	}
+	if err := d.Set("instance_id", *address.InstanceId); err != nil {
```
catsby: We need to use `address.AllocationId` and `address.InstanceId` here, instead of dereferencing them, to avoid dereferencing nil pointers, like we do for `address.NetworkInterfaceId` et. al. below

builtin/providers/aws/resource_aws_eip_association_test.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+
+		return fmt.Errorf("Error finding instance/address")
+	}
+}
+
+func testAccCheckAWSEIPAssociationDestroy(s *terraform.State) error {
+	for _, rs := range s.RootModule().Resources {
+		log.Printf("\n\n----- This is never called")
```
catsby: Is this debugging, or comments? It should be removed for the former, a comment for the latter ... 

Shouldn't we check that the association is gone? E.g. query the API and make sure this EIP no longer has an association? Isn't that "destroyed", in this context?

builtin/providers/aws/resource_aws_eip_association_test.go
```diff
@@ -0,0 +1,118 @@
...
+						"aws_eip.bar.0", &a),
+					testAccCheckAWSEIPAssociationExists(
+						"aws_eip_association.by_allocation_id", &i, &a),
+					testAccCheckInstanceExists(
+						"aws_instance.foo.1", &i),
+					testAccCheckAWSEIPExists(
+						"aws_eip.bar.1", &a),
+					testAccCheckAWSEIPAssociationExists(
+						"aws_eip_association.by_public_ip", &i, &a),
+				),
```
catsby: Can we check the ENI association as well?

catsby:
Looks good so far, I've left some comments that we need addressed or discussed before we can merge. Thanks for taking this up, @stack72 !

builtin/providers/aws/resource_aws_eip_association.go
stack72: When an EIP is associated with something (ENI, Instance, Public IP etc), it get's an association-id. This was used as the ID. Therefore, the call to get the association uses the ID rather than the different scenarios it can be. An association_id will always be specific AFAICT

stack72:
Test's still pass as expected:

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEIPAssociation' 2>~/tf.log
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEIPAssociation -timeout 120m
=== RUN   TestAccAWSEIPAssociation_basic
--- PASS: TestAccAWSEIPAssociation_basic (163.37s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	163.391s
```

builtin/providers/aws/resource_aws_eip_association.go
catsby: Ah, I misread and thought allocation was the ID

catsby:
Good to go here

PR #6555, opened 2016-05-09
provider/aws: Opsworks Agent has a default, needs to be computed

catsby:
`agent_version` was introduced to Opsworks Stacks in https://github.com/hashicorp/terraform/pull/6493, however as an optional attribute that will have a default if not specified. This PR makes the attribute computed, if the user omits it. Fixes failing Acc tests:

```
--- FAIL: TestAccAWSOpsworksStackVpc (15.79s)
	testing.go:172: Step 0 error: After applying this step, the plan was not empty:
		
		DIFF:
		
		UPDATE: aws_opsworks_stack.tf-acc
		  agent_version: "3437-20160504095744" => ""
		
		STATE:
```


apparentlymart:
:+1: LGTM!

PR #6559, opened 2016-05-09
core: Add support for marking outputs as sensitive

jen20:
This pull request allows an output to be marked "sensitive", in which case the value is redacted in the post-refresh and post-apply list of outputs.

The configuration:

```tf
variable "input" {
    default = "Hello world"
}

output "notsensitive" {
    value = "${var.input}"
}

output "sensitive" {
    sensitive = true
    value = "${var.input}"
}
```

Would result in the output:

```
terraform apply

Outputs:

  notsensitive = Hello world
  sensitive    = <sensitive>
```

The `terraform output` command continues to display the value as before.

Note that sensitivity is not tracked internally, so if the output is interpolated in another module into a resource, the value will be displayed. The value is still present in the state.

### DIFF
--- a/command/apply.go
+++ b/command/apply.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/terraform/config"
 	"github.com/hashicorp/terraform/terraform"
 )
 
@@ -250,7 +251,7 @@ func (c *ApplyCommand) Run(args []string) int {
 	}
 
 	if !c.Destroy {
-		if outputs := outputsAsString(state); outputs != "" {
+		if outputs := outputsAsString(state, ctx.Module().Config().Outputs); outputs != "" {
 			c.Ui.Output(c.Colorize().Color(outputs))
 		}
 	}
@@ -376,7 +377,7 @@ Options:
 	return strings.TrimSpace(helpText)
 }
 
-func outputsAsString(state *terraform.State) string {
+func outputsAsString(state *terraform.State, schema []*config.Output) string {
 	if state == nil {
 		return ""
 	}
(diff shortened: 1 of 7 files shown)


phinze:
Needs docs - basically just convert what you wrote for the commit message over onto this page https://www.terraform.io/docs/configuration/outputs.html

command/apply.go
```diff
@@ -250,7 +251,7 @@ func (c *ApplyCommand) Run(args []string) int {
 	}
 
 	if !c.Destroy {
-		if outputs := outputsAsString(state); outputs != "" {
+		if outputs := outputsAsString(state, ctx.Module().Config().Outputs); outputs != "" {
```
phinze: `Module()` and `Config()` are guaranteed to be non-nil?

command/apply.go
```diff
@@ -400,11 +406,18 @@ func outputsAsString(state *terraform.State) string {
 		for _, k := range keys {
 			v := outputs[k]
 
-			outputBuf.WriteString(fmt.Sprintf(
-				"  %s%s = %s\n",
-				k,
-				strings.Repeat(" ", keyLen-len(k)),
-				v))
+			if schemaMap[k].Sensitive {
```
phinze: Could this panic if I have an output in the state but remove it from config?

phinze:
Functionality and code looking good. Two Qs inline.

command/apply.go
jen20: It shouldn't at this stage - apply has already run so the state will be modified.

command/apply.go
jen20: Yes, if we have got to this stage.

website/source/docs/configuration/outputs.html.md
```diff
@@ -57,3 +57,27 @@ output NAME {
...
+`terraform refresh`, sensitive outputs are redacted, with `<sensitive>`
+displayed in place of their value.
+
+### Limitations of Sensitive Outputs
+
+* the values of sensitive outputs are still stored in the Terraform
+  state, and available using the `terraform output` command, so cannot be
+  relied on as a sole means of protecting values.
+* sensitivity is not tracked internally, so if the output is interpolated in
+  another module into a resource, the value will be displayed.
```
phinze: nit - capitalize first words

phinze:
Tiny casing nit - this LGTM

website/source/docs/configuration/outputs.html.md
jen20: :+1:

PR #6564, opened 2016-05-09
fix typo vars PATh -> PATH

curx:


stack72:
Hi @curx 

thanks so much for the PR here to fix the example. I will get this merged when we finish the release process for 0.6.16

Paul

PR #6310, opened 2016-04-22
provider/google: Support static private IP addresses

evandbrown:
The private address of an instance's network interface may now be specified.
If no value is provided, an address will be chosen by Google Compute Engine
and that value will be read into Terraform state.

PR #6574, opened 2016-05-10
provider/google: correct error messages in acceptance tests

billf:


billf:
i can squash these, but i left them separate because it's not clear why region was set to `us-central-1` to begin with: the API won't accept that as a value.

billf:
safe per #6310

evandbrown:
lgtm, ready to merge.

jen20:
Thanks @billf!

Issue #6561, opened 2016-05-09
Terraform v0.6.16 Release Process


jen20:
Hello again @apparentlymart, @jtopjian, @sparkprime, @stack72, @radeksimko, @svanharmelen , @lwander and @ack!

It's Terraform release time again, this time for v0.6.16

Please hold off on merges until this issue is closed - thanks!

phinze:
Reopening until dev-0.7 merge is done.

jen20:
This is done! Thanks everyone.

PR #6584, opened 2016-05-10
fix(Makefile): remove 'updatedeps' from .PHONY

samber:
Not used anymore

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,4 +84,4 @@ fmt:
 fmtcheck:
 	@sh -c "'$(CURDIR)/scripts/gofmtcheck.sh'"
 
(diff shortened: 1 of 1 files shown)


stack72:
Hi @samber 

Thanks so much for this PR - I will merge this when the release process for 0.6.16 has officially ended

Paul

apparentlymart:
I merged this before I realized that the 0.6.16 release process issue #6561 actually remains open, which was my bad... but I'm going to leave it there because it's a minor change that shouldn't affect any ongoing release processes and so doesn't seem worth the history noise of reverting it and re-merging it again.

PR #6585, opened 2016-05-10
aws_instance BC between 0.6.14 and 0.6.15+ in Changelog

radeksimko:
Wondering why does terraform want to recreate my `aws_instance`s I dug deeper to find this commit https://github.com/hashicorp/terraform/commit/d85df6352621e172da0c3c2037f84a1e9a679694 which apparently introduced BC between `0.6.14` and `0.6.15`.

This made me realise I'm the bad guy 🙈  that is still using `security_groups` for VPC security groups, which I obviously shouldn't be doing.

I however think that user should know about this change in behaviour (previously this was silently ignored).

```
-/+ aws_instance.nat
...
    security_groups.#:           "0" => "1" (forces new resource)
    security_groups.1521004543:  "" => "sg-11be0033" (forces new resource)
...
```

cc @mitchellh ^

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,10 @@ BUG FIXES:
 
 ## 0.6.15 (April 22, 2016)
 
+BACKWARDS INCOMPATIBILITIES / NOTES:
+
+ * `aws_instance` - if you still use `security_groups` field for SG IDs - i.e. inside VPC, this will generate diffs during `plan` and `apply` will **recreate** the resource. Terraform expects IDs (VPC SGs) inside `security_group_ids`.
+
 FEATURES:
 
  * **New command:** `terraform fmt` to automatically normalize config file style ([#4955](https://github.com/hashicorp/terraform/issues/4955))


apparentlymart:
This was reported in a number of separate issues (which sadly I was too lazy to track down here) and seems to be tripping up lots of people.

Looks like the `Read` code is normalizing the values to be in the correct location.

Good news is that other folks had success just changing the config to use `vpc_security_groups`, after which the diff goes away and the instance doesn't need to get rebuilt. Bad news is that it's not immediately obvious to people what they should do here...

CHANGELOG.md
```diff
@@ -90,6 +90,10 @@ BUG FIXES:
 
 ## 0.6.15 (April 22, 2016)
 
+BACKWARDS INCOMPATIBILITIES / NOTES:
+
+ * `aws_instance` - if you still use `security_groups` field for SG IDs - i.e. inside VPC, this will generate diffs during `plan` and `apply` will **recreate** the resource. Terraform expects IDs (VPC SGs) inside `security_group_ids`.
```
vitroth: The correct field is `vpc_security_group_ids` not `security_group_ids`.

radeksimko: Good catch! I'll fix that.

stack72:
Hi @radeksimko 

I went ahead and made this change for you while pruning some PRs - https://github.com/hashicorp/terraform/commit/3de2ab04373ba90a17f20e5d538a12460d4d0410

P.

PR #6587, opened 2016-05-10
provider/fastly: add support for custom VCL configuration

mattmoyer:
This PR adds a new `vcl` configuration block to support Fastly [custom VCL](https://docs.fastly.com/guides/vcl/uploading-custom-vcl). VCL is the [Varnish Configuration Language](https://www.varnish-cache.org/docs/2.1/tutorial/vcl.html), which is a compiled C-like DSL.

[Fastly's API model](https://docs.fastly.com/api/config#vcl) for custom VCL is that you have zero or more VCL files associated with your service, with one of them set as the "main" block. The main block needs to define some entrypoints that are invoked directly by the Varnish request handling flow, the other non-main blocks are available to include as libraries in the main block.

#### Usage
```hcl
resource "fastly_service_v1" "demo" {
  name = "demofastly"

  domain {
    name    = "demo.notexample.com"
    comment = "demo"
  }

  backend {
    address = "127.0.0.1"
    name    = "localhost"
    port    = 80
  }

  force_destroy = true

  vcl {
    name = "my_custom_main_vcl"
    content = "${file("${path.module}/my_custom_main.vcl")}"
    main = true
  }

  vcl {
    name = "my_custom_library_vcl"
    content = "${file("${path.module}/my_custom_library.vcl")}"
  }
}
```

#### Caveats
Fastly does not allow custom VCL by default on new accounts. I think it is only available on paid accounts if you request access and describe your planned use case. I noted this in the documentation, but because of this I didn't (yet) write any acceptance tests for this functionality.

I'm open to any other ideas about how to represent VCL in this provider.

### DIFF
--- a/builtin/providers/fastly/resource_fastly_service_v1.go
+++ b/builtin/providers/fastly/resource_fastly_service_v1.go
@@ -396,11 +396,40 @@ func resourceServiceV1() *schema.Resource {
 					},
 				},
 			},
+
+			"vcl": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": &schema.Schema{
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "A name to refer to this VCL configuration",
+						},
+						"content": &schema.Schema{
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "The contents of this VCL configuration",
+						},
+						"main": &schema.Schema{
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Default:     false,
+							Description: "Should this VCL configuation be the main configuration",
+						},
+					},
+				},
+			},
 		},
 	}
 }
 
 func resourceServiceV1Create(d *schema.ResourceData, meta interface{}) error {
+	if err := validateVCLs(d); err != nil {
+		return err
+	}
+
 	conn := meta.(*FastlyClient).conn
 	service, err := conn.CreateService(&gofastly.CreateServiceInput{
 		Name:    d.Get("name").(string),
@@ -416,6 +445,10 @@ func resourceServiceV1Create(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
+	if err := validateVCLs(d); err != nil {
+		return err
+	}
+
 	conn := meta.(*FastlyClient).conn
 
 	// Update Name. No new verions is required for this
@@ -443,6 +476,7 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
 		"gzip",
 		"s3logging",
 		"condition",
+		"vcl",
 	} {
 		if d.HasChange(v) {
 			needsChange = true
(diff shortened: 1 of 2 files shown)


mattmoyer:
I originally had this implemented on top of https://github.com/hashicorp/terraform/pull/6508 (`ValidateFunc` for lists and sets), but I don't think I have the time to finish implementing that so I went ahead and tweaked this to just validate on `apply`.

builtin/providers/fastly/resource_fastly_service_v1.go
```diff
@@ -854,6 +888,73 @@ func resourceServiceV1Update(d *schema.ResourceData, meta interface{}) error {
...
+			if newVCLVal == nil {
+				newVCLVal = new(schema.Set)
+			}
+
+			oldVCLSet := oldVCLVal.(*schema.Set)
+			newVCLSet := newVCLVal.(*schema.Set)
+
+			remove := oldVCLSet.Difference(newVCLSet).List()
+			add := newVCLSet.Difference(oldVCLSet).List()
+			_ = add
```
catsby: Is this leftover from debugging, or doing something I don't yet understand?

mattmoyer: Oops, that was just debugging cruft leftover. Assigning to `_` lets you get around `add declared and not used` errors while you're in the middle of things.

website/source/docs/providers/fastly/r/service_v1.html.markdown
```diff
@@ -204,6 +239,12 @@ compressed. Default `0`
 Apache Common Log format (`%h %l %u %t %r %>s`)
 * `timestamp_format` - (Optional) `strftime` specified timestamp formatting (default `%Y-%m-%dT%H:%M:%S.000`).
 
+The `vcl` block supports:
+
+* `name` - (Required) A unique name for this configuration block
+* `content` - (Required) The custom VCL code to upload.
+* `main` - (Optional) If `true`, use this block as the main configuration. If
```
catsby: Should not that _There Can Be Only One_ `main` block

builtin/providers/fastly/resource_fastly_service_v1.go
catsby: yup, thought so, thanks!

catsby:
Wonderful! This was on my very short list of things to do, thanks! 
As I have not gotten to it myself, I had not yet requested VCL support for our testing account. I've just emailed support to request it, so as soon as I get word I'll probably just merge this and add tests myself later.

The code looks good otherwise so I may just merge today anyway.

Thanks!

website/source/docs/providers/fastly/r/service_v1.html.markdown
mattmoyer: Ha, that is probably really a good thing to note here, though.

mattmoyer:
@catsby sounds great. I just rebased some of those little fixes so I think this should be fine to merge whenever you're ready.

catsby:
Fastly already got back to me and enabled VCL support on our test account ❤️ 

I'll work on some tests and try to get this in today. Thanks again!

catsby:
I opened #6662 to continue this work here by adding an acceptance test, so I'm going to close this now. Thanks for the contribution! #6662 includes your commits so you'll still get attribution 😄

PR #6557, opened 2016-05-09
core: Fix interp error msgs on module vars during destroy

phinze:
Wow this one was tricky!

This bug presents itself only when using planfiles, because when doing a
straight `terraform apply` the interpolations are left in place from the
Plan graph walk and paper over the issue. (This detail is what made it
so hard to reproduce initially.)

Basically, graph nodes for module variables are visited during the apply
walk and attempt to interpolate. During a destroy walk, no attributes
are interpolated from resource nodes, so these interpolations fail.

This scenario is supposed to be handled by the `PruneNoopTransformer` -
in fact it's described as the example use case in the comment above it!

So the bug had to do with the actual behavor of the Noop transformer.
The resource nodes were not properly reporting themselves as Noops
during a destroy, so they were being left in the graph.

This in turn triggered the module variable nodes to see that they had
another node depending on them, so they also reported that they could
not be pruned.

Therefore we had two nodes in the graph that were effectively noops but
were being visited anyways. The module variable nodes were already graph
leaves, which is why this error presented itself as just stray messages
instead of actual failure to destroy.

Fixes #5440
Fixes #5708
Fixes #4988
Fixes #3268

Issue #5440, opened 2016-03-03
Error destroying (but it worked)


sethvargo:
```
Terraform v0.6.12

Setting up remote state...
Downloading remote modules...
Refreshing Terraform state prior to plan...
aws_instance.nomad_client.0: Destroying...
aws_route_table_association.hashicorp-training: Destroying...
aws_route_table_association.hashicorp-training: Destruction complete
aws_route_table.hashicorp-training: Destroying...
aws_instance.server.1: Destroying...
aws_instance.server.0: Destroying...
aws_instance.server.2: Destroying...
aws_route_table.hashicorp-training: Destruction complete
aws_internet_gateway.hashicorp-training: Destroying...
aws_instance.nomad_client.0: Destruction complete
aws_instance.nomad_server.0: Destroying...
aws_instance.server.1: Destruction complete
aws_instance.server.0: Destruction complete
aws_instance.server.2: Destruction complete
aws_instance.nomad_server.0: Destruction complete
aws_instance.nomad_server.2: Destroying...
aws_instance.nomad_server.1: Destroying...
aws_instance.nomad_server.2: Destruction complete
aws_internet_gateway.hashicorp-training: Destruction complete
aws_instance.nomad_server.1: Destruction complete
aws_subnet.hashicorp-training: Destroying...
aws_key_pair.hashicorp-training: Destroying...
aws_security_group.hashicorp-training: Destroying...
aws_key_pair.hashicorp-training: Destruction complete
aws_security_group.hashicorp-training: Destruction complete
aws_subnet.hashicorp-training: Destruction complete
aws_vpc.hashicorp-training: Destroying...
aws_vpc.hashicorp-training: Destruction complete
Error applying plan:

3 error(s) occurred:

* Resource 'aws_key_pair.hashicorp-training' does not have attribute 'key_name' for variable 'aws_key_pair.hashicorp-training.key_name'
* Resource 'aws_security_group.hashicorp-training' does not have attribute 'id' for variable 'aws_security_group.hashicorp-training.id'
* Resource 'aws_subnet.hashicorp-training' does not have attribute 'id' for variable 'aws_subnet.hashicorp-training.id'

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

Terraform configs: https://www.dropbox.com/s/krwdfk52xbj8xyx/nomad-repro.zip?dl=0

[...]

Issue #3268, opened 2015-09-17
DO Droplet atlas_artifact destroy error


bensojona:
Provisioning a DO droplet with `terraform apply` that uses an `atlas_artifact` `digitalocean.droplet`, then doing a `terraform destroy` destroys the resource appropriately but throws the below error.

`* Resource 'atlas_artifact.web' does not have attribute 'id' for variable 'atlas_artifact.web.id'`
 
See [gist](https://gist.github.com/bensojona/423a8f9c37291cf508dd) to reproduce. Will need to set environment variables `DIGITALOCEAN_TOKEN`, `ATLAS_TOKEN`, and `ATLAS_USERNAME`.

- Create DO droplet using Packer droplet.json template
- Upload public key `example.pub` to your DO account
- `terraform get`
- `TF_VAR_digital_ocean_token=$DIGITALOCEAN_TOKEN TF_VAR_atlas_token=$ATLAS_TOKEN TF_VAR_atlas_username=$ATLAS_USERNAME TF_VAR_ssh_fingerprint=3c:e3:4d:a6:81:22:ac:6e:41:c1:f6:85:e1:4b:0f:ec terraform apply`
- `TF_VAR_digital_ocean_token=$DIGITALOCEAN_TOKEN TF_VAR_atlas_token=$ATLAS_TOKEN TF_VAR_atlas_username=$ATLAS_USERNAME TF_VAR_ssh_fingerprint=3c:e3:4d:a6:81:22:ac:6e:41:c1:f6:85:e1:4b:0f:ec terraform destroy`
- `TF_VAR_digital_ocean_token=$DIGITALOCEAN_TOKEN TF_VAR_atlas_token=$ATLAS_TOKEN TF_VAR_atlas_username=$ATLAS_USERNAME TF_VAR_ssh_fingerprint=3c:e3:4d:a6:81:22:ac:6e:41:c1:f6:85:e1:4b:0f:ec terraform plan`
- `TF_VAR_digital_ocean_token=$DIGITALOCEAN_TOKEN TF_VAR_atlas_token=$ATLAS_TOKEN TF_VAR_atlas_username=$ATLAS_USERNAME TF_VAR_ssh_fingerprint=3c:e3:4d:a6:81:22:ac:6e:41:c1:f6:85:e1:4b:0f:ec terraform destroy`

[...]

PR #6599, opened 2016-05-10
terraform: Correct fix for destroy interp errors

phinze:
The fix that landed in #6557 was unfortunately the wrong subset of the
work I had been doing locally, and users of the attached bugs are still
reporting problems with Terraform v0.6.16.

At the very last step, I attempted to scope down both the failing test
and the implementation to their bare essentials, but ended up with a
test that did not exercise the root of the problem and a subset of the
implementation that was insufficient for a full bugfix.

The key thing I removed from the test was a _referencing output_ for the
module, which is what breaks down the #6557 solution.

I've re-tested the examples in #5440 and #3268 to verify this solution
does indeed solve the problem.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -2549,19 +2549,16 @@ func TestContext2Apply_destroyModuleWithAttrsReferencingResource(t *testing.T) {
 			Providers: map[string]ResourceProviderFactory{
 				"aws": testProviderFuncFixed(p),
 			},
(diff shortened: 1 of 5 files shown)


phinze:
After review, I'll backport this onto the `maint-0.6` branch to stage it in case we want to do a 0.6.17 maintenance release.

jen20:
This LGTM assuming Travis is happy!

PR #6606, opened 2016-05-11
build: "dev" and "quickdev" targets build separate binaries

apparentlymart:
The usual dev workflow is to run separate builds of core (using `make core-dev`) and each plugin (using `make plugin-dev`), but before the plugins were internalized there was also `make dev` and `make quickdev` to build core and the separate plugin binaries in a single step. With plugins internalized these became confusingly incompatible with the `core-dev` and `plugin-dev` targets, by producing a bundled binary.

In order to restore consistency to these dev targets, "dev" is changed to always mean "separated plugins", even when building the core and plugins in a single step using `dev` and `quickdev`.

As a Terraform developer, the idea would be to sync up with the latest code each morning (or thereabouts) by running `git pull` followed by `make dev`, and the throughout the day do one-off builds of either core or individual plugins, using unbundled plugin binaries in all cases.

A bundled binary can still be produced using `make bin`, but it's expected that this will be rare in the development workflow because it is much slower than building core and the plugins individually as needed. This target is primarily for release engineering.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -17,6 +17,7 @@ GIT_DIRTY=$(test -n "`git status --porcelain`" && echo "+CHANGES" || true)
 # Determine the arch/os combos we're building for
 XC_ARCH=${XC_ARCH:-"386 amd64 arm"}
 XC_OS=${XC_OS:-linux darwin windows freebsd openbsd solaris}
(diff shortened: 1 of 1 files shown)


apparentlymart:
(@phinze, this is as we discussed earlier today)

PR #6607, opened 2016-05-11
build: "make fmt" to ignore vendor packages

apparentlymart:
Several of our vendered dependencies are not gofmt-compliant, but we don't want to fix that since the vendored code is supposed to exactly match upstream.


jen20:
:+1: - been meaning to do this for a while!

PR #6616, opened 2016-05-11
IPv6 support for the communicator.

kristinn:
Now you can provision machines over IPv6.

### DIFF
--- a/communicator/shared/shared.go
+++ b/communicator/shared/shared.go
@@ -1,0 +1,26 @@
+package shared
+
+import (
+	"fmt"
+	"net"
+)
+
+// HostFormatter helps out with host formatting.
+// For instance, the host parameter differs when connecting to an IPv4 address vs. an IPv6 address.
+type HostFormatter interface {
+	Format(string) string
+}
+
+// HostFormatterImpl implements the HostFormatter interface.
+type HostFormatterImpl struct{}
+
+// Format formats the host/IP correctly, so we don't provide IPv6 address in an IPv4 format during node communication.
+func (h *HostFormatterImpl) Format(host string) string {
+	ip := net.ParseIP(host)
+	// Return the host as is if it's either a hostname or an IPv4 address.
+	if ip == nil || ip.To4() != nil {
+		return host
+	}
+
+	return fmt.Sprintf("[%s]", host)
+}
--- a/communicator/shared/shared.go
+++ b/communicator/shared/shared.go
@@ -5,22 +5,13 @@ import (
 	"net"
 )
 
-// HostFormatter helps out with host formatting.
-// For instance, the host parameter differs when connecting to an IPv4 address vs. an IPv6 address.
-type HostFormatter interface {
-	Format(string) string
-}
-
-// HostFormatterImpl implements the HostFormatter interface.
-type HostFormatterImpl struct{}
-
-// Format formats the host/IP correctly, so we don't provide IPv6 address in an IPv4 format during node communication.
-func (h *HostFormatterImpl) Format(host string) string {
-	ip := net.ParseIP(host)
-	// Return the host as is if it's either a hostname or an IPv4 address.
-	if ip == nil || ip.To4() != nil {
-		return host
+// IpFormat formats the IP correctly, so we don't provide IPv6 address in an IPv4 format during node communication. We return the ip parameter as is if it's an IPv4 address or a hostname.
+func IpFormat(ip string) string {
+	ipObj := net.ParseIP(ip)
+	// Return the ip/host as is if it's either a hostname or an IPv4 address.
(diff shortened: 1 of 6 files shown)


kristinn:
This change is needed to be able to provision over SSH.

I haven't done any tests towards winrm. Can someone with proper Azure/winrm knowledge comment on this? If this is the case, then I will also create an Azure test.

chrislovecnm:
@jen20 @stack72 @phinze you mind reviewing?

chrislovecnm:
@kristinn do we need to update docs?  Appreciate the tests!!

kristinn:
@stack72 it's not only vsphere improvement, it's just general provisioning improvement. So far providing ipv6 addresses to the provisioners haven't worked.

kristinn:
@chrislovecnm I'll look into it, see if the docs need to be updated.

sprokopiak:
This pull request looks to have solved my issue using the vsphere provider and winrm getting a poperly formatted IPv6 address. However, I'm unable to fully test if winrm will connect properly as I just found out my org has disabled IPv6 in vsphere.

chrislovecnm:
@stack72 @jen20 what testing needs to happen in core to have this merged?

Thanks

chrislovecnm:
@stack72 @jen20 @phinze ~ how do we get this into the code base?

Thanks

Chris

cc: @markpeek

chrislovecnm:
Bueller?... Bueller?... Bueller? ... this is blocking users who have ipv6 ... and surprisingly we have a few :)

chrislovecnm:
@jen20 @stack72 @phinze @catsby how do we get this merged?  Help is appreciated!!

chrislovecnm:
@jen20 can we get this merged?

chrislovecnm:
@jen20 @stack72 this resolves one of our open issues.  When can we get this merged?

communicator/shared/shared.go
```diff
@@ -0,0 +1,26 @@
...
+// For instance, the host parameter differs when connecting to an IPv4 address vs. an IPv6 address.
+type HostFormatter interface {
+	Format(string) string
+}
+
+// HostFormatterImpl implements the HostFormatter interface.
+type HostFormatterImpl struct{}
+
+// Format formats the host/IP correctly, so we don't provide IPv6 address in an IPv4 format during node communication.
+func (h *HostFormatterImpl) Format(host string) string {
```
josephholsten: is there a reason you're implementing this with a method receiver instead of just a function? Are you actually using this interface with any other types?

josephholsten:
I like it. I think you could simplify to just a function for the host formatter, but I'll live either way.

josephholsten:
I like it. I think you could simplify to just a function for the host formatter, but I'll live either way.

communicator/shared/shared.go
kristinn: No not really. I just usually do it like this since it makes creating tests easier. In this case the tests aren't really taking advantage of this, so in this case it's not needed, unless we want to "overwrite" the Format() function in a test case.

I can change this to be just a single function. What is your opinion on that?

josephholsten: I'd keep it simple until we actually need it.

kristinn: Ok. I will change it tomorrow.

kristinn:
@josephholsten @chrislovecnm I just updated this pull request to make it simpler.

kristinn:
I also added missing tests for the winrm provisioner.

kristinn:
@chrislovecnm @jen20 @stack72 @phinze @catsby is there a reason why this isn't being merged?

communicator/shared/shared.go
```diff
@@ -0,0 +1,17 @@
...
+
+// IpFormat formats the IP correctly, so we don't provide IPv6 address in an IPv4 format during node communication. We return the ip parameter as is if it's an IPv4 address or a hostname.
+func IpFormat(ip string) string {
+	ipObj := net.ParseIP(ip)
+	// Return the ip/host as is if it's either a hostname or an IPv4 address.
+	if ipObj == nil || ipObj.To4() != nil {
+		return ip
+	}
+
+	return fmt.Sprintf("[%s]", ip)
```
jtopjian: Just to confirm: if the address already has brackets, `net.ParseIP` will return nil, the address will be returned as-is, and used as the connection address?

kristinn: That is correct.

If the input is not a parseable IP or if it's a valid IPv4 address, the input is returned as is.

Otherwise the input is returned surrounded by brackets.

jen20:
Thanks @kristinn, sorry this took a while to get around to!

kristinn:
No problem. Happy it's merged. 😄

PR #6619, opened 2016-05-11
provider/vsphere: Virtual machine update

dkalleg:
Appears to be working in happy path. Concerns around tfstate file
best practices.  If we write out all disk info we can query from
vSphere, we end up detecting diffs on successive apply/plan calls.
If best practice is to only write out what user provided, how best to
map a potentially long list of disks without uuids to a list queried
from vSphere? If we can assume ording is always preserved, then the
problem is simplified.

Posting as a still WIP to generate discussion on how stateful info is handled.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -39,11 +39,13 @@ type networkInterface struct {
 }
 
 type hardDisk struct {
+	name       string
 	size       int64
 	iops       int64
 	initType   string
 	vmdkPath   string
 	controller string
+	bootable   bool
 }
 
 //Additional options Vsphere can use clones of windows machines
@@ -82,7 +84,7 @@ type virtualMachine struct {
 	timeZone              string
 	dnsSuffixes           []string
 	dnsServers            []string
-	bootableVmdk          bool
+	hasBootableVmdk       bool
 	linkedClone           bool
 	skipCustomization     bool
 	windowsOptionalConfig windowsOptConfig
@@ -320,21 +322,28 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 			},
 
 			"disk": &schema.Schema{
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Required: true,
-				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"uuid": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
+						"key": &schema.Schema{
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+
 						"template": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
-							ForceNew: true,
 						},
 
 						"type": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
-							ForceNew: true,
 							Default:  "eager_zeroed",
 							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 								value := v.(string)
@@ -349,34 +358,37 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 						"datastore": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
-							ForceNew: true,
 						},
 
 						"size": &schema.Schema{
 							Type:     schema.TypeInt,
 							Optional: true,
-							ForceNew: true,
+						},
+
+						"name": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
 						},
 
 						"iops": &schema.Schema{
 							Type:     schema.TypeInt,
 							Optional: true,
-							ForceNew: true,
 						},
 
 						"vmdk": &schema.Schema{
 							// TODO: Add ValidateFunc to confirm path exists
 							Type:     schema.TypeString,
 							Optional: true,
-							ForceNew: true,
-							Default:  "",
 						},
 
 						"bootable": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
-							Default:  false,
-							ForceNew: true,
+						},
+
+						"keep_on_remove": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
 						},
 
 						"controller_type": &schema.Schema{
@@ -442,11 +454,6 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 		rebootRequired = true
 	}
 
-	// do nothing if there are no changes
-	if !hasChanges {
-		return nil
-	}
-
 	client := meta.(*govmomi.Client)
 	dc, err := getDatacenter(client, d.Get("datacenter").(string))
 	if err != nil {
@@ -459,6 +466,100 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 	if err != nil {
 		return err
 	}
+
+	if d.HasChange("disk") {
+		hasChanges = true
+		oldDisks, newDisks := d.GetChange("disk")
+		oldDiskSet := oldDisks.(*schema.Set)
+		newDiskSet := newDisks.(*schema.Set)
+
+		addedDisks := newDiskSet.Difference(oldDiskSet)
+		removedDisks := oldDiskSet.Difference(newDiskSet)
+
+		// Removed disks
+		for _, diskRaw := range removedDisks.List() {
+			if disk, ok := diskRaw.(map[string]interface{}); ok {
+				devices, err := vm.Device(context.TODO())
+				if err != nil {
+					return fmt.Errorf("[ERROR] Update Remove Disk - Could not get virtual device list: %v", err)
+				}
+				virtualDisk := devices.FindByKey(int32(disk["key"].(int)))
+
+				keep := false
+				if v, ok := d.GetOk("keep_on_remove"); ok {
+					keep = v.(bool)
+				}
+
+				err = vm.RemoveDevice(context.TODO(), keep, virtualDisk)
+				if err != nil {
+					return fmt.Errorf("[ERROR] Update Remove Disk - Error removing disk: %v", err)
+				}
+			}
+		}
+		// Added disks
+		for _, diskRaw := range addedDisks.List() {
+			if disk, ok := diskRaw.(map[string]interface{}); ok {
+
+				var datastore *object.Datastore
+				if disk["datastore"] == "" {
+					datastore, err = finder.DefaultDatastore(context.TODO())
+					if err != nil {
+						return fmt.Errorf("[ERROR] Update Remove Disk - Error finding datastore: %v", err)
+					}
+				} else {
+					datastore, err = finder.Datastore(context.TODO(), disk["datastore"].(string))
+					if err != nil {
+						log.Printf("[ERROR] Couldn't find datastore %v.  %s", disk["datastore"].(string), err)
+						return err
+					}
+				}
+
+				var size int64
+				if disk["size"] == 0 {
+					size = 0
+				} else {
+					size = int64(disk["size"].(int))
+				}
+				iops := int64(disk["iops"].(int))
+				controller_type := disk["controller"].(string)
+
+				var mo mo.VirtualMachine
+				vm.Properties(context.TODO(), vm.Reference(), []string{"summary", "config"}, &mo)
+
+				var diskPath string
+				switch {
+				case disk["vmdk"] != "":
+					diskPath = disk["vmdk"].(string)
+				case disk["name"] != "":
+					snapshotFullDir := mo.Config.Files.SnapshotDirectory
+					split := strings.Split(snapshotFullDir, " ")
+					if len(split) != 2 {
+						return fmt.Errorf("[ERROR] createVirtualMachine - failed to split snapshot directory: %v", snapshotFullDir)
+					}
+					vmWorkingPath := split[1]
+					diskPath = vmWorkingPath + disk["name"].(string)
+				default:
+					return fmt.Errorf("[ERROR] resourceVSphereVirtualMachineUpdate - Neither vmdk path nor vmdk name was given")
+				}
+
+				log.Printf("[INFO] Attaching disk: %v", diskPath)
+				err = addHardDisk(vm, size, iops, "thin", datastore, diskPath, controller_type)
+				if err != nil {
+					log.Printf("[ERROR] Add Hard Disk Failed: %v", err)
+					return err
+				}
+			}
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// do nothing if there are no changes
+	if !hasChanges {
+		return nil
+	}
+
 	log.Printf("[DEBUG] virtual machine config spec: %v", configSpec)
 
 	if rebootRequired {
@@ -646,49 +747,82 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 	}
 
 	if vL, ok := d.GetOk("disk"); ok {
-		disks := make([]hardDisk, len(vL.([]interface{})))
-		for i, v := range vL.([]interface{}) {
-			disk := v.(map[string]interface{})
-			if i == 0 {
+		if diskSet, ok := vL.(*schema.Set); ok {
+
+			disks := []hardDisk{}
+			hasBootableDisk := false
+			for _, value := range diskSet.List() {
+				disk := value.(map[string]interface{})
+				newDisk := hardDisk{}
+
 				if v, ok := disk["template"].(string); ok && v != "" {
-					vm.template = v
-				} else {
-					if v, ok := disk["size"].(int); ok && v != 0 {
-						disks[i].size = int64(v)
-					} else if v, ok := disk["vmdk"].(string); ok && v != "" {
-						disks[i].vmdkPath = v
-						if v, ok := disk["bootable"].(bool); ok {
-							vm.bootableVmdk = v
-						}
-					} else {
-						return fmt.Errorf("template, size, or vmdk argument is required")
+					if v, ok := disk["name"].(string); ok && v != "" {
+						return fmt.Errorf("Cannot specify name of a template")
 					}
+					vm.template = v
+					if hasBootableDisk {
+						return fmt.Errorf("[ERROR] Only one bootable disk or template may be given")
+					}
+					hasBootableDisk = true
 				}
+
+				if v, ok := disk["type"].(string); ok && v != "" {
+					newDisk.initType = v
+				}
+
 				if v, ok := disk["datastore"].(string); ok && v != "" {
 					vm.datastore = v
 				}
-			} else {
+
 				if v, ok := disk["size"].(int); ok && v != 0 {
-					disks[i].size = int64(v)
-				} else if v, ok := disk["vmdk"].(string); ok && v != "" {
-					disks[i].vmdkPath = v
-				} else {
-					return fmt.Errorf("size or vmdk argument is required")
+					if v, ok := disk["template"].(string); ok && v != "" {
+						return fmt.Errorf("Cannot specify size of a template")
+					}
+
+					if v, ok := disk["name"].(string); ok && v != "" {
+						newDisk.name = v
+					} else {
+						return fmt.Errorf("[ERROR] Disk name must be provided when creating a new disk")
+					}
+
+					newDisk.size = int64(v)
 				}
 
+				if v, ok := disk["iops"].(int); ok && v != 0 {
+					newDisk.iops = int64(v)
+				}
+
+				if v, ok := disk["controller_type"].(string); ok && v != "" {
+					newDisk.controller = v
+				}
+
+				if vVmdk, ok := disk["vmdk"].(string); ok && vVmdk != "" {
+					if v, ok := disk["template"].(string); ok && v != "" {
+						return fmt.Errorf("Cannot specify a vmdk for a template")
+					}
+					if v, ok := disk["size"].(string); ok && v != "" {
+						return fmt.Errorf("Cannot specify size of a vmdk")
+					}
+					if v, ok := disk["name"].(string); ok && v != "" {
+						return fmt.Errorf("Cannot specify name of a vmdk")
+					}
+					if vBootable, ok := disk["bootable"].(bool); ok {
+						hasBootableDisk = true
+						newDisk.bootable = vBootable
+						vm.hasBootableVmdk = vBootable
+					}
+					newDisk.vmdkPath = vVmdk
+				}
+				// Preserves order so bootable disk is first
+				if newDisk.bootable == true || disk["template"] != "" {
+					disks = append([]hardDisk{newDisk}, disks...)
+				} else {
+					disks = append(disks, newDisk)
+				}
 			}
-			if v, ok := disk["iops"].(int); ok && v != 0 {
-				disks[i].iops = int64(v)
-			}
-			if v, ok := disk["type"].(string); ok && v != "" {
-				disks[i].initType = v
-			}
-			if v, ok := disk["controller_type"].(string); ok && v != "" {
-				disks[i].controller = v
-			}
+			vm.hardDisks = disks
+			log.Printf("[DEBUG] disk init: %v", disks)
 		}
-		vm.hardDisks = disks
-		log.Printf("[DEBUG] disk init: %v", disks)
 	}
 
 	if vL, ok := d.GetOk("cdrom"); ok {
@@ -722,7 +856,7 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 }
 
 func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{}) error {
-	log.Printf("[DEBUG] reading virtual machine: %#v", d)
+	log.Printf("[DEBUG] virtual machine resource data: %#v", d)
 	client := meta.(*govmomi.Client)
 	dc, err := getDatacenter(client, d.Get("datacenter").(string))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -336,15 +340,6 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							ForceNew: true,
-							Default:  "eager_zeroed",
```
dkalleg: FYI, I removed the default here because the default + ForceNew: true creates an interesting situation.  If the user were to not set the type, the .tfstate file would still end up with an entry for type = "eager_zeroed".  If you run terraform plan immediately after, it would tell you that you are attempting to alter the "type" state from "eager_zeroed" to "" (since your .tf script does not set "type").  With the addition of the ForceNew, this would blow away the VM.  Bad news bears.  It seems that Default values don't play nicely with Update scenarios.  I moved the Default setting into the Create() block, which is functionally equivalent.

dkalleg:
Heres an example of the tfstate file concerns I'm referring to.  I make a vm with three disks attached, defined so:
```
    disk {
        vmdk = "images/ubuntu.vmdk"
        bootable = true
    }
    disk {
        size = 2
    }
    disk {
        vmdk = "images/test.vmdk"
    }
```

With the current implementation and no changes in the .tf script file, I run terraform plan and it detects various "differences" :
```
~ vsphere_virtual_machine.dansTerraformVm
    disk.0.datastore:                       "local" => ""
    disk.0.size:                            "8" => "0"
    disk.1.datastore:                       "local" => ""
    disk.1.type:                            "thin" => ""
    disk.1.vmdk:                            "tfTestVmDisk_31/tfTestVmDisk.vmdk" => ""
    disk.2.datastore:                       "local" => ""
    disk.2.size:                            "8" => "0"
```

Disk 0, I did not specify the datastore or size.  After the read() at the end of the first terraform apply, we discover the datastore and size of the disk.  Since they are written out to tfstate file (via d.Set()), the following terraform plan see's the "difference" between the .tf script (which does not specify a datastore or size, meaning the resolve out to "" and 0), and what is recorded in the tfstate file ("local" and 8).

Apparently TF is smart enough to realize if you want to "update" the variable to an empty init value (ie. "local" => "") it does not trigger the ForceNew.  Yay.  But it does mean we will see that in the output whenever terraform plan is run.  

Functionally, in regard to disk updates, this seems to be working.  I'll be putting it through some more involved / less "happy-path" testing to confirm this.

My last note has to do with the initType of a disk.  When creating a new disk, I seem to always find the disk properties returned from vSphere show its 'ThinProvisioned' value as true and 'EagerlyScrub' as nil, regardless if what I set in my .tf script.  This is suggesting that something is broken around the initType setting when creating the disk.  Since that is outside the scope of this PR I'm mostly ignoring it.  I'm setting the value of a disks init type to "thin" whenever I see ThinProvisioned is true and ignoring other cases.  TODO there for once this problem is figured out.

Phew.

@phinze @stack72 Please review at your earliest convenience.  Please excuse all my debug logging for now :)

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -356,7 +351,7 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 						"size": &schema.Schema{
 							Type:     schema.TypeInt,
 							Optional: true,
-							ForceNew: true,
```
chrislovecnm: Reason again for `ForceNew: true`?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -451,6 +439,84 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 	if err != nil {
 		return err
 	}
+
+	if d.HasChange("disk") {
+		hasChanges = true
+		oldDisks, newDisks := d.GetChange("disk")
```
chrislovecnm: Does this ever return an error?  May want to look at error handling.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -451,6 +439,84 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+		oldDiskList := oldDisks.([]interface{})
+		newDiskList := newDisks.([]interface{})
```
chrislovecnm: Dumb question but can either of these return as nil?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -451,6 +439,84 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+			removedDisks := oldDiskList[len(newDiskList):]
+			log.Printf("[DEBUG] resourceVSphereVirtualMachineUpdate - Disks to remove: %#v", removedDisks)
+
+			for _, diskRaw := range removedDisks {
+				if disk, ok := diskRaw.(map[string]interface{}); ok {
+					log.Printf("[FINDME] disk spec to remove: %#v", disk)
+
+					devices, err := vm.Device(context.TODO())
+					if err != nil {
+						log.Printf("[ERROR] Update Remove Disk - Could not get virtual device list")
```
chrislovecnm: We going do error handling here?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -673,7 +742,13 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 				disks[i].iops = int64(v)
 			}
 			if v, ok := disk["type"].(string); ok && v != "" {
+				if v != "thin" && v != "eager_zeroed" {
+					return fmt.Errorf(
```
chrislovecnm: We might want to do this as error handling in the `schema.Resource` section.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -323,7 +323,7 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 			"disk": &schema.Schema{
 				Type:     schema.TypeList,
 				Required: true,
-				ForceNew: true,
+				ForceNew: false,
```
dagnello: forceNew: false is the default, so it not needed

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -762,30 +838,94 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
+			log.Printf("[FINDME] virtualDisk thin?: %v", *virtualDisk.Backing.(*types.VirtualDiskFlatVer2BackingInfo).ThinProvisioned)
+			log.Printf("[FINDME] vd: %#v", vd)
+
+			thin := virtualDisk.Backing.(*types.VirtualDiskFlatVer2BackingInfo).ThinProvisioned
+			if thin != nil {
+				isThin := *thin
+				if isThin {
+					disk["type"] = "thin"
+				}
+				// TODO Investigate why disks created as either thin or eager_zeroed
```
chrislovecnm: Did you want to file an issue with the govmomi peeps?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -762,30 +838,94 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
-	log.Printf("[DEBUG] %#v", mvm.Guest.Net)
+	log.Printf("[DEBUG] Datacenter - %#v", dc)
+	log.Printf("[DEBUG] mvm.Summary.Config - %#v", mvm.Summary.Config)
+	log.Printf("[DEBUG] mvm.Guest.Net - %#v", mvm.Guest.Net)
+
+	disks := make([]map[string]interface{}, 0)
+	for _, device := range mvm.Config.Hardware.Device {
+		if vd, ok := device.(*types.VirtualDisk); ok {
+
+			virtualDisk := vd.GetVirtualDevice()
```
chrislovecnm: Can this return an error?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -762,30 +838,94 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
+				// } else {
+				// 	disk["type"] = "eager_zeroed"
+				// }
+			}
+
+			disk["key"] = virtualDisk.Key
+
+			diskPath := virtualDisk.Backing.(*types.VirtualDiskFlatVer2BackingInfo).FileName
+			diskPathSplit := strings.Split(diskPath, " ")
+			if len(diskPathSplit) == 2 {
```
chrislovecnm: Do we want to document how we are doing this syntax?  Not sure if it is documented ... probably is

chrislovecnm:
Lots of comments 😄  Appreciate the work!  Do we need to tweak or improve our tests?  Also I assume that you are going to be pushing more updates to the PR.  Until that is completed please mark [WIP]

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1368,6 +1565,7 @@ func (vm *virtualMachine) createVirtualMachine(c *govmomi.Client) error {
 	for _, hd := range vm.hardDisks {
```
dagnello: In order to keep track of hard disk state that can later be used to deal with upgrade, we need to persist two pieces of information.  The first is the path to the VMDK (required to know if disk is renamed, moved, etc.) and the second being the device name/id for the disk mounted to a vm (required to detach disk from VM at a later point).

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: No error returned from GetChange()
https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L69

dkalleg:
Took a new approach to use TypeSet instead of TypeList.  This allows us a dynamically scalable list of disks that aren't dependent on position in the disk list.  Without it, removing the n-1 disk would result in the nth disk being diff'ed against the n-1 disk; this is a limitation of the TypeList schema type.  TypeSet provides a deeper check for what was added/removed.  

Bootable disks positioning is still respected.  Matching of vSphere disks against .tf script disks is done by vmdk file names since they are guaranteed by vSphere to be uniquely identifiable.  To accomplish this for disks we create on behalf of the user, we now require a unique `name` attribute for each disk.  This name becomes the filename of the vmdk within the vm's working directory.  

Templates are considered to be immutable so there is not update case for templates.  If a user creates a vm from a template and decides they want to use a different template, they'll destroy their vm and create a new one.  

@chrislovecnm @stack72 @phinze @jen20 - Please review at your earliest convenience.

FYI, during testing, a handful a tests were giving me a failure due to "vsphere_virtual_machine.bar: path '*' resolves to multiple networks", which I'm not sure is causing it.  But the new test I added for disk update is passing for me.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -443,6 +441,98 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+
+		addedDisks := newDiskSet.Difference(oldDiskSet)
+		removedDisks := oldDiskSet.Difference(newDiskSet)
+
+		// Removed disks
+		for _, diskRaw := range removedDisks.List() {
+			if disk, ok := diskRaw.(map[string]interface{}); ok {
+				devices, err := vm.Device(context.TODO())
+				if err != nil {
+					log.Printf("[ERROR] Update Remove Disk - Could not get virtual device list")
```
chrislovecnm: Do we need to bail here??  If we cannot remove a disk I think TF needs to exit.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -443,6 +441,98 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+				}
+				virtualDisk := devices.FindByKey(int32(disk["key"].(int)))
+
+				keep := false
+				if v, ok := d.GetOk("keep_on_remove"); ok {
+					keep = v.(bool)
+				}
+
+				vm.RemoveDevice(context.TODO(), keep, virtualDisk)
```
chrislovecnm: Does this call return err??

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -443,6 +441,98 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+		}
+		// Added disks
+		for _, diskRaw := range addedDisks.List() {
+			if disk, ok := diskRaw.(map[string]interface{}); ok {
+
+				var datastore *object.Datastore
+				if disk["datastore"] == "" {
+					datastore, err = finder.DefaultDatastore(context.TODO())
+					if err != nil {
+						return err
```
chrislovecnm: Log before returning 😄 please.  Also u returned here, but did not return @ 459.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -443,6 +441,98 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
...
+				}
+
+				var size int64
+				if disk["size"] == 0 {
+					size = 0
+				} else {
+					size = int64(disk["size"].(int))
+				}
+
+				iops := int64(disk["iops"].(int))
```
chrislovecnm: is `disk["iops"]` guaranteed to not be `nil` here?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -630,46 +720,84 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
...
 	if vL, ok := d.GetOk("disk"); ok {
-		disks := make([]hardDisk, len(vL.([]interface{})))
-		for i, v := range vL.([]interface{}) {
-			disk := v.(map[string]interface{})
-			if i == 0 {
+		if diskSet, ok := vL.(*schema.Set); ok {
+
+			disks := []hardDisk{}
+			hasBootableDisk := false
+			for _, value := range diskSet.List() {
```
chrislovecnm: Dumb question ... could `diskSet.List()` throw a panic if the list is empty or nil??

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -630,46 +720,84 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
...
+				disk := value.(map[string]interface{})
+				newDisk := hardDisk{}
+
 				if v, ok := disk["template"].(string); ok && v != "" {
```
chrislovecnm: Have we checked that `disk["template"]` is not nil by here??  I cannot find it in the diff.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -727,30 +855,90 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
 	collector := property.DefaultCollector(client.Client)
-	if err := collector.RetrieveOne(context.TODO(), vm.Reference(), []string{"guest", "summary", "datastore"}, &mvm); err != nil {
+	if err := collector.RetrieveOne(context.TODO(), vm.Reference(), []string{"guest", "summary", "datastore", "config"}, &mvm); err != nil {
 		return err
 	}
 
-	log.Printf("[DEBUG] %#v", dc)
-	log.Printf("[DEBUG] %#v", mvm.Summary.Config)
-	log.Printf("[DEBUG] %#v", mvm.Guest.Net)
+	log.Printf("[DEBUG] Datacenter - %#v", dc)
```
chrislovecnm: THANK-YOU ... lol  much more readable.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -727,30 +855,90 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
...
+
+			// Separate datastore and path
+			diskFullPathSplit := strings.Split(diskFullPath, " ")
+			if len(diskFullPathSplit) != 2 {
+				return fmt.Errorf("[ERROR] Failed trying to parse disk path: %v", diskFullPath)
+			}
+			diskPath := diskFullPathSplit[1]
+			// Isolate filename
+			diskNameSplit := strings.Split(diskPath, "/")
+			diskName := diskNameSplit[len(diskNameSplit)-1]
```
chrislovecnm: Do we have an edge case where `diskPath` will not have a `/` in it?  I am wondering if would should test that it has a `/` ... kinda on the fence.  Your call.

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -67,7 +67,7 @@ func TestAccVSphereVirtualMachine_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						"vsphere_virtual_machine.foo", "disk.#", "2"),
 					resource.TestCheckResourceAttr(
-						"vsphere_virtual_machine.foo", "disk.0.template", template),
+						"vsphere_virtual_machine.foo", "disk.3582676876.template", template),
```
chrislovecnm: Can we not do two disks on the basic test??  My vision is that the basic test would be 'basic' ...  one disk 😄

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -921,6 +921,130 @@ func testAccCheckVSphereVirtualMachineExists(n string, vm *virtualMachine) resou
...
+	}
+	if v := os.Getenv("VSPHERE_DATASTORE"); v != "" {
+		datastoreOpt = fmt.Sprintf("        datastore = \"%s\"\n", v)
+	}
+	template := os.Getenv("VSPHERE_TEMPLATE")
+	gateway := os.Getenv("VSPHERE_IPV4_GATEWAY")
+	label := os.Getenv("VSPHERE_NETWORK_LABEL")
+	ip_address := os.Getenv("VSPHERE_IPV4_ADDRESS")
+
+	resource.Test(t, resource.TestCase{
```
chrislovecnm: I have been thinking about this ~ a new pattern that we may need to start using as a pre-req for testing.  

A test checks which env vars are required for the test.  The test fails if the env vars are not set before the test runs.

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
```diff
@@ -921,6 +921,130 @@ func testAccCheckVSphereVirtualMachineExists(n string, vm *virtualMachine) resou
...
+					resource.TestCheckResourceAttr(
+						"vsphere_virtual_machine.foo", "vcpu", "2"),
+					resource.TestCheckResourceAttr(
+						"vsphere_virtual_machine.foo", "memory", "4096"),
+					resource.TestCheckResourceAttr(
+						"vsphere_virtual_machine.foo", "memory_reservation", "4096"),
+					resource.TestCheckResourceAttr(
+						"vsphere_virtual_machine.foo", "disk.#", "2"),
+					resource.TestCheckResourceAttr(
+						"vsphere_virtual_machine.foo", "disk.3582676876.template", template),
```
chrislovecnm: Do I need any setup for this?? `disk.3582676876.template`  If so make function notes on how to setup the template

website/source/docs/providers/vsphere/r/virtual_machine.html.markdown
```diff
@@ -113,6 +113,7 @@ The `disk` block supports:
 * `template` - (Required if size and bootable_vmdk_path not provided) Template for this disk.
 * `datastore` - (Optional) Datastore for this disk
 * `size` - (Required if template and bootable_vmdks_path not provided) Size of this disk (in GB).
+* `name` - (Required if size is provided) Name of the new disk file to create.
```
chrislovecnm: Can we just generate a name if they don't give us one?  Or document what the name means :)

chrislovecnm:
@dkalleg nice work btw ... The effort and time is GREATLY appreciated!!!!

@markpeek @stack72 @jen20  ... can I get some comments from you golang gurus?

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: Good call, updating this.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: Yup, updating this.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: I believe so.  This is grabbing disk info from the schema, which will always pad with zero-values if the user doesn't not provide a value.  Meaning, for iops, we'd get a 0, not a nil.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: Since the disk schema attribute is set to Required: true, we shouldn't be able to get to that scenario; however, looking at the set.List() implementation, it looks safe for empty sets.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: Similar to my other comment, this is grabbed from the schema which will zero-value out each attribute, so we would get an empty string here instead of a nil.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: If we split on "/" and there is no "/", we'll get back a slice of len 1.  Indexing by len(diskNameSplit)-1 should always return the last index, which is safe in any scenario.

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
dkalleg: Maybe thats better suited for a different PR since design of the basic test is outside the scope of this PR.  
FYI:  The changes you see here where we reference "disk.1234123423" is the hash of the Set value.  Looking at other providers tests, I've realized the set hash value is reproducible given that your disk declaration is constant, which it is in the tests.  So I had to update each test that uses a disk to use a hash instead of an array index (ex. "disk.0")

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
dkalleg: Could yield more readable error messages.  Current implementation would fail if some env vars are not specified, which may or may not be difficult to debug.  This also may be good for a separate PR to update all the tests at once to follow this, in order to give uniformity.

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
dkalleg: Only setup required for this test is the same setup as other tests that require a template.  The long number string your seeing is the set hash value.  Since the disk schema uses a set instead of a list now, it is indexed by these hash values instead of incrementing digits starting from 0.  These hash values should be static across multiple runs of this test.  The hash would only change if the disk declaration at the bottom of this file is changed.

builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go
chrislovecnm: key word

> some

website/source/docs/providers/vsphere/r/virtual_machine.html.markdown
dkalleg: We could come up with a random string.  I would think providing a name would give good readability in the vSphere environment when the time comes to clean out old disks.  If we auto generate a name, we would have to do checks to ensure it is unique among other (randomly named) disks on that vm.  Would have to read down each disk each time before creating a new disk to ensure no duplicates, or possibly store it in a calculated attribute in the disk schema.

chrislovecnm: how about this?

`name` - (Required if size is provided and creating a new disk) This "name" is used as the disk name in vSphere, when the new disk is created.

dkalleg: Sounds good to me! I'll update this. I was operating on the assumption that size is only specified when we are creating a disk for the user.  Admittedly, this is not obvious to users so explicitly calling it out in the docs is smart.

dkalleg:
Chris' comments addressed.  Still looking into name auto-generation.  Ready for more reviews.

chrislovecnm:
@dkalleg "The command "make test vet" exited with 2."

dkalleg:
Whoopsie.  New patch inbound.

dkalleg:
Name auto-generation looks to be non-trivial.  Having the value in the schema allows us something to compare against in the read() phase.  We would have to do a d.Set in the create() phase on the disks set in order to write the generated disk name, which doesn't seem like Terraform standard practice.  Someone can correct me on this. 

Hoping to generate more discussion around this before investing much time.

chrislovecnm:
@dkalleg the doc update is fine.  Don't worry about name gen.

chrislovecnm:
@stack72 @jen20 LGTM ... if we could get a quick code review and if nothing jumps out at you please merge.

dkalleg:
FYI, It would be a valid argument to say the template attribute of a disk should be a ForceNew, since if the user wants to change the template, it is nearly equivalent to creating an entirely new vm.  In testing when I set template for ForceNew: true, I get an unexpected result.

Any time I modify a disk and expect the Update() functionality to take over, a `terraform plan` will show that the ForceNew of disks not being deleted (disks being added or disks not being modified) are triggering the ForceNew:

```
-/+ vsphere_virtual_machine.dansTerraformVm
    cdrom.#:                                "1" => "1"
    cdrom.0.datastore:                      "local" => "local"
    cdrom.0.path:                           "dan/cidataDan.iso" => "dan/cidataDan.iso"
    datacenter:                             "Datacenter" => "Datacenter"
    disk.#:                                 "2" => "2"
    disk.1833630372.bootable:               "false" => "false"
    disk.1833630372.datastore:              "" => ""
    disk.1833630372.iops:                   "1002" => "0"
    disk.1833630372.keep_on_remove:         "false" => "false"
    disk.1833630372.name:                   "three" => ""
    disk.1833630372.size:                   "2" => "0"
    disk.1833630372.template:               "" => ""
    disk.1833630372.type:                   "eager_zeroed" => ""
    disk.1833630372.vmdk:                   "" => ""
    disk.3360429507.bootable:               "" => ""
    disk.3360429507.datastore:              "" => ""
    disk.3360429507.iops:                   "" => ""
    disk.3360429507.keep_on_remove:         "" => ""
    disk.3360429507.key:                    "" => "<computed>"
    disk.3360429507.name:                   "" => "two"
    disk.3360429507.size:                   "" => "2"
    disk.3360429507.template:               "" => "" (forces new resource)
    disk.3360429507.type:                   "" => "eager_zeroed"
    disk.3360429507.uuid:                   "" => "<computed>"
    disk.3360429507.vmdk:                   "" => ""
    disk.433444201.bootable:                "true" => "true"
    disk.433444201.datastore:               "" => ""
    disk.433444201.iops:                    "0" => ""
    disk.433444201.keep_on_remove:          "false" => ""
    disk.433444201.key:                     "0" => "<computed>"
    disk.433444201.name:                    "" => ""
    disk.433444201.size:                    "0" => ""
    disk.433444201.template:                "DansTfTest/ubuntuTemplate" => "DansTfTest/ubuntuTemplate" (forces new resource)
    disk.433444201.type:                    "eager_zeroed" => "eager_zeroed"
    disk.433444201.uuid:                    "" => "<computed>"
    disk.433444201.vmdk:                    "" => ""
```
In this experiment, I boot a vm with 2 disks (1 template, 1 new disk), then I modify my .tf to remove the new disk and add a different new disk.  The *unmodified* template disk triggers the ForceNew as well as the new disk to be added.  

Because of this, I'm leaving ForceNew out of all attributes.

chrislovecnm:
@dkalleg thanks for catching that.  We ready for a merge?

cc: @stack72 @catsby @jen20 @phinze

chrislovecnm:
@dkalleg we good for a merge??

chrislovecnm:
@dkalleg can you rebase this and then let everyone know that it is ready for a merge?

dkalleg:
@chrislovecnm You bet.  Shooting for some time today.

dkalleg:
@chrislovecnm @stack72 @catsby @jen20 @phinze 
Rebased and ready for review

chrislovecnm:
LGTM can we get a merge?

chrislovecnm:
@stack72 thanks!!!

Issue #6626, opened 2016-05-12
Feature: Ability to enable Amazon S3 bucket acceleration


premist:
AWS [announced](https://aws.amazon.com/blogs/aws/aws-storage-update-amazon-s3-transfer-acceleration-larger-snowballs-in-more-regions/) new feature called **Amazon S3 Transfer Acceleration** which you can use to accelerate transfer speed for cross-country transfer of large objects.

You can enable this feature by using [PUT Bucket accelerate](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTaccelerate.html) endpoint via AWS API. 

We're performing this action manually after our initial `terraform apply`, but it would be really handy if Terraform supports enabling bucket for acceleration.

You can read more about this feature on Amazon S3 Documentation: [Amazon S3 Transfer Acceleration](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html)

stack72:
Hi @premist 

PR has just been opened to add this feature :)

Paul

premist:
Hi @stack72, that was fast! Awesome :)

catsby:
This was added in #6628

PR #6628, opened 2016-05-12
provider/aws: Add support for S3 Bucket Acceleration

stack72:
As requested in #6626, this adds support for S3 Bucket Transfer Accelection

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSS3Bucket_basic' 2>~/tf.log
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSS3Bucket_ -timeout 120m
=== RUN   TestAccAWSS3Bucket_basic
--- PASS: TestAccAWSS3Bucket_basic (47.14s)
=== RUN   TestAccAWSS3Bucket_acceleration
--- PASS: TestAccAWSS3Bucket_acceleration (100.95s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	101.941s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket.go
@@ -293,6 +293,13 @@ func resourceAwsS3Bucket() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+
+			"acceleration_status": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateS3BucketAccelerationStatus,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_s3_bucket.go
```diff
@@ -1290,6 +1334,19 @@ func normalizeRegion(region string) string {
 	return region
 }
 
+func validateS3BucketAccelerationStatus(v interface{}, k string) (ws []string, errors []error) {
+	value := strings.ToLower(v.(string))
+	validTypes := map[string]struct{}{
+		"Enabled":   struct{}{},
+		"Suspended": struct{}{},
```
catsby: `TestAccAWSS3Bucket_acceleration` fails for me, because we lowercase the value here but the `validTypes` are capitalized... odd

stack72: Whoops! I need to admit that I tried to be really clever here and add a unit test after I created the PR that would check for the types :(

Good spot!!!

Fixed now

catsby:
LGTM thanks!

amivit:
How does this work - not an option but always on? Shouldn't it be a setting since it costs extra?

PR #6635, opened 2016-05-12
provider/vsphere: fix gateway settings in dhcp tests 

thetuxkeeper:
Fix for DHCP tests after merging #6522.
cc @chrislovecnm , @jen20

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -290,7 +290,7 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 						"ipv4_gateway": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
-							ForceNew: true,
+							Computed: true,
 						},
 
 						"ipv6_address": &schema.Schema{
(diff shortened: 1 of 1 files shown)


stack72:
LGTM! pending travisci

thetuxkeeper:
@stack72: Wow, that was a fast response :D 
Btw. Is it really the right way to set it only as `Computed`? I mean if you set this setting (not computed) it won't be updated. But `ForceNew` is somehow also wrong since it would redeploy if you change the DHCP gateway not set by terraform.

stack72:
if it can be set by the user and can also be computed then computed and optional is correct

thetuxkeeper:
Ok, thanks. Then we would have to implement the update logic that it works like a user would expect. But that's already on the roadmap.

stack72:
@thetuxkeeper for now, until we get the update logic in place, should these be `Optional`, `Computed` (to fix the bug) and `ForceNew`?

thetuxkeeper:
With `ForceNew`, would it recreate the VM if you use DHCP (`*_gateway` not set and computed) and change the gateway with DHCP?
If it's like that we have two choices while the update logic is not there yet:
1. unnecessary recreation when changing the gateway over DHCP (`ForceNew`)
2. or not applying the setting if you set it in the config

But I think an unnecessary recreation is better than not applying changed. But I'm not sure.

If we change it for the gateway settings to `ForceNew`+`Computed`, should we change it for the other network settings? The IP address settings should have the same behavior since only `Computed` is set

stack72:
Got it - let's (IMO) go with the least dangerous one - the not updating (as you have it) and then revisit the update later

thetuxkeeper:
Ok. Then I know how to set it if something like that appears somewhere else.
Thanks for merging :)

PR #6643, opened 2016-05-12
state/remote tests: bump timeout to work around failing test [maint-0.6]

phinze:
It seems that the 50ms timeout has become too tight for this assertion.

Ideally we'd rework the test to not be time based, but that is a battle
for another day! :)


josephholsten:
This is already in master, though it's got an incorrect error message: https://github.com/hashicorp/terraform/blob/master/state/remote/atlas_test.go#L162-L164

phinze:
Yep - this targets the `maint-0.6` branch to get that test passing on the 0.6 series. Good catch on the message in master though!

PR #6644, opened 2016-05-12
Terraform 0.6.17 - state upgrade/downgrade changes

jen20:
Opened primarily to get reliable CI.

cc @phinze.

### DIFF
--- a/builtin/providers/terraform/resource_state.go
+++ b/builtin/providers/terraform/resource_state.go
@@ -62,7 +62,11 @@ func resourceRemoteStateRead(d *schema.ResourceData, meta interface{}) error {
 
 	var outputs map[string]string
 	if !state.State().Empty() {
-		outputs = state.State().RootModule().Outputs
+		outputValueMap := make(map[string]string)
+		for key, output := range state.State().RootModule().Outputs {
+			//This is ok for 0.6.17 as outputs will have been strings
+			outputValueMap[key] = output.Value.(string)
+		}
 	}
 
 	d.SetId(time.Now().UTC().String())
--- a/builtin/providers/terraform/resource_state.go
+++ b/builtin/providers/terraform/resource_state.go
@@ -62,10 +62,9 @@ func resourceRemoteStateRead(d *schema.ResourceData, meta interface{}) error {
 
 	var outputs map[string]string
 	if !state.State().Empty() {
-		outputValueMap := make(map[string]string)
 		for key, output := range state.State().RootModule().Outputs {
 			//This is ok for 0.6.17 as outputs will have been strings
-			outputValueMap[key] = output.Value.(string)
+			outputs[key] = output.Value.(string)
 		}
 	}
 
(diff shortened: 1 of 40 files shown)


phinze:
(rebase over the travis.yml update in `maint-0.6` to get CI running)

builtin/providers/terraform/resource_state.go
```diff
@@ -62,7 +62,11 @@ func resourceRemoteStateRead(d *schema.ResourceData, meta interface{}) error {
 
 	var outputs map[string]string
 	if !state.State().Empty() {
-		outputs = state.State().RootModule().Outputs
+		outputValueMap := make(map[string]string)
+		for key, output := range state.State().RootModule().Outputs {
+			//This is ok for 0.6.17 as outputs will have been strings
+			outputValueMap[key] = output.Value.(string)
+		}
```
phinze: outputValueMap looks like it's thrown away here as a local variable?

I think this should just be populating `outputs`, the var from the higher level scope.

jen20: Good point - this needs better test coverage as well.

command/output.go
```diff
@@ -98,7 +98,13 @@ func (c *OutputCommand) Run(args []string) int {
 		return 1
 	}
 
-	c.Ui.Output(v)
+	switch v.Type {
+	case "string":
+		c.Ui.Output(v.Value.(string))
+	default:
+		panic(fmt.Errorf("Unknown output type: %T", v.Value.(string)))
```
phinze: I'd just `c.Ui.Error` and `return 1` here, but no big deal either way - this shouldn't be hit in normal operating conditions.

terraform/state.go
```diff
@@ -6,30 +6,35 @@ import (
...
 // State keeps track of a snapshot state-of-the-world that Terraform
 // can use to keep track of what real world resources it is actually
-// managing. This is the latest format as of Terraform 0.3
+// managing. This is the latest format as of Terraform 0.6.17.
 type State struct {
-	// Version is the protocol version. Currently only "1".
+	// Version is the protocol version.
 	Version int `json:"version"`
 
+	// ReadVersion is the version of the state that was originall read.
```
phinze: s/originall/originally/

terraform/state_v1_test.go
```diff
@@ -1,118 +1,147 @@
...
-	"encoding/gob"
-	"errors"
-	"io"
 	"reflect"
-	"sync"
+	"strings"
 	"testing"
 
-	"github.com/mitchellh/hashstructure"
+	"github.com/davecgh/go-spew/spew"
```
phinze: Stray spew here - don't think we have this vendored now.

phinze:
This came out nice! LGTM pending travis. A few optional nits inline.

terraform/state_v1_test.go
jen20: This isn't stray - we use it to output usable test expected/actual in the case of failure (e.g. L96)

jen20: Just checked - spew is still vendored on the `maint-0.6` branch so we should be :+1: here.

phinze:
LGTM!

PR #4241, opened 2015-12-09
Make the OpsWorks cookbooks SSH key write only

b-ryan:
Fixes #3635

This follows the suggestion of @apparentlymart in
https://github.com/hashicorp/terraform/issues/3635#issuecomment-151000068
to fix the issue of OpsWorks stacks always complaining about the custom
cookbooks SSH key needing to be changed.

Functional tests:

* Created a new stack and gave it an SSH key. The key was written to
  OpsWorks properly.
* Ran "plan" again and terraform indicated it needed to change the SSH
  key, which is expected since terraform cannot read what the existing
  SSH is.
* Removed the key from my resource and this time, "plan" did not have
  any changes. The `tfstate` file indicated the SSH key was "" (empty
  string).
* Changed an unrelated property of the stack. Previously this was not
  working for me due to terraform attempting to change the SSH key.

Issue #6648, opened 2016-05-12
AWS OpsWorks ssh_key in app_source keeps re-applying


iroller:
Hi there,

It looks like there's the same issue as https://github.com/hashicorp/terraform/issues/3635 existing in aws_opsworks_application -> app_source -> ssh_key.

Config:

``` hcl
resource "aws_opsworks_application" "app" {
  name        = "application"

  app_source = {
    type     = "git"
...
    ssh_key  = "${file("~/.ssh/private-key")}"
  }
```

On terraform plan/apply it keeps re-applying the ssh_key even though it's the same.

Also if it's filtered it shouldn't be displayed:

```
$ terraform plan

~ aws_opsworks_application.app
    app_source.0.ssh_key:  "*****FILTERED*****" => "-----BEGIN RSA PRIVATE KEY--- ... my-private-key-here"
```
### Terraform Version

v0.6.15
### Affected Resource(s)
- aws_opsworks_application
- app_source

stack72:
Closed by #6649

iroller:
Sad but the pull request wasn't complete. There's no support for "WriteOnly" attributes for aws_opsworks_application. It's only supported in aws_opsworks_layers.

To get that bug actually fixed we'll need to add "WriteOnly" to the main schema or there should be a separate struct, e.g. opsworksApplicationAttribute for "app_source" just like it's done for aws_opsworks_layers's opsworksLayerTypeAttribute.

wookieb:
Ping. Could you please open the issue? I'm using terraform 0.6.16 and got the same issue.

u2mejc:
@wookieb  There are a few of these immutable resources. As a stop gap, I've avoided these issues by simply using the lifecycle ignore_changes feature, like this: 

```
lifecycle {
    ignore_changes = ["custom_cookbooks_source.0.ssh_key"]
  }
```

solin86:
The lifecycle ignore_changes workaround @u2mejc proposed doesn't work for us.

Whenever another attribute of the resource (`aws_opsworks_stack` in our case) changes, `custom_cookbooks_source.0.ssh_key` will be set to empty string in OpsWorks, effectively breaking the whole thing for us as I can't see any kind of workaround.

If interested, I have a very minimal `stack.tf` that I could enhance with comments on how to reproduce. Using `0.6.16`.

PR #6649, opened 2016-05-12
Make the OpsWorks app_source SSH key write only

iroller:
Similarly to https://github.com/hashicorp/terraform/pull/4241 prevent ssh_key in OpsWorks application app_source key from re-applying every time.


iroller:
Related to https://github.com/hashicorp/terraform/issues/6648

stack72:
Hi @iroller 

Thanks so much for the PR here - @apparentlymart is the man to give this the once over :)

Paul

apparentlymart:
Oh my, another one of these... :disappointed: 

Thanks!

PR #6650, opened 2016-05-12
master: atlas: update test err msg to reflect real timeout

josephholsten:


stack72:
LGTM! When travis goes green, merge away @josephholsten :)

Issue #6416, opened 2016-04-29
Security group --> vpc_security_group_ids change bug


boxrick:
### Terraform Version
0.6.15 ( Works in 0.6.14 )

### Affected Resource(s)
- aws_instance

### Terraform Configuration Files

 security_groups 
 vpc_security_group_ids 

### Info

In Terraform 0.6.14 security_groups was valid from custom VPC, in 0.6.15 it takes it as a value and works but then upon a refresh it tries to re-create the resource.

This means from upgrading to 0.6.14 to 0.6.15 the it re-creates these resources each time:

    security_groups.#:          "0" => "2" (forces new resource)
    security_groups.3326659650: "" => "sg-6b56000f" (forces new resource)
    security_groups.3522483665: "" => "sg-74560010" (forces new resource)

Even though the resource has these 2 security groups attached and working. This either needs documenting or some more verbose error to say that this security group is not valid.

stumyp:
Have the same issue, forced to revert back to 0.6.14, not fixed in current master either.

apparentlymart:
Sorry for the trouble, folks!

The referenced issue #6369 has a workaround for this, which appears to allow you to get around the problem by adjusting the config to match what Terraform is reading back from AWS.

It sounded like @catsby wanted to use this ticket to investigate this and either implement a fix or document the workaround, but hopefully that workaround is useful in the mean time.

stumyp:
thanks @apparentlymart , that worked.
I think it is safe to close this one :+1:

apparentlymart:
In #6585 @radeksimko is retroactively adding a note to the changelog about this.

@catsby, did you want to add something to the main `aws_instance` docs about this, or perhaps change the provider implementation to give better feedback for this case?

apparentlymart:
(A potential fix for this was merged in #5193 but then reverted because it broke for EC2-Classic users.)

apparentlymart:
I have closed several other reports of this same issue to consolidate discussion here. Let's not close this one until we're sure we've got a handle on the plan for this problem.

catsby:
Hey friends – we identified the regression that's causing this to start happening, it was introduced in v0.6.15. I have a pull request open here https://github.com/hashicorp/terraform/pull/6664 that will revert the regression.

[...]

PR #6664, opened 2016-05-13
provider/aws: Revert security group / vpc security group depreciation 

catsby:
Reverts a depreciation that was  added where we correctly save/update `security_groups` and `vpc_security_groups` and no longer allows them to be interchangeable. 

As discussed we'll likely re-enable this in v0.7.0 with explicit notes in the CHANGELOG and likely further validation in the resource.


phinze:
LGTM

catsby:
Fixes #6416 and others referenced therein

lime:
Will there be a version 0.6.17 with this fix included? If I understood correctly, this is still unreleased and the regression is present in the two latest 0.6 versions.

PR #6665, opened 2016-05-13
command: Add remote state capability to output subcommand

vancluever:
**UPDATE:** Moved to `terraform remote output` - see comments.

This patch provides the ability to view **separate** remote state via the `terraform output` sub-command.

Some use cases:

 * You need output variables out-of-band from Terraform, ie: you want to pull the outputs without needing the whole repo or config, or having to run `terraform remote config` first.
 * You want to use `terraform_remote_state`, but you need to provide the outputs to variables/parameters that ultimately influence resource `count`s (my use case). In this instance, you can use this to view the outputs of a separate remote state without affecting your current locally configured remote state.

Command-line usage is similar to `terraform remote config`, but with some changes to make things semantic for the specific use context.

Also, I've updated the docs and the help a little bit for the `-module` flag, which only works when you are specifying a certain output to display, due to the fact that `command.outputsAsString()` only acts on the root module.

I was going to make this a separate project (https://github.com/paybyphone/terraform-remote-env) but that kind of seemed silly. ;)

### DIFF
--- a/command/apply.go
+++ b/command/apply.go
@@ -4,12 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"sort"
 	"strings"
 
 	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/go-multierror"
-	"github.com/hashicorp/terraform/config"
 	"github.com/hashicorp/terraform/terraform"
 )
 
@@ -250,8 +248,8 @@ func (c *ApplyCommand) Run(args []string) int {
 			c.Meta.StateOutPath())))
 	}
 
-	if !c.Destroy {
-		if outputs := outputsAsString(state, ctx.Module().Config().Outputs, true); outputs != "" {
+	if !c.Destroy && state != nil {
+		if outputs := allOutputsAsString(state.RootModule(), ctx.Module().Config().Outputs, true); outputs != "" {
 			c.Ui.Output(c.Colorize().Color(outputs))
 		}
 	}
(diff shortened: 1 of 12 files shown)


apparentlymart:
Interesting idea! (Of course, your second use-case should be eliminated by being able to interpolate data resource attributes into count, but still!)

I wonder if this would be more intuitive as a separate command `terraform remote output`... currently all of the remote-related stuff lives under that prefix, and then it could have exactly the same argument names as `terraform remote config` and not have confusing conflicts between arguments (`-state` vs. `-remote=true`).

Don't feel that strongly about it, though. What do you think?

vancluever:
@apparentlymart, thanks! Yeah this idea mainly came up as a stop-gap idea for me while #4169 is fully hashed out, but I figured there might be value for remote state being readable from Terraform for out-of-band tasks as well.

I was torn too on where exactly in the command tree to put this, but settled on `output` because it currently seems that `remote` is more oriented around managing/configuring remote state. One thing that I do see now again after reviewing and mulling it is that a lot of the logic for fetching and display the outputs seems to be hardcoded within the `command.OutputCommand.Run()` function. Mind you, we could probably move some of that logic to some common functions and just use that.

Let me see what I can do here. :)

vancluever:
Alright, this is now done. The functionality has been moved to `remote output`, and the options changed so that they completely reflect that of `remote config`, with the appropriate bits from the plain `output` command added in.

In this process I also did a bit of a re-factoring on the `output` command's functionality so that it could be re-used with `remote output`, **and** I fixed it so that all outputs could be dumped when `-module` is specified, versus the old behaviour of just silently dumping the outputs for the root module.

vancluever:
Just a FYI I'm about to close this. There has been some changes and divergences to the way that outputs are handled it looks like (work that @jen20 has been doing to ensure that list and map outputs are properly formatted and handled) which has added some complexity to keeping this up to date. Further to that, I have found that I can replace a good chunk of this functionality with a simple `jq` filter on the remote state's JSON, so the usefulness of having this sub-command has significantly diminished for me.

PR #4848, opened 2016-01-27
New resource - aws_availability_zones

vancluever:
This adds a new resource - `aws_availability_zones` - to get a list of availability zones that the account/region combination can use.

Precedent: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getavailabilityzones.html

I've exposed the API parameters (docs in the PR).

One thing - the test is general, as depending on where you are set up, doing something like limiting your region to `us-east-1`, etc, will give no results. I've tested locally here though with the options and had success.

Also - `buildEc2Filters` is similar to the function I have in #4396, which I will move to use this if/when things are merged.

PR #6671, opened 2016-05-14
[WIP]: provider/aws: Add aws_availability_zones data source

jen20:
This PR adds a data source with a single set, `available` for the schema which gets populated with the availability zones to which an account has access.

I implemented it as part of the review process of data sources so thought I would open an additional PR for feedback from @apparentlymart about this implementation.

One area we need to consider is nomenclature (cc @phinze, @catsby, @mitchellh) - many of these sources will return lists, and we should take care to keep naming arity consistent - I am expecting the actual names here to change when we make that decision.

TODO: Acceptance test.


vancluever:
@jen20 you might want to check #4848 as well, that one sorts the AZs (to make sure the order is predictable) and there's an acceptance test there too.

apparentlymart:
@jen20 my thought was that to start we'd mainly focus on data sources that return single objects... so `aws_ami` returns a single AMI matching given criteria, etc... because right now you can't really do much with attributes that are lists of resources in Terraform anyway.

However, this one I think makes sense because ultimately it's just a list of strings, which Terraform is pretty good at dealing with. And so calling it `aws_availability_zones` makes sense.

I was planning to build out `aws_ami` (using @vancluever's earlier work as a starting point) as an initial non-trivial data source example, establishing a pattern for how AWS `Describe*` actions can map on to data sources... but I was planning to wait until #6598 lands before building that, in case we make any changes to the interface before it is merged.

website/source/docs/providers/aws/d/availability_zones.html.markdown
```diff
@@ -0,0 +1,39 @@
+---
+layout: "aws"
+page_title: "AWS: aws_availability_zones"
+sidebar_current: "docs-aws-datasource-availability-zones"
+description: |-
+    Provides a list of availability zones which can be used by an AWS account
+---
+
+# aws\_availability_zones
```
apparentlymart: Does this second underscore also need to be escaped, or will Markdown ignore it because there's an odd number of them? (In other files I've seen all the underscores escaped and have been following that pattern, but not sure if that's actually necessary...)

builtin/providers/aws/data_source_availability_zones.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsAvailabilityZones() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsAvailabilityZonesRead,
+
+		Schema: map[string]*schema.Schema{
+			"available": &schema.Schema{
+				Type:     schema.TypeSet,
```
apparentlymart: I think some folks were interested in having these in a predictable order so that they can do stuff like pluck the first item from the list and use it as a "primary AZ" in configuration... so might be better to make this a `TypeList` and sort it alphabetically in the implementation so that the ordering is dependable.

apparentlymart:
Looks good to me, minor inline niggles aside. :trumpet:

builtin/providers/aws/data_source_availability_zones.go
```diff
@@ -0,0 +1,50 @@
...
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
```
apparentlymart: Certainly doesn't have to be in the first version of this data source, but I was thinking that when mapping `Describe*` calls onto data sources we'd map each of the filters onto an `Optional` attribute and then include them in the call.

In this case, that could be attributes `regions`, which would be a list of regions to constrain to, and `states`, allowing to filter by state.

So then I'd use it something like this:

```hcl
data "aws_availability_zones" "us" {
    regions = ["us-west-1", "us-west-2", "us-east-1"]
    states = ["available", "information", "impaired"] # don't include "unavailable" AZs
}
```

builtin/providers/aws/data_source_availability_zones.go
jen20: That is a good plan I think.

builtin/providers/aws/data_source_availability_zones.go
jen20: @apparentlymart I'm not sure that I see the value in _this_ case, since the configured provider will determine which region is valid for use? In general though, I agree with this.

website/source/docs/providers/aws/d/availability_zones.html.markdown
jen20: Probably. I hate markdown ;-)

jen20:
I have no idea why the diff is so screwed up here - I thought I had rebased onto origin/master. I'll close this one and open a new PR for this soon.

PR #6672, opened 2016-05-14
Logical Resources for Random Values

apparentlymart:
This PR introduces a new provider called "random" that provides a generic way for Terraform to work with random values while avoiding "perma-diffs", by modelling the random values as resources themselves and thus allowing the random values to be retained between runs until configuration directs them to change.

This provider starts off with two resources that are each directed at a different use-case:

* `random_id` allows the generation of unique ids in a generic way, primarily for use in adding uniqueness to the names of resources that have `create_before_destroy` set but whose implementations do not have built-in support for such "uniqing".
* `random_shuffle` allows the generation of random permutations of lists, primarily for shuffling the results of data sources like `aws_availability_zones` (see #6671) to distribute resources "randomly" within the available availability zones. This is intended to avoid the need to build randomness into multiple data sources individually.

There's more exposition and examples of this in the documentation updates that you can find in the diff of this PR.


jen20:
@apparentlymart I am pro on this idea - one question: should it be a data source in the new world?

apparentlymart:
@jen20 I think these along with most of the resources in the `tls` provider are interesting examples where the idea of a "logical resource" still makes sense: although the random resources exist entirely within Terraform's logic, they still have a create/delete lifecycle and retain state.

If they were data sources they would generate a fresh random number on each run, which would then fail to address the requirement that the values remain stable most of the time.

jen20:
:+1: agree with your thinking there. This LGTM!

PR #6674, opened 2016-05-15
provider/openstack: Remove Floating IP Support From Instances

jtopjian:
Inspired by @Fodoj's request to remove the top-level `floating_ip` instance attribute, I wanted to explore entirely removing Floating IPs from instances.

Management of Floating IPs from within instances has to account for several different use-cases which becomes very complicated over time. As can be seen from this patch, removing Floating IPs gets rid of quite a bit of code from the instance resource -- including the increasingly complicated Access IP logic.

Moving the association and disassociation of Floating IPs to the Floating IP resources keeps those actions in more of a natural area of Terraform. 

In addition, this solves issus such as #4118 for these two specific resources.

There are downsides, though:

1. This is a breaking change.
2. The instance will never know of the Floating IP. This kind of matches normal Floating IP behavior, since the operating system within the instance never knows about the Floating IP either -- just its local IPs.
3. All connection configuration must now be configured to use the Floating IP resource `openstack_compute_floatingip_v2.myip.address`.

I'm not 100% sure this is a good idea yet. This WIP PR is more of an exploration as well as polling others for input.

At the moment, this PR only accounts for `openstack_compute_floatingip_v2`. I will work on `openstack_networking_floatingip_v2` next.

Edit: `openstack_networking_floatingip_v2` is done.

Input on this PR is greatly appreciated.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_floatingip_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_floatingip_v2.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/floatingip"
 )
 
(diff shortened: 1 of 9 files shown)


jtopjian:
After exploring this for a few days, I think the current implementation is a better choice. I'm closing this PR, but it'll be in the history if anyone discovers it in the future and might have some input on.

PR #6682, opened 2016-05-16
Fixes #6076 - Adjusts check to allow for instance-id reset on aws_route

### LINKED ISSUE #6076: aws_route Error: more than 1 target specified.
#5867 and/or #5321 appear to have broken aws_route when the underlying instance_id is recreated/tainted. I'm unsure what is causing it but the route table entry in AWS still exists and points to a dandling eni (inactive black hole). Terraform then proceeds to apply the aws_route to the instance_id, but aws errors due to only target can be specified.

```
* aws_route.public_to_openvpn_clients: Error: more than 1 target specified. Only 1 of gateway_idnat_gateway_id, instance_id, network_interface_id, route_table_id orvpc_peering_connection_id is allowed.
```

Sample config:
```
resource "aws_route" "public_to_openvpn_clients" {
    route_table_id = "${aws_route_table.public.id}"
    destination_cidr_block = "${var.openvpn_cidr}"
    instance_id = "${aws_instance.openvpn_server.id}"
}
resource "aws_instance" "openvpn_server" {
 #...
}
resource "aws_route_table" "public" {
 # ...
}
```
To reproduce:
* terraform apply
* terraform taint aws_instance.openvpn_server
* terraform apply

jrnt30:
fixes #6076 

As @atward described, when using the `instance_id` attribute for the association, in the background AWS is also assigning the `network_interface_id`.  The current way of checking would essentially eliminate any possibility of reallocating the route to a new instance id due to the `numTargets` check always returning 2 in that case and fast failing. 

- Update order of precedence to ensure that `allowedTargets` is evaluated so that `instance_id` comes after `network_interface_id` to allow for check 
- If there are multiple but they are `instance_id` and `network_interface_id` is the other let AWS handle the `network_interface_id`

NOTE: One thing I would have liked to add a test for but had a family emergency come up is tainting the original instance and reapplying. I did not see any direct example of doing this and was considering simply standing up 2 instances in the test and running a re-apply to associate the 2nd instance.  If I get a chance before this is merged I will try and do so however I do not have cycles immediately.


atward:
:+1: Tested in #6076

atward:
@jrnt30 close #6076 if this is fixed.

Issue #5975, opened 2016-04-01
Add nsone (DNS) provider


sarguru:
It would be nice to add the terraform provider for nsone (https://github.com/bobtfish/terraform-provider-nsone) into the core built-ins to increase the number of built-in providers.

Would be great if any of the maintainers would have a look at the code and have any suggestions on if/how-to add the provider to the list of built-ins.

cc @bobtfish

bobtfish:
cc @phinze we discussed this a while back and you suggested this could/should be a thing that you'd like to see go into core. @sarguru++ has done a bunch of fixing it up and adding tests etc so that it's suitable for inclusion.

There aren't really any docs yet, deliberately, to save us the effort of having to transcribe them when we move stuff to core - we *will* be adding docs with the core PR if you give us a :thumbsup: that this looks suitable and we should go ahead with starting the work of renaming / moving it all.

Thoughts?
TIA
<3

sarguru:
Just a gentle bump, to see if there is any update . TIA :)

bobtfish:
ping @mitchellh @phinze @radeksimko - what's missing here for us to join the party?

bobtfish:
ping @mitchellh  @phinze @radeksimko @jen20 @stack72  - can one of you pitch in here?

josephholsten:
@bobtfish honestly, let's just do this. Do you need any help throwing together a PR?

I'll take a look this week to see if there's any obvious improvements/simplifications we can make. But if you all are using it for production work, it's probably solid enough to move forward with.

bobtfish:
@sarguru lets go ahead and PR? I'll let you drive this, yell at me if you want review / doc review before you submit :)

josephholsten:
@bobtfish how's #6683 look to you?

josephholsten:
ok, we're treating #6744 as the active branch

stack72:
Closing in favour of the (super close to being complete) #10172

PR #6683, opened 2016-05-16
Import NS1 Provider

josephholsten:
First attempt at #5975, importing the entire history. 

This was accomplished with:

    git subtree add -P nsone https://github.com/josephholsten/terraform-provider-nsone.git move-upstream


josephholsten:
deprecated in favor of #6692

Issue #6466, opened 2016-05-03
aws_elasticache_parameter_group crash when change exists outside of tfstate


paulruiz:
Panic crash in _aws.flattenElastiCacheParameters_ when running terraform apply, plan or destroy with existing _aws_elasticache_parameter_group_ where change to a parameter is made outside of terraform.  Verified reproducible with terraform -v "0.6.15" and "0.6.14"

To reproduce
* terraform apply the aws_elasticache_parameter_group.tf included in the gist below
* change the timeout parameter value in AWS console or by awscli
```
aws elasticache modify-cache-parameter-group --cache-parameter-group-nam bug-repro --parameter-name-values ParameterName=timeout,ParameterValue=0
```
* terraform plan
* crash :(

aws_elasticache_parameter_group.tf, crash.log and terraform.tfstate are included in the gist
https://gist.github.com/paulruiz/06878773487ed51b1923dc600ff691f2

paulruiz:
Sorry.  Actually the crash only happens if you make the change in AWS console.  Updating the parameter via awscli produces the expected diff from plan.

paulruiz:
Further testing in AWS console revealed that no parameter change is necessary.  A crash can be triggered by just selecting the parameter group, clicking "Edit Parameters", then clicking "Save Changes", without making any changes to parameters.  Then run terraform plan.

catsby:
Hey @paulruiz thanks for opening this – I've sent a patch in https://github.com/hashicorp/terraform/issues/6687 to only save attributes with valid values that come with them. Turns out there's one added that has no value...

catsby:
https://github.com/hashicorp/terraform/pull/6687 has been merged, closing this, thanks!

PR #6687, opened 2016-05-16
provider/aws: Fix crash in ElastiCache param group

catsby:
Fixes crash reported in https://github.com/hashicorp/terraform/issues/6466 

When you edit/save the defaults on the web UI (even without changing anything), you get a lot of parameters with default values added. One such parameter (`notify-keyspace-events`) is returned that has no value 😱 

```
{
  DataType: "string",
  Description: "The keyspace events for Redis to notify Pub/Sub clients about. By default all notifications are disabled",
  IsModifiable: true,
  MinimumEngineVersion: "2.8.6",
  ParameterName: "notify-keyspace-events",
  Source: "user"
},
```

This PR only sets parameters locally if there is a value found


jen20:
LGTM!

PR #6691, opened 2016-05-16
pin github.com/bobtfish/go-nsone-api

josephholsten:


stack72:
Closing this - there is another PR open that takes care of this using govendor

PR #6692, opened 2016-05-16
nsone: import existing providers from https://github.com/bobtfish/terraform-provider-nsone

josephholsten:
This includes #6691 so that it will pass tests

* [ ] fix `golint` warnings


josephholsten:
deprecated in favor of #6744

PR #6703, opened 2016-05-17
website: docs for azurerm custom images

igable:
Adds docs on booting custom images with the azurerm_virtual_machine
resource; this was previously undocumented.


jen20:
:+1: thanks @igable!

PR #6704, opened 2016-05-17
core: Fix destroy when modules vars are used in resource counts

phinze:
A new problem was introduced by the prior fixes for destroy
interpolation messages when resources depend on module variables with
a _count_ attribute, this makes the variable crucial for properly
building the graph - even in destroys. So removing all module variables
from the graph as noops was overzealous.

By borrowing the logic in `DestroyEdgeInclude` we are able to determine
if we need to keep a given module variable relatively easily.

I'd like to overhaul the `Destroy: true` implementation so that it does
not depend on config at all, but I want to continue for now with the
targeted fixes that we can backport into the 0.6 series.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -2622,6 +2622,90 @@ module.child:
 	}
 }
 
+func TestContext2Apply_destroyWithModuleVariableAndCount(t *testing.T) {
+	m := testModule(t, "apply-destroy-mod-var-and-count")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+
+	var state *State
+	var err error
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Module: m,
+			Providers: map[string]ResourceProviderFactory{
+				"aws": testProviderFuncFixed(p),
+			},
+		})
+
+		// First plan and apply a create operation
+		if _, err := ctx.Plan(); err != nil {
+			t.Fatalf("plan err: %s", err)
+		}
+
+		state, err = ctx.Apply()
+		if err != nil {
+			t.Fatalf("apply err: %s", err)
+		}
+	}
+
+	h := new(HookRecordApplyOrder)
+	h.Active = true
+
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Destroy: true,
+			Module:  m,
+			State:   state,
+			Hooks:   []Hook{h},
+			Providers: map[string]ResourceProviderFactory{
+				"aws": testProviderFuncFixed(p),
+			},
+		})
+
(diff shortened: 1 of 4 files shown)


phinze:
cc @jen20 - not sure if this is something you noticed yourself or can be linked to an active issue

vancluever:
@phinze I'm definitely seeing this, and this just fixed the issue, so thank you very much :)

I was getting lots of destroy messages like:

```
* 1:20: unknown variable accessed: var.public_subnet_addresses in:

${length(split(",", var.public_subnet_addresses))}
* 1:20: unknown variable accessed: var.cidr_blocks in:

${length(split(",", var.cidr_blocks))}
* 1:20: unknown variable accessed: var.port_list in:

${length(split(",", var.port_list)) * length(split(",", var.cidr_blocks))}

```

The resources in question definitely do compute count off module parameters.

vancluever:
@phinze there might be a little more to this, the patch definitely fixed some of the issues, but I still have stuff abridged in the below gist:

https://gist.github.com/vancluever/61731817680b5788ffd240f3bc0cdb2c

Looks like the instances being referenced are in the root module.

vancluever:
More updates: Just looking at the module referenced here, and it looks like the variables being referenced are used as variables in nested modules that ultimately leads to variables used in count. Maybe `DestroyEdgeInclude` needs a bit of tweaking?

jen20:
I will verify against my reproductions of this here and review. I think @stack72 hit some also. Looks like it might need some additional work as per @vancluever though.

stack72:
Hi @phinze / @jen20 

I just tested this - it now allows me to destroy the resources I was seeing an issue with when I first reported this to @jen20 

I will leave the code review to the professionals ;)

P.

vancluever:
@phinze check out https://github.com/paybyphone/terraform/commit/c975906b810befc2bbc496068b0b72f0838a1fee, which makes it work recursively (I think!). Going to give this a shot at the office soon.

vancluever:
OK, looks like that fixed it for me - going to re-deploy and destroy the config I was working on it with and see if I see full intended behaviour  - there has to be some variables in there that should be no-ops.

phinze:
WOW nice work @vancluever! Taking a look now and lmk how your testing goes.

vancluever:
@phinze Confirmed working :) gist below has a successful noop on a module var:

https://gist.github.com/vancluever/b5fdc794f2def766860bcf22b97f6e2f

jen20:
@vancluever thanks for your work on this! I'm about to integrate it into the maintenance release we are about to push, and then into 0.7.

vancluever:
Woohoo! Glad I could help guys!

jen20:
Closed in favour of #6723.

PR #6710, opened 2016-05-17
Update doc for AWS allowed termination policies

jkburges:
Include the list of allowed values for AWS auto scaling group termination policies.

This saves the terraform user from having to go look up AWS doc.  It's done in other places in the terraform doc already, e.g.: the `enabled_metrics` attribute on this same page.

Ref: http://docs.aws.amazon.com/autoscaling/latest/userguide/as-instance-termination.html#custom-termination-policy


stack72:
Hi @jkburges 

This looks good - thanks so much :)

Paul

PR #6715, opened 2016-05-17
core: honor "destroy" diffs for data resources

apparentlymart:
Previously the `planDestroy` pass would correctly produce a destroy diff, but the "apply" pass would just ignore it and make a fresh diff, turning it back into a "create" because data resources are always eager to refresh.

Now we consider the previous diff when re-diffing during apply and so we can preserve the plan to destroy and then ultimately actually "destroy" the data resource (remove from the state) when we get to `ReadDataApply`.

This ensures that the state is left empty after `terraform destroy`; previously we would leave behind data resource states.

### DIFF
--- a/terraform/eval_read_data.go
+++ b/terraform/eval_read_data.go
@@ -12,12 +12,14 @@ type EvalReadDataDiff struct {
 	OutputState **InstanceState
 	Config      **ResourceConfig
 	Info        *InstanceInfo
+
+	// Set Previous when re-evaluating diff during apply, to ensure that
+	// the "Destroy" flag is preserved.
+	Previous **InstanceDiff
 }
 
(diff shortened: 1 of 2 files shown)


jen20:
@apparentlymart Should we consider adding a test for empty state following destroy around this behaviour?

apparentlymart:
Honestly this whole area has pretty weak test coverage, since it was derived from similar code for managed resources that also has pretty weak test coverage.

I could add some tests around this `EvalNode` in particular, but probably the behavior we actually want to test is that a configuration with a data source in it cleans itself up properly, and that's probably something better tested over a broader set of code than just a unit test.

Do you have any recommendations about what is the appropriate level to test this at?

jen20:
Seems to me like we should be able to write a context test for this - I'll give that a shot now.

apparentlymart:
Looks like this was merged

PR #6718, opened 2016-05-17
Fixes an vet error.

buzztroll:
The head of master currently breaks when running make with the error below.  This PR fixes that:

go tool vet -all .
builtin/providers/openstack/resource_openstack_compute_instance_v2.go:393: github.com/hashicorp/terraform/vendor/github.com/rackspace/gophercloud/openstack/compute/v2/extensions/bootfromvolume.CreateOptsExt composite literal uses unkeyed fields

Vet found suspicious constructs. Please check the reported constructs
and fix them if necessary before submitting the code for review.
make: *** [vet] Error 1


jtopjian:
LGTM! Acceptance tests pass with this patch in place. Thank you!

PR #6731, opened 2016-05-17
[WIP] AWS APIGateway Custom Authorizer

johnjelinek:
I see that there is already an AWS APIGateway Authorizer. I'm going to try to wire up the resource to an AWS APIGateway method. Looking for feedback along the way :). This is a WIP placeholder unless someone else can beat me to it.

### DIFF
--- a/builtin/providers/aws/resource_aws_api_gateway_authorizer.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_authorizer.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -200,7 +201,11 @@ func resourceAwsApiGatewayAuthorizerDelete(d *schema.ResourceData, meta interfac
 	log.Printf("[INFO] Deleting API Gateway Authorizer: %s", input)
 	_, err := conn.DeleteAuthorizer(&input)
 	if err != nil {
-		return fmt.Errorf("Deleting API Gateway Authorizer failed: %s", err)
+		// XXX: Figure out a way to delete the method that depends on the authorizer first
+		// otherwise the authorizer will be dangling until the API is deleted
+		if !strings.Contains(err.Error(), "ConflictException") {
+			return fmt.Errorf("Deleting API Gateway Authorizer failed: %s", err)
+		}
 	}
 
 	return nil
--- a/builtin/providers/aws/resource_aws_api_gateway_method.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_method.go
@@ -45,6 +45,11 @@ func resourceAwsApiGatewayMethod() *schema.Resource {
 				Required: true,
 			},
 
+			"authorizer_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
 			"api_key_required": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -89,6 +94,7 @@ func resourceAwsApiGatewayMethodCreate(d *schema.ResourceData, meta interface{})
 		// TODO reimplement once [GH-2143](https://github.com/hashicorp/terraform/issues/2143) has been implemented
 		RequestParameters: aws.BoolMap(parameters),
 		ApiKeyRequired:    aws.Bool(d.Get("api_key_required").(bool)),
+		AuthorizerId:      aws.String(d.Get("authorizer_id").(string)),
 	})
 	if err != nil {
 		return fmt.Errorf("Error creating API Gateway Method: %s", err)
--- a/builtin/providers/aws/resource_aws_api_gateway_method_test.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_method_test.go
@@ -44,12 +44,45 @@ func TestAccAWSAPIGatewayMethod_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSAPIGatewayMethod_customauthorizer(t *testing.T) {
+	var conf apigateway.Method
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSAPIGatewayMethodDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSAPIGatewayMethodConfigWithCustomAuthorizer,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAPIGatewayMethodExists("aws_api_gateway_method.test", &conf),
+					testAccCheckAWSAPIGatewayMethodAttributes(&conf),
+					resource.TestCheckResourceAttr(
+						"aws_api_gateway_method.test", "http_method", "GET"),
+					resource.TestCheckResourceAttr(
+						"aws_api_gateway_method.test", "authorization", "CUSTOM"),
+					resource.TestCheckResourceAttr(
+						"aws_api_gateway_method.test", "request_models.application/json", "Error"),
+				),
+			},
+
+			resource.TestStep{
+				Config: testAccAWSAPIGatewayMethodConfigUpdate,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAPIGatewayMethodExists("aws_api_gateway_method.test", &conf),
+					testAccCheckAWSAPIGatewayMethodAttributesUpdate(&conf),
+				),
+			},
+		},
+	})
+}
+
 func testAccCheckAWSAPIGatewayMethodAttributes(conf *apigateway.Method) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		if *conf.HttpMethod != "GET" {
 			return fmt.Errorf("Wrong HttpMethod: %q", *conf.HttpMethod)
 		}
-		if *conf.AuthorizationType != "NONE" {
+		if *conf.AuthorizationType != "NONE" && *conf.AuthorizationType != "CUSTOM" {
 			return fmt.Errorf("Wrong Authorization: %q", *conf.AuthorizationType)
 		}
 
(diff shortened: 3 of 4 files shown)


johnjelinek:
I'm trying to figure out how to handle the following:
```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAPIGatewayMethod_customauthorizer'
/Users/administrator/go/bin/stringer
2016/05/17 20:42:37 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAPIGatewayMethod_customauthorizer -timeout 120m
=== RUN   TestAccAWSAPIGatewayMethod_customauthorizer
--- FAIL: TestAccAWSAPIGatewayMethod_customauthorizer (28.45s)
	testing.go:234: Step 1 error: Error applying: 1 error(s) occurred:

		* aws_api_gateway_authorizer.test: Deleting API Gateway Authorizer failed: ConflictException: Cannot delete authorizer 'tf-acc-test-authorizer', is referenced in method: GET//test
			status code: 409, request id: e66e977f-1c99-11e6-ab1a-5d2160b20552
	testing.go:295: Error destroying resource! WARNING: Dangling resources

		Error: Error applying: 1 error(s) occurred:

		* aws_api_gateway_authorizer.test: Deleting API Gateway Authorizer failed: ConflictException: Cannot delete authorizer 'tf-acc-test-authorizer', is referenced in method: GET//test
			status code: 409, request id: e871449c-1c99-11e6-af10-131e5fb9adc0

		State: aws_api_gateway_authorizer.test:
		  ID = 87kvkh
		  authorizer_credentials = arn:aws:iam::116738426468:role/tf_acc_api_gateway_auth_invocation_role
		  authorizer_result_ttl_in_seconds = 0
		  authorizer_uri = arn:aws:apigateway:region:lambda:path/2015-03-31/functions/arn:aws:lambda:us-west-2:116738426468:function:tf_acc_api_gateway_authorizer/invocations
		  identity_source = method.request.header.Authorization
		  identity_validation_expression =
		  name = tf-acc-test-authorizer
		  rest_api_id = opatedcbg0
		  type = TOKEN

		  Dependencies:
		    aws_api_gateway_rest_api.test
		    aws_lambda_function.authorizer
		    aws_iam_role.invocation_role
		aws_api_gateway_rest_api.test:
		  ID = opatedcbg0
		  description =
		  name = test
		  root_resource_id = 3vmcy4llok
		aws_iam_role.iam_for_lambda:
		  ID = tf_acc_iam_for_lambda_api_gateway_authorizer
		  arn = arn:aws:iam::116738426468:role/tf_acc_iam_for_lambda_api_gateway_authorizer
		  assume_role_policy = {
		  "Version": "2012-10-17",
		  "Statement": [
		    {
		      "Action": "sts:AssumeRole",
		      "Principal": {
		        "Service": "lambda.amazonaws.com"
		      },
		      "Effect": "Allow",
		      "Sid": ""
		    }
		  ]
		}

		  name = tf_acc_iam_for_lambda_api_gateway_authorizer
		  path = /
		  unique_id = AROAJJNV5T5L7XCZIN5VO
		aws_iam_role.invocation_role:
		  ID = tf_acc_api_gateway_auth_invocation_role
		  arn = arn:aws:iam::116738426468:role/tf_acc_api_gateway_auth_invocation_role
		  assume_role_policy = {
		  "Version": "2012-10-17",
		  "Statement": [
		    {
		      "Action": "sts:AssumeRole",
		      "Principal": {
		        "Service": "apigateway.amazonaws.com"
		      },
		      "Effect": "Allow",
		      "Sid": ""
		    }
		  ]
		}

		  name = tf_acc_api_gateway_auth_invocation_role
		  path = /
		  unique_id = AROAIGQR5VMFQTQX4U43A
		aws_lambda_function.authorizer:
		  ID = tf_acc_api_gateway_authorizer
		  arn = arn:aws:lambda:us-west-2:116738426468:function:tf_acc_api_gateway_authorizer
		  description =
		  filename = test-fixtures/lambdatest.zip
		  function_name = tf_acc_api_gateway_authorizer
		  handler = exports.example
		  last_modified = 2016-05-18T01:43:05.135+0000
		  memory_size = 128
		  role = arn:aws:iam::116738426468:role/tf_acc_iam_for_lambda_api_gateway_authorizer
```
any tips?

johnjelinek:
@radeksimko: could you have a look here?

johnjelinek:
I got a repro, where if the authorizer is deleted prior to the apigateway method, then it results in the conflict exception:
```
aws apigateway delete-authorizer --rest-api-id opatedcbg0 --region=us-west-2 --authorizer-id 87kvkh

A client error (ConflictException) occurred when calling the DeleteAuthorizer operation: Cannot delete authorizer 'tf-acc-test-authorizer', is referenced in method: GET//
```

builtin/providers/aws/resource_aws_api_gateway_method_test.go
```diff
@@ -154,6 +187,108 @@ func testAccCheckAWSAPIGatewayMethodDestroy(s *terraform.State) error {
...
+  parent_id = "${aws_api_gateway_rest_api.test.root_resource_id}"
+  path_part = "test"
+}
+
+resource "aws_api_gateway_method" "test" {
+  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
+  resource_id = "${aws_api_gateway_resource.test.id}"
+  http_method = "GET"
+  authorization = "CUSTOM"
+  authorizer_id = "${aws_api_gateway_authorizer.test.id}"
```
johnjelinek: I think this is the tricky part. Because I have a reference to the `authorizer_id` here, terraform wants to delete the authorizer before deleting the method, but the AWS SDK API expects all methods to be deleted first. Maybe instead, the authorizer should have a collection of methods it should attach to, instead. What do you think?

johnjelinek:
If I set an explicit dependency like this:
```
resource "aws_api_gateway_authorizer" "test" {
  name = "tf-acc-test-authorizer"
  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
  authorizer_uri = "arn:aws:apigateway:region:lambda:path/2015-03-31/functions/${aws_lambda_function.authorizer.arn}/invocations"
  authorizer_credentials = "${aws_iam_role.invocation_role.arn}"
  depends_on = ["aws_api_gateway_method.test"]
}
```
then I get an error like this:
```
--- FAIL: TestAccAWSAPIGatewayMethod_customauthorizer (0.01s)
	testing.go:234: Step 0 error: Configuration is invalid.

		Warnings: []string(nil)

		Errors: []string{"1 error(s) occurred:\n\n* Cycle: aws_api_gateway_authorizer.test, aws_api_gateway_method.test"}
```

johnjelinek:
I added a `TODO` for anyone who knows how to change the destroy order without setting cyclical dependencies. I'm gonna update the docs and then, I guess this will be ready for merging.

johnjelinek:
@radeksimko: docs are in place for review.

johnjelinek:
any feedback?

gthole:
This is pretty important for our use case, and the `authorizer` resource is useless without completing the integration to the `method` like this.  Can this be reviewed please?  (@johnjelink possibly remove the WIP tag if it's ready for review.)

I'm going to cherry-pick this into my fork and test it against our stack.  I'm happy to report back here if it will be helpful.

johnjelinek:
That would be super helpful if you could test. I can remove WIP also.

gthole:
I built the binaries with this commit on top of v0.6.16 (`make dev` apparently doesn't create all the plugins on `master` anymore, which is the only reason I used the v0.6.16 tag as my base).  It worked great for me and I've been able to actually hook up the custom `authorizer` resource to something.

johnjelinek:
That's great! Glad to hear it. Can we get this merged?

builtin/providers/aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -200,7 +201,11 @@ func resourceAwsApiGatewayAuthorizerDelete(d *schema.ResourceData, meta interfac
 	log.Printf("[INFO] Deleting API Gateway Authorizer: %s", input)
 	_, err := conn.DeleteAuthorizer(&input)
 	if err != nil {
-		return fmt.Errorf("Deleting API Gateway Authorizer failed: %s", err)
+		// XXX: Figure out a way to delete the method that depends on the authorizer first
+		// otherwise the authorizer will be dangling until the API is deleted
+		if !strings.Contains(err.Error(), "ConflictException") {
```
radeksimko: As long as the authorizer is managed in the same terraform scope and is referenced properly (i.e. **not hardcoded** authorizer IDs in methods) Terraform will schedule the deletion correctly by default - i.e. for complete destruction `1st method => 2nd authorizer`. Have a look at the dependency graph via `terraform graph`. 😉 

The only issue that may theoretically arise is eventual consistency of the AWS API (or the implementation of it) - i.e. the `authorizer_id` change may take time to propagate. We usually just retry the deletion in such cases with a reasonable timeout.

Did you experience this problem yourself @johnjelinek ?

builtin/providers/aws/resource_aws_api_gateway_method_test.go
radeksimko: > terraform wants to delete the authorizer before deleting the method

in what situation exactly? on complete `terraform destroy`? I believe the opposite is reality:
https://gist.github.com/radeksimko/3d9b0637ad821b34de7be4e7e2c993fe - snippet below:

```
aws_api_gateway_method.MyDemoMethod: Destruction complete
...
aws_api_gateway_authorizer.demo: Destroying...
aws_api_gateway_authorizer.demo: Destruction complete
```

builtin/providers/aws/resource_aws_api_gateway_method.go
```diff
@@ -89,6 +94,7 @@ func resourceAwsApiGatewayMethodCreate(d *schema.ResourceData, meta interface{})
 		// TODO reimplement once [GH-2143](https://github.com/hashicorp/terraform/issues/2143) has been implemented
 		RequestParameters: aws.BoolMap(parameters),
 		ApiKeyRequired:    aws.Bool(d.Get("api_key_required").(bool)),
+		AuthorizerId:      aws.String(d.Get("authorizer_id").(string)),
```
radeksimko: I think we'll need to update the other functions too so that user can _change_ the `authorizer_id` and Terraform can read it back if it's changed outside of Terraform.

i.e.

`resourceAwsApiGatewayMethodRead` will need `d.Set("authorizer_id", out.AuthorizerId)` - I noticed we're missing some other fields there - if you want to fix that too, that's great, but we can fix that in a separate PR.

and `resourceAwsApiGatewayMethodUpdate` will need `if d.HasChange("authorizer_id") {` block.

johnjelinek: If this all works for you, let's :shipit:

builtin/providers/aws/resource_aws_api_gateway_method_test.go
```diff
@@ -154,6 +187,108 @@ func testAccCheckAWSAPIGatewayMethodDestroy(s *terraform.State) error {
 	return nil
 }
 
+const testAccAWSAPIGatewayMethodConfigWithCustomAuthorizer = `
+resource "aws_api_gateway_rest_api" "test" {
+  name = "test"
```
radeksimko: Probably a nitpick, but it may be better to give the API unique name, e.g. `tf-acc-test-authorizer`. Should these tests leak (i.e. leave undeleted resources behind) we will be able to track the test more easily.

radeksimko:
Hi @johnjelinek ,
I did try to run attached acceptance test couple of times (5x to be exact) and it always passed. I was not able to reproduce the `ConflictException` mentioned in your initial post.

If you saw this error while using the exact same config, it must be eventual consistency and we should add [Retry block](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_iam_role.go#L108-L120) in that case. If you do decide to add such retry block, I'd suggest you check the error code via [exact match](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_iam_role.go#L133) instead of using `strings.Contains()`.

I left you one more comment about reading + updating the field. Once that's resolved I'm happy to merge this 😉

builtin/providers/aws/resource_aws_api_gateway_method.go
radeksimko: Have a look at my other comments - there is a few other things before we can :shipit: 😉

radeksimko: It works for common workflow - i.e. create/destroy, doesn't work for update and won't work for import either when we add it.

zanona:
+1 for merging.

stack72:
Closed in favour of #8535

PR #6742, opened 2016-05-18
provider/azurerm: Add support for exporting the `azurerm_storage_account` access keys

### LINKED ISSUE #6739: Azure RM: azurerm_storage_account should export storage keys
The azurerm_storage_account should export storage keys for use in later Terraform operations.  This would be particularly helpful when creating a storage account, then using it in an Azure RM template deployment where the storage key is an input param (ex: HD Insights cluster).

stack72:
Fixes #6739 

Please note that we do NOT have the ability to manage the access keys -
we are just getting the keys that the account creates for us. To manage
the keys, you would need to use the azure portal still

### DIFF
--- a/builtin/providers/azurerm/resource_arm_storage_account.go
+++ b/builtin/providers/azurerm/resource_arm_storage_account.go
@@ -90,6 +90,16 @@ func resourceArmStorageAccount() *schema.Resource {
 				Computed: true,
 			},
 
+			"primary_access_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"secondary_access_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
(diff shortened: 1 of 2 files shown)


jen20:
LGTM.

Preskton:
Such response time! <3

dinnak:
thoughts on when this might be available via a build?

stack72:
Hi @dinnak

This will be part of the terraform 0.7 release

Paul

ryhoward:
Awesome! Any idea on when 0.7 will be pushed out exactly? A couple weeks?

stack72:
Hi @ryhoward 

I'm.not 100% sure here. I am not a member of the core team so one of @phinze or @jen20 would have to reply

Paul

ryhoward:
thanks @stack72

dinnak:
thanks @stack72

PR #6747, opened 2016-05-18
core: Write V1 state by default in 0.6.17 release

jen20:
If there is no state, we want to write V1 rather than V2 by default. In this case ReadVersion will be set to 0 (the same as binary state - if we encounter this we want to upgrade to V1 JSON).

The test inspects the JSON directly to avoid any bad results owing to upgrade on read behaviour that may be hard to reason about.


phinze:
LGTM!

PR #6751, opened 2016-05-18
[WIP] providers/aws: expose network interface id

zsalzbank:
Expose the network interface ID that is created with a new instance.

This can be useful when associating an existing elastic IP to the
default interface on an instance that has multiple network interfaces.

I wasn't sure if any tests were required for this. When I looked at the existing tests, I didn't see any for attributes that are returned by the provider. If I need to add a test, can somebody point me in the right direction for where to add them?


builtin/providers/aws/resource_aws_instance.go
```diff
@@ -491,10 +496,12 @@ func resourceAwsInstanceRead(d *schema.ResourceData, meta interface{}) error {
 		for _, ni := range instance.NetworkInterfaces {
 			if *ni.Attachment.DeviceIndex == 0 {
 				d.Set("subnet_id", ni.SubnetId)
+				d.Set("network_interface_id", ni.NetworkInterfaceId)
 			}
 		}
 	} else {
 		d.Set("subnet_id", instance.SubnetId)
+		d.Set("network_interface_id", "")
```
stack72: Hi @zsalzbank 

Why do we set the `network_interface_id` to an empty string in this scenario?

P.

zsalzbank: Because there are no network interfaces to get it from. If there is one from somewhere else that I don't know about, I'd be happy to change it.

stack72: got it - thanks

PR #6752, opened 2016-05-18
core: Forward port OutputState to 0.7

jen20:
The 0.6.17 release (on the maint-0.6 branch) introduced OutputState as a complex type in the state JSON V2, and a bunch of migrations for dealing with it. This PR forward ports that and another fix to the 0.7 series.


phinze:
That ported nicely! LGTM

PR #6753, opened 2016-05-18
: core: Fix destroy w/module vars used in counts

jen20:
_This is the same PR as #6723 but targeted at the 0.7 release rather than 0.6.17._

This PR combines the work of @phinze in #6704 and @vancluever on the paybyphone fork of Terraform.

First from @phinze:

A new problem was introduced by the prior fixes for destroy interpolation messages when resources depend on module variables with a count attribute, this makes the variable crucial for properly building the graph - even in destroys. So removing all module variables from the graph as noops was overzealous.

By borrowing the logic in DestroyEdgeInclude we are able to determine if we need to keep a given module variable relatively easily.

I'd like to overhaul the Destroy: true implementation so that it does not depend on config at all, but I want to continue for now with the targeted fixes that we can backport into the 0.6 series.

Then from @vancluever in paybyphone@c975906:

Building on b10564a, adding tweaks that allow the module var count search to act recursively, ensuring that a sitaution where something like var.top gets passed to module middle, as var.middle, and then to
module bottom, as var.bottom, which is then used in a resource count.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -2706,6 +2706,92 @@ module.child:
 	}
 }
 
+func TestContext2Apply_destroyWithModuleVariableAndCountNested(t *testing.T) {
+	m := testModule(t, "apply-destroy-mod-var-and-count-nested")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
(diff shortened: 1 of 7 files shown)


phinze:
LGTM

PR #6756, opened 2016-05-18
[WIP] provider/openstack: Remove Volume Support from Instances

jtopjian:
This commit removes volume management from the openstack_compute_instance_v2
resource and moves all functionality into the openstack_blockstorage_volume_v1
resource. Attaching and detaching volumes is now handled in the volume
resource.

Booting from a volume is still handled in the instance resource.


jtopjian:
This is beginning to go down a rabbit hole that is already solved with the current implementation. I'm going to close this PR, but it'll be in the PR history for others to find in the future and maybe have some input on.

PR #6761, opened 2016-05-19
fix aws instance read subnet eth0

Djuke:
Reading the aws instance it was assumed that eth0 was the first in the list of network interfaces. This is true when there is only one interface, but if you attach multiple network interfaces with different subnets then you don't know which subnet of which ni is read.

With this fix I assume that the subnet you want is the one of the ni with device index 0. The ni upon creating the instance gets device index 0.


Djuke:
this my conflict with https://github.com/hashicorp/terraform/pull/6387, but it has the same fix

radeksimko:
I managed to reproduce the bug with the following config: https://gist.github.com/radeksimko/4ea29792dfd8522c04de5e3343ed705b
which I had to `destroy` & `apply` couple of times.

I originally wanted to ask you to add a test covering this, but it seems there's no easy way to test it as it's intermittent issue (AWS returns attached ENIs in no particular order).

Thanks for catching and fixing this @Djuke 👍

Issue #4129, opened 2015-12-01
Question: Configurable timeout for terraform apply on OpenStack


afritzler:
Hi,

is there a way to specify a timeout when running terraform apply against the OpenStack provider (or any other providers)?

Regards,
Andreas

jtopjian:
Almost all OpenStack resources have a timeout, though the timeout is set pretty high. It's not configurable, at the moment, either. AFAIK, this is how other resources in other providers are configured, too.

Are you running into a resource that does not have a timeout yet? Or are you running into a situation where the timeout should be customizable?

afritzler:
Since I'm trying to create/delete pretty big instances on OpenStack, I am sometimes running into timeout situations. So I was wondering if there is a way to overwrite the default settings. Btw. what are the default timeouts for creating/deleting instances?

jtopjian:
Instance creation and deletion are set to 10 minutes. If the state does not change to `ACTIVE` within 10 minutes, Terraform will error / timeout.

Would you say the instances you're creating are taking longer than that? If not, you might be running into a different issue.

nicgrayson:
I would also like this timeout to be configurable. I have some image copying that takes almost 20 minutes to run.

jtopjian:
I'm going to go ahead and close this ticket. Per comments in #4223, configuring the timeout is probably much larger in scope than just OpenStack. If there is a specific area where you're hitting the timeout, please let me know where and I'll increase it (or submit a PR similar to #4223).

tjlee0909:
This is slightly embarassing, but at Box we actually have an image that is so large that it takes close to an hour to provision, depending on network traffic and whether or not that particular KVM host has a cached version of the image or not.

Do you know if there are any immediate plans to support timeouts? We are currently evaluating terraform to see if it fits out requirements.

jtopjian:
@tjlee0909 I'm going to re-open this question and investigate options.

ZZelle:
I have exactly the same trouble: i need to import large images (>300Go) and start them on VMware and 30min is not enough to copy the VM image or to boot the 1st VM instance with this image (during 1st boot the image is copied from image service backend to compute storage)

Fodoj:
Timeout for resising instance is also too low now.

mattsoftware:
The timeout is also too low for me, trying to create an aws aurora cluster from a large snapshot, takes more than the 40 minutes hardcoded into the binary.

pat610:
[...]

PR #6764, opened 2016-05-19
provider/openstack Increase timeouts for image resize, subnets and routers

Fodoj:
https://github.com/hashicorp/terraform/issues/4129 related

I increased timeouts for some resources. For others timeouts seems to be okay. but re-sizing instances and deleting subnets\routers regulary lead to timeout errors.


jtopjian:
👍

PR #6768, opened 2016-05-19
Normalizes availability_set_id to lowercase to avoid spurious diffs

Bowbaq:
Azure RM seems to return uppercased availability set ids from the API, which causes spurious diffs if the case used in the definition is anything else than uppercase. This normalizes the id to lowercase.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_virtual_machine.go
+++ b/builtin/providers/azurerm/resource_arm_virtual_machine.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/arm/compute"
@@ -69,6 +70,9 @@ func resourceArmVirtualMachine() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
+				StateFunc: func(id interface{}) string {
+					return strings.ToLower(id.(string))
+				},
 			},
 
 			"license_type": &schema.Schema{
@@ -480,7 +484,7 @@ func resourceArmVirtualMachineRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	if resp.Properties.AvailabilitySet != nil {
-		d.Set("availability_set_id", resp.Properties.AvailabilitySet.ID)
+		d.Set("availability_set_id", strings.ToLower(*resp.Properties.AvailabilitySet.ID))
 	}
 
 	d.Set("vm_size", resp.Properties.HardwareProfile.VMSize)


stack72:
Hi @Bowbaq 

Thanks for the PR here - it's awesome to see someone else try to fix this. I have been thinking about this a lot and I am not quite sure it is as easy as this. The Azure API is actually case-sensitive i.e.

parsing a resource ID for VirtualMachines will return no result but virtualMachines will (notice the lowercase v here)

Therefore, we need to run some tests on this to make sure that the virtual machine is actually being added to the correct availability_set when using lowercase

Does this make sense? Have you tried the tests?

Paul

Bowbaq:
Seems to work fine in this case

stack72:
This is good for me :) We can merge it and then let it bake in the nightly tests :) Thanks so much for the work here

dougt1:
Hi @stack72 & @Bowbaq  - i think this change may have introduced a bug. I am running v 0.7.0 in Azure (Resource Manager) with 2 VMs in an availability set - and now on every run my location is seen as different (as appears case sensitive) - this results in the forced creation of new resources within the availability group - resulting in the deletion and re-creation of all vm's in the availability set. It doesn't seem to matter what case I specify the location as - on subsequent runs it always returns as a change and wants to recreate the resource group and the VM's within the group.

Results from a terraform plan below:-

-/+ module.availability_set_test.azurerm_availability_set.availability_set
    location:                     "UKSouth2" => "uksouth2" (forces new resource)
    name:                         "availability-set-test" => "availability-set-test"
    platform_fault_domain_count:  "3" => "3"
    platform_update_domain_count: "5" => "5"
    resource_group_name:          "test-prod-rg" => "test-prod-rg"
    tags.%:                       "1" => "1"
    tags.environment:             "production" => "production"

Also - when i look at the Travis CI build results, unless I am reading it wrong it looks like there are no tests for Azure RM:-

ok  	github.com/hashicorp/terraform/builtin/bins/provider-azure	0.006s
?   	github.com/hashicorp/terraform/builtin/bins/provider-azurerm	[no test files]
?   	github.com/hashicorp/terraform/builtin/bins/provider-chef	[no test files]

so maybe we need to look at adding in some automated tests as well?

Thanks.

stack72:
Morning @dougt1 

So this PR only touched `availability_set_id`, not location. The issue here is that you are using a non-normalized version of the location. In your case, the lowercase `uksouth2` or `UK South 2` would work - this specific issue you found is another bug

Also, FYI, the make test is a unit tests thing - we have a full suite of automated tests for AzureRM :) https://travis-ci.org/hashicorp/terraform/builds/152307805

We run these each night

I will try and dig into the location thing - can you open a new issue to mark the work on that?

P.

PR #6785, opened 2016-05-19
provider/vSphere Disk Controller Type

dkalleg:
Gives the user the ability to specify the controller type they would
like to connect their disk to.  Supported options are scsi and ide.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -39,10 +39,11 @@ type networkInterface struct {
 }
 
 type hardDisk struct {
-	size     int64
-	iops     int64
-	initType string
-	vmdkPath string
+	size       int64
+	iops       int64
+	initType   string
+	vmdkPath   string
+	controller string
 }
 
 //Additional options Vsphere can use clones of windows machines
@@ -377,6 +378,21 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Default:  false,
 							ForceNew: true,
 						},
+
+						"controller_type": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "scsi",
+							ForceNew: true,
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+								value := v.(string)
+								if value != "scsi" && value != "ide" {
+									errors = append(errors, fmt.Errorf(
+										"only 'scsi' and 'ide' are supported values for 'controller_type'"))
+								}
+								return
+							},
+						},
 					},
 				},
 			},
@@ -667,6 +683,9 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 			if v, ok := disk["type"].(string); ok && v != "" {
 				disks[i].initType = v
 			}
+			if v, ok := disk["controller_type"].(string); ok && v != "" {
+				disks[i].controller = v
+			}
 		}
 		vm.hardDisks = disks
 		log.Printf("[DEBUG] disk init: %v", disks)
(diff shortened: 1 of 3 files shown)


builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -377,6 +378,11 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Default:  false,
 							ForceNew: true,
 						},
+
+						"controller_type": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
```
stack72: if `scsi` is the default, then we should specify that here and now have to do the if / else on it further down. We just need to document it is `scsi`

stack72:
hi @dkalleg 

I left a commend inline about the default but the other thing to note is that we are only using this value on `Create`. Therefore, if someone changes it, it won't make the suggested changes - if this is the expected case (that we don't allow updates), should it not be marked as `ForceNew`?

P.

dkalleg:
Hey @stack72 Thanks for the feedback.  In my other PR for adding disk updates:
https://github.com/hashicorp/terraform/pull/6619
I removed Defaults because they were causing side effects to the diff calls in the update scenarios.. for the life of me I can't remember the details.  I just tried reproducing whatever I saw and couldn't :)  Maybe due to switching from TypeList to TypeSet... I'll put the Defaults back in for this and my disk update PR.  

I also didn't touch update here since my other PR sets up the framework for dealing with disk updates.  I've put myself into a chicken & egg scenario.  I'd think the path of least resistance would be to see this PR merge (sans update story) then I can rebase my disk update PR to include this, and address the controller_type update there.  If I try to address update in this PR it'll be a diff'ing nightmare.  Thoughts on this?

I was stuck in the mindset of not having ForceNew on disk attributes since the disk update story removes them, which is why I left it out here :)  But for this iteration, I'd agree that ForceNew belongs on controller_type (until disk update lands).  I'll add this as well.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
chrislovecnm: @stack72 agreed

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -377,6 +378,21 @@ func resourceVSphereVirtualMachine() *schema.Resource {
...
+							Default:  "scsi",
+							ForceNew: true,
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+								value := v.(string)
+								if value != "scsi" && value != "ide" {
+									errors = append(errors, fmt.Errorf(
```
chrislovecnm: We also can check this in the schema with a validator ... Which we should be doing more.  I will help us long term to keep the variable error handling in the schema.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -873,14 +892,14 @@ func resourceVSphereVirtualMachineDelete(d *schema.ResourceData, meta interface{
 }
 
 // addHardDisk adds a new Hard Disk to the VirtualMachine.
-func addHardDisk(vm *object.VirtualMachine, size, iops int64, diskType string, datastore *object.Datastore, diskPath string) error {
+func addHardDisk(vm *object.VirtualMachine, size, iops int64, diskType string, datastore *object.Datastore, diskPath string, controller_type string) error {
```
chrislovecnm: We need to refactor this method ... passing all of these args is nuts.  Nothing you need to do in this pr ...

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -873,14 +892,14 @@ func resourceVSphereVirtualMachineDelete(d *schema.ResourceData, meta interface{
...
 	devices, err := vm.Device(context.TODO())
 	if err != nil {
 		return err
 	}
 	log.Printf("[DEBUG] vm devices: %#v\n", devices)
 
-	controller, err := devices.FindDiskController("scsi")
```
chrislovecnm: We need to update a unit test.  You could use one that adds two disk, add one scsi the other ide :)

dkalleg:
@chrislovecnm Added some controller settings to an existing test.  Passing for me.

I'm on board for more validators and some future refactors.

Please let me know if theres something else you'd like to see addressed here that I haven't yet.

chrislovecnm:
Nope ... Question though... Will you two PRs collide?

dkalleg:
@chrislovecnm I'm sure there will be some refactoring necessary after one of them is merged.  Can't avoid that :)

chrislovecnm:
@stack72 @phinze @jen20 @catsby LGTM ... can we get a merge?

PR #6793, opened 2016-05-20
Update EC2 example to modern AMI & instance

AdamWhittingham:
The current intro docs give an example config to get you going.

However, this uses and older `t1.micro` instance and deprecated AMI, so users following the guide will get the following error when running `terraform apply`:

```
Error launching source instance: InvalidParameterCombination: Non-Windows instances with a virtualization type of 'hvm' are currently not supported for this instance type.
```

This PR updates the example config to a `t2.micro` instance and the latest Amazon Linux AMI to make the guide easier to follow again.

### DIFF
--- a/website/source/intro/getting-started/build.html.md
+++ b/website/source/intro/getting-started/build.html.md
@@ -59,8 +59,8 @@ provider "aws" {
 }
 
 resource "aws_instance" "example" {
-	ami = "ami-408c7f28"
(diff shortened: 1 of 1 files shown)


catsby:
Hey @AdamWhittingham thanks for the patch here – In the end I had to do a bit more refactoring of the docs so that EC2 Classic users could still follow along. I made the updates in #6801 if you're interested. Thanks!

PR #6800, opened 2016-05-20
Added a missing 'not' to the docs

davidpellerin:


catsby:
Hey @davidpellerin thanks for the docs update. I'm curious, do you have a config that demonstrates this? The AWS API Docs are a bit confusing here:

- http://docs.aws.amazon.com/opsworks/latest/APIReference/API_CreateStack.html#API_CreateStack_RequestSyntax

>DefaultAvailabilityZone
The stack's default Availability Zone, which must be in the specified region. For more information, see Regions and Endpoints. If you also specify a value for DefaultSubnetId, the subnet must be in the same zone. For more information, see the VpcId parameter description.

> VpcId
The ID of the VPC that the stack is to be launched into. The VPC must be in the stack's region. All instances are launched into this VPC. You cannot change the ID later.

> * If your account supports EC2-Classic, the default value is no VPC.
* If your account does not support EC2-Classic, the default value is the default VPC for the specified region.

> If the VPC ID corresponds to a default VPC and you have specified either the DefaultAvailabilityZone or the DefaultSubnetId parameter only, AWS OpsWorks infers the value of the other parameter. If you specify neither parameter, AWS OpsWorks sets these parameters to the first valid Availability Zone for the specified region and the corresponding default VPC subnet ID, respectively.

> If you specify a nondefault VPC ID, note the following:

>* It must belong to a VPC in your account that is in the specified region.
* You must specify a value for DefaultSubnetId.
For more information on how to use AWS OpsWorks with a VPC, see Running a Stack in a VPC. For more information on default VPC and EC2-Classic, see Supported Platforms.

>Type: String

>Required: No

My interpretation there is that omitting the `default_availability_zone` is completely fine. Do you have a config that's showing otherwise?

davidpellerin:
@catsby I agree that the AWS documentation is somewhat confusing.  I guess the reason I changed it was more for clarity.  Since it says it's an "optional" parameter, I found the wording to be confusing.  

This is not required unless you set `vpc_id`.

could also say

This is required **IF** you set `vpc_id`

That might sound better.

stack72:
Hi all

I have spent some time with the docs on this one and agree it is confusing. I believe the following to be true:

* `default-subnet-id` is required if you specify a value for the VpcId
* The only limition of adding a `default-availability-zone` is that is must be (according to the docs):

```
If you also specify a value for DefaultSubnetId , the subnet must be in the same zone
```
* If you specify neither `default-subnet-id` or `default-availability-zone`, AWS OpsWorks sets these parameters to the first valid Availability Zone for the specified region and the corresponding default VPC subnet ID, respectively.

Therefore, I am going to go ahead and close this one out. Adding this to the docs may be confusing to users who also cross examine the AWS docs when creating their terraform config. This has also been confirmed by the AWS console, which auto-selects the AV / Subnet

Thanks for all the work here

Paul

PR #6801, opened 2016-05-20
provider/aws: Update documentation to use t2.micro, provide Classic options

catsby:
This updates our Getting Started guide to use `t2.micro` instances and us-east-1 compatible AMIs for most users who are not on EC2 Classic. It also provides an alternative AMI that works for `t1.micro`.

On `Building`:
screen shot 2016-05-20 at 4 17 02 pm

On `Change`:
screen shot 2016-05-20 at 1 12 21 pm

Fixes: 
- https://github.com/hashicorp/terraform/pull/6793


stack72:
@catsby :) Yay! Nice work

PR #6805, opened 2016-05-20
provider/aws: Add aws_availability_zones source

jen20:
This pull request adds a data source with a single list, `available`, in the schema. The list is populated with the availability zones to which an account has access within the region for which the AWS provider is configured. The list is sorted alphabetically so should remain in a predictable order.

### DIFF
--- a/builtin/providers/aws/data_source_availability_zones.go
+++ b/builtin/providers/aws/data_source_availability_zones.go
@@ -1,0 +1,52 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"sort"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsAvailabilityZones() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsAvailabilityZonesRead,
+
+		Schema: map[string]*schema.Schema{
+			"instance": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func dataSourceAwsAvailabilityZonesRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	log.Printf("[DEBUG] Reading availability zones")
+	d.SetId(time.Now().UTC().String())
+
+	req := &ec2.DescribeAvailabilityZonesInput{DryRun: aws.Bool(false)}
+	azresp, err := conn.DescribeAvailabilityZones(req)
+	if err != nil {
+		return fmt.Errorf("Error listing availability zones: %s", err)
+	}
+
+	raw := make([]string, len(azresp.AvailabilityZones))
+	for i, v := range azresp.AvailabilityZones {
+		raw[i] = *v.ZoneName
+	}
+
+	sort.Strings(raw)
+
+	if err := d.Set("instance", raw); err != nil {
+		return fmt.Errorf("[WARN] Error setting availability zones")
+	}
+
+	return nil
+}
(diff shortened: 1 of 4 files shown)


builtin/providers/aws/data_source_availability_zones.go
```diff
@@ -0,0 +1,52 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceAwsAvailabilityZonesRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	log.Printf("[DEBUG] Reading availability zones")
+	d.SetId(time.Now().UTC().String())
```
stack72: why use the time as the Id? Is it just to set something?

builtin/providers/aws/data_source_availability_zones.go
```diff
@@ -0,0 +1,52 @@
...
+
+	req := &ec2.DescribeAvailabilityZonesInput{DryRun: aws.Bool(false)}
+	azresp, err := conn.DescribeAvailabilityZones(req)
+	if err != nil {
+		return fmt.Errorf("Error listing availability zones: %s", err)
+	}
+
+	raw := make([]string, len(azresp.AvailabilityZones))
```
stack72: Will we get these back in the same order each time? Or should we sort them?

website/source/layouts/aws.erb
```diff
@@ -10,6 +10,15 @@
                     <a href="/docs/providers/aws/index.html">AWS Provider</a>
                 </li>
 
+                <li<%= sidebar_current(/^docs-aws-datasource/) %>
```
stack72: I think the AWS docs page will have to be redesigned - the more data sources we add, the longer that left hand nav is going to be :)

This is obviously a good thing though

stack72:
Looks awesome! Cannot wait to use it :) I currently use this: https://github.com/terraform-community-modules/tf_aws_availability_zones

stack72:
@jen20 so, I'm *REALLY* impatient and I put this through some tests :)

```
data "aws_availability_zones" "zones" {}

output "azs" {
  value = "${data.aws_availability_zones.zones.available}"
}
```

I got the following:

```

Outputs:

azs = [
  us-west-2a
  us-west-2b
  us-west-2c
]
```

apparentlymart:
In a moment of brief forgetfulness I re-implemented this as part of #6819... so if this gets merged first I'll go back and tear that out of there.

@stack72 that issue you're seeing with "unknown resource type" is one that @jen20 saw first and I saw over the weekend too, on many different data sources. It seems like a bug in the core handling of data sources during destroy, but I didn't track it down yet.

jen20:
I think we've decided on a "convention" of naming the list on singleton resources such as this `instance`. So it should read `${data.aws_availability_zones.instance}`. I'll update the PR and merge this. I can't reproduce the core bug with all of @apparentlymart's patches applied, although I could reliably reproduce it previously, so I think we can consider this fixed unless we come up with a new repro (on a new issue!). I'll update this PR now to line up with the new convention.

builtin/providers/aws/data_source_availability_zones.go
jen20: This is just the convention set by @apparentlymart. It's as good as anything else :-)

website/source/layouts/aws.erb
jen20: Yes, the long navigation bar is clearly unsustainable for larger providers. I guess that will be looked at soonish.

phinze:
Tested and it Just Works ™️ with multi-provider support:

```tf
provider "aws" {
  alias  = "east1"
  region = "us-east-1"
}

provider "aws" {
  alias  = "west2"
  region = "us-west-2"
}

data "aws_availability_zones" "east" {
  provider = "aws.east1"
}

data "aws_availability_zones" "west" {
  provider = "aws.west2"
}

output "east" {
  value = "${data.aws_availability_zones.east.instance}"
}

output "west" {
  value = "${data.aws_availability_zones.west.instance}"
}
```

zsh

COOL! 😍 

LGTM

PR #6816, opened 2016-05-21
core: context tests for data resources

apparentlymart:
Two context tests for data source lifecycle. Not 100% comprehensive but better than the zero tests from before.

Earlier we had a bug where data resources would not yet removed from the state during a destroy. This was fixed in cd0c4522eed540baf40e13e7c7fdb1b2195ecb2a, and the former of these tests will hopefully make sure it stays fixed.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -340,6 +340,59 @@ func TestContext2Apply_destroyComputed(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_destroyData(t *testing.T) {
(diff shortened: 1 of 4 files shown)


jen20:
Thanks @apparentlymart! This LGTM. Context tests cover the majority of the "internal" surface area of Terraform, so are pretty useful in cases like this.

jen20:
These commits landed as b832fb3 and ed9b8f9 - not sure how I managed to rebase them instead of merge with no fast forward...

PR #6817, opened 2016-05-21
core: restore data resource creation diffs

apparentlymart:
cd0c4522eed540baf40e13e7c7fdb1b2195ecb2a contained a bug where the creation diff for a data resource was put into a new local variable within the else block rather than into the diff variable in the parent scope, causing a null diff to always be produced.

This restores the expected behavior: a computed data resource appears in the diff, so it can then be fetched during the apply walk.

It also introduces a context test so this is less likely to get broken again.

### DIFF
--- a/terraform/eval_read_data.go
+++ b/terraform/eval_read_data.go
@@ -38,7 +38,8 @@ func (n *EvalReadDataDiff) Eval(ctx EvalContext) (interface{}, error) {
 		provider := *n.Provider
 		config := *n.Config
 
-		diff, err := provider.ReadDataDiff(n.Info, config)
+		var err error
+		diff, err = provider.ReadDataDiff(n.Info, config)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


apparentlymart:
@jen20 this fixes a regression caused by that other patch of mine you merged yesterday.

Thanks for giving me the pointer to the concept of context tests... I wasn't aware of those before, and that gave me the missing thing I needed to test and catch this sort of bug.

terraform/eval_read_data.go
```diff
@@ -38,7 +38,8 @@ func (n *EvalReadDataDiff) Eval(ctx EvalContext) (interface{}, error) {
 		provider := *n.Provider
 		config := *n.Config
 
-		diff, err := provider.ReadDataDiff(n.Info, config)
```
jen20: Good catch. This is annoying feature of Go that the GitHub level of context on diffs don't really help with!

PR #6818, opened 2016-05-21
helper/resource: ignore data resource diffs during destroy

### LINKED ISSUE #6713: `data` blocks don't work within the acceptance test harness
The acceptance test harness always expects the diff to be empty after apply, but data sources will *always* generate a diff because they are always eager to refresh themselves when given the opportunity.

This means that right now `data` blocks can't be used in acceptance tests. I missed this initially because in porting `terraform_remote_state` I neglected to update its test to use `data` block instead of a `resource` block.

We need some sort of exception to allow this to work as expected.

apparentlymart:
When testing destroy, the test harness calls Refresh followed by Plan, with the expectation that the resulting diff will be empty.

Data resources challenge this expectation, because they will always be instantiated during refresh if their configuration isn't computed, and so the subsequent diff will want to destroy what was instantiated.

To work around this, we make an exception that data resource destroy diffs may appear in the plan but nothing else.

This fixes #6713.

### DIFF
--- a/helper/resource/testing_config.go
+++ b/helper/resource/testing_config.go
@@ -3,6 +3,7 @@ package resource
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform/terraform"
 )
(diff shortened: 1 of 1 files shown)


jen20:
This LGTM for now - it's probably better to bake in the knowledge that we shouldn't care about data resources, and this implementation give us the ability to detect stray "other" resources (however they might come about). :+1:

PR #6819, opened 2016-05-22
provider/aws: data sources for AWS network planning

apparentlymart:
This is a bit of a grab-bag of AWS provider data sources that are aimed at a particular use-case: building and using VPCs and Subnets easily.

Some of the data sources in this set are not directly towards that goal, but enable it.

Here we have:

* `aws_subnet`
* `aws_vpc`
* `aws_region`
* `aws_availability_zone` and `aws_availability_zones`

Here is a simple example showing how a module can take a subnet id and automatically find the associated VPC:

```hcl
variable "aws_subnet_id" {
    description = "Id of the subnet where the EC2 instance will be created"
}

data "aws_subnet" "target" {
    id = "${var.aws_subnet_id}"
}

resource "aws_security_group" "main" {
    vpc_id = "${data.aws_subnet.target.vpc_id}"

    ingress {
        from_port = 22
        to_port = 22
        protocol = "tcp"
        cidr_blocks = ["${var.cidr_block}"]
    }

    # ...
}

resource "aws_instance" "main" {
    # ...

    vpc_security_group_ids = ["${aws_security_group.main.id}"]
    subnet_id = "${var.aws_subnet_id}"
}

output "aws_instance_id" {
    value = "${aws_instance.main.id}"
}
```

This is an initial implementation of [some conventions I came up with for mapping EC2 `Describe` actions to data sources](https://gist.github.com/apparentlymart/e62c81db609eca1fe08dc8226dd48485). It includes `ec2_filters.go`, which is a re-usable helper function for easily implementing EC2 data sources that fit these conventions.

### DIFF
--- a/builtin/providers/aws/data_source_aws_availability_zone.go
+++ b/builtin/providers/aws/data_source_aws_availability_zone.go
@@ -1,0 +1,89 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsAvailabilityZone() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsAvailabilityZoneRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"name_suffix": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"state": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsAvailabilityZoneRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	req := &ec2.DescribeAvailabilityZonesInput{}
+
+	if name := d.Get("name"); name != "" {
+		req.ZoneNames = []*string{aws.String(name.(string))}
+	}
+
+	req.Filters = buildEC2AttributeFilterList(
+		map[string]string{
+			"state": d.Get("state").(string),
+		},
+	)
+	if len(req.Filters) == 0 {
+		// Don't send an empty filters list; the EC2 API won't accept it.
+		req.Filters = nil
+	}
+
+	log.Printf("[DEBUG] DescribeAvailabilityZones %s\n", req)
+	resp, err := conn.DescribeAvailabilityZones(req)
+	if err != nil {
+		return err
+	}
+	if resp == nil || len(resp.AvailabilityZones) == 0 {
+		return fmt.Errorf("no matching AZ found")
+	}
+	if len(resp.AvailabilityZones) > 1 {
+		return fmt.Errorf("multiple AZs matched; use additional constraints to reduce matches to a single AZ")
+	}
+
+	az := resp.AvailabilityZones[0]
+
+	// As a convenience when working with AZs generically, we expose
+	// the AZ suffix alone, without the region name.
+	// This can be used e.g. to create lookup tables by AZ letter that
+	// work regardless of region.
+	nameSuffix := (*az.ZoneName)[len(*az.RegionName):]
+
+	d.SetId(*az.ZoneName)
+	d.Set("id", az.ZoneName)
+	d.Set("name", az.ZoneName)
+	d.Set("name_suffix", nameSuffix)
+	d.Set("region", az.RegionName)
+	d.Set("state", az.State)
+
+	return nil
+}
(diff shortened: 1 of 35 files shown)


apparentlymart:
Note to self: once this is more complete, consider doing a little bit of rework of the code from #6911 to use the general `ec2_filters.go` function added here, rather than doing its own tag/filter handling.

apparentlymart:
A different patch introduced `aws_availability_zones` so this needs to be reworked to apply to that rather than introducing its own.

Since this is just some new data sources and doc updates I'm going to hold this until after 0.7 is out and try to land it for the first patch release after that.

apparentlymart:
Since 0.7 was delayed, maybe this can get in before that after all... I'm going to optimistically add the release-0.7 tag here but happy to remove it if this patch requires more iterations, rather than having those iterations block the release.

jen20:
@apparentlymart For now I'm going to remove `release-0.7` from this, and tentatively put it in for a 0.7.1 - seem reasonable?

apparentlymart:
Totally fine @jen20, and sorry for the delayed reply.

stack72:
Hi @apparentlymart 

Can you rebase this? :)

P.

apparentlymart:
@stack72... rebased as requested. :grinning:

stack72:
LOL @apparentlymart we need another rebase please :)

apparentlymart:
Rebased again. These conflicts have just been other changes to the data source list in the provider, so don't actually affect the implementations of these data sources... if someone could review the main implementation then I can resolve any minor conflicts myself before merging. :grinning:

builtin/providers/aws/data_source_aws_subnet_test.go
```diff
@@ -0,0 +1,125 @@
...
+				attr["vpc_id"],
+				vpcRs.Primary.Attributes["id"],
+			)
+		}
+
+		if attr["cidr_block"] != "172.16.123.0/24" {
+			return fmt.Errorf("bad cidr_block %s", attr["cidr_block"])
+		}
+		if attr["availability_zone"] != "us-west-2a" {
+			return fmt.Errorf("bad cidr_block %s", attr["cidr_block"])
```
apparentlymart: this should be printing out `availablility_zone`, not `cidr_block`.

builtin/providers/aws/data_source_aws_availability_zone.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if name := d.Get("name"); name != "" {
+		req.ZoneNames = []*string{aws.String(name.(string))}
+	}
+
+	req.Filters = buildEC2FilterList(d, map[string]string{
+		"state": "state",
+	})
+
+	log.Printf("[DEBUG] DescribeAvailabilityZones %#v\n", req)
```
catsby: General _going forward_ nit, all* of the AWS structs implement GoStringer, so you can use `%s` to output them to string instead of `%#v`


\* probably

catsby:
Overall this looks good to me, however...

I don’t care for the `buildEC2FilterList` method. It strikes me as magical and maybe _does too much_. You give it an `attr` map to map keys/values (which I feel the key/values are backwards, but I digress…) and you give it `meta` to get them from (itself essentially a black box of _everything_).  But If you happen to have a `filter` block in that black box, or a `tags` block, you’ll magically get filters from them, even if you don't want to. 

The method itself is also undocumented, so how/when I would use this and how/when I would not isn't so clear, and there are no tests for it either. Is it only applicable for data sources? Or is it meant to start being used throughout the AWS provider?  

I feel like a better method would be one that just accepts a `map[string]interface{}` of `filter_name` and `value`:

```golang
req.Filter = buildEC2FilterList(map[string]interface{}{
  "cidr": d.Get("cidr").(string),
  "tags": d.Get("tags"),
})
```

It requires more explicitness (read: work) from the person using it, but I think that's good and as a result you get more clarity as to what's happening. 

This would need some special tweaking for the `filters` block if we had one, maybe a special/reserved key string in the map. 

I hope this feedback doesn't come off too much as a bikeshed argument, as the vast majority of this PR looks great! 

Thoughts?

zms:
@stack72: any chance we can have this in 0.7.5?

apparentlymart:
@zms I have a bit more work to do first, to respond to the review feedback... I hope to get to this soon.

apparentlymart:
Hi @catsby! Thanks for the review and sorry for the delay in replying.

I have reorganized what was originally the `buildEC2FilterList` function into three smaller functions, each dealing with a different part of this problem:

* `buildEC2AttributeFilterList` produces a list of filters from a map from EC2 filter keys to single values. It's now the caller's responsibility to stringify the non-string attributes in the way expected by the AWS API, and it doesn't know anything about `schema.ResourceData`.
* `buildEC2TagFilterList` produces a list of filters from a list of `ec2.Tag` objects.
* `buildEC2CustomFilterList` takes a value conforming to the schema produced by `ec2CustomFiltersSchema` and uses it to produce a list of filters based directly on the user's provided values.

I loosely modeled this after how EC2 tags are handled in `tags.go`, and indeed make use of the `tagsForMap` function from there in order to extract the tag values from the `ResourceData` to pass into `buildEC2TagFilterList`, rather than re-inventing that wheel as I was before.

All three of these now have basic unit tests and hopefully-comprehensive-enough documentation.

Hopefully this new structure is easier to follow and makes the individual data source implementations more readable.

imduffy15:
I'm fairly sure the answer is no but.... is it possible to utilize this and the aws_vpc resource to ensure the default VPC is deleted in an account?

PR #6820, opened 2016-05-22
Add explanation for type on Fastly header block

premist:
This PR adds additional explanation for 'type' attribute on Fastly header block.


stack72:
Hi @premist 

thanks for the PR here. I am going to make 1 slight modifcation, the values allowed are: `request`, `fetch`, `cache` & `response`

Thanks

Paul

premist:
@stack72 Thanks for the correction!

PR #2479, opened 2015-06-25
terraform: fail early if module validation fails

mitchellh:
Fixes #1861 

This just bails earlier if we have module validation errors. There isn't a test for this because its mostly a UX change. All errors that are caught at this stage are also caught at the graph stage, but the errors are nicer UX earlier if we bail.

PR #6833, opened 2016-05-23
Interpolate variable during Input and Validate

### LINKED ISSUE #5322: Syntax error interpolating count attribute when value passed between modules
From a [report on the Terraform mailing list](https://groups.google.com/d/msgid/terraform-tool/e0790b65-d599-4f20-b81c-fe13db55e0c6%40googlegroups.com).

## Repro config

```tf
# main.tf
module "source" {
  source = "./source"
}
module "child" {
  source  = "./child"
  list = "${module.source.list}"
}

# source/main.tf
output "list" { value = "foo,bar,baz" }

# child/main.tf
variable "list" { }
resource "template_file" "temp" {
  count     = "${length(split(",", var.list))}"
  template  = "foo"
}
```

## Steps

Run `terraform plan`.

## Expected behavior

A plan to create 3 templates.

## Observed behavior

```

* strconv.ParseInt: parsing "${length(split(\",\", var.list))}": invalid syntax
```

Related issues:

 * #1497 
 * #3884 
 * #4738 

This might actually be a dupe of one of those, but I wanted to get this repro out there. :+1:

jbardin:
fixes #5322

### DIFF
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -130,16 +130,6 @@ func (i *Interpolater) valueModuleVar(
 	v *config.ModuleVariable,
 	result map[string]ast.Variable) error {
 
-	// If we're computing all dynamic fields, then module vars count
-	// and we mark it as computed.
-	if i.Operation == walkValidate {
-		result[n] = ast.Variable{
-			Value: config.UnknownVariableValue,
-			Type:  ast.TypeString,
-		}
-		return nil
-	}
-
 	// Build the path to the child module we want
 	path := make([]string, len(scope.Path), len(scope.Path)+1)
 	copy(path, scope.Path)
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -129,6 +129,7 @@ func (i *Interpolater) valueModuleVar(
 	n string,
 	v *config.ModuleVariable,
 	result map[string]ast.Variable) error {
+
 	// If we're computing all dynamic fields, then module vars count
 	// and we mark it as computed.
 	if i.Operation == walkValidate {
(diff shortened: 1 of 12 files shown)


jen20:
LGTM @jbardin! :tada:

terraform/interpolate.go
```diff
@@ -129,15 +129,6 @@ func (i *Interpolater) valueModuleVar(
...
 	result map[string]ast.Variable) error {
-	// If we're computing all dynamic fields, then module vars count
-	// and we mark it as computed.
-	if i.Operation == walkValidate {
-		result[n] = ast.Variable{
-			Value: config.UnknownVariableValue,
-			Type:  ast.TypeString,
-		}
-		return nil
-	}
```
phinze: What I'd do here is walk back to the PR / commit where this stanza was introduced and determine what the intent was just to double check that we're not reversing some prior decision.

The test suite should cover us, so the fact that everything is passing is probably enough, but I think for such a subtle area of the code base it's worth it to do extra due diligence.

Let me know what the git archeology turns up!

phinze:
We hashed this out in Slack, I submitted one extra test case I was concerned about. If you pull that into your branch this LGTM for merge! 🚢

terraform/interpolate.go
jbardin: This behavior was introduced in PR #432.
The latter test still exists, and the first test was removed via PR #2479.

ddegoede:
@jbardin @phinze , should this be fixed in 0.7.0rc2? I think I am running into this issue, and if I replace the parameter with the actual contents of the module output, it works fine.

jbardin:
@ddegoede, yes that patch should be in rc2. I wonder if this is somehow a different manifestation of the same problem. 

Can you provide a standalone example that reproduces this in a new issue? Thanks!

ddegoede:
@jbardin I created an new issue for my error with TF plan and modules

PR #6836, opened 2016-05-23
Add aws elasticache replication group

lian:
this continues @stack72's `Initial scaffolding for Elasticache Replication Groups` commit.

done: rebased, fixed api changes, fixed create timeouts and states, added basic read func.
missing: complete read func.

aws's response to `DescribeReplicationGroups` is a little weird in terms of detail. seems most of the terraform configured params are passed down to the actual CacheCluster resources. a complete read func might need to enum the replicationgroup's cacheclusters and read state from there to match/refresh against terraform defined state.

opening pr to get more feedback on what read func should and shouldn't read.

### DIFF
--- a/builtin/providers/aws/resource_aws_elasticache_cluster.go
+++ b/builtin/providers/aws/resource_aws_elasticache_cluster.go
@@ -192,6 +192,12 @@ func resourceAwsElasticacheCluster() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"replication_group_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -268,6 +274,10 @@ func resourceAwsElasticacheClusterCreate(d *schema.ResourceData, meta interface{
 		req.PreferredAvailabilityZones = azs
 	}
 
+	if v, ok := d.GetOk("replication_group_id"); ok {
+		req.ReplicationGroupId = aws.String(v.(string))
+	}
+
 	resp, err := conn.CreateCacheCluster(req)
 	if err != nil {
 		return fmt.Errorf("Error creating Elasticache: %s", err)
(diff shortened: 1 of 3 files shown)


dankrieger:
I was working out finishing off the read function recently as well: https://github.com/dankrieger/terraform/commits/aws-elasticache-replication_groups

I'm not a go developer so I'm not confident in my ability to complete a valid pull request.  I'm not sure what a complete read function requires, but I felt like I got pretty close to a reasonable implementation.

lian:
@dankrieger nice. yea the details on what read function should read is also my problem. for example in your branch cache_node_groups doesn't have to be a list, because there is always just one element in that list (at least api docs state so). i like your getClustersToRemove and createCluster methods. if you don't mind i will combine our two branches and push them into this PR.

dankrieger:
@lian Yeah, I definitely struggled with figuring out the read method.  I have a feeling whatever ends up being a "good enough" representation of the read function will be draft one, and it can be expanded later when necessary.  Feel free to take whatever you want from my code, I just want to see this functionality approved and in the mainline.

samber:
@lian @dankrieger nice work both !

In fact, ```createCacheCluster``` and ```deleteCacheCluster``` seems the right way to update members.

Last Tuesday I also worked on my side on improving @dankrieger branch. I mostly focused my work on:
- Making AwsElasticacheCluster and AwsElasticacheReplicationGroup resources more consistant with closer schemas
- Large refactorization of ```cacheClusterReplicationGroupStateRefreshFunc```, double-checking the state of clusters in ```cacheClusterReplicationClusterStateRefresh``` and verifying the number of nodes up.
- ```MaxItems = 1``` on ```cache_node_group``` and refactorization of ```setCacheNodeGroupsData``` because we cannot have more than one NodeGroup in a replication group.

I was still testing that when I stopped working on this last Tuesday. It's not perfect, but maybe it can help you.

samber:
@lian What is the status on this ? 🙏

lian:
@samber hope if find some time this week :| otherwise push what you have already

nanoz:
This feature would be so helpful in a HA setup 🙏

samber:
To move forward with this PR, I will try to make some progresses on the feature this week.

builtin/providers/aws/resource_aws_elasticache_cluster.go
```diff
@@ -192,6 +192,12 @@ func resourceAwsElasticacheCluster() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"replication_group_id": &schema.Schema{
```
samber: In this PR, the ElasticacheReplicationGroup don't use the ElasticacheCluster resource. Is there any reason to do that ?!
I might miss something ;-)

stack72:
@lian / @samber as you both have made much better progress than me (I have not been able to get back to my PR), I have closed my PR. It would be awesome if either of you would be able to let me know how you are getting on?

P

someara:
Totally rooting for this feature.
I believe in you.
You can do this.

-s

stack72:
Hi @lian 

I finally got back to the work i was doing on this and opened the complete PR in #8275 - this includes a common schema for ElastiCache Cluster and Replication Group as they overlapped a lot!

Thanks for the work you put in here

Paul

stack72:
Closed via #8275

PR #6844, opened 2016-05-24
provider/openstack: Rebuild On Network Changes

### LINKED ISSUE #6740: provider/openstack Adding fixed_ip_v4 doesnt work
1. Create an instance without specifying fixed ip v4
2. Add fixed_ip_v4
3. Run apply

Result:

Each run terraform will show that ip address is added, but actually not. So each run it will show `  network.0.fixed_ip_v4: "192.168.0.41" => "192.168.0.21"` without every actually modifying it.

jtopjian:
This commit makes it so that openstack_compute_instance_v2 resources
are recreated when any network setting (except Floating IPs) is
changed.

Fixes #6740


Fodoj:
Does  changing port really requires recreation?

jtopjian:
I would think so. In similar ways that changing the network name/uuid and fixed IPs will cause a rebuild, then a port ID is very similar.

PR #6848, opened 2016-05-24
core: Keep errors/warning out of stdout for ease of automation

Bowbaq:
Some warnings about internal plugins were cluttering the output and confusing scripts depending on that output. This seems to keep the warnings & such out of `stdout`


phinze:
Ah yes this is a good call!

And the 0.7 release will be a good time to change this behavior.

Pulling this in and noting it in the changelog 👍

Issue #6767, opened 2016-05-19
Invalid memory address or nil pointer dereference


TeaBough:
### Terraform Version
Terraform v0.6.16

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_network_interface

### Terraform Configuration Files
```hcl
resource "aws_network_interface" "eni-53a81d09" {
  subnet_id         = "subnet-89d9f1cf"
  private_ips       = ["10.0.0.239"]
  security_groups   = []
  source_dest_check = false
  description = "Primary network interface"
}
```

### Panic Output
https://gist.githubusercontent.com/TeaBough/8bbe93352826906ef3750e0a46f8a639/raw/9f6e51edf5ca48cd001fae63dfa548757e8c93fc/crash.log

### Expected Behavior
Should have detect that the network interfaces are already created

### Actual Behavior
Terraform crashed : 
```
2016/05/19 14:47:18 [DEBUG] terraform-provider-aws: panic: runtime error: invalid memory address or nil pointer dereference
```

### AWS State
This happens when the network-interface has no security-groups (on AWS). I'm using terraforming to fetch my current aws architecture. Before it was possible to have a network-interface with no security attached, now it's not. I think that's where the problem is.

TeaBough:
PR : https://github.com/hashicorp/terraform/pull/6851

catsby:
Thanks for reporting/fixing this!

ivanjaros:
Just got the same error with the official 0.6.16. Using Virtualbox with Ubuntu 16.4 :(
Any info on 0.6.17?

jen20:
Hi @ivanjaros! I just tried to repro this on `master` (slated for 0.7 release shortly) and saw no crash. We aren't anticipating a 0.6.17 release, but 0.7 will be out soon.

The config I used was:

```
resource "aws_vpc" "test" {
  cidr_block = "10.0.0.0/16"
}

resource "aws_subnet" "test" {
  cidr_block = "10.0.1.0/24"
  vpc_id     = "${aws_vpc.test.id}"
}

resource "aws_network_interface" "eni-53a81d09" {
  subnet_id         = "${aws_subnet.test.id}"
  private_ips       = ["10.0.1.239"]
  security_groups   = []
  source_dest_check = false
  description       = "Primary network interface"
}
```

ivanjaros:
```
panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #6851, opened 2016-05-24
Handled case when instanceId is absent in network interfaces

TeaBough:
fix : https://github.com/hashicorp/terraform/issues/6767


catsby:
Thanks!

PR #6856, opened 2016-05-24
Revert "core: Keep errors/warning out of stdout for ease of automation"

phinze:
Reverts hashicorp/terraform#6848

All errors are swallowed after this merge. Oops! 😊  

zsh

We'll have to revisit this w/ a bit more testing.


phinze:
cc @Bowbaq - my bad I should have pulled down the PR before merge. Guess we need some better test coverage on this particular thing as well!

Issue #6794, opened 2016-05-20
Update of a lambda function stored in unversioned S3 bucket does not work


alexef:
Saw that this should have already be fixed by #4770 and #5370, but I'm facing the following issue:

Changing the configuration from:

    resource "aws_lambda_function" "test" {
      s3_bucket = "example"
      s3_key = "test.zip"
      function_name = "testFunction"
    }

to

    resource "aws_lambda_function" "test" {
      s3_bucket = "example"
      s3_key = "test2.zip"
      function_name = "testFunction"
    }

raises the error:

    * aws_lambda_function.testFunction: Error modifying Lambda Function Code testFunction: InvalidParameter: 1 validation error(s) found.
    - minimum field size of 1, UpdateFunctionCodeInput.S3ObjectVersion.

Looking at the the `UpdateFunctionCodeInput` what I suppose it happens is that terraform checks the tfstate and expects that the `s3_object_version` field has at least one character, so for updates it becomes mandatory.

apparentlymart:
Thanks for reporting this @alexef!

I can confirm this bug as I too encountered it a few days ago and hadn't got around to opening an issue for it yet.

I think the fix here is actually quite simple: during Create, [we set the version field only if it's non-empty in the config](https://github.com/hashicorp/terraform/blob/f6a21e789954a4d9e770ba9b5378b117f3d0721a/builtin/providers/aws/resource_aws_lambda_function.go#L155:L157), so it gets sent as `null` to AWS. But during `Update`, [we set it unconditionally](https://github.com/hashicorp/terraform/blob/f6a21e789954a4d9e770ba9b5378b117f3d0721a/builtin/providers/aws/resource_aws_lambda_function.go#L311), causing *the empty string* to be sent when it's empty, which the AWS API implementation then complains about.

I think updating the update implementation to look like the create one will fix this. I'll probably get to this eventually if no-one else does, but if someone else happens upon this while looking for a "first Terraform change" this is probably a good, straightforward one to start with and make sure there's an acceptance test for updating so we can catch any other issues where the Create/Update codepaths are different.

alexef:
Thank you @apparentlymart for making it seem so easy! Here comes my first PR.

I'm really not sure about the acceptance test you mentioned above - just copied and adapted an existing one.

ProTip:
This was tripping me up today as well on the latest release.

joncolby:
[...]

PR #6860, opened 2016-05-24
Fix: check updated s3 code for aws_lambda_function

alexef:
See #6794 for more details.

Disclaimer: this is my first Go PR, I don't have any prior experience with it. Ran `make` to test my changes.

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -318,17 +318,19 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
 		s3Key, keyOk := d.GetOk("s3_key")
 		s3ObjectVersion, versionOk := d.GetOk("s3_object_version")
 
-		if bucketOk && d.HasChange("s3_bucket") {
-			codeReq.S3Bucket = aws.String(s3Bucket.(string))
-			codeUpdate = true
-		}
-		if keyOk && d.HasChange("s3_key") {
-			codeReq.S3Key = aws.String(s3Key.(string))
-			codeUpdate = true
-		}
-		if versionOk && d.HasChange("s3_object_version") {
-			codeReq.S3ObjectVersion = aws.String(s3ObjectVersion.(string))
-			codeUpdate = true
+		if bucketOk && keyOk {
+			if d.HasChange("s3_bucket") || d.HasChange("s3_key") {
+				codeReq.S3Bucket = aws.String(s3Bucket.(string))
+				codeReq.S3Key = aws.String(s3Key.(string))
+				codeUpdate = true
+			}
+
+			if versionOk && d.HasChange("s3_object_version") {
+				codeReq.S3Bucket = aws.String(s3Bucket.(string))
+				codeReq.S3Key = aws.String(s3Key.(string))
+				codeReq.S3ObjectVersion = aws.String(s3ObjectVersion.(string))
+				codeUpdate = true
+			}
 		}
 	}
 
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -304,20 +304,32 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
 	}
 
 	codeUpdate := false
-	if d.HasChange("filename") || d.HasChange("source_code_hash") {
-		name := d.Get("filename").(string)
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_lambda_function.go
```diff
@@ -297,19 +297,34 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
...
+		}
+	} else {
+		s3Bucket, bucketOk := d.GetOk("s3_bucket")
+		s3Key, keyOk := d.GetOk("s3_key")
+		s3ObjectVersion, versionOk := d.GetOk("s3_object_version")
+
+		if bucketOk && keyOk {
+			if d.HasChange("s3_bucket") || d.HasChange("s3_key") {
+				codeReq.S3Bucket = aws.String(s3Bucket.(string))
+				codeReq.S3Key = aws.String(s3Key.(string))
```
apparentlymart: Do you think we should set these even if they haven't changed?

I'm thinking about the case where only `s3_object_version` has changed and `s3_bucket` and `s3_key` are still the same... in that case, I think we'd try to send a request with an empty `S3Bucket` and `S3Key`, right?

I think it might work for us to separate the "has it changed?" logic from the construction of the object. So as pseudo-code it would be like:

```go
if any of s3_bucket, s3_key, s3_object_version, filename or source_code_hash have changed {
    codeReq = &lambda.UpdateFunctionCodeInput{
        FunctionName: aws.String(d.Id()),
    }
    if filename is set {
        codeReq.ZipFile = loadFileContent(v.(string))
    }
    else {
        codeReq.S3Bucket = d.Get("s3_bucket")
        codeReq.S3Key = d.Get("s3_key")
        if version is set {
            codeReq.S3ObjectVersion = d.Get("s3_object_Version")
        }
    }
    call UpdateFunctionCode with codeReq, handle errors
    call setPartial for all the code-related attributes
}
```

Do you think that makes sense? I think this way we'll always send S3Bucket and S3Key if there have been *any* changes to S3 attributes.

alexef: Yes, you're right. IMO `s3_bucket` and `s3_key` must be set for every request, even though they haven't changed.

Your pseudocode is more readable than my code, I will update the PR to reflect it.

alexef:
@apparentlymart I've updated the code, but looking at it, shouldn't the next following lines be updated as well? (configUpdate).

I'll be afk for two weeks, please feel free to do anything to it.

apparentlymart:
Thanks @alexef! I'll take a closer look at this soon.

alexef:
I'm back. Let me know if I can help w/ anything :)

alexef:
Rebased against master.

apparentlymart:
Hi @alexef... sorry for the silence here. I've had to take a little break for Terraform work to deal with some "real work" at work but I do intend to come back and look at this again at some point, unless someone else beats me to it.

alexef:
No worries. I'll keep ping from time to time :)

apparentlymart:
Hi @alexef! Sorry for the delay here. I took a look at this today.

The logic here seems plausible, but when I tried to run the acceptance tests I ran into some issues:
* The new test is looking for a function called `tf_acc_lambda_name_s3`, but the function that's created is actually called `tf_acc_lambda_name_s3_unversioned`.
* After I tweaked the tests to address that, I got a different error that wasn't so easy to explain:

```
	testing.go:265: Step 1 error: Expected a non-empty plan, but got an empty plan!
```

It seems like something's amiss in the test steps that's causing no diff to be detected on the second step, but I wasn't able to spot it after trying a few things. Maybe you can spot it?

The other thing I noticed (while trying to debug the problem above) is that it seems that on the update step it creates a new S3 object `lambda-func-modified.zip` rather than simply updating the original one `lambda-func.zip` in place. Perhaps it would be better to preserve the key name here and test that we're able to trigger an update based on the `source_code_hash` argument.

Sorry again for the delay. Please let me know if anything isn't clear here!

alexef:
@apparentlymart I'm not sure how should I continue with this PR. At the time I wrote it, the change was trivial and the tests passed.

It's likely that the tests were broken when I rebased this PR from master. I'll have a second look at the code, but lacking the time/any experience with Go, I cannot guarantee I can fix it :(

apparentlymart:
Hi again @alexef! Thanks for taking the time to update this.

I found that the acceptance tests were not working due to some mismatches between function name in the test config vs the test checks, but the fix was easy and so I just fixed it and merged this rebased version of your changes at 7c4a3bf9678034f23eebabbb13fcd8534fbe0f62.

PR #6864, opened 2016-05-25
provider/openstack: Ensure CIDRs Are Lower Case

### LINKED ISSUE #6451: OpenStack Security Group IPv6 Capitalization
When using the OpenStack provider, if you specify an IPv6 cidr with capital letters, terraform will continually try to re-create the group (and fail) because the state returned and stored for the cidr is in lowercase. The provider should do a case-insensitive compare. 

### Terraform Version
Terraform v0.6.15

### Affected Resource(s)
- openstack_compute_secgroup_v2

```hcl
resource "openstack_compute_secgroup_v2" "test_uppercase" {
  name = "test_access_uppercase"
  description = "Securty Group Test - IPv6 Uppercase"
  rule {
    from_port = 61000
    to_port = 61000
    ip_protocol = "tcp"
    cidr = "2001:558:FC00::/39"
  }
}
```

### Debug Output
https://gist.github.com/lewg/da85fb63f7c8eb605a647f124a63125e

### Expected Behavior

The security group exists, there should be no modification. 

### Actual Behavior

It believes it needs to be changed, which then fails because that cidr already exists. 

### Steps to Reproduce
1. `terraform apply`
3. `terraform apply`

jtopjian:
This commit ensures that CIDR arguments are converted into lower
case values, specifically for IPv6 addresses.

Fixes #6451

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_secgroup_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_secgroup_v2.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/hashcode"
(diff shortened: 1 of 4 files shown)


stack72:
LGTM :) Nice simple fix

Issue #5710, opened 2016-03-18
Aurora updates


catsby:
Right now, most if not all of AWS RDS Cluster Instance attributes are `ForceNew`. Because they are essentially regular RDS Instances, we should be able to update them via `ModifyDBInstance`:

- http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBInstance.html

I believe this to be true, but haven't confirmed. We should investigate, since recreating a db instance because we changed the name sounds silly

stack72:
@catsby I believe we can close this now

The `aws_rds_cluster` resource now has an Updatefunc and it also has started having modification operations on it

Thoughts?

catsby:
@stack72 not quite; this ticket is referring to RDS Cluster Instance, which still is all `ForceNew`

catsby:
We've added some updates in #6865

Issue #6841, opened 2016-05-23
aws RDS (Aurora), instance parameter and changes


arcadiatea:
Terraform 0.7, Aurora RDS instance
Issue 1. The current terraform did not support parameter group at instance level. 
Issue 2. The current terraform did not have code to detected any changes at Aurora RDS Instance level to call a ModifyInstance action. The critical parameter lack of the ModifyInstance action is instance class change. Any change of db instance type will end with a destroy and rebuild. 

I have fixed code ready to submit later today.

catsby:
Hello –

> The current terraform also did not detect the changes of db instance type. The db instance type changes result in a resource destroy and rebuild.

I'm not sure I understand here. You mention it did not detect the change of db instance type (I assume `instance_class` config parameter), but you also mention the change resulted in a destroy/rebuild.. which suggests it was detected. Maybe I'm misunderstanding. Do you have a configuration that demonstrates this behavior?

arcadiatea:
@catsby 
I updated the issue definition.
The current terraform model be able to detect the change of instance class, but instead of do a ModifyInstance action, it go for the destroy and recreate route.  The current terraform has no feature on cluster instance update (resourceAwsRDSClusterInstanceUpdate). I will submit a pull request for just fix the instance class change and parameter group today. 
I am looking other feature which needed to be changed at Aurora Instance level and open different issues and pull requests

arcadiatea:
@catsby 
Current plan when instance class change.

`-/+ aws_rds_cluster_instance.www_instance
    cluster_identifier:   "prd00000111-linda2-acctmgr-aurora-www" => "prd00000111-linda2-acctmgr-aurora-www"
    db_subnet_group_name: "prd00000111.linda2.acctmgr-aurora" => "prd00000111.linda2.acctmgr-aurora"
    endpoint:             "prd00000111-linda2-acctmgr-aurora-www-0.cv5zmqt8xpne.us-east-1.rds.amazonaws.com" => "<computed>"
    identifier:           "prd00000111-linda2-acctmgr-aurora-www-0" => "prd00000111-linda2-acctmgr-aurora-www-0"
    instance_class:       "db.r3.xlarge" => "db.r3.2xlarge" (forces new resource)
    port:                 "3306" => "<computed>"
    publicly_accessible:  "false" => "false"
    tags.#:               "3" => "3"
    tags.Environment:     "linda2" => "linda2"
[...]

Issue #5436, opened 2016-03-02
AWS Aurora - Changing Instance size forces destruct - Doesn't recognise instance size change


boxrick:
AWS allows you to change the instance size "On the fly" without any downtime, this is also supported on a normal RDS instance on Terraform.

1) If you are using Cluster DB ( Aurora ), changing the instance size causes a destruction of the resource. This is obviously wrong and un-needed 

2) A similar issue with this, is if the instance size is changed manually Terraform does not detect and try to rectify the issue.

chrisfu:
+1. Terraform seems to want to destroy even when making even the most trivial of changes to a cluster... even when it comes to adding/removing CIDR from an associated security group.

bsiegel:
Seems like this will be addressed by #5710

stack72:
Fixed via #6865

PR #6865, opened 2016-05-25
support aurora instance's parameter group and modifyinstance

arcadiatea:
This change for #6841.

### DIFF
--- a/builtin/providers/aws/resource_aws_rds_cluster_instance.go
+++ b/builtin/providers/aws/resource_aws_rds_cluster_instance.go
@@ -71,11 +71,6 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
 				Computed: true,
 			},
 
-			"status": &schema.Schema{
-				Type:     schema.TypeString,
-				Computed: true,
-			},
-
 			// apply_immediately is used to determine when the update modifications
 			// take place.
 			// See http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Modifying.html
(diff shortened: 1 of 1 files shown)


catsby:
Hey @arcadiatea – am I correct that we can simply use the existing `aws_db_parameter_group`, correct?

builtin/providers/aws/resource_aws_rds_cluster_instance.go
```diff
@@ -58,13 +58,31 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
...
-				ForceNew: true,
+			},
+
+			"db_parameter_group_name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"status": &schema.Schema{
```
catsby: This attribute doesn't seem to be set anywhere, what is it meant to record? The `DBInstanceStatus`, or `status` from `StatusInfos.DBInstanceStatusInfo.N` ?

- http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBInstance.html

catsby:
I have a question about the `status` attribute but I did test out this branch and it worked as expected 👍

arcadiatea:
Hey, @catsby
This change is to allow Aurora instance be able to use customized parameter group set by aws_db_parameter_group.
The current issues is Aurora instance always created with default db parameter group and not be able to detect the modify requirement.

catsby:
@arcadiatea yep! I tried it out and it works as expected. I do have an question about the added `status` parameter; it's not being set, and it's not clear to me what value this should be. Do you recall?

arcadiatea:
@catsby  that status attribute is not be used for this modification , you are right.
I am working on another enhancement feature on aws_rds_cluster_instance (support security_group changes)  and I can include the usage of status there.

catsby:
Ok, thanks for clarifying! If you remove that attribute from this commit, I'll go ahead and merge this

arcadiatea:
@catsby status attributes is removed. This change probably also fix #5710, #5436

catsby:
Thanks! I'm going to merge this with an addition to the test file as well.

PR #6879, opened 2016-05-25
Add top-level ELB Attachment resource

### LINKED ISSUE #3999: provider/aws: Add top-level ELB Attachment resource
Based on an ML discussion:

https://groups.google.com/d/msgid/terraform-tool/e9ef8081-db96-4329-9b62-adf8c0b40c81%40googlegroups.com

It seems like it might be worth splitting out instance <-> ELB attachments as their own resource, in the way that we've done in several other scenarios now.

The use case from the thread is the desire to be able to create an ELB w/ instances and to have provisioners on the instances reference details about the ELB. You can't do that today w/o a cycle because the only way to attach instances to an ELB is via the `instances` parameter on `aws_elb`.

If we had a top level attachment resource you could do something like this:

```tf
# ==> Proposed config, does not work today <==

variable "count" { default = 3 }

resource "aws_elb" "app" {
  # no instances field specified
}

resource "aws_instance" "app" {
  count = "${var.count}"
  # ... provisioners here, reference ELB normally
}

# New proposed resource:
resource "aws_elb_attachment" "app" {
  count    = "${var.count}"
  elb      = "${aws_elb.app.id}"
  instance = "${element(aws_instance.app.*.id, count.index)}"
}
```

jbardin:
Add an aws_elb_attachment resource so that the attment of instances to
an ELB can be managed separately from an aws_elb and prevent dependency
cycles.

fixes #3999

### DIFF
--- a/builtin/providers/aws/resource_aws_elb_attachment.go
+++ b/builtin/providers/aws/resource_aws_elb_attachment.go
@@ -1,0 +1,121 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsElbAttachment() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsElbAttachmentCreate,
+		Read:   resourceAwsElbAttachmentRead,
+		Delete: resourceAwsElbAttachmentDelete,
+
+		Schema: map[string]*schema.Schema{
+			"elb": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"instance": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceAwsElbAttachmentCreate(d *schema.ResourceData, meta interface{}) error {
+	elbconn := meta.(*AWSClient).elbconn
+	elbName := d.Get("elb").(string)
+
+	instance := d.Get("instance").(string)
+
+	registerInstancesOpts := elb.RegisterInstancesWithLoadBalancerInput{
+		LoadBalancerName: aws.String(elbName),
+		Instances:        []*elb.Instance{{InstanceId: aws.String(instance)}},
+	}
+
+	log.Printf("[INFO] registering instance %s with ELB %s", instance, elbName)
+
+	_, err := elbconn.RegisterInstancesWithLoadBalancer(&registerInstancesOpts)
+	if err != nil {
+		return fmt.Errorf("Failure registering instances with ELB: %s", err)
+	}
+
+	d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", elbName)))
+
+	return nil
+}
+
+func resourceAwsElbAttachmentRead(d *schema.ResourceData, meta interface{}) error {
+	elbconn := meta.(*AWSClient).elbconn
+	elbName := d.Get("elb").(string)
+
+	// only add the instance that was previously defined for this resource
+	expected := d.Get("instance").(string)
+
+	// Retrieve the ELB properties to get a list of attachments
+	describeElbOpts := &elb.DescribeLoadBalancersInput{
+		LoadBalancerNames: []*string{aws.String(elbName)},
+	}
+
+	resp, err := elbconn.DescribeLoadBalancers(describeElbOpts)
+	if err != nil {
+		if isLoadBalancerNotFound(err) {
+			log.Printf("[ERROR] ELB %s not found", elbName)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error retrieving ELB: %s", err)
+	}
+	if len(resp.LoadBalancerDescriptions) != 1 {
+		log.Printf("[ERROR] Unable to find ELB: %s", resp.LoadBalancerDescriptions)
+		d.SetId("")
+		return nil
+	}
+
+	// only set the instance Id that this resource manages
+	found := false
+	for _, i := range resp.LoadBalancerDescriptions[0].Instances {
+		if expected == *i.InstanceId {
+			d.Set("instance", expected)
+			found = true
+		}
+	}
+
+	if !found {
+		log.Printf("[WARN] instance %s not found in elb attachments", expected)
+		d.SetId("")
+	}
+
+	return nil
+}
+
+func resourceAwsElbAttachmentDelete(d *schema.ResourceData, meta interface{}) error {
+	elbconn := meta.(*AWSClient).elbconn
+	elbName := d.Get("elb").(string)
+
+	instance := d.Get("instance").(string)
+
+	log.Printf("[INFO] Deleting Attachment %s from: %s", instance, elbName)
+
+	deRegisterInstancesOpts := elb.DeregisterInstancesFromLoadBalancerInput{
+		LoadBalancerName: aws.String(elbName),
+		Instances:        []*elb.Instance{{InstanceId: aws.String(instance)}},
+	}
+
+	_, err := elbconn.DeregisterInstancesFromLoadBalancer(&deRegisterInstancesOpts)
+	if err != nil {
+		return fmt.Errorf("Failure deregistering instances from ELB: %s", err)
+	}
+
+	return nil
+}
--- a/builtin/providers/aws/resource_aws_elb_attachment_test.go
+++ b/builtin/providers/aws/resource_aws_elb_attachment_test.go
@@ -1,0 +1,232 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSELBAttachment_basic(t *testing.T) {
+	var conf elb.LoadBalancerDescription
+
+	testCheckInstanceAttached := func(count int) resource.TestCheckFunc {
+		return func(*terraform.State) error {
+			if len(conf.Instances) != count {
+				return fmt.Errorf("instance count does not match")
+			}
+			return nil
+		}
+	}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_elb.bar",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckAWSELBDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSELBAttachmentConfig1,
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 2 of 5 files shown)


website/source/docs/providers/aws/r/elb.html.markdown
```diff
@@ -10,6 +10,12 @@ description: |-
 
 Provides an Elastic Load Balancer resource.
 
+~> **NOTE on ELB Instances and ELB Attachments:** Terraform currently
+provides both a standalone [ELB Attachment resource](elb_attachment.html)
+(describing the instances attached to an ELB), and an ELB resource with
+`instances` defined in-line. At this time you cannot use an ELB with in-line
+instaces in conjunction with an ELB Attachment resource. Doing so will cause a
+conflict of rule settings and will overwrite rules.
```
radeksimko: That said user still has to define at least 1 listener inside `aws_elb` as ELBs without listeners are disallowed on the API level 😢 

I'm wondering if it's better to implement the `aws_elb_attachment` as a data source that would be somehow folded into `aws_elb.dynamic_listeners`. I haven't looked into how the data sources work under the hood as yet, but reversing the dependency (ELB / listeners) here would IMO help as all listeners would be defined in a single place and generating diff/plan would be easier I think.

Also we could then mark `aws_elb.listener` as `ConflictsWith` `dynamic_listeners` which would prevent users from doing stupid things.

radeksimko: Forget it, I completely misunderstood the purpose of this PR!

radeksimko:
This LGTM 👍  with a few nitpicks:

1. Link to the new resource page in docs is missing in menu - see `website/source/layouts/aws.erb`
2. I'm thinking that not only `aws_elb` & `aws_elb_attachment` are mutually exclusive, but also multiple `aws_elb_attachment` resources for the same ELB would conflict with each other since `aws_elb_attachment` inevitably manages _all_ attached instances - shouldn't that be also mentioned in the docs?

radeksimko:
Also I reckon enhancement to this resource will be requested soon - support for ASG/ELB attachment, but I think that can be part of another PR.

jbardin:
@radeksimko thanks for the feedback!

Yes, we were discussing aws_elb_attachment resources additive rather than exclusive so that attachments could be defined in multiple places. 

Do we have the same dependency cycle problems with ASG attachments?  the ELB and and LC are both defined in the ASG, so I'm not sure if there's a clear benefit there.

radeksimko:
> Yes, we were discussing making aws_elb_attachment resources additive rather than exclusive so that attachments could be defined in multiple places.

radeksimko:
> Yes, we were discussing making aws_elb_attachment resources additive rather than exclusive so that attachments could be defined in multiple places.

I think that anyone wanting to do blue/green deployments in a way that each "colour" is managed in a separate Terraform context/directory may benefit from the _additive_ approach. 

I think it can be still a _list_ of instances to attach - i.e. `aws_elb_attachment` can still manage more than 1 relationship (more than 1 instance) - so the schema can remain the same, but it would only remove attachments it has created.

It also makes sense in the light of Terraform being nice to existing infrastructure outside of Terraform - imagine a situation where user had existing ELB and wanted to use TF only for managing the relationship EC2/ELB.

If you still believe that the _exclusive_ approach is better, then I think the resource may be renamed to `aws_elb_attachments` to make it clear that it's managing _all_ attachments.

> Do we have the same dependency cycle problems with ASG attachments?

What kind of dependency cycles do you mean? I'm not sure I follow.

jbardin:
Ok, this just got a refactor and simplification. After discussing this more with @phinze, we're going to leave this as a single instance per attachment resource, rather than a list. The aligns with other resources we've broken out in this manner, and works better within in the limitations of the terraform internals. We can work on refactoring them all when we have a way to effectively manage this type of interdependent set of resources. 

@radeksimko the resource cycle I was referring to was what's described in issue #3999, but that isn't the only reason to add other things as separate resources either.

builtin/providers/aws/resource_aws_elb_attachment.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	log.Printf("[INFO] registering instance %s with ELB %s", instance, elbName)
+
+	_, err := elbconn.RegisterInstancesWithLoadBalancer(&registerInstancesOpts)
+	if err != nil {
+		return fmt.Errorf("Failure registering instances with ELB: %s", err)
+	}
+
+	d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", elbName)))
```
radeksimko: Is there any reason why not just make this `elbName` + instance ID?

builtin/providers/aws/resource_aws_elb_attachment.go
```diff
@@ -0,0 +1,120 @@
...
+	if len(resp.LoadBalancerDescriptions) != 1 {
+		log.Printf("[ERROR] Unable to find ELB: %s", resp.LoadBalancerDescriptions)
+		d.SetId("")
+		return nil
+	}
+
+	// only set the instance Id that this resource manages
+	for _, i := range resp.LoadBalancerDescriptions[0].Instances {
+		if expected == *i.InstanceId {
+			d.Set("instance", expected)
```
radeksimko: What if this `d.Set` is never reached? i.e. what if someone detaches the instance from the ELB outside of Terraform?
The common pattern we use elsewhere for similar cases is to log `WARN` message + destroy the resource via `d.SetId("")`.

builtin/providers/aws/resource_aws_elb_attachment.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func resourceAwsElbAttachmentRead(d *schema.ResourceData, meta interface{}) error {
+	elbconn := meta.(*AWSClient).elbconn
+	elbName := d.Get("elb").(string)
+
+	// only add the instance that was previously defined for this resource
+	expected := d.Get("instance").(string)
+	if expected == "" {
+		return fmt.Errorf("no instance defined for ELB Attachment")
```
radeksimko: This should never happen as it is `Required` field or am I missing something obvious?

radeksimko:
> we're going to leave this as a single instance per attachment resource, rather than a list

ok, I didn't have a strong opinion about this as much as I had about exclusive/additive approach, so I think we're pretty much on the same page here. 👌 

I left you some other comments/questions in the code.

builtin/providers/aws/resource_aws_elb_attachment.go
jbardin: An instance attached to the same ELB could change to a new aws_elb_attachment resource, creating a new resource with the same ID (I'm not sure if that's a problem in practice, but was being cautious).

builtin/providers/aws/resource_aws_elb_attachment.go
jbardin: You're right. I removed that check because of a misunderstanding of the schema behavior, but I'll add it back with another acceptance test.

builtin/providers/aws/resource_aws_elb_attachment.go
jbardin: right again, just defensive programming gone awry.

builtin/providers/aws/resource_aws_elb_attachment.go
radeksimko: I see what do you mean, but both fields are `ForceNew: true` anyway, so it's going to happen _anyway_.

My only concern about this is that it may make future resource importing slightly tricker. I haven't looked at the import functionality yet though, so maybe I'm wrong - in which case feel free to ignore this concern.

builtin/providers/aws/resource_aws_elb_attachment.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsElbAttachment() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsElbAttachmentCreate,
+		Read:   resourceAwsElbAttachmentRead,
+		Delete: resourceAwsElbAttachmentDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
radeksimko: I don't think this will actually work. The passthrough function is only supposed to work for ID-only imports as far as I know (and was able to read from the code I skimmed through).

It would be probably better to omit `Importer` completely for now, until we figure out how to deal with import for such resources.

builtin/providers/aws/resource_aws_elb_attachment.go
radeksimko: The import won't work anyway since the lookup is not based on ID anyway, so it doesn't matter - see my other note above.

builtin/providers/aws/resource_aws_elb_attachment.go
jbardin: Yes, thanks for catching that, it was left from an earlier attempt. 

I think importing would have to rely on the ELB importing them as a single list, which I think is probably fine. Either way we don't need Importer here, and can sort it out later if there are other options.

builtin/providers/aws/resource_aws_elb_attachment_test.go
```diff
@@ -0,0 +1,232 @@
...
+	# us-west-2
+	ami = "ami-043a5034"
+	instance_type = "t1.micro"
+}
+
+resource "aws_elb_attachment" "foo1" {
+  elb      = "${aws_elb.bar.id}"
+  instance = "${aws_instance.foo1.id}"
+}
+`
```
phinze: Spacing nit - can you `hclfmt` these? Two space indent, aligned equals.

website/source/docs/providers/aws/r/elb.html.markdown
```diff
@@ -10,6 +10,12 @@ description: |-
 
 Provides an Elastic Load Balancer resource.
 
+~> **NOTE on ELB Instances and ELB Attachments:** Terraform currently
+provides both a standalone [ELB Attachment resource](elb_attachment.html)
+(describing an instance attached to an ELB), and an ELB resource with
+`instances` defined in-line. At this time you cannot use an ELB with in-line
+instaces in conjunction with a ELB Attachment resources. Doing so will cause a
+conflict of rule settings and will overwrite rules.
```
phinze: `s/conflict of rule settings/conflict and will overwrite attachments/`

website/source/docs/providers/aws/r/elb_attachment.html.markdown
```diff
@@ -0,0 +1,34 @@
...
+`instances` defined in-line. At this time you cannot use an ELB with in-line
+instaces in conjunction with an ELB Attachment resource. Doing so will cause a
+conflict of rule settings and will overwrite rules.
+## Example Usage
+
+```
+# Create a new load balancer attachment
+resource "aws_elb_attachment" "baz" {
+  elb      = "${aws_elb.bar.id}"
+  instances = ["${aws_instance.foo.id}"]
```
phinze: Switch to singular here.

phinze:
Looking great! A few final spacing / docs nits and this LGTM after those are addressed. 👍

jbardin:
```
$ TF_ACC=1 go test -v -run ELBAttachment
=== RUN   TestAccAWSELBAttachment_basic
--- PASS: TestAccAWSELBAttachment_basic (131.55s)
=== RUN   TestAccAWSELBAttachment_drift
--- PASS: TestAccAWSELBAttachment_drift (95.12s)
PASS
```

phinze:
LGTM

PR #3124, opened 2015-08-30
aws_iam_policy_document logical resource

apparentlymart:
A helper to construct IAM policy documents using familiar Terraform concepts. Makes Terraform-style interpolations easier and resolves the syntax conflict between Terraform interpolations and IAM policy variables by changing the latter to use ``&{...}`` for its interpolations.

Its use is completely optional and users are free to go on using literal heredocs, file interpolations or whatever else; this just adds another option that fits more naturally into a Terraform config.

* [x] Resource Implementation
* [x] Tests
* [x] Documentation

PR #4278, opened 2015-12-12
aws_s3_bucket - Fix policy normalization to match AWS API

vancluever:
There are a couple of situations I have encountered where a S3 bucket policy that is technically correct gets changed by AWS:

 * When a policy statement has no `Sid` specified, it will create a `Sid` entry with blank data.
 * When a policy statement only has one action, but is specified in an array, it will truncate the array and use a single element string action instead. (ie: `"Action": [ "s3:ListBucket" ]` becomes `"Action": "s3:ListBucket"`)
 * Multiple actions will be sorted alphabetically.

This PR expands on the current `normalizeJson` functionality that exists as a `StateFunc` hook and adds features to correct this:

 * `sortJSONArray` will recursively search for JSON arrays in the policy and sort them
 * `normalizePolicyStatements` sets the empty `Sid` when one does not exist and also truncates single-element `Action` entries.

Related to #3634, #3124, #4245, and possibly others, basically there seems to be a lot of other situations where this is an issue, each with possibly different quirks.

PR #6881, opened 2016-05-25
provider/aws: aws_iam_policy_document data source

jen20:
This brings over the work done by @apparentlymart and @radeksimko in PR #3124, and converts it into a data source for the AWS provider.

It adds a helper to construct IAM policy documents using familiar Terraform concepts. It makes Terraform-style interpolations easier and resolves the syntax conflict between Terraform interpolations and IAM policy variables by changing the latter to use &{...} for its interpolations.

Its use is completely optional and users are free to go on using literal heredocs, file interpolations or whatever else; this just adds another option that fits more naturally into a Terraform config.

Currently the tests are failing, which needs investigation.

### DIFF
--- a/website/source/docs/providers/aws/d/iam_policy_document.html.markdown
+++ b/website/source/docs/providers/aws/d/iam_policy_document.html.markdown
@@ -1,0 +1,141 @@
+---
+layout: "aws"
+page_title: "AWS: aws_iam_policy_document"
+sidebar_current: "docs-aws-resource-iam-policy-document"
+description: |-
+  Generates an IAM policy document in JSON format
+---
+
+# aws\_iam\_policy\_document
+
+Generates an IAM policy document in JSON format.
+
+This is a data source which can be used to construct a JSON representation of
+an IAM policy document, for use with resources which expect policy documents,
+such as the `aws_iam_policy` resource.
+
+```
+data "aws_iam_policy_document" "example" {
+    statement {
+        actions = [
+            "s3:ListAllMyBuckets",
+            "s3:GetBucketLocation",
+        ]
+        resources = [
+            "arn:aws:s3:::*",
+        ]
+    }
+
+    statement {
+        actions = [
+            "s3:ListBucket",
+        ]
+        resources = [
+            "arn:aws:s3:::${var.s3_bucket_name}",
+        ]
+        condition {
+            test = "StringLike"
+            variable = "s3:prefix"
+            values = [
+                "",
+                "home/",
+                "home/&{aws:username}/",
+            ]
+        }
+    }
+
+    statement {
+        actions = [
+            "s3:*",
+        ]
+        resources = [
+            "arn:aws:s3:::${var.s3_bucket_name}/home/&{aws:username}",
+            "arn:aws:s3:::${var.s3_bucket_name}/home/&{aws:username}/*",
+        ]
+    }
+
+}
+
+resource "aws_iam_policy" "example" {
+    name = "example_policy"
+    path = "/"
+    policy = "${data.aws_iam_policy.example.json}"
+}
+```
+
+Using this data source to generate policy documents is *optional*. It is also
+valid to use literal JSON strings within your configuration, or to use the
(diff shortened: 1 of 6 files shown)


website/source/docs/providers/aws/d/iam_policy_document.html.markdown
```diff
@@ -0,0 +1,145 @@
...
+# aws\_iam\_policy\_document
+
+Generates an IAM policy document in JSON format.
+
+This is a data source which can be used to construct a JSON representation of
+an IAM policy document, for use with resources which expect policy documents,
+such as the `aws_iam_policy` resource.
+
+```
+resource "aws_iam_policy_document" "example" {
```
apparentlymart: This should say `data` instead of `resource` now. :grinning:

apparentlymart:
Apart from that docs nit this LGTM, though I suppose I'm kinda reviewing my own code here. :grinning:

phinze:
Apart from the outstanding docs tweak and test issue, this PR LGTM!

jen20:
Thanks - I reviewed the code as I ported it and it looked fine. I'll push another commit that more accurately reflects the provenance of this work though since it's almost entirely unaltered from @apparentlymart's original PR! The docs have been updated - I think we should add a giant note at the top about the differing native interpolation syntax though.

jen20:
There is stil the issue of the failing acceptance test which I believe is exposing a core bug in destroy:

```
	testing.go:300: Error destroying resource! WARNING: Dangling resources

		Error: Error applying: 1 error(s) occurred:

		* data.aws_iam_policy_document.test: unknown resource type: aws_iam_policy_document

		State: <no state>
```

vancluever:
Just mentioned one thing in #4278 and I think I see one more too:

 * There needs to be some logic to ensure that array items like `Action`s get converted to `string`s instead of `[]string`s if there is only one element (this was one of the things that was causing recurring diffs in unchanged policies). So those elements probably need to be `interface{}` and then handled accordingly.
 * Looks like some elements might be misspelled - at least according to the AWS docs and what I've always used - I'll annotate the code.

Looking forward to this one too as it's one less branch I need to maintain on my side!

vancluever:
Just gave this a whirl, gist below has the diff from the S3 bucket policy:

https://gist.github.com/vancluever/526e9c1d8153436e58244847b8fb20f1

Few things probably need to be done here:

 * `omitempty` needs to be removed from `Sid`
 * We have to sort arrays, in reverse alphabetical order
 * Anything that takes a list value needs to be able to take a string value as well, which will be used in place of single element lists.

vancluever:
Hey guys, here's what I got now:

https://github.com/paybyphone/terraform/commit/c85d0ee5c50f2f23dfbf7784f5b4e35c2bca19c7

Still gotta work on fixing the tests, will probably grab a couple from the policy generator and toss them in.

vancluever:
Hmm, okay, the policy generator is generating policies that are not what the API would give back either. I'm just going to adjust the test, maybe if we get some people trying this out and send in any feedback if they get a diff that would be good.

vancluever:
Updated the commit with the updated test. Removed the blank element from `condition`, not too sure if that's a condition that we really need to handle, hopefully people won't be putting blank values in their document, however I'm not too sure if this might come up in interpolation?

phinze:
@jen20 If you drop the `[WIP]` and comment w/ the passing test this LGTM!

jen20:
Tests now passing:

```
make testacc TEST=./builtin/providers/aws TESTARGS="-run TestAccAWSIAMPolicyDocument"
2016/05/31 12:49:58 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run TestAccAWSIAMPolicyDocument -timeout 120m
=== RUN   TestAccAWSIAMPolicyDocument
--- PASS: TestAccAWSIAMPolicyDocument (6.40s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	6.416s
```

PR #6884, opened 2016-05-26
Allow specifying a default value to lookup()

### LINKED ISSUE #4474: Failed lookup() short-circuits interpolation
Consider this terraform code:

    variable "food" {}

    variable "type_lookup" {
        default = {
            apple = "fruit"
            banana = "fruit"
            tomato = "vegetable"
            beef = "meat"
            spam = ""
        }
    }

    output "food_type" {
        value = "${coalesce(lookup(var.type_lookup, var.food), "unknown")}"
    }

So, given an input variable `food`, I would expect the module provide an output `food_type` containing the lookup value from the `type_lookup` map, or to return `unknown` if the food is not specified in the map (or if the specified value is an empty string, as with the key `spam` in the example).

However, what happens is that if the key provided to the lookup function is not found, the output `food_type` is just not returned at all.

Examples:

    $ terraform apply
    var.food
      Enter a value: spam
    
    Outputs:
      food_type = unknown
    
    $ terraform apply
    var.food
      Enter a value: apple
    
    Outputs:
      food_type = fruit
    
    $ terraform apply
    var.food
      Enter a value: durian
    
    [no outputs]

In other words, a failed `lookup()` call appears to short-circuit the entire interpolation. This is not what I would expect. Further, even reordering the arguments to `coalesce`, or adding fixed text on either side of the interpolation, or even just using the `lookup()` call alone also fails silently in the same way. So eg, this additional output:

    output "second_try" { value = "hello: ${lookup(var.type_lookup, var.food)}" }

This results in no output value for `second_try` if the value of `var.food` is not a key in the `var.type_lookup` map.

IMO, `lookup()` should either return an empty string on a failed lookup or provide a way to specify some default value. It definitely should not short-circuit the entire output block.

daveadams:
This should fix #4474 in which lookup() calls fail and short-circuit the entire interpolation when the given key is not found in the map. This allows an optional third argument for providing a default value when the key does not exist in the map, which will allow better composition of lookup() with coalesce() et al.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -548,17 +548,32 @@ func interpolationFuncSplit() ast.Function {
 // dynamic lookups of map types within a Terraform configuration.
 func interpolationFuncLookup(vs map[string]ast.Variable) ast.Function {
 	return ast.Function{
-		ArgTypes:   []ast.Type{ast.TypeMap, ast.TypeString},
-		ReturnType: ast.TypeString,
+		ArgTypes:     []ast.Type{ast.TypeMap, ast.TypeString},
+		ReturnType:   ast.TypeString,
+		Variadic:     true,
+		VariadicType: ast.TypeString,
 		Callback: func(args []interface{}) (interface{}, error) {
+			defaultValue := ""
+			defaultValueSet := false
+			if len(args) > 2 {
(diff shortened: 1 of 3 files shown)


jen20:
Thanks @daveadams - this is a nice addition.

PR #6892, opened 2016-05-26
Adding a simple way to support authenticate registry.

IsmailMehdi:
This change will allow terraform to support the docker mechanism for authenticated private registry. It is simple in the sense that it does not require any changes to the manifest. The credentials are provided via the normal docker config.json authentication scheme.


kyhavlov:
👍 this is a very convenient way to pass registry credentials to terraform

stack72:
Hi @IsmailMehdi 

Thanks for the PR here. Please can you add an acceptance test on how we can test this works as expected?

Thanks

Paul

IsmailMehdi:
@stack72 Would love to add a test, unfortunately I am not sure how to establish a private registry with authentication enabled as part of the test. Any thoughts on that ?

IsmailMehdi:
@stack72 Any thoughts on this ?

PR #6895, opened 2016-05-27
core: Write resource mode to state

phinze:
Before, the resource mode was implicitly encoded into the
ResourceStateKey (if the key was prefixed with "data."). Here, we add an
explicit "mode" field to the state which makes it easier to interpret a
data source in the state.

### DIFF
--- a/terraform/eval_state.go
+++ b/terraform/eval_state.go
@@ -1,6 +1,10 @@
 package terraform
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/config"
+)
 
 // EvalReadState is an EvalNode implementation that reads the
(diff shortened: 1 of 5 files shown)


phinze:
@jen20 so I believe this is complete, but I'm having trouble finding a place where this actually makes a difference. The `transform_orphan` code is already baked around `ResourceStateKey` and I don't think it would make it simpler to look at mode there.

Trying to think through where else this implementation would help make things cleaner. Any ideas?

Barring that seems like it's just extra code and we shouldn't do it yet?

phinze:
Punting on this for now for the above reasons. If we come up with a good use case that makes adding this clearly beneficial we can pick it back up 👍

PR #6906, opened 2016-05-27
Fix mixing internal and external plugins

cbednarski:
- Only display the warning in the main process, which prevents the plugin handshake from failing
- Added `TF_PLUGIN_DEV` environment variable; if set the external plugin warning will not be displayed

/cc @stack72 @jen20


jen20:
:+1: Looks good to me!

PR #6907, opened 2016-05-27
core: No destroy nodes necessary for data sources

jen20:
We don't need to construct destroy nodes for data sources, as there is nothing to do beyond removing them from state. This patch checks the mode during expansion and bails early if we are dealing with a data source.

Prior to this patch, destruction of data sources would (usually) fail with an "unknown resource type x" error, as demonstrated in the tests of the aws_iam_policy_document resource.

### DIFF
--- a/terraform/graph_config_node_resource.go
+++ b/terraform/graph_config_node_resource.go
@@ -245,6 +245,10 @@ func (n *GraphNodeConfigResource) DestroyNode() GraphNodeDestroy {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


jen20:
This was discussed with @apparentlymart (who implemented the data sources feature), and demonstrably solves a real problem, so I'm going to merge it and let @phinze review retroactively :-)

jen20:
Changed my mind on this - the fix actually isn't correct. I'll come back to this shortly.

PR #6912, opened 2016-05-28
aws:Adding in retry to allow ec2/DescribeRouteTables consistency

### LINKED ISSUE #5335: Intermittent AWS Eventual Consistency Issues on CircleCI?
I have a terraform configuration whose job is to create a VPC.  The basic structure is this:

```
prod-vpc
- creates vpc
- adds vpc peering route with mgmt-vpc

stage-vpc
- creates vpc
- adds vpc peering route with mgmt-vpc

mgmt-vpc
- adds vpc peering routes with prod and stage

create-vpc-peering-connections
```

As you can see, there are many interdependencies, however `apply`ing and `destroy`ing this from my macbook consistently works fine.

When I run this same terraform configuration as part of a CircleCI build, however, I get intermittent failures.  Here are some of the errors I've gotten:

```
* aws_route.internet-gateway: 
error finding matching route for Route table (rtb-dff811b8) and destination CIDR block (0.0.0.0/0)
```

```
* aws_route.vpc-peering.1: 
error finding matching route for Route table (rtb-a39b72c4) and destination CIDR block (10.110.0.0/18)
```

Notice that sometimes I'm getting the same error, but for a different resource, though the second one occurred twice.  Here are some other errors.  Note that each build will fail with one of these, or in some rare cases, succeed.

```
4 error(s) occurred:

* aws_internet_gateway.main: InvalidInternetGatewayID.NotFound: The internetGateway ID 'igw-eab0118e' does not exist
    status code: 400, request id: 
* Resource 'aws_eip.nat' does not have attribute 'id' for variable 'aws_eip.nat.id'
* aws_route.vpc-peering.0: 
error finding matching route for Route table (rtb-eb99708c) and destination CIDR block (10.100.0.0/18)
* aws_route.vpc-peering.1: 
error finding matching route for Route table (rtb-eb99708c) and destination CIDR block (10.110.0.0/18)
```

```
1 error(s) occurred:

* aws_nat_gateway.nat.0: Error waiting for NAT Gateway (nat-0526389328c14da4e) to become available: unexpected state 'failed', wanted target '[available]'
```

```
4 error(s) occurred:

* aws_route.internet-gateway: 
error finding matching route for Route table (rtb-5ca64f3b) and destination CIDR block (0.0.0.0/0)
* Resource 'aws_eip.nat' does not have attribute 'id' for variable 'aws_eip.nat.*.id'
* Resource 'aws_eip.nat' does not have attribute 'id' for variable 'aws_eip.nat.*.id'
* Resource 'aws_eip.nat' does not have attribute 'id' for variable 'aws_eip.nat.*.id'
```

Every so often the build succeeds, but even then I sometimes receive some non-fatal `diffs didn't match during apply` warnings:

```
* aws_iam_instance_profile.instance_profile: diffs didn't match during apply. This is a bug with Terraform and should be reported as a GitHub Issue.

Terraform Version: 0.6.12
    Resource ID: aws_iam_instance_profile.instance_profile
    Mismatch reason: attribute mismatch: roles.2877750799
    Diff One (usually from plan): *terraform.InstanceDiff{Attributes:map[string]*terraform.ResourceAttrDiff{"create_date":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "unique_id":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "name":*terraform.ResourceAttrDiff{Old:"", New:"lc-instance-profile-${var.app_name}-${var.vpc_name}", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:true, Type:0x0}, "path":*terraform.ResourceAttrDiff{Old:"", New:"/", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:true, Type:0x0}, "roles.#":*terraform.ResourceAttrDiff{Old:"", New:"1", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "roles.2877750799":*terraform.ResourceAttrDiff{Old:"", New:"lc-instance-role-${var.app_name}-${var.vpc_name}", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "arn":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}}, Destroy:false, DestroyTainted:false}
    Diff Two (usually from apply): *terraform.InstanceDiff{Attributes:map[string]*terraform.ResourceAttrDiff{"name":*terraform.ResourceAttrDiff{Old:"", New:"lc-instance-profile-asg-example-app-stg", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:true, Type:0x0}, "path":*terraform.ResourceAttrDiff{Old:"", New:"/", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:true, Type:0x0}, "roles.#":*terraform.ResourceAttrDiff{Old:"", New:"1", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "roles.4293975883":*terraform.ResourceAttrDiff{Old:"", New:"lc-instance-role-asg-example-app-stg", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "arn":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "create_date":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}, "unique_id":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Type:0x0}}, Destroy:false, DestroyTainted:false}
```

So, based on these data points, and the fact that it runs fine in my localdev, I'm guessing there's something about the CircleCI environment.  Most (all?) of these errors indicate AWS eventual consistency issues, but I'm struggling to explain why the CircleCI environment would be more likely to trigger them?

Some other datapoints:
- My localdev is a Macbook Pro with 8 cores
- My localdev's `GOMAXPROCS` = 27984
- CircleCI's environment has 32 cores (found from SSHinto into the build)
- CircleCI's `GOMAXPROCS` = ~4000000
- CircleCI's environment is running Ubuntu (I believe 14.04 LTS)
- Setting `GOMAXPROCS` = 1 via an env var in CircleCI did yield a successful build, but obviously slowed things down.
- Using `terraform apply -parallelism=3` (just to reduce this down from 10) did not seem to affect things, but I'm guessing `GOMAXPROCS` and `parallelism` ultimately have the same effect.

Does anyone have any ideas for how I might further debug this or why this is happening?  Thanks for your help and for this outstanding piece of software!

jrnt30:
Fixes #5335 - This adds in additional retry logic for the addition of the route.  As it stands today, there have been several issues with the AWS API not returning the explicit route via `ec2/DescribeRouteTables` for some period of time.  

The previous version just retried when the creation of the route was unsuccessful.

### DIFF
--- a/builtin/providers/aws/resource_aws_route.go
+++ b/builtin/providers/aws/resource_aws_route.go
@@ -184,7 +184,7 @@ func resourceAwsRouteCreate(d *schema.ResourceData, meta interface{}) error {
 		route, err = findResourceRoute(conn, d.Get("route_table_id").(string), d.Get("destination_cidr_block").(string))
 
 		if err != nil {
-			log.Print("[DEBUG] Attempting to find route in route table %s again", d.Get("route_table_id").(string))
+			log.Printf("[DEBUG] Attempting to find route in route table %s again", d.Get("route_table_id").(string))
 			return resource.RetryableError(err)
 		}
 
--- a/builtin/providers/aws/resource_aws_route.go
+++ b/builtin/providers/aws/resource_aws_route.go
@@ -179,7 +179,18 @@ func resourceAwsRouteCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error creating route: %s", err)
 	}
 
-	route, err := findResourceRoute(conn, d.Get("route_table_id").(string), d.Get("destination_cidr_block").(string))
+	var route *ec2.Route
+	err = resource.Retry(2*time.Minute, func() *resource.RetryError {
+		route, err = findResourceRoute(conn, d.Get("route_table_id").(string), d.Get("destination_cidr_block").(string))
+
+		if err != nil {
+			log.Print("[DEBUG] Attempting to find route in route table %s again", d.Get("route_table_id").(string))
+			return resource.RetryableError(err)
+		}
+
+		return nil
+	})
+
 	if err != nil {
 		return err
 	}


catsby:
Do you know what kind of error we get when the situation you describe in #5335 happens, when a route is created but the follow up `DescribeRoutes` doesn't have it? I'm curious if this retry logic is better suited to be in `findResourceRoute`

jrnt30:
There was no error per se.  It was that the response was empty, which ultimately led to the issue being generated that the resource didn't have an id (and would fail on subsequent runs due to an overlap of CIDR blocks).  

I had considered the `findResourceRoute` however as it stands it would also potentially impact the read operation (since we are just retrying until we actually get a response back).  In the missing case, that would incur the retry cost.  I could however add in an additional param around the retry if that would be more desirable.

PR #6911, opened 2016-05-28
provider/aws: aws_ami_description - data source edition (now aws_ami)

vancluever:
This is #4396 in data source form. See that issue/PR for details.

Incidentally, porting this was very easy, with one small testing caveat - to test the "resource" properly, I had to prefix the name with `data` to ensure that it could be looked up in the state. This may just be semantics we have to live with, or could be the basis for a quality of life update.

I also ran into the same bug that @jen20 is running into on #6881, for which I have a fix incoming shortly.

PR #6913, opened 2016-05-28
[WIP] core: Ensure that DataSourcesMap is handled during apply w/destroy

vancluever:
Fixes #6881, which came up for me in #6911 as well. 

WIP because I'm still trying to find a good way to test the `*schema.Provider` part.

### DIFF
--- a/helper/schema/provider.go
+++ b/helper/schema/provider.go
@@ -168,7 +168,15 @@ func (p *Provider) Apply(
 	info *terraform.InstanceInfo,
 	s *terraform.InstanceState,
 	d *terraform.InstanceDiff) (*terraform.InstanceState, error) {
-	r, ok := p.ResourcesMap[info.Type]
+	m := p.ResourcesMap
+	for k, v := range p.DataSourcesMap {
+		if _, ok := m[k]; ok {
+			panic(fmt.Errorf("Data source %s is also a resource. This is a bug and should be reported.", k))
+		}
+		m[k] = v
+	}
+
+	r, ok := m[info.Type]
 	if !ok {
 		return nil, fmt.Errorf("unknown resource type: %s", info.Type)
 	}
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -133,11 +133,12 @@ func (r *Resource) Apply(
 
 	if d.Destroy || d.RequiresNew() {
 		if s.ID != "" {
-			// Destroy the resource since it is created
-			if err := r.Delete(data, meta); err != nil {
-				return r.recordCurrentSchemaVersion(data.State()), err
+			if r.Delete != nil {
+				// Destroy the resource since it is created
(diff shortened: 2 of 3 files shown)


apparentlymart:
@vancluever do you think this is the same issue @jen20 was trying to address in #6907?

We noticed during some debugging that we're generating "destroy" and "destroy tainted" nodes for data resources, which isn't supposed to happen and isn't handled properly in core... the "destroy" nodes always do the managed resource behavior, regardless of resource type.

He ultimately decided that this wasn't the right fix and we didn't discuss why, but just want to make sure we're not trying to solve the same problem in a multitude of different ways...

vancluever:
@apparentlymart this is a little different - the intention isn't to do anything with the graph, but one thing I did notice is that `*schema.Provider.Apply()` definitely does not look at the `DataSourcesMap` when it does its thing, and I didn't necessarily see anything that would lead me to believe that it would only apply to destroys, so it might have been something that should be checked anyway.

If anything I'd say the latter part is more around fixing the destroy walk issue, where we just ignore a missing `Delete` function and just proceed as normal. I'd image that won't be needed if a destroy walk is not happening on data sources.

I did notice that @jen20 did say that the destroy stuff was needed in some respect, I figured that you guys might have talked about it though :)

Some additional thoughts: I'm not too sure if ignoring `Destroy` when deleting is the best idea, as it might create a case where bugs with resources (versus data sources) that are not properly designed don't panic when they should, and I'm wondering if the logic put into `*schema.Provider.Apply()` should be put into the other functions there as well (namely `*schema.Provider.Diff()` and `*schema.Provider.Refresh()`).

apparentlymart:
The intention here is that we never enter `schema.Provider.Apply` for a data resource, since they are supposed to use the separate function `ReadDataApply`. So if we *are* ending up in this codepath for data resources then that's a bug in itself , and I think what @jen20 was looking at for #6907.

If I'm understanding this patch correctly then it makes it illegal to have both a resource and a data source both called `aws_ami`, which is in conflict with the intent here: we actually *want* to have matching names, so that e.g. `resource "aws_ami"` will create and manage an AMI while `data "aws_ami"` will retrieve the attributes of one that already exists, ideally with the attributes as consistent as possible between the two.

I suspect that whatever patch replaces #6907 will make this patch moot, since it will prevent us from reaching any of the branches you've created here... but we can wait until @jen20 has found a new path and confirm if this remains true.

vancluever:
Yeah, this is something I was wondering about. (I was thinking about renaming the data source in #6911 to `aws_ami`, incidentally).

Maybe my efforts are better focused on trying to figure out why the root of the issue exists then versus trying to work around it... seeing as I am still trying to figure out how to test `schema.Provider.Apply` properly and coming up blank, I'm not going to complain!

vancluever:
I think I've found the root cause, closing this now and will be putting in another PR in a sec.

PR #6915, opened 2016-05-28
provider/aws: remove ah and esp as aws endpoint does not support the string versions

### LINKED ISSUE #6888: provider/aws: aws_security_group_rule fails with protocol 50 or 51
### Terraform Version
0.6.16

### Affected Resource(s)
- aws_security_group_rule

```hcl
resource "aws_security_group" "vpn" {
  description = "Security Group for East to West VPN"
  vpc_id = "${aws_vpc.us-east-1-doximity.id}"
}

resource "aws_security_group_rule" "vpn_protocol_50" {
  type = "ingress"
  from_port = 0
  to_port = 0
  protocol = "50"
  cidr_blocks = ["${aws_eip.us-west-2a-vpn.public_ip}/32"]
  security_group_id = "${aws_security_group.vpn.id}"
}

resource "aws_security_group_rule" "vpn_protocol_51" {
  type = "ingress"
  from_port = 0
  to_port = 0
  protocol = "51"
  cidr_blocks = ["${aws_eip.us-west-2a-vpn.public_ip}/32"]
  security_group_id = "${aws_security_group.vpn.id}"
}
```

### Debug Output
https://gist.github.com/sarkis/2909062add54496b4dba0a69eaecd375

### Panic Output
N/A

### Expected Behavior
Security group rule created / updated in state

### Actual Behavior
400 error from AWS

### Steps to Reproduce
1. Set up an aws_security_group_rule with protocol set to any of: 50, "50", 51, "51", "ah", "esp"
2. `terraform apply`

### Important Factoids
Nothing unique to our infra, we are running in EC2 VPC and have not had this issue until 0.6.16.

### References
Couldn't find any issues that directly relate to this one... Although, since we didn't have these problems with 0.6.14, it may be directly related to this PR:

https://github.com/hashicorp/terraform/pull/5881

sarkis:
The AWS endpoint accessed via the aws-go-sdk does not support the string version of the "ah" and "esp" protocols. When attempting to create security group rules (via the go sdk) using the string representations the following errors are returned:

```
InvalidParameterValue: Invalid value 'ah' for IP protocol. Unknown protocol.
        status code: 400, request id: 3fb7c0d8-8fc9-40d7-948b-5b438f671cfc

InvalidParameterValue: Invalid value 'esp' for IP protocol. Unknown protocol.
        status code: 400, request id: 3fb7c0d8-8fc9-40d7-948b-5b438f671cfc
```

The fix for this problem is simply removing ah and esp from the `protocolIntegers` map which will end up in the expected behavior as there is a proper fall through here: https://github.com/hashicorp/terraform/blob/1c81aa3471ee85368cce4f5e1d8030b788012dc4/builtin/providers/aws/resource_aws_security_group.go#L878-L880

Fixes #6888

### DIFF
--- a/builtin/providers/aws/network_acl_entry.go
+++ b/builtin/providers/aws/network_acl_entry.go
@@ -82,8 +82,6 @@ func protocolIntegers() map[string]int {
 	var protocolIntegers = make(map[string]int)
 	protocolIntegers = map[string]int{
 		// defined at https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
-		"ah":   51,
-		"esp":  50,
 		"udp":  17,
 		"tcp":  6,
 		"icmp": 1,


jrnt30:
I was thinking it may be worthwhile to approach this slightly differently.  Since TCP/UDP/ICMP can also be sent over as the proto numbers, what if we keep support for Proto Names (and we could even expand it) but simply convert those to the numeric equivalent prior to sending to AWS?

sarkis:
@jrnt30 I thought about that being a possible solution, but I was discouraged a bit when I looked at all the possibilities: http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

jrnt30:
Yes, that is true.  I agree this needs to change for the to allow the provider to actually work but it does seem strange to basically be reversing functionality that was implemented for another issue. 

In making the change as is the original author of the 'AH/ESP' support will now have failing configuration again.  It seemed like normalizing it in TF code would be the easier way of handling the existing functionality & ensuring the lookup works across the different quirky AWS APIs.

sarkis:
I don't think ah/esp works anywhere currently in 0.6.16 and master... @catsby looping you in as you were the PR owner for the original change. Is there somewhere outside of aws_security_group_rule that ah/esp protocols work?

sarkis:
Okay - just got the time to test this out. It seems that "ah" and "esp" protocols do work for network acls and just not for security groups.

catsby:
Sorry for being late to this party – https://github.com/hashicorp/terraform/pull/8999 should sort this out. Thanks!

PR #6925, opened 2016-05-29
add an example of cidrsubnet() being used with ipv6

RsrchBoy:
...as this will hopefully clue people in that this function will indeed work
to manipulate ipv6 networks.

Not that I completely spaced on that for quite some time, or anything like
that.

Nope, not me.  Not at all.


RsrchBoy:
No problem -- I was more frustrated with myself than anything, really :)

Issue #6099, opened 2016-04-08
Support name prefix everywhere


joshuaspence:
I generally try to use `name_prefix` instead of `name` where possible, but not all resources support `name_prefix`. For example, `aws_security_group` does but `aws_iam_role` does not. It would be great to have support for `name_prefix` everywhere that `name` is supported.

stack72:
@joshuaspence I am going to go ahead and close this out. The bulk of this has been added :) 

Thanks for all the work you put in on this

PR #6939, opened 2016-05-30
Add `name_prefix` to `aws_iam_instance_profile` and `aws_iam_role`

joshuaspence:
Ref #6099. Add a `name_prefix` attribute to the `aws_iam_instance_profile` and `aws_iam_role` resources.

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_instance_profile.go
+++ b/builtin/providers/aws/resource_aws_iam_instance_profile.go
@@ -8,6 +8,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -23,18 +24,23 @@ func resourceAwsIamInstanceProfile() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
 			"create_date": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
 			"unique_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
 			"name": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"name_prefix"},
 				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 					// https://github.com/boto/botocore/blob/2485f5c/botocore/data/iam/2010-05-08/service-2.json#L8196-L8201
 					value := v.(string)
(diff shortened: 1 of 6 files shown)


stack72:
Hi @joshuaspence 

when you said that it doesn't quite work yet, what do you mean? :)

P;

joshuaspence:
Oh, the acceptance tests are failing and I don't understand why.

builtin/providers/aws/resource_aws_iam_instance_profile.go
```diff
@@ -23,18 +24,22 @@ func resourceAwsIamInstanceProfile() *schema.Resource {
...
 				Computed: true,
 			},
+
 			"name": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
```
stack72: @joshuaspence as you do not specify a name but do specify a name_prefix, you need to make name Computed. This means that when we set it in the Read, we will not get any alternative plan

Doing this locally gave me this:

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSIAMInstanceProfile'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSIAMInstanceProfile -timeout 120m
=== RUN   TestAccAWSIAMInstanceProfile_basic
--- PASS: TestAccAWSIAMInstanceProfile_basic (16.34s)
=== RUN   TestAccAWSIAMInstanceProfile_namePrefix
--- PASS: TestAccAWSIAMInstanceProfile_namePrefix (17.01s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	33.367s
```

builtin/providers/aws/resource_aws_iam_role.go
```diff
@@ -23,14 +24,17 @@ func resourceAwsIamRole() *schema.Resource {
...
 			"unique_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
 			"name": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
```
stack72: Again this needs to be computed so that when we don't set it as part of our config, the Read can still set it and allow it to not cause Updates :)

stack72:
@joshuaspence ok, 2 small changes needed here and we are in a working situation :)

P.

joshuaspence:
Thanks, this should be ready to go now. I didn't quite understand why the `name` field needs to be marked as computed because it may not necessarily be so (for example, if the `name` attribute of the prefix is specified rather than `name_prefix`).

builtin/providers/aws/resource_aws_iam_instance_profile.go
```diff
@@ -49,12 +55,33 @@ func resourceAwsIamInstanceProfile() *schema.Resource {
...
 			},
+
+			"name_prefix": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					// https://github.com/boto/botocore/blob/2485f5c/botocore/data/iam/2010-05-08/service-2.json#L8196-L8201
+					value := v.(string)
+					if len(value) > 64 {
```
joshuaspence: This value probably isn't correct... are there any guarantees made about the length of the string returned by the `uuidV4` function?

stack72:
Hi @joshuaspence 

I'll give this a final set of tests when I am back online. Just a note on Computed - this is used when there is a chance that the value get's set on the Read but not in the initial config. The Computed means that Terraform will understand why there would be changes in that case

P.

stack72:
@joshuaspence these looks great now :)

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSRole'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSRole -timeout 120m
=== RUN   TestAccAWSRolePolicyAttachment_basic
--- PASS: TestAccAWSRolePolicyAttachment_basic (33.34s)
=== RUN   TestAccAWSRole_basic
--- PASS: TestAccAWSRole_basic (16.32s)
=== RUN   TestAccAWSRole_namePrefix
--- PASS: TestAccAWSRole_namePrefix (16.08s)
=== RUN   TestAccAWSRole_testNameChange
--- PASS: TestAccAWSRole_testNameChange (39.93s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	105.703s
```

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSIAMInstanceProfile'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSIAMInstanceProfile -timeout 120m
=== RUN   TestAccAWSIAMInstanceProfile_basic
--- PASS: TestAccAWSIAMInstanceProfile_basic (18.32s)
=== RUN   TestAccAWSIAMInstanceProfile_namePrefix
--- PASS: TestAccAWSIAMInstanceProfile_namePrefix (19.99s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	38.329s
```

Thanks for the hard work!

PR #6946, opened 2016-05-31
provider/aws: Add aws_s3_bucket_object data source

radeksimko:
There are 2 main use-cases I can think of:

1. Streamlined Lambda deployment where ZIP files with new versions are uploaded to S3 outside of Terraform (i.e. new version ID is generated)
2. Some kind of configuration stored in S3 object (kinda like `consul_keys`). Plaintext values per object should be easy to read and use out of the box, JSON/YAML and other formats might need some extra care.

### `helper/resource` - `PreventPostDestroyRefresh`

**TL;DR:** This is to reduce number of TestCases from 3 to 2. It might be possible to reduce it down to 1, but I wanted to test the behaviour of data source at the beginning of the lifecycle, not in the middle.

The S3 object + S3 bucket we're testing there is set up as part of the test case and we run `refresh` _after_ destroying all resources by default - probably to verify that all resources are gone. This is ok for data sources which read data outside of Terraform's control. Destroy operation however doesn't destroy data sources which means that `refresh` on `aws_s3_object` reading object that was destroyed would fail.

### Test plan

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=DataSourceAWSS3BucketObject'
```
```
TF_ACC=1 go test ./builtin/providers/aws -v -run=DataSourceAWSS3BucketObject -timeout 120m
=== RUN   TestAccDataSourceAWSS3Object_basic
--- PASS: TestAccDataSourceAWSS3Object_basic (56.73s)
=== RUN   TestAccDataSourceAWSS3Object_readableBody
--- PASS: TestAccDataSourceAWSS3Object_readableBody (57.21s)
=== RUN   TestAccDataSourceAWSS3Object_kmsEncrypted
--- PASS: TestAccDataSourceAWSS3Object_kmsEncrypted (317.53s)
=== RUN   TestAccDataSourceAWSS3Object_allParams
--- PASS: TestAccDataSourceAWSS3Object_allParams (66.85s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	498.339s
```

### DIFF
--- a/builtin/providers/aws/data_source_aws_s3_object.go
+++ b/builtin/providers/aws/data_source_aws_s3_object.go
@@ -1,0 +1,221 @@
+package aws
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsS3BucketObject() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsS3BucketObjectRead,
+
+		Schema: map[string]*schema.Schema{
+			"body": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"bucket": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"cache_control": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"content_disposition": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"content_encoding": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"content_language": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"content_length": &schema.Schema{
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"content_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"etag": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"expiration": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"expires": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"key": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"last_modified": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"metadata": &schema.Schema{
+				Type:     schema.TypeMap,
+				Computed: true,
+			},
+			"range": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"server_side_encryption": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"sse_kms_key_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"storage_class": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"version_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"website_redirect_location": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsS3BucketObjectRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).s3conn
+
+	bucket := d.Get("bucket").(string)
+	key := d.Get("key").(string)
+
+	input := s3.HeadObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+	}
+	if v, ok := d.GetOk("range"); ok {
+		input.Range = aws.String(v.(string))
+	}
+	if v, ok := d.GetOk("version_id"); ok {
+		input.VersionId = aws.String(v.(string))
+	}
+
+	versionText := ""
+	uniqueId := bucket + "/" + key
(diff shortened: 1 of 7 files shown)


builtin/providers/aws/provider.go
```diff
@@ -113,6 +113,7 @@ func Provider() terraform.ResourceProvider {
 		DataSourcesMap: map[string]*schema.Resource{
 			"aws_ami":                dataSourceAwsAmi(),
 			"aws_availability_zones": dataSourceAwsAvailabilityZones(),
+			"aws_s3_object":          dataSourceAwsS3Object(),
```
apparentlymart: Should we instead call this `aws_s3_bucket_object` so it matches with the resource of the same name? So far we've been trying to make the data sources match the corresponding resources as much as possible (in terms of attributes too) so it's relatively easy to switch between them.

helper/resource/testing_config.go
```diff
@@ -101,10 +101,12 @@ func testStep(
 	}
 
 	// And another after a Refresh.
-	state, err = ctx.Refresh()
-	if err != nil {
-		return state, fmt.Errorf(
-			"Error on follow-up refresh: %s", err)
+	if !step.Destroy || (step.Destroy && !step.PreventPostDestroyRefresh) {
```
apparentlymart: Hmm... some of the code below (comment starting with "Data sources are tricky") is there with the intent that this wouldn't be necessary, but I assume when you were testing this you found that it still was for some reason?

apparentlymart: Oh, sorry... I re-read your description and I think I understand what you were going for here now: the refresh fails because the bucket (managed by Terraform in this case) is deleted.

I wonder if this would be necessary if your `aws_s3_object` data block interpolated the bucket name and object key from the two resources... I expect that in that case Terraform sould understand that the data source depends on the bucket and the object and would thus *not* try to refresh them when the resources are not present. Unfortunately I expect that this will not work as desired, because the bucket name and object key are not computed in the configuration so Terraform will think it "knows" these values even before those objects are created. This seems like a bug but I'm not sure what the *right* behavior would be given the building blocks this is made of.

builtin/providers/aws/data_source_aws_s3_object.go
```diff
@@ -0,0 +1,221 @@
...
+	if isContentTypeAllowed(out.ContentType) {
+		input := s3.GetObjectInput{
+			Bucket: aws.String(bucket),
+			Key:    aws.String(key),
+		}
+		if v, ok := d.GetOk("range"); ok {
+			input.Range = aws.String(v.(string))
+		}
+		if v, ok := d.GetOk("version_id"); ok {
+			input.VersionId = aws.String(v.(string))
```
apparentlymart: Should we set `input.VersionId` to be `out.VersionId` unconditionally here? That way in the case where the `version_id` argument was *not* provided by the caller we should still be sure to get the body at the same version as we got the head.

Admittedly this race is relatively minor compared to other races in Terraform -- and certainly there are lots of them -- but doesn't seem like it'd add complexity to handle it given that we already know the version id by the time we get down here.

apparentlymart:
Aside from some nits, this LGTM. Thanks... I was just the other day wishing for this data source!

builtin/providers/aws/provider.go
radeksimko: Sure, that's no problem at all, I can rename it.

Generally speaking though I'd rather see resource called `aws_s3_object` as it just makes more sense when you pronounce it (IMO). 😃 

I assume that consistency with past choices is probably easier & safer to implement than deprecating past choices though.

helper/resource/testing_config.go
radeksimko: It probably is a JFDI approach to this problem.

I think the "right approach" ™️  would be (optionally) performing selective post-destroy refresh (i.e. exclude all or some data sources), but that would probably require more digging and more +/- LOC and I didn't want to end up in that 🐰  hole.

> I expect that in that case Terraform sould understand that the data source depends on the bucket and the object and would thus not try to refresh them when the resources are not present. > Unfortunately I expect that this will not work as desired ...

I can imagine the edge cases that aren't quite well covered yet which is exactly why I tried to keep data sources as isolated from resources as possible - after all that's going to be the common use case.

There is no reason why should user be attaching S3 object data source to an S3 object resource - they would use resource attributes instead. We should be testing common use cases. 😉 

i.e. I agree it should behave like you described, but that behaviour is not something we want to test in this specific context (s3-object-resource / s3-object-data-source).

builtin/providers/aws/data_source_aws_s3_object.go
radeksimko: I'm tempted to agree that we should do it, but I will need to double check the behaviour for unversioned bucket - basically I need to check if

```go
GetObjectInput{Bucket: aws.String(bucket), Key: aws.String(key)}
```
is equal to
```go
GetObjectInput{Bucket: aws.String(bucket), Key: aws.String(key), VersionId: aws.String("")}
```
if that's the case, then I'm happy to make that change.

vancluever:
@radeksimko I too was thinking about this data source! Question: would it be too tough to get in KMS support or possible customer SSE support as well? It would make in-band use of encrypted S3 data so much easier (and open up the playing field for using S3 as an easy secret store for Terraform even).

vancluever:
@radeksimko ahh yeah, you are right, and `GetObject` doesn't even have a way to use KMS versus SSE, so there you go.

Very nice - might have to rebuild our fork in the next couple of days with this! Thanks a ton!

builtin/providers/aws/data_source_aws_s3_object.go
radeksimko: So apparently it is not the same:

```
* data.aws_s3_bucket_object.yada: Failed getting S3 object: InvalidArgument: Version id cannot be the empty string
	status code: 400, request id: D2F18E0DCD0D3BC4
```

so we can't remove the condition, but we can totally turn it into `if out.VersionId != nil {` to avoid the race condition you mentioned.

radeksimko:
@apparentlymart I address 2 of your valid comments + provided some explanation on the third, I also reran acceptance tests which all passed.

Will you give it final 👍  / 👎  ? 😉

apparentlymart:
I didn't have time to test again but by my eyes it looks great! :+1:

radeksimko:
^ FYI @gpeng @surminus this may be useful for your Lambda deployments

(landing in `0.7` soon) 😉

surminus:
@radeksimko great thanks!

Djuke:
Will there be support for interpolating the values of arguments `bucket` and `key`?

PR #6955, opened 2016-05-31
provider/aws: Add iam_arn to aws_cloudfront_origin_access_identity

vancluever:
I've noticed that when S3 normalizes a bucket policy it takes the `CanonicalUser` principal and converts it to an `AWS` one, with an ARN like:

```
arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity ABCD123456
```

This of course causes more of the spurious diffs that we all know and love ;)

So, I've added the `iam_arn` to the `aws_cloudfront_origin_access_identity` resource and included some notes in the docs.


catsby:
Thanks!

vancluever:
Thanks @catsby!

xenoterracide:
reading this commit, I thought this would work... http://serverfault.com/q/837851/6852

PR #6957, opened 2016-05-31
Updated eip_association example

Charg:
**Reasoning for docs update**: The example is referencing a non-existent variable, `allocation_id`, within the `aws_eip` resource. I believe this should actually be `aws_eip.example.id` instead of `aws_eip.example.allocation_id`. Following the example will result in an error when applying your terraform configuration.

**Relevant Terraform version:** Terraform v0.6.16


stack72:
Hi @Charg 

thanks for the PR here. EIP does have an allocation_id but we currently do not set it :)

Therefore, your PR has raised a bigger issue for us to fix!

I am going to merge this PR as `id` is set as the allocation_id in a VPC environment

Thanks

Paul

PR #6960, opened 2016-06-01
api.opscode.com should be api.chef.io

juliandunn:
The old name still works, but it's better to use the new one.


stack72:
Hi @juliandunn 

Thanks for the PR here. Sorry the URL was incorrect!

Paul

PR #6966, opened 2016-06-01
provider/vsphere: added mac address parameter

thetuxkeeper:
I added the mac address parameter that we are able to use DHCP with fixed leases/IP addresses for important machines (like consul master nodes)

cc @chrislovecnm


stack72:
Hi @thetuxkeeper 

This LGTM! Nice addition here :)

Paul

PR #6976, opened 2016-06-02
deps: Upgrade Azure SDK for Go

jen20:
This pull-request upgrades the Azure SDK for Go. Every resource that doesn't use Riviera is touched by this.

Acceptance tests for most resources have been run - however ones which are unacceptably slow (because of the upstream services) have been left as an excercise for Travis - these are:

- Template deployments
- Virtual machines
- CDN resources


stack72:
Once Travis is green, we should just merge this and check the acceptance tests over the next few nights to make sure all is well

jen20:
Agreed. Hopefully the ASM provider is now green, was missing a dep after upgrade and had some storage changes.

jen20:
OK we have a green build! I'll merge this now and see what the acceptance tests look like overnight...

cc @phinze, @stack72, @jbardin

PR #6977, opened 2016-06-02
Fix errata Opsworks Documentation

### LINKED ISSUE #6885: docs: aws_opsworks_instance ebs_block_device errata
I very painfully discovered an issue with the docs on https://www.terraform.io/docs/providers/aws/r/opsworks_instance.html

`encrypted` is not a valid option for block device mappings in opsworks

this can be confirmed by looking up the docs for the opsworks ebs block device model in their various apis (.NET or java, for example)

attempting to use the flag results in an error (unsupported)

I believe it was copied from the aws_instance section, and thus just needs to be updated to remove the erroneous `encrypted` option.

optimisticanshul:
fixes #6885 
/cc @catsby @stack72


stack72:
Yha ks for the fix here @optimisticanshul  - I will get this to the website today

P.

PR #6979, opened 2016-06-02
provider/aws: aws_s3_object -> aws_s3_bucket_object (docs + filename)

radeksimko:
See [`builtin/providers/aws/provider.go`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/provider.go#L117)


stack72:
LGTM! Nice one @radeksimko

Issue #3639, opened 2015-10-26
Docker images version change do not trigger container update


lvjp:
# Problem
When i do docker image update i should run it in two step:
    1. Update the image
    2. Update container associated with the updated image

I think that terraform can be smart enough for handle this in one step.

# My environment
## My .tf declaration:
```
provider "docker" {
    host = "unix:/var/run/docker.sock"
}
resource "docker_container" "my-docker-container" {
    name = "my-docker-container"
    image = "${docker_image.my-docker-image.latest}"
}
resource "docker_image" "my-docker-image" {
    name="my/docker-image:0.5.2"
}
```

### My version number
```shell
$ terraform version
Terraform v0.6.6
```

### First plan
```shell
$ terraform plan
[... refresh cutted ...]
~ docker_image.my-docker-image
    name: "my/docker-image:0.5.2" => "my/docker-image:0.6.2"
```

### First apply
```shell
docker_image.my-docker-image: Refreshing state... (ID: 4f744e3faa29693abefe6ef45dddc11fbd15b388b5238b46f1a0828279dc2440my/docker-image:0.5.2)
docker_container.my-docker-container: Refreshing state... (ID: 6b2cd1316e16097aa5c18bc3d3d99d3b249cc0b0a8b6ba47aed338cbc20b706f)
docker_image.my-docker-image: Modifying...
  name: "my/docker-image:0.5.2" => "my/docker-image:0.6.2"
docker_image.my-docker-image: Modifications complete

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
[... Help cutted ...]
```

### Second plan
```shell
[... refresh cutted ...]
-/+ docker_container.my-docker-container
    bridge:           "" => "<computed>"
    gateway:          "172.17.42.1" => "<computed>"
    image:            "4f744e3faa29693abefe6ef45dddc11fbd15b388b5238b46f1a0828279dc2440" => "1e39e488bd5b2789b19eb56349702c682430cf507597bfee1dcc81b70955d203" (forces new resource)
    ip_address:       "172.17.0.49" => "<computed>"
    ip_prefix_length: "16" => "<computed>"
    must_run:         "true" => "1"
    name:             "my-docker-container" => "my-docker-container"
```

### Second apply
```shell
docker_image.my-docker-image: Refreshing state... (ID: 4f744e3faa29693abefe6ef45dddc11fbd15b388b5238b46f1a0828279dc2440my/docker-image:0.5.2)
[...]

PR #7000, opened 2016-06-03
provider/docker: Data source for docker image

kyhavlov:
This PR is to add a data source for fetching the version of a docker image from a registry, so that terraform can correctly show the update of the image and its dependent containers in the same run, as discussed in #3639. 

I've left the existing docker_image resource unchanged except for adding a "registry_id" field to it with `ForceNew = true` that can be passed in from this data source to determine whether the image needs to be re-pulled.

An example use of this data source:
```hcl
data "docker_image" "nomad" {
  name = "kyhavlov/nomad:latest"
}

resource "docker_image" "nomad" {
  name = "${data.docker_image.nomad.name}"
  registry_id = "${data.docker_image.nomad.id}"
}

resource "docker_container" "nomad-server" {
  image        = "${docker_image.nomad.latest}"
  name         = "nomad-server"
}
```

I've still got to add an acceptance test for this (new to Go and haven't had experience with terraform acceptance tests so any input here is welcome) and update the docs, but the code itself is finished.

### DIFF
--- a/builtin/providers/docker/data_source_docker_registry_image.go
+++ b/builtin/providers/docker/data_source_docker_registry_image.go
@@ -1,0 +1,166 @@
+package docker
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceDockerRegistryImage() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDockerRegistryImageRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"sha256_digest": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceDockerRegistryImageRead(d *schema.ResourceData, meta interface{}) error {
+	pullOpts := parseImageOptions(d.Get("name").(string))
+
+	// Use the official Docker Hub if a registry isn't specified
+	if pullOpts.Registry == "" {
+		pullOpts.Registry = "registry.hub.docker.com"
+	} else {
+		// Otherwise, filter the registry name out of the repo name
+		pullOpts.Repository = strings.Replace(pullOpts.Repository, pullOpts.Registry+"/", "", 1)
+	}
+
+	// Docker prefixes 'library' to official images in the path; 'consul' becomes 'library/consul'
+	if !strings.Contains(pullOpts.Repository, "/") {
+		pullOpts.Repository = "library/" + pullOpts.Repository
+	}
+
+	if pullOpts.Tag == "" {
+		pullOpts.Tag = "latest"
+	}
+
+	digest, err := getImageDigest(pullOpts.Registry, pullOpts.Repository, pullOpts.Tag, "", "")
+
+	if err != nil {
+		return fmt.Errorf("Got error when attempting to fetch image version from registry: %s", err)
+	}
+
+	d.SetId(digest)
+	d.Set("sha256_digest", digest)
+
+	return nil
+}
+
+func getImageDigest(registry, image, tag, username, password string) (string, error) {
+	client := http.DefaultClient
+
+	req, err := http.NewRequest("GET", "https://"+registry+"/v2/"+image+"/manifests/"+tag, nil)
+
+	if err != nil {
+		return "", fmt.Errorf("Error creating registry request: %s", err)
+	}
+
+	if username != "" {
+		req.SetBasicAuth(username, password)
+	}
+
+	resp, err := client.Do(req)
+
+	if err != nil {
+		return "", fmt.Errorf("Error during registry request: %s", err)
+	}
+
+	switch resp.StatusCode {
+	// Basic auth was valid or not needed
+	case http.StatusOK:
+		return resp.Header.Get("Docker-Content-Digest"), nil
+
+	// Either OAuth is required or the basic auth creds were invalid
+	case http.StatusUnauthorized:
+		if strings.HasPrefix(resp.Header.Get("www-authenticate"), "Bearer") {
+			auth := parseAuthHeader(resp.Header.Get("www-authenticate"))
+			params := url.Values{}
+			params.Set("service", auth["service"])
+			params.Set("scope", auth["scope"])
+			tokenRequest, err := http.NewRequest("GET", auth["realm"]+"?"+params.Encode(), nil)
+
+			if err != nil {
+				return "", fmt.Errorf("Error creating registry request: %s", err)
+			}
+
+			if username != "" {
+				tokenRequest.SetBasicAuth(username, password)
+			}
+
+			tokenResponse, err := client.Do(tokenRequest)
+
+			if err != nil {
+				return "", fmt.Errorf("Error during registry request: %s", err)
+			}
+
+			if tokenResponse.StatusCode != http.StatusOK {
+				return "", fmt.Errorf("Got bad response from registry: " + tokenResponse.Status)
+			}
+
+			body, err := ioutil.ReadAll(tokenResponse.Body)
+			if err != nil {
+				return "", fmt.Errorf("Error reading response body: %s", err)
+			}
+
+			token := &TokenResponse{}
+			err = json.Unmarshal(body, token)
+			if err != nil {
+				return "", fmt.Errorf("Error parsing OAuth token response: %s", err)
+			}
+
+			req.Header.Set("Authorization", "Bearer "+token.Token)
+			digestResponse, err := client.Do(req)
+
+			if err != nil {
+				return "", fmt.Errorf("Error during registry request: %s", err)
+			}
+
+			if digestResponse.StatusCode != http.StatusOK {
+				return "", fmt.Errorf("Got bad response from registry: " + digestResponse.Status)
+			}
+
+			return digestResponse.Header.Get("Docker-Content-Digest"), nil
+		} else {
+			return "", fmt.Errorf("Bad credentials: " + resp.Status)
+		}
+
+	// Some unexpected status was given, return an error
+	default:
+		return "", fmt.Errorf("Got bad response from registry: " + resp.Status)
+	}
+}
+
+type TokenResponse struct {
+	Token string
+}
+
+// Parses key/value pairs from a WWW-Authenticate header
+func parseAuthHeader(header string) map[string]string {
+	parts := strings.SplitN(header, " ", 2)
+	parts = strings.Split(parts[1], ",")
+	opts := make(map[string]string)
+
+	for _, part := range parts {
+		vals := strings.SplitN(part, "=", 2)
+		key := vals[0]
+		val := strings.Trim(vals[1], "\", ")
+		opts[key] = val
+	}
+
+	return opts
+}
--- a/builtin/providers/docker/data_source_docker_registry_image_test.go
+++ b/builtin/providers/docker/data_source_docker_registry_image_test.go
@@ -1,0 +1,52 @@
+package docker
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+var registryDigestRegexp = regexp.MustCompile(`\A[A-Za-z0-9_\+\.-]+:[A-Fa-f0-9]+\z`)
+
+func TestAccDockerRegistryImage_basic(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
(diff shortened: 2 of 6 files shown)


radeksimko:
Great you came up with this PR, this was the next on my list of data sources I wanted to see in Terraform! 👍  😄 

I feel there are two use-cases for this data source:

1. Define tag, typically `latest` or `v1` (generally tags we expect to be repointed to different hashes) and translate that into hash
2. Just define repo URL (`library/nginx`) and translate that either into hash or tag which we don't expect to move.

The obvious question re 2) is how to determine which tag/hash is the latest - hopefully the API returns these in a particular order, if not we'd probably need to do some sorting.

Also having the _computed_ tag can be useful when you want to use that as human-readable output - e.g.

```hcl
data "docker_image" "nomad" {
  name = "kyhavlov/nomad"
}

resource "docker_image" "nomad" {
  name = "${data.docker_image.nomad.name}"
  image_id = "${data.docker_image.nomad.latest_tag}"
}
output "version_deployed" {
  value = "${data.docker_image.nomad.latest_tag}"
}
```

```
Outputs:

version_deployed: v0.4.5
```

also diff like this is much easier to read by humans:

```
~ docker_image.nomad
    image_id: "0.3.0" => "0.3.1"
```

rather than
```
~ docker_image.nomad
    image_id: "308e519aac60" => "493d82594c15"
```

kyhavlov:
For 2) the API doesn't return them in any order, so you have to loop through them and check the manifest for each one and compare dates. 

That said, is wanting to detect the newest tag a common thing? Docker already has the convention of latest being an alias to the most up to date tag, and I'd have thought that people would want to be explicit with the tags they give when deploying, rather than letting terraform implicitly find a tag for them. It's also counter to the docker behavior for tags, which is to default to latest when a tag isn't specified (instead of discovering the most recent).

kyhavlov:
This should be ready for review now; I added a section in the docs and a couple acceptance tests for the new data source.

kyhavlov:
@radeksimko @jen20 Any comments on what else (if anything) I should do for this?

kyhavlov:
Can anyone review this? @phinze @jefferai ? This addresses a pretty major issue with using the docker provider.

kyhavlov:
Can anyone give feedback on this? @stack72

phinze:
Hi @kyhavlov - thanks for your work on this so far!

I think @radeksimko is the best person to pull in for a review now that you've got a full working draft up here.

One concern I have is about the additional library dependency load this implementation uses. Importing ~70 new files seems hefty for what amounts to a single API call. Is there a bunch of work that the docker-registry-client library is doing, or can we inline an HTTP request + JSON parse to simplify this?

I'll let Radek comment more on the modeling questions, and we'll keep moving this forward! 🚀

kyhavlov:
@phinze I'd like to do this without adding the dependency but I'm not sure how. The problem this is fixing stems from the fact that terraform needs to know whether it will have to update an image. Actually doing the pull just to do read() doesn't work, because then it affects the resource, so I don't see any other way except checking the registry. Getting the image manifest from the registry involves a couple API calls and dealing with oauth endpoints, even on the public docker hub, so it's doing a decent amount of work that we'd have to replicate even for just the basic case.

hmcgonig:
This is awesome, can't wait for this. 👍

kyhavlov:
@phinze @radeksimko any suggestions about the pull request/dependency?

radeksimko:
Hi @kyhavlov
sorry for the delay. I took some time to read the current implementation of `docker_image` resource to understand the history & relationship to this new data source and also differences between [Docker Remote API](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/) & [Docker Registry API](https://docs.docker.com/registry/spec/api/).

#### Data source VS resource

The `docker_image` resource reads the metadata about the image; mostly translates `latest` to hash by **pulling** the latest image/layer and then it also **removes** the old image/layer - all of that using `docker` provider client connection (Docker Remote API). Pulling the image can be useful as it is then cached as part of the terraform apply run even before the new image replaces the old one.

Your new data source uses the Docker Registry API (which is btw. what makes it very thin and quick 👍 ) to read the image metadata.

It IMO makes total sense to keep providing both functionalities as long as the separation is clear for the user and makes sense from maintenance perspective.

#### Provider separation

I feel like your data source and anything else that doesn't need Docker API access deserves a separate provider - called e.g. `docker_registry`. Basically https://github.com/hashicorp/terraform/pull/6507 turned into a provider. This would make the data source useful for people who don't schedule containers via docker API directly (e.g. ECS, K8S) too.

#### `docker_image` resource

Instead of `registry_id` I suggest we follow the concept of `triggers` we have in `null_resource`, see https://www.terraform.io/docs/provisioners/null_resource.html and/or https://github.com/hashicorp/terraform/blob/master/builtin/providers/null/resource.go#L22-L26 and/or just expect the user to make `name` a reference - I think both use cases make sense - e.g. if you want to trigger `pull` of related containers when one is updated.

In addition to that we should make it very clear in the documentation that **this resource doesn't pull new layers by default** and encourage user to either interpolate `data.docker_registry_image` in the `name` field or use `trigger_pull`.

#### Example

Basically I'm thinking about it like this:

```hcl
provider "docker_registry" {
  username = ""
  password = ""
}

data "docker_registry_image" "nomad" {
  name = "nomad"
  tag = "latest"
}

provider "docker" {
  host = "/my/lovely/docker/unix/socket"
}

# First use case
resource "docker_image" "nomad" {
  name = "${data.docker_registry_image.nomad.name}:${data.docker_registry_image.nomad.digest}"
}

# Second use case
resource "docker_image" "nomad" {
  name = "my_special_app:latest"
  trigger_pull = ["${data.docker_registry_image.nomad}"]
}

# Third use case
resource "aws_ecs_task_definition" "nomad" {
  family = "nomad"
  container_definitions = <<DEFINITION
[
  {
    "name": "jenkins",
    "image": "${data.docker_registry_image.nomad.name}:${data.docker_registry_image.nomad.digest}",
    "cpu": 10,
    "memory": 500,
  }
]
DEFINITION
}
```

I think we could remove the `keep_updated` field as it just causes the misbehaviour w/ double-plan/apply described in #3639 and has no further value.

@phinze I also feel the pain of dependency accumulation but I think it's ok to pull that library in given the complexity of docker registry auth - we need auth even for anonymous access.
I'm actually thinking that more of the logic could be ported into `heroku/docker-registry-client` in the future - e.g. construction of the registry URL and logic for maintaining schema v1/v2 compatibility.

radeksimko:
As @xsellier rightly mentioned in https://github.com/hashicorp/terraform/pull/6507#issuecomment-231601088 there are some cases where multiple different registries can be used to deploy a set of containers onto the same docker host. That weakens my arguments for `provider` separation as you would need `N * 2` provider blocks (`N` being number of unique registries) if you wanted to use that data source + `docker_image` resource to pull each image.

In that sense I don't mind keeping it under `docker` provider as long as we make `host` field `Optional`, but the rest of my arguments remain valid.

kyhavlov:
Ok, so just to clarify, I can replace `registry_id` and `keep_updated` with a `trigger_pull` list, change the docker provider to make the host field optional for people who just want the data source, and update the docs to explain how to keep the image updated with the data source rather than the `keep_updated` field.

builtin/providers/docker/data_source_docker_image.go
```diff
@@ -0,0 +1,78 @@
...
+	return &schema.Resource{
+		Read: dataSourceDockerImageRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"id": &schema.Schema{
```
radeksimko: I believe that `id` is in the schema by default settable via `d.SetId()` and readable via `d.Id()` - I'm actually surprised the internal schema validation isn't erroring out. It should probably be catching such cases.

radeksimko: In case you'll have any issues with using `d.SetId()` for an actual image hash I would suggest adding a new field called `digest` or `sha256_digest` as that's what it's called in the latest Docker API schema.

builtin/providers/docker/data_source_docker_image.go
```diff
@@ -0,0 +1,78 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceDockerImageRead(d *schema.ResourceData, meta interface{}) error {
+	pullOpts := parseImageOptions(d.Get("name").(string))
+	auth := getAuthConfig(pullOpts)
```
radeksimko: Would you mind moving this auth logic (basically from this line `32` to `57`) into provider as discussed in the other linked PR? Also the registry URL should be settable via provider field, e.g. `registry_url` or something along those lines.

radeksimko:
@kyhavlov Sounds 👌  Besides the comments about provider-centric auth - assuming we're keeping the data source inside `docker` provider, shall we name the new data source something like `docker_registry_image` to make it clear that no image is being pulled and it's only metadata from the Registry API?

Any suggestions on how to make the separation more obvious are welcomed.
Feel free to ping me if you want to discuss the solution or review updated PR.

kyhavlov:
@radeksimko I'm mostly done making changes to address PR comments, just need to figure out how we want to add acceptance tests for the registry auth stuff.

radeksimko:
@kyhavlov for the purpose of a code review it would be very helpful if you could either submit the dependency (`0989ee66bab1e0309b34d1bb389f1e28da3fdd47`) as a separate PR - I can merge that straight away - or squash the other three commits into a single one.

Acceptance tests discussed over IRC. 😃

kyhavlov:
@radeksimko I squashed the non-vendor commits and added acceptance tests for pulling/reading an image from a private registry.

radeksimko:
Thanks for squashing commits.

_To share what we discussed w/ @kyhavlov over IRC in the spirit of transparency:_

I started testing this PR with some private registry providers to see how well does it work with those.

 - Docker Hub (`registry.hub.docker.com` / `index.docker.io` / `registry-1.docker.io`)
 - AWS ECR (`<ACCOUNT_ID>.dkr.ecr.us-east-1.amazonaws.com`)
 - Google GCR (`gcr.io/project-id/image`)
 - Artifactory (`custom-org.artifactoryonline.com/custom-repo` / `custom-repo.jfrog.io/custom-repo`)
 - Quay.io (`quay.io/repo/image`)

and I achieved different levels of success per registry provider for different reasons. I reckon either Heroku don't stick to the Registry API strictly or those registry providers don't (maybe both or maybe the API definition isn't as strict as it should be).

I would be happy to go ahead with a limited support (e.g. only Docker Hub, which currently just works), but I'm afraid we will need to rethink the auth mechanisms completely. I would be especially worried about docker config parser & association of auth w/ normalised hostnames.

All of the above makes both of us think it would be best to focus on **getting data source merged for public images only** for now - i.e. take baby 👶  steps 👣  .

Thinking about support public registry only in this iteration we also discussed the possibility of removing the `heroku` library as dependency completely and relying on raw `net/http` & `encoding/json` for now.

@kyhavlov Feel free to add anything I may have missed from our conversation.

kyhavlov:
I removed the auth portions from the PR and replaced the heroku docker-registry-client lib with some code to get an image sha256 digest from a given registry endpoint (including the flow for oauth). So far I've tested it with a private registry running v.2.4.0, the Docker Hub, Google GCR, and Quay.io.

radeksimko:
It only works for Docker Hub unless the URL explicitly specifies a port number, but I'm ok to call this 1st iteration as we discussed earlier.

The map below describes current behaviour of `parseImageOptions`:

```go
cases := map[string]dc.PullImageOptions{
		"alpine":                                       dc.PullImageOptions{Repository: "alpine", Registry: "", Tag: ""},
		"alpine:latest":                                dc.PullImageOptions{Repository: "alpine", Registry: "", Tag: "latest"},
		"gcr.io/google_containers/busybox":             dc.PullImageOptions{Repository: "gcr.io/google_containers/busybox", Registry: "", Tag: ""},
		"gcr.io/google_containers/pause:0.8.0":         dc.PullImageOptions{Repository: "gcr.io/google_containers/pause", Registry: "", Tag: "0.8.0"},
		"gcr.io:443/google_containers/pause:0.8.0":     dc.PullImageOptions{Repository: "gcr.io:443/google_containers/pause", Registry: "gcr.io:443", Tag: "0.8.0"},
		"https://gcr.io/google_containers/busybox":     dc.PullImageOptions{Repository: "https://gcr.io/google_containers/busybox", Registry: "https:", Tag: "latest"},
		"library/alpine":                               dc.PullImageOptions{Repository: "library/alpine", Registry: "", Tag: ""},
		"library/alpine:latest":                        dc.PullImageOptions{Repository: "library/alpine", Registry: "", Tag: "latest"},
		"registry.hub.docker.io/library/alpine:latest": dc.PullImageOptions{Repository: "registry.hub.docker.io/library/alpine", Registry: "", Tag: "latest"},
	}
```

I will merge this + document the limitations.

Thank you for your patience and persistence! 😄

PR #7004, opened 2016-06-03
provider/aws: Fix panics in datasource AMI

catsby:
`TestAccAWSAmiDataSource_owners` is throwing panics:

- https://travis-ci.org/hashicorp/terraform/jobs/134931713

The SnaphotID was coming back nil. This PR checks that and conditionally adds it. 

@vancluever / ~~@stack72~~ @jen20  did either of you come across something like this in https://github.com/hashicorp/terraform/pull/6911 ?

### DIFF
--- a/builtin/providers/aws/data_source_aws_ami.go
+++ b/builtin/providers/aws/data_source_aws_ami.go
@@ -345,7 +345,6 @@ func amiBlockDeviceMappings(m []*ec2.BlockDeviceMapping) *schema.Set {
 			ebs := map[string]interface{}{
 				"delete_on_termination": fmt.Sprintf("%t", *v.Ebs.DeleteOnTermination),
 				"encrypted":             fmt.Sprintf("%t", *v.Ebs.Encrypted),
-				"snapshot_id":           *v.Ebs.SnapshotId,
 				"volume_size":           fmt.Sprintf("%d", *v.Ebs.VolumeSize),
 				"volume_type":           *v.Ebs.VolumeType,
 			}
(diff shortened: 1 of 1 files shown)


catsby:
```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAmiDataSource -timeout 120m
=== RUN   TestAccAWSAmiDataSource_natInstance
--- PASS: TestAccAWSAmiDataSource_natInstance (9.37s)
=== RUN   TestAccAWSAmiDataSource_windowsInstance
--- PASS: TestAccAWSAmiDataSource_windowsInstance (10.78s)
=== RUN   TestAccAWSAmiDataSource_instanceStore
--- PASS: TestAccAWSAmiDataSource_instanceStore (9.90s)
=== RUN   TestAccAWSAmiDataSource_owners
--- PASS: TestAccAWSAmiDataSource_owners (30.48s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	60.548s

```

stack72:
LGTM!

vancluever:
@catsby looks good over here too. I never came across this in any of the testing that I have done on the resource or data source, and I wrote the acceptance tests hoping that I got good coverage of most of the types of AMI configuration. Interesting that there's an AMI that has an EBS volume that doesn't have a snapshot that's associated with it!

Incidentally I think that the `TestAccAWSAmiDataSource_owners` might actually be a good gauge of how resilient the data source will be - the test is general enough that it may come upon these kinds of things in the future. Possibly if having so many attributes is too fragile, we can reduce the scope (at the time I just wanted to make sure I didn't miss anything but in retrospect I wonder how much is actually useful in a TF context).

PR #7007, opened 2016-06-03
core: Remove tests from provider binaries

jen20:
These tests run each time Travis builds, causing additional noise and a (negligible) speed decrease. However, since the advent of internal plugins, these tests are unnecessary, and each file only carries a package declaration anyway - so there are no tests actually executed!


jen20:
cc @phinze

phinze:
Oh hey. Nice catch. Less code yay.

PR #7009, opened 2016-06-03
Fix Google Cloud Two-Tier Example

cblecker:
There were a few issues with the Google Two-Tier example as provided in the Terraform repo:
- The credentials file path variable in the example file didn't match the real variable being called. Fixed the example file.
- The installation script calls a number of privileged commands, but logs in with an unprivileged user. Added a sudo to the inline remote_exec call.
- Bumped the Ubuntu image to the latest for 14.04-trusty


stack72:
Hi @cblecker 

Thanks for the PR here - this LGTM!

P.

PR #7010, opened 2016-06-03
Update examples using concat with strings

mbainter:
Since using concat() to join strings is deprecated and soon to be removed, this patch updates the sha examples to use interpolation instead.


jen20:
Good call - thanks @mbainter!

PR #7012, opened 2016-06-04
OpenStack LBaaS v2 Support

dkalleg:
Gives support for LBaaS V2.  Relies on a recent PR that's been merged to GopherCloud master:  https://github.com/rackspace/gophercloud/pull/575


dkalleg:
Can anyone give some guidance on how to bump the gophercloud version up to 67139b9485d6fd682c5314e963b0915e18f7947a ?  Once that is done the integration tests should be able to build and run.

jtopjian:
@dkalleg This is so awesome - thank you!

I did a quick read through the code and it looks great. There's some nits with regard to the embedded formatting in the tests, but that always happens and wouldn't block a merge. I think this should be able to get merged in time for 0.7.

With regard to the gophercloud version, Terraform _just_ switched from `godep` to `govender`. [This diff](https://github.com/hashicorp/terraform/commit/f91d870537c248842433bf393941541fb0b67302) shows the difference between the two. If you'd like to give it a shot, go for it, and we can get someone to approve it. Or, if you'd like, I can try it out in a separate PR.

builtin/providers/openstack/provider.go
```diff
@@ -105,6 +105,11 @@ func Provider() terraform.ResourceProvider {
 			"openstack_lb_monitor_v1":                  resourceLBMonitorV1(),
 			"openstack_lb_pool_v1":                     resourceLBPoolV1(),
 			"openstack_lb_vip_v1":                      resourceLBVipV1(),
+			"openstack_lbaas_loadbalancer_v2":          resourceLoadBalancerV2(),
```
jtopjian: I get that LBaaS v1 used the `/lb/` URL and LBaaS v2 uses the `lbaas` URL, so I understand why you chose to use `lbaas` in the name... but I wonder if the resources should keep `lb` and `LB` in their names for consistency... Thoughts?

dkalleg: I have no strong feeling either way.  Unless someone else chimes in before Monday, I'll push an update for this.

dkalleg:
Not sure what the conflict github is reporting is.  Side effect of changing a vendor project? Thoughts?

jtopjian:
@dkalleg It looks like there were 117 files pulled in from gophercloud. That seems like a very large number... I would imagine <10 just for the lbaasv2 stuff. Maybe that's where the conflict is coming from?

dkalleg:
@jtopjian My intention was to bring up gophercloud to its latest revision, which would include a lot of other changes.  Are you recommending I try to pull in just the lbaasv2 gophercloud changes?

jtopjian:
@dkalleg hm... my understanding from using Godep was that the PR contains the diff between what's currently in `vendor` and what's required to bring it up to date.

jtopjian:
@dkalleg ah! @phinze pointed out that the hundred-some files all look like `_test.go` files. Godep was nice enough to strip those out, but it seems Govendor is adding them back in.

Can you check and see if you're using the latest Govendor or if there is a flag to not add in the test files?

dkalleg:
@jtopjian I removed the _test.go files from the gophercloud update, down to 38 files changed from 117 :) .  I also refactored the resource naming from lbaas to lb.

vendor/vendor.json
```diff
@@ -1,5 +1,6 @@
 {
 	"comment": "",
+<<<<<<< 8c3ff933951a5ab0111c0ef6b3553e42e4ce14af
```
jtopjian: This looks leftover from a merge conflict?

jtopjian:
@dkalleg Thank you for your patience with this one! It's never fun to be caught in between a core refactor like the current Godep to Govendor one right now...

vendor/vendor.json
dkalleg: Yep, sorry.  Fixed.

dkalleg:
@jtopjian No worries!  Was mostly a painless process and taught me about govendor.

jtopjian:
@dkalleg I'm in the process of setting up the [OpenStack acceptance test environment](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/devstack/deploy.sh) for LBaaS v2. Once that's done, I'll run these acceptance tests through it.

In the meantime, I'm going to do a more thorough review of the code and make some inline comments.

builtin/providers/openstack/resource_openstack_lbaas_listener_v2.go
```diff
@@ -0,0 +1,317 @@
...
+	d.Set("name", listener.Name)
+	d.Set("protocol", listener.Protocol)
+	d.Set("tenant_id", listener.TenantID)
+	d.Set("description", listener.Description)
+	d.Set("protocol_port", listener.ProtocolPort)
+	d.Set("admin_state_up", listener.AdminStateUp)
+	d.Set("default_pool_id", listener.DefaultPoolID)
+	d.Set("connection_limit", listener.ConnLimit)
+	d.Set("sni_container_refs", listener.SniContainerRefs)
+	d.Set("default_tls_container_ref", listener.DefaultTlsContainerRef)
```
jtopjian: For `d.Set` to work correctly, the arguments must have `Computed: true`.

builtin/providers/openstack/resource_openstack_lbaas_listener_v2_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+func testAccCheckLBV2ListenerDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
+	networkingClient, err := config.networkingV2Client(OS_REGION_NAME)
+	if err != nil {
+		return fmt.Errorf("(testAccCheckLBV2ListenerDestroy) Error creating OpenStack networking client: %s", err)
+	}
+
+	for _, rs := range s.RootModule().Resources {
+		log.Printf("[FINDME] rs TYPE is: %T", rs)
```
jtopjian: Looks like this might have been some leftover debugging?

builtin/providers/openstack/resource_openstack_lbaas_loadbalancer_v2.go
```diff
@@ -0,0 +1,258 @@
...
+	log.Printf("[DEBUG] Retreived OpenStack LoadBalancerV2 %s: %+v", d.Id(), lb)
+
+	d.Set("name", lb.Name)
+	d.Set("description", lb.Description)
+	d.Set("vip_subnet_id", lb.VipSubnetID)
+	d.Set("tenant_id", lb.TenantID)
+	d.Set("vip_address", lb.VipAddress)
+	d.Set("admin_state_up", lb.AdminStateUp)
+	d.Set("flavor", lb.Flavor)
+	d.Set("provider", lb.Provider)
```
jtopjian: ditto on `Computed: true`.

builtin/providers/openstack/resource_openstack_lbaas_loadbalancer_v2.go
```diff
@@ -0,0 +1,258 @@
...
+	if d.HasChange("name") {
+		updateOpts.Name = d.Get("name").(string)
+	}
+	if d.HasChange("description") {
+		updateOpts.Description = d.Get("description").(string)
+	}
+	if d.HasChange("admin_state_up") {
+		asu := d.Get("admin_state_up").(bool)
+		updateOpts.AdminStateUp = &asu
+	}
```
jtopjian: `admin_state_up` is configured with `ForceNew: true`, so this block here would never be evaluated. I'd say either `ForceNew` or this block needs removed.

builtin/providers/openstack/resource_openstack_lbaas_listener_v2.go
```diff
@@ -0,0 +1,317 @@
...
+			for _, v := range raw.([]interface{}) {
+				sniContainerRefs = append(sniContainerRefs, v.(string))
+			}
+		}
+		updateOpts.SniContainerRefs = sniContainerRefs
+	}
+	if d.HasChange("admin_state_up") {
+		asu := d.Get("admin_state_up").(bool)
+		updateOpts.AdminStateUp = &asu
+	}
```
jtopjian: Per other comment, I think `ForceNew` and `HasChange` can't be used together.

builtin/providers/openstack/resource_openstack_lbaas_loadbalancer_v2.go
```diff
@@ -0,0 +1,258 @@
...
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	var updateOpts loadbalancers.UpdateOpts
+	if d.HasChange("name") {
+		updateOpts.Name = d.Get("name").(string)
+	}
+	if d.HasChange("description") {
+		updateOpts.Description = d.Get("description").(string)
+	}
```
jtopjian: `ForceNew` set on description.

builtin/providers/openstack/resource_openstack_lbaas_loadbalancer_v2.go
```diff
@@ -0,0 +1,258 @@
...
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(d.Get("region").(string))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	var updateOpts loadbalancers.UpdateOpts
+	if d.HasChange("name") {
+		updateOpts.Name = d.Get("name").(string)
+	}
```
jtopjian: `ForceNew` set on name.

jtopjian:
@dkalleg Looks like there are three areas to work on:

1. The `FINDME` debug lines, though this is more of a nitpick
2. If you're using `d.Set()` with any argument, it must be configured with `Computed: true`.
3. Terraform won't evaluate the `HasChange` portion for aguments that are configured with `ForceNew: true`. There seems to be a lot of arguments in these resources with `ForceNew` set -- maybe a review of what needs this?

Once devstack is set up, I'll run everything through and see what the result is.

By far, though, this is a great PR! I think the above items are minor compared to the rest of the work you put into this (docs, great tests, etc).

builtin/providers/openstack/resource_openstack_lbaas_listener_v2.go
dkalleg: Can you explain what doesn't work correctly if `Computed: false`?  I ask because in testing I:
1) Create a listener w/ TF.
2) Note the "name" field (just calling out this attr arbitrarily) was properly populated.
3) Manually change the listeners name in OpenStack.
4) Call `terraform plan` with no changes to the .tf script
5) Note the plan output wants to change the listeners name from what I manually set to what is in the .tf script:  `name:   "foo" => "tf_test_listener" (forces new resource)`

I do notice that the terraform.tfstate does not show the name as "foo" after the plan.  I tried adding `Computed: true` to the name attr and re-ran through this process with the same results.

dkalleg: Also, fyi, I realized I don't need a ForceNew: true on name :) I'll fix that and go review all the other attrs.

builtin/providers/openstack/resource_openstack_lbaas_listener_v2.go
```diff
@@ -0,0 +1,317 @@
...
+
+	"github.com/rackspace/gophercloud"
+	"github.com/rackspace/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners"
+)
+
+func resourceListenerV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceListenerV2Create,
+		Read:   resourceListenerV2Read,
+		Delete: resourceListenerV2Delete,
```
dkalleg: @jtopjian I just realized, I missed adding the `Update: ` entry here!  But.. how could the test cases have passed if I missed this?  Fixing this in my next patch.

builtin/providers/openstack/resource_openstack_lbaas_listener_v2_test.go
dkalleg: Whoopsie :)  Removing.

builtin/providers/openstack/resource_openstack_lbaas_loadbalancer_v2.go
dkalleg: Good catch.  I'm taking ForceNew out of the schema definition.

dkalleg:
@jtopjian I fixed #1 and 3.  I posted a question above about your #2 comment.  Thanks for reviewing!

jtopjian:
@dkalleg Re #2: I ran some tests and fiddled with some code and surprised myself - it looks like `d.Set` works just fine on attributes that aren't computed. I was always under the impression that the two had to go together. Sorry about that.

I was able to get a working devstack installation with LBaaS. I found that I had to increase the timeout of the `resource_openstack_lb_loadbalancer_v2` resource because it takes a while for the vm to start and provision. IMO, raising it from 2 minutes to 20 minutes is probably a safe thing to do.

Once I made that change, all tests passed!

But I also noticed that the files were still called `resource_openstack_lbaas_*.go` - any chance you can rename these to the `lb` version?

Once that's done, I'd say squash everything into two commits: one for the Terraform resources and one for the Govendor dependencies. Then I'll see if someone else can sign off on Govendor stuff just to make sure the commit doesn't mess anything up.

_Then_ we should be good to go. 😄

dkalleg:
@jtopjian Done

vendor/vendor.json
jtopjian: Ping on this one. Looks like it still exists?

jtopjian:
@dkalleg Looks like that one line is still floating around `vendor.json`:

https://github.com/hpcloud/terraform/blob/6aa655fde4aa85ac90eff9efa6e2b8f30a79c35d/vendor/vendor.json#L3

dkalleg:
@jtopjian Ah, sorry.  Should be fixed now.

phinze:
Great work on this @dkalleg! And thanks as always @jtopjian for the review. 👍

dkalleg:
@jtopjian @phinze Thanks!

PR #7015, opened 2016-06-04
#7013 add tls config support to consul provider

### LINKED ISSUE #7013: Add support for setting TLS configuration in Consul provider
I would like to have the ability to do client authentication when using the Consul provider in Terraform. I'm thinking something like this would do just fine:

    {
        "consul": {
            "address": "my.consul.server",
            "datacenter": "dc0",
            "scheme": "https",
            "tls": {
                "ca_file": "ca-bundle.cert.pem",
                "cert_file": "terraform.cert.pem",
                "key_file": "terraform.key.pem"
            }
        }
    }

maxenglander:
Resolves #7013.

Note that I updated the vendored Consul API by running `govendor fetch github.com/hashicorp/consul/api`.


stack72:
Hi @maxenglander 

Thanks for the PR here, sorry it has taken so long to get to it! I think in order to merge it, we would really love to see some updates to the documentation to reflect this. Have you been able to test that this works as expected? Could you possibly also turn in an acceptance test as well?

Thanks

Paul

maxenglander:
Hey @stack72 thanks, and no worries! Will add documentation. Have been using it in production for some time, but will also add an acceptance test. Will ping again when the changes are in.

stack72:
Awesome! Thanks :)

maxenglander:
Hey @stack72, I created a test that validates the provider config when `ca_file`, `key_file` and `ca_cert` are used. This test does not connect to a Consul server.

It's also possible to run all of the existing Consul acceptance tests so that any which connect to a Consul agent will use TLS. This requires first setting some environment variables, and pointing the Consul provider to an agent that serves (and optionally authenticates) HTTPS requests. I included a README which will hopefully make this clear.

I hope this is satisfactory - let me know if there's anything I can do to improve this.

builtin/providers/consul/config.go
```diff
@@ -26,9 +30,21 @@ func (c *Config) Client() (*consulapi.Client, error) {
 	if c.Scheme != "" {
 		config.Scheme = c.Scheme
 	}
+
+	tlsConfig := &consulapi.TLSConfig{}
+	tlsConfig.CAFile = c.CAFile
+	tlsConfig.CertFile = c.CertFile
+	tlsConfig.KeyFile = c.KeyFile
+	cc, err := consulapi.SetupTLSConfig(tlsConfig)
```
stack72: Can these values be empty? Or should we check they are set in the config object before being set as a tlsConfig object?

builtin/providers/consul/config.go
```diff
@@ -26,9 +30,21 @@ func (c *Config) Client() (*consulapi.Client, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	config.HttpClient.Transport.(*http.Transport).TLSClientConfig = cc
```
stack72: what happens if the TLSClientConfig object is empty?

builtin/providers/consul/resource_provider.go
```diff
@@ -20,11 +20,34 @@ func Provider() terraform.ResourceProvider {
 			"address": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{
+					"CONSUL_ADDRESS",
+					"CONSUL_HTTP_ADDR",
```
stack72: We currently check for this (CONSUL_HTTP_ADDR) in the provider_test but not CONSUL_ADDRESS - should we also do that?

stack72:
Hi @maxenglander 

So i ran the tests here and they look good. I have a few small questions inline - so if you can have a think about those before i merge, then that would be awesome

```
TF_ACC=1 go test ./builtin/providers/consul -v -run= -timeout 120m
--- PASS: TestAccDataConsulKeys_basic (0.03s)
=== RUN   TestAccConsulAgentService_basic
--- PASS: TestAccConsulAgentService_basic (0.02s)
=== RUN   TestAccConsulCatalogEntry_basic
--- PASS: TestAccConsulCatalogEntry_basic (0.03s)
=== RUN   TestAccConsulKeyPrefix_basic
--- PASS: TestAccConsulKeyPrefix_basic (0.08s)
=== RUN   TestConsulKeysMigrateState
--- PASS: TestConsulKeysMigrateState (0.00s)
=== RUN   TestConsulKeysMigrateState_empty
--- PASS: TestConsulKeysMigrateState_empty (0.00s)
=== RUN   TestAccConsulKeys_basic
--- PASS: TestAccConsulKeys_basic (0.05s)
=== RUN   TestAccConsulNode_basic
--- PASS: TestAccConsulNode_basic (0.03s)
=== RUN   TestAccConsulService_basic
--- PASS: TestAccConsulService_basic (0.03s)
=== RUN   TestResourceProvider
--- PASS: TestResourceProvider (0.00s)
=== RUN   TestResourceProvider_impl
--- PASS: TestResourceProvider_impl (0.00s)
=== RUN   TestResourceProvider_Configure
--- PASS: TestResourceProvider_Configure (0.00s)
=== RUN   TestResourceProvider_ConfigureTLS
--- PASS: TestResourceProvider_ConfigureTLS (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/consul	0.289s
```

Thanks for all the work here!

Paul

builtin/providers/consul/config.go
maxenglander: `consul/api.TLSConfig{CAFile,CertFile,KeyFile}` are all `string`. Their default values are `""`, so checking to see if `terraform/consul.TLSConfig.{CAFile,CertFile,KeyFile}` (also `string`) are `""` before assigning is unnecessary.

(Answer to your other question below.)

builtin/providers/consul/config.go
maxenglander: Here's my thinking:

   1. By [#L37](https://github.com/hashicorp/terraform/pull/7015#discussion-diff-74521998R37), we have a zero-valued object of `consul/api.TLSConfig`, the equivalent of `&consulapi.TLSConfig{}`
   1. Passing the above into [`consul/api.SetupTLSConfig`](https://github.com/hashicorp/consul/blob/c2cbbdec3e236d325363506cc3181b58613483fb/api/api.go#L245) produces a zero-valued `tls.Config`, the equivalent of `&tls.Config{}`
   1.  Assigning the above to `net/http.Transport.TLSClientConfig` does indeed change its default value (`<nil>`), however, according to the [go docs](https://github.com/golang/go/blob/master/src/net/http/transport.go#L113):

> 	// TLSClientConfig specifies the TLS configuration to use with
	// tls.Client. If nil, the default configuration is used.

So, in the end, what we get when we assign an empty `tls.Config` to `net/http.Transport.TLSClientConfig` is the same configuration that Go [chooses at runtime](https://github.com/golang/go/blob/7a622740655bb5fcbd160eb96887032314842e6e/src/net/http/transport.go#L1014) if we didn't do the assignment.

builtin/providers/consul/resource_provider.go
maxenglander: Yes! Done in 10cca03.

stack72:
@maxenglander thanks for the very good explanations :) the changes work on a non-tls enabled cluster so we look good to merge!

maxenglander:
@stack72 Awesome - thanks for the review and merge!

PR #7023, opened 2016-06-05
Add vsphere virtual machine migration (Fixes #7008)

### LINKED ISSUE #7008: Upgrading to v0.6.16 will delete and rebuild all vSphere hosts
Upgraded to `Terraform v0.6.16` .  Changelog mentions:
provider/vsphere: Add skip_customization option to vsphere_virtual_machine resources (#6355)

All of the resources of this vsphere  are showing:
```
 skip_customization:                     "" => "false" (forces new resource)
```
even explicitly setting `skip_customization = ""` doesn't help.   All of the servers will be delete and rebuild -- data loss.

markpeek:
Fixes vsphere virtual machine migration issues with disk controller_type and skip_customization.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -111,6 +111,9 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 		Update: resourceVSphereVirtualMachineUpdate,
 		Delete: resourceVSphereVirtualMachineDelete,
 
+		SchemaVersion: 1,
+		MigrateState:  resourceVSphereVirtualMachineMigrateState,
+
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine_migrate.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine_migrate.go
@@ -1,0 +1,58 @@
+package vsphere
+
+import (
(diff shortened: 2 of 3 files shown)


markpeek:
@chrislovecnm @dkalleg @thetuxkeeper please review this migration change

stack72:
LGTM! Nice fix here :)

thetuxkeeper:
Looks good!
Nice feature you're using. Will keep that in mind if I change something that needs state migration (like the one you fixed :) ).
Just one question. What if we had a new version "2" (or let's say "10" in a year or two :) ), what would you have to change? Is the migration function called for each version or only once?
If only once, is it possible to prepare it for multiple version upgrades/migrations? Something like 
```
while schema_version < target_version 
{
    migrate(current_version, current_version+1)
}
```
I think all at once would be too much at some time in the future, so my idea would be to do it step by step from one version to the next until we reach the latest one.
What do you think?

chrislovecnm:
@markpeek thanks for the work!!  

@thetuxkeeper you have time to run the tests?  Otherwise let me know, and I will run them this eveningish :P

markpeek:
@thetuxkeeper Here's an [example](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance_migrate.go#L13) on how it is done for the Google Compute Instance. Note the fallthrough between the cases in the switch statement.

thetuxkeeper:
@markpeek : ah, I see. Thanks for the example, didn't know that there's such a statement :)
@chrislovecnm : not today, I'll let them run tomorrow, won't hurt if we both let them run. :)

PR #7026, opened 2016-06-06
Prevent overwriting states with other unrelated states

apparentlymart:
Through real-world use of Terraform in production I have encountered [a variety of "gotchas" relating to state management](https://gist.github.com/apparentlymart/657885e730d1e5abc6ea). This PR is an attempt to tackle one of these where I think I have a reasonable path forward: protecting against overwriting an remote states.

One common way this occurs is through clumsy use of `terraform remote config`. One variant of this is to forget to enable remote state and then try to enable it afterwards:
- `terraform apply`
- oh whoops! I forgot to enable remote state and now I have duplicated my resources
- `terraform destroy` (deletes all of the stuff in the local state, leaving an empty state)
- `terraform remote config ...`
- whoops! I've now overwritten that remote state with my empty local state.

Another variant is when switching from one remote to another:
- `terraform remote config ...`
- `terraform apply`
- `terraform remote config ...`
- whoops! I've now overwritten the second remote state with the contents of the first.

---

This PR first introduces the idea  of "state lineage", which allows us to determine when one state is an earlier or later version of another state vs. when the two states are entirely unrelated. Comparing `Serial` is only expected to produce meaningful results when `Lineage` matches.

Building on this, when syncing local and remote state, fail with an error if the two states are of different lineage. This includes running `terraform remote config` to enable or re-configure remote state when a local state is already present.

As a special case, if local cache has different lineage but it is entirely empty of resources then just silently drop it and replace it with the remote state, since the loss of an _empty_ state is trivial to recover from and this will reduce friction when users are bootstrapping a new config.

The primary goal here is to return an error when the user seems to be accidentally doing something dangerous, with as little impact as possible to "legitimate" workflows. In future version of Terraform we may make more fundamental changes to these features to help the user not make these mistakes in the first place, but this is intended as a short-term fix to reduce the risk of a state-related catastrophe.

This is the same set of changes as from #6540 except that the change to prevent applying a stale plan is removed, since it no longer works in light of the `plan` behavior change introduced in #6811.

### DIFF
--- a/command/remote_config_test.go
+++ b/command/remote_config_test.go
@@ -406,6 +406,128 @@ func TestRemoteConfig_enableRemote(t *testing.T) {
 	testRemoteLocalBackup(t, true)
 }
 
+func TestRemoteConfig_conflictRemote(t *testing.T) {
+	tmp, cwd := testCwd(t)
+	defer testFixCwd(t, tmp, cwd)
+
+	s := terraform.NewState()
+	s.Serial = 5
+	// Conflict is detected only when state has resources
+	s.Modules = []*terraform.ModuleState{
+		{
+			Path: []string{"root"},
+			Resources: map[string]*terraform.ResourceState{
+				"test_instance.foo": {},
+			},
+		},
+	}
+	conf, srv := testRemoteState(t, s, 200)
+	defer srv.Close()
+
+	s = s.DeepCopy()
+	// Different lineage locally means this is a conflict, because remote
+	// already exists but is describing an unrelated state lineage.
+	s.Lineage = "different-lineage"
+	s.Remote = conf
+
+	// Store the local state
+	statePath := filepath.Join(tmp, DefaultDataDir, DefaultStateFilename)
+	if err := os.MkdirAll(filepath.Dir(statePath), 0755); err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	f, err := os.Create(statePath)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	err = terraform.WriteState(s, f)
+	f.Close()
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	ui := new(cli.MockUi)
+	c := &RemoteConfigCommand{
+		Meta: Meta{
+			ContextOpts: testCtxConfig(testProvider()),
+			Ui:          ui,
+		},
+	}
+	args := make([]string, 0, 5)
+	args = append(args, "-backend="+conf.Type)
+	for bck, bcv := range conf.Config {
+		args = append(args, "-backend-config", bck+"="+bcv)
+	}
+	if code := c.Run(args); code == 0 {
+		t.Fatalf("bad: got success, but want error")
+	}
+}
+
+func TestRemoteConfig_localEmptyRemoteExists(t *testing.T) {
+	tmp, cwd := testCwd(t)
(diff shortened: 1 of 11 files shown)


bigkraig:
Yes!!

phinze:
This LGTM on initial review - would love to get @mitchellh's thoughts on overall history and UX here as well as @jen20's eyes since he's been spending a lot of time upgrading state for 0.7!

bigkraig:
@phinze @jen20 Since we are talking about state and 0.7, how about a look at #6732 too? It's preventing us from using 0.7

command/remote_config_test.go
```diff
@@ -406,6 +406,128 @@ func TestRemoteConfig_enableRemote(t *testing.T) {
...
+	}
+	f, err := os.Create(statePath)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	err = terraform.WriteState(s, f)
+	f.Close()
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
```
mitchellh: The helper `testStateFile` should be used here to save you a lot of lines and cleanup. :)

command/remote_config_test.go
```diff
@@ -406,6 +406,128 @@ func TestRemoteConfig_enableRemote(t *testing.T) {
...
```
mitchellh: Same as above.

command/state.go
```diff
@@ -274,3 +282,30 @@ func remoteStateFromPath(path string, refresh bool) (*state.CacheState, error) {
...
+ran concurrently and have created divergent states. In this case, some
+resources may be duplicated across the two states.
+
+In either case, caution is advised. Your local state cache is preserved
+at .terraform/terraform.tfstate and it will not overwrite the remote until
+the confict is resolved.
+
+At this point it is safe to disable remote state, which will retain the
+local state on this computer and leave the remote state untouched:
+    terraform remote config -disable -pull=false
```
mitchellh: I would reword this error message to provide more actions. My questions as a user would be:

* I definitely want my local state to override the remote state, what do I do?

* I definitely want my remote state to override my local state, what do I do?

phinze:
@bigkraig i know @jen20 was working on remote state today - i'll make sure we address that issue before the next RC 👍

mitchellh:
I made a few comments that need to be addressed. One other thing I'd recommend is actually documenting some of this behavior in perhaps a "state" internals page for: https://www.terraform.io/docs/internals/index.html

That's probably outside the scope of this PR but its something that I think that should be documented soon since the state is slowly becoming more complex [to protect users]

apparentlymart:
Thanks for all the feedback! It looks like the feedback was on the remote state checking part, and I probably won't have time to iterate on this one more time before 0.7 final is cut, so I split the lineage part off into #7109 in the hope that it can get included in 0.7 in isolation. That will have no effect on the UX but will get this in the state format so that the UX change(s) can follow in a subsequent releases.

lbernail:
I overwrote a remote state earlier today by mistake
I was able to recover thanks to versioning on s3

When trying to understand when a bogus local state would overwrite the remote one (my understanding is that today it is only based on serial number) I discovered the new lineage parameter in the state file which lead me to this PR

I must say that this will be *really* helpful!

apparentlymart:
I believe there are some more fundamental changes to remote state coming in the near future, and this is a pretty old patch anyway, so I'm going to close this out for now and maybe revisit this later once it's clearer how state management is changing.

PR #6893, opened 2016-05-26
Adding Govomi Debug Logging Capability and refactoring of integration tests

chrislovecnm:
Also refactored some of the integration tests.

PR #7027, opened 2016-06-06
Don't require vsphere debug paths to be set

markpeek:
The change to add govmomi debug paths (#6893) required user input even
when the env was not set. This change defaults those values appropriately.


markpeek:
@chrislovecnm please review

chrislovecnm:
Nice catch ... LGTM!!

chrislovecnm:
@markpeek do we have any others of these running around.  Thanks again for the catch ... still learning golang!

PR #7037, opened 2016-06-06
provider/cloudflare: Upgrade library to github.com/cloudflare/cloudflare-go

catsby:
This upgrades our CloudFlare provider to use the official upstream library 

- https://github.com/cloudflare/cloudflare-go

Note that at the time of this PR the above library is under active development:

> Note: This library is under active development as we expand it to cover our (expanding!) API. Consider the public API of this package a little unstable as we work towards a v1.0.


phinze:
LGTM

PR #7040, opened 2016-06-06
Clarify HTTP URL module source docs

jritsema:
Updated [Module Sources](https://www.terraform.io/docs/modules/sources.html) website documentation to be more clear about how `HTTP URL` module sources work.  The following line in the documentation seems to indicate that you can serve up modules over HTTP.  Also the fact that `HTTP URLs` is listed alongside the other options makes this misleading.

> Any HTTP endpoint can serve up Terraform modules.

I made a tweak to indicate that HTTP URLs are just redirects to other module sources rather than being an actual module source in and of itself.

### DIFF
--- a/website/source/docs/modules/sources.html.markdown
+++ b/website/source/docs/modules/sources.html.markdown
@@ -29,6 +29,8 @@ Terraform supports the following sources:
 
   * HTTP URLs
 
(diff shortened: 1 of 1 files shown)


phinze:
Looks like a good clarification - thanks @jritsema!

PR #7044, opened 2016-06-07
DigitalOcean - uppercase region forces droplet recreation

bchelli:
The region returned by the API is always lowercase therefore when you specify a region uppercase in your config file it forces the droplet to be regenerated on every ```terraform apply``` (even when it is not needed).


stack72:
Hi @bchelli 

Thanks so much for the fix here. This LGTM!

Paul

Issue #6173, opened 2016-04-14
aws s3 - disallow creating bucket names with underscore character (_)


spyrospph:
Hi,

Was trying to troubleshoot an aws EMR issue while using aws data pipeline and I bumped onto an amazing aws gotcha.

Bucket names which contain underscores "_" are allowed to be created only in the us-east-1 region and they create issues with aws products such as EMR. Specifically when EMR tries to store data to S3, it breaks because the bucket name contains an "_".

Could terraform prohibit such bucket names?

PR #7052, opened 2016-06-07
provider/aws: Add support for bucket name validation

kjmkznr:
Support bucket name validation.

https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket.go
@@ -26,9 +26,10 @@ func resourceAwsS3Bucket() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"bucket": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Required:     true,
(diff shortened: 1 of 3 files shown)


catsby:
Hey @kjmkznr thanks for the PR! Does this make any accommodations for the more relaxed restrictions on Buckets in us-east-1, mentioned not that doc page? I like this validation, but we need to be flexible for people still on us-east-1 and allowed to have more relaxed names

kjmkznr:
Yes, mentioned that doc page. And I know lax only in us-east-1 region.
I wanted this validation,  I tried to put out the pull request.
And I agree to what you say.

kjmkznr:
link #6173

catsby:
Hey @kjmkznr – I didn't mean to suggest you close your pull request, I apologize for giving that impression! 

I do like this change, we just need to think on how we'd approach this, gracefully if we can. As the doc says, this will be enforced in us-east-1 eventually. For us, it's a matter of what to do with users who are currently using the relaxed names.

kjmkznr:
@catsby 
Sorry!
I see.

It is one of the idea, What about to add a ‘ignore_validation‘ parameter?

PR #7056, opened 2016-06-07
Update vpc_peering documentation

mbainter:
The documentation wording implies that in all cases you have to manually accept peering requests.  This change is intended to clarify where this is required.  The documentation also separates between "basic usage" and "basic usage with tags", but the expanded usage didn't actually provide much additional useful information.  Expanded a bit to show the use of auto_accept since both VPCs are created by the content and to show setting the Name tag for proper display in the console.

The basic example was also just a basic example of the vpc creation stanza, so I replaced that with the simplified peering connection stanza.


stack72:
Hi @mbainter 

Thanks for the PR here. Good spot on the basic usage example - that wasn't very useful at all :)

This LGTM!

P.

Issue #7046, opened 2016-06-07
PROPOSAL: rename `parameter_group_name` to `db_cluster_parameter_group_name`


ijin:
Amazon Aurora supports options for configuring `db parameter group` and `db cluster parameter group`.

Currently, `db parameter group` is configured using the `db_parameter_group_name` argument in the `aws_rds_cluster_instance` resource and `db cluster parameter group` configured using the `parameter_group_name` argument in the `aws_rds_cluster` resource.

In other words, 

aurora option | tf argument | tf resource
------------------ | --------------- | ---------------
db parameter group | db_parameter_group_name | aws_rds_cluster_instance
db cluster parameter group | parameter_group_name | aws_rds_cluster

```
resource "aws_rds_cluster_instance" "cluster_instances" {
    db_parameter_group_name = "${aws_db_parameter_group.aurora.name}"
}

resource "aws_rds_cluster" "default" {
    parameter_group_name = "${aws_rds_cluster_parameter_group.aurora.name}"
}
```
Using `parameter_group_name` is confusing and should be renamed to something like `db_cluster_parameter_group_name`, to conform to AWS's `CreateDBCluster` API request parameter: **DBClusterParameterGroupName**.

ref: http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html

PR #7068, opened 2016-06-08
Modify parameter_group_name to db_cluster_parameter_group_name

ijin:
for #7046

### DIFF
--- a/builtin/providers/aws/resource_aws_rds_cluster.go
+++ b/builtin/providers/aws/resource_aws_rds_cluster.go
@@ -61,7 +61,7 @@ func resourceAwsRDSCluster() *schema.Resource {
 				Computed: true,
 			},
 
-			"parameter_group_name": &schema.Schema{
+			"db_cluster_parameter_group_name": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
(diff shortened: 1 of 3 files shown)


catsby:
Thanks @ijin – I opened https://github.com/hashicorp/terraform/pull/7083 that builds off of this. It keeps the old parameter name around but marks it as deprecated. I realize its likely the number of people using `master` or the release candidate and also using this parameter is probably pretty small, but I though it a nice thing to do. What do you think?

ijin:
@catsby Thanks. I just thought it would be easier to change the parameter name now before the official release of v0.7.0. Although it feels weird having a deprecation message from RC, I do understand your concerns for maintaining compatibility. How far in the future would removing the old parameter make sense? I could submit a pullreq when the time comes.

catsby:
It's absolutely easier to change it now, since v7 isn't officially out! 
I did however personally point a handful of people to _this particular_ config attribute, so I was happy to put in the little bit of extra code to make things smoother. 

I've made a self-note to remove the attribute in the future. Our versioning and release schedule is not fixed, so it's hard to say when it will happen, likely a release or three. Thanks!

ijin:
Ok, got it! 👍

PR #7070, opened 2016-06-08
provider/cloudstack: add root_disk_size

serbaut:
Add support for setting rootdisksize.

From http://cloudstack.apache.org/api/apidocs-4.8/root_admin/deployVirtualMachine.html
```
rootdisksize
	Optional field to resize root disk on deploy. Value is in GB. Only applies to template-based deployments. Analogous to details[0].rootdisksize, which takes precedence over this parameter if both are provided
```


svanharmelen:
Thanks @serbaut! LGTM!

PR #7073, opened 2016-06-08
BGP AS 60000 is assigned to EQUINIX-CLOUD-EXCHANGE-AMSTERDAM-ASN.

ottenhoff:
AS65000 is reserved for private use. 

  https://en.wikipedia.org/wiki/Autonomous_system_\(Internet\)


jen20:
Great catch @ottenhoff! Thanks for the pull request!

PR #7074, opened 2016-06-08
provider/cloudstack: do not force a new resource when updating instance user_data

serbaut:
Adds user_data_update (as an alternative to user_data) that does not
force a new resource to be created.

### DIFF
--- a/builtin/providers/cloudstack/resource_cloudstack_instance.go
+++ b/builtin/providers/cloudstack/resource_cloudstack_instance.go
@@ -122,7 +122,6 @@ func resourceCloudStackInstance() *schema.Resource {
 			"user_data": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
 				StateFunc: func(v interface{}) string {
 					switch v.(type) {
 					case string:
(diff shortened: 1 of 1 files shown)


svanharmelen:
Hi @serbaut, why would you create a dedicated attribute for this one? If we can update the `user_data` without having to rebuild the VM, then let's just remove the `ForceNew` flag from the `user_data` parameter add the needed logic to `resourceCloudStackInstanceUpdate`. 

I don't see why having two separate fields for the same thing would make any sense...

serbaut:
I added a new attribute to avoid breaking anything. If you want I can change the behavior of user_data instead.

svanharmelen:
Yes, please do that. I don't think you will break anything by changing the behaviour so it better matches the workings and behaviour of CloudStack :wink:

svanharmelen:
Sorry for the delay, LGTM! Thanks!

serbaut:
Any pointers on how to set up all variables to run the integration tests? I started but didn't get very far.  An example would probably be enough.

svanharmelen:
@serbaut I normally just export all the requested vars and off it goes. Maybe we should open a new issue for this in which you can share some more details and what errors you get?

Issue #7057, opened 2016-06-07
keys() interpolation function ordering


imbriaco:
Hi Terraform Team,

In [[1]](https://github.com/hashicorp/terraform/blob/master/config/interpolate_funcs.go#L613) I see an `interpolationFuncKeys` function defined that returns a list of keys from a map. This is very useful! 

After reviewing the implementation, it seems that the keys are sorted before they are returned. I'm wondering whether it is safe for me to rely upon this behavior or if I need to treat the order of results that I get back from `keys()` as undefined? Unfortunately, this function does not appear to be documented (though I found some confirmation in another issue that it is at least intended to be public) so I wasn't able to answer this for myself.

I'd like to advocate for defining the result of `keys()` as sorted (or at least as having consistent ordering) and updating the documentation and test cases to reflect this as necessary. My justification is that when I think about ways that I might like to make use of this, an obvious case might look something like the following.

This code is off the top of my head without a real editor, so please excuse any typos or missing parens. I think the idea should come across at least.

```
variable "subnets" {
  default = {
    us-east-1a = "1.2.3.4/24"
    us-east-1b = "1.2.3.5/24"
  }
}

resource "aws_subnet" "subnet" { 
  ...
  availability_zone = "${element(keys(var.subnets), count.index)}"
  cidr_block = "${lookup(var.subnets, element(keys(var.subnets), count.index))}"
  count = "${length(keys(var.subnets))}"
}
```

If the ordering of the list of keys were to change, this could some undesirable (and unpredictable) effects on resources that are constructed in a similar way.

Thanks!

cc: @phinze

phinze:
Thanks for filing this @imbriaco - agreed that the `keys()` and `values()` functions become much more useful if their ordering is consistent.

jen20:
The missing documentation here is not good - we will resolve that also. You can rely on the output of keys being sorted alphabetically. If you need `values()` to be in a deterministic order you can use the new `sort()` interpolation function, though I'm not necessarily opposed to making that sorted either. Thoughts @phinze?

apparentlymart:
Given that the `sort` function exists to help sorting the values by themselves, I'd vote for `values()` by default using the same ordering as `keys()`, since that way the two will be "aligned" by default.

josephholsten:
+1 to returning `values()` in matched order to `keys()`

jbardin:
This is taken care of in #7418

We added a unit test to make sure the order of keys and values don't break, and added those interpolation functions to the docs.

Issue #2143, opened 2015-05-29
Tag names containing dots are always deleted


gar1t:
This tags definition:

```
tags {                                                                      
    "foo.bar" = "123"                                                       
}
```

Will cause an existing tag named "foo.bar" to be deleted. terraform apply always shows this:

```
tags.foo.bar: "<existing value>" => ""
```

mitchellh:
There is another issue tracking this, but the unfortunate thing is that periods in tag names simply can't be supported with the current state format. It'd be a silly dance to do so. We'll add an error.

We're working on fixing this but in the short term could you not use dots? :( sorry.

gar1t:
-Shakes fist at universe- ;)

Yeah, unfortunately we're using tag names from another application and making that change would be very hard indeed as it's managing servers in independent contexts.

I'm wondering about an escaping mechanism in your names that could be used to specify unicode characters, which could be passed along to AWS [1] - crazy?

[1] http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html

gar1t:
One other thought here - if Terraform didn't insist on its tags being the entire tagset, this would not be an issue either. Terraform could concede that other applications may use tags and act only to ensure that its tagset is a correct subset. If someone wants to delete a tag explicitly, they can specify the name with an empty string.

To maintain compatibility with the current regime, you might call this thing "ExtraTags" (or similar).

tony612:
:+1:

pdf:
This limitation of the parser has wider implications too - if you need to pass keys containing periods to a provider, you have to make users do some dance where they replace periods with some other character - that must not overlap with any other valid characters - and then replace them before sending to the backend.

I'd really like to have quoted keys in maps treated as literals (syntactically preferable IMO), or failing that, allow periods to be escaped via `\.`, but I don't know how realistic either option is.

marekrogala:
:+1: 
pretty important to me

tony612:
I just wonder is there any progress for this problem? Dot is a convention for the names of our tags, so it's not easy to change :(

filete:
:thumbsup:
Also breaks out tag name convention based on dots...

bigkraig:
:+1: this is important for the api gateway

coen-hyde:
As linked above this is breaking `advanced_options` for AWS's Elasticsearch service. Would be great to get this fix.

nicolai86:
[...]

PR #7082, opened 2016-06-08
core: Make lists and maps distinguishable in state

jen20:
At a high level this PR achieves the following:

1. Introduction of state V3
2. Migration of state V2 to V3
3. Change of representation of maps from .# for the count to .%

The commit messages are significantly more detailed and should be used as a reference while reviewing.


helper/hilmapstructure/hilmapstructure.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var hilMapstructureDecodeHookEmptySlice []interface{}
+var hilMapstructureDecodeHookStringSlice []string
+var hilMapstructureDecodeHookEmptyMap map[string]interface{}
+
+// WeakDecode behaves in the same way as mapstructure.WeakDecode but has a
+// DecodeHook which defeats the backward compatibility mode of mapstructure
+// which WeakDecodes []interface{}{} into an empty map[string]interface{}. This
+// allows us to use WeakDecode (desirable), but not fail on empty lists.
```
phinze: 👍  nice

helper/schema/schema.go
```diff
@@ -1145,8 +1143,6 @@ func (m schemaMap) validateMap(
 		// If raw and reified are equal, this is a string and should
 		// be rejected.
 		reified, reifiedOk := c.Get(k)
-		log.Printf("[jen20] reified: %s", spew.Sdump(reified))
-		log.Printf("[jen20]     raw: %s", spew.Sdump(raw))
```
phinze: 😛

terraform/interpolate.go
```diff
@@ -566,49 +572,70 @@ func (i *Interpolater) interpolateComplexTypeAttribute(
...
 
-	if len(numericKeys) == len(allKeys) {
-		// This is a list
 		var members []string
-		for _, key := range numericKeys {
+		for _, key := range keys {
 			members = append(members, attributes[key])
 		}
+		// This behaviour still seems very broken to me... it retains BC but is
+		// probably going to cause problems in future
```
phinze: Interesting related convo in #7057

terraform/interpolate.go
```diff
@@ -566,49 +572,70 @@ func (i *Interpolater) interpolateComplexTypeAttribute(
...
+
 		members := make(map[string]interface{})
-		for _, key := range allKeys {
+		for _, key := range keys {
 			members[key] = attributes[resourceID+"."+key]
 		}
 		return hil.InterfaceToVariable(members)
 	}
+
+	return ast.Variable{}, fmt.Errorf("No complex type %s found", resourceID)
```
phinze: I wonder if the solution to #2143 is hiding somewhere in an unnecessarily greedy regexp matcher in here, or if it's a `helper/schema` bug instead.

Non-critical color commentary shall continue! 😄

terraform/interpolate_test.go
```diff
@@ -333,7 +334,7 @@ func TestInterpolator_resourceMultiAttributes(t *testing.T) {
 	testInterpolate(t, i, scope, "aws_route53_zone.yada.nothing",
 		interfaceToVariableSwallowError([]interface{}{}))
 
-	// Maps still need to work
+	//// Maps still need to work
```
phinze: Comment of a comment is a comment.

terraform/interpolate_test.go
```diff
@@ -527,7 +528,9 @@ func testInterpolate(
 		"foo": expectedVar,
 	}
 	if !reflect.DeepEqual(actual, expected) {
-		t.Fatalf("%q: actual: %#v\nexpected: %#v", n, actual, expected)
+		spew.Config.DisableMethods = true
+		t.Fatalf("%q:\n\n  actual: %#v\nexpected: %#v\n\n%s\n\n%s\n\n", n, actual, expected,
+			spew.Sdump(actual), spew.Sdump(expected))
```
phinze: Stray spew here - I think we've still managed to avoid actually vendoring that lib so far?

phinze: (Not that I'd mind all that much if we pulled it in.)

terraform/state.go
```diff
@@ -1379,23 +1379,32 @@ type jsonStateVersionIdentifier struct {
...
+func testForV0State(buf *bufio.Reader) error {
+	start, err := buf.Peek(len("tfstate"))
+	if err != nil {
+		return fmt.Errorf("Failed to check for magic bytes: %v", err)
+	}
+	if string(start) == "tfstate" {
+		return fmt.Errorf("Terraform 0.7 no longer supports upgrading the binary state\n" +
+			"format which was used prior to Terraform 0.3. Please upgrade\n" +
+			"this state file using Terraform 0.6.16 prior to using it with\n" +
+			"Terraform 0.7.")
```
phinze: Perfect 👍

terraform/state.go
```diff
@@ -1477,6 +1508,38 @@ func ReadStateV2(jsonBytes []byte) (*State, error) {
...
+				state.TFVersion)
+		}
+	}
+
+	// Sort it
+	state.sort()
+
+	return state, nil
+}
+
```
phinze: Ended up going back to the "each version updates to latest" impl rather than the "each version updates to next" impl?

I assume this was a "for now" thing, which I'm AOK with just clarifying.

phinze: Oh NM was misreading it's all helper functions but it is incremental. LGTM

terraform/state_upgrade_v2_to_v3.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+		}
+	}
+
+	return new, nil
+}
+
+func upgradeAttributesV2ToV3(instanceState *InstanceState) error {
+	collectionKeyRegexp := regexp.MustCompile(`^(.*\.)#$`)
+	collectionSubkeyRegexp := regexp.MustCompile(`^([^\.]+)\..*`)
```
phinze: I'm looking for a unit test to express the behavior here - am I missing where that might live?

Specifically I'm thinking about greedy / nongreedy matchers and nested complex types, e.g.

 * `aset.123.amap.#`
 * `alist.1.alist.#`

etc

terraform/interpolate.go
jen20: This seems only tangentially related? This comment (which was probably supposed to be marked as a TODO :P) relates to lists, not map keys.

phinze:
Okay! Overall PR LGTM - a few optional nits inline that are mostly drivel with maybe some good bits here and there. Would like to hear about that last testing one before we do the final merge though.

terraform/state_upgrade_v2_to_v3.go
jen20: Test to be added. I have a good test case using the test provider that excercises many combinations.

terraform/interpolate_test.go
jen20: We already have it (transitively) - these are a lot more useful for debugging failures than a `%#v`!

jen20:
@phinze I've addressed the testing and other points here, so will merge now and deal with any additional review as a post-mortem in order to unblock other work.

builtin/providers/test/resource.go
```diff
@@ -13,38 +13,94 @@ func testResource() *schema.Resource {
...
+			"list_of_map": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeMap,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+				},
+			},
```
phinze: 👏

PR #7083, opened 2016-06-08
provider/aws: Rename parameter_group_name to db_cluster_parameter_group_name (supersedes #7068)

catsby:
Build off of #7068 but still include the `parameter_group_name` for backwards compatibility. 
While `parameter_group_name` was only introduced in the v0.7.0rc, I still think it important to not break existing configs. I have a note to remove this parameter after future versions are released.

`db_cluster_parameter_group_name` takes precedence over the two.

This is a _better safe than sorry_ PR. I think it's not totally unreasonable to simply merge #7068 as-is simple because the config parameter in question was introduced in v0.7.0, however, for anyone on `master` or the release candidate and using that parameter would be greeted with a configuration error, and I thought it was easy enough to avoid with this addition 

Fixed https://github.com/hashicorp/terraform/issues/7046

### DIFF
--- a/builtin/providers/aws/resource_aws_rds_cluster.go
+++ b/builtin/providers/aws/resource_aws_rds_cluster.go
@@ -61,7 +61,7 @@ func resourceAwsRDSCluster() *schema.Resource {
 				Computed: true,
 			},
 
-			"parameter_group_name": &schema.Schema{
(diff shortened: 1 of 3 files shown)


stack72:
LGTM! Nice note -I think we will need that migration when it's removed so well remembered!

PR #7088, opened 2016-06-09
provider/vSphere - Add DiskEnableUUID option

dkalleg:
Since the custom_configuration_parameters can't take dots, we cannot
set 'disk.EnableUUID'.  This adds a parameter for this options that gets
added to a configSpec. This option causes the vm to mount disks by uuid
on the guest OS.


dkalleg:
@stack72 @phinze @chrislovecnm Could you take a look when you can?  Hoping we can squeeze this into the 0.7.0 release.

stack72:
Hi @dkalleg 

Thanks for this :) Looks good to me

P.

dkalleg:
@stack72 Wow thanks!

chrislovecnm:
Thanks Paul

PR #7090, opened 2016-06-09
docs: explain instantiating terraform module multiple times

geekq:
1. For me it was surprising, that the string after the `module` keyword is a name of the instance of the module, not a module name.
2. Lot of people see modules as a vehicle to reduce repetition (DRY) and some ask, e.g. in google groups, if is is possible, to instantiate a module multiple times

So I thought, it would be a good idea to document such a use case for modules


jen20:
Thanks @geekq! This is a definite improvement!

PR #7091, opened 2016-06-09
Serialization for hash panics on TypeMap

jbardin:
The serializeCollectionMemberForHash helper can't be called for the
MapType values, because MapType doesn't have a schema.Elem. Instead, we
can write the key/value pairs directly to the buffer. This still doesn't
allow for nested maps or lists, but we need to define that use case
before committing to it here.


phinze:
Nice! This LGTM

Issue #7043, opened 2016-06-07
DigitalOcean - droplet update


bchelli:
Hi Guys,

Thanks (again) for terraform :)

The DigitalOcean API does not support down-sizing droplet (even if it is not mentioned on the API documentation), therefore when I down-size one droplet, [terraform always fail on applying the modifications](https://github.com/hashicorp/terraform/blob/d1ac7d3b2e1be2f32092275340f33a2d73f7b31a/builtin/providers/digitalocean/resource_digitalocean_droplet.go#L261).

It feels like Terraform should not allow the update of the size of an instance and instead mark the size as a ForceNew field. I don't know if this is the best and only way to make it work, but I think this behavior is a more reliable than the current one.

Thanks,

stack72:
Hi @bchelli 

Thanks for opening the issue here. I am currently looking into this 

Paul

stack72:
@bchelli 

So the error is as follows:

```
		* digitalocean_droplet.foobar: Error resizing droplet (17090364): POST https://api.digitalocean.com/v2/droplets/17090364/actions: 422 Size can not decrease size of Droplet's disk image
```

The DO API doesn't allow sizing down, it only allows sizing up. I am not quite sure why this is the case.

There is currently a test in Terraform that allows changing size, but I will document that it only allows changing to be a bigger droplet

Thanks

Paul

bchelli:
Thanks Paul

stack72:
Closed via #7094

PR #7094, opened 2016-06-09
provider/digitalocean: Update of droplet document to add a note about resize

stack72:
When resizing a DO droplet, you can only increase the size not
descrease. If you try and go down in size, the API will return this
error:

```
 * digitalocean_droplet.foobar: Error resizing droplet (17090364):
   POST https://api.digitalocean.com/v2/droplets/17090364/actions:
   422 Size can not decrease size of Droplet's disk image
```

documenting DO droplet error as found in #7043


phinze:
LGTM 👍

PR #7107, opened 2016-06-10
make linux amd64 binaries static again.

### LINKED ISSUE #6714: Terraform v0.6.16 dynamically linked
Hi,

I'm having some issues running Terraform v0.6.16 on Alpine Linux. Version 0.6.15 and 0.6.14 run great but for some reason the binary distribution for v0.6.16 seems dynamically linked which is giving me some issues with glibc. On Alpine 3.3.3 I'm getting the following results after installing the binary:
```
/home/user # file /usr/local/bin/terraform
/usr/local/bin/terraform: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, stripped
/home/user # /usr/local/bin/terraform
sh: /usr/local/bin/terraform: not found
```

Manually compiling terraform 0.6.16 from the tag in the repo gives me the following result:
```
/home/user # file /usr/local/bin/terraform
terraform: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
/home/user # /usr/local/bin/terraform version
Terraform v0.6.16
```

Was this an intended change? If so I'll need to switch building my containers to building from source

mindw:
fixed #6714.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -35,6 +35,9 @@ if ! which gox > /dev/null; then
     go get -u github.com/mitchellh/gox
 fi
 
+# instruct gox to build statically linked binaries
+export CGO_ENABLED=0
+
 LD_FLAGS="-X main.GitCommit=${GIT_COMMIT}${GIT_DIRTY}"
 # In relase mode we don't want debug information in the binary
 if [[ -n "${TF_RELEASE}" ]]; then


phinze:
Thanks for this, @mindw! I put a bit more context in the #6714 thread but this is definitely what we want to do for our builds. 👌

PR #7109, opened 2016-06-10
core: State "Lineage" concept

apparentlymart:
The lineage of a state is an identifier shared by a set of states whose serials are meaningfully comparable because they are produced by progressive Refresh/Apply operations from the same initial empty state.

This is initialized as a type-4 (random) UUID when a new state is initialized and then preserved on all other changes.

Since states before this change will not have lineage but users may wish to set a lineage for an existing state in order to get the safety benefits it will grow to imply, an empty lineage is considered to be compatible with all lineages.

This is a single commit from the set that I originally submitted as #7026. I've split it up into its own commit because this is the part that requires changes to the state format and is thus best merged before 0.7 is cut. This patch alone makes no change to the UX and provides no safety guarantees, but adds the building block that will allow them to be added later. I will continue to work on the other parts for including in a 0.7 patch release.

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -7,12 +7,15 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"reflect"
 	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/hashicorp/go-version"
+	"github.com/satori/go.uuid"
+
 	"github.com/hashicorp/terraform/config"
 	"github.com/mitchellh/copystructure"
 )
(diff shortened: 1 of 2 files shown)


apparentlymart:
@phinze sorry for all the fussing here... but I split off the "lineage" commit from #7026 because the other part needs more work but I don't expect I will have time to do it before 0.7, and I'd like to get this part in for 0.7 if possible since it represents a change to the state format which I'd like to release along with all of the other format changes in 0.7.

phinze:
This LGTM, @apparentlymart! Thanks for being flexible around the release process here.

I was originally expecting "lineage assignment" to happen explicitly in a State version upgrade, but I think the `init()` behavior along with the zero value handling should make this implementation work just fine with older states. /cc @jen20 for a double check on that logic

In the meantime, I'm going to pull this in as-is. I'm omitting any changelog entry for now - we can add notes to the changelog when there are user-facing changes based on this data.

Issue #57, opened 2014-07-28
Store arrays in variables


gwagner:
I have lists of security groups and or lists of ec2 instances that i would like to to keep in arrays ( separated by files which are auto loaded by *.tf ) so i know when i change things in one place, it gets changed in another.  

Right now vars can only hold strings or key value maps, but not arrays (if i am reading the docs and error messages correctly)

ojongerius:
:+1:

justinclayton:
:+1:

glongman:
:+1:

kubek2k:
:+1: + same for maps

sbward:
:+1:

joescii:
:+1:, y'all

joegoggins:
+1. Looks like this is done already and slated for [release in 0.4.0](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md)

btw: the way you guys maintain your changelog.md is down right inspiring and incredibly useful. 
You don't see that often enough in the open source world--props!

mitchellh:
@joegoggins Not completely fixed but there is certainly a workaround in 0.4.0.

mitchellh:
And thanks for the compliments on the changelog! :)

clstokes:
What's the status of this? My use case is the same as the OP's. I don't see mention of this in the 0.4.0 release notes...

clstokes:
Split/join is the workaround for now I take it?

mitchellh:
Yes, for now.

lyrixx:
:+1: and then we will need to be able to do something like:

```
        cidr_blocks = ${var.ip1} +  ${var.ip2} + ["0.0.0.0/0"]
```

jamesob:
Is someone currently at work on this? I'm willing to jump in and start a patch if not.

jamesob:
I'm not able to get the `split`/`join` hack working for `aws_instance.security_groups`. I've got a module configured this way:

```
# foo.tf

resource "aws_instance" "server" {
  ...
  security_groups = "${split(",", var.sec_groups)}"
}


# variables.tf

variable "sec_groups" {                                                                                                                                                                                                              
   default = "sg-xx,sg-x,sg-xxx"
   description = "The security groups to instantiate the EC2 instance under."                                                                                                                                                                                                                                                                        
}
```
and using it thusly
```
# foo_user.tf

module "foo" {
[...]

PR #7110, opened 2016-06-10
Update tests for new map count indicator

catsby:
A recent change modified how things like maps were counted in the statefile. This PR updates the test files accordingly 

@jen20 there are two files touched there that I wasn't expecting, can you check them out? The unit tests seem to pass either way:

- `helper/schema/schema_test.go`
- `terraform/interpolate_test.go`


phinze:
This LGTM - we can swing back around to those two extras you mentioned:

> "#57 Computed map without config that's known to be empty does not generate diff"

^^ We should make sure that is still testing its intent.

>  `getInterpolaterFixture`

^^ I think this was just an overly embellished fixture (having attrs that aren't necessary for the tests that use it), so that's probably AOK.

Issue #7106, opened 2016-06-10
New interpolation function uniq


janschumann:
I would like to be able to remove duplicate elements from a list of strings.

Here is a use case:

```hcl
variable "administrators" {
  default = "user1,user2,user3"
}

variable "developers" {
  default = "user1,user3"
}

resource "aws_iam_user" "user" {
  count = "${length(uniq(concat(split(",", var.developers), split(",", var.administrators))))}"

  name = "${element(uniq(concat(split(",", var.developers), split(",", var.administrators)), count.index))}"
}

resource "aws_iam_group_membership" "developers" {
  name = "developers-membership"

  users = ["${split(",", var.developers)}"]

  group = "${aws_iam_group.developers.name}"
}

resource "aws_iam_group_membership" "administrators" {
  name = "administrators-membership"

  users = [ "${split(",", var.administrators)}" ]

  group = "${aws_iam_group.administrators.name}"
}
```

jen20:
This was merged as #7174. Thanks @janschumann!

PR #7111, opened 2016-06-10
Remove duplicate elements from a list via interpolation function uniq()

janschumann:
@see #7106


jen20:
Hi @janschumann! This looks like a reasonable thing to add. I'm not sure about the name `uniq` however. Personally I would tend towards `distinct` to mirror standard functional programming terminology, and SQL. Any thoughts, @phinze?

jen20:
This was merged as #7174. Thanks @janschumann!

PR #7113, opened 2016-06-10
provider/azurerm: DNS CNAME resource wasn't posting records

stack72:
Azure changed the API for CNAME at some point and since then we haven't
been creating CNAME records. The API changes from []records to a single
record

This PR changes the schema for dns cnames to have a record parameter and
adds a deprecation warning around records. Talked with @jen20 on this
and we decided that it's currently broken and we should handle this as
part of 0.7 where there are other breaking changes

```
TF_LOG=1 make testacc TEST=./builtin/providers/azurerm
TESTARGS='-run=TestAccAzureRMDnsCNameRecord' 2>~/tf.log
TF_ACC=1 go test ./builtin/providers/azurerm -v
-run=TestAccAzureRMDnsCNameRecord -timeout 120m
=== RUN   TestAccAzureRMDnsCNameRecord_basic
--- PASS: TestAccAzureRMDnsCNameRecord_basic (97.22s)
=== RUN   TestAccAzureRMDnsCNameRecord_subdomain
--- PASS: TestAccAzureRMDnsCNameRecord_subdomain (94.94s)
=== RUN   TestAccAzureRMDnsCNameRecord_updateRecords
--- PASS: TestAccAzureRMDnsCNameRecord_updateRecords (116.62s)
```

### DIFF
--- a/builtin/providers/azurerm/resource_arm_dns_cname_record.go
+++ b/builtin/providers/azurerm/resource_arm_dns_cname_record.go
@@ -34,10 +34,16 @@ func resourceArmDnsCNameRecord() *schema.Resource {
 			},
 
 			"records": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Elem:       &schema.Schema{Type: schema.TypeString},
+				Set:        schema.HashString,
+				Deprecated: "Use `record` instead. This attribute will be removed in a future version",
+			},
+
+			"record": &schema.Schema{
+				Type:     schema.TypeString,
 				Required: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
-				Set:      schema.HashString,
 			},
 
 			"ttl": &schema.Schema{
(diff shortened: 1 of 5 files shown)


builtin/providers/azurerm/resource_arm_dns_cname_record.go
```diff
@@ -34,10 +34,16 @@ func resourceArmDnsCNameRecord() *schema.Resource {
 			},
 
 			"records": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Elem:       &schema.Schema{Type: schema.TypeString},
+				Set:        schema.HashString,
+				Deprecated: "Use `record` instead. This attribute will be removed in a future version",
```
phinze: I believe we have `Removed` which is an error instead of a warning. I think that's what you want here since we don't handle "records" anymore right?

phinze:
Agreed with the "currently broken - fix w/ 0.7" strategy.

LGTM! Merge is all yours and add a note to the top of the changelog when you do. 👍

PR #7116, opened 2016-06-10
provider/aws: Clean up some map/set test checks

catsby:
Follow up to https://github.com/hashicorp/terraform/pull/7110 which was a bit overreaching. Seems some of these resources have `TypeSet`, where `#` is appropriate. 

There are other failures in https://travis-ci.org/hashicorp/terraform/builds/136731250 but they are unrelated and will be addressed outside of this PR


stack72:
LGMT! Nice to catch these early

phinze:
LGTM! Pulling in so we catch these for tonight's run.

PR #7124, opened 2016-06-11
testing: Allow acceptance test to run as unit test

jen20:
This commit adds a flag to acceptance tests in order to make appropriately named tests work during `make test` irrespective of the TF_ACC environment variable. This should only be used on tests which are known to be fast.

The actual use case here is for the remote state tests (to follow) which use the framework but have never actually been running on a regular basis (quelle horreur), which led to #6989 (and a ton of core work!)


jen20:
cc @phinze, @jbardin. I'm going to merge this pending Travis to unblock myself but please feel free to make any changes you think are necessary as a post-mortem!

PR #7127, opened 2016-06-11
core + provider/terraform: Fix outputs in remote state

jen20:
The work integrated in hashicorp/terraform#6322 "silently" broke the ability to use remote state correctly. This was widely reported as an issue in Terraform 0.7.0-rc1 both via the mailing list and the HangOps chat.

The commit messages are detailed and should be the primary reference for review: some core work on the helper/schema package was required.

cc: @phinze, @jbardin, @mitchellh

### DIFF
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/hil/ast"
 	"github.com/hashicorp/terraform/config"
 	"github.com/hashicorp/terraform/config/module"
+	"github.com/hashicorp/terraform/flatmap"
 )
 
 const (
@@ -213,6 +214,7 @@ func (i *Interpolater) valueResourceVar(
 	result map[string]ast.Variable) error {
 	// If we're computing all dynamic fields, then module vars count
 	// and we mark it as computed.
+
 	if i.Operation == walkValidate {
 		result[n] = ast.Variable{
 			Value: config.UnknownVariableValue,
@@ -343,6 +345,7 @@ func (i *Interpolater) valueUserVar(
 func (i *Interpolater) computeResourceVariable(
 	scope *InterpolationScope,
 	v *config.ResourceVariable) (*ast.Variable, error) {
+
 	id := v.ResourceId()
 	if v.Multi {
 		id = fmt.Sprintf("%s.%d", id, v.Index)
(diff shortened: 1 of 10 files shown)


terraform/interpolate.go
```diff
@@ -589,21 +590,19 @@ func (i *Interpolater) interpolateComplexTypeAttribute(
...
 		}
+		sort.Strings(keys)
 
 		var members []string
 		for _, key := range keys {
 			members = append(members, attributes[key])
 		}
-		// This behaviour still seems very broken to me... it retains BC but is
-		// probably going to cause problems in future
-		sort.Strings(members)
```
phinze: We just discussed this in chat - sorting by keys rather than members here is strictly better, and a major release is definitely the time to make this change.

In order to address the BC issues here of list interpolations unexpectedly shuffling on upgrade, we're going to introduce a `sort()` interpolation function to sort the contents of a list. This should allow users seeing unexpected diffs due to list shuffling on an 0.7 upgrade to retain the old behavior.

terraform/interpolate.go
```diff
@@ -620,19 +619,16 @@ func (i *Interpolater) interpolateComplexTypeAttribute(
...
 			}
 		}
 
-		members := make(map[string]interface{})
-		for _, key := range keys {
-			members[key] = attributes[resourceID+"."+key]
-		}
-		return hil.InterfaceToVariable(members)
+		expanded := flatmap.Expand(resourceFlatMap, resourceID)
+		return hil.InterfaceToVariable(expanded)
```
phinze: I think we should add a unit test that covers this change.

jen20: Added (and squashed into the commit that made the changes)

phinze:
LGTM pending travis. IMO the `sort()` func we can do in a separate PR.

terraform/interpolate.go
jen20: This is implemented in #7128.

chandy:
Looks like the docs were not updated to reflect the dropping of .output from the data source.

https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/terraform/index.html.markdown
https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/terraform/d/remote_state.html.md

jen20:
@chandy thanks for pointing that out! There is a lot of documentation to update for lists and maps, which I will work on over the remainder of this week and early next week.

chandy:
you saved all the most fun for last :)

ste-bah:
hi guys just wanted to know if this is in the latest release of v 0.7X ?

PR #7128, opened 2016-06-11
core: Add sort() interpolation function

jen20:
This may be required to maintian the ordering of sorted multi-variable lists in Terraform 0.7, since values are now sorted by list index rather than value. Discussed during review of #7127.

cc @phinze

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -75,6 +75,7 @@ func Funcs() map[string]ast.Function {
 		"sha1":         interpolationFuncSha1(),
 		"sha256":       interpolationFuncSha256(),
 		"signum":       interpolationFuncSignum(),
+		"sort":         interpolationFuncSort(),
 		"split":        interpolationFuncSplit(),
 		"trimspace":    interpolationFuncTrimSpace(),
 		"upper":        interpolationFuncUpper(),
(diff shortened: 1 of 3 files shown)


config/interpolate_funcs.go
```diff
@@ -529,6 +530,33 @@ func interpolationFuncSignum() ast.Function {
...
+		Callback: func(args []interface{}) (interface{}, error) {
+			inputList := args[0].([]ast.Variable)
+
+			// Ensure that all the list members are strings and
+			// create a string slice from them
+			members := make([]string, len(inputList))
+			for i, val := range inputList {
+				if val.Type != ast.TypeString {
+					return nil, fmt.Errorf(
+						"sort() may only be used with lists containg strings.")
```
phinze: s/containg/containing/

Would be nice if this message printed unexpected type a la `fmt.Errorf("list has non-string element (%T)", value.Type.String())` above.

website/source/docs/configuration/interpolation.html.md
```diff
@@ -195,6 +195,12 @@ The supported built-in functions are:
       Example: `element(split(",", var.r53_failover_policy), signum(count.index))`
       where the 0th index points to `PRIMARY` and 1st to `FAILOVER`
 
+  * `sort(list)` - Returns a lexographically sorted list of the strings contained in
+      the list passed as an argument. Sort may only be used with lists which contain only
+      strings.
+      Example: `sort(aws_instance.foo.*.id)`
+      Example: `sort(var.list_of_strings)`
```
phinze: I think the newlines won't print well on the site - might want to just merge these up inline like `Examples: `foo`, `bar``

phinze:
LGTM outside of two nits

jen20:
Cool, both of those are fixed up. Will merge pending Travis.

PR #7130, opened 2016-06-12
Add AccTest for tags with dots for Triton provider

sodre:
Test will fail on metadata_3 but not on _2 as discussed with @jen20 on 11 June 2016.

- Tests for issue hashicorp/terraform#2143.
- Fixed package names since Triton deprecated the g3-series.
- Must supply SDC_URL when invoking tests.

### DIFF
--- a/builtin/providers/triton/resource_machine_test.go
+++ b/builtin/providers/triton/resource_machine_test.go
@@ -233,6 +233,8 @@ func TestAccTritonMachine_metadata(t *testing.T) {
 	machineName := fmt.Sprintf("acctest-%d", acctest.RandInt())
 	basic := fmt.Sprintf(testAccTritonMachine_metadata_1, machineName)
 	add_metadata := fmt.Sprintf(testAccTritonMachine_metadata_1, machineName)
+	add_metadata_2 := fmt.Sprintf(testAccTritonMachine_metadata_2, machineName)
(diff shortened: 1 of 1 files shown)


jen20:
Hi @sodre - thanks for the repro here. I'll investigate this and fix it up.

jen20:
Tagging this as `core` and `release-0.7` as it's unclear whether this is related to map handling or is provider-specific at this stage.

jen20:
Hi @sodre! I've been investigating this and am _reasonably_ sure this isn't a core bug, and is down to something unusual the Triton provider is doing. The thing that makes me relatively sure that there isn't a problem with our core map key handling is this template which plans and applies successfully:

```
resource "aws_vpc" "test" {
  cidr_block = "10.0.0.0/16"

  tags {
    Name                = "test"
    Thing.Thing2.Thing3 = "test2"
  }
}
```

And this configuration which also applies successfully:

```
resource "test_resource" "test" {
  required = "hello"

  required_map = {
    key.key2.key3 = "value"
    key.key4.key5 = "value2"
  }
}
```

I'm going to keep digging here, but I suspect something in the handling of the tags map in the Triton provider is causing this.

sodre:
@jen20, since you reopened #2143 do you still think this is a Triton specific issue?

jen20:
Hi @sodre, it's not Triton specific.

jen20:
Closed in favour of #7654 - thanks for your work here @sodre!

Issue #7047, opened 2016-06-07
Can't destroy using modules on 0.7rc01


dbcoliveira:
Hi there,

I am currently using terraform 0.7rc01 and I can't destroy something that have been applied using modules (docker provider) just seconds before.

A snippet of the module source goes like:

```
variable docker_host {}
variable container_name {}
variable servers {}
variable arguments {}

provider "docker" {
    host = "tcp://${var.docker_host}:2375/"
}
.
.
.
```

A snippet of the module goes as:
```
variable "ipaddress" {
        default = "10.64.78.75,10.64.78.76,10.64.78.77"
}

module "etcd01" {
  source = "module/etcd/"
  docker_host = "${element(split(",",var.ipaddress),0)}"
  container_name = "${var.cname}"
  servers = "${var.servers}"
  arguments = ""
}
.
.
.
```

So applying with:

`terraform apply -target=module.etcd01 -target=module.etcd02 -target=module.etcd03 -var-file=secret.tfvars
`

```
module.etcd02.docker_container.etcd-viewer: Creation complete
module.etcd03.docker_container.etcd-viewer: Creation complete

Apply complete! Resources: 9 added, 0 changed, 0 destroyed.
```

Everything goes well and confirmed that it really applies.
On the destroy part it fails big:

`terraform destroy -target=module.etcd01 -target=module.etcd02 -target=module.etcd03 -var-file=secret.tfvars `
```
Do you really want to destroy?
  Terraform will delete the following infrastructure:
  	module.etcd01
  	module.etcd02
  	module.etcd03
  There is no undo. Only 'yes' will be accepted to confirm

  Enter a value: yes

.........
Error applying plan:

3 error(s) occurred:

* 1:9: unknown variable accessed: var.docker_host in:

tcp://${var.docker_host}:2375/
* 1:9: unknown variable accessed: var.docker_host in:

tcp://${var.docker_host}:2375/
* 1:9: unknown variable accessed: var.docker_host in:

tcp://${var.docker_host}:2375/
```

Note that I know the variable list declaration implementation was introduced on 0.7 so I've actually try to perform the same code without the ipaddress list but using the actual ips. the result is the same.

Any ideas ?

Thank you for the feedback

phinze:
Hi @dbcoliveira - thanks for the report, and sorry for the trouble!

Here is a full repro config that I have confirmed works in 0.6.16 and does not work on 0.7.0-rc1 or master:

```tf
# main.tf
variable "in" {
[...]

PR #7131, opened 2016-06-12
core: Fix destroy when module vars used in provider config

phinze:
For `terraform destroy`, we currently build up the same graph we do for `plan` and `apply` and we do a walk with a special Diff that says "destroy everything".

We have fought the interpolation subsystem time and again through this code path. Beginning in #2775 we gained a new feature to selectively prune out problematic graph nodes. The past chain of destroy fixes I have been involved with (#6557, #6599, #6753) have attempted to massage the "noop" definitions to properly handle the edge cases reported.

"Variable is depended on by provider config" is another edge case we add here and try to fix.

This dive only makes me more convinced that the whole `terraform destroy` code path needs to be reworked.

For now, I went with a "surgical strike" approach to the problem expressed in #7047. I found a couple of issues with the existing Noop and DestroyEdgeInclude logic, especially with regards to flattening, but I'm explicitly ignoring these for now so we can get this particular bug fixed ahead of the 0.7 release. My hope is that we can circle around with a fully specced initiative to refactor `terraform destroy`'s graph to be more state-derived than config-derived.

Until then, this fixes #7407

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -443,6 +443,45 @@ func TestContext2Apply_destroySkipsCBD(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_destroyModuleVarProviderConfig(t *testing.T) {
+	m := testModule(t, "apply-destroy-mod-var-provider-config")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+	state := &State{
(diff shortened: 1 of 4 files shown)


phinze:
cc @jen20

jen20:
LGTM - seems to work from this side. I agree that `terraform destroy` needs reworking entirely, though it's somewhat unclear how we'd end up materialising the relationships for destroy order purely from state.

PR #7132, opened 2016-06-12
core: Fix detection of empty list/map defaults

jen20:
This commit changes config parsing from weak decoding lists and maps into `[]string` and `map[string]string` respectively to decode into `[]interface{}` and `map[string]interface{}` respectively. This is in order to take advantage of the work integrated in #7082 to defeat the backward compatibility features of the mapstructure library.

Test coverage of loading empty variables and validating their default types against expectation.

cc @phinze, @jbardin


phinze:
LGTM

PR #7133, opened 2016-06-12
core: Format empty lists and maps in output

jen20:
`terraform output` and it's brethren now consolidate empty maps and lists on a single line of output instead of the pathological behaviour of taking three lines previously. The same code paths are used across all output mechanisms:

```
$ cat main.tf
variable "emptystring" {
    type = "string"
    default = ""
}

variable "emptylist" {
    type = "list"
    default = []
}

variable "emptymap" {
    type = "map"
    default = {}
}

output "emptystring" {
    value = "${var.emptystring}"
}

output "emptylist" {
    value = "${var.emptylist}"
}

output "emptymap" {
    value = "${var.emptymap}"
}

$ terraform apply

Outputs:

emptylist = []
emptymap = {}
emptystring =

$ terraform output
emptylist = []
emptymap = {}
emptystring =

$ terraform show

Outputs:

emptylist = []
emptymap = {}
emptystring =
```

### DIFF
--- a/command/output.go
+++ b/command/output.go
@@ -148,6 +148,7 @@ func formatListOutput(indent, outputName string, outputList []interface{}) strin
 	keyIndent := ""
 
 	outputBuf := new(bytes.Buffer)
+
 	if outputName != "" {
 		outputBuf.WriteString(fmt.Sprintf("%s%s = [", indent, outputName))
 		keyIndent = "  "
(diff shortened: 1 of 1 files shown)


jen20:
cc @phinze, @jbardin

phinze:
👍

PR #7135, opened 2016-06-12
core: Fix empty multi-variable type

### LINKED ISSUE #7002: core: regression in 0.7: a problem with passing interpolated parameters to modules
Hi,

I believe I found a bug in latest terraform 0.7 build which is not present in 0.6.16. This is a regression since terraform is no longer able to parse tf files which it used to parse before. It is probably related to support of lists and maps in the module parameters.

### Terraform Version
Build from git v0.7.0-rc1-53-gfcb952d

### Affected Resource(s)
Tested only with the following since it was the easiest:

* template_file

But I think it is a terraform core problem (see below).

mod/mod.tf
```hcl
variable "param" {}
```
test.tf
```hcl
resource "template_file" "tmpl" {
  count = "0"
  template = "abc"
}

module "mod" {
  source = "mod"
  param = "${join(",", template_file.tmpl.*.rendered)}"
}
```
### Expected Behavior
This is what terraform 0.6.16 returns when you run plan:
```
Refreshing Terraform state prior to plan...


doesn't need to do anything.
```

### Actual Behavior
This is what happens with terraform v0.7.0-rc1-53-gfcb952d
```

* At column 3, line 1: join: argument 1 should be TypeList, got TypeString in:

${join(",", template_file.tmpl.*.rendered)}
```

### Steps to Reproduce

1. `terraform get`
2. `terraform plan`

### Important Factoids
If you comment out module and add the following to the bottom of `test.tf`, everything works as expected. So it seems to be a problem with module parameters in particular.

```hcl
output "out" {
  value = "${join(",", template_file.tmpl.*.rendered)}"
}
```

What is more, `count` value does not seem to have any possible effect on this failure. `terraform plan` still fails in the same way.

jen20:
Previously, interpolation of multi-variables was returning an empty variable if the resource count was 0. The empty variable was defined as TypeString, Value "". This means that empty resource counts fail type checking for interpolation functions which operate on lists.

Instead, return an empty list if the count is 0. A context test tests this against further regression. Also add a regression test covering the case of a single count multi-variable.

In order to make the context testing framework deal with this change it was necessary to special case empty lists in the test diff function.

Fixes #7002.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -48,6 +48,69 @@ func TestContext2Apply(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_resourceCountOneList(t *testing.T) {
+	m := testModule(t, "apply-resource-count-one-list")
+	p := testProvider("null")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+	ctx := testContext2(t, &ContextOpts{
+		Module: m,
+		Providers: map[string]ResourceProviderFactory{
+			"null": testProviderFuncFixed(p),
+		},
+	})
+
+	if _, err := ctx.Plan(); err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	state, err := ctx.Apply()
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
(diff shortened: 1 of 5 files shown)


jen20:
cc @phinze

phinze:
LGTM

Issue #7103, opened 2016-06-10
Terraform plan panic 


ddegoede:
Testing using a complex variable (map) with a module, as a comment in #1336 suggested it should work in 0.7.0, but it resulted in panic

### Terraform Version

Terraform version: 0.7.0 rc1 301da85f30239e87b30db254a25706a6d41c2522

### Terraform Configuration Files

```
variable "cs_api_url" {}
variable "cs_api_key" {}
variable "cs_secret_key" {}

variable "server" {
    type = "map"
    default = {
        instance_count = 1
        hostname = "moduletest"
        offering = "offeringbla"
        template = "templatebla"
        network = "networkbla"
    }
}

module "server" {
    source        = "./server"
    
    server        = "${var.server}"
    cs_api_url    = "${var.cs_api_url}"
    cs_api_key    = "${var.cs_api_key}"
    cs_secret_key = "${var.cs_secret_key}"
}
```

and module 

```
variable "server" { type = "map" }
variable "cs_api_url" {}
variable "cs_api_key" {}
variable "cs_secret_key" {}
variable "cszone" {
    default = "NL2"
}

provider "cloudstack" {
    api_url    = "${var.cs_api_url}"
    api_key    = "${var.cs_api_key}"
    secret_key = "${var.cs_secret_key}"
}

resource "cloudstack_instance" "server" {
    count            = "${var.server.instance_count}"
    expunge          = true
    name             = "${lower(format("%s%02d",var.server.hostname,count.index + 1))}"
    service_offering = "${var.server.offering}"
    template         = "${var.server.template}"
    network          = "${var.server.network}"
    zone             = "${var.cszone}"
}
```

### Panic Output

[https://gist.github.com/ddegoede/68a88d19d9f6b236cf5dcf25b7b13e6b.js](https://gist.github.com/ddegoede/68a88d19d9f6b236cf5dcf25b7b13e6b.js)

### Expected Behavior

Success or an error, telling me I cannot use complex variables with modules.

### Actual Behavior

It crashed.

### Steps to Reproduce

terraform plan

phinze:
Thanks for the report @ddegoede! We'll dig into this and get back to you with what we find.

phinze:
Hi @ddegoede - the core issue here is the 0.6 -> 0.7 removal of the "dot indexing" syntax used in the reproducing config. I've made a proper error message for it on master which takes care of the panic.

On master (and in the next RC) you'll see an error message like this:

```
[...]

PR #7138, opened 2016-06-12
core: Better error for dot indexing on user vars

phinze:
Dot indexing worked in the "regexps and strings" world of 0.6.x, but it
no longer works on the 0.7 series w/ proper List / Map types.

There is plenty of dot-indexed config out in the wild, so we need to do
what we can to point users to the new syntax.

Here is one place we can do it for user variables (`var.somemap`). We'll
also need to address Resource Variables and Module Variables in a
separate PR.

This fixes the panic in #7103 - a proper error message is now returned.

### DIFF
--- a/config/interpolate.go
+++ b/config/interpolate.go
@@ -282,6 +282,10 @@ func NewUserVariable(key string) (*UserVariable, error) {
 		name = name[:idx]
 	}
 
+	if len(elem) > 0 {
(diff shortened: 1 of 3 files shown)


phinze:
cc @jen20

phinze:
Merging for now, post-hoc review welcome.

PR #7144, opened 2016-06-12
corrected "here doc" syntax notes

veselinkantsev:


phinze:
The `EOT` syntax there is also valid, I think "end of text" is what that was going for. I suppose `EOF` ("end of file") is slightly more conventional, though, so sure I'll pull this in! 😀

Issue #7030, opened 2016-06-06
Crash when using concat on lists of maps


oillio:
### Terraform Version
Terraform v0.7.0-rc1 (301da85f30239e87b30db254a25706a6d41c2522)

### Terraform Configuration Files
Module:
```hcl
variable "foo" {
  type = "list"
}
variable "bar" {
  type = "list"
}

output "outFoo" {
  value = "${concat(var.foo,var.bar)}"
}
```

Calling:
```hcl
module "envTesting" {
  source = "../modules"
  foo {
    value     = "aVal"
  }
  foo {
    value     = "anotherVal"
  }
  bar {
    value     = "thirdVal"
  }
  bar {
    value     = "finalVal"
  }
}
```

### Panic Output
https://gist.github.com/oillio/ad293c464de43f7ff10efe8b0ceac466

phinze:
Hi @oillio - thanks for this report. I've reproduced and fixed up the crash with a quick solution for now in #7145, which turns it into an error message. Getting `concat()` to support lists of complex types is its own task that I'm spinning up into a fresh issue, which I'll link back here. 👍

phinze:
Ok just double checked that the crasher is now fixed. Continued discussion of expected behavior can happen over in #7146

PR #7145, opened 2016-06-12
core: Fix panic on concat() w/ list of nonprimitives

phinze:
The `concat()` interpolation function does not yet support types other
than strings / lists of strings. Make it an error message instead of a
panic when a list of non-primitives is supplied.

Fixes the panic in #7030


phinze:
Going to merge this now. /cc @jen20 @jbardin for post hoc review

PR #7148, opened 2016-06-13
core: Fix crash with tainted resource

jen20:
This commit fixes a crash in `terraform show` where there is no primary resource, but there is a tainted resource, because of the changes made to tainted resource handling in 0.7.


gechr:
Can confirm this fixes the issue! 👍

PR #7150, opened 2016-06-13
provider/azurerm: Change `azurerm_virtual_machine` to use MaxItems

stack72:
There were lots of hacky checks in the ARM VM resource to check for a
length of 1 param block

MaxItems was introduced so this PR updates to use MaxItems


jen20:
LGTM!

Issue #7159, opened 2016-06-14
Documentation for aws_availability_zones uses interpolation on count


madpawel:
The documentation for [aws_availability_zones](https://github.com/hashicorp/terraform/blob/v0.7.0-rc2/website/source/docs/providers/aws/d/availability_zones.html.markdown) currently refers to interpolation on a count, which is not permitted. This should probably refer to a distributing IP ranges from a variable over the availability zones instead?

apparentlymart:
This was fixed as part of #7162 by updating the docs. Thanks for the report, @madpawel!

Since we already have #1497 open for the idea that more interesting interpolations should be allowed in `count`, I'm going to close this one out to consolidate any further discussion about this over there. Hopefully one day we get to roll back my doc fix in #7162 because the original example works. :grinning:

PR #7162, opened 2016-06-14
aws_availability_zones output as "names" rather than "instance"

apparentlymart:
Since this resource produces a list it feels more intuitive to give its attribute a plural name, and since the noun "instance" already means something specific in the AWS provider that doesn't apply here we use "names" to indicate that these are availability zone names.

Didn't bother with a deprecation cycle on this since it's never been in a stable release, only in RCs.

Also includes updating the docs to not show a dynamic count example for now, since we don't support that yet.


apparentlymart:
(@jen20 this is as we discussed, though while I was there I also did a doc fix for #7159 in case that doesn't get fixed some other way before 0.7 lands)

stack72:
Sitting with @jen20 and he said this LGTH :)

PR #7154, opened 2016-06-13
[WIP] provider/vsphere: reliable network interface indices

thetuxkeeper:
Before implementing the update of the network interfaces, we need a reliable way to match the config and the devices in VMware. 
This is a WIP since I have problems getting the data into the TF schema object structure. 
I want a map[string]map[string] structure. The outer map with the device id/device key as index and the inner map like the current structure with ip addresses, gateway,... 
Is there a way to override the hash function of the TypeSet ? With TypeMap  I get an error at runtime (will add the message later),  but the core message was that it was an invalid input for the schema object. 
cc @chrislovecnm

PR #7167, opened 2016-06-14
Refreshing devices list after adding a disk or cdrom controller

dkalleg:
Without refreshing the devices list, we can't find the controller we just created.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -414,7 +414,6 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							Default:  "scsi",
-							ForceNew: true,
 							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 								value := v.(string)
 								if value != "scsi" && value != "ide" {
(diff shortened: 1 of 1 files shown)


dkalleg:

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1153,9 +1153,14 @@ func addHardDisk(vm *object.VirtualMachine, size, iops int64, diskType string, d
...
 		vm.AddDevice(context.TODO(), c)
+		// Update our devices list
+		devices, err := vm.Device(context.TODO())
+		if err != nil {
+			return err
+		}
 		controller, err = devices.FindDiskController(controller_type)
 		if err != nil {
-			return fmt.Errorf("[ERROR] Could not find the controller we just created")
+			return fmt.Errorf("[ERROR] Could not find the new %v controller: %v", controller_type, err)
```
chrislovecnm: please log `err`, or check that it is logged in stack above.

chrislovecnm:
@dkalleg is this an edge case or should have we caught this in testing?

thetuxkeeper:
@dkalleg : Interesting. It's the same way I do it in #7154 for the network interfaces. Little sad that we have to spam api calls (create device function with created device as return value would be great)
@chrislovecnm : Since it's all before the `Read()` I think it's "just" a bug. the tfstate will include the created controller, so the tests won't see it.
Additionally I think (not tested) it will only fail if we add a controller type that wasn't present before (will find the old/wrong one). So we need a test that adds a non-default controller (I think a VM has a controller automatically, should be scsi) or make sure that we have no controllers at all and add one.

dkalleg:
@stack72 @phinze @thetuxkeeper @chrislovecnm 
Any actionable feedback?

chrislovecnm:
LGTM!!

dkalleg:
Thanks @stack72

PR #7174, opened 2016-06-15
core: Rebase #7111 and rename unique() to distinct()

jen20:
This pull request rebases #7111 onto master, and renames the new interpolation function `distinct()` instead of `uniq()`. It also adds documentation.


jen20:
Fixes #7111. Thanks @janschumann!

Issue #7129, opened 2016-06-11
Openstack - graceful instance termination


alkersan:
Hello @jtopjian. Is it possible to make removal of compute instances graceful? Currently Openstack provider destroys instances with [`Delete`](http://gophercloud.io/docs/compute/#delete-server) call (a.k.a. `terminate`) which don't give a chance for vm to stop it's services. E.g. consul agent, if configured with [`leave_on_terminate`](https://www.consul.io/docs/agent/options.html#leave_on_terminate) will not receive `SIGTERM` and will not deregister itself, thus leaving unhealthy node visible to cluster members.

What if prior to deletion, call [Stop](http://gophercloud.io/docs/compute/#stop-server) (from [`extensions/startstop`](https://godoc.org/github.com/rackspace/gophercloud/openstack/compute/v2/extensions/startstop) package), which should transit machine to `SHUTOFF`? This can be done at best effort basis, with lower timeout (say 1~3 mins, instead of default 10 mins for creation/deletion), doing deletion in any case afterwards.

jtopjian:
@alkersan interesting idea!

What do you think of having this as an opt-in feature? For example, a new argument called `stop_before_destroy`?

alkersan:
Yes, this enhancement is behavior changer and opt-in is needed. And as far as I understand, `start\stop` lives in `extensions` namespace because it was added later to `nova` api, thus it may simply not work for every openstack installation.

I've started working on it, will add PR soon.

jtopjian:
Closing this as it was fixed in #7184

PR #7184, opened 2016-06-15
provider/openstack stop before destroy

alkersan:
@jtopjian Please review this patch for #7129 behavior. It includes 1 added dependency to `gophercloud` package and basic opt-in implementation of stop_before_destroy flag.

Need advice on acceptance test for it. Currently it was tested manually within corporate openstack setup.

First scenario was a mini consul cluster, where clients were configured with [`leave_on_terminate`](https://www.consul.io/docs/agent/options.html) option. When I scaled down consul machines - they correctly left cluster and disappeared from node catatlog, thus it subjectively proves that Nova sent ACPI shutdown signal to guests OS. Without `stop_before_destroy` consul shows all nodes in failing state.

Another check is to peek at instance [action history API](http://developer.openstack.org/api-ref/compute/?expanded=list-actions-for-server-detail#servers-actions-servers-os-instance-actions). With  `stop_before_destroy` enabled, action history should include `stop`:
```
+--------+------------------------------------------+---------+----------------------------+
| Action | Request_ID                               | Message | Start_Time                 |
+--------+------------------------------------------+---------+----------------------------+
| create | req-e0d5c543-161a-43a0-9fb5-527b802faa03 | -       | 2016-06-15T20:01:39.000000 |
| stop   | req-b86d88e3-8916-47e3-a139-8b0cf5cfb35a | -       | 2016-06-15T20:03:47.000000 |
+--------+------------------------------------------+---------+----------------------------+
```
But, unfortunately, I didn't find implementation of history API in `gophercloud`

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/keypairs"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/schedulerhints"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/secgroups"
+	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/startstop"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/tenantnetworks"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/extensions/volumeattach"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/flavors"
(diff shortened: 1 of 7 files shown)


jtopjian:
@alkersan Nice work!

I agree -- creating an acceptance test that confirms end-to-end functionality for this is difficult. I think creating a simple test that verifies the new argument works without error is good enough. In this case, just make a copy of the "basic" acceptance test but add `stop_before_destroy = true` 😄

alkersan:
@jtopjian ok, simple acc test is there and branch was rebased on fresh master

jtopjian:
@alkersan this looks great! I just ran the acceptance test suite and everything passed.

Would you be able to squash the commits into two: one for the Gophercloud deps and the other for everything else? Once that's done, I'll go ahead and merge this in 😄

alkersan:
@jtopjian done

jtopjian:
@alkersan Thank you, sir!

Issue #7172, opened 2016-06-15
Terraform does not respect ordering of firewall rules in an Openstack firewall's policy


scrossan:
Hi Hashicorp,

I need to apply firewall rules in my Openstack firewall policy in a specific order in order to allow/deny certain things. However, Terraform does not seem to respect the order of the rules passed to the `openstack_fw_policy_v1` resource and this is causing me problems because the firewall therefore does not behave as I expect it to.

Previously we explicitly ordered the rules by number depending on their position in a list. However, the result of this was that the firewall was ripped out and replaced on each apply (because the rules would be deleted/created if their position in the list changed), causing confusion in the diffs which did not give us the confidence to apply in production. Ideally, Terraform would use the array of `rules` to guide the positioning of new rules and insert before/after as necessary.

Please let me know if you need any further information from me and I will be happy to assist.

### Terraform Version
`Terraform v0.6.16`

### Affected Resource(s)
- openstack_fw_policy_v1

### Terraform Configuration Files
Please find an example in the following gist, which I have found to demonstrate the same behaviour we find with our own configuration files - which I am unable to share with you: https://gist.github.com/scrossan/e5d5ba1d154e313cd21887879810f8e9

Interestingly, what seems to be the case is that when the destination IP addresses are ordered (e.g. for each of the three rules, 2.2.2.1, 2.2.2.2 and 2.2.2.3) the rules are ordered the correct way in the policy. However, if I change the way the rules are ordered in the `rules` array in the `openstack_fw_policy_v1` resource, leaving the ordered IPs the same, then Terraform will not detect this as a change.

In this example the destination IPs are all the same, resulting in mis-ordered rules in the policy: rule_2, rule_1 and rule_3 despite the ordering in the array of `rules`.

### Debug Output
https://gist.github.com/scrossan/00659bd731e51db60f237f63f395529b

### Expected Behavior
I expected Terraform to order the firewall rules in the policy according to the order I passed them to the policy in the array of `rules`.

### Actual Behavior
Terraform re-ordered the rules and they are ordered differently in the firewall policy than the order given in the array of `rules`. 

### Steps to Reproduce
1. `terraform apply`
2. Observe mis-ordered rules in Openstack
3. `terraform destroy`
4. `terraform apply`
5. Observe rules ordered differently than in 2, but still mis-ordered

### Important Factoids
Our cloud provider is using Openstack Kilo.

jen20:
[...]

PR #7194, opened 2016-06-16
provider/openstack: Change rules type to List

scrossan:
Allows for ordering of rules that are applied to the firewall policy,
including addition and removal of rules in certain positions.

Fixes https://github.com/hashicorp/terraform/issues/7172.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_fw_policy_v1.go
+++ b/builtin/providers/openstack/resource_openstack_fw_policy_v1.go
@@ -49,10 +49,9 @@ func resourceFWPolicyV1() *schema.Resource {
 				Computed: true,
 			},
 			"rules": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
-				Set:      schema.HashString,
 			},
 		},
 	}
@@ -66,13 +65,13 @@ func resourceFWPolicyV1Create(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
 	}
 
-	v := d.Get("rules").(*schema.Set)
+	v := d.Get("rules").([]interface{})
 
 	log.Printf("[DEBUG] Rules found : %#v", v)
-	log.Printf("[DEBUG] Rules count : %d", v.Len())
+	log.Printf("[DEBUG] Rules count : %d", len(v))
 
-	rules := make([]string, v.Len())
-	for i, v := range v.List() {
+	rules := make([]string, len(v))
+	for i, v := range v {
 		rules[i] = v.(string)
 	}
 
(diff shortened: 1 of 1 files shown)


jen20:
Hi @scrossan! This change looks good to me in principle - does it give a reasonable looking diff for those migrating from the previous format?

scrossan:
Hi @jen20 - I believe it does. If we expand the example in https://github.com/hashicorp/terraform/issues/7172 to include 11 firewall rules and apply that with the current release, Terraform plans the following when it is rebuilt with this change:
```
$ ~/Developer/go/src/github.com/hashicorp/terraform/pkg/darwin_amd64/terraform plan
...
~ openstack_fw_policy_v1.policy
    rules.0:  "" => "5c7c8680-15be-4884-958c-9ffcb80dc2e3"
    rules.1:  "" => "630826d8-522d-4751-af9a-ab6c0bad6abc"
    rules.10: "" => "62e68220-1c92-411b-9671-c41d03b588f3"
    rules.2:  "" => "e1c55715-15d0-491a-a485-ebe17efca03a"
    rules.3:  "" => "50a41c95-0d7a-42ec-81d0-face533226cb"
    rules.4:  "" => "ec9b1702-9961-47fb-b6cf-12a271c76782"
    rules.5:  "" => "be45af29-e57b-4a7b-8f07-d2471de5888b"
    rules.6:  "" => "7da6e008-4505-47f2-93fc-dd1fd569c6de"
    rules.7:  "" => "c7928376-e54b-401f-b647-cdd04de77a4f"
    rules.8:  "" => "c58efbfb-f8e4-4468-866a-458a2f2e7057"
    rules.9:  "" => "2c195512-cd82-4112-81cb-924d98eff4cb"


```
It lists the rules alphabetically rather than numerically, but now wants to order the rules as they are specified in the array of `rules` in the firewall policy resource. When I check the firewall policy in the OpenStack GUI I can see that the rules have actually been applied in the same order.

For our own configuration, with 144 rules that have seemingly so far have been applied in no particular order, Terraform wants to do the same as it does above: order the rules as they are ordered in the array of `rules`. We will need to check that applying Terraform with this change will not lead to any interruption of service by way of re-ordering the existing firewall policy, but I believe that Terraform is now behaving as we originally expected it to do so.

jtopjian:
> We will need to check that applying Terraform with this change will not lead to any interruption of service by way of re-ordering the existing firewall policy

This is probably unavoidable, but I think as long as it's detailed in the CHANGELOG, it should be sufficient.

To confirm: the old method of ordering was alphabetical by ID and now the ordering is position in the list?

> but I believe that Terraform is now behaving as we originally expected it to do so.

Indeed. I dug through the Neutron code last night and confirmed that firewall rules are applied as an array/list in order of element (Firewall rules are passed into the policy on the command-line by a space-separated string. The string is then split by whitespace and applied).

Are you able to run the fwaas acceptance tests and see if everything still passes? If so, I'm good with having this merged. If you aren't able to run them, I will be able to in the next day or so.

scrossan:
> To confirm: the old method of ordering was alphabetical by ID and now the ordering is position in the list?

In the diff ordering is still alphabetical (as above), but as for how the rules are actually applied - with this change - it's now based on position in the list, whereas before it does look like it was alphabetical by ID although I'm not certain about that.

Regarding the acceptance tests, I ran the following command which completed successfully:
```bash
make testacc TEST=./builtin/providers/openstack TESTARGS='-run=TestAccFW'
```
Just wanted to double check that I ran the correct tests before claiming that they were passing successfully!

jtopjian:
@scrossan Yup - that's the right command.

@jen20 I'm good with merging this if you feel the schema change has been done correctly.

jtopjian:
@scrossan Thank you for your work on this! I just reviewed and tested everything and I think it's good to go!

PR #7204, opened 2016-06-16
[WIP] helper/schema: Don't run StateFunc if value is computed

### LINKED ISSUE #3076: Error when using a template as the policy for an s3 bucket
When I try to use a template as the policy for my s3 bucket:

```hcl
resource "template_file" "woofound-policy" {
    filename = "${path.module}/s3-policy.tpl"

    vars {
        env = "${var.env}"
        iam = "${var.iam}"
    }
}

resource "aws_s3_bucket" "woofound" {
  bucket = "woofound-images-${var.env}"
  policy = "${template_file.woofound-policy.rendered}"
}
```
I get the following message (not an error) when I run "terraform plan --module-depth=1":

```
+ module.woofound_stag.aws_s3_bucket.woofound
    bucket:           "" => "woofound-images-stag"
    force_destroy:    "" => "0"
    policy:           "" => "Error parsing JSON: invalid character '$' looking for beginning of value"
```

However, when I run "terraform apply" everything works fine, and the policy is correctly set on S3.

phinze:
Should fix #3076 but I need to add some tests. :)

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -959,8 +959,8 @@ func (m schemaMap) diffString(
 	all bool) error {
 	var originalN interface{}
 	var os, ns string
-	o, n, _, _ := d.diffChange(k)
-	if schema.StateFunc != nil && n != nil {
+	o, n, _, computed := d.diffChange(k)
+	if schema.StateFunc != nil && n != nil && !computed {
 		originalN = n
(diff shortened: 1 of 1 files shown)


phinze:
That bug was fixed via other means, so this PR is defunct. 👋

PR #7205, opened 2016-06-16
core: don't check any parts of a computed set in InstanceDiff.Same

### LINKED ISSUE #2027: aws_security_group - diffs didn't match
Received the following error message:

aws_security_group.nat: diffs didn't match during apply. This is a bug with Terraform and should be reported.

Not sure that this is enough info to reproduce, and clearly there's some issue. 

```
resource "aws_security_group" "nat" {
	vpc_id = "${aws_vpc.main.id}"
	name = "nat"
	description = "For nat servers "

	ingress {
		from_port = 22
		to_port = 22
		protocol = "tcp"
		security_groups = ["${aws_security_group.jump.id}"]
	}

	ingress {
		from_port = 22
		to_port = 22
		protocol = "tcp"
		security_groups = ["${aws_security_group.provision.id}"]
	}

	ingress {
		from_port = 23
		to_port =  65535
		protocol = "tcp"
		security_groups = ["${aws_security_group.ssh_base.id}"]
	}

	ingress {
		from_port = -1
		to_port = -1
		protocol = "icmp"
		security_groups = ["${aws_security_group.ssh_base.id}"]
	}

	ingress {
		from_port = 23
		to_port =  65535
		protocol = "udp"
		security_groups = ["${aws_security_group.ssh_base.id}"]
	}

	egress {
		from_port = 0
		to_port = 0
		protocol = "-1"
		cidr_blocks = ["0.0.0.0/0"]
	}
}
```

jbardin:
When checking for "same" values in a computed hash, not only might some
of the values differ between versions changing the hash, but there may be
fields not included at all in the original map, and different overall
counts.

Instead of trying to match individual set fields with different hashes,
remove any hashed key longer than the computed key with the same base
name.

Fixes #2027 (and hopefully a number of other "diffs didn't match during apply" issues)

### DIFF
--- a/terraform/diff.go
+++ b/terraform/diff.go
@@ -443,35 +443,30 @@ func (d *InstanceDiff) Same(d2 *InstanceDiff) (bool, string) {
 			// values. So check if there is an approximate hash in the key
 			// and if so, try to match the key.
 			if strings.Contains(k, "~") {
-				// TODO (SvH): There should be a better way to do this...
 				parts := strings.Split(k, ".")
-				parts2 := strings.Split(k, ".")
+				parts2 := append([]string(nil), parts...)
+
 				re := regexp.MustCompile(`^~\d+$`)
 				for i, part := range parts {
 					if re.MatchString(part) {
+						// we're going to consider this the base of a
+						// computed hash, and remove all longer matching fields
+						ok = true
+
 						parts2[i] = `\d+`
+						parts2 = parts2[:i+1]
+						break
 					}
 				}
-				re, err := regexp.Compile("^" + strings.Join(parts2, `\.`) + "$")
+
+				re, err := regexp.Compile("^" + strings.Join(parts2, `\.`))
 				if err != nil {
 					return false, fmt.Sprintf("regexp failed to compile; err: %#v", err)
 				}
+
 				for k2, _ := range checkNew {
 					if re.MatchString(k2) {
 						delete(checkNew, k2)
-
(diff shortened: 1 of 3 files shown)


terraform/diff.go
```diff
@@ -443,35 +443,30 @@ func (d *InstanceDiff) Same(d2 *InstanceDiff) (bool, string) {
 			// values. So check if there is an approximate hash in the key
 			// and if so, try to match the key.
 			if strings.Contains(k, "~") {
-				// TODO (SvH): There should be a better way to do this...
 				parts := strings.Split(k, ".")
-				parts2 := strings.Split(k, ".")
+				parts2 := append([]string(nil), parts...)
```
phinze: The `append` with `[]string(nil)` base is a new one on me - looks like it makes sure parts2 is a freshly allocated slice?

terraform/diff.go
```diff
@@ -443,35 +443,30 @@ func (d *InstanceDiff) Same(d2 *InstanceDiff) (bool, string) {
 			// values. So check if there is an approximate hash in the key
 			// and if so, try to match the key.
 			if strings.Contains(k, "~") {
-				// TODO (SvH): There should be a better way to do this...
```
phinze: haha -> @jbardin responds... "there is"!

terraform/diff.go
jbardin: yup!  equivalent to  `parts2 := make([]string, len(parts)); copy(parts2, parts)`

phinze:
After chatting on phone about detail - this LGTM!

Issue #7168, opened 2016-06-14
vSphere virtual machine must be PowerOn to destroy


vipulsabhaya:
Hi there,

There seems to be an issue with deleting a VM in vsphere if it's not powered on.  There is logic in the 'get vm' code path that expects to find an IP.  When a VM is not powered on, there is no IP available.

### Terraform Version
2016/06/14 11:35:52 [INFO] Terraform version: 0.7.0 dev e7ce5a233eba8eb948d852883e9f5c94dbf43128

### Affected Resource(s)
- vsphere_virtual_machine

### Debug Output
```vsphere_virtual_machine.testdiskuuid: Refreshing state... (ID: vipul-testuuid)
[...]

PR #7206, opened 2016-06-16
vSphere Provider - Fix destroy when vm is powered off or has no networks

dkalleg:
This patch adds a wait when powering on a vm so setupVirtualMachine does
not return until the vm is actually powered on.  This allows other
functions to work off the assumption that the current state of the vm
is not in flux. During resourceVSphereVirtualMachineRead(), the wait for
IP would cause a hang for any VM with no network interfaces or for vms
that had been powered off for any reason.  This also means that the user
could not delete a vm with no network interfaces or that is powered off.
Checking power state before trying to check for network interfaces.
Resolves https://github.com/hashicorp/terraform/issues/7168

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -619,12 +619,6 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 		}
 	}
 
-	ip, err := vm.WaitForIP(context.TODO())
-	if err != nil {
-		return err
-	}
-	log.Printf("[DEBUG] ip address: %v", ip)
-
 	return resourceVSphereVirtualMachineRead(d, meta)
 }
 
@@ -897,14 +891,20 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 		return nil
 	}
 
-	var mvm mo.VirtualMachine
-
-	// wait for interfaces to appear
-	_, err = vm.WaitForNetIP(context.TODO(), true)
+	state, err := vm.PowerState(context.TODO())
 	if err != nil {
 		return err
 	}
 
+	if state == types.VirtualMachinePowerStatePoweredOn {
+		// wait for interfaces to appear
+		_, err = vm.WaitForNetIP(context.TODO(), true)
+		if err != nil {
+			return err
+		}
+	}
+
+	var mvm mo.VirtualMachine
 	collector := property.DefaultCollector(client.Client)
 	if err := collector.RetrieveOne(context.TODO(), vm.Reference(), []string{"guest", "summary", "datastore", "config"}, &mvm); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


dkalleg:
This does add an interesting caveat where if you try to run `terraform plan` when a vm is powered off, it will find a diff between the "discovered networks" and the "declared networks"
```    
    name:                                   "tfDestroyFix" => "tfDestroyFix"
    network_interface.#:                    "0" => "1" (forces new resource)
    network_interface.0.ip_address:         "" => "<computed>"
    network_interface.0.ipv4_address:       "" => "<computed>"
    network_interface.0.ipv4_gateway:       "" => "<computed>"
    network_interface.0.ipv4_prefix_length: "" => "<computed>"
    network_interface.0.ipv6_address:       "" => "<computed>"
    network_interface.0.ipv6_gateway:       "" => "<computed>"
    network_interface.0.ipv6_prefix_length: "" => "<computed>"
    network_interface.0.label:              "" => "VM Public" (forces new resource)
    network_interface.0.mac_address:        "" => "<computed>"
    network_interface.0.subnet_mask:        "" => "<computed>"
    skip_customization:                     "false" => "false"
```

This is sort of a grey area since handling a powered off vm is not really a handled case at the moment.  This prompts a discussion about the ability to handle/recover from different power states, but that should be in a separate issue/PR :)  .  This PR resolves a known issue of not being able to destroy a vm that is powered off or has no network interfaces.

dkalleg:

chrislovecnm:
When should we be waiting for all ips, one ip, or customization to be done?

This has been a constant thorn in our side, and this PR looks like a good point to address it ;)

chrislovecnm:
When should we be waiting for all ips, one ip, or customization to be done?

This has been a constant thorn in our side, and this PR looks like a good point to address it ;)

thetuxkeeper:
the `WaitForNetIP` in the update function (line 622) can be removed. I don't see a reason why it's still there since it's also in the read function, which will be called just a few lines later.
Everything else looks good. Nice improvement/bug fix :)

dkalleg:
@thetuxkeeper I did a test with that WaitForNetIP() removed and the Update() called the Read() before the vm was up, which resulted in the Read() not finding the networks and messing up the tfstate.  I think it's important to have at the end of the Update() so that the Read() happens once the vm is in a stable state.
@chrislovecnm Hi Paranoid, I'm Dan. :)  Adding that check.  Considering that we're populating `networkInterfaces` in a way that may or may not have the ipv4_address key, its a good idea.

chrislovecnm:
@dkalleg we are dev ops guys, not be paranoid gets us woken up at 2am 😄

thetuxkeeper:
@dkalleg : that's strange. that was the reason why I put the `property.DefaultCollector` after the `WaitForNetIP` call in `Read()` (was the other way round before the `WaitForNetIP` PR and had this exact problem). That needs some more testing or at least logs since it's not intended that we need to call it twice. It makes it just harder for future changes (bigger refactoring or stuff like that).
But I don't think it's a blocker for this PR. We can think about it in another issue/PR. Will try to reproduce it, when I have time.

chrislovecnm:
@thetuxkeeper you have a chance to test this?

dkalleg:
@thetuxkeeper @stack72 @phinze @jen20 @chrislovecnm 
Any further actionable feedback?

dkalleg:
@thetuxkeeper @stack72 @phinze @jen20 @chrislovecnm 
Just checking in.  Any thoughts on this?

thetuxkeeper:
Nothing from me. I think we can merge it. If I find the time to run the tests, I'll do it. But would be safer if someone else did it (I'm still struggling with this new cloud stuff on detail level ... :D )

dkalleg:
@stack72 Could you take a look at this?

chrislovecnm:
@dkalleg do we need a test for this?  Otherwise, what is the status of you running the integration tests?  Overall looks great, and I appreciate the effort.

dkalleg:
@chrislovecnm Sorry, missed this.  Tests are passing on my end.  This might be a little overly specific of a use case to make a test for.

Issue #7155, opened 2016-06-13
Terraform against local Triton lab setup


neuroserve:
### Terraform Version
Terraform v0.6.16

### Affected Resource(s)
triton_machine

### Terraform Configuration Files
provider "triton" {
  account      = ""
  key_id       = ""
  key_material = ""
  url          = ""
}

resource "triton_machine" "dcos-bootstrap" {
  name    = "dcos-bootstrap"
  image   = "d8e65ea2-1f3e-11e5-8557-6b43e0a88b38"
  package = "sample-2G-kvm"

  tags {
    role = "bootstrap"
  }

  connection {
    host        = "${self.primaryip}"
    private_key = "${file("/root/.ssh/sdc-docker-bla.id_rsa")}"
  }

  provisioner "remote-exec" {
    inline = ["/bin/true"]
  }
}

### Panic Output
https://gist.github.com/neuroserve/6d900714fab98a1e496236c8bacbf3bb

### Expected Behavior
I would have expected to see the provisioning of a triton kvm-instance

### Actual Behavior
Terraform crashed.

### Important Factoids
I'm running terraform against my local Triton Lab (not against Joyents public cloud). I'm not sure, whether that could be a factor.

jen20:
Hi @neuroserve! Can you confirm what the URL you are using for your local setup is? The crash to me suggests a bug in Joyent's gosign library which is incorrectly detecting local setups:

https://github.com/joyent/gosign/blob/master/auth/auth.go#L117-L122

cc @BrianHicks, @misterbisson

neuroserve:
It is a local URL. It does not resolve in DNS. Its configured in /etc/hosts in the VM that's running terraform. It's like https://cloudapi.my-dc-1.cns.mylab.de/

sodre:
@neuroserve This is almost surely the issue in joyent/gosign#8. Can you remove the "trailing slash" in your SDC_URL and try again?

@jen20, I will put a PR to update gosign to the latest version which has the fix.

neuroserve:
Removing the trailing slash works! Provisioning is now possible. Thx.

PR #7208, opened 2016-06-17
vendor: joyent/gosign/auth update to master

sodre:
Updates Joyent's gosign to the latest version. 
On the vendor side it addressed issue joyent/gosign#8 which in turn should close hashicorp/terraform#7155.


stack72:
Hi @sodre 

Thanks so much for this PR - this LGTM!


Paul

PR #7211, opened 2016-06-17
provider/mysql: Empty Provider Credentials Caused Panic

### LINKED ISSUE #7199: Empty mysql endpoint results in unexpected EOF
terraform 0.6.16

```
provider "mysql" {
  endpoint = ""
  username = "user"
  password = "password"
}

resource "mysql_database" "db" {
  name = "somedb"
}
```

The above config will result in the following error
```

* unexpected EOF
```

Instead I'd expect it to error out that the endpoint is not set or that it was unable to connect to the provider.

stack72:
Fixes #7199 There are currently no checks on username and endpoint in the provider
schema from being an empty value. This PR adds support to make sure that
endpoint and username are not empty strings as that can cause a panic

Results of the PR:

```
 % terraform apply
fix these before continuing.

Errors:

  * provider.mysql: Endpoint must not be an empty string
```

### DIFF
--- a/builtin/providers/mysql/provider.go
+++ b/builtin/providers/mysql/provider.go
@@ -17,12 +17,28 @@ func Provider() terraform.ResourceProvider {
 				Type:        schema.TypeString,
 				Required:    true,
 				DefaultFunc: schema.EnvDefaultFunc("MYSQL_ENDPOINT", nil),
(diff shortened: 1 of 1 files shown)


jen20:
LGTM! Thanks @stack72!

Issue #5907, opened 2016-03-29
AWS Opsworks Rails App Layer ignores manage_bundler


ldhertert:
Any rails app later created by terraform has "Install and manage Bundler" set to "no", regardless of the value placed in the terraform config.  Using terraform v0.6.14.

apparentlymart:
Thanks for reporting this, @ldhertert.

I did some initial investigation here but didn't have time yet to dig deep. The rest of this comment is just some notes for future-me or anyone else that wants to try to debug this:

The opsworks layer attributes are implemented in a generic, schema-driven way so I think there are two obvious ways in which this could be broken:

* The attribute name Terraform is using in the underlying API call is incorrect. This would mean it would affect only this attribute on this layer type.
* The way Terraform is serializing boolean values is incorrect. This would mean that all of the boolean attributes could be broken across all layer types, which seems less likely since I remember seeing other booleans work during my earlier testing.

[The documentation](http://docs.aws.amazon.com/opsworks/latest/APIReference/API_CreateLayer.html) shows that there is indeed an attribute called `ManageBundler`, which is the name that Terraform is using.

As far as I can tell there is no real documentation for what values are expected for each attribute, so the values used by Terraform were figured out by trying things in the UI and then checking how they appeared in the API responses. However, it's possible that there are some inconsistencies in these that I didn't catch in the original implementation work.

When I get a chance I'll try to repro this and see how what Terraform is writing is different to what the AWS console writes when populating this attribute.

janschumann:
I think I found a similar bug while testing #4419 (opsworks app). There some optional arguments can be passed to the aws api depending on the application type. These are passed as `Arguments` of type `map[string]*string`. Probably these two bugs share the same source of pain?

u2mejc:
I poked at this a little today. First issue is here in opsworks_layers.go

```go
var (
   opsworksTrueString  = "1"
   opsworksFalseString = "0"
)
```
Should be:
```go
var (
   opsworksTrueString  = "true"
   opsworksFalseString = "false"
)
```

That small change actually resolves @ldhertert initial issue. If you `terraform apply`, it works. But after digging, the root issue is the live (remote AWS) layer attributes are not being written to the statefile. This breaks changing anything inside the attributes map and then the eval loop fails to see any changes. I poked around some, but it didn't bubble up to me yet. I'll poke at it more as time allows, but I don't have a patch staged at this time.

iroller:
[...]

PR #7219, opened 2016-06-17
Correct true and false OpsWorks string values to fix manage_bundler config bug

dlanner:
This PR implements the small change provided by @u2mejc in [his comment](https://github.com/hashicorp/terraform/issues/5907#issuecomment-218038835) on issue #5907, in order to fix a bug where the `aws_opsworks_rails_app_layer` resource's `manage_bundler` parameter is ignored, so no matter if you set it to `true`, `false`, or don't even set it, the created Rails layer's "Install and manage Bundler" setting is unconditionally "No".

There are no existing acceptance tests for `resource_aws_opsworks_rails_app_layer`. Unfortunately, I don't have the time to write them, although it seems like it would be a good opportunity to learn a little Go.

But I did compile Terraform in Vagrant (`make dev`) with this change and tested my Terraform template file which set `manage_bundler = true`, and it resulted in the created Rails layer having "Install and manage Bundler" set to "yes". So it does fix that issue, as @u2mejc said.

### DIFF
--- a/builtin/providers/aws/opsworks_layers.go
+++ b/builtin/providers/aws/opsworks_layers.go
@@ -39,8 +39,8 @@ type opsworksLayerType struct {
 }
 
 var (
-	opsworksTrueString  = "1"
-	opsworksFalseString = "0"
+	opsworksTrueString  = "true"
+	opsworksFalseString = "false"
 )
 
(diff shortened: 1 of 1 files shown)


stack72:
Hi @dlanner 

Thanks so much for this PR - this LGTM!

Paul

PR #7220, opened 2016-06-17
vSphere provider - Graceful read miss

dkalleg:
For both the file and virtual_disk resource, Stat is used during read,
but if Stat returns an error, read() will return that error. In doing
so, if a resource is deleted manually, the TF user would then not be
able to destroy the resource because the read would block the Delete()
call. With this patch, read() will only return an error if that error
is NOT a DatastoreNoSuchFileError.


dkalleg:

jen20:
LGTM! Thanks @dkalleg.

chrislovecnm:
LGTM ... @jen20 @stack72 can we get a merge?

Issue #6624, opened 2016-05-12
aws_elastic_beanstalk_application: panic when applying 


thomasvandoren:
### Terraform Version
Terraform v0.6.15

### Affected Resource(s)
- terraform-provider-aws

### Terraform Configuration Files
```hcl
# main.tf
module "nexus" {
  source = "../modules/nexus"

  name ="nexus3"
  ssl_cert_id = "arn:aws:iam::REDACTED:server-certificate/REDACTED.com-wildcard-cert"
  cname = "CNAME HERE"
  instance_type = "m3.medium"
  keypair_name = "nexus3_keypair"
  keypair_public_key = "PUBLIC_KEY HERE"
}

# ../modules/nexus/main.tf
variable "name" {
  type = "string"
  default = ""
}

variable "ssl_cert_id" {
  type = "string"
  default = ""
}

variable "cname" {
  type = "string"
  default = ""
}

variable "instance_type" {
  type = "string"
  default = "t2.micro"
}

variable "keypair_name" {
  type = "string"
  default = ""
}

variable "keypair_public_key" {
  type = "string"
  default = ""
}

resource "aws_key_pair" "nexus3_kp" {
    key_name = "${var.keypair_name}"
    public_key = "${var.keypair_public_key}"
}

resource "aws_elastic_beanstalk_application" "nexus3_app" {
  name = "${var.name}-app"
}

resource "aws_elastic_beanstalk_configuration_template" "nexus3_config" {
  name = "tf-nexus3-template-config"
  application = "${aws_elastic_beanstalk_application.nexus3_app.name}"
  solution_stack_name = "64bit Amazon Linux 2016.03 v2.1.0 running Docker 1.9.1"

  // FIXME: use a vpc here!!!
  /*setting {
    namespace = "AWSEBLoadBalancerSecurityGroup.aws:ec2:vpc"
    name = "VPCId"
    value = "${}"
  }*/

  setting {
    namespace = "aws:autoscaling:asg"
    name = "MaxSize"
    value = "1"
  }

  setting {
    namespace = "aws:autoscaling:asg"
    name = "MinSize"
    value = "1"
  }

  setting {
    namespace = "aws:autoscaling:launchconfiguration"
    name = "EC2KeyName"
    value = "${var.keypair_name}"
  }

  setting {
    namespace = "aws:autoscaling:launchconfiguration"
    name = "InstanceType"
    value = "${var.instance_type}"
  }

  setting {
[...]

PR #7222, opened 2016-06-18
provider/aws: elastic beanstalk invalid setting crash

dharrisio:
This fixes the crash in #6624/#6707/#7197. It also adds functionality to relay any failures from the elastic beanstalk environment events. As an example, the config

```
provider "aws" {
  region = "us-east-1"
}

resource "aws_elastic_beanstalk_application" "tftest" {
  name = "tf-test-name"
  description = "tf-test-desc"
}

resource "aws_elastic_beanstalk_environment" "tftest" {
  name = "tf-test-name"
  application = "${aws_elastic_beanstalk_application.tftest.name}"
  solution_stack_name = "64bit Amazon Linux 2016.03 v2.1.0 running Go 1.4"

setting {
    namespace = "aws:elb:loadbalancer"
    name = "SecurityGroups"
    value = "{\"Fn::GetAtt\":[\"AWSEBLoadBalancerSecurityGroup\",\"GroupId\"]},{\"Ref\":\"AWSEBLoadBalancerSecurityGroup\"}"
  }
}
```
Results in the following output
```
aws_elastic_beanstalk_application.tftest: Creating...
  description: "" => "tf-test-desc"
  name:        "" => "tf-test-name"
aws_elastic_beanstalk_application.tftest: Creation complete
aws_elastic_beanstalk_environment.tftest: Creating...
  all_settings.#:               "" => "<computed>"
  application:                  "" => "tf-test-name"
  autoscaling_groups.#:         "" => "<computed>"
  cname:                        "" => "<computed>"
  cname_prefix:                 "" => "<computed>"
  instances.#:                  "" => "<computed>"
  launch_configurations.#:      "" => "<computed>"
  load_balancers.#:             "" => "<computed>"
  name:                         "" => "tf-test-name"
  queues.#:                     "" => "<computed>"
  setting.#:                    "" => "1"
  setting.2468988437.name:      "" => "SecurityGroups"
  setting.2468988437.namespace: "" => "aws:elb:loadbalancer"
  setting.2468988437.value:     "" => "{\"Fn::GetAtt\":[\"AWSEBLoadBalancerSecurityGroup\",\"GroupId\"]},{\"Ref\":\"AWSEBLoadBalancerSecurityGroup\"}"
  solution_stack_name:          "" => "64bit Amazon Linux 2016.03 v2.1.0 running Go 1.4"
  tier:                         "" => "WebServer"
  triggers.#:                   "" => "<computed>"
  wait_for_ready_timeout:       "" => "10m"
aws_elastic_beanstalk_environment.tftest: Still creating... (10s elapsed)
aws_elastic_beanstalk_environment.tftest: Still creating... (20s elapsed)
aws_elastic_beanstalk_environment.tftest: Still creating... (30s elapsed)
Error applying plan:

1 error(s) occurred:

* aws_elastic_beanstalk_environment.tftest: 
2016-06-18 00:45:09.529 +0000 UTC: Stack named 'awseb-e-5s3a3gisjc-stack' aborted operation. Current state: 'CREATE_FAILED'  Reason: The following resource(s) failed to create: [AWSEBLoadBalancer]. 
2016-06-18 00:44:57.692 +0000 UTC: Creating load balancer failed Reason: Security group(s) can be applied to only an ELB in VPC.

```
Most of the tests on this pass, with one exception. The worker tier test fails, I was only able to reproduce the issue in us-west-2, so it might be temporary.

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
@@ -370,7 +370,7 @@ func resourceAwsElasticBeanstalkEnvironmentRead(d *schema.ResourceData, meta int
 		return err
 	}
 
-	if tier == "WebServer" {
+	if tier == "WebServer" && env.CNAME != nil {
 		beanstalkCnamePrefixRegexp := regexp.MustCompile(`(^[^.]+).\w{2}-\w{4,9}-\d.elasticbeanstalk.com$`)
 		var cnamePrefix string
 		cnamePrefixMatch := beanstalkCnamePrefixRegexp.FindStringSubmatch(*env.CNAME)
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
@@ -225,6 +225,8 @@ func resourceAwsElasticBeanstalkEnvironmentCreate(d *schema.ResourceData, meta i
 		createOpts.TemplateName = aws.String(templateName)
 	}
 
+	// Get the current time to filter describeBeanstalkEvents messages
+	t := time.Now()
 	log.Printf("[DEBUG] Elastic Beanstalk Environment create opts: %s", createOpts)
 	resp, err := conn.CreateEnvironment(&createOpts)
 	if err != nil {
@@ -248,6 +250,11 @@ func resourceAwsElasticBeanstalkEnvironmentCreate(d *schema.ResourceData, meta i
 		return fmt.Errorf(
 			"Error waiting for Elastic Beanstalk Environment (%s) to become ready: %s",
 			d.Id(), err)
+	}
+
+	err = describeBeanstalkEvents(conn, d.Id(), t)
+	if err != nil {
+		return err
 	}
 
 	return resourceAwsElasticBeanstalkEnvironmentRead(d, meta)
@@ -293,6 +300,8 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
 		updateOpts.TemplateName = aws.String(d.Get("template_name").(string))
 	}
 
+	// Get the current time to filter describeBeanstalkEvents messages
+	t := time.Now()
 	log.Printf("[DEBUG] Elastic Beanstalk Environment update opts: %s", updateOpts)
 	_, err = conn.UpdateEnvironment(&updateOpts)
 	if err != nil {
@@ -313,6 +322,11 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
 		return fmt.Errorf(
 			"Error waiting for Elastic Beanstalk Environment (%s) to become ready: %s",
 			d.Id(), err)
+	}
+
+	err = describeBeanstalkEvents(conn, d.Id(), t)
+	if err != nil {
+		return err
 	}
 
 	return resourceAwsElasticBeanstalkEnvironmentRead(d, meta)
@@ -514,6 +528,8 @@ func resourceAwsElasticBeanstalkEnvironmentDelete(d *schema.ResourceData, meta i
 		TerminateResources: aws.Bool(true),
 	}
 
+	// Get the current time to filter describeBeanstalkEvents messages
+	t := time.Now()
 	log.Printf("[DEBUG] Elastic Beanstalk Environment terminate opts: %s", opts)
 	_, err = conn.TerminateEnvironment(&opts)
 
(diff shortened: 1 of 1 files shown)


catsby:
Thanks! I appreciate the work you did digging in and fixing these!

builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -642,3 +663,26 @@ func dropGeneratedSecurityGroup(settingValue string, meta interface{}) string {
 
 	return strings.Join(legitGroups, ",")
 }
+
+func describeBeanstalkEvents(conn *elasticbeanstalk.ElasticBeanstalk, environmentId string, t time.Time) error {
```
jbardin: A better name might be `describeBeanstalkErrors` since this only checks for ERRORs.

I'd also prefer a `[]string` of the errors, (or even a `[]*elasticbeanstalk.EventDescription` rather than concatenating them into an error value. A function returning an error generally indicated the call failed.

jbardin:
oops, just missed the merge ;) My comments were only stylistic, so nothing that can't be cleaned up later if need be.

builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
dharrisio: @jbardin Thanks for the feedback here, sorry it took a bit of time for me to get around to this.

1.  I agree with changing the name to describeBeanstalkErrors. Originally, I was contemplating having a parameter to set the level of events returned, but I think error is all we really need. 
2. I tried two approaches here, but ran into issues with both. Can you elaborate on how to do this? I tried both returning a slice of errors and a slice of strings. I pushed the slice of errors example to branch on my fork: https://github.com/dharrisio/terraform/commit/950de0f78f1bcc801b00fc64034b530c0c21b5db
What I ran into were type errors about returning `[]error` as type `error`. There was a similar issue with using a slice of strings: `[]string` as type `string` in argument to `fmt.Errorf`.

jbardin:
Hi @dharrisio, sorry for the delayed reply.

This is mostly a style comment, and really only has an impact if someone else tries to reuse this function elsewhere in the code, so don't worry about it too much.

re: #2, This makes more sense once you change the name to `describeBeanstalkErrors`. The only error that indicates the call failed is the error from `DescribeEvents`, which is only logged. Since the purpose of the function is to "describe errors", those errors would be the expected return value, and any problem encountered while describing those would be returned as a separate error (if at all).

As for the return type, (IMHO) I think it shouldn't return an `error` type if the call succeeds (though `error` may be OK if the function were called say `checkBeanstalkEvents`, this is all about style and self-documenting APIs after all).

Though if you do ever need to bundle multiple errors like this, all you need is to define an Error() method on a container for those errors. Internally we use these for added functionality:

 - github.com/hashicorp/go-multierror
 - github.com/hashicorp/errwrap

dharrisio:
@jbardin Thanks for the explanation on the errors, that helps a lot!

My reasoning for having this return an error is that the Beanstalk API doesn't return an error in some cases. When this happens we don't know if the resource was not created or updated until we get the errors from describe events. When this happens we don't want Terraform to continue creating/updating other resources in the graph. From my understanding, in that case, we would want this to be returned as an error?

As an example, expanded on from #6624

```
provider "aws" {
  region = "us-east-1"
}

resource "aws_elastic_beanstalk_application" "tftest" {
  name = "tf-test-name"
  description = "tf-test-desc"
}

resource "aws_elastic_beanstalk_environment" "tftest" {
  name = "tf-test-name"
  application = "${aws_elastic_beanstalk_application.tftest.name}"
  solution_stack_name = "64bit Amazon Linux 2016.03 v2.1.0 running Go 1.4"

  setting {
    namespace = "aws:elb:loadbalancer"
    name = "SecurityGroups"
    value = "{\"Fn::GetAtt\":[\"AWSEBLoadBalancerSecurityGroup\",\"GroupId\"]},{\"Ref\":\"AWSEBLoadBalancerSecurityGroup\"}"
  }
}

resource "aws_route53_record" "default" {
  zone_id = "${var.zone_id}"
  name    = "${aws_elastic_beanstalk_environment.tftest.name}"
  type    = "${var.record_type}"
  ttl     = "${var.record_ttl}"
  records = ["${aws_elastic_beanstalk_environment.tftest.cname}"]
}
```

In this case, the environment creation will fail, and we would not want to create/update the r53 record to point to the bad environment.

To further complicate and confuse things, we probably do not want errors when deleting the resource because it is possible to get some false positives. In this case I had to run `terraform destroy` twice, because of errors trying to delete an autoscaling group that doesn't exist.

Example output using multierror: https://gist.github.com/dharrisio/8024591dfaabb0bfdd73ee39d9ff17e8

PR #7223, opened 2016-06-18
website/remote: added example usage for -disable

leecalcote:
Added example usage for remote config subcommand, showing how to move from remote state to local state.


jen20:
LGTM, thanks @leecalcote!

PR #7225, opened 2016-06-18
provisioner/local-exec add powershell support

tommynsong:
add support to start a new PS session directly instead of via cmd /C which prove to be very problematic.


teamterraform:
Hi folks! Now that terraform 0.12 is released, we are undertaking a general GitHub issue/PR cleanup. I am going to close this very old PR as most of the relevant code paths have changed a(and the provider is in its own repository now). Thanks!

PR #7227, opened 2016-06-18
Stop swallowing errors from DescribeAlarms

dtolnay:
This caused alarms to be created repeatedly if Terraform is running without DescribeAlarms permission.


jen20:
Thanks @dtolnay! This looks good to me.

PR #7231, opened 2016-06-19
Add scaleway provider

nicolai86:
this PR adds a new provider for scaleway. It is an extended, tested & properly integrated version of the scaleway provider found here: https://github.com/hex-sh/terraform-provider-scaleway. 

the provider build here allows scaleway users to manage servers, volumes, IPs, security groups & security group rules. The only thing missing at the moment is runtime volume attachment, which is not exposed via the scaleway API, and requires multiple API calls to succeed in sequence (stopping the running server, attaching the volume, starting the server again; reverse for detaching) which is why I did not implement it for now; I'd prefer a top level resource for this, similar to how aws handles it;

However, before diving into an implementation for this I'd prefer some feedback.

### DIFF
--- a/builtin/bins/provider-scaleway/main.go
+++ b/builtin/bins/provider-scaleway/main.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"github.com/hashicorp/terraform/builtin/providers/scaleway"
+	"github.com/hashicorp/terraform/plugin"
+)
+
+func main() {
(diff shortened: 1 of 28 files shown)


nicolai86:
I just saw that the saleway api sdk has some rather undesirable dependencies (logrus, http2curl, anonuuid) - I'll try to get this fixed at the source and will create a new PR as soon as the dependencies are removed (ideally)

PR #7237, opened 2016-06-20
[WIP] Support "hiding" ignored parameters from state.

### LINKED ISSUE #516: Storing sensitive values in state files
#309 was the first change in Terraform that I could find that moved to store sensitive values in state files, in this case the `password` value for Amazon RDS. This was a bit of a surprise for me, as previously I've been sharing our state files publicly. I can't do that now, and feel pretty nervous about the idea of storing state files in version control at all (and definitely can't put them on github or anything).

If Terraform is going to store secrets, then some sort of field-level encryption should be built in as well. In the meantime, I'm going to change things around to use https://github.com/AGWA/git-crypt on sensitive files in my repos.

modax:
Add a lifecycle flag "hide_ignored" to remove "ignored" resource
parameters from the final state. While definitely not perfect, this
kind of simple solution should make it possible to prevent sensitive
from being written to tfstate.

"hide_ignored" uses "ignore_changes" list since it does not make sense
to omit non-ignored prameters from state file as it would break terraform
flow.

Fixes #516

The work is not complete yet as it has no tests and documentation. What is more, probably parameter name could be better as well, I'm open to suggestions.

I would just like to hear from development team if you would accept this kind of solution and I should pursue to improve it (won't take long) or just drop the idea. While definitely not perfect, this patch is kind of simple, the flag is opt-in and I think it would be useful for many users out there (me included) as passwords in the state file pose a huge security problem. For example, I can encrypt terraform.tfvars myself to store it securely but I basically have no control of how terraform.tfstate gets stored remotely.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -110,6 +110,7 @@ type ResourceLifecycle struct {
 	CreateBeforeDestroy bool     `mapstructure:"create_before_destroy"`
 	PreventDestroy      bool     `mapstructure:"prevent_destroy"`
 	IgnoreChanges       []string `mapstructure:"ignore_changes"`
+	HideIgnored         bool     `mapstructure:"hide_ignored"`
 }
 
 // Copy returns a copy of this ResourceLifecycle
--- a/config/loader_hcl.go
+++ b/config/loader_hcl.go
@@ -675,7 +675,7 @@ func loadManagedResourcesHcl(list *ast.ObjectList) ([]*Resource, error) {
 		var lifecycle ResourceLifecycle
 		if o := listVal.Filter("lifecycle"); len(o.Items) > 0 {
 			// Check for invalid keys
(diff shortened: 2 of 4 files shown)


stack72:
Hi @modax 

What is the current status of this PR? IS this something you want to proceed with?

Paul

modax:
Hi @stack72,

I still think this approach is not a bad idea (because currently there are no other options) and I'm willing to work on finishing it (if you approve it makes sense). However there is one limitation I noticed that when the "hide_ignored" flag gets removed, the parameter is treated as new and hence might trigger recreation of some resources.

stack72:
Hi

We see your WIP hasn't had any activity for over 28 days, so we're going to close it. We're still interested in taking a look so if it is ready for review/merge, please reopen it. 

Thank you!

Paul

PR #7242, opened 2016-06-20
provider/cloudstack: add affinity_group_names

mcanevet:


builtin/providers/cloudstack/resource_cloudstack_instance.go
```diff
@@ -81,6 +81,15 @@ func resourceCloudStackInstance() *schema.Resource {
 				Set:      schema.HashString,
 			},
 
+			"affinity_group_names": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
```
svanharmelen: Are you sure changes to the attribute require a new instance to take effect?

svanharmelen:
@mcanevet thanks for the PR! I have two small question and you need to rebase the PR, other then that it looks good.

builtin/providers/cloudstack/resource_cloudstack_instance.go
svanharmelen: Just checked and this is not needed: http://cloudstack.apache.org/docs/api/apidocs-4.8/root_admin/updateVMAffinityGroup.html So we can lose the `ForceNew`

builtin/providers/cloudstack/resource_cloudstack_instance.go
```diff
@@ -81,6 +81,15 @@ func resourceCloudStackInstance() *schema.Resource {
...
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
```
svanharmelen: It does not only make sense, but is also required as the params are mutually exclusive according to the docs: http://cloudstack.apache.org/docs/api/apidocs-4.8/root_admin/updateVMAffinityGroup.html

mcanevet:
@svanharmelen fixed

svanharmelen:
@mcanevet thanks... Noticed the build fails, which is because of the `Computed` flag. Since we do not read back this data in the read func, the `Computed` flag doesn't make much sense here. So it's save to just remove that flag and then we should be good to go...

mcanevet:
@svanharmelen Travis is happy now!

svanharmelen:
Thanks!

Issue #7178, opened 2016-06-15
aws_db_security_group tries to delete the resource while trying to add a new ingress


vivienschilis:
Hi there,

### Terraform Version
Terraform v0.6.16

### Affected Resource(s)
- aws_db_security_group

### Terraform Configuration Files
```hcl
resource "aws_db_security_group" "remote-auto" {
  description = "foo"
  name = "remote-auto"

  ingress {
     cidr = "80.15.144.202/32"
  }
  
  ingress {
     cidr = "120.223.36.167/32"
  }
}
```

### Debug Output
https://gist.github.com/vivienschilis/3b088e7857ef8ca25722efac172bb646

### Panic Output
None

### Expected Behavior
It should have appended a new ingress on the existing security group.

### Actual Behavior
It tries to delete the current security group in order to add new ingress. This will create a downtime to the RDS access. 

Also if the db is already associated with a security group, it will fail deleting it because my db is not yet managed by terraform, that's expected obviously

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Add one ingress
2. `terraform apply`
3. Add a second ingress
4. `terraform plan`

### Important Factoids
Running in EC2 Classic

Thanks guys

catsby:
Thanks @vivienschilis – I opened #7245 to address this

vivienschilis:
Thanks @catsby :). It's interesting to the solution!

catsby:
Added in #7245

PR #7245, opened 2016-06-20
provider/aws: Add inplace edit/update DB Security Group Rule Ingress

catsby:
Fixes https://github.com/hashicorp/terraform/issues/7178


stack72:
This LGTM but it would be good to add an acceptance test to prove that it works :)

PR #7249, opened 2016-06-20
core: don't compare map diffs for computed values

jbardin:
Adding a failing test before we decide how to fix this.

issue #7244


jbardin:
This will remove all computed map values from the `InstanceDiff.Same` check. That function (and our diff process) needs some close review in general, but I wanted to add this with the minimal change set possible. 

This will prevent spurious "diffs didn't match during apply" errors caused by computed map values.

phinze:
Nice catch here. LGTM

PR #7258, opened 2016-06-21
docs: Make middleman work offline out of the box

radeksimko:
Details & reasoning behind solution discussed in https://github.com/hashicorp/middleman-hashicorp/pull/28

### Test plan

1. `bundle exec middleman`
2. Go to `http://localhost:4567/downloads.html` to verify there's a single default build

This behaviour is now default no matter if you're online or offline (in `development` environment).

The assumption this is built on is that most contributors/maintainers update content rather than tune downloads page. If anyone needs to verify changes in the downloads page they'll need to do this:

```
set :releases_enabled, true
```

### DIFF
--- a/website/config.rb
+++ b/website/config.rb
@@ -1,7 +1,12 @@
 set :base_url, "https://www.terraform.io/"
 
+configure :development do
+  set :releases_enabled, false
+end
+
 activate :hashicorp do |h|
-  h.name        = "terraform"
-  h.version     = "0.6.16"
-  h.github_slug = "hashicorp/terraform"
+  h.name             = "terraform"
+  h.version          = "0.6.16"
+  h.github_slug      = "hashicorp/terraform"
+  h.releases_enabled = config[:releases_enabled]
 end


jen20:
cc @sethvargo

website/config.rb
```diff
@@ -1,7 +1,12 @@
...
+end
+
 activate :hashicorp do |h|
-  h.name        = "terraform"
-  h.version     = "0.6.16"
-  h.github_slug = "hashicorp/terraform"
+  h.name             = "terraform"
+  h.version          = "0.6.16"
+  h.github_slug      = "hashicorp/terraform"
+  h.releases_enabled = config[:releases_enabled]
```
sethvargo: Do you need to make these changes? I think just setting the config value should be fine?

jen20:
Personally I have no clue about how any of the Middleman stuff works, so I'll leave this discussion as being between @sethvargo and @radeksimko here (or anyone else with sufficient knowledge to determine what the correct course of action is here)!

jen20:
Hi @radeksimko and @sethvargo - not sure if this was ever resolved but I'll close this as it won't merge cleanly anyway. If it's still necessary we can fix it up in future.

PR #6717, opened 2016-05-17
provider/template: convert resources to data sources

apparentlymart:
The template resources don't actually need to retain any state, so they are good candidates to be data sources.

PR #7265, opened 2016-06-22
Force the re-creation of the template resource if its content is modifed

### LINKED ISSUE #6899: template_cloudinit_config needs two applys to propagate a change
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.
### Terraform Version

0.6.16
### Affected Resource(s)
- template_cloudinit_config

``` hcl
resource "template_file" "test" {
  template = "# foo ${bar} baz"
  vars {
    bar = "howdy"
  }
}

resource "template_cloudinit_config" "test" {
  gzip = false
  base64_encode = false

  part {
    content_type = "text/cloud-config"
    content = "${template_file.test.rendered}"
    merge_type = "list(append)+dict(recurse_array)+str()"
  }
}

  ami = "ami-5c207736"
  instance_type = "m3.medium"

  user_data = "${template_cloudinit_config.test.rendered}"
}
```
### Expected Behavior

The second `terraform apply` should change all three resources.
### Actual Behavior

Running `terraform apply` the first time does the right thing.  Then change the value of bar variable in template file.  When doing the second `terraform apply`,  only changes template_file and template_cloudinit_config, but change doesn't propagate to the aws_instance that is using the template_cloudinit_config.  A third `terraform apply` picks up that change.
### Steps to Reproduce
1. `terraform apply` All three resources get created
2. Change the value of bar variable in template_file 
3. `terraform apply` only changes template_file and template_cloudinit_config, but change doesn't propagate to the aws_instance that is using the template_cloudinit_config
4. `terraform apply` The aws_instance finally gets modified

ColinHebert:
Fixes #6899

At the moment the provider makes the assumption that the template can be "modified" (while still being the same resource), this is not going to work as expected because the resource doesn't notify the system to mention that the resource has changed.

In the case of templates, there is no "update" it gets thrown away and recreated.

### DIFF
--- a/builtin/providers/template/resource_cloudinit_config.go
+++ b/builtin/providers/template/resource_cloudinit_config.go
@@ -27,6 +27,7 @@ func resourceCloudinitConfig() *schema.Resource {
 			"part": &schema.Schema{
 				Type:     schema.TypeList,
 				Required: true,
+				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"content_type": &schema.Schema{
--- a/builtin/providers/template/resource_cloudinit_config.go
+++ b/builtin/providers/template/resource_cloudinit_config.go
@@ -19,7 +19,6 @@ func resourceCloudinitConfig() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceCloudinitConfigCreate,
 		Delete: resourceCloudinitConfigDelete,
-		Update: resourceCloudinitConfigCreate,
 		Exists: resourceCloudinitConfigExists,
 		Read:   resourceCloudinitConfigRead,
 


jen20:
Hi @ColinHebert! Thanks for opening a pull request. This doesn't look quite correct however - the internal validation of the provider is failing with this changeset:

```
--- FAIL: TestProvider (0.00s)
	provider_test.go:11: err: resource template_cloudinit_config: All fields are ForceNew or Computed w/out Optional, Update is superfluous
FAIL
FAIL	github.com/hashicorp/terraform/builtin/providers/template	0.006s
```

I'll try to fix this up at some point, unless you get a chance to take another look at it!

ColinHebert:
Heh, if I read it correctly, it's complaining about the fact that there is an update path when the fields are by design forcing new resources (therefore the update path is irrelevant).

That's awesome because that's exactly what I want \o/

modax:
Won't this problem go away when template_file is converted to data source so it no longer has state?

apparentlymart:
@modax yes I think you are right here... the `template_file` data source should get refreshed automatically before planning begins and so dependent resources will automatically see the newly-rendered result, without needing to `apply` first.

ColinHebert:
Oh neato!

jen20:
Thanks for the ping on #6717, @apparentlymart! I'll get that reviewed soon.

PR #7270, opened 2016-06-22
provider/dns: Implementing dns provider.

BSick7:
## Purpose

This introduces a provider to read dns records.
This PR is inspired by https://github.com/Shopify/terraform-provider-dns.
Code was adapted from that repository to work as `data` sources instead of `resources`.
## Checklist
- [x] Implement `dns` provider
- [x] Implement `data "dns_a_record"`
- [x] Test `data "dns_a_record"`
- [x] Implement `data "dns_cname_record"`
- [x] Test `data "dns_cname_record"`
- [x] Implement `data "dns_txt_record"`
- [x] ~~Test `data "dns_txt_record"`~~
- [x] Add `dns` provider docs to site

### DIFF
--- a/builtin/providers/dns/data_dns_a_record.go
+++ b/builtin/providers/dns/data_dns_a_record.go
@@ -26,15 +26,6 @@ func dataSourceDnsARecord() *schema.Resource {
 				Default:  true,
 			},
 
-			// Optionally filter IPv6 records from DNS replies.
-			// This is helpful for other resources that do no support IPv6 yet,
-			// such as AWS security groups
-			"ipv4": &schema.Schema{
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  false,
-			},
-
 			"addrs": &schema.Schema{
 				Type:     schema.TypeList,
 				Elem:     &schema.Schema{Type: schema.TypeString},
@@ -53,16 +44,13 @@ func dataSourceDnsARecordRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	addrs := make([]string, 0)
-	ipv4only := d.Get("ipv4").(bool)
 	sortingEnabled := d.Get("sort").(bool)
 
 	for _, ip := range records {
-		if ipv4only {
-			if ipv4 := ip.To4(); ipv4 != nil {
-				addrs = append(addrs, ipv4.String())
-			}
-		} else {
-			addrs = append(addrs, ip.String())
+		// LookupIP returns A (IPv4) and AAAA (IPv6) records
+		// Filter out AAAA records
+		if ipv4 := ip.To4(); ipv4 != nil {
+			addrs = append(addrs, ipv4.String())
 		}
 	}
 
--- a/builtin/providers/dns/data_dns_a_record.go
+++ b/builtin/providers/dns/data_dns_a_record.go
@@ -1,0 +1,77 @@
+package dns
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"net"
+	"sort"
+)
+
+func dataSourceDnsARecord() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDnsARecordRead,
+		Schema: map[string]*schema.Schema{
+			"host": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			// Optionally sort A records in alphabetical order.
+			// This is helpful when a name uses round-robin DNS, which may
+			// sort records with multiple addresses in a non-deterministic order.
+			// This random sorting can cause flapping in terraform plans, where
+			// the changes in sort order cause dependent resources to update
+			// despite having no real change in the set of addresses.
+			"sort": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			// Optionally filter IPv6 records from DNS replies.
+			// This is helpful for other resources that do no support IPv6 yet,
+			// such as AWS security groups
+			"ipv4": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			"addrs": &schema.Schema{
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceDnsARecordRead(d *schema.ResourceData, meta interface{}) error {
+	host := d.Get("host").(string)
+
+	records, err := net.LookupIP(host)
+	if err != nil {
+		return err
+	}
+
+	addrs := make([]string, 0)
+	ipv4only := d.Get("ipv4").(bool)
+	sortingEnabled := d.Get("sort").(bool)
+
+	for _, ip := range records {
+		if ipv4only {
+			if ipv4 := ip.To4(); ipv4 != nil {
+				addrs = append(addrs, ipv4.String())
+			}
+		} else {
+			addrs = append(addrs, ip.String())
+		}
+	}
+
+	if sortingEnabled {
+		sort.Strings(addrs)
+	}
+
+	d.Set("addrs", addrs)
+	d.SetId(host)
+
+	return nil
+}
--- a/builtin/providers/dns/data_dns_a_record.go
+++ b/builtin/providers/dns/data_dns_a_record.go
@@ -14,18 +14,6 @@ func dataSourceDnsARecord() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			// Optionally sort A records in alphabetical order.
-			// This is helpful when a name uses round-robin DNS, which may
-			// sort records with multiple addresses in a non-deterministic order.
-			// This random sorting can cause flapping in terraform plans, where
-			// the changes in sort order cause dependent resources to update
-			// despite having no real change in the set of addresses.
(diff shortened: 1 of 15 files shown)


BSick7:
cc @thegedge @burke @dalehamel

BSick7:
Every time I execute my tests, I get
```
testing.go:255: Step 0 error: Check failed: Check 1/2 error: Not found: data.dns_a_record.google
```

@apparentlymart any chance you can weigh in as to why this may be happening (since you built the much loved `data` source)

BSick7:
Nevermind @apparentlymart. I figured it out.
Make sure it's very obvious to consumers that `SetId(...)` must be called in a data source read func.

BSick7:
I don't have a solid way of building acceptance tests for TXT records.
I'm ready for review @stack72

builtin/providers/dns/data_dns_a_record.go
```diff
@@ -0,0 +1,77 @@
...
+			"sort": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			// Optionally filter IPv6 records from DNS replies.
+			// This is helpful for other resources that do no support IPv6 yet,
+			// such as AWS security groups
+			"ipv4": &schema.Schema{
```
thegedge: Technically speaking, `A` records are only IPV4, and `AAAA` records are IPV6. `net.LookupIP` unfortunately looks like it does both, so should we do the filtering always? Some food for thought.

BSick7: Great feedback. I will remove and always filter.
Perhaps I could add `data "dns_aaaa_record"` next. (which looks likes someone is studdering :smile: )

BSick7: Done.

builtin/providers/dns/data_dns_a_record.go
```diff
@@ -0,0 +1,77 @@
...
+			// This is helpful when a name uses round-robin DNS, which may
+			// sort records with multiple addresses in a non-deterministic order.
+			// This random sorting can cause flapping in terraform plans, where
+			// the changes in sort order cause dependent resources to update
+			// despite having no real change in the set of addresses.
+			"sort": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
```
thegedge: :+1: Good call on determinism here. Any thoughts on making this happen by default? I think a lot of other resources generally do something like that.

thegedge:
Just a quick skim for a review, with a couple of small comments. Thanks for taking the time to make this a data source! We haven't played with data sources yet, but everything does look great. We'll let someone from the team give the extensive review 😄

builtin/providers/dns/data_dns_a_record.go
BSick7: My inner self was telling me to let the developer have the ability to choose.
As you challenge it, I can't honestly think of a case where someone would want to see terraform plans change just because the records changed order.

I am going to remove sort, always sort the records.
If somebody wants different, they can always add it in later.

BSick7: Done. IP addresses are always sorted.

nanoz:
Being able to read SRV records would be so helpful for load balancer configs in a DNS service discovery setup ! :)

radeksimko:
Hi @BSick7 
thanks for the PR.
I just quickly skimmed over the code and a few high-level questions came up to my mind:

I'm thinking the motivation behind this can be to speed up network connections by performing DNS lookup early instead of leaving it up to runtime which is understandable/reasonable. Are there any other motivations behind this provider? More specific use cases would be very helpful.

The significant (probably obvious?) downside - especially in environments like AWS (ignoring EIPs for now) - is that DNS records may change and point to different IPs/names.
Users who don't realise this downside/effect may not realise that translated DNS names/IPs don't get updated until they run `terraform apply`.

I assume this provider just uses default DNS servers set in the system. Was this an intention or do you have any plans on allowing the user to specify DNS servers?

-----

@nanoz Would you mind providing example(s) of load balancer(s) that need to be fed with translated DNS SRV records and can't do the translation/lookup themselves?

thegedge:
@radeksimko I'll fill you in on our use case, and the reason we originally wrote the read-only resource plugin referenced by the OP.

We have an AMI builder on the side, and whenever it built a new AMI it would update a TXT record with the id of that AMI. We didn't need to constantly update our existing instances with new AMIs because chef was making sure everything was up-to-date, so whenever new instances were brought up it would use the latest AMI we built (by creating a new `dns_txt_record` resource).

steve-jansen:
@radeksimko I'm happy to answer your questions on the value of this data provider

@BSick7 and I use [Shopify's custom DNS provider](https://github.com/Shopify/terraform-provider-dns) today to whitelist network egress to third party systems by FQDN.  What's interesting is how stable DNS records can be - we've found in practice that our whitelists resolve to pretty stable IPs.  Our own 3rd parties seem to use virtual IPs on load balancers that are very stable.

I agree that new users could misinterpret how a DNS data resource works in the context of the Terraform plan/apply lifecycle.   Perhaps we just call this out in the doc page for this resource?

For our own use case, Consul checks generate alerts when a whitelisted FQDN is no longer reachable.  This allows us to plan/apply when needed.  We also have the option for Atlas to automatically plan/apply on a schedule.

While it's not a perfect solution, we've found it to be a pragmatic approach that has been better than simply allowing egress on 443 to 0.0.0.0.  We're working on a more robust approach outside Terraform; this is our stopgap until that new approach is ready.  I think others could benefit from this as well.

radeksimko:
@thegedge I see, so basically you are using TXT records to achieve something you can now achieve via https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/d/ami.html.markdown ?

> @BSick7 and I use Shopify's custom DNS provider today to whitelist network egress to third party systems by FQDN.

That's interesting use case, thanks for sharing it! So your 3rd parties expose something like this https://ip-ranges.amazonaws.com/ip-ranges.json in a form of DNS records?

thegedge:
> @thegedge I see, so basically you are using TXT records to achieve something you can now achieve via https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/d/ami.html.markdown ?

Yep, that looks like it could probably work, as long as it can be locked down for existing instances (which we prefer to avoid recreating if there's a newer AMI).

BSick7:
> @thegedge I see, so basically you are using TXT records to achieve something you can now achieve via https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/d/ami.html.markdown ?

That is an awesome data source @radeksimko!

> That's interesting use case, thanks for sharing it! So your 3rd parties expose something like this https://ip-ranges.amazonaws.com/ip-ranges.json (specific to the 3rd party, not AWS, obviously) in a form of DNS records?

Yup! A very topical example is that we have a security group to allow scada egress from our nodes to reach atlas to join the cluster (consul, nomad).

jason-riddle:
Would querying SRV records fall into this as well? Or should I just open a pull request to add that in after this has been merged?

steve-jansen:
@radeksimko do you think this PR might land in Terraform 0.7?

apparentlymart:
I have some use-cases that a DNS provider could help with, though I'd need some additional functionality:

* Make the `TXT` record data source try to parse the value with [the syntax for name/value pairs in DNS-SD](https://tools.ietf.org/html/draft-cheshire-dnsext-dns-sd-11#section-6.3) and expose them as an additional map attribute `dns_sd` if successful. If not successful, I'd just leave the map empty so it can still be used for TXT records that aren't for DNS-SD.
* Support `SRV` records.

As I'm sure you can guess from the above, the use-case is an application of the DNS service discovery protocol that is part of Bonjour, although we're using these conventions in static DNS rather than with multicast/dynamic DNS as is usually the case with Bonjour. We use this as a sort of early-stage service discovery mechanism for some things we need to find in order to bootstrap our Consul clusters. (We're currently using this information only in some provisioning scripts that get run by Terraform, but it'd be pretty cool to be able to use this data within Terraform itself too.)

Certainly not meaning to suggest that this additional scope is required for landing this PR... but something I'd probably add to it in a subsequent PR were it merged. (I'll leave the final call on that to @radeksimko since he's already been reviewing here.)

radeksimko:
FYI: I'm aware this PR was submitted a while back, I'd just appreciate your input in https://github.com/hashicorp/terraform/pull/8299 which may conflict with this PR.

Thanks.

drebes:
Hi, just got aware of this through @radeksimko comment in #8299. The idea of that DNS provider is to update DNS records using TSIG, but I also considered using data providers to query data from the  servers and split the currently single dns_record resource into multiple dns_<type>_record.

I could see we making the credentials optional and bounding the zone to the records, that way I believe we would be able to combine the two providers. But have you considering basing your provider in https://github.com/miekg/dns It's a great library to handle DNS in golang.

steve-jansen:
Hi @radeksimko this PR has been open for many months.  

How can we land this and/or #8299?

catsby:
Hey all – the plan as it stands is to merge #8299 and then combine this PR into the Provider included in it. 

Basically, 8299 has many resources for creating/managing DNS and this PR has data sources, so we'd like both under a single `dns` provider. 

I'll be doing most of the heavy lifting here, but please let me know if you have any concerns or objections.

Thanks!

BSick7:
Closing.
Moving to new PR: https://github.com/hashicorp/terraform/pull/12744

PR #7272, opened 2016-06-22
core: Correctly ensure that State() is a copy

jen20:
The previous mechanism for testing state threw away the mutation made on the state by calling State() twice - this commit corrects the test to match
the comment.

In addition, we replace the custom copying logic with the copystructure library to simplify the code.

cc @phinze, @jbardin

### DIFF
--- a/state/testing.go
+++ b/state/testing.go
@@ -118,9 +118,11 @@ func TestState(t *testing.T, s interface{}) {
 			t.Fatalf("bad: expected %d, got %d", serial, reader.State().Serial)
 		}
 
-		// Check that State() returns a copy
-		reader.State().Serial++
-		if reflect.DeepEqual(reader.State(), current) {
+		// Check that State() returns a copy by modifying the copy and comparing
+		// to the current state.
+		stateCopy := reader.State()
+		stateCopy.Serial++
+		if reflect.DeepEqual(stateCopy, current) {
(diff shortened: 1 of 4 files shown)


state/testing.go
```diff
@@ -118,9 +118,11 @@ func TestState(t *testing.T, s interface{}) {
 			t.Fatalf("bad: expected %d, got %d", serial, reader.State().Serial)
 		}
 
-		// Check that State() returns a copy
-		reader.State().Serial++
-		if reflect.DeepEqual(reader.State(), current) {
+		// Check that State() returns a copy by modifying the copy and comparing
+		// to the current state.
+		stateCopy := reader.State()
+		stateCopy.Serial = 1000000
```
jbardin: should we ++ here like before? if only to get rid of the magic number (and very slim chance of collision)

jen20: We definitely could `++` here - I originally changed this to put some kind of easy-to-spot sentinal value for observability during debugging, but can change it back.

jbardin: Oh, a sentinel is fine too is this is visible somewhere. I' like the MaxInt64 value myself (1<<63-1) ;) , but this is OK

jbardin:
LGTM. Using the same copystructure for all the deep copies should be more consistent.

PR #7273, opened 2016-06-22
core: Don't try to copy files over themselves

### LINKED ISSUE #7238: terraform init blanks files
### Terraform Version

    Terraform v0.7.0-dev (fe540b408f590ee2e410537672160598e300d056)

### Affected Resource(s)

- local files

### Expected Behavior

From `terraform init` documentation:

> Init will download the module from SOURCE and copy it into the DIR (which defaults to the current working directory). Version control information from the module (such as Git history) will not be copied.

> The directory being initialized must be empty of all Terraform configurations. If the module has other files which conflict with what is already in the directory, they will be overwritten.

### Actual Behavior

If you check the second paragraph that I mentioned earlier it shouldn't blank any file that doesn't conflict. Or at least this is what I understand from reading it.

### Steps to Reproduce

```
$ mkdir tf

$ echo "alex was here" > tf/1

$ ls -l tf
total 8
-rw-r--r--  1 alex  wheel  14 20 Jun 15:57 1

$ terraform init tf

$ ls -l tf
total 8
-rw-r--r--  1 alex  wheel  14 20 Jun 15:57 1

$ cd tf

$ terraform init .

$ ls -l
total 0
-rw-r--r--  1 alex  wheel  0 20 Jun 15:57 1
```

If I run the `init` from outside the folder it doesn't blank the files.

jbardin:
When copying a config module, make sure the full path for src and dst
files don't match, and also check the inode in case we resolved a
different path to the same file.

Make a note about the unsafe usage of reusing a tempDir path.

fixes #7238

### DIFF
--- a/config/module/copy_dir.go
+++ b/config/module/copy_dir.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 )
 
 // copyDir copies the src directory contents into dst. Both directories
@@ -19,6 +20,7 @@ func copyDir(dst, src string) error {
 		if err != nil {
 			return err
 		}
+
 		if path == src {
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


jen20:
This LGTM in principle (pending the Travis failure being fixed up)! Great catch!

jbardin:
hmm, where did I miss the tests locally ...

jbardin:
Forgot to check for IsNotExist. Tests should pass now

PR #7276, opened 2016-06-22
core: output expected HCL types when evaluating config

### LINKED ISSUE #7139: Error message on module input type mismatch should not reference Go types
### Terraform Version

master / 2127466280172a7cd0aa1d863cb932aa0107c986 

### Affected Resource(s)

core

```tf
# main.tf
variable "amap" {
  type    = "map"
  default = {  }
}
module "mod" {
  source   = "./mod"
  mapparam = "${var.amap}"
}
```

```tf
# mod/main.tf
variable "mapparam" { }
```

### Expected Behavior

An error message that does not reference Go types.

### Actual Behavior

```

* variable mapparam in module mod should be type string, got map[string]interface {}
```

jbardin:
Don't refer to Go types when an unexpected type is encountered in the
config.

Fixes #7139

### DIFF
--- a/terraform/eval_variable.go
+++ b/terraform/eval_variable.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/hashicorp/terraform/config"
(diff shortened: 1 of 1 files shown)


jbardin:
Is this the correct language for the types we're expecting? I used HCL as a reference, but I'm not sure if this is supposed to be more terraform specific.

terraform/eval_variable.go
```diff
@@ -161,3 +162,26 @@ func (n *EvalVariableBlock) Eval(ctx EvalContext) (interface{}, error) {
...
+func hclTypeName(i interface{}) string {
+	switch k := reflect.Indirect(reflect.ValueOf(i)).Kind(); k {
+	case reflect.Bool:
+		return "boolean"
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32,
+		reflect.Uint64, reflect.Uintptr, reflect.Float32, reflect.Float64:
+		return "number"
+	case reflect.Array, reflect.Slice:
+		return "array"
```
jen20: We have been calling these "lists" in Terraform so far - do we want to keep to this or use the Go names here?

terraform/eval_variable.go
```diff
@@ -161,3 +162,26 @@ func (n *EvalVariableBlock) Eval(ctx EvalContext) (interface{}, error) {
...
+	case reflect.Map:
+		return "object"
```
jen20: Similarly here we have been talking about maps rather than objects

jen20:
Hi @jbardin! I left a couple of comments on where we have differed from HCL terminology - personally I think it's better to remain consistent with the `type` field on variable declarations?

jbardin:
Yes, I wanted to be consistent, I just wasn't sure at exactly whether these were supposed to be targeted more to "HCL" or "Terraform".

I'll change up the terminology.

jen20:
@jbardin I fixed this up locally, rebased and merged - the merge commit is at `ad27dc9`.

PR #7277, opened 2016-06-22
Fix readme numbered indentation

beedub:
The markdown wasn't being rendered as expected because there was a lack of indentation


stack72:
Hi @beedub 

Thanks for the PR here - this LGTM!

Paul

PR #7281, opened 2016-06-23
provider/aws: Support import functionality for `aws_ecr_repository`

stack72:
Changes required:

* ECR Read func doesn't need to take ``registry_id` as it uses the
current account Id
* `name` wasn't being set in the ECR Read so the import was failing as
name wasn't found
* the tests were being run against us-east-1 as this used to be the only supported region - that restriction has been removed

```
make testacc TEST=./builtin/providers/aws
TESTARGS='-run=TestAccAWSEcrRepository_'
-run=TestAccAWSEcrRepository_ -timeout 120m
=== RUN   TestAccAWSEcrRepository_importBasic
--- PASS: TestAccAWSEcrRepository_importBasic (17.37s)
=== RUN   TestAccAWSEcrRepository_basic
--- PASS: TestAccAWSEcrRepository_basic (16.05s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    33.437s
```

### DIFF
--- a/builtin/providers/aws/import_aws_ecr_repository_test.go
+++ b/builtin/providers/aws/import_aws_ecr_repository_test.go
@@ -1,0 +1,28 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSEcrRepository_importBasic(t *testing.T) {
+	resourceName := "aws_ecr_repository.default"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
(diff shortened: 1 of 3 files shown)


jen20:
@stack72 - Given the change to region, should we also update the note on the [documentation page](https://www.terraform.io/docs/providers/aws/r/ecr_repository.html)?

stack72:
@jen20 luckily we future proofed the documentation :)

```
NOTE on ECR Availability: The EC2 Container Registry is not yet rolled out in all regions - available regions are listed
[the AWS Docs](https://docs.aws.amazon.com/general/latest/gr/rande.html#ecr_region).
```

jen20:
:+1: SGTM.

PR #7290, opened 2016-06-23
provider/aws: Support Import for `aws_iam_user`

stack72:
The Id wasn't being set until after the Read func returned from the API.
I needed to move that Id set up until just after the Create response
returned

The same Id's have been set - username

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSUser_'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSUser_
-timeout 120m
=== RUN   TestAccAWSUser_importBasic
--- PASS: TestAccAWSUser_importBasic (14.24s)
=== RUN   TestAccAWSUser_basic
--- PASS: TestAccAWSUser_basic (24.99s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    39.261s
```

### DIFF
--- a/builtin/providers/aws/import_aws_iam_user_test.go
+++ b/builtin/providers/aws/import_aws_iam_user_test.go
@@ -1,0 +1,28 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSUser_importBasic(t *testing.T) {
+	resourceName := "aws_iam_user.user"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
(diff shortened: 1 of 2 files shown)


jen20:
:+1: LGTM! Thanks @stack72!

russmac:
I'm having an issue using a user list with count, In that the import makes the users resource name aws_iam_user.$name and the terraform configuration names the resources aws_iam_user.$count

Is this expected behavior? To need to write a single user resource for every user?

jen20:
@russmac Can you post your configuration? This sounds like you're missing an interpolation context.

jen20:
Oh, sorry I misread - yes - if you import your resources will be named in the form `aws_iam_user.name`, vs the `aws_iam_user.name.index` that a list would require. You could probably do manual state surgery here, or `terraform state` set of operations might allow you to do what you need.

PR #7293, opened 2016-06-23
HTTP Basic Auth

mrcrilly:
When using an HTTP remote, just like a Consul remote, I should be able to provide HTTP Basic Auth credentials so that I can "lock down" my HTTP end point and not leave it open to the world or rely on secret URL elements, such as UUIDs in the URL, to prevent unauthorized access.

This change is practically identical to hashicorp@1a19f43

### DIFF
--- a/state/remote/http.go
+++ b/state/remote/http.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mrcrilly:
Closing this off due to confusion on my side.

PR #7298, opened 2016-06-23
code: Add proper build constraints for GH-7273

jbardin:
The syscall.Stat_t type doesn't exist on windows, so the inode lookup
needs to be in a file with proper build constraints.


jen20:
LGTM - I don't know if there is a way of determining the same thing on Windows.

PR #7304, opened 2016-06-23
core: Fix bug with interpolation of unknown multi-variables

jen20:
This PR includes a number of commits necessary to fix interpolation during the Input walk when using the `element()` function with a multi-variable generated by the splat syntax. We now correctly proceed on to the planning stage during the Input stage rather than panicing, as non-reified multi-variables are treated as `UnknownVariable` rather than empty lists.


stack72:
Nice work on testing this edge case - this LGTM!

jen20:
cc @phinze or @mitchellh for some post-mortem review on this.

mitchellh:
I won't be able to take a look until Monday (on vacation!), but noting that I opened this tab so I do look when I get back.

mitchellh:
LGTM! :)

PR #7315, opened 2016-06-24
provider/cloudstack: make ACL's swappable, unless you want to stop using an ACL

svanharmelen:
In CloudStack you can dynamically start using an ACL and once you use
an ACL you can dynamically swap ACL’s. But once your using an ACL, you
can no longer stop using an ACL without rebuilding the network.

This change makes the `ForceNew` value dynamic so that it only returns
`true` if you are reverting from using an ACL to not using an ACL
anymore, making this functionally inline with the behaviour CloudStack
offers.


svanharmelen:
@phinze if you have a second to 👀  this one... Thx!

PR #7325, opened 2016-06-25
Add a link to the cloudwatch dimensions aws page in the docs

pshima:
Thought it would be handy to add the link directly here.


radeksimko:
👍

PR #7326, opened 2016-06-25
provider/aws: Documentation update for Elasticache Snapshot

### LINKED ISSUE #4947: AWS elastic cache specifying snapshot_retention_limit doesnt allow to create t2.micro instance
I have a Elastic cache template that have snapshot_retention_limit  in it.
But as appears it cant be specified for t2.micro instances.
```
* aws_elasticache_cluster.redis: Error creating Elasticache: InvalidParameterCombination: Instance type does not support snapshotting. Snapshot retention limit parameter should not be specified.
	status code: 400, request id: a0e8c652-c934-11e5-92e5-9f4a1b5228e7
```
I tried to set 0, still no luck. So basically its means that i can provision micro instance with the same template as other size instances since terraform always passing snapshot_retention_limit

stack72:
Fixes #4947: To avoid confusion on using snapshots and t2 instance
sizes, a note has been added to the documentation


catsby:
👍

PR #7352, opened 2016-06-27
provider/aws: Support Import `aws_dynamodb_table`

stack72:
There were some changes required to the Read func to get this working.
The initial set of tests showed the following:

```
testing.go:255: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) {
        }

        (map[string]string) (len=8) {
         (string) (len=8) "hash_key": (string) (len=16) "TestTableHashKey",
         (string) (len=23) "local_secondary_index.#": (string) (len=1) "1",
         (string) (len=36) "local_secondary_index.884610231.name": (string) (len=12) "TestTableLSI",
         (string) (len=52) "local_secondary_index.884610231.non_key_attributes.#": (string) (len=1) "0",
         (string) (len=47) "local_secondary_index.884610231.projection_type": (string) (len=3) "ALL",
         (string) (len=41) "local_secondary_index.884610231.range_key": (string) (len=15) "TestLSIRangeKey",
         (string) (len=4) "name": (string) (len=38) "TerraformTestTable-2710929679033484576",
         (string) (len=9) "range_key": (string) (len=17) "TestTableRangeKey"
        }
```

On investigation, this was telling me that `hash_key`, `range_key`, `name` and `local_secondary_index` were not being set on the Read func

When they were being set, all looks as expected:

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSDynamoDbTable_'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSDynamoDbTable_ -timeout 120m
=== RUN   TestAccAWSDynamoDbTable_importBasic
--- PASS: TestAccAWSDynamoDbTable_importBasic (20.39s)
=== RUN   TestAccAWSDynamoDbTable_basic
--- PASS: TestAccAWSDynamoDbTable_basic (39.99s)
=== RUN   TestAccAWSDynamoDbTable_streamSpecification
--- PASS: TestAccAWSDynamoDbTable_streamSpecification (50.44s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	110.841s
```

### DIFF
--- a/builtin/providers/aws/import_aws_dynamodb_table_test.go
+++ b/builtin/providers/aws/import_aws_dynamodb_table_test.go
@@ -1,0 +1,28 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSDynamoDbTable_importBasic(t *testing.T) {
+	resourceName := "aws_dynamodb_table.basic-dynamodb-table"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSDynamoDbTableDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSDynamoDbConfigInitialState(),
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
--- a/builtin/providers/aws/resource_aws_dynamodb_table.go
+++ b/builtin/providers/aws/resource_aws_dynamodb_table.go
@@ -34,6 +34,9 @@ func resourceAwsDynamoDbTable() *schema.Resource {
 		Read:   resourceAwsDynamoDbTableRead,
 		Update: resourceAwsDynamoDbTableUpdate,
 		Delete: resourceAwsDynamoDbTableDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"arn": &schema.Schema{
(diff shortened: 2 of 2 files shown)


builtin/providers/aws/resource_aws_dynamodb_table.go
```diff
@@ -601,6 +604,44 @@ func resourceAwsDynamoDbTableRead(d *schema.ResourceData, meta interface{}) erro
...
+		lsi := map[string]interface{}{
+			"name":            *lsiObject.IndexName,
+			"projection_type": *lsiObject.Projection.ProjectionType,
+		}
+
+		for _, attribute := range lsiObject.KeySchema {
+
+			if *attribute.KeyType == "RANGE" {
+				lsi["range_key"] = *attribute.AttributeName
+			}
```
radeksimko: Did you possibly forget about `KeyType == "HASH"` here for local secondary indexes? I know these will always be the same as primary hash key, but AFAIK it doesn't mean they should be omitted from the API request. See line `260` in `Create` function.

builtin/providers/aws/resource_aws_dynamodb_table.go
```diff
@@ -601,6 +604,44 @@ func resourceAwsDynamoDbTableRead(d *schema.ResourceData, meta interface{}) erro
...
+		}
+		nkaList := make([]string, len(lsiObject.Projection.NonKeyAttributes))
+		for _, nka := range lsiObject.Projection.NonKeyAttributes {
+			nkaList = append(nkaList, *nka)
+		}
+		lsi["non_key_attributes"] = nkaList
```
radeksimko: I reckon that not setting `non_key_attributes` VS setting it to empty slice has the exact same effect - if that's the case we should be safe 👍 

If not, we should consider adding a the same condition as we have in `Create` above - i.e. only set `non_key_attributes` for `*lsiObject.Projection.ProjectionType == "INCLUDE"`

radeksimko:
Besides two comments above this LGTM.

builtin/providers/aws/resource_aws_dynamodb_table.go
stack72: Luckily, these are not computed values so these are fine to keep in the read :)

stack72: AFAICT this is safe, this isn't giving any adverse effects in the import. The import would show state that is unexpected

stack72:
@radeksimko addressed 1 of the 2 comments - AFAICT the addition of the empty slice is fine. The import would have showed the issue if there were one

builtin/providers/aws/import_aws_dynamodb_table_test.go
```diff
@@ -0,0 +1,30 @@
...
+			resource.TestStep{
+				Config: testAccAWSDynamoDbConfigInitialState(),
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				//ImportStateVerifyIgnore: []string{
+				//	"description"},
```
radeksimko: Forgotten code? 😛

builtin/providers/aws/resource_aws_dynamodb_table.go
radeksimko: As discussed via Slack this is actually correct since `KeyType` `HASH` is taken from `hash_key` field.

radeksimko:
LGTM (pending Travis).

PR #7370, opened 2016-06-27
Show Terraform version on state version mismatch.

tpounds:
Minor improvement that shows the current Terraform version when the detected state version is unsupported.


phinze:
Good call! Thanks for this @tpounds 👍

PR #5304, opened 2016-02-24
Added support for redshift destination to firehose delivery streams

dminnear:
Previously, firehose delivery streams only supported writing to s3. This PR adds redshift as a destination for firehose delivery streams. There is also some refactoring done to clean up the s3 configuration so existing delivery stream resources will be broken.

The related issue is #5152.

PR #7375, opened 2016-06-27
provider/aws: Added support for redshift destination to firehose delivery streams (supersedes #5304)

catsby:
This PR continues #5304 to add support for Redshift as a destination for Kinesis Firehose streams (and opens up the way to support ElasticSearch in the future).

It builds off of #5304 and adds:

- re-worked Acc tests to confirm remove values
- converts `s3_configuration` and `redshift_configuration` to `TypeList`; There Can Be Only One™ (of each) so that's fine and it helps with..
- migrating old state of flat parameters into the nested state of the `TypeList`:  
    `role_arn` => `s3_configuration.0.role_arn`
- provides removal warnings on the old attributes (re-added them as `Computed`, `Removed`)

Some of this is mine, but most of it was generously contributed by @dminnear . The reason for opening this new pr was because I had sat on #5304 for a while, waiting for a good time to introduce a backwards incompatible change (which we get in v0.7.0), and I didn't want to ask them to re-work or add a bunch of things after having waited so patiently for so long. 

Running acc tests here:

- [ ] Updated (https://github.com/hashicorp/terraform/commit/20314ce7e3c9ff1694cfa573a84dabfa923d0fc3): https://travis-ci.org/hashicorp/terraform/builds/140845034

### DIFF
--- a/builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
+++ b/builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
@@ -20,6 +20,8 @@ func resourceAwsKinesisFirehoseDeliveryStream() *schema.Resource {
 		Update: resourceAwsKinesisFirehoseDeliveryStreamUpdate,
 		Delete: resourceAwsKinesisFirehoseDeliveryStreamDelete,
 
+		SchemaVersion: 1,
+		MigrateState:  resourceAwsKinesisFirehoseMigrateState,
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
```diff
@@ -91,49 +186,179 @@ func resourceAwsKinesisFirehoseDeliveryStream() *schema.Resource {
...
 
+func validateConfiguration(d *schema.ResourceData) error {
+	destination := d.Get("destination").(string)
+	if destination != "s3" && destination != "redshift" {
+		return fmt.Errorf("[ERROR] Destination must be s3 or redshift")
+	}
+
+	s3Configuration := d.Get("s3_configuration").([]interface{})
+	if len(s3Configuration) > 1 {
+		return fmt.Errorf("[ERROR] You can only define a single s3_configuration per delivery stream")
```
apparentlymart: Is this also true of `redshift_configuration`?

stack72: We no longer need to do this, we can add `MaxItems: 1` to the schema for `s3_configuration` :)

builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
```diff
@@ -91,49 +186,179 @@ func resourceAwsKinesisFirehoseDeliveryStream() *schema.Resource {
...
+}
+
+func createRedshiftConfig(d *schema.ResourceData, s3Config *firehose.S3DestinationConfiguration) (*firehose.RedshiftDestinationConfiguration, error) {
+	redshiftRaw, ok := d.GetOk("redshift_configuration")
+	if !ok {
+		return nil, fmt.Errorf("[ERR] Error loading Redshift Configuration for Kinesis Firehose: redshift_configuration not found")
+	}
+	rl := redshiftRaw.([]interface{})
+	if len(rl) == 0 || len(rl) > 1 {
+		return nil, fmt.Errorf("[ERR] You can only define a single s3_configuration per delivery stream")
```
stack72: This error should says `redshift_configuration` rather than `s3_configuration`

builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
catsby: Updated in 20314ce, yay less code!

builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
catsby: D'oh, nice catch! Update in 20314ce

builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
```diff
@@ -91,49 +188,168 @@ func resourceAwsKinesisFirehoseDeliveryStream() *schema.Resource {
...
+	destination := d.Get("destination").(string)
+	if destination != "s3" && destination != "redshift" {
+		return fmt.Errorf("[ERROR] Destination must be s3 or redshift")
+	}
+
+	return nil
+}
+
+func createS3Config(d *schema.ResourceData) *firehose.S3DestinationConfiguration {
+	s3 := d.Get("s3_configuration").([]interface{})[0].(map[string]interface{})
```
phinze: Just noting the types are guaranteed by helper/schema and the length is guaranteed by Required: true + MaxItems: 1

phinze:
LGTM!

PR #7383, opened 2016-06-28
aws: Add new region (Mumbai)

radeksimko:
https://aws.amazon.com/about-aws/whats-new/2016/06/announcing-the-aws-asia-pacific-mumbai-region/

New ID verifiable via

```
aws ec2 describe-regions | jq .Regions[].RegionName | sort
```

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -328,9 +328,20 @@ func (c *Config) Client() (interface{}, error) {
 // ValidateRegion returns an error if the configured region is not a
 // valid aws region and nil otherwise.
 func (c *Config) ValidateRegion() error {
-	var regions = [12]string{"us-east-1", "us-west-2", "us-west-1", "eu-west-1",
-		"eu-central-1", "ap-southeast-1", "ap-southeast-2", "ap-northeast-1",
-		"ap-northeast-2", "sa-east-1", "cn-north-1", "us-gov-west-1"}
+	var regions = [12]string{
+		"ap-northeast-1",
+		"ap-northeast-2",
+		"ap-southeast-1",
+		"ap-southeast-2",
+		"cn-north-1",
+		"eu-central-1",
+		"eu-west-1",
+		"sa-east-1",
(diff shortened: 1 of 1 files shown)


stack72:
Nice work! LGTM! Merge when ready

stephenchu:
Mind me asking, why is Terraform keeping a list of hardcoded regions for its purpose?

radeksimko:
> why is Terraform keeping a list of hardcoded regions for its purpose?

I think this is purely for static/offline validation of the aws provider configuration - so that the user gets a human-friendly error for cases like this:

```hcl
provider "aws" {
  region = "not-exists"
}
```

We cannot verify this online, even though it's available from the API, because we'd need to choose a valid region (and hard-coding a single region is not a good idea since that region becomes a single point of failure) + have valid credentials so this is kind of a 🐔  & 🍳  problem when validating `provider` which is responsible for setting up the connection.

Also the way most SDKs work is that they simply take the given region as a string and make a hostname of it, so the error message user would see if Terraform wouldn't perform this offline validation is something like this:

```
Could not connect to the endpoint URL: "https://sts.not-exists.amazonaws.com/"
```

What I think would make sense to do though is to keep such list of valid regions in the SDK and let the SDK perform the validation, but SDK doesn't currently offer such functionality neither a list of regions.

boxrick:
Will this make it into the RC3 at some point shortly?

stack72:
@boxrick not sure what the release plans are - that's a question for @phinze :)

boxrick:
@phinze Will this make it into TF 0.7? Fairly simple change but it is stopping some work and I would appreciate if it could be merged in :)

stack72:
Hi @boxrick 

this has already been merged - so it will definitely be in 0.7

Paul

stack72:
FWIW, it's already part of the RC3

radeksimko:
@stephenchu FYI - I just opened https://github.com/hashicorp/terraform/pull/11358 to add `skip_region_validation`. Feedback welcomed there.

PR #7389, opened 2016-06-28
provider/azurerm: Add MaxItems to `azurerm_virtual_machine` data_disk

### LINKED ISSUE #7321: count on azure_data_disk
Hi,

would appreciate adding support for creating an array of data disks on azure. 

https://www.terraform.io/docs/providers/azure/r/data_disk.html

stack72:
Fixes #7321

When testing to see how many data_disks could be added to a machine, I
got the following response from the Azure RM API

```
{
	"error": {
		"code": "OperationNotAllowed",
		"target": "dataDisks",
		"message": "The maximum number of data disks allowed to be attached to a VM is 1."
	}
}
```

We should ensure the schema copes with this so as to avoid waiting on the creation to throw the error


stack72:
Not the case, data disks are limited by machine type - not a hard count :)

PR #7268, opened 2016-06-22
[WIP] schema: Get() of a nested Set in List outputs empty set

radeksimko:
Related to https://github.com/hashicorp/terraform/issues/7253 & https://github.com/hashicorp/terraform/pull/7254

```
$ make test TEST=./helper/schema
```
```
TF_ACC= go test ./helper/schema  -timeout=30s -parallel=4
--- FAIL: TestMultiLevelFieldReader_ReadField_SetInList (0.00s)
	field_reader_multi_test.go:431: Expected value to be not nil: []interface {}{interface {}(nil)}
--- FAIL: TestSetInsideList_simple (0.00s)
	resource_data_test.go:877: Expected main_list to be not nil: []interface {}{interface {}(nil)}
--- FAIL: TestSetInsideList_complex (0.00s)
	resource_data_test.go:946: Expected exactly 1 instance of main_list, got 2
FAIL
FAIL	github.com/hashicorp/terraform/helper/schema	0.024s
make: *** [test] Error 1
```

I managed to fix two simple failing test cases <sup>[1]</sup>, then came up with more complicated ResourceData test case (`84f8186` / `TestSetInsideList_complex`) which is failing:

```
TF_ACC= go test ./helper/schema  -timeout=30s -parallel=4
--- FAIL: TestSetInsideList_complex (0.00s)
	resource_data_test.go:956: Given: []interface {}{}

		Expected: []interface {}{"blue", "green"}
FAIL
FAIL	github.com/hashicorp/terraform/helper/schema	0.022s
make: *** [test] Error 1
```

-------

<sup>[1]</sup> Bugfix is not part of this PR yet as I wanted to demonstrate the problem 1st before coming with a solution.

PR #7254, opened 2016-06-21
Fix CloudFront cache_behavior precedence

optimisticanshul:
Fixes #7253

Issue #7253, opened 2016-06-21
Allow configuring CloudFront cache_behavior precedence


randomeizer:
### Terraform Version

0.6.16
### Affected Resource(s)

Please list the resources as a list, for example:
- cloudfront_distribution > cache_behavior
### Terraform Configuration Files

`````` hcl
# Distro pointing at the specified domain.
resource "aws_cloudfront_distribution" "cloudfront" {
  origin {
    domain_name = "${var.lde_vm_domain_name}"
    origin_id = "lde_vm"
    custom_origin_config {
      http_port = 80
      https_port = 443
      origin_protocol_policy = "match-viewer"
      origin_ssl_protocols = [
        "SSLv3",
        "TLSv1"]
    }
    custom_header = {
      name = "${var.lde_custom_header_name}"
      value = "${var.lde_custom_header_value}"
    }
  }
  aliases = ["${var.public_domain_name}"]
  enabled = true
  comment = "CloudFront Server"
  default_cache_behavior {
    target_origin_id = "lde_vm"
    allowed_methods = [
      "DELETE",
      "GET",
      "HEAD",
      "OPTIONS",
      "PATCH",
      "POST",
      "PUT"]
    cached_methods = [
      "GET",
      "HEAD"]
    compress = true
    smooth_streaming = false
    forwarded_values {
      cookies {
        forward = "all"
      }
      headers = ["Host"]
      query_string = true
    }
    viewer_protocol_policy = "allow-all"
    min_ttl = 0
    default_ttl = 3600
    max_ttl = 31536000
  }

# private digital content cache
  cache_behavior {
    path_pattern = "user/digital_content/download/*"
    target_origin_id = "lde_vm"
    allowed_methods = [
      "GET",
      "HEAD",
      "OPTIONS"]
    cached_methods = [
      "GET",
      "HEAD"]
    compress = true
    smooth_streaming = false
    forwarded_values {
      cookies {
        forward = "none"
      }
      headers = ["Host"]
      query_string = true
    }
    trusted_signers = ["self"]
    viewer_protocol_policy = "allow-all"
    min_ttl = 0
    default_ttl = 3600
    max_ttl = 31536000
  }

  # JS files
  cache_behavior {
    path_pattern = "js/*"
    target_origin_id = "lde_vm"
    allowed_methods = [
      "GET",
      "HEAD",
      "OPTIONS"]
    cached_methods = [
      "GET",
      "HEAD"]
    compress = true
    smooth_streaming = false
[...]

PR #7393, opened 2016-06-28
helper/schema: Make nested Set(s) in List(s) work

radeksimko:
Related to #7253 & #7254
Supersedes/fixes #7268

-----

Issue explained below in (hopefully) simple ASCII art.

**resource schema**
```
.
├── TypeInt
└── TypeList
    └── TypeSet
    └── TypeString
    └── TypeInt
    └── TypeBool
```
**resource CRUD**
```
resourceXyzCreate() {
  d.Get("TypeInt") # OK
  d.Get("TypeList") # OK
}
resourceXyzRead() {
  d.Get("TypeInt") # OK
  d.Get("TypeList") # OK
}
resourceXyzUpdate() {
  d.Get("TypeInt") # OK
  d.Get("TypeList") # returns *Set(TypeSet=NIL, TypeString, TypeInt, TypeBool)
}
```
The nested TypeSet inside TypeList was set to `nil` and any data further nested in that Set were ignored inside `Update` function of a resource **unless that `TypeSet` field was actually contained in the diff**.

------

### Summary (couple of whys)

 - Previously there was a common _exported_ `ReadList` function shared between all field readers (`Diff`, `Map`, `Config`), this was replaced with _unexported_ implementations per each reader
   - it turns out each reader requires different approach, especially `DiffFieldReader`
   - this is a first step towards getting rid of the magical `nestedConfigFieldReader`
   - reader-specific function can call `readPrimitive` and doesn't have to perform the data-type detection anymore - theoretically should save us some CPU cycles and prevent the infinite loops (one of the reasons why `nestedConfigFieldReader` was there)
 - `readSet` implementation of the `Diff` reader was only reading the diff and worked fine for creation/deletion of whole sets, but not so well for cases where only part of the set was being updated (contained in the diff).
   - `readSet` now fills in the gaps by getting the missing data from previous reader

### Tests, yey

About 84% LOC of this PR are just tests covering the broken (now fixed) functionality - all contained in a separate commit.

-----

Another PR will follow to fix the original CloudFront bug with data types.

### DIFF
--- a/helper/schema/field_reader_config.go
+++ b/helper/schema/field_reader_config.go
@@ -83,7 +83,7 @@ func (r *ConfigFieldReader) readField(
 	case TypeBool, TypeFloat, TypeInt, TypeString:
 		return r.readPrimitive(k, schema)
 	case TypeList:
-		return readListField(&nestedConfigFieldReader{r}, address, schema)
+		return r.readList(address, schema)
 	case TypeMap:
 		return r.readMap(k)
 	case TypeSet:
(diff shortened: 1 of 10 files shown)


helper/schema/field_reader_config.go
```diff
@@ -216,13 +216,65 @@ func (r *ConfigFieldReader) readPrimitive(
...
+		addrPadded[len(addrPadded)-1] = idx
+		rawResult, err := r.readField(addrPadded, true)
+		if err != nil {
+			return FieldReadResult{}, err
+		}
+		if !rawResult.Exists {
+			// This should never happen, because by the time the data
+			// gets to the FieldReaders, all the defaults should be set by
+			// Schema.
+			panic("missing field in list: " + strings.Join(addrPadded, "."))
```
phinze: +1 for panic here - as soon as I saw the "This should never happen" in the old shared impl i was like 💭  "we should panic then!" 😀

helper/schema/field_reader_diff.go
```diff
@@ -93,13 +95,13 @@ func (r *DiffFieldReader) readMap(
...
 	var resultVal interface{}
-	if resultSet {
+	if exists {
 		resultVal = result
 	}
 
 	return FieldReadResult{
 		Value:  resultVal,
-		Exists: resultSet,
+		Exists: exists,
```
phinze: <3 good clean up

helper/schema/field_reader_config.go
radeksimko: Right, that's more or less inherited from the original generic `ReadList` function, but it's still valid here too. 😄

helper/schema/field_reader_map.go
```diff
@@ -91,6 +92,57 @@ func (r *MapFieldReader) readPrimitive(
...
 }
 
+func (r *MapFieldReader) readList(
+	address []string, schema *Schema) (FieldReadResult, error) {
+
+	addrPadded := make([]string, len(address)+1)
+	copy(addrPadded, address)
+
+	// Get the number of elements in the list
+	addrPadded[len(addrPadded)-1] = "#"
```
phinze: This should be a `%` nowadays, no?

phinze: Oh nm, it's a TypeList in the MapFieldReader. Saw "Map" and assumed TypeMap. All good!

phinze:
Okay! This looks great @radeksimko, especially the extensive testing work.

Merge at will. 👍 🚀

catsby:
Hello –

We've found some regressions in our tests and suspect this PR. I'm going to issue a revert, but first, here are some examples:

```
--- FAIL: TestAccAWSSecurityGroup_Change (25.93s)
    testing.go:255: Step 1 error: Error applying: 1 error(s) occurred:

        * aws_security_group.web: Error authorizing security group ingress rules: InvalidParameterValue: Invalid value '' for IP protocol. No protocol value given.
```

```
--- FAIL: TestAccFastlyServiceV1_gzips_basic (15.47s)
    testing.go:255: Step 1 error: Error applying: 1 error(s) occurred:
        
        * fastly_service_v1.foo: 400 - Bad Request
        Message: Name can't be blank
```

From https://travis-ci.org/hashicorp/terraform/builds/141247532 and https://travis-ci.org/hashicorp/terraform/builds/141254784, respectively.

PR #7376, opened 2016-06-27
provider/aws: Elastic Beanstalk scheduledaction

dharrisio:
Add support for scheduled actions in Elastic Beanstalk option settings
by adding optional `resource` attribute for option setting resource. #6476 

```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSBeanstalkEnv_resource -timeout 120m
=== RUN   TestAccAWSBeanstalkEnv_resource
--- PASS: TestAccAWSBeanstalkEnv_resource (382.48s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    382.500s
```

PR #7398, opened 2016-06-28
provider/aws: fix Elastic Beanstalk settings diff

dharrisio:
Fixes an issue where terraform plan or apply will always have a diff.
The Elastic Beanstalk API returns data for the `resource` attribute for
some settings that are not documented. This limits the `resource`
attribute to settings in the `aws:autoscaling:scheduledaction`
namespace.

@stack72 This is a follow up for #7376.


stack72:
@dharrisio thanks for the follow up here! :)

PR #7402, opened 2016-06-28
core: increment the state serial whenever we upgrade

jbardin:
Always increment the state serial whenever we upgrade the state version.
This prevents possible version conflicts between local and remote state
when one has been upgraded, but the serial numbers match.


phinze:
LGTM

PR #7403, opened 2016-06-28
core: Don't set Modules to nil during state upgrade

### LINKED ISSUE #7394: core: crash during eval with empty V1 state
### Terraform Version
v0.7.0-rc2-183-g5e52bba

terraform.tfstate:
```json
{
    "version": 1,
    "serial": 0,
    "modules": [
        {
            "path": [
                "root"
            ],
            "outputs": {},
            "resources": {}
        }
    ]
}
```

main.tf:
```hcl
resource "aws_security_group" "baz" {
   name_prefix = "baz-"
   description = "Used in the terraform acceptance tests"
}
```

running `terraform plan` will panic

### Panic Output
https://gist.github.com/jbardin/3be806770a3f2694fbf41616e705521c

jbardin:
The state Modules map is accessed and always expected to be non-nil. The
V1->V2 upgrade was setting the map to nil if the len == 0.

Fixes #7394

### DIFF
--- a/terraform/state_upgrade_v1_to_v2.go
+++ b/terraform/state_upgrade_v1_to_v2.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	"fmt"
+
 	"github.com/mitchellh/copystructure"
 )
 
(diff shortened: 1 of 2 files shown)


phinze:
Conceptually seems right - can we drop a unit test in for this?

jbardin:
Confirmed that since the module init method initialized the maps by default, they are probably expected to be non-nil throughout.

PR #7392, opened 2016-06-28
provider/docker: Docker DNS Setting Enhancements

dmportella:
Added support for docker dns settings: --dns-opt and --dns-search

Also fixed vet issues on aws provider builtin

Tests updated

You can now specify the following on your docker container resource

```
resource "docker_container" "website" {
	count = 1
	image = "${docker_image.website.name}"
	name = "website"
	dns = ["8.8.8.8"]

	#new fields	
	dns_opts = ["rotate", "user-vc"]
	dns_search = ["example.com."]

	ports {
		internal = 8080
		external = 8080
	}
}

resource "docker_image" "website" {
	name = "website:1.0.0"
}
```

PR #7412, opened 2016-06-29
provider/docker: Docker documentation and additional test message

dmportella:
As discussed on: https://github.com/hashicorp/terraform/pull/7392

Added

- additional information in case of docker test fail on `memory_swap` assert.
- updated website documentation to reflect new additions on https://github.com/hashicorp/terraform/pull/7392
  - added `dns_opts`, `dns_search` information.
  - modified `memory_swap` to highlight that if 'terraform apply` is run on a system that doesn't support memory swap the property will always compute to -1 as it is returned like this by docker rest api.

### DIFF
--- a/builtin/providers/docker/resource_docker_container_test.go
+++ b/builtin/providers/docker/resource_docker_container_test.go
@@ -95,7 +95,7 @@ func TestAccDockerContainer_customized(t *testing.T) {
 		}
 
 		if c.HostConfig.MemorySwap != (2048 * 1024 * 1024) {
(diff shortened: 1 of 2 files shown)


dmportella:
@stack72 changes are here. Let me know if you need anything else done here.

stack72:
@dmportella

Thanks for the work here to round off that PR :)

P.

dmportella:
@stack72 no worries

Will be doing more in the future there is a lot of properties missing from docker which are just simple maps.

stack72:
@dmportella

We welcome the PRs :)

P.

PR #7418, opened 2016-06-29
core: Add test to ensure key/values interpolated order

jbardin:
Test that interpolated values are returned in the order of the sorted keys.


phinze:
Nice - LGTM

PR #7424, opened 2016-06-30
provider/vsphere: virtual_machine support for more features

ChloeTigre:
## Abstract
We need to tell vSphere what OS is run into the VM.
We also need to be able to specify explicitly what kind of adapter is wanted in the NIC.

So here is the action:
Allow to specify explicitly guest_id in VM and adapter_type in network section of VM.

## Default changes
None. By default it will perform the same as before.

## What's possible with this?
Specify explicitly what kind of network adapter you want. Specify what OS will be deployed in the VM, straight from Terraform.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -76,7 +76,7 @@ type memoryAllocation struct {
 	reservation int64
 }
 
-type virtualMachine struct{
+type virtualMachine struct {
 	name                  string
 	folder                string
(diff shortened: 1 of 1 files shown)


ChloeTigre:
Bump. There was a missing go fmt.

ChloeTigre:
Note: added support for thick_lazy disk provisioning.

ChloeTigre:
I won't move this branch any more. The refactor is to be seen at #7742. I'll add the other features of this PR there so this may be closed.
@stack72 if you want to review the other, It'll be ready today.

PR #7439, opened 2016-06-30
core: Fix plan output for data sources

### LINKED ISSUE #6823: Data resources in non-root modules render incorrectly in plan output
terraform-nested-module-data

The plan renderer has a special case for resources whose names begin with `data.`, but when data resources appear inside nested modules they don't actually have this prefix and so the diff is presented as if it were an update of a managed resource.

jbardin:
Data sources are given a special plan output when the diff would show it
creating a new resource, to indicate that there is only a logical data
resource being read. Data sources nested in modules weren't formatted in
the plan output in the same way.

This checks for the "data." part of the path in nested modules, and adds
acceptance tests for the plan output.

Fixes #6823

### DIFF
--- a/command/format_plan.go
+++ b/command/format_plan.go
@@ -78,6 +78,8 @@ func formatPlanModuleExpand(
 			continue
 		}
 
+		dataSource := strings.HasPrefix(name, "data.")
+
 		if moduleName != "" {
 			name = moduleName + "." + name
 		}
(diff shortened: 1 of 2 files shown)


phinze:
> and adds acceptance tests for the plan output.

Looks like we need a `git add` here. 😁

jbardin:
hmm, I did, must have un-added it somehow with a `push -f` :/

phinze:
Solid, LGTM!

Issue #3605, opened 2015-10-22
provisioner/chef: unregister the client from chef server when replacing an instance


pmvilaca:
It would be useful to get terraform deleting the client from chef server for the instances that are being replaced. Otherwise, the replacement operation will always fail during the provisioning task because the node already exists on the chef server (assuming that you're defining the node_name)

Example:

```
provisioner "chef"  {
    run_list = ["role[common]"]
    node_name = "nat_instance_ffm_1a"
   ...
```

Result:
```
aws_instance.nat_1b (chef): ================================================================================
aws_instance.nat_1b (chef): Chef encountered an error attempting to create the client "nat_instance_ffm_1b"
aws_instance.nat_1b (chef): ================================================================================
```

partamonov:
I guess this is related to #649 #2010 #2720. And I agree this is nice to have.

apparentlymart:
I left [a comment about this](https://github.com/hashicorp/terraform/issues/386#issuecomment-154596719) in #386 way back in the days of three-digit issues. :grinning:

Right now this sort of thing is not supported by Terraform's architecture since it doesn't allow for actions on delete, but that issue #386 is about supporting that, which could then in turn solve this more-specific problem of leaving junk nodes in Chef Server. In our case we were naming the nodes after the EC2 instance id and so the outcome was hundreds of defunct nodes rather than a naming conflict, but still annoying especially since paid chef uses "number of nodes" as a quota metric.

bitglue:
> not supported by Terraform's architecture since it doesn't allow for actions on delete

Isn't the destroy action on a resource exactly that?

It would seem to me that what's missing is a Terraform resource that represents the membership in a chef server. Then you not only have a way to add clients, and remove clients, but also reconcile state and recover from failures.

rberger:
Does it look like this is something anyone is working on? Its a huge pain point at least for me and I would assume anyone who is using Terraform and Chef.

svanharmelen:
@rberger no I don't think so unfortunately... It's a thing that needs to be added to Terraform core and I didn't had time to have a look at it lately...

orclev:
Just voicing my wish for this as well, currently before we can run a terraform apply we need to manually go in and delete all the registered clients in chef.

svanharmelen:
This should be fixed by PR #8577 which introduces a new attribute called `recreate_client`. If set to `true`, it will first delete the existing node and client before registering the new one.

PR #7440, opened 2016-06-30
provisioner/chef: Add bootstrapping with knife to support chef-vault

jmccann:
## Use Case
I would like the ability to bootstrap a node that has a `run_list` that utilizes vaults from chef-vault

## Problem
Current chef provisioner bootstraps a node by using a validator and running chef-client on the node.  This does not allow you to add the node to chef-vaults and therefore can not have an initial `run_list` that includes cookbooks/recipes that utilize a chef-vault.

## Fix
I am adding code to the chef provisioner to allow bootstrapping a node with `knife bootstrap` rather then running chef-client and using a validator.  Newer versions of knife included with Chef allow "validatorless" bootstrapping and supports flag `--bootstrap-vault-json` which allows specifying chef-vaults to add the node to.

### DIFF
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -46,6 +46,50 @@ func (p *Provisioner) linuxInstallChefClient(
 	return p.runCommand(o, comm, fmt.Sprintf("%srm -f install.sh", prefix))
 }
 
+func (p *Provisioner) linuxCreateValidationKey(
+	o terraform.UIOutput,
+	comm communicator.Communicator) error {
+	if err := p.deployValidationKey(o, comm, linuxTmpDir); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Provisioner) linuxBootstrapSetup(
+	o terraform.UIOutput,
+	comm communicator.Communicator) error {
+
+	if err := p.runCommand(o, comm, "ssh-keygen -t rsa -N '' -f /root/.ssh/id_rsa"); err != nil {
+		return err
+	}
+	if err := p.runCommand(o, comm, "cp /home/$(whoami)/.ssh/authorized_keys /home/$(whoami)/.ssh/authorized_keys.pre_bootstrap"); err != nil {
+		return err
+	}
+	if err := p.runCommand(o, comm, "bash -c \"cat /root/.ssh/id_rsa.pub >> /home/$(whoami)/.ssh/authorized_keys\""); err != nil {
+		return err
+	}
+	if err := p.runCommand(o, comm, "/opt/chef/embedded/bin/gem install chef-vault"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Provisioner) linuxBootstrapCleanup(
+	o terraform.UIOutput,
+	comm communicator.Communicator) error {
+
+	if err := p.runCommand(o, comm, "cp /home/$(whoami)/.ssh/authorized_keys.pre_bootstrap /home/$(whoami)/.ssh/authorized_keys"); err != nil {
+		return err
+	}
+	if err := p.runCommand(o, comm, "rm -f /home/$(whoami)/.ssh/authorized_keys.pre_bootstrap /root/.ssh/id_rsa* /tmp/validation.pem"); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (p *Provisioner) linuxCreateConfigFiles(
 	o terraform.UIOutput,
 	comm communicator.Communicator) error {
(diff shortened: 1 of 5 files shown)


svanharmelen:
@jmccann thanks for the PR! I can tell you invested some time into this one, but I must admit that it feels a little weird to me to use the `knife bootstrap` approach for this... IMO it would make much more sense to just call the `knife bootstrap` command from a `local-exec` provisioner, instead of calling it from the remote host with a target address of `127.0.0.1`.

I understand that you miss some functionality in the current `Chef` provisioner, but I wonder if this is the best way to add that functionality. I will need to find some time to take a closer look at this PR (will try to get around to that shortly) and will also do some thinking about if there are better approaches or that this is the only way to achieve your goals...

jmccann:
@svanharmelen Thanks for taking a quick look at this and for your comments.  I understand some of the concerns you bring up and initially had the same concerns.  I'll explain why I feel this is an OK approach and how it helps me and people I work with.

First let me clarify I'm by no means a terraform expert and am learning things as I go.  ;)

I'm trying to use source control and CI/CD to manage my infrastructure.  Specifically I'm using Drone with [drone-terraform](https://github.com/drone-plugins/drone-terraform) for managing my infra with terraform.  Unfortunately drone-terraform does not contain the Chef stack inside of it so `local-exec` using that would not work.

What I had been doing prior to creating this code and PR was passing `user_data` to my openstack instances that did pretty much the same thing.  I have multiple openstack instances that I do not have direct access to.  The problem with this approach was every time I try to update some minor Chef attribute it wants to rebuild the system.  Another problem was it was hard to explain to people what was going on.

I then saw other co-workers using the chef provisioner to get around both of these issues by setting a bastion host for the connection.  This was pretty much exactly what I wanted and was very easy to use except that it does not support chef-vaults.

One thing I did think of before working on this was using `remote-exec` to run the knife bootstrap.  But that also by default runs on the system I am provisioning.  So I would basically end up doing in `remote-exec` what I've tried to add to `provisioner/chef` (bootstrapping a node from itself).  But it would be harder for people to consume later.

For me having a good pattern to give people I work with for provisioning Chef, configuring Chef and having it have access to chef-vaults on the first Chef run is what I'm trying to accomplish.  A lot of people are already familiar with using `provisioner/chef`.  So I thought extending it to also support bootstrapping a node into a chef-vault would be good.

jmccann:
@svanharmelen I was wondering if you have some time to review this and think it over more.  Again, my use case is the ability to bootstrap a system with access to chef-vaults using the chef provisioner.

I think the primary concern should be attributes I've added to the provisioner (`bootstrap` and `vaults_json`).  If the frontend is good then people can start using this feature as is.  Then as people have time and/or find better ways to implement the backend it can then be refactored without impact to the user. For example, I've thought of removing `bootstrap` and having the bootstrap method determined solely based on the presence of `vaults_json`.

I'm hoping this is a wanted feature and that we can work through any implementation details.

Thanks!

svanharmelen:
@jmccann sorry for the lack of response here! I'm currently on holiday, but will be back coming week (Tuesday) and will make time to process this one so we can move it forward.

So please give me another few days and then see if we can get this (or a slightly adjusted version) merged end of next week.

Thx!

svanharmelen:
So I ended up ill the day before getting back from holiday an so I didn't have a look yet... Did think about possible solutions and I think I have a good approach which I will try to work out today/tomorrow.

Will keep you posted!

svanharmelen:
So I think I have a very nice solution, but I will need to run some tests on OSX, Linux and Windows to confirm it all works as expected first... Will update and share (if all works as expected) the solution after I'm done testing.

jmccann:
@svanharmelen Awesome!  Excited to see what you have.

svanharmelen:
The solution I'm working on also fixes issue #3605. I managed to test and verify that part, but got disturbed with other work when I was about to go and test the vault part... So that will have to wait until tomorrow, but it looks good :wink:

svanharmelen:
@jmccann again sorry for all the delays, but please have a look at PR #8577.

The only code changes that are directly related to this PR are in this [file](https://github.com/hashicorp/terraform/pull/8577/files#diff-033150dd6357bb8d6e1a2c44b80d689b) and as you can see there are effectively two new functions here. One for fixing issue #3605 and one for adding Chef Vault support.

The rest is mainly updating and cleaning code and tests in order to get the Chef provisioner up-to-spec again.

Let me know what you think of this approach and if you are able to build from this branch and give it a test run, that would be great.

mrmarbury:
Hi,

does this also include vaults that are already on the server and would be added to the bootstrap command via `--bootstrap-vault-item` or is it just for vaults that are created with the new Node?

Cheers

jmccann:
@mrmarbury This is to allow bootstrapping a new node to existing vaults on the server.  This work is being done under #8577 now.

bottkv488:
@jmccann, @svanharmelen 
Thank you so much for taking care of this, as
running ```knife vault refresh foo bar``` inside a ```local-exec``` block is such a time consuming nuisance

jmccann:
Closing in favor of #8577

PR #7441, opened 2016-06-30
provider/aws: Allow `port` on `aws_db_instance` to be updated

### LINKED ISSUE #2439: aws_db_instance PORT change forces new resources - should just modify existing
I created an rds resource, populated it with data, and would like to change its name now because it will conflict with other VPCs that use my terraform configs.

`identifier:                       "rds-db" => "rds-db-production" (forces new resource)`

In the aws console you are able to modify the ID of the resources, it does require a reboot, however does not destroy the resource.

Temp workaround: We just used the console to modify the resource, manually modified the state file, and ran a terraform plan and it runs clean - all that was updated was the route53 record that points to the rds instance, which is expected.

stack72:
Fixes #2439

Port used to ForceNew, it has now been changed to allow it to be
updated. 2 changes were needed:

1. Setting the port back to state
2. Adding the wait for state function to the Update func

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSDBInstance_portUpdate'
-run=TestAccAWSDBInstance_portUpdate -timeout 120m
=== RUN   TestAccAWSDBInstance_portUpdate
--- PASS: TestAccAWSDBInstance_portUpdate (699.84s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    699.861s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_db_instance.go
+++ b/builtin/providers/aws/resource_aws_db_instance.go
@@ -157,7 +157,6 @@ func resourceAwsDbInstance() *schema.Resource {
 				Type:     schema.TypeInt,
 				Optional: true,
 				Computed: true,
-				ForceNew: true,
 			},
 
 			"publicly_accessible": &schema.Schema{
@@ -650,6 +649,7 @@ func resourceAwsDbInstanceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("publicly_accessible", v.PubliclyAccessible)
 	d.Set("multi_az", v.MultiAZ)
 	d.Set("kms_key_id", v.KmsKeyId)
+	d.Set("port", v.DbInstancePort)
 	if v.DBSubnetGroup != nil {
 		d.Set("db_subnet_group_name", v.DBSubnetGroup.DBSubnetGroupName)
 	}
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_db_instance.go
```diff
@@ -916,13 +916,37 @@ func resourceAwsDbInstanceUpdate(d *schema.ResourceData, meta interface{}) error
...
+
 	log.Printf("[DEBUG] Send DB Instance Modification request: %t", requestUpdate)
 	if requestUpdate {
 		log.Printf("[DEBUG] DB Instance Modification request: %s", req)
 		_, err := conn.ModifyDBInstance(req)
 		if err != nil {
 			return fmt.Errorf("Error modifying DB Instance %s: %s", d.Id(), err)
 		}
+
+		log.Println("[INFO] Waiting for DB Instance to be available")
```
catsby: Probably change this to be "updated" instead of available. Does this refresh change anything else? Meaning, if you have apply immediately, does it wait until the change is applied?

stack72: The state of the database goes back to `available` after it has finished updating

This will wait until the changes have applied - the update didn't before

stack72:
@catsby 

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSDBInstance_iops_update'                    
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSDBInstance_iops_update -timeout 120m
=== RUN   TestAccAWSDBInstance_iops_update
--- PASS: TestAccAWSDBInstance_iops_update (4553.91s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	4553.932s
```

builtin/providers/aws/resource_aws_db_instance.go
catsby: > The state of the database goes back to available after it has finished updating

I think this is true for some attributes, like `port`, which happen to change quickly. I'm not sure what others there that behave like this, but things like `allocated_storage` do not. We still see a plan/diff on follow up if we're not using the `apply_immediately` flag. 

That said, it prevents unnecessary plan/diffs for things like `port`, so we'll go with it 👍

PR #7445, opened 2016-06-30
provider/aws: Read Elastic Beanstalk stack name

### LINKED ISSUE #7300: Elastic Beanstalk Solution Stack Name stale in State
Running terraform 0.6.16.

In my statefile stored on atlas, it has: `"solution_stack_name": "64bit Amazon Linux 2016.03 v2.1.1 running Multi-container Docker 1.9.1 (Generic)"`

However when I query the same environment using AWS's API, it is still on `v2.1.0`

```
[~]$ aws --region us-east-1 elasticbeanstalk describe-environments --application-name tf-rise-runtime
{
    "Environments": [
        {
            "ApplicationName": "tf-rise-runtime",
            "EnvironmentName": "rise-runtime",
            "VersionLabel": "0f37b4c81b01d3e83c1162b86595a3e9bb1bdf72-jenkins-eb-deploy-stage-1152",
            "Status": "Ready",
            "EnvironmentLinks": [],
            "EnvironmentId": "e-3m3iciyvzs",
            "EndpointURL": "awseb-e-3-AWSEBLoa-S3Z7AKU3CQJL-583246180.us-east-1.elb.amazonaws.com",
            "SolutionStackName": "64bit Amazon Linux 2016.03 v2.1.0 running Multi-container Docker 1.9.1 (Generic)",
            "CNAME": "rise-runtime.yuh2qyr258.us-east-1.elasticbeanstalk.com",
            "Health": "Green",
            "AbortableOperationInProgress": false,
            "Tier": {
                "Version": " ",
                "Type": "Standard",
                "Name": "WebServer"
            },
            "HealthStatus": "Ok",
            "DateUpdated": "2016-06-22T15:43:54.439Z",
            "DateCreated": "2016-06-01T23:01:17.618Z"
        }
    ]
}
```

As you can see, the solution stack name is actually `"SolutionStackName": "64bit Amazon Linux 2016.03 v2.1.0 running Multi-container Docker 1.9.1 (Generic)"`

I do not know why terraform isn't trying to keep this in sync and update it to `2.1.1`

dharrisio:
Fixes #7300 where an Elastic Beanstalk error or change outside of Terraform modifies the `solution_stack_name` attribute, which results in a stale value in the state file.

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
@@ -404,6 +404,10 @@ func resourceAwsElasticBeanstalkEnvironmentRead(d *schema.ResourceData, meta int
 		}
 	}
 
+	if err := d.Set("solution_stack_name", *env.SolutionStackName); err != nil {
+		return err
+	}
+
 	if err := d.Set("autoscaling_groups", flattenBeanstalkAsg(resources.EnvironmentResources.AutoScalingGroups)); err != nil {
 		return err
 	}


stack72:
Hi @dharrisio 

Just ran through the tests and they look good to me :)

Thanks for this

Paul

stack72:
Actually, @dharrisio 

On second look at the tests, we are getting a plan cycle

```
=== RUN   TestAccAWSBeanstalkEnv_config
--- FAIL: TestAccAWSBeanstalkEnv_config (401.23s)
	testing.go:255: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		UPDATE: aws_elastic_beanstalk_environment.tftest
		  solution_stack_name: "64bit Amazon Linux running Python" => ""

```

dharrisio:
@stack72 The tests fails because the `solution_stack_name` and `template_name` are mutually exclusive. When the `template_name` is used, the `solution_stack_name` is still read from the environment. When this happens, Terraform will try and correct this by setting `solution_stack_name` to nothing.

I was able to get around this by setting `solution_stack_name` to computed, but this means it can no longer use `ConflictsWith` on the `template_name`. With this change, if the Terraform document has both attributes, the Elastic Beanstalk API will return an error. I think that will be ok, but wanted to check if there was a better way to handle this.

stack72:
Hi @dharrisio 

If this is Computed, then let's make it so :) Do they both need to be Computed?

P.

dharrisio:
I think that only the `solution_stack_name` is computed. This will have a value if `solution_stack_name` or `template_name` is used. `template_name` is optional, and should only have a value if it is explicitly used in the Terraform document. 

Running a few more tests to verify, but should have a new pr for this up today.

PR #7446, opened 2016-06-30
core: rerun resource validation before plan and apply

### LINKED ISSUE #7170: Crash report from  0.7.0 rc2
### Debug Output

https://gist.github.com/trauts2/6df7d91eb743b0258dc2ab94479f681c

### Panic Output

Gist with debug attached

[crash.log.zip](https://github.com/hashicorp/terraform/files/315275/crash.log.zip)

phinze:
In #7170 we found two scenarios where the type checking done during the
`context.Validate()` graph walk was circumvented, and the subsequent
assumption of type safety in the provider's `Diff()` implementation
caused panics.

Both scenarios have to do with interpolations that reference Computed
values. The sentinel we use to indicate that a value is Computed does
not carry any type information with it yet.

That means that an incorrect reference to a list or a map in a string
attribute can "sneak through" validation only to crop up...

 1. ...during Plan for Data Source References
 2. ...during Apply for Resource references

In order to address this, we:

 * add high-level tests for each of these two scenarios in `provider/test`
 * add context-level tests for the same two scenarios in `terraform`
   (these tests proved _really_ tricky to write!)
 * place an `EvalValidateResource` just before `EvalDiff` and `EvalApply` to
   catch these errors
 * add some plumbing to `Plan()` and `Apply()` to return validation
   errors, which were previously only generated during `Validate()`
 * wrap unit-tests around `EvalValidateResource`
 * add an `IgnoreWarnings` option to `EvalValidateResource` to prevent
   active warnings from halting execution on the second-pass validation

Eventually, we might be able to attach type information to Computed
values, which would allow for these errors to be caught earlier. For
now, this solution keeps us safe from panics and raises the proper
errors to the user.

Fixes #7170

### DIFF
--- a/terraform/eval_validate.go
+++ b/terraform/eval_validate.go
@@ -108,11 +108,13 @@ type EvalValidateResource struct {
 	ResourceName string
 	ResourceType string
 	ResourceMode config.ResourceMode
+
+	// IgnoreWarnings means that warnings will not be passed through. This allows
+	// "just-in-time" passes of validation to continue execution through warnings.
+	IgnoreWarnings bool
 }
 
 func (n *EvalValidateResource) Eval(ctx EvalContext) (interface{}, error) {
-	// TODO: test
-
 	provider := *n.Provider
 	cfg := *n.Config
 	var warns []string
@@ -127,16 +129,15 @@ func (n *EvalValidateResource) Eval(ctx EvalContext) (interface{}, error) {
 		warns, errs = provider.ValidateDataSource(n.ResourceType, cfg)
 	}
 
-	// If the resouce name doesn't match the name regular
-	// expression, show a warning.
+	// If the resource name doesn't match the name regular
+	// expression, show an error.
 	if !config.NameRegexp.Match([]byte(n.ResourceName)) {
 		errs = append(errs, fmt.Errorf(
 			"%s: resource name can only contain letters, numbers, "+
-				"dashes, and underscores."+
-				n.ResourceName))
+				"dashes, and underscores.", n.ResourceName))
 	}
 
-	if len(warns) == 0 && len(errs) == 0 {
+	if (len(warns) == 0 || n.IgnoreWarnings) && len(errs) == 0 {
 		return nil, nil
 	}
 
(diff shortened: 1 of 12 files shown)


phinze:
Doozy for ya @jbardin 😄

phinze:
Okay I think I've addressed the failing tests w/ the added last two bullet points on the list above:

 * wrap unit-tests around `EvalValidateResource`
 * add an `IgnoreWarnings` option to `EvalValidateResource` to prevent
   active warnings from halting execution on the second-pass validation

terraform/eval_validate.go
```diff
@@ -108,11 +108,13 @@ type EvalValidateResource struct {
...
 	ResourceMode config.ResourceMode
+
+	// IgnoreWarnings means that warnings will not be passed through. This allows
+	// "just-in-time" passes of validation to continue execution through warnings.
+	IgnoreWarnings bool
 }
 
 func (n *EvalValidateResource) Eval(ctx EvalContext) (interface{}, error) {
-	// TODO: test
-
```
phinze: One down...

```
❯ git grep -i 'TODO: test' | wc -l
      32
```

32 to go! 😄

terraform/transform_resource.go
```diff
@@ -454,7 +464,16 @@ func (n *graphNodeExpandedResource) managedResourceEvalNodes(resource *Resource,
...
+				// Re-run validation to catch any errors we missed, e.g. type
+				// mismatches on computed values.
+				&EvalValidateResource{
+					Provider:       &provider,
+					Config:         &resourceConfig,
+					ResourceName:   n.Resource.Name,
+					ResourceType:   n.Resource.Type,
+					ResourceMode:   n.Resource.Mode,
+					IgnoreWarnings: true,
+				},
```
phinze: So this eval sequence is the same for `walkDestroy` and `walkApply`... and I have a sneaking suspicion that we should be skipping this node for `walkDestroy`.

No tests have failed, and I'm not quite sure how to construct a failing test to prove my hunch. It would need to have the properties:

 * Resource passes validation during `walkValidate`
 * But fails validation during `walkDestroy`

Since all our cases so far have involved computed values being resolved, I can't think of any real world example that would trigger this.

So given all that, not sure if I want to blindly add it to `walkDestroy`, or blindly _avoid_ adding it to `walkDestroy`. Thoughts welcome!

phinze: Oops this applies but to the lower one.

jbardin:
LGTM per our conversation.

PR #5838, opened 2016-03-24
provider/aws: Guard against empty responses from Lambda Permissions

catsby:
Fixes #5536 by guarding against a nil response (or nil `Statement`) from `lambda.AddPermission`. It's not clear if just the `Statement` is nil, or the entire response is nil, so we guard against both

PR #7448, opened 2016-07-01
Timing out is not success

dtolnay:
Practically every caller of `Retry` depends on this. Anything that looks like:

```go
var out *Type
err = Retry(10*time.Second, func() *RetryError {
    out, err = f()
    return // ...
})
if err != nil {
    return // ...
}
use(out.Field)
```

Without this change, the last line will panic if `f` hangs.

### DIFF
--- a/helper/resource/wait.go
+++ b/helper/resource/wait.go
@@ -36,12 +36,20 @@ func Retry(timeout time.Duration, f RetryFunc) error {
 		},
 	}
 
-	c.WaitForState()
+	_, waitErr := c.WaitForState()
 
 	// Need to acquire the lock here to be able to avoid race using resultErr as
 	// the return value
 	resultErrMu.Lock()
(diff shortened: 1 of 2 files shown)


dtolnay:
I suspect this is why silly things like #5838 were necessary. Those sorts of checks are all over the place. AWS wouldn't actually return nil, nil.

phinze:
Wow thanks @dtolnay - this is a great catch!

I'm surprised we've managed to sustain a bug like this for so long, but - as your linked issues suggest -we probably papered over it by generally assuming nil panics were caused by funky API issues and adding extra nil checking.

phinze:
This LGTM, but I'd like to get one more set of eyes on it before we merge. Perhaps a @jbardin?

jbardin:
LGTM2, thanks!

PR #7454, opened 2016-07-01
Ignore AWS internal tags

andskli:
See the following issue: #4513.

AWS creates internal tags for some resources (for example, resources that are created by CloudFormation), which are immutable. Thus, terraform are unable to modify these tags and detects them as stuff that's supposed to be changed. They're all prefixed with `aws:`.

This proposed solution ignores the following two regex pattern: `^aws:*`.

Feedback very much welcome!

### DIFF
--- a/builtin/providers/aws/autoscaling_tags.go
+++ b/builtin/providers/aws/autoscaling_tags.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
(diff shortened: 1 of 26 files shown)


jen20:
Hi @andskli! This looks reasonable to me, though I wonder whether the provider should also allow opting out of this? cc @catsby.

stack72:
Hi @andskli 

Thanks for the PR here. We just discussed this internally. For this to be merged, we would need the PR to be extended to include all of the tag* files (e.g. tagsRDS). This would mean that we are able to do this for the entire provider

Thanks for the work here

Paul

andskli:
Blindly extended to all tag* files and their corresponding tests!

Phlamethrower:
@andskli - It looks like you've missed the autoscaling code (autoscaling_tags.go).

Thanks for the changes so far!

andskli:
@Phlamethrower - thanks for noticing, and sorry for the late reply as I've been on vacation. However, the latest push should fix the missing piece. :-)

bassrock:
Is there a way to monkey patch this in?

nickm4062:
bump

u2mejc:
@andskli looks like the only conflict is a couple of new imports, would you be able to rebase master on your branch and accept the new imports? Alternatively, if you'd like to grant me write access on your repo, I'll be happy to help shepherd it into master.

Hi Paul! This is a great feature improvement as it allows the importation of existing resources originally created with AWS tools. The issue currently is exasperated because you can't add these tags in Terraform without running into a parsing error. Think we could get this fast-tracked into the next release?

andskli:
Conflict solved.

u2mejc:
Awesome! @stack72, anything else you'd like to see before this is merged?

u2mejc:
Hi Paul, happy friday! 🍻  Anything else you'd like to see to have this merged? Test coverage looking good to you?

cc: @mitchellh

stack72:
HI @andskli 

Thanks for the work here - this is looking good now! Sorry it has taken so long to get back to you on 

Paul

stack72:
This was manually merged in 73d68f7bcc0c201ea2a593dd70692c6263745132

PR #7459, opened 2016-07-01
helper/resource: Consolidate unit test override

phinze:
I noticed we had two mechanisms for unit test override. One that dropped
a sentinel into the env var, and another with a struct member on
TestCase. This consolidates the two, using the cleaner struct member
internal mechanism and the nicer `resource.UnitTest()` entry point.


jbardin:
LGTM

PR #7461, opened 2016-07-01
Remove unused variables

dtolnay:
For some reason these prevent [errcheck](https://github.com/kisielk/errcheck) from running successfully.

Side note: errcheck reports 2590 unchecked errors (excluding tests and vendored dependencies). It would have prevented the longstanding bug in #7448.


phinze:
Solid, thanks!

I use `errcheck` locally in my editor, but never connected the mental pathway that it should just go into CI. 🙈 

I'll look into seeing if we can drop it in on an opt-in basis so we don't have to bite off all those >2500 errors at once.

dtolnay:
I filed #8537 to follow up on integrating errcheck into CI.

PR #7462, opened 2016-07-01
Return nonnegative hash if int is 32 bits

dtolnay:
`hashcode.String("2338615298")` was negative on 32-bit.


jbardin:
Good catch, thanks!
@phinze, LGTM

PR #7464, opened 2016-07-01
core: Return an error when there's no remote state

### LINKED ISSUE #7455: Panic creating/reading remote state
### Terraform Version
`Terraform v0.7.0-dev`

### Affected Resource(s)

- terraform_remote_state using s3 bucket.

```hcl
resource "aws_s3_bucket" "state_bucket" {
  bucket = "terraform-state-abcdefg123"
  acl = "private"
}

resource "terraform_remote_state" "on_s3_by_environment_and_prefix" {
  backend = "s3"

  config {
    bucket = "${aws_s3_bucket.state_bucket.bucket}"
    key    = "terraform.tfstate" # "${var.prefix}something.tfstate"
  }
}
```

### Debug Output
I can privately share it or anonymized it if needed. 

### Panic Output
I can privately share it or anonymized it if needed. 

### Expected Behavior

- Create a bucket and upload the sate to it.

### Actual Behavior

- The bucket is created but the state file isn't.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids
We are using roles for the accounts.

jbardin:
When refreshing remote state, indicate when no state file was found with
an ErrRemoteStateNotFound error. This prevents us from inadvertantly
getting a nil state into a terraform.State where we assume there's
always a root module.

Fixes #7455

### DIFF
--- a/state/remote/remote_test.go
+++ b/state/remote/remote_test.go
@@ -41,3 +41,21 @@ func testClient(t *testing.T, c Client) {
 		t.Fatalf("bad: %#v", p)
 	}
 }
+
+func TestRemoteClient_noPayload(t *testing.T) {
+	s := &State{
+		Client: nilClient{},
+	}
+	if err := s.RefreshState(); err != ErrRemoteStateNotFound {
+		t.Fatal("expected ErrRemoteStateNotFound, got", err)
+	}
+}
+
+// nilClient returns nil for everything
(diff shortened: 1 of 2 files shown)


phinze:
Cool so this addresses the `terraform_remote_state` issue but also will toss a sane error if anything funky happens w/ the client. 👍 

LGTM

PR #7469, opened 2016-07-02
Convert tls_cert_request to be a data source

apparentlymart:
This resource (unlike the others in this provider) isn't stateful, so it is a good candidate to be a data source.

The old resource form is preserved via the standard shim in helper/schema, which will generate a deprecation warning but will still allow the resource to be used.


jen20:
LGTM - we have an unrelated failing TLS acceptance test though: `TestLocallySignedCert`.

PR #7487, opened 2016-07-05
aws: AMI data source docs fixed

radeksimko:
The extra spacing was causing the following formatting:

screen shot 2016-07-05 at 10 38 41


stack72:
LGTM!

PR #7493, opened 2016-07-05
terraform: allow literal maps to be passed to modules

### LINKED ISSUE #7140: Unable to pass map literals as module parameters
### Terraform Version

master / 2127466280172a7cd0aa1d863cb932aa0107c986 

### Affected Resource(s)

core

```tf
# main.tf
module "mod" {
  source   = "./mod"
  mapparam = { }
}
```

```tf
# mod/main.tf
variable "mapparam" { type = "map" }
```

### Steps to Reproduce

```
terraform get
terraform apply
```

### Expected Behavior

No error message.

### Actual Behavior

```

* variable mapparam in module mod should be type map, got []interface {}
```

phinze:
Passing a literal map to a module looks like this in HCL:

    module "foo" {
      source = "./foo"
      somemap {
        somekey = "somevalue"
      }
    }

The HCL parser always wraps an extra list around the map, so we need to
remove that extra list wrapper when the parameter is indeed of type "map".

Fixes #7140

### DIFF
--- a/terraform/eval_variable.go
+++ b/terraform/eval_variable.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	"fmt"
+	"log"
 	"reflect"
 	"strings"
 
@@ -21,10 +22,6 @@ import (
 //       declared
 //     - the path to the module (so we know which part of the tree to
 //       compare the values against).
-//
-// Currently since the type system is simple, we currently do not make
-// use of the values since it is only valid to pass string values. The
-// structure is in place for extension of the type system, however.
 type EvalTypeCheckVariable struct {
 	Variables  map[string]interface{}
 	ModulePath []string
@@ -50,10 +47,6 @@ func (n *EvalTypeCheckVariable) Eval(ctx EvalContext) (interface{}, error) {
 	}
 
 	for name, declaredType := range prototypes {
-		// This is only necessary when we _actually_ check. It is left as a reminder
-		// that at the current time we are dealing with a type system consisting only
-		// of strings and maps - where the only valid inter-module variable type is
-		// string.
 		proposedValue, ok := n.Variables[name]
 		if !ok {
 			// This means the default value should be used as no overriding value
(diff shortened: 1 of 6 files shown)


terraform/eval_variable.go
```diff
@@ -21,10 +21,6 @@ import (
 //       declared
 //     - the path to the module (so we know which part of the tree to
 //       compare the values against).
-//
-// Currently since the type system is simple, we currently do not make
-// use of the values since it is only valid to pass string values. The
-// structure is in place for extension of the type system, however.
```
phinze: This bit of the comment was no longer valid by the time I got here.

terraform/eval_variable.go
```diff
@@ -81,6 +77,12 @@ func (n *EvalTypeCheckVariable) Eval(ctx EvalContext) (interface{}, error) {
 			case map[string]interface{}:
 				continue
 			default:
+				if list, ok := proposedValue.([]interface{}); ok && len(list) == 1 {
+					if m, ok := list[0].(map[string]interface{}); ok {
+						n.Variables[name] = m
+						continue
+					}
+				}
```
phinze: I don't like doing the coercion mid-type-check. I'm planning to pull this out into its own `EvalNode`.

Just wanted to get visibility on the strategy before i go further

jbardin:
LGTM (emphasis on the TM here, I'm not sure if there are other implications to consider)

phinze:
👌  will wait for a @jen20 double check here

terraform/eval_variable.go
```diff
@@ -50,10 +47,6 @@ func (n *EvalTypeCheckVariable) Eval(ctx EvalContext) (interface{}, error) {
 	}
 
 	for name, declaredType := range prototypes {
-		// This is only necessary when we _actually_ check. It is left as a reminder
-		// that at the current time we are dealing with a type system consisting only
-		// of strings and maps - where the only valid inter-module variable type is
-		// string.
```
jen20: :+1: on removing this comment (added for a late-version 0.6.x)

jen20:
Yup, this LGTM too.

PR #7496, opened 2016-07-05
Fix nested module "unknown variable" during dest

jbardin:
During a destroy with nested modules, access a variable between them
causes an "unknown variable accessed" during destroy.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -4704,3 +4704,83 @@ aws_instance.foo:
 		t.Fatalf("bad: \n%s", actual)
 	}
 }
+
+// https://github.com/hashicorp/terraform/issues/7378
+func TestContext2Apply_destroyNestedModuleWithAttrsReferencingResource(t *testing.T) {
+	m := testModule(t, "apply-destroy-nested-module-with-attrs")
+	p := testProvider("null")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+
+	var state *State
+	var err error
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Module: m,
+			Providers: map[string]ResourceProviderFactory{
+				"null": testProviderFuncFixed(p),
+			},
+		})
+
+		// First plan and apply a create operation
+		if _, err := ctx.Plan(); err != nil {
+			t.Fatalf("plan err: %s", err)
+		}
+
+		state, err = ctx.Apply()
+		if err != nil {
+			t.Fatalf("apply err: %s", err)
+		}
+	}
+
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Destroy: true,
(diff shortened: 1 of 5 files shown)


terraform/test-fixtures/apply-destroy-nested-module-with-attrs/middle/middle.tf
```diff
@@ -0,0 +1,8 @@
+variable param {}
```
phinze: technically valid to leave the var name bare, but conventionally we always quote it -> `variable "param" {}`

terraform/test-fixtures/apply-destroy-nested-module-with-attrs/main.tf
```diff
@@ -0,0 +1 @@
+
```
phinze: This file can be dropped, nothing requires a `main.tf` to exist.

terraform/context_apply_test.go
```diff
@@ -4704,3 +4704,84 @@ aws_instance.foo:
...
+		ctx := testContext2(t, &ContextOpts{
+			Destroy: true,
+			Module:  m,
+			State:   state,
+			Providers: map[string]ResourceProviderFactory{
+				"null": testProviderFuncFixed(p),
+			},
+		})
+
+		// First plan and apply a create operation
```
phinze: This comment no longer accurate - this is the destroy now.

jbardin: I think this was for another comment (I've since removed)? This block is doing a Plan+Apply.

phinze: Oh weird - yeah last time I made this comment it was on the below block. Weird diff hunk issue maybe.

terraform/graph_config_node_variable.go
```diff
@@ -130,17 +130,22 @@ func (n *GraphNodeConfigVariable) hasDestroyEdgeInPath(opts *NoopOpts, vertex da
 	if vertex == nil {
 		vertex = opts.Vertex
 	}
+
+	var root graphNodeRoot
```
phinze: I see how this works, but it feels like a bit of an odd way to test if a node is the root node. Maybe worth wrapping this up in a `IsGraphNodeRoot()` predicate stored near the `graphNodeRoot` impl?

Or here's another idea - IMHO this impl is a bit less confusing than the local var / `==` test:

```go
_, isRoot := v.(graphNodeRoot)
if isRoot || cv.DestroyEdgeInclude(v) { // ...
```

What do you think?

jbardin: Yeah, checking that c is a root node is a little cleaner. Unless we had a single global rootNode value to compare to, I agree the struct comparison felt weird. I'll update

phinze:
LGTM

jbardin:
This fixes #7378

PR #7510, opened 2016-07-06
terraform: add test helper for inline config loading

phinze:
In scenarios with a lot of small configs, it's tedious to fan out actual
dir trees in a test-fixtures dir. It also spreads out the context of the
test - requiring the reader fetch a bunch of scattered 3 line files in
order to understand what is being tested.

Our config loading code still only reads from disk, but in
the `helper/resource` acc test framework we work around this by writing
inline config to temp files and loading it from there. This helper is
based on that strategy.

Eventually it'd be great to be able to build up a `module.Tree` from
config directly, but this gets us the functionality today.

Example Usage:

    testModuleInline(t, map[string]string{
      "top.tf": `
        module "middle" {
          source = "./middle"
        }
      `,
      "middle/mid.tf": `
        module "bottom" {
          source = "./bottom"
          amap {
            foo = "bar"
          }
        }
      `,
      "middle/bottom/bot.tf": `
        variable "amap" {
          type = "map"
        }
      `,
    }),

### DIFF
--- a/terraform/terraform_test.go
+++ b/terraform/terraform_test.go
@@ -2,6 +2,7 @@ package terraform
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


phinze:
cc @jen20 @jbardin

terraform/terraform_test.go
```diff
@@ -60,6 +61,54 @@ func testModule(t *testing.T, name string) *module.Tree {
...
 }
 
+// testModuleInline takes a map of path -> config strings and yields a config
+// structure with those files loaded from disk
+func testModuleInline(t *testing.T, config map[string]string) *module.Tree {
+	cfgPath, err := ioutil.TempDir("", "tf-test")
+	if err != nil {
+		t.Errorf("Error creating temporary directory for config: %s", err)
+	}
+	// defer os.RemoveAll(cfgPath)
```
jbardin: can we clean this up, or do the files need to exist later for the `*module.Tree`?

phinze: Oop no commented while I was verifying the behavior, fixing.

jbardin:
I like it! LGTM

Issue #6934, opened 2016-05-30
aws_s3_bucket_notification multiple notifications for bucket


fuCtor:
Try configure multiple notification for bucket. But provider override it by last definition.
In state file I have:

```
  "aws_s3_bucket_notification.bucket_notify.0": {
      "type": "aws_s3_bucket_notification",
      "depends_on": [
          "aws_s3_bucket.general",
          "aws_sns_topic.exchange"
      ],
      "primary": {
          "id": "test-bucket",
          "attributes": {
              "bucket": "test-bucket",
              "id": "test-bucket",
              "lambda_function.#": "0",
              "queue.#": "0",
              "topic.#": "1",
              "topic.0.events.#": "1",
              "topic.0.events.3356830603": "s3:ObjectCreated:*",
              "topic.0.filter_prefix": "chunks",
              "topic.0.filter_suffix": "",
              "topic.0.id": "s3-chunks-notify",
              "topic.0.topic_arn": "arn:aws:sns:XXX:s3-chunks-create-topic"
          }
      }
  },
  "aws_s3_bucket_notification.bucket_notify.1": {
      "type": "aws_s3_bucket_notification",
      "depends_on": [
          "aws_s3_bucket.general",
          "aws_sns_topic.exchange"
      ],
      "primary": {
          "id": "test-bucket",
          "attributes": {
              "bucket": "test-bucket",
              "id": "test-bucket",
              "lambda_function.#": "0",
              "queue.#": "0",
              "topic.#": "1",
              "topic.0.events.#": "1",
              "topic.0.events.3356830603": "s3:ObjectCreated:*",
              "topic.0.filter_prefix": "trace",
              "topic.0.filter_suffix": "",
              "topic.0.id": "s3-trace-notify",
              "topic.0.topic_arn": "arn:aws:sns:XXX:s3-trace-create-topic"
          }
      }
  },
```

Config file contains:

```
resource "aws_s3_bucket_notification" "bucket_notify" {
  count = "${var.topic_count}"
  bucket = "${aws_s3_bucket.general.id}"
  topic {
    id = "s3-${lookup(var.topics, count.index)}-notify"
    topic_arn = "${element(aws_sns_topic.exchange.*.arn, count.index)}"
    events = ["s3:ObjectCreated:*"]
    filter_prefix = "${lookup(var.topics, count.index)}"
  }
}
```

As you can see, this resource have same ID.

When I try add ID key for resource, receive error:

```
    Errors:
[...]

PR #7518, opened 2016-07-06
Update S3 notification documentation

bschwind:
When adding multiple notifications from one S3 bucket to one SQS queue, it wasn't immediately intuitive how to do this. At first I created two `aws_s3_bucket_notification` configs and it seemed to work fine, however the config for one event will overwrite the other. In order to have multiple events, you can defined the `queue` key twice, or use an array if you're working with the JSON syntax. I tried to make this more clear in the documentation.


bschwind:
Related with #6934

phinze:
LGTM - thanks for this @bschwind! 👍

Issue #6870, opened 2016-05-25
10 Second Polling on Elastic Beanstalk causes API limit issues.


tecnobrat:
In 0.6.16 when you update an elastic beanstalk environment it polls every 10 seconds waiting for the environment to finish updating.  If you have many environments that need updating at the same time, it ends up hitting API limits.

Can we get this as a configurable value?

dharrisio:
I think that this would be a good thing to add as well. Any thoughts on the name for that attribute? Maybe something like `wait_for_ready_poll_frequency`?

tecnobrat:
Sounds good .. however I wonder if there is a more global setting, this could be needed for any number of resources. Perhaps someone like @catsby could comment on whether it should be a globally name variable to deal with the polling interval like in here.

catsby:
Hey Friends – I'm sorry to hear that you're hitting API limits here. Can you give me an idea how many Beanstalk Environments you're managing, and what kind of updates you're doing? I'm trying to see this in person with 20 environments but I'm not yet hitting API limits. Of course, I'm only testing Environments and working in a test environment, so this isn't exactly your situation(s).

Regardless, there are some ideas here. 

The easiest and most isolated is to increase the `MinTimeout` in both the `create` and `update` methods:

- https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go#L318-L324

It's currently set at 10; by bumping it to 20 we can cut the API calls in half _just for Beanstalk Environments_. 

The suggestion of a configuration for the polling timeouts is a request we've received before. We faced the same dilemma as you touched on here; is this a global timeout configuration (the `aws` provider), per resource _type_ (`aws_elasticbeanstalk_environment`), or per resource (`aws_elasticbeanstalk_environment.env-1`)? In your case, maybe it's just the later, but that requires implementation in each resource. In other cases, our users seem to want to change this in one place for their large infrastructure. 

So, which option do you think would suit _you_ best? 
I opened #7523 as a band-aid for ElasticBeanstalk specifically, but as I mentioned I wasn't able to hit this limit myself so perhaps not.

tecnobrat:
Hey @catsby we have about 25 environments.  The limit is hit more often when we do a mass update to all of them, for instance updating to the latest `solution_stack_name`.

[...]

PR #7523, opened 2016-07-06
provider/aws: Beanstalk environments, bump the minimum timeout between API calls

catsby:
This is a band-aid to https://github.com/hashicorp/terraform/issues/6870. Beanstalk environments can take a long time to setup, and if you are managing many of them you can hit API rate limits for your account. 

A more permanent solution may be to expose a knob to set a customizable timeout / poll interval, but as I mention in #6870 the design of such a knob hasn't been planed out yet.

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
@@ -137,7 +137,6 @@ func resourceAwsElasticBeanstalkEnvironment() *schema.Resource {
 			"poll_interval": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "10s",
 				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 					value := v.(string)
 					duration, err := time.ParseDuration(value)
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -249,7 +249,7 @@ func resourceAwsElasticBeanstalkEnvironmentCreate(d *schema.ResourceData, meta i
 		Refresh:    environmentStateRefreshFunc(conn, d.Id()),
 		Timeout:    waitForReadyTimeOut,
 		Delay:      10 * time.Second,
```
tecnobrat: Isn't it the delay you want to update?

catsby: Delay is the amount of time before we start. MinTimeout is the minimum amount of time between refreshes:

```go
Delay          time.Duration    // Wait this time before starting checks
MinTimeout     time.Duration    // Smallest time to wait before refreshes
```

I'll be refactoring this PR to expose at least `MinTimeout`, stay tuned!

builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -134,6 +134,24 @@ func resourceAwsElasticBeanstalkEnvironment() *schema.Resource {
...
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					duration, err := time.ParseDuration(value)
+					if err != nil {
+						errors = append(errors, fmt.Errorf(
+							"%q cannot be parsed as a duration: %s", k, err))
+					}
+					if duration < 10*time.Second || duration > 60*time.Second {
+						errors = append(errors, fmt.Errorf(
+							"%q must be between 10s and 60s", k))
```
catsby: `10s` is our current minimum. `60s` seemed like a reasonable max, but we can tune it higher if need be

tecnobrat: `60s` seems like a reasonable max, I'd agree. You don't want atlas jobs taking forever because of someone's value here.  Its a fine line.

tecnobrat: Also I like `poll_interval` as the variable, seems to be generic that could be added in another providers / resources in the future.

phinze:
I'm not sure if `MinTimeout` does what we want here:

https://github.com/hashicorp/terraform/blob/2a45b1fb2520554a706527ba9329c16fdd17ff88/helper/resource/state.go#L77-L83

Looks like it's capped at 10s based on that conditional.

That's all based on doing exponential backoff. Maybe we need to plumb through a fresh option for a static poll interval if the user requests it?

catsby:
Paul and I discussed this and found that `MinTimeout` was in fact "working" as I expected it, however, we discovered that it would only do so for about 10 minutes, until the calculated back off time exceeded the 60s, in which case the polling would return to every 10 seconds 😵 

I'm going to take another pass and try adding a separate config option that's preferable to the min timeout

catsby:
@phinze I refactored this to use `poll_interval` as a separate thing in `WaitForState`.

phinze:
LGTM

Issue #6222, opened 2016-04-18
provider-aws: Intermittent Race Condition with Ec2 Metadata service/credentials


jrnt30:
### Issue:

While `plan` or `apply`ing our infrastructure when using the EC2 Metadata Service for TF AWS credentials we are seeing intermittent issues with our larger TF stack where we receive varying numbers of `No valid credential sources found for AWS Provider.` error messages.  

A few things to note: 
- Doesn't ever happen on some of our smaller TF stacks 
- If we pass the credentials via environment variables explicitly it works every time 
- Configuring the credentials file explicitly on the node it will work fine
- If we use the IAM Role/Metadata provided credential provider it works intermittently (maybe 20% of the time) 
- If we limit the parallelism with the IAM Role credential provider it works most of the time as well when reduced down to 3.  5 and 7 still fail relatively frequently 
- Get the "No valid credential sources found for AWS Provider." error on both Plan and Execute 
- Increasing log level to "TRACE" slows things down enough for us to prevent the issue from occurring :) 

My guess is that there is an API rate limiting occurring when using the IAM Role provided credentials with the default parallelism, curious if others have run into similar issues. 
### Terraform Version:

terraform-0.6.14
### Affected:
- AWS Provider configuration w/ IAM Roles/Metadata service
### Debug Output (Stripped to just relevant counts):

```
Count for [./credentials_file_2016-04-18-171349-BROKE - Building AWS auth structure]  24
Count for [./credentials_file_2016-04-18-171349-BROKE - EC2 Metadata service found, adding EC2 Role Credential Provider]  24
Count for [./credentials_file_2016-04-18-171349-BROKE - EC2 Metadata service not found, not adding EC2 Role Credential Provider]  0
```
### Panic Output:

No panic output, however do get: 

```
* No valid credential sources found for AWS Provider.
  Please see https://terraform.io/docs/providers/aws/index.html for more information on
  providing credentials for the AWS Provider
```
### Expected Behavior

Reduce contention on Metadata service calls or handle/retry responses

jrnt30:
@phinze Per our conversation in chats today. Thanks for your help as well

jrnt30:
@phinze I was able to confirm that the metadata service does start sending 429s when it's being hit heavily.  In my concurrency test it seems to break down after 25 concurrent connections are made.

Solisol:
Is there a fix for this coming? *friendlybump*

dcarley:
[...]

PR #7524, opened 2016-07-06
[WIP] provider/aws: Cache client setup

phinze:
This prevents a the bunches of API calls that emanate from the
`provider.Config` calls as we walk the graph.

Historically, the `provider.Config` code path has not involved any
remote API calls. Partially because of this - core has not been stingy
with the number of times it asks the provider to configure itself.

As of recently, the AWS provider started performing several api calls:
- `iam.GetUser`
- `sts.GetCallerIdentity`

It turns out core calls this code path 6-9x per module, which really
adds up! Here, we introduce a very naive caching mechanism that relies
on the fact that only one provider process will be launched.

Taking samples from the `aws_vpc` basic acceptance test, this reduces
the number of config-related API calls down to one.

### DIFF
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
+	"github.com/mitchellh/hashstructure"
 )
 
 // Provider returns a terraform.ResourceProvider.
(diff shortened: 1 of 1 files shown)


phinze:
This works well in initial testing, but not super sure about all the implications of a global-var-map cache. Could pretty easily wrap a mutex around the sucker if that would be helpful.

/cc @jbardin @jen20 @catsby

phinze:
While not a root cause fix, I think this will likely address the problems being reported in #6222.

builtin/providers/aws/provider.go
```diff
@@ -337,6 +348,14 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		Insecure:         d.Get("insecure").(bool),
 	}
 
+	hash, err := hashstructure.Hash(config, nil)
+	if err != nil {
+		return nil, err
+	}
+	if client, ok := cachedClients[hash]; ok {
+		return client, nil
+	}
```
phinze: @jen20 in discussing this with you in person, you suggested that this mapping was unnecessary, because each distinct AWS config (one per alias) will be in its own process already. Am I remembering that correctly?

jrnt30:
One thing to note with the IAM roles is that they do rotate credentials under the covers automatically.  It may be necessary for AWS at least to take this into consideration or provide a way of evicting the cache should the credentials fail. 

I could see this occurring in a situation where you were right on the cusp of a credential change or a very long running TF deploy.

phinze:
I still think it'd be a good idea to pursue this as an optimization, but this PR as old and I've no plans to pick it up anytime soon, so closing this to clean up.

Issue #7202, opened 2016-06-16
vSphere Virtual Machine Resource - Cannot specify type of scsi controller


dkalleg:
Hi there,

Today we can only specify the controller types 'ide' and 'scsi', but govmomi supports 4 types of scsi controllers:
LSI Logic, BusLogic, VMWare Paravirtual, and LSI Logic SAS

[https://github.com/hashicorp/terraform/blob/master/vendor/github.com/vmware/govmomi/object/virtual_device_list.go#sourcegraph&def=GoPackage/github.com/vmware/govmomi/object/-/SCSIControllerTypes&L43-56](url)

Today, Terraform defaults to LSI Logic when the user specifies SCSI.  Terraform should support each.

Regarding implementation, I believe a `scsi_type` attribute should be a vm level attr not a disk level attr so that each scsi device is uniformly connected to the same type of controller.

chrislovecnm:
`scsi_type` at vm level.  Does that mean that we only can have one type per vm?

dkalleg:
Yes.  My understanding is that there is no use case for having multiple different types of scsi controllers on a vm.

chrislovecnm:
I have done it before :P Some disks paravirtual some lsi

chrislovecnm:
Maybe I should not have

mixacha:
I believe Microsoft recommends Lsi controllers when doing MSC in virtual environments. IT would be great if we could do at disk level, as I'm sure that as Terraform gains momentum in the enterprises this will be a good "selling" point.

sputnik13:
@mixacha I'm sure that's LSI SAS not LSI Parallel, we tested windows server 2012 r2 on vsphere 5.5 today and it (windows) didn't even recognize an LSI Parallel controller out of the box while LSI SAS was recognized and worked right away.  The issue we're seeing is that LSI Parallel was being selected as default.

mixacha:
@sputnik13  Yeah, thanks for additional clarification.

stack72:
Closed via #7525

stack72:
Closed via #7525

PR #7525, opened 2016-07-07
Additional SCSI controller types support

dkalleg:
This allows the user to specify new controller types.  Before when
specifying 'scsi', govmomi defaults to lsilogic-parallel. This patch
allows the user to now specify 'scsi-lsi-parallel', 'scsi-buslogic',
scsi-paravirtual', and 'scsi-lsi-sas'. Resolves issue
https://github.com/hashicorp/terraform/issues/7202

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -26,6 +26,15 @@ var DefaultDNSServers = []string{
 	"8.8.4.4",
 }
 
+var DiskControllerTypes = []string{
+	"scsi",
+	"scsi-lsi-parallel",
+	"scsi-buslogic",
+	"scsi-paravirtual",
+	"scsi-lsi-sas",
+	"ide",
+}
+
 type networkInterface struct {
 	deviceName       string
 	label            string
@@ -421,9 +430,15 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Default:  "scsi",
 							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 								value := v.(string)
-								if value != "scsi" && value != "ide" {
+								found := false
+								for _, t := range DiskControllerTypes {
+									if t == value {
+										found = true
+									}
+								}
+								if !found {
 									errors = append(errors, fmt.Errorf(
-										"only 'scsi' and 'ide' are supported values for 'controller_type'"))
+										"Supported values for 'controller_type' are %v", strings.Join(DiskControllerTypes, ", ")))
 								}
 								return
 							},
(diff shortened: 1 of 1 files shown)


sputnik13:
It would be great if this could be merged for 0.7.0 or a point release soon thereafter.  We cannot use terraform to create Windows VMs on vSphere at this time because LSI Parallel (which is the only scsi controller type supported by terraform at this time) is not supported by Windows 2012 R2 out of the box.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -421,9 +421,16 @@ func resourceVSphereVirtualMachine() *schema.Resource {
 							Default:  "scsi",
 							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 								value := v.(string)
-								if value != "scsi" && value != "ide" {
```
stack72: I think we need to do something about this ValidateFunc. We need to refactor it to something like declaring a []string of values and then checking that the value is in the slice. The cyclomatic complexity of if && && && etc doesn't really read the best :)

dkalleg: Good point.  I added a []string with each controller type and the validate func now loops over it.  I wish golang had the python-ism of "if x in y" instead of having to loop through.  Apparently that is a go-ism. See what you think about this patch.

stack72: Hey @dkalleg 

Just a FYI, I wrote a validateFunc that looks like this:

```
func validateArmStorageBlobType(v interface{}, k string) (ws []string, errors []error) {
	value := strings.ToLower(v.(string))
	validTypes := map[string]struct{}{
		"blob": struct{}{},
		"page": struct{}{},
	}

	if _, ok := validTypes[value]; !ok {
		errors = append(errors, fmt.Errorf("Blob type %q is invalid, must be %q or %q", value, "blob", "page"))
	}
	return
}
```

Notice the use of map and not having to iterate over it ;)

P.

dkalleg: I like not having to iterate! To do this wouldn't I have to redeclare the set of controller types inside of the validateFunc?  My mentality was to have them defined just once, so if a new controller type is supported later, it would only have to be added in one place.  

Is this about what you were thinking:
```
ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
	value := v.(string)
	validTypes := map[string]struct{}{
		"scsi": struct{}{},
		"scsi-lsi-parallel": struct{}{},
		"scsi-buslogic": struct{}{},
		"scsi-paravirtual": struct{}{},
		"scsi-lsi-sas": struct{}{},
		"ide": struct{}{},
	}
	if _, ok := validTypes[value]; !ok {
		errors = append(errors, fmt.Errorf("Supported values for 'controller_type are %v", strings.Join(DiskControllerTypes, ", ")))
	}
```

Do you find it better to redeclare the list to prevent an iteration here?  I'm on the fence so I'll let you decide.

stack72: This looks good as is - i'm just nit-picking :)

sputnik13: bit of a moot point since it's merged already, but...  the use of a map may improve lookup performance, but for a small set like this it's probably negligible and it wouldn't matter much unless the operation was frequent.  Also, there's value in having DiskControllerTypes as an exported list for users of the module.  Turning the list in to a map in that case would be detrimental as DiskControllerTypes would have to be iterated as a map given there's no key extraction facility for maps in go.

PR #7526, opened 2016-07-07
Makefile: Support `make test` if Terraform itself is vendored

jwbowler:
Just in case this is relevant to anyone else out there.

This assumes, of course, that nobody has Terraform cloned into a directory not named `terraform`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-TEST?=$$(go list ./... | grep -v '/vendor/' | grep -v '/builtin/bins/')
+TEST?=$$(go list ./... | grep -v '/terraform/vendor/' | grep -v '/builtin/bins/')
 VETARGS?=-all
 GOFMT_FILES?=$$(find . -name '*.go' | grep -v vendor)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,7 +29,8 @@ core-dev: generate
 
 # Shorthand for quickly testing the core of Terraform (i.e. "not providers")
 core-test: generate
-	@echo "Testing core packages..." && go test -tags 'core' $(TESTARGS) $(shell go list ./... | grep -v -E 'builtin|vendor')
+	@echo "Testing core packages..." && \
+		go test -tags 'core' $(TESTARGS) $(shell go list ./... | grep -v -E 'builtin|terraform/vendor')
```
jen20: Probably working putting a `terraform/` in front of `builtin` here too?

jen20:
Hi @jwbowler! This looks reasonable to me, and won't break anything from our end. I left a comment in line, not sure if it will affect your use cases or not?

jwbowler:
@jen20 No problem, that works for us.

jen20:
LGTM - I'll merge this now. Thanks @jwbowler!

Issue #7146, opened 2016-06-12
Should concat() support lists of complex types?


phinze:
Over in #7145 I've submitted a fix for the reported panic in #7030. This fix is the stop gap measure of properly reporting the lack of support in `concat()` for anything but lists of strings.

We should decide what we want the semantics of `concat()` to be for the following:

 * List of maps
 * List of lists
 * Arbitrarily nested versions of the above

It would also help us decide how to prioritize making this functionality work if we could gather together a few use cases for the various scenarios.

phinze:
@oillio did you have a use case in mind when you ran into the panic in #7030?

oillio:
I don't have a use case that actually works :)

I was trying to find a way to use multiple sources of values to configure an aws_elastic_beanstalk_environment.setting field.  I couldn't find a way to use variable data in the setting, so I don't currently have a use for concat on complex variable types.
I will want to use concat like this if a feature like #7034 is ever implemented.

jen20:
This is fairly incompatible with the type checker in HIL, since `concat` has to explicitly declare the return type (hence the backward compatibility note of no longer being able to use `concat()` with strings). We should probably consider whether to make the type checker smarter or whether to add differently named mechanisms for concatenating other data types.

phinze:
I think the return type for `concat()` will always be a TypeList here - it's just the type of the contents of said list that change.

A shot at target semantics:

 * `concat()` - combine lists, e.g. `concat(["foo", "bar"], ["baz"]) => ["foo, "bar", "baz"])`
 * `merge()` - combine maps e.g. `merge({foo = 1}, {bar = 2}) => {foo = 1, bar = 2}`

tecnobrat:
Looks like this was fixed by the mentioned pr?

phinze:
Indeed, this was taken care of in #7528 by @jbardin

ju2wheels:
@phinze Im not following what the overall decision here was. Is `concat` supposed to work on lists of maps or only lists of strings?

Using 0.9.6, Im trying to do the following:

```
resource "aws_rds_cluster_parameter_group" "my_rcpg" {
  description = "My Cluster - ${var.my_module_instance_name}"
  family      = "${var.my_rds_cluster_parameter_group_family}"
  name        = "mymodule-${var.my_module_instance_name}"
[...]

PR #7528, opened 2016-07-07
core: Add list() interpolation function

### LINKED ISSUE #7460: Allow creating list in interpolations
### Terraform Version

### Affected Resource(s)

Interpolations

N/A
### Debug Output
N/A
### Panic Output
N/A
### Expected Behavior

I have some variables that I'd like to pass into a resource.  Sometimes, some of those variables can be empty string, so I'd like to be able to use compact.  However, I don't have those variables in a list format, so cannot use compact.  It would be nice if there was a varag list() for creating the list so that I could pass it into compact.

I can do something like:

    value = "${compact(split(",", format("%s,%s", var.var1, var.var2)))}"

So the capability is there, but would be much nicer if I could do:

    value = "${compact(list(var.var1, var.var2))}"

### Actual Behavior
N/A
### Steps to Reproduce
N/A
### Important Factoids
N/A
### References
N/A

jen20:
The list() interpolation function provides a way to add support for list literals to HIL without having to invent new syntax for it and modify the HIL parser.

It presents as a function, thus:

    - list() -> []
    - list("a") -> ["a"]
    - list("a", "b") -> ["a", "b"]

Thanks to @wr0ngway for the idea of this approach, fixes #7460.

Assuming this is the route we wish to follow, we should make it work with types other than strings - the initial quick implementation is to get a feel for using a function rather than syntax.

cc @phinze, @mitchellh, @jbardin

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -69,6 +69,7 @@ func Funcs() map[string]ast.Function {
 		"join":         interpolationFuncJoin(),
 		"jsonencode":   interpolationFuncJSONEncode(),
 		"length":       interpolationFuncLength(),
+		"list":         interpolationFuncList(),
 		"lower":        interpolationFuncLower(),
 		"md5":          interpolationFuncMd5(),
 		"uuid":         interpolationFuncUUID(),
@@ -80,6 +81,26 @@ func Funcs() map[string]ast.Function {
 		"split":        interpolationFuncSplit(),
 		"trimspace":    interpolationFuncTrimSpace(),
 		"upper":        interpolationFuncUpper(),
+	}
+}
+
+// interpolationFuncList creates a list from the parameters passed
+// to it.
+func interpolationFuncList() ast.Function {
+	return ast.Function{
+		ArgTypes:     []ast.Type{},
+		ReturnType:   ast.TypeList,
+		Variadic:     true,
+		VariadicType: ast.TypeString,
+		Callback: func(args []interface{}) (interface{}, error) {
+			var outputList []string
+
+			for _, val := range args {
+				outputList = append(outputList, val.(string))
+			}
+
+			return stringSliceToVariableValue(outputList), nil
+		},
 	}
 }
 
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -258,10 +258,8 @@ func interpolationFuncCoalesce() ast.Function {
 	}
 }
 
-// interpolationFuncConcat implements the "concat" function that
-// concatenates multiple strings. This isn't actually necessary anymore
-// since our language supports string concat natively, but for backwards
-// compat we do this.
+// interpolationFuncConcat implements the "concat" function that concatenates
+// multiple lists.
 func interpolationFuncConcat() ast.Function {
 	return ast.Function{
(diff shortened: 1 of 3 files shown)


phinze:
Direction looks good. +1 for this as a stop-gap solution until we can parse non-primitives directly in HIL.

jen20:
I've cleaned this up:

- added a test to verify behaviour if a non-string is passed
- removed the type checking code (which would never be hit) in favour of letting the type checker do its work
- fixed formatting issues

I'll remove the WIP now - cc @phinze.

phinze:
LGTM for strings. Agreed we should make it work for lists and maps as well.

jbardin:
Expanded the list interpolation to accept lists and maps, and the concat function to accept lists of lists and lists of maps. 

This now also fixes #7146

config/interpolate_funcs.go
```diff
@@ -218,44 +258,52 @@ func interpolationFuncCoalesce() ast.Function {
...
-
-				// Otherwise variables
-				if argument, ok := arg.([]ast.Variable); ok {
-					for _, element := range argument {
-						t := element.Type
-						switch t {
+				switch arg := arg.(type) {
+				case string:
+					// we also allow bare strings for backwards compatibility
+					outputList = append(outputList, ast.Variable{Type: ast.TypeString, Value: arg})
```
phinze: Let's nuke the BC here. I believe we've already gone talking about this case like it's gone! 😀

config/interpolate_funcs_test.go
```diff
@@ -5,13 +5,114 @@ import (
...
+				Value: map[string]ast.Variable{
+					"key2": {
+						Type:  ast.TypeString,
+						Value: "baz",
+					},
+				},
+			},
+		},
+	})
+}
```
phinze: 👏

config/interpolate_funcs.go
```diff
@@ -218,44 +258,52 @@ func interpolationFuncCoalesce() ast.Function {
...
 				}
+			}
 
-				return nil, fmt.Errorf("arguments to concat() must be a string or list of strings")
+			// we don't support heterogeneous types, so make sure all types match the first
+			if len(outputList) > 0 {
+				firstType := outputList[0].Type
+				for _, v := range outputList[1:] {
+					if v.Type != firstType {
+						return nil, fmt.Errorf("unexpected type %s in list of type %s", v.Type, firstType)
```
phinze: I believe "printable" is a thing for types to make them more error string friendly.

phinze:
Nice work @jbardin!! A few inline nits, but this looks super solid overall.

config/interpolate_funcs.go
jbardin: I wasn't sure, because we also stringer-ify them.  The Printable() output is all lower case, but there's no grammar-correct way to remove the stutter from that output (not that this is much better :shrug:)

phinze: Ah gotchya - yeah I was really just making sure we are treating this as a user-facing error message and making sure it's reasonably decipherable from out front. I'll defer to whatever you think works best here. 👍

jbardin:
Used Printable, since it will usually make a better user-facing error.
Squashed the minor changes.

phinze:
LGTM

PR #7533, opened 2016-07-07
[docs] Clarify 'join' produces a string

mlosapio:
It's not clear that join produces a string and not a list - after reading all the available functions "concat" produces a list and that is properly noted.


jen20:
Thanks @mlosapio! I'll merge this - I have further clarifications to make regarding which types of lists may be used here which I will address shortly.

PR #7534, opened 2016-07-07
Allow empty names in aws_route53_record

macheins:
Empty names are supported by DNS and AWS, but do not work, because in terraform they result in `.zone` instead of only `zone`.

### DIFF
--- a/builtin/providers/aws/resource_aws_route53_record.go
+++ b/builtin/providers/aws/resource_aws_route53_record.go
@@ -695,7 +695,11 @@ func expandRecordName(name, zone string) string {
 	rn := strings.ToLower(strings.TrimSuffix(name, "."))
 	zone = strings.TrimSuffix(zone, ".")
 	if !strings.HasSuffix(rn, zone) {
-		rn = strings.Join([]string{name, zone}, ".")
+		if len(name) == 0 {
+			rn = zone
+		} else {
+			rn = strings.Join([]string{name, zone}, ".")
+		}
 	}
 	return rn
 }
--- a/builtin/providers/aws/resource_aws_route53_record_test.go
+++ b/builtin/providers/aws/resource_aws_route53_record_test.go
@@ -339,6 +339,23 @@ func TestAccAWSRoute53Record_TypeChange(t *testing.T) {
 	})
 }
 
+func TestAccAWSRoute53Record_empty(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_route53_record.empty",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckRoute53RecordDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccRoute53RecordConfigEmptyName,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRoute53RecordExists("aws_route53_record.empty"),
+				),
+			},
+		},
+	})
+}
+
 func testAccCheckRoute53RecordDestroy(s *terraform.State) error {
 	conn := testAccProvider.Meta().(*AWSClient).r53conn
 	for _, rs := range s.RootModule().Resources {
--- a/builtin/providers/aws/resource_aws_route53_record_test.go
+++ b/builtin/providers/aws/resource_aws_route53_record_test.go
@@ -972,3 +972,17 @@ resource "aws_route53_record" "sample" {
   records = ["127.0.0.1", "8.8.8.8"]
 }
 `
+
+const testAccRoute53RecordConfigEmptyName = `
+resource "aws_route53_zone" "main" {
+	name = "notexample.com"
+}
+
+resource "aws_route53_record" "empty" {
+	zone_id = "${aws_route53_zone.main.zone_id}"
+	name = ""
+	type = "A"
+	ttl = "30"
+	records = ["127.0.0.1"]
+}
+`
--- a/builtin/providers/aws/resource_aws_route53_record_test.go
+++ b/builtin/providers/aws/resource_aws_route53_record_test.go
@@ -992,7 +992,7 @@ resource "aws_route53_record" "sample" {
 
 const testAccRoute53RecordConfigEmptyName = `
 resource "aws_route53_zone" "main" {
-	name = "notexample.com"
+	name = "not.example.com"
 }
 
 resource "aws_route53_record" "empty" {


stack72:
Hi @macheins 

Thanks so much for the PR here! Do you have any code that we can use to turn into a test for this? This would ensure that the functionality works as expected going forward

Thanks

Paul

macheins:
Hey @stack72 

I tried to add some test. But since I am not familiar with go, the testing framework and do not have much spare time right now. So I can't verify if it is correct. Please check it. Thanks.

stack72:
thanks @macheins 

The test code will help us!

P.

stack72:
Hi @macheins 

thanks again for the work here. I just ran the tests and got the following:

```
=== RUN   TestAccAWSRoute53Record_empty
--- FAIL: TestAccAWSRoute53Record_empty (89.17s)
	testing.go:328: Error destroying resource! WARNING: Dangling resources

		Error: Check failed: InvalidInput: FATAL problem: DomainNameEmpty (Domain name is empty) encountered with ''
			status code: 400, request id: a72f5a70-4b64-11e6-9aea-31a4d7f8b68c

		State: <no state>
FAIL
exit status 1
```

I think this means that the AWS API doesn't allow the record name to be empty - would this be correct?

P.

macheins:
Hey @stack72 

I assume the test I wrote is wrong, because DNS and AWS (in UI and CLI) support it an I am already using the patch in production without any flaws.

I double checked it by applying the following Terraform configuration file:
```
resource "aws_route53_zone" "example" {
  name = "tf.example.org"
}

resource "aws_route53_record" "empty" {
  zone_id = "${aws_route53_zone.example.zone_id}"
  name = ""
  type = "A"
  ttl = 300
  records = ["127.0.0.1"]
}
```

`terraform plan` of the above config returns:
```
+ aws_route53_record.empty
    fqdn:               "<computed>"
    records.#:          "1"
    records.3619153832: "127.0.0.1"
    ttl:                "300"
    type:               "A"
    zone_id:            "${aws_route53_zone.example.zone_id}"

+ aws_route53_zone.example
    comment:        "Managed by Terraform"
    name:           "tf.example.org"
    name_servers.#: "<computed>"
    vpc_region:     "<computed>"
    zone_id:        "<computed>"
```

`terraform apply` applies the config without any problems.

Checking again with `aws route53 list-resource-record-sets --hosted-zone-id <INSERTID>` returns:
```json
{
    "ResourceRecordSets": [
        {
            "ResourceRecords": [
                {
                    "Value": "127.0.0.1"
                }
            ], 
            "Type": "A", 
            "Name": "tf.example.org.", 
            "TTL": 300
        }, 
        {
            "ResourceRecords": [
                {
                    "Value": "ns-1737.awsdns-25.co.uk."
                }, 
                {
                    "Value": "ns-1031.awsdns-00.org."
                }, 
                {
                    "Value": "ns-238.awsdns-29.com."
                }, 
                {
                    "Value": "ns-547.awsdns-04.net."
                }
            ], 
            "Type": "NS", 
            "Name": "tf.example.org.", 
            "TTL": 172800
        }, 
        {
            "ResourceRecords": [
                {
                    "Value": "ns-1737.awsdns-25.co.uk. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400"
                }
            ], 
            "Type": "SOA", 
            "Name": "tf.example.org.", 
            "TTL": 900
        }
    ]
}
```

I also checked `terraform destroy` and it works correctly.

So it is definitely supported and the patch is definitely working. Perhaps someone might help with checking the test, which is certainly the cause of error.

macheins:
@stack72 I am not sure if the second-level domain is the cause of error. Changed test to use a third-level domain.

stack72:
Hi @macheins 

Just a FYI, I have worked out what the test issue was. It was happening in this:

```
func testAccCheckRoute53RecordDestroy(s *terraform.State) error {
	conn := testAccProvider.Meta().(*AWSClient).r53conn
	for _, rs := range s.RootModule().Resources {
		if rs.Type != "aws_route53_record" {
			continue
		}

		parts := strings.Split(rs.Primary.ID, "_")
		zone := parts[0]
		name := parts[1]
		rType := parts[2]

		lopts := &route53.ListResourceRecordSetsInput{
			HostedZoneId:    aws.String(cleanZoneID(zone)),
			StartRecordName: aws.String(en),
			StartRecordType: aws.String(rType),
		}

		resp, err := conn.ListResourceRecordSets(lopts)
		if err != nil {
			if awsErr, ok := err.(awserr.Error); ok {
				// if NoSuchHostedZone, then all the things are destroyed
				if awsErr.Code() == "NoSuchHostedZone" {
					return nil
				}
			}
			return err
		}
		if len(resp.ResourceRecordSets) == 0 {
			return nil
		}
		rec := resp.ResourceRecordSets[0]
		if FQDN(*rec.Name) == FQDN(name) && *rec.Type == rType {
			return fmt.Errorf("Record still exists: %#v", rec)
		}
	}
	return nil
}
```

it was missing: `en := expandRecordName(name, "notexample.com")`

P.

stack72:
The test results now report as:

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSRoute53Record_'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSRoute53Record_ -timeout 120m
=== RUN   TestAccAWSRoute53Record_basic
--- PASS: TestAccAWSRoute53Record_basic (91.67s)
=== RUN   TestAccAWSRoute53Record_basic_fqdn
--- PASS: TestAccAWSRoute53Record_basic_fqdn (103.92s)
=== RUN   TestAccAWSRoute53Record_txtSupport
--- PASS: TestAccAWSRoute53Record_txtSupport (91.51s)
=== RUN   TestAccAWSRoute53Record_spfSupport
--- PASS: TestAccAWSRoute53Record_spfSupport (91.99s)
=== RUN   TestAccAWSRoute53Record_generatesSuffix
--- PASS: TestAccAWSRoute53Record_generatesSuffix (92.46s)
=== RUN   TestAccAWSRoute53Record_wildcard
--- PASS: TestAccAWSRoute53Record_wildcard (143.37s)
=== RUN   TestAccAWSRoute53Record_failover
--- PASS: TestAccAWSRoute53Record_failover (95.56s)
=== RUN   TestAccAWSRoute53Record_weighted_basic
--- PASS: TestAccAWSRoute53Record_weighted_basic (96.06s)
=== RUN   TestAccAWSRoute53Record_alias
--- PASS: TestAccAWSRoute53Record_alias (98.33s)
=== RUN   TestAccAWSRoute53Record_s3_alias
--- PASS: TestAccAWSRoute53Record_s3_alias (110.01s)
=== RUN   TestAccAWSRoute53Record_weighted_alias
--- PASS: TestAccAWSRoute53Record_weighted_alias (195.86s)
=== RUN   TestAccAWSRoute53Record_geolocation_basic
--- PASS: TestAccAWSRoute53Record_geolocation_basic (98.17s)
=== RUN   TestAccAWSRoute53Record_latency_basic
--- PASS: TestAccAWSRoute53Record_latency_basic (94.33s)
=== RUN   TestAccAWSRoute53Record_TypeChange
--- PASS: TestAccAWSRoute53Record_TypeChange (145.35s)
=== RUN   TestAccAWSRoute53Record_empty
--- PASS: TestAccAWSRoute53Record_empty (99.36s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	1647.991s
```

The small change was made in 1c07582

stack72:
Manually merged :)

Issue #7241, opened 2016-06-20
Using "computed" values in maps not possible


Alars-ALIT:
### Terraform Version
Terraform v0.7.0-rc2 

### Affected Resource(s)
Terraform core

### Terraform Configuration Files
```hcl
resource "null_resource" "intermediates" {
    triggers = {
        test = "testing..."
    }
}


module "test_mod" {
  source = "./mod"

  services = {
    "name"             = "user-service"
    "min"              = "1"
    "max"              = "1"
    "desired"          = "1"
    "instance_type"    = "t2.small"
    "elb"              = "${null_resource.intermediates.triggers.test}" 
  }
}
```

```hcl
variable "services"                { type = "list" }

resource "aws_autoscaling_group" "asg-test" {
  count                     = "${length(var.services)}"
  availability_zones        = ["test"]
  min_size                  = "${lookup(var.services[count.index], "min")}"
  max_size                  = "${lookup(var.services[count.index], "max")}"
  desired_capacity          = "${lookup(var.services[count.index], "desired")}"
  launch_configuration      = "any"
  vpc_zone_identifier       = ["any"]
  load_balancers            = ["${lookup(var.services[count.index], "elb")}"]  
}
```
### Expected Behavior
Expected the map to contain a entry for "elb". I seems like "computed" values can't be passed in maps

### Actual Behavior
Errors:

  * lookup: lookup failed to find 'elb' in:

${lookup(var.services[count.index], "elb")}

### Steps to Reproduce
terraform plan

jen20:
Hi @Alars-ALIT! I can reproduce this - thanks for the concise example. I'll investigate what is going on here, as I see no reason in principle that this should not work.

jen20:
Note to self for investigative context - in `interpolationFuncLookup`:

```
2016/06/24 14:49:03 [INFO] args: ([]interface {}) (len=2 cap=2) {
 (map[string]ast.Variable) (len=5) {
  (string) (len=3) "min": (ast.Variable) {Variable (TypeString): 1},
  (string) (len=3) "max": (ast.Variable) {Variable (TypeString): 1},
  (string) (len=7) "desired": (ast.Variable) {Variable (TypeString): 1},
  (string) (len=13) "instance_type": (ast.Variable) {Variable (TypeString): t2.small},
  (string) (len=4) "name": (ast.Variable) {Variable (TypeString): user-service}
 },
 (string) (len=3) "elb"
}
```

jen20:
[...]

PR #7551, opened 2016-07-08
core: Allow use of computed values in maps

jen20:
This pull-request is the (long-awaited!) reproduction and fix for #7241. The commit messages contain far more detail, but the TL;DR on the fix is that UnknownVariableValue must be set for specific map keys regardless of where in the variable structure they appear, rather than just at the top level in EvalVariableBlock.

cc @phinze.

### DIFF
--- a/config/interpolate_walk.go
+++ b/config/interpolate_walk.go
@@ -54,6 +54,9 @@ type interpolationWalkerContextFunc func(reflectwalk.Location, ast.Node)
 
 func (w *interpolationWalker) Enter(loc reflectwalk.Location) error {
 	w.loc = loc
+	if loc == reflectwalk.WalkLoc {
+		w.sliceIndex = -1
+	}
 	return nil
 }
 
@@ -72,6 +75,7 @@ func (w *interpolationWalker) Exit(loc reflectwalk.Location) error {
 		w.cs = w.cs[:len(w.cs)-1]
 	case reflectwalk.SliceElem:
 		w.csKey = w.csKey[:len(w.csKey)-1]
+		w.sliceIndex = -1
 	}
 
 	return nil
@@ -85,7 +89,13 @@ func (w *interpolationWalker) Map(m reflect.Value) error {
 func (w *interpolationWalker) MapElem(m, k, v reflect.Value) error {
 	w.csData = k
 	w.csKey = append(w.csKey, k)
-	w.key = append(w.key, k.String())
+
+	if w.sliceIndex != -1 {
+		w.key = append(w.key, fmt.Sprintf("%d.%s", w.sliceIndex, k.String()))
+	} else {
+		w.key = append(w.key, k.String())
+	}
+
 	w.lastValue = v
 	return nil
 }
@@ -164,6 +174,7 @@ func (w *interpolationWalker) Primitive(v reflect.Value) error {
 		} else if replaceVal == UnknownVariableValue {
 			remove = true
 		}
+
 		if remove {
 			w.removeCurrent()
 			return nil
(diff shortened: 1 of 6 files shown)


config/interpolate_walk.go
```diff
@@ -54,6 +54,9 @@ type interpolationWalkerContextFunc func(reflectwalk.Location, ast.Node)
 
 func (w *interpolationWalker) Enter(loc reflectwalk.Location) error {
 	w.loc = loc
+	if loc == reflectwalk.WalkLoc {
```
phinze: I had to look this up in the code. `WalkLoc` seems to be "the top of a given walk"

jen20: Correct - this is in lieu of a constructor to set the default value for the interpolate walker struct

config/interpolate_walk.go
```diff
@@ -85,7 +89,13 @@ func (w *interpolationWalker) Map(m reflect.Value) error {
 func (w *interpolationWalker) MapElem(m, k, v reflect.Value) error {
 	w.csData = k
 	w.csKey = append(w.csKey, k)
-	w.key = append(w.key, k.String())
+
+	if w.sliceIndex != -1 {
+		w.key = append(w.key, fmt.Sprintf("%d.%s", w.sliceIndex, k.String()))
+	} else {
+		w.key = append(w.key, k.String())
+	}
```
phinze: I read this as "If we are in a slice, the key for this MapElem needs to be prefixed with the slice index."

Seems sort of weird that only MapElems get this treatment - I'd expect the SliceElem visit to do this work. Is there a reason that MapElems need to be special cased that I am missing?

Related: it'd be really great if we could wrap a unit test around this, not sure how much of a nightmare that would be though.

terraform/eval_variable.go
```diff
@@ -154,15 +156,60 @@ func (n *EvalVariableBlock) Eval(ctx EvalContext) (interface{}, error) {
...
+			tCurrent := current.(map[string]interface{})
+			if val, hasVal := tCurrent[pathComponents[i]]; hasVal {
+				current = val
+				continue
+			}
+
+			tCurrent[pathComponents[i]] = config.UnknownVariableValue
+			break
+		}
+	}
```
phinze: This turned out much nicer since we can walk the path!

It seems to me that a few tweaks to this could allow an element inside a list to be properly marked as computed. Or does that already work via another mechanism?

phinze: In other words, on first glance I'd expect the "list item computed" case to be able to work like:

 * `ComputedKeys` has element like `mylist.4`
 * We recognize that in this walk and mark that list element as computed.

phinze:
Ok! Left some notes/questions inline, but this overall is looking good!

config/interpolate_walk.go
jen20: The reason it is like this is because of the lack of Enter and Exit for Map and Slice elements - meaning we have no opportunity to just mutate the key during slice visit (at least, not in a way that is clean). Not really sure how easy it is to test - however, there is test coverage that this is correct via the next layer up.

terraform/eval_variable.go
jen20: We can't necessarily do mylist.4 without taking into account the movement of other slice elements - if one is removed, the numbering moves during the unknown key determination! I _believe_ lists already work via another mechanism but I will investigate further here and add another context test.

jen20:
Rebased onto master

phinze:
LGTM w/ responses. Can address the question of computed list elements via a separate investigation/PR.

Issue #3923, opened 2015-11-14
Ability to use AMI name as well as ID for AWS instances


nickithewatt:
When creating a new instance within AWS, the AMI referenced (via the `ami` field is specifically to the ID to be used i.e.
```
# Create a new instance of the ami-1234 on an m1.small node
resource "aws_instance" "web" {
    ami = "ami-1234"
    instance_type = "m1.small"
}
```

We are currently uploading / creating our own custom AMIs, which have been given more human friendly logical names. It would be nice to be able to reference this AMI by name instead of just the ID, perhaps an option to have an `ami_name` as an alternative to the current `ami` field which honours just the AMI ID? i.e.

```
# Create a new instance using the AMI with name my-logical-ami-name-v1 and 
# backed by ID ami-1234 on an m1.small node
resource "aws_instance" "web" {
    ami_name = "my-logical-ami-name-v1"
    instance_type = "m1.small"
}
```

stack72:
@nickithewatt what would you expect to be able to input for this? Would it be the ami_name always or would you want to be able to select the ami based on tag?

If it is *always* the ami_name then this can be done easily

nickithewatt:
@stack72 For our use case we would simply need to specify the ami_name, no need for tags at this stage :)

stack72:
Thanks @nickithewatt - I am just working on a PR for this now :)

stack72:
@nickithewatt so the functionality I proposed in #3935 allows you to do as follows:

```
resource "aws_security_group" "tf_test_foo" {
	name = "tf_test_foo"
	description = "foo"

	ingress {
		protocol = "icmp"
		from_port = -1
		to_port = -1
		cidr_blocks = ["0.0.0.0/0"]
	}
}

resource "aws_instance" "foo" {
    ami_name = "BOSH-04e04008-85a5-4f5d-9f37-5eda69f57f81"
    instance_type = "m1.small"
    availability_zone = "us-west-2a"
    security_groups = ["${aws_security_group.tf_test_foo.name}"]
}
```

This will use the Describe-Images API to get the correct AMI Id for that image

catsby:
Hey all – the problem I see with this (and noted in #3935) is that without the ID in the configuration, Terraform won't know to make any changes to your infrastructure if the ID changes. Terraform will impose your configuration on your infrastructure; if the AMI name is still the same, no action regarding the ID will be taken; it will simply make note (store in state file) that the ID changed. 

[...]

PR #7567, opened 2016-07-09
docs/aws: Promote aws_ami data source more

radeksimko:
This makes the example work out of the box for any region and promote the new AMI data source in documentation better.

It should also make it easier to adopt the example for users that don't have `t1.micro` available anymore.

Related: https://github.com/hashicorp/terraform/issues/3923


jen20:
LGTM!

PR #7568, opened 2016-07-09
provider/aws: Discover supported EC2 platforms (EC2 Classic/VPC)

radeksimko:
This will make it possible/easier to make decisions based on supported EC2 platforms (VPC and/or EC2 Classic) in the context of any AWS resource which needs make such decisions.

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html#differences-ec2-classic-vpc

This would help in https://github.com/hashicorp/terraform/pull/6387, https://github.com/hashicorp/terraform/issues/9744 and _possibly_ in `aws_instance`, `aws_db_instance` (`security_groups` vs `vpc_security_group_ids`). I'm not 100% sure about the latter because we may actually need to allow `security_groups` for VPC accounts too, because that's how you reference SGs in the default VPCs.

```
make test TEST=./builtin/providers/aws TESTARGS='-run=TestGetSupportedEC2Platforms -v'
 2>/dev/null
go test -i ./builtin/providers/aws || exit 1
echo ./builtin/providers/aws | \
		xargs -t -n4 go test -run=TestGetSupportedEC2Platforms -v -timeout=30s -parallel=4
=== RUN   TestGetSupportedEC2Platforms
--- PASS: TestGetSupportedEC2Platforms (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	0.031s
```

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -136,6 +136,7 @@ type AWSClient struct {
 	r53conn               *route53.Route53
 	partition             string
 	accountid             string
+	supportedplatforms    []string
 	region                string
 	rdsconn               *rds.RDS
 	iamconn               *iam.IAM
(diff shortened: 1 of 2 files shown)


radeksimko:
I won't mention this in Changelog as it has no impact on the user.

PR #7169, opened 2016-06-14
Adding disk keep_on_remove support on delete case

dkalleg:
This both fixes the keep_on_remove reference in the disk update case
as well as adds logic to safely eject disks in the destroy case.

PR #7579, opened 2016-07-11
Revert the "Revert "Adding disk keep_on_remove support on delete case""

stack72:
Reverts hashicorp/terraform#7572

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -509,8 +509,8 @@ func resourceVSphereVirtualMachineUpdate(d *schema.ResourceData, meta interface{
 				virtualDisk := devices.FindByKey(int32(disk["key"].(int)))
 
 				keep := false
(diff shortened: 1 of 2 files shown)


stack72:
@dkalleg just a FYI, I merged your PR (#7169) but it caused a build failure:

```
# github.com/hashicorp/terraform/builtin/providers/vsphere
builtin/providers/vsphere/resource_vsphere_virtual_machine_test.go:1186: assignment count mismatch: 7 = 8
```

Need you to have a look at this if possible please

dkalleg:
Hey Paul,

I saw you added the change to https://github.com/hashicorp/terraform/pull/7169 to update the test, thanks for doing that.  That should solve the "assignment count mismatch" issue.  I think the number of return values to testCheckFuncBasic() must have changed since I originally wrote the patch.

Did you see it pass after that?  Is there anything further for me to do?

Thanks,
Dan

stack72:
@dialled

All is looking good now :) thanks for messaging back

P.

PR #7586, opened 2016-07-11
Update aws api gateway docs to add some examples

mclarke47:
Reasoning for docs update: 

I have recently been using terraform to define aws api gateways and thought that request_parameters_in_json was fairly ambiguous, so I have added some examples for requests/responses.

I have also corrected an error in the gateway deployment example where there was a missing aws_api_gateway_integration and aws_api_gateway_deployment.stage_name.

Relevant Terraform version:

Tested against latest master.

Any feedback would be appreciated.


radeksimko:
LGTM, thanks.

PR #7589, opened 2016-07-11
command: Do not count data sources in plan totals

### LINKED ISSUE #7483: Counts are wrong with data sources
(Issue deliberately terse and self-assigned!)

Refreshed data sources appear to be included in the counts, probably the hook needs updating.

Same issue as @clstokes found previously.

phinze:
Fixes #7483

### DIFF
--- a/command/hook_count.go
+++ b/command/hook_count.go
@@ -1,6 +1,7 @@
 package command
 
 import (
+	"strings"
 	"sync"
 
 	"github.com/hashicorp/terraform/terraform"
@@ -89,6 +90,11 @@ func (h *CountHook) PostDiff(
 	terraform.HookAction, error) {
 	h.Lock()
 	defer h.Unlock()
+
+	// We don't count anything for data sources
+	if strings.HasPrefix(n.Id, "data.") {
+		return terraform.HookActionContinue, nil
+	}
 
 	switch d.ChangeType() {
 	case terraform.DiffDestroyCreate:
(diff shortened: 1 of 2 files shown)


phinze:
cc @jen20 / @jbardin

command/hook_count.go
```diff
@@ -90,6 +91,11 @@ func (h *CountHook) PostDiff(
 	h.Lock()
 	defer h.Unlock()
 
+	// We don't count enything for data sources
+	if strings.HasPrefix(n.Id, "data.") {
```
apparentlymart: Might this have the same problem as discussed in #6823? Not sure if `n.Id` here is a module-local name or one that might potentially begin with `module.foo` for data resources in nested modules.

phinze: Good question! I thought the same thing, but from reading [`HumanId()`](https://github.com/hashicorp/terraform/blob/edd67547bc2aa12244fcf1620bcb209ee73e008e/terraform/resource.go#L75-L84) I am fairly certain that `InstanceInfo` keeps `Id` unscoped and maintains `ModulePath []string` separately.

Will verify manually as well here.

phinze: Yup manual testing confirms this fix works for data sources in modules. 👍

jbardin:
LGTM if we're sure `.data` is always a prefix

jen20:
LGTM I _think_, but I am somewhat concerned about whether the behaviour here will be different if the resource shim (for backwards compatibilty) is used. I'll check and report back...

jen20:
Yup, data sources using the resource shim appear in counts still - however I think this is correct as they will actually evaluate as resources. So LGTM!

PR #7611, opened 2016-07-12
provider/aws: Fix data.aws_iam_policy_document

### LINKED ISSUE #7499: aws_iam_policy_document data don't generate ids properties
### Terraform Version

### Affected Resource(s)
- aws_iam_policy_document

```hcl
data "aws_iam_policy_document" "key_store_kms" {
  id = "key-consolepolicy-2"
  statement {
    id = "Enable IAM User Permissions"
    actions = ["kms:*"]
    resources = ["*"]
  }
}
```

### Expected Behavior
Adding the `id` arguments to the policy document and the statements I would expect the generated json would have the equivalent `Id` and `Sid` properties fot the policy and the statement.

```json
{
  "Id": "key-consolepolicy-2",
  "Statement": [
    {
      "Sid": "Enable IAM User Permissions",
      "Action": "kms:*",
      "Resource": "*"
    }
  ]
}
```

### Actual Behavior
The policy `Id` and the statement `Sid` are omited in the json

### Steps to Reproduce
1. `terraform apply`

jen20:
Adds ID and Version support to policy documents.

Fixes #7499.

### DIFF
--- a/website/source/docs/providers/aws/d/iam_policy_document.html.markdown
+++ b/website/source/docs/providers/aws/d/iam_policy_document.html.markdown
@@ -17,6 +17,7 @@ such as the `aws_iam_policy` resource.
 ```
 data "aws_iam_policy_document" "example" {
     statement {
+        sid = "1"
         actions = [
             "s3:ListAllMyBuckets",
             "s3:GetBucketLocation",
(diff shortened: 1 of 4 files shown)


website/source/docs/providers/aws/d/iam_policy_document.html.markdown
```diff
@@ -71,14 +72,15 @@ valid to use literal JSON strings within your configuration, or to use the
 
 The following arguments are supported:
 
-* `id` (Optional) - An ID for the policy document.
+* `version` (Optional) - An policy version. Defaults to `2012-10-17`.
```
apparentlymart: We should probably document that if you set this to `2008-10-17` then you can't use policy variables.

(Though I don't know why anyone would set this to anything other than what it defaults to anyway.)

jen20: @apparentlymart the reason I had was for existing ones what shouldn't be modified (for whatever reason) - happy to point out in the docs that the default should be fine for almost all use cases though

jen20: Actually having researched this I am in agreement @apparentlymart, and will remove version again.

PR #7616, opened 2016-07-13
Setting admin_state_up default to true

dkalleg:
This is to be inline with OpenStack defaulting admin_state_up to true.


jtopjian:
LGTM :)

jtopjian:
Posterity note: since the LBaaS v2 resources were added in the same release cycle, this change won't be documented in CHANGELOG.

dkalleg:
Thanks @jtopjian !

PR #7619, opened 2016-07-13
Aws/EMR implementation

gitlabbin:
Based on:
https://github.com/hashicorp/terraform/issues/2098

Covered most of the major parameters in TF specs.
Hope get some feed back, 
and more help welcome on the test cases.

Thanks.

### DIFF
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -182,6 +182,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_elastictranscoder_preset":                 resourceAwsElasticTranscoderPreset(),
 			"aws_elb":                                      resourceAwsElb(),
 			"aws_elb_attachment":                           resourceAwsElbAttachment(),
+			"aws_emr":                                      resourceAwsEMR(),
 			"aws_flow_log":                                 resourceAwsFlowLog(),
 			"aws_glacier_vault":                            resourceAwsGlacierVault(),
 			"aws_iam_access_key":                           resourceAwsIamAccessKey(),
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -183,6 +183,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_elb":                                      resourceAwsElb(),
 			"aws_elb_attachment":                           resourceAwsElbAttachment(),
 			"aws_emr":                                      resourceAwsEMR(),
+			"aws_emr_task_group":                           resourceAwsEMRTaskGroup(),
 			"aws_flow_log":                                 resourceAwsFlowLog(),
 			"aws_glacier_vault":                            resourceAwsGlacierVault(),
 			"aws_iam_access_key":                           resourceAwsIamAccessKey(),
--- a/builtin/providers/aws/resource_aws_emr.go
+++ b/builtin/providers/aws/resource_aws_emr.go
@@ -121,6 +121,11 @@ func resourceAwsEMR() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"visible_to_all_users": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 		},
 	}
 }
@@ -186,7 +191,7 @@ func resourceAwsEMRCreate(d *schema.ResourceData, meta interface{}) error {
 		JobFlowRole:       aws.String(instanceProfile),
 		ReleaseLabel:      aws.String(d.Get("release_label").(string)),
 		ServiceRole:       aws.String(serviceRole),
-		VisibleToAllUsers: aws.Bool(true),
+		VisibleToAllUsers: aws.Bool(d.Get("visible_to_all_users").(bool)),
 	}
 
 	if v, ok := d.GetOk("log_uri"); ok {
--- a/builtin/providers/aws/resource_aws_emr.go
+++ b/builtin/providers/aws/resource_aws_emr.go
@@ -220,6 +220,36 @@ func resourceAwsEMRCreate(d *schema.ResourceData, meta interface{}) error {
 }
 
 func resourceAwsEMRRead(d *schema.ResourceData, meta interface{}) error {
+	emrconn := meta.(*AWSClient).emrconn
+
+	req := &emr.DescribeClusterInput{
+		ClusterId: aws.String(d.Id()),
+	}
+
+	resp, err := emrconn.DescribeCluster(req)
+	if err != nil {
+		return fmt.Errorf("Error reading EMR cluster: %s", err)
+	}
+	fmt.Println(resp)
+
+	if resp.Cluster == nil {
+		d.SetId("")
+		return nil
+	}
+
+	instance := resp.Cluster
+
+	if instance.Status != nil {
+		if *resp.Cluster.Status.State == "TERMINATED" {
+			d.SetId("")
+			return nil
+		}
+
+		if *resp.Cluster.Status.State == "TERMINATED_WITH_ERRORS" {
+			d.SetId("")
+			return nil
+		}
+	}
 
 	return nil
 }
--- a/builtin/providers/aws/resource_aws_emr.go
+++ b/builtin/providers/aws/resource_aws_emr.go
@@ -35,6 +35,7 @@ func resourceAwsEMR() *schema.Resource {
 			"core_instance_type": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 			"core_instance_count": &schema.Schema{
 				Type:     schema.TypeInt,
@@ -216,7 +217,7 @@ func resourceAwsEMRCreate(d *schema.ResourceData, meta interface{}) error {
 	fmt.Println(resp)
 	d.SetId(*resp.JobFlowId)
 
-	return nil
+	return resourceAwsEMRRead(d, meta)
 }
 
 func resourceAwsEMRRead(d *schema.ResourceData, meta interface{}) error {
@@ -249,6 +250,12 @@ func resourceAwsEMRRead(d *schema.ResourceData, meta interface{}) error {
 			d.SetId("")
 			return nil
 		}
+	}
+
+	instanceGroups, errGrps := loadGroups(d, meta)
+	if errGrps == nil {
+		coreGroup := findGroup(instanceGroups, "CORE")
+		d.Set("core_instance_type", coreGroup.InstanceType)
 	}
 
 	return nil
@@ -321,6 +328,19 @@ func expandApplications(apps []interface{}) []*emr.Application {
 		appOut = append(appOut, app)
 	}
 	return appOut
+}
+
+func loadGroups(d *schema.ResourceData, meta interface{}) ([]*emr.InstanceGroup, error) {
+	emrconn := meta.(*AWSClient).emrconn
+	reqGrps := &emr.ListInstanceGroupsInput{
+		ClusterId: aws.String(d.Id()),
+	}
+
+	respGrps, errGrps := emrconn.ListInstanceGroups(reqGrps)
+	if errGrps != nil {
+		return nil, fmt.Errorf("Error reading EMR cluster: %s", errGrps)
+	}
+	return respGrps.InstanceGroups, nil
 }
 
 func findGroup(grps []*emr.InstanceGroup, typ string) *emr.InstanceGroup {
--- a/builtin/providers/aws/resource_aws_emr.go
+++ b/builtin/providers/aws/resource_aws_emr.go
@@ -421,12 +421,10 @@ func readHttpJson(url string, target interface{}) error {
 func readLocalJson(localFile string, target interface{}) error {
 	file, e := ioutil.ReadFile(localFile)
 	if e != nil {
-		fmt.Printf("File error: %v\n", e)
 		log.Printf("[ERROR] %s", e)
 		return e
 	}
-	log.Printf("[DEBUG] %s\n", string(file))
-
+	
 	return json.Unmarshal(file, target)
 }
 
--- a/builtin/providers/aws/resource_aws_emr.go
+++ b/builtin/providers/aws/resource_aws_emr.go
@@ -248,14 +248,14 @@ func resourceAwsEMRRead(d *schema.ResourceData, meta interface{}) error {
 
 	if instance.Status != nil {
 		if *resp.Cluster.Status.State == "TERMINATED" {
-			d.SetId("")
 			log.Printf("[DEBUG] EMR Cluster (%s) was TERMINATED already", d.Id())
+			d.SetId("")
 			return nil
 		}
 
 		if *resp.Cluster.Status.State == "TERMINATED_WITH_ERRORS" {
-			d.SetId("")
 			log.Printf("[DEBUG] EMR Cluster (%s) was TERMINATED_WITH_ERRORS already", d.Id())
+			d.SetId("")
 			return nil
 		}
 	}
@@ -302,7 +302,7 @@ func resourceAwsEMRUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Modify EMR Cluster done...")
 
-	return nil
+	return resourceAwsEMRRead(d, meta)
 }
 
 func resourceAwsEMRDelete(d *schema.ResourceData, meta interface{}) error {
--- a/builtin/providers/aws/resource_aws_emr.go
+++ b/builtin/providers/aws/resource_aws_emr.go
@@ -6,11 +6,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/emr"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 )
 
 func resourceAwsEMR() *schema.Resource {
@@ -223,6 +226,23 @@ func resourceAwsEMRCreate(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Created EMR Cluster done...")
 	d.SetId(*resp.JobFlowId)
+
+	log.Println(
+		"[INFO] Waiting for EMR Cluster to be available")
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"STARTING", "BOOTSTRAPPING"},
+		Target:     []string{"WAITING", "RUNNING"},
+		Refresh:    resourceAwsEMRClusterStateRefreshFunc(d, meta),
+		Timeout:    40 * time.Minute,
+		MinTimeout: 10 * time.Second,
+		Delay:      30 * time.Second, // Wait 30 secs before starting
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("[WARN] Error waiting for EMR Cluster state to be \"WAITING\": %s", err)
+	}
 
 	return resourceAwsEMRRead(d, meta)
 }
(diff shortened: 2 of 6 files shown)


builtin/providers/aws/provider.go
```diff
@@ -182,6 +182,8 @@ func Provider() terraform.ResourceProvider {
 			"aws_elastictranscoder_preset":                 resourceAwsElasticTranscoderPreset(),
 			"aws_elb":                                      resourceAwsElb(),
 			"aws_elb_attachment":                           resourceAwsElbAttachment(),
+			"aws_emr":                                      resourceAwsEMR(),
```
stack72: is this a wrapper resource for `emr_cluster`?

gitlabbin: yes, it is.

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"master_instance_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"core_instance_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
catsby: If this attribute is omitted, will AWS provide it? Should we make it `Optional` and `Computed` and read it?

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+	if v, ok := d.GetOk("core_instance_type"); ok {
+		coreInstanceType = v.(string)
+	}
+	coreInstanceCount := d.Get("core_instance_count").(int)
+
+	applications := d.Get("applications").(*schema.Set).List()
+	var userKey, subnet, extraMasterSecGrp, extraSlaveSecGrp, emrMasterSecGrp, emrSlaveSecGrp, instanceProfile, serviceRole string
+	instanceProfile = "EMR_EC2_DefaultRole"
+	ec2Attributes := d.Get("ec2_attributes").([]interface{})
+	if len(ec2Attributes) == 1 {
```
catsby: we should probably error if they define 2+ blocks of EC2 Attributes. As is, if they have 2 blocks then nothing will happen. You can also leverage the `MaxItems` attribute of a schema, if you want to enforce that:

- https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L108

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+			EmrManagedMasterSecurityGroup: aws.String(emrMasterSecGrp),
+			EmrManagedSlaveSecurityGroup:  aws.String(emrSlaveSecGrp),
+		},
+		Name:         aws.String(d.Get("name").(string)),
+		Applications: emrApps,
+
+		JobFlowRole:       aws.String(instanceProfile),
+		ReleaseLabel:      aws.String(d.Get("release_label").(string)),
+		ServiceRole:       aws.String(serviceRole),
+		VisibleToAllUsers: aws.Bool(true),
```
catsby: This seems like an optional attribute we may want to expose, but with a default of `true`. Not a blocker thought 😄

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+	if v, ok := d.GetOk("tags"); ok {
+		tagsIn := v.(map[string]interface{})
+		params.Tags = expandTags(tagsIn)
+	}
+	if v, ok := d.GetOk("configurations"); ok {
+		confUrl := v.(string)
+		params.Configurations = expandConfigures(confUrl)
+	}
+
+	resp, err := conn.RunJobFlow(params)
```
catsby: nitpick: let's log that we're executing here, just a simple log of the params:

- https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_rds_cluster.go#L355

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+
+	if err != nil {
+		log.Printf("[ERROR] %s", err)
+		return err
+	}
+
+	log.Printf("[DEBUG] Created EMR Cluster done...")
+	fmt.Println(resp)
```
catsby: debug statement

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+
+	req := &emr.DescribeClusterInput{
+		ClusterId: aws.String(d.Id()),
+	}
+
+	resp, err := emrconn.DescribeCluster(req)
+	if err != nil {
+		return fmt.Errorf("Error reading EMR cluster: %s", err)
+	}
+	fmt.Println(resp)
```
catsby: debug

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+
+	if resp.Cluster == nil {
+		d.SetId("")
```
catsby: as a pattern we should log that we didn't find the cluster we were looking for, so we're removing from state:

- https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_rds_cluster.go#L398

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+			d.SetId("")
+			return nil
+		}
+
+		if *resp.Cluster.Status.State == "TERMINATED_WITH_ERRORS" {
+			d.SetId("")
+			return nil
+		}
+	}
+
```
catsby: Is there nothing worth saving from the `DescribeCluster` call? Seems there is a `Status` struct, does it contain anything we could use, if for nothing else to update the status of the cluster and it's operations? 

- http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_DescribeCluster.html

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,413 @@
...
+	log.Printf("[DEBUG] Modify EMR cluster")
+	req := &emr.ListInstanceGroupsInput{
+		ClusterId: aws.String(d.Id()),
+	}
+
+	respGrps, errGrps := conn.ListInstanceGroups(req)
+	if errGrps != nil {
+		return fmt.Errorf("Error reading EMR cluster: %s", errGrps)
+	}
+	fmt.Println(respGrps)
```
catsby: debug

builtin/providers/aws/resource_aws_emr.go
gitlabbin: By default, "core_instance_type" will be the same as "master_instance_type", 
- [x] As you suggested make it into "Computed"

builtin/providers/aws/resource_aws_emr.go
gitlabbin: Cool, `MaxItems` make it simple
- [x] apply `MaxItems ` for ec2Attributes

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] a simple log of the params for create EMR cluster

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] remove debug statement

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] remove `fmt.Println(resp)`

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] remove `fmt.Println(respGrps)`

builtin/providers/aws/resource_aws_emr.go
gitlabbin: Follow TF resource not found pattern.
- [x] Add log for EMR cluster not found

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] Expose `visible_to_all_users` as an option, default will be `true`

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+		log.Printf("[DEBUG] EMR Cluster (%s) not found", d.Id())
+		return nil
+	}
+
+	instance := resp.Cluster
+
+	if instance.Status != nil {
+		if *resp.Cluster.Status.State == "TERMINATED" {
+			d.SetId("")
+			log.Printf("[DEBUG] EMR Cluster (%s) was TERMINATED already", d.Id())
```
catsby: Minor nit: need to log before we call `d.SetId("")` here, otherwise we'll log an empty string 😄

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+	_, errModify := conn.ModifyInstanceGroups(params)
+	if errModify != nil {
+		log.Printf("[ERROR] %s", errModify)
+		return errModify
+	}
+
+	log.Printf("[DEBUG] Modify EMR Cluster done...")
+
+	return nil
```
catsby: let's end this method with `return resourceAwsEMRRead(d, meta)`

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+	instanceGroups, errGrps := loadGroups(d, meta)
+	if errGrps == nil {
+		coreGroup := findGroup(instanceGroups, "CORE")
+		d.Set("core_instance_type", coreGroup.InstanceType)
+	}
+
+	return nil
+}
+
+func resourceAwsEMRUpdate(d *schema.ResourceData, meta interface{}) error {
```
catsby: For any attribute that can be updated, we should check for a change first before calling out to modify. See [AWS DB Instance](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_instance.go#L793) as a possible example. This way we don't call out to the API to modify a cluster if there has been no updates. 

Is `core_instance_count` the only attribute that can change at the moment?

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+	defer r.Body.Close()
+
+	return json.NewDecoder(r.Body).Decode(target)
+}
+
+func readLocalJson(localFile string, target interface{}) error {
+	file, e := ioutil.ReadFile(localFile)
+	if e != nil {
+		fmt.Printf("File error: %v\n", e)
```
catsby: debug statement

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+		log.Printf("[ERROR] %s", e)
+		return e
+	}
+	log.Printf("[DEBUG] %s\n", string(file))
```
catsby: another debug (need to do a sweep of these, some are relevant, others seem just for development)

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"release_label": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"master_instance_type": &schema.Schema{
```
catsby: I see `master_instance_type` and a few other attributes are not updatable it seems; there's no code in `UPDATE` to change them. Is this intentional? I have not used EMR so maybe this is normal? 

If these are attributes that you cannot update on a _live_ EMR cluster, then they should be marked as `ForceNew` so that changing them re-creates the cluster

builtin/providers/aws/resource_aws_emr_task_group.go
```diff
@@ -0,0 +1,114 @@
...
+		Delete: resourceAwsEMRTaskGroupDelete,
+		Schema: map[string]*schema.Schema{
+			"cluster_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"instance_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
```
catsby: I see in the UPDATE method here that `instance_type` is not something the API supports updating; should this attribute then be marked as `ForceNew` so that we'll destroy->create on change?

builtin/providers/aws/resource_aws_emr_task_group.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+			"instance_count": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  60,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
catsby: what happens if you don't specify a `name` here? From the code below, we still try to send `""`, is the API fine with this?

builtin/providers/aws/resource_aws_emr_task_group.go
```diff
@@ -0,0 +1,114 @@
...
+
+	log.Printf("[DEBUG] Created EMR task group finished: %#v", resp)
+	d.SetId(*resp.InstanceGroupIds[0])
+
+	return nil
+}
+
+func resourceAwsEMRTaskGroupRead(d *schema.ResourceData, meta interface{}) error {
+
+	return nil
```
catsby: Is there a reason we don't read any attributes from the API here? Are they unavailable?

builtin/providers/aws/resource_aws_emr_task_group.go
```diff
@@ -0,0 +1,114 @@
...
+		log.Printf("[ERROR] %s", err)
+		return err
+	}
+
+	log.Printf("[DEBUG] Modify EMR task group finished: %#v", resp)
+
+	return nil
+}
+
+func resourceAwsEMRTaskGroupDelete(d *schema.ResourceData, meta interface{}) error {
```
catsby: seems there is no true `delete` in the api here. would it be better for use to scale it down to zero and/or terminate the instances associated with it?

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+		tagsIn := v.(map[string]interface{})
+		params.Tags = expandTags(tagsIn)
+	}
+	if v, ok := d.GetOk("configurations"); ok {
+		confUrl := v.(string)
+		params.Configurations = expandConfigures(confUrl)
+	}
+
+	log.Printf("[DEBUG] EMR Cluster create options: %s", params)
+	resp, err := conn.RunJobFlow(params)
```
catsby: We need to poll the after creation process and validate that the EMR Cluster enters the state ([status defined here](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_ClusterStatus.html)) we're expecting. I used [the example config from the test](https://github.com/hashicorp/terraform/pull/7619/files#diff-d53b8d8e18462f63430bdce6475994b2R82) and while Terraform reported everything was ok, I checked the console and after some time the console reported an error:

emr_error


See AWS DB for example:

- https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_instance.go#L603-L622

builtin/providers/aws/resource_aws_emr_test.go
```diff
@@ -0,0 +1,111 @@
...
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
+		}
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No cluster id set")
+		}
+		conn := testAccProvider.Meta().(*AWSClient).emrconn
+		_, err := conn.DescribeCluster(&emr.DescribeClusterInput{
```
catsby: For this test, let's check that the response lists the EMR Cluster that we're looking for, and possibly even check the `status` to be in an acceptable range

builtin/providers/aws/resource_aws_emr.go
```diff
@@ -0,0 +1,439 @@
...
+	instance := resp.Cluster
+
+	if instance.Status != nil {
+		if *resp.Cluster.Status.State == "TERMINATED" {
+			d.SetId("")
+			log.Printf("[DEBUG] EMR Cluster (%s) was TERMINATED already", d.Id())
+			return nil
+		}
+
+		if *resp.Cluster.Status.State == "TERMINATED_WITH_ERRORS" {
```
catsby: I discovered that `TERMINATED_WITH_ERRORS` does not mean _destroyed_; the resource still existed in the console. This will leave a dangling resource if we remove the ID from the state. I don't know if we'd hit this however if we polled in the creation process to verify that the cluster gets setup correctly

catsby: Maybe I misunderstand here. We can still read it in the API and view it in the console, but even in the console there doesn't seem to be a way to delete it any further. I suppose it goes away after some time?

builtin/providers/aws/resource_aws_emr.go
gitlabbin: Good catch :)
- [x] move up one line

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] apply this as return

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] add `d.HasChange("core_instance_count") ` checking

You are right
`core_instance_count` is the only attribute can be changed

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] clean the log

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] clean up debug log

builtin/providers/aws/resource_aws_emr.go
gitlabbin: - [x] add `ForceNew: true,` for `master_instance_type `

builtin/providers/aws/resource_aws_emr_task_group.go
gitlabbin: - [x] add `ForceNew: true,` to task group `instance_type`

builtin/providers/aws/resource_aws_emr_task_group.go
gitlabbin: - The aws-sdk doesn't care the name

builtin/providers/aws/resource_aws_emr_task_group.go
gitlabbin: - the only thing you can touch is the count of the task group

builtin/providers/aws/resource_aws_emr_task_group.go
gitlabbin: You are right, task group still control by the cluster, there is no real delete happen.

builtin/providers/aws/resource_aws_emr.go
gitlabbin: * I did implement some polling code before, it's too random for how long time AWS will take to finish the provisioning and bootstrapping. 
* The test case need some improvement. the security group circular dependency still a problem for the  testing

builtin/providers/aws/resource_aws_emr.go
gitlabbin: you are right,  AWS-SDK allows to terminate, no delete at all, you still can query the cluster information, even you terminate it.

builtin/providers/aws/resource_aws_emr_test.go
gitlabbin: - [x] add checking for EMR Cluster id and status

catsby:
Hey @gitlabbin thanks for all the work here! We decided to push it over the finish line in #9106 so as not to bug you with the last mile work, after you already did the bulk of it. Thanks so much!

catsby:
#9106 was just merged, thanks again for your contribution!

gitlabbin:
@catsby @stack72 , my pleasure work with you guys. great product you are creating.

Best Regards,
Brian Chen

Issue #7195, opened 2016-06-16
0.7-rc2: cannot pass "indexed" output from splat resources within module A to module B


modax:
Hi,

I think there is a problem with outputting attributes from splat resources from modules. When I later try to pass such an output to another module by extracting element from it, terraform fails to do that. I attach a test case which triggers this bug.

### Terraform Version
v0.7.0-rc2-22-g5fdcf5d

### Affected Resource(s)

Core issue. Related to support for lists in module parameters.

### Terraform Configuration Files

mod/mod.tf
```hcl
variable "param" {
  type = "list"
}

resource "template_file" "tmpl" {
  count = "2"
  template = "doesnt"
}

output "out_from_splat" {
  type = "list"
  value = [ "${template_file.tmpl.*.rendered}" ]
}
```

test.tf
```hcl
module "mod1" {
  source = "mod"
  param = [ "this", "one", "works" ]
}

module "mod2" {
  source = "mod"
  param = [ "${module.mod1.out_from_splat[0]}" ]
  #param = [ "${element(module.mod1.out_from_splat, 0)}" ]
}
```

### Expected Behavior
`terraform plan` should have succeeded.

### Actual Behavior
`terraform plan` fails with:

```
Error configuring: 1 error(s) occurred:

* At column 3, line 1: list "module.mod1.out_from_splat" does not have any elements so cannot determine type. in:

${module.mod1.out_from_splat[0]}
```

### Steps to Reproduce
1. `terraform get`
2. `terraform plan`

### Important Factoids
If you try the same with ${element} (uncomment the line in `test.tf`), terraform will panic with division by 0.

phinze:
Thanks for the detailed report here @modax, we'll definitely look at this soon. :ok_hand:

jen20:
We should try to reproduce this once the fix for #7143 is merged (shortly).

modax:
Hi @jen20,

I have tested current build and the bug is fixed. However, #7143 is not a PR which fixes it. I have bisected the commit to:

```
2356afde84ba499a2d977360f5a52dfe205227e3 is the first new commit
commit 2356afde84ba499a2d977360f5a52dfe205227e3
Author: James Nugent <james@jen20.com>
Date:   Thu Jun 23 21:14:09 2016 +0100

    core: Fix interpolation of unknown multi-variables
    
    This commit test "TestContext2Input_moduleComputedOutputElement"
    by ensuring that we treat a count of zero and non-reified resources
[...]

Issue #7143, opened 2016-06-12
Unable to make splat reference to attribute of type map


phinze:
### Terraform Version

master / 2127466280172a7cd0aa1d863cb932aa0107c986 

### Affected Resource(s)

core

### Terraform Configuration Files
```hcl
resource "aws_security_group" "foo" {
  count = 3
  tags {
    Name = "sg-${count.index}"
  }
}

resource "aws_instance" "web" {
  count                  = 2
  instance_type          = "t2.nano"
  ami                    = "ami-06b94666"
  vpc_security_group_ids = ["${aws_security_group.foo.*.id}"]
}

output "listoflists" { value = ["${aws_instance.web.*.vpc_security_group_ids}"] }
output "listofmaps" { value = ["${aws_security_group.foo.*.tags}"] }
```

### Steps to Reproduce

1. `TF_LOG=warn terraform apply`

### Expected Behavior

Outputs including lists of lists and list of maps.

### Actual Behavior

The `listoflists` output is flattened (perhaps this is intentional behavior?):

```
listoflists = [
  sg-553ce633
  sg-543ce632
  sg-523ce634
  sg-553ce633
  sg-543ce632
  sg-523ce634
]
```

And the `listofmaps` output cannot be referenced. Two warnings are printer:

```
2016/06/12 12:17:01 [WARN] Output interpolation "listofmaps" failed: Resource 'aws_security_group.foo' does not have attribute 'tags' for variable 'aws_security_group.foo.*.tags'
2016/06/12 12:17:01 [WARN] Output interpolation "listofmaps" failed: Resource 'aws_security_group.foo' does not have attribute 'tags' for variable 'aws_security_group.foo.*.tags'
```

PR #7628, opened 2016-07-13
core: Add context test for module var from splat

jen20:
This adds additional coverage of the situation reported in #7195 to prevent against regression. The actual fix was in 2356afd, in response to #7143.


jen20:
cc @phinze

PR #7634, opened 2016-07-13
docs: Improve the digitalocean_tag usage example

joonas:
I noticed the comment on the example was incorrect and since there wasn't an example of how to use them with the `digitalocean_droplet` resource, I figured I'd include one too.


stack72:
Thanks @joonas 

Will merge this as soon as the build goes green!

P.

PR #7638, opened 2016-07-14
provider/aws: Bump the SDK version to v1.2.5

stack72:
This will allow for new changes to the ECS


jen20:
LGTM. We should merge this sooner rather than later to get at least one acctest run with it

PR #7643, opened 2016-07-14
provider/vSphere: Fix for IPv6 only environment creation.

kristinn:
The code only waited until one or more IPv4 interfaces came online.
If you only had IPv6 interfaces attached to your machine, then the
machine creation process would completely stall.


chrislovecnm:
Will review when I have a chance.

chrislovecnm:
LGTM ... @catsby can we get a merge?

catsby:
Thanks!

PR #7657, opened 2016-07-14
Increased lambda event mapping creation timeout

kellydavid:
Increased the retry timeout from 1 to 5 minutes due to the time for IAM permission propagation.


radeksimko:
Hi @kellydavid 
thanks for the PR. This makes sense to me.

Would you mind bumping the timeout to 5mins in Update function too?
Since IAM is eventually consistent the `Create` operation may succeed and subsequent `Update` operation may fail (i.e. may need retry too).

kellydavid:
@radeksimko Thanks, made that change.

radeksimko:
👍

PR #7659, opened 2016-07-14
[Vagrantfile] set resources for the provider 'parallels'

cblecker:
Small change to set memory and CPU resources for the [Parallels Vagrant provider](https://parallels.github.io/vagrant-parallels/docs/configuration.html).


stack72:
hi @cblecker 

Paul

PR #7661, opened 2016-07-15
provider/openstack: Support Import of OpenStack Networking Resources

jtopjian:
Router-based resources are not included. They will be added later.


stack72:
LGTM!

PR #7667, opened 2016-07-15
provider/aws: Fixup skip_final_snapshot import

phinze:
Neither skip_final_snapshot nor final_snapshot_identifier can be fetched
from any API call, so we need to default skip_final_snapshot to false
during import so that final_snapshot_identifier is not required

### DIFF
--- a/builtin/providers/aws/import_aws_db_instance_test.go
+++ b/builtin/providers/aws/import_aws_db_instance_test.go
@@ -23,7 +23,10 @@ func TestAccAWSDBInstance_importBasic(t *testing.T) {
 				ImportState:       true,
 				ImportStateVerify: true,
 				ImportStateVerifyIgnore: []string{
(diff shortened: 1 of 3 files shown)


phinze:
Waiting on the acc test to pass locally, then we can apply this same logic to the other `skip_snapshot_identifier` resources. /cc @stack72

phinze:
Haha I got the bool backwards! Silly negative-named boolean params. 😝

Fixed up and re-running test. See you in ~380 seconds. 😑

catsby:
👍

phinze:
After fighting the test harness for a bit, we got there!

```
=== RUN   TestAccAWSDBInstance_importBasic
--- PASS: TestAccAWSDBInstance_importBasic (515.54s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    515.569s
```

Issue #7269, opened 2016-06-22
Unable to add rule for ICMP type 0 (Echo Reply) using aws_network_acl_rule


kwilczynski:
At the moment, using `GetOk()` over `schema.TypeInt` with the value of 0 for `icmp_type` (the `icmp_code` is also affected by this) results in the value of 0 being discarded as empty and/or "zero value", as per the following: [resource_aws_network_acl_rule.go#L106-L115](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_network_acl_rule.go#L106-L115)

```go
// Specify additional required fields for ICMP
if p == 1 {
	params.IcmpTypeCode = &ec2.IcmpTypeCode{}
	if v, ok := d.GetOk("icmp_code"); ok {
		params.IcmpTypeCode.Code = aws.Int64(int64(v.(int)))
	}
	if v, ok := d.GetOk("icmp_type"); ok {
		params.IcmpTypeCode.Type = aws.Int64(int64(v.(int)))
	}
}
```

This affects version 0.6.16 and newer, probably since the code was first introduced.

An example:

```hcl
resource "aws_network_acl_rule" "rule_ingress_allow_icmp_echo_reply" {
  network_acl_id = "${module.acl.id}"

  rule_number = 1
  rule_action = "allow"

  protocol   = "icmp"
  cidr_block = "0.0.0.0/0"

  icmp_type = 0
  icmp_code = -1
}
```

Would produce an error:

```
* aws_network_acl_rule.rule_ingress_allow_icmp_echo_reply: Error Creating Network Acl Rule: MissingParameter: The request must contain the parameter icmpTypeCode.type
	status code: 400, request id: 16a1bb97-ced4-4ef7-bc3e-24eca1294f9d
```

I think, that since there is no concept of an "_undefined variable_" that would uniformly be some sort of a special magic value which could be used to indicate that the attribute is indeed undefined, plus changing the way how Schema is designed (especially with the helper functions like the aforementioned `GetOk()`) would be too much of a change, the simplest fix, which I can think of, would be to make these two attributes to be of a string type (as it has its default "zero value" set to be an empty string internally), and then check whether string is not empty. If not, then try `strconv.ParseInt()` over the value and handle error correctly should there be any.

In other words, an empty string would indicate unset value, and any other value that does not cause parsing error would be a valid integer ICMP type, including 0.

**Edit**: List of ICMP types and codes, for posterity: http://www.nthelp.com/icmp.html

[...]

PR #7669, opened 2016-07-15
Fix icmp_type and icmp_code in aws_network_acl_rule.

kwilczynski:
The ICMP type 0 (Echo Reply) was not handled correctly. This commit changes the
type of attributes "icmp_type" and "icmp_code" from TypeInt to TypeString,
allowing for the string value to be manually converted into an integer. This
enables an integer values such as -1, 0, 8, etc., coming from the resource
definition in the template to be handled correctly.

### DIFF
--- a/builtin/providers/aws/resource_aws_network_acl_rule.go
+++ b/builtin/providers/aws/resource_aws_network_acl_rule.go
@@ -63,14 +63,16 @@ func resourceAwsNetworkAclRule() *schema.Resource {
 				ForceNew: true,
 			},
 			"icmp_type": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateICMPArgumentValue,
 			},
 			"icmp_code": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateICMPArgumentValue,
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


kwilczynski:
Resolves #7269.

kwilczynski:
@phinze done! Let me know how does this look to you.

builtin/providers/aws/resource_aws_network_acl_rule.go
```diff
@@ -63,12 +63,12 @@ func resourceAwsNetworkAclRule() *schema.Resource {
 				ForceNew: true,
 			},
 			"icmp_type": &schema.Schema{
-				Type:     schema.TypeInt,
+				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
```
phinze: Let's also add a `ValidateFunc` to these fields that attempts the `Atoi` and returns an error when parsing fails. This will let users get much faster feedback if they typo something that is not an integer.

phinze:
Looking good so far! If you can get the `ValidateFunc` added I think this is ready to go.

I'll drop a `[WIP]` (work in progress) prefix on the title - edit the title to pull it back out once you have a `ValidateFunc` done. And let me know if you need any help! 👍

kwilczynski:
@phinze leave it with me.

kwilczynski:
Test is passing:

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSNetworkAclRule_basic'
2016/07/17 04:19:21 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSNetworkAclRule_basic -timeout 120m
=== RUN   TestAccAWSNetworkAclRule_basic
--- PASS: TestAccAWSNetworkAclRule_basic (54.03s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	54.055s
```

kwilczynski:
@phinze over to you! 🚀

phinze:
This looks great - thanks for all your work on this, @kwilczynski!

PR #7672, opened 2016-07-16
provider/aws: Clean up aws config path a bit

phinze:
Rearrange client setup, and remove the extraneous log lines we make per
connection. There's no need to log one line per API client - we're just
setting up structs for most of them.

Since this collapses the file down quite a bit, switch to alphabetized
client setup, since previously there wasn't much of an order to things.


catsby:
LGTM

PR #7674, opened 2016-07-16
provider/azurerm: azurerm_storage_account now returns storage keys value instead of their names

shanoor:
`azurerm_storage_account` was returning storage keys name when accessing `primary_access_key` and `secondary_access_key`, which is pretty much useless for now. Now it returns the key values.

Next step would be to return a map of the keys with their name and value but I'm not familiar enough with `go` to do it right away.


stack72:
Hi @shanoor 

Thanks for picking this up! LGTM

Paul

PR #7678, opened 2016-07-17
provider/consul: consul_keys data source

apparentlymart:
Previously the `consul_keys` resource did double-duty as both a reader and writer of values from the Consul key/value store, but that made its interface rather confusing and complex, as well as having all of the other general problems associated with read-only resources.

Here we split the functionality such that reading is done with the `consul_keys` data source while writing is still done with the `consul_keys` resource. (or, indeed, the `consul_key_prefix` resource)

The old read behavior of the resource is still supported, but it's no longer documented (except as a deprecation note) and will generate deprecation warnings when used.

In future it should be possible to simplify the consul_keys resource by removing all of the read support, but that is deferred for now to give users a chance to gracefully migrate to the new data source.

### DIFF
--- a/builtin/providers/consul/data_source_consul_keys.go
+++ b/builtin/providers/consul/data_source_consul_keys.go
@@ -1,0 +1,96 @@
+package consul
+
+import (
+	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceConsulKeys() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceConsulKeysRead,
+
+		Schema: map[string]*schema.Schema{
+			"datacenter": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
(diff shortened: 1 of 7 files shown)


apparentlymart:
I added the release-0.7 tag here in the hope that this patch is good to merge and can go into 0.7 with relatively little fanfare. However, if it seems problematic then I'm happy to defer to 0.7.1 so that any revisions don't block the release.

jen20:
Hi @apparentlymart, I'm happy to merge this for 0.7, I'll review on Monday.

mgaffney:
@jen20 Have you had a chance to review this?

stack72:
Hey @apparentlymart 

Just had a look through this and ran the tests :)

```
% make testacc TEST=./builtin/providers/consul TESTARGS='-run=TestAccDataConsulKeys_'
TF_ACC=1 go test ./builtin/providers/consul -v -run=TestAccDataConsulKeys_ -timeout 120m
--- PASS: TestAccDataConsulKeys_basic (0.03s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/consul	0.045s
```

This LGTM!

PR #7681, opened 2016-07-17
provider/azurerm: fix vault_certificates in vm/vm scale_sets

smesch:
When setting the `certificate_url` and `certificate_store` values in
`os_profile_secrets` / `vault_certificates` for a Windows VM in AzureRM, I
was getting the following error:

```
[DEBUG] Error setting Virtual Machine Storage OS Profile Secrets:
&errors.errorString{s:"Invalid address to set:
[]string{\"os_profile_secrets\", \"0\", \"vault_certificates\"}"}
```

This seemed to have resolve the issue when creating a regular VM (`resource_arm_virtual_machine.go`) and I went ahead and made the corresponding changes for VM scale sets as well (`resource_arm_virtual_machine_scale_set.go`).

Regarding the tests, I wasn't sure if it would be possible to handle this scenario as from what I can see, there is no way to create/interact with Azure Key Vaults from Terraform.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_virtual_machine.go
+++ b/builtin/providers/azurerm/resource_arm_virtual_machine.go
@@ -339,7 +339,7 @@ func resourceArmVirtualMachine() *schema.Resource {
 						},
 
 						"vault_certificates": {
-							Type:     schema.TypeSet,
+							Type:     schema.TypeList,
 							Optional: true,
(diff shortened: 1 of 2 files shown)


stack72:
Hi @smesch 

Thanks so much for the fix here - this LGTM!

Paul

PR #7688, opened 2016-07-18
Allowing ap-south-1 (Mumbai) as valid AWS region

avichalbadaya:


stack72:
Hi @avichalbadaya 

Thanks so much for the PR here. The work has already been done for this - we are just waiting on the 0.7 release

thanks

Paul

avichalbadaya:
@stack72 Thanks for letting me know. Any rough date range when TF 0.7 will be released ?

radeksimko:
@stack72 I think @avichalbadaya's PR is still helpful - have a look at the code changes.

@avichalbadaya There's just one change I'd like you to make before merging this - the S3 website endpoint for the new region is different per [documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html) - contains a dot instead of a dash. This means we'll need to update the [`WebsiteDomain` function](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_s3_bucket.go#L1027-L1037) that generates the domain. That should also make Travis green. 😉

avichalbadaya:
@radeksimko : Awesome :) . Thanks ! Just made that change and it's green now.

stack72:
Hi @avichalbadaya and @radeksimko 

My apologies here - i should have looked at this in much more depth! (good spot @radeksimko!)

This LGTM now!

Paul

PR #7689, opened 2016-07-18
core: Convert context vars to map[string]interface{}

jen20:
This is the first step in allowing overrides of map and list variables.
We convert Context.variables to map[string]interface{} from
map[string]string and fix up all the call sites. There are a few areas
commented out behind a TODO for once go-atlas is updated to reflect the
new structures - these will need to be restored.

cc @phinze


phinze:
First pass LGTM

phinze:
👍 🚢

PR #7694, opened 2016-07-19
RabbitMQ Provider

jtopjian:
This commit adds support for a RabbitMQ provider. Included are the following resources:

* `rabbitmq_binding`
* `rabbitmq_exchange`
* `rabbitmq_permissions`
* `rabbitmq_policy`
* `rabbitmq_queue`
* `rabbitmq_user`
* `rabbitmq_vhost`

Importing of resources is also supported.

Docs included.

### DIFF
--- a/builtin/providers/rabbitmq/resource_binding.go
+++ b/builtin/providers/rabbitmq/resource_binding.go
@@ -1,0 +1,195 @@
+package rabbitmq
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/michaelklishin/rabbit-hole"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceBinding() *schema.Resource {
+	return &schema.Resource{
+		Create: CreateBinding,
+		Read:   ReadBinding,
+		Delete: DeleteBinding,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"source": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"vhost": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"destination": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"destination_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"properties_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"routing_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"arguments": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func CreateBinding(d *schema.ResourceData, meta interface{}) error {
+	rmqc := meta.(*rabbithole.Client)
+
+	vhost := d.Get("vhost").(string)
+	bindingInfo := rabbithole.BindingInfo{
+		Source:          d.Get("source").(string),
+		Destination:     d.Get("destination").(string),
+		DestinationType: d.Get("destination_type").(string),
+		RoutingKey:      d.Get("routing_key").(string),
+		PropertiesKey:   d.Get("properties_key").(string),
+		Arguments:       d.Get("arguments").(map[string]interface{}),
+	}
+
+	if err := declareBinding(rmqc, vhost, bindingInfo); err != nil {
+		return err
+	}
+
+	name := fmt.Sprintf("%s/%s/%s/%s/%s", vhost, bindingInfo.Source, bindingInfo.Destination, bindingInfo.DestinationType, bindingInfo.PropertiesKey)
+	d.SetId(name)
+
+	return ReadBinding(d, meta)
+}
+
+func ReadBinding(d *schema.ResourceData, meta interface{}) error {
+	rmqc := meta.(*rabbithole.Client)
+
+	bindingId := strings.Split(d.Id(), "/")
+	if len(bindingId) < 5 {
+		return fmt.Errorf("Unable to determine binding ID")
+	}
+
+	vhost := bindingId[0]
+	source := bindingId[1]
+	destination := bindingId[2]
+	destinationType := bindingId[3]
+	propertiesKey := bindingId[4]
+
+	bindings, err := rmqc.ListBindingsIn(vhost)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] RabbitMQ: Bindings retrieved: %#v", bindings)
+	bindingFound := false
+	for _, binding := range bindings {
+		if binding.Source == source && binding.Destination == destination && binding.DestinationType == destinationType && binding.PropertiesKey == propertiesKey {
+			log.Printf("[DEBUG] RabbitMQ: Found Binding: %#v", binding)
+			bindingFound = true
+
+			d.Set("vhost", binding.Vhost)
+			d.Set("source", binding.Source)
+			d.Set("destination", binding.Destination)
+			d.Set("destination_type", binding.DestinationType)
+			d.Set("routing_key", binding.RoutingKey)
+			d.Set("properties_key", binding.PropertiesKey)
+			d.Set("arguments", binding.Arguments)
+		}
+	}
+
+	// The binding could not be found,
+	// so consider it deleted and remove from state
+	if !bindingFound {
+		d.SetId("")
+	}
+
+	return nil
+}
+
+func DeleteBinding(d *schema.ResourceData, meta interface{}) error {
+	rmqc := meta.(*rabbithole.Client)
+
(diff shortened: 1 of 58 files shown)


builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/michaelklishin/rabbit-hole"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceBinding() *schema.Resource {
+	return &schema.Resource{
+		Create: CreateBinding,
+		Update: nil,
```
stack72: If all are ForceNew then we don't even need to include the Update part of the schema

builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+
+func CreateBinding(d *schema.ResourceData, meta interface{}) error {
+	rmqc := meta.(*rabbithole.Client)
+	vhost := d.Get("vhost").(string)
+
+	bindingInfo := rabbithole.BindingInfo{
+		Source:          d.Get("source").(string),
+		Destination:     d.Get("destination").(string),
+		DestinationType: d.Get("destination_type").(string),
+		RoutingKey:      d.Get("routing_key").(string),
```
stack72: As this is optional, if we send "", will this be an issue?

builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+		PropertiesKey:   d.Get("properties_key").(string),
+		Arguments:       d.Get("arguments").(map[string]interface{}),
```
stack72: Will sending an empty Map be an issue here?

builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if err := declareBinding(rmqc, vhost, bindingInfo); err != nil {
+		return err
+	}
+
+	name := fmt.Sprintf("%s/%s/%s/%s/%s", vhost, bindingInfo.Source, bindingInfo.Destination, bindingInfo.DestinationType, bindingInfo.PropertiesKey)
+	d.SetId(name)
```
stack72: Any reason why the ID is split with / ?

builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+	rmqc := meta.(*rabbithole.Client)
+	bindingParts := strings.Split(d.Id(), "/")
+	bindings, err := rmqc.ListBindingsIn(bindingParts[0])
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] RabbitMQ: Bindings retrieved: %#v", bindings)
+	for _, binding := range bindings {
+		if binding.Source == bindingParts[1] && binding.Destination == bindingParts[2] && binding.DestinationType == bindingParts[3] && binding.PropertiesKey == bindingParts[4] {
```
stack72: This feels kinda weird - is there not an easier way to guarantee we have the correct binding?

builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if err := declareBinding(rmqc, vhost, bindingInfo); err != nil {
+		return err
+	}
+
+	name := fmt.Sprintf("%s/%s/%s/%s/%s", vhost, bindingInfo.Source, bindingInfo.Destination, bindingInfo.DestinationType, bindingInfo.PropertiesKey)
+	d.SetId(name)
+
+	return nil
```
stack72: I think for consistency with other providers, Create should call Read

builtin/providers/rabbitmq/resource_binding.go
jtopjian: The / itself is just arbitrary - something that would not be used in any of the pieces that make up the ID.

In order to look up the binding, all of these pieces are required -- there's no single UUID that can pull this. So my thinking was to build a UUID with an arbitrary character that can be guaranteed to split each piece later.

Is there a better way to do this? It does feel clunky.

builtin/providers/rabbitmq/resource_binding.go
jtopjian: I don't believe so: https://github.com/michaelklishin/rabbit-hole/blob/master/bindings.go#L115

But if it's better to conditionally check each optional argument for a non-zero value, I'd be happy to refactor.

builtin/providers/rabbitmq/resource_binding.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func ReadBinding(d *schema.ResourceData, meta interface{}) error {
```
stack72: I think the Read should check for 404 and remove from state if not found - this will allow us to avoid errors when someone deletes a binding manually

builtin/providers/rabbitmq/resource_exchange.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/michaelklishin/rabbit-hole"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceExchange() *schema.Resource {
+	return &schema.Resource{
+		Create: CreateExchange,
+		Update: nil,
```
stack72: Again no need for Update to even be included

builtin/providers/rabbitmq/resource_exchange.go
```diff
@@ -0,0 +1,213 @@
...
+	d.SetId(id)
+
+	return ReadExchange(d, meta)
+}
+
+func ReadExchange(d *schema.ResourceData, meta interface{}) error {
+	rmqc := meta.(*rabbithole.Client)
+	exchParts := strings.Split(d.Id(), "@")
+
+	exchangeSettings, err := rmqc.GetExchange(exchParts[1], exchParts[0])
```
stack72: Can this return a 404? If so, we should refresh from state IMO

builtin/providers/rabbitmq/resource_exchange.go
```diff
@@ -0,0 +1,213 @@
...
+	e["durable"] = exchangeSettings.Durable
+	e["auto_delete"] = exchangeSettings.AutoDelete
+	e["arguments"] = exchangeSettings.Arguments
+	exchange[0] = e
+	d.Set("settings", exchange)
+
+	return nil
+}
+
+func UpdateExchange(d *schema.ResourceData, meta interface{}) error {
```
stack72: Not needed if everything is ForceNew :)

jtopjian:
@stack72 Definitely worthwhile to revisit this after 6 weeks - I found quite a few inconsistencies.

The resulting diff would have made everything messy, so I squashed each resource's changes into the individual commit.

Let me know if you have any questions and please feel free to go to town critiquing this -- at your convenience :)

builtin/providers/rabbitmq/resource_binding.go
jtopjian: confirmed this is OK.

builtin/providers/rabbitmq/resource_binding.go
jtopjian: confirmed this is OK

stack72:
Tests all pass :) 

```
% make testacc TEST=./builtin/providers/rabbitmq                                                                                                                                                        1 ↵
2016/09/01 17:29:23 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/rabbitmq -v  -timeout 120m
=== RUN   TestAccBinding_importBasic
--- PASS: TestAccBinding_importBasic (0.20s)
=== RUN   TestAccExchange_importBasic
--- PASS: TestAccExchange_importBasic (0.15s)
=== RUN   TestAccPermissions_importBasic
--- PASS: TestAccPermissions_importBasic (0.46s)
=== RUN   TestAccPolicy_importBasic
--- PASS: TestAccPolicy_importBasic (0.15s)
=== RUN   TestAccQueue_importBasic
--- PASS: TestAccQueue_importBasic (0.14s)
=== RUN   TestAccUser_importBasic
--- PASS: TestAccUser_importBasic (0.04s)
=== RUN   TestAccVhost_importBasic
--- PASS: TestAccVhost_importBasic (0.12s)
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccBinding
--- PASS: TestAccBinding (0.18s)
=== RUN   TestAccExchange
--- PASS: TestAccExchange (0.46s)
=== RUN   TestAccPermissions
--- PASS: TestAccPermissions (0.59s)
=== RUN   TestAccPolicy
--- PASS: TestAccPolicy (0.21s)
=== RUN   TestAccQueue
--- PASS: TestAccQueue (0.17s)
=== RUN   TestAccUser
--- PASS: TestAccUser (0.08s)
=== RUN   TestAccVhost
--- PASS: TestAccVhost (0.11s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/rabbitmq      	3.073s
```

I haven't tested this against an SSL setup cluster - but i did use a docker host for this:

```
alias devrabbitmq="docker run -d  -p 15672:15672 rabbitmq:3-management"
```

I think this is in fine shape to get going with :)

jtopjian:
> I haven't tested this against an SSL setup cluster

I did this morning and it works well. :)

Per our discussion, clarification about the ENDPOINT attribute: https://github.com/hashicorp/terraform/pull/7694/commits/d1e6f5c3dc98756b7705ca8db3ec90874e5bf5ab#diff-10ce1218e14de08884cd27718757e88bR48

Let me know how that reads :)

stack72:
LGTM! Merging now :)

PR #7709, opened 2016-07-19
Speed up WaitForState

dtolnay:
- Do not sleep twice before first attempt. There is already `time.Sleep(conf.Delay)` which controls the delay before the first check.
- Do not double the wait time when resource is already in the target state. Exponential backoff is for waiting for it to enter the target state.

### DIFF
--- a/helper/resource/state.go
+++ b/helper/resource/state.go
@@ -2,7 +2,6 @@ package resource
 
 import (
 	"log"
-	"math"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


dtolnay:
Before, `aws_kms_key` took a minimum of 2+4+8+10+10+10+10+10+10+10 = 84 seconds to delete, even when the deletion happened instantly. Now it will take 2*9 = 18 seconds.

jen20:
Hi @dtolnay! I've rebased this to account for the `PollInterval` change that @catsby made to support Elastic Beanstalk, and reopened this as #8370. Thanks for contributing to Terraform - this is a nice improvement!

PR #7710, opened 2016-07-20
core: Allow variable overrides to contain lists and maps

jen20:
This commit introduces the concept of lists and maps being parsed from HCL in TF_VAR_x overrides. Additional work is required to allow `-var` and objects from .tfvars files.

<del>Unit tests are required.</del>

### DIFF
--- a/command/flag_kv.go
+++ b/command/flag_kv.go
@@ -3,21 +3,46 @@ package command
 import (
 	"fmt"
 	"io/ioutil"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/hcl"
 	"github.com/mitchellh/go-homedir"
 )
 
-// FlagKV is a flag.Value implementation for parsing user variables
-// from the command-line in the format of '-var key=value'.
-type FlagKV map[string]string
+// FlagTypedKVis a flag.Value implementation for parsing user variables
+// from the command-line in the format of '-var key=value', where value is
+// a type intended for use as a Terraform variable
+type FlagTypedKV map[string]interface{}
 
-func (v *FlagKV) String() string {
+func (v *FlagTypedKV) String() string {
 	return ""
 }
 
-func (v *FlagKV) Set(raw string) error {
+func (v *FlagTypedKV) Set(raw string) error {
+	key, value, err := parseVarFlagAsHCL(raw)
+	if err != nil {
+		return err
+	}
+
+	if *v == nil {
+		*v = make(map[string]interface{})
+	}
+
+	(*v)[key] = value
+	return nil
+}
+
+// FlagStringKV is a flag.Value implementation for parsing user variables
+// from the command-line in the format of '-var key=value', where value is
+// only ever a primitive.
+type FlagStringKV map[string]string
+
+func (v *FlagStringKV) String() string {
+	return ""
+}
+
+func (v *FlagStringKV) Set(raw string) error {
 	idx := strings.Index(raw, "=")
 	if idx == -1 {
 		return fmt.Errorf("No '=' value in arg: %s", raw)
@@ -34,7 +59,7 @@ func (v *FlagKV) Set(raw string) error {
 
 // FlagKVFile is a flag.Value implementation for parsing user variables
 // from the command line in the form of files. i.e. '-var-file=foo'
-type FlagKVFile map[string]string
+type FlagKVFile map[string]interface{}
 
 func (v *FlagKVFile) String() string {
 	return ""
@@ -47,7 +72,7 @@ func (v *FlagKVFile) Set(raw string) error {
 	}
 
 	if *v == nil {
-		*v = make(map[string]string)
+		*v = make(map[string]interface{})
 	}
 
 	for key, value := range vs {
@@ -57,7 +82,7 @@ func (v *FlagKVFile) Set(raw string) error {
 	return nil
 }
 
-func loadKVFile(rawPath string) (map[string]string, error) {
+func loadKVFile(rawPath string) (map[string]interface{}, error) {
 	path, err := homedir.Expand(rawPath)
 	if err != nil {
 		return nil, fmt.Errorf(
@@ -78,7 +103,7 @@ func loadKVFile(rawPath string) (map[string]string, error) {
 			"Error parsing %s: %s", path, err)
 	}
 
-	var result map[string]string
+	var result map[string]interface{}
 	if err := hcl.DecodeObject(&result, obj); err != nil {
 		return nil, fmt.Errorf(
 			"Error decoding Terraform vars file: %s\n\n"+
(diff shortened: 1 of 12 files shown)


terraform/context.go
```diff
@@ -76,6 +77,23 @@ type Context struct {
 	runCh               <-chan struct{}
 }
 
+func parseVariableAsHCL(input string) (interface{}, error) {
+	const sentinalValue = "SENTINAL_TERRAFORM_VAR_OVERRIDE_KEY"
```
apparentlymart: Sorry for this total nit-pick, but the correct spelling of this word is "sentinel".

jen20: Indeed it is :+1:

terraform/context.go
```diff
@@ -119,24 +120,112 @@ func NewContext(opts *ContextOpts) (*Context, error) {
...
-		variables[k] = v
-	}
-	for k, v := range opts.Variables {
-		variables[k] = v
+			// Strip off the prefix and get the value after the first "="
+			idx := strings.Index(v, "=")
+			k := v[len(VarEnvPrefix):idx]
+			v = v[idx+1:]
+
+			// Override the command line set variable. Note that *not* finding the variable
```
jbardin: This is overriding config, not command line, correct?

jen20: Correct, I'll fix this now

jbardin:
LGTM, but feel free to wait for phinze to check it out

jen20:
:+1: probably a significant enough set of changes for @phinze to want to take a look here

jen20:
I've made some fairly significant modifications to this to support `-var-file` and simplify the implementation, which need additional test coverage, so I'm going to add the `[WIP]` flag back.

command/flag_kv.go
```diff
@@ -3,21 +3,46 @@ package command
...
 	"strings"
 
 	"github.com/hashicorp/hcl"
 	"github.com/mitchellh/go-homedir"
 )
 
-// FlagKV is a flag.Value implementation for parsing user variables
-// from the command-line in the format of '-var key=value'.
-type FlagKV map[string]string
+// FlagKVString is a flag.Value implementation for parsing user variables
```
phinze: s/FlagKVString/FlagTypedKV/

command/flag_kv.go
```diff
@@ -3,21 +3,46 @@ package command
...
+// from the command-line in the format of '-var key=value', where value is
+// a type intended for use as a Terraform variable
+type FlagTypedKV map[string]interface{}
 
-func (v *FlagKV) String() string {
+func (v *FlagTypedKV) String() string {
 	return ""
```
phinze: This seems a bit insidious. Is there a functional reason for the empty Stringer impl?

command/flag_kv.go
```diff
@@ -3,21 +3,46 @@ package command
...
 }
 
-func (v *FlagKV) Set(raw string) error {
+func (v *FlagTypedKV) Set(raw string) error {
+	key, value, err := parseVarFlagAsHCL(raw)
+	if err != nil {
+		return err
+	}
+
+	if *v == nil {
```
phinze: is this equivalent to `v == nil`?

command/flag_kv.go
```diff
@@ -3,21 +3,46 @@ package command
...
+		*v = make(map[string]interface{})
+	}
+
+	(*v)[key] = value
+	return nil
+}
+
+// FlagKVString is a flag.Value implementation for parsing user variables
```
phinze: s/FlagKVString/FlagStringKV/

command/flag_kv.go
```diff
@@ -103,3 +128,44 @@ func (v *FlagStringSlice) Set(raw string) error {
 
 	return nil
 }
+
+// parseVariableAsHCL parses the value of a single variable as would have been specified
```
phinze: s/parseVariableAsHCL/parseVarFlagAsHCL/

command/flag_kv.go
jen20: This is following the pattern set in `FlagKV` (as was) - I don't know if there is a good reason for it.

command/flag_kv_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func TestFlagKV_impl(t *testing.T) {
-	var _ flag.Value = new(FlagKV)
+	var _ flag.Value = new(FlagStringKV)
 }
 
 func TestFlagKV(t *testing.T) {
```
phinze: Should update test names to match the func name.

phinze: Probably want some unit tests around `FlagTypedKV` here as well

command/meta.go
```diff
@@ -36,9 +36,9 @@ type Meta struct {
 
 	// Variables for the context (private)
 	autoKey       string
-	autoVariables map[string]string
+	autoVariables map[string]interface{}
```
phinze: Did you figure out what these things were? 😆

command/flag_kv.go
jen20: Need to verify there - I suspect the receiver can be nil while the underlying map has not been constructed. Again, this came from the original `FlagKV` implementation.

terraform/context.go
```diff
@@ -119,24 +120,106 @@ func NewContext(opts *ContextOpts) (*Context, error) {
 		par = 10
 	}
 
-	// Setup the variables. We first take the variables given to us.
-	// We then merge in the variables set in the environment.
+	// Set up the variables in the following sequence:
+	//    0 - Take default values from the configuration
+	//    1 - Take values from TF_VAR_x environment variables
+	//    2 - Take values specified in -var flags, overriding values
+	//        set by environment variables if necessary.
```
phinze: (2) comprises both `-var` and `-var-file` variables yes?

jen20: Yes

command/meta.go
jen20: Yes - these are variables loaded from `terraform.tfvars` irrespective of whether a `-var-file` is passed.

terraform/context.go
```diff
@@ -119,24 +120,106 @@ func NewContext(opts *ContextOpts) (*Context, error) {
...
+					case string:
+						if typedDefault == "" {
+							continue
+						}
+						variables[v.Name] = typedDefault
+					case int:
+						variables[v.Name] = fmt.Sprintf("%d", typedDefault)
+					case bool:
+						variables[v.Name] = fmt.Sprintf("%t", typedDefault)
+					}
```
phinze: Ah this is sadness. Totally understand why it is necessary. There might be some edge cases here (do big ints become int64? are floats valid?)

jen20: I _think_ we only support ints < int64 in HCL, but will verfy. It might be better to use reflection and kind categories here. Alternatively we could add all the various types that could show up to the type switch.

command/flag_kv.go
phinze: Ah OK I bet the flags package uses this somehow.

command/flag_kv.go
phinze: We chatted about this in Slack. I'm 👌

jen20: `v` could be pointing to a map variable declared, without the map ever having been constructed - I don't think these are equivalent.

terraform/context.go
jen20: I've added an additional comment here to clarify.

command/flag_kv_test.go
```diff
@@ -62,6 +63,86 @@ func TestFlagKV(t *testing.T) {
...
+			"key=foo=bar",
+			map[string]interface{}{"key": "foo=bar"},
+			false,
+		},
+
+		{
+			"map.key=foo",
+			map[string]interface{}{"map.key": "foo"},
+			false,
+		},
```
phinze: So this is just a string variable called "map.key" - not actually a dot-notation set map key?

phinze:
Looking good! It really is surprising how little code this took. Nice work 👍

command/flag_kv_test.go
jen20: Yes - this would be a good point to inject a better error message around overriding map keys, actually.

PR #7723, opened 2016-07-20
Add tests for maps with dots

jbardin:
This adds some unit tests for config maps with dots in the key values.
We check for maps with keys which have overlapping names. There are
however still issues with nested maps which create overlapping flattened
names, as well as nested lists with dots in the key.


stack72:
LGTM! As it's a set of tests, if it builds, then it works :)

PR #7726, opened 2016-07-20
Explain domain and name parameters.

goloroden:


stack72:
Hi @goloroden 

Thanks so much for this PR! If possible, please can we change this to follow the note notation (e.g. 

```
~> **NOTE on Route Tables and Routes:** Terraform currently
provides both a standalone Route resource and a [Route Table](route_table.html) resource with routes
defined in-line. At this time you cannot use a Route Table with in-line routes
in conjunction with any Route resources. Doing so will cause
a conflict of rule settings and will overwrite rules.
```

Also, please can you provide an example of what you are describing as part of that note?

Thanks

Paul

stack72:
ping @goloroden :)

goloroden:
@stack72 - sorry for not having answered yet, I'm currently missing the time to handle this. But I'll have a look at it again at some point in the future!

stack72:
No worries at all @goloroden 

If you want, you can drop a short explanation here and I can update the PR on your behalf :)

P.

goloroden:
The point here is simply that the current naming does not explain that it's about the domain and the subdomain. Instead it sounds as if you should provide the FQDN to the field `domain` (which of course then leaves you wondering what to enter into `name`).

It took me a while to figure out that if you want to use the domain `foo.bar.com` you have to enter `foo` as the name, and `bar.com` as the domain. Hence the note I wanted to add.

stack72:
Hi @goloroden 

just a FYI, I actually changed the examples here a little:

```
## Example Usage

# Add a record to the root domain
resource "dnsimple_record" "foobar" {
	domain = "${var.dnsimple_domain}"
	name = ""
	value = "192.168.0.11"
	type = "A"
	ttl = 3600
}

# Add a record to a sub-domain
resource "dnsimple_record" "foobar" {
	domain = "${var.dnsimple_domain}"
	name = "terraform"
	value = "192.168.0.11"
	type = "A"
	ttl = 3600
}
```

stack72:
I merged this manually in https://github.com/hashicorp/terraform/commit/3392b3a76462eb3d3116bdfde96344be7d649596

PR #7731, opened 2016-07-20
Fix race condition in resource.Retry

dtolnay:
If `WaitForState()` returns from a timeout and `f()` finishes at the same time, the read of `resultErr` after `WaitForState()` races with the write of `resultErr` after `f()`.


dtolnay:
@radeksimko is there anything I can do to move this along?

jen20:
Hi @dtolnay! I was just looking through this, it appears the same fix was caught in a sweep of race conditions that @jbardin did towards the end of July (in 074be9). Sorry this wasn't reviewed quicker and merged - but you'll no doubt be glad to know that the fix has been in at least two releases now!

PR #7732, opened 2016-07-20
Fix bad message from TimeoutError

dtolnay:
Before:

> Error creating IAM Role my-role: timeout while waiting for state to become 'success'. last error: %!s(\<nil\>)

### DIFF
--- a/helper/resource/error.go
+++ b/helper/resource/error.go
@@ -44,9 +44,10 @@ type TimeoutError struct {
 }
 
 func (e *TimeoutError) Error() string {
-	return fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


catsby:
Thanks @dtolnay – do you have any clues as to why that last error is `nil`?

dtolnay:
Yes, it should be fixed in #7733. It means the first try never finished.

dtolnay:
@catsby is there anything I can do to move this along?

jen20:
Hi @dtolnay, I rebased and merged this - thanks for your contribution to Terraform!

PR #7733, opened 2016-07-20
provider/aws: Increase timeout for creating IAM role

dtolnay:
I am seeing `terraform apply` fail with:

> Error creating IAM Role my-role: timeout while waiting for state to become 'success'. last error: %!s(<nil>)

(The message is fixed in #7732.)

All subsequent `terraform apply`s fail with:

> Error creating IAM Role my-role: EntityAlreadyExists: Role with name my-role already exists.

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_role.go
+++ b/builtin/providers/aws/resource_aws_iam_role.go
@@ -105,10 +105,10 @@ func resourceAwsIamRoleCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	var createResp *iam.CreateRoleOutput
-	err := resource.Retry(10*time.Second, func() *resource.RetryError {
+	err := resource.Retry(30*time.Second, func() *resource.RetryError {
 		var err error
 		createResp, err = iamconn.CreateRole(request)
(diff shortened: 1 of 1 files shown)


dtolnay:
@stack72 is there anything I can do to move this along?

radeksimko:
Hi @dtolnay 
thanks for the PR.

I took some time to read through the git log and related issues to understand why we need the retry logic here in the 1st place, because I was pretty sure this is not the reason:

```
// IAM roles can take ~30 seconds to propagate in AWS:
```

Even though **the statement is true on its own**, it's not why we'd retry `iam:CreateRole`, this _would be_ a reason for retrying creation of any resource that depends on this IAM Role (e.g. EC2 instance, ECS service, etc.).

The reason we retry `iam:CreateRole` here is because there may be other IAM resources, e.g. IAM Users referenced in the Role assume policy and these **IAM Users take time to propagate**.

IAM Role with assume policy that doesn't reference any other IAM resource should not require any retry (ain't saying we should put it away though).

It may all sound like nitpicking a 1-2 line comment, but I felt the need for explaining this as I was confused by that one. 😃 
I will update the comment in a separate PR.

PR #7734, opened 2016-07-20
provider/aws: pull iamconn setup earlier

phinze:
Fixes problem introduced in re-arrangement of config

Refs #7672


stack72:
Verified locally - will merge on green build!

PR #7742, opened 2016-07-21
vSphere provider: refactor resource_vsphere_virtual_machine.go

ChloeTigre:
This file is huge.

It implements a lot of features but its lack of readability impairs the possibility of evolutions.

Hence I refactored it.

I think we should try to achieve smaller functions and more readable code.

I will add new features from the result of this merge.


ChloeTigre:
Sorry for the noise. I needed to fix my code generators which were bogus.

PR #7747, opened 2016-07-21
provider/fastly: Update go-fastly SDK

catsby:
Update the go-fastly SDK to the latest version, fixing a bug when using `ignore_if_set`. b0022f57de5ab7f324150ac1dc88355c722917f7 adds `ignore_if_set` to an existing test, and will fail on `master` (at time of writing).


catsby:
Tried spinning up a custom TravisCI run for these tests, but TravisCI is having a blip, so I'm running them manual. Will update when they finish

catsby:
The tests pass and besides the dependency update, it's a minor change, so I'm going to pull this in.

PR #7749, opened 2016-07-21
Update vmware-govcd plugin

robinbowes:
This PR updates the vendored vmware-govcd plugin

It passes all tests and builds for me (on OSX).


stack72:
Hi @robinbowes 

Do you still need this merged?

Paul

robinbowes:
Well, I'm not currently working on either terraform or vcloud, but that's no reason to not keep the vcloud provider up-to-date :)

stack72:
Got it! I am looking through the back log to see what is / isn't needed anymore

Thanks for the info

stack72:
Closing as there is no activity on this - thanks for all the work here

robinbowes:
I'm not clear why you've not merged this?

You're vendoring the plugin; this MR simply updates the plugin to a more recent version.

stack72:
Hi @robinbowes 

There is a merge conflict and we are not in a position to be able to test the VCD configuration. I can easily fix the conflict, but, as this is not a core provider, if there are issues with it, then it could cause issues for the userbase

We want to mitigate that as much as possible. If you really need this to be merged then we can do that 

Paul

PR #7750, opened 2016-07-21
Removed azurerm_network_interface from scale set example

gambrose:
Azure scale sets encapsulates the creation of network interfaces so it is not valid in this example.

I'm also think there maybe a issue with some of the parameter docs in the example they are prefixed with 'virtual_machine_' but not descriptions below the example. I was going to fix this as well but wasn't sure which was correct. The tests use the prefix but the schema in code doesn't appear to?


stack72:
Hi @gambrose 

You are correct - this shouldn't be in the example :) Thanks for finding this

Paul

PR #7755, opened 2016-07-21
Fix endpoint documentation for aws_rds_cluster

WolverineFan:
Sorry for all the whitespace changes.  Atom automatically removes whitespace at the end of lines.  I can create a different patch if it's too annoying.


catsby:
LGTM, thanks!

PR #7756, opened 2016-07-21
core: send version info to atlas

jbardin:
Set a default `Terraform-Version` header in atlas requests.

Also adds a helper function to build the version string, since we were formatting that in various places throughout the code.

### DIFF
--- a/builtin/providers/azurerm/config.go
+++ b/builtin/providers/azurerm/config.go
@@ -92,13 +92,7 @@ func withRequestLogging() autorest.SendDecorator {
 }
 
 func setUserAgent(client *autorest.Client) {
-	var version string
-	if terraform.VersionPrerelease != "" {
-		version = fmt.Sprintf("%s-%s", terraform.Version, terraform.VersionPrerelease)
-	} else {
-		version = terraform.Version
-	}
-
+	version := terraform.VersionString()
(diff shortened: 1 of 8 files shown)


radeksimko:
I'm wondering if there's any way to incorporate this new helper function into the AWS provider too...
https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/config.go#L321-L327

Maybe the AWS SDK interface doesn't allow it, in which case feel free to ignore me... but it would be nice to keep this consistent.

jbardin:
Yes, I looked at that, but I mistakingly looked at the function definition and thought it was in our code and required further refactoring. I think it's probably fine to use the same string there too. I'll update the PR.

builtin/providers/azurerm/config.go
```diff
@@ -92,13 +92,7 @@ func withRequestLogging() autorest.SendDecorator {
...
 
 func setUserAgent(client *autorest.Client) {
-	var version string
-	if terraform.VersionPrerelease != "" {
-		version = fmt.Sprintf("%s-%s", terraform.Version, terraform.VersionPrerelease)
-	} else {
-		version = terraform.Version
-	}
-
+	version := terraform.VersionString()
```
stack72: Love that you added it to AzureRM as well - thanks!

stack72:
LGTM! nice one :)

PR #7766, opened 2016-07-21
provider/aws: Delete access keys before deleting IAM user

dtolnay:
Without this change:

```
Error deleting IAM User my-user: DeleteConflict: Cannot delete entity, must delete access keys first.
```

We add access keys outside of Terraform because we do not want the secret keys in the tfstate.

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_user.go
+++ b/builtin/providers/aws/resource_aws_iam_user.go
@@ -157,25 +157,6 @@ func resourceAwsIamUserDelete(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return fmt.Errorf("Error removing user %q from all groups: %s", d.Id(), err)
 	}
-
-	// All access keys for the user must be removed
-	var accessKeys []string
-	if d.Get("force_destroy").(bool) {
-		listAccessKeys := &iam.ListAccessKeysInput{
-			UserName: aws.String(d.Id()),
-		}
-		pageOfAccessKeys := func(page *iam.ListAccessKeysOutput, lastPage bool) (shouldContinue bool) {
-			for _, g := range page.AccessKeyMetadata {
-				accessKeys = append(accessKeys, *g.AccessKeyId)
-			}
-			return true
-		}
-		err = iamconn.ListAccessKeysPages(listAccessKeys, pageOfAccessKeys)
-		if err != nil {
-			return fmt.Errorf("Error removing access keys of user %s: %s", d.Id(), err)
-		}
-	}
-
 	for _, g := range groups {
 		// use iam group membership func to remove user from all groups
 		log.Printf("[DEBUG] Removing IAM User %s from IAM Group %s", d.Id(), g)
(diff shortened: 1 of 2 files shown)


radeksimko:
Hi @dtolnay 
thank you for the PR. I understand the intentions here - cleaning up manually created keys. That's a valid use case 👍 

One of the core Terraform's principles is to *not* destroy/change things it doesn't manage however. It's sometimes hard to decide what is actually being managed - i.e. just user or user *and* their keys, but if keys can be managed separately via Terraform I think it's very clear that it should be that resource responsible for management of keys:
https://www.terraform.io/docs/providers/aws/r/iam_access_key.html

The current behaviour with the current schema is as intended - i.e. Terraform _should_ error out if there are existing access keys that are managed outside of Terraform. It is possible that they just simply manage the keys elsewhere and they forgot to delete those - these keys may potentially be in use and the user may want to verify that.

We can however let users **opt in** to the destructive behaviour via special field, e.g. `force_destroy = true` (defaults to `false`). See how we do this in [`aws_s3_bucket`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_s3_bucket.go#L291)

Would you mind making the behaviour opt-in - i.e. explicit rather than implicit?

----

I saw that piece of code responsible for detaching the user from IAM Groups - and I suggest we treat that as a bug which will need to be fixed at some point. Generally speaking if Terraform is deleting/changing something outside the scope of a given resource, it's most likely (exceptions apply) a bug that should be reported and/or fixed.

dtolnay:
Makes sense, thanks for the feedback. I added a `force_destroy` option to control removal of non-Terraform-managed access keys.

builtin/providers/aws/resource_aws_iam_user.go
```diff
@@ -173,6 +184,16 @@ func resourceAwsIamUserDelete(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	for _, k := range accessKeys {
```
radeksimko: Would you mind moving this for loop along with `var accessKeys []string` into the `force_destroy` condition too?
It IMO makes sense from the context point of view.

dtolnay: :+1: done

builtin/providers/aws/resource_aws_iam_user.go
```diff
@@ -132,39 +138,25 @@ func resourceAwsIamUserUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		for _, g := range page.Groups {
 			groups = append(groups, *g.GroupName)
 		}
-
-		// if there's a marker present, we need to save it for pagination
-		if r.Marker != nil {
-			*marker = *r.Marker
-		}
-		*truncated = *r.IsTruncated
+		return true
```
radeksimko: Isn't this going to loop forever if `shouldContinue = true`? Shouldn't this be something like

```go
if !lastPage {
  return true
}
```

The previous implementation used `ListGroupsForUserOutput.IsTruncated` to make decisions when paginating. I don't mind changing the implementation as long as the behaviour remains the same. Can you confirm the behaviour is exactly the same?

dtolnay: I did confirm that the ListAccessKeysPages function is smart about not looping forever. My account is limited to 2 access keys per user so I set MaxItems to 1, changed this line to `return false` and confirmed that only 1 access key was deleted, then changed this line to `return true` and confirmed that both access keys were deleted (and it did not loop forever).

I have not tested ListGroupsForUserPages specifically but it goes through the same codepath. I can test later today or I can change it back if you prefer.

The `shouldContinue` is really only for iterating over a limited number of pages or searching for something until you find it.

radeksimko:
I will have a look into how the pagination works in AWS SDK to verify it works as expected and we don't end up in endless loop or something like that. 😃  Your implementation may be completely correct - I just haven't seen that one anywhere else.

Otherwise this is looking pretty good - just one small thing I'd like to ask you for - can you document this new field in https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/r/iam_user.html.markdown ?

Thanks.

builtin/providers/aws/resource_aws_iam_user.go
dtolnay: I changed both to `return !lastPage` for clarity.

dtolnay:
I added `force_destroy` to the website.

radeksimko:
I have looked at the implementation of [EachPage function in the SDK](https://github.com/aws/aws-sdk-go/blob/master/aws/request/request_pagination.go#L80-L104) to confirm that our assumptions were correct here. It also explains why `return true` (i.e. `shouldContinue = true`) won't cause endless loop. That function won't be called at all if there are no pages to iterate over. That said `return !isLastPage` is IMO cleaner.

Thanks. 👍

PR #7771, opened 2016-07-22
provider/aws: `aws_redshift_cluster` `number_of_nodes` was having the wrong value set to state

stack72:
we used `len(ClusterNodes)` rather than NumberOfNodes :)

This was picked up by the nightly tests! <3


catsby:
LGTM!

It's a bit baffling why these two things don't produce the same result :/

PR #7777, opened 2016-07-22
provider/aws: Fix issue updating ElasticBeanstalk Environment Settings

### LINKED ISSUE #6890: Removing Elastic Beanstalk Environment Variables
On terraform 0.6.16

If I remove a `setting` with the namespace of `aws:elasticbeanstalk:application:environment`, terraform says its removed, however the AWS console still has that setting, as well as the API still returns a value for that setting.

Also, if I add a environment variable manually in the AWS console and then re-run a plan, terraform doesn't seem to see that value and try to remove it.

It looks like when terraform refreshes its state from AWS, it doesn't properly catalog `aws:elasticbeanstalk:application:environment` keys, certainly not ensuring that there are not extra or missing keys.

catsby:
Fixes the logic that updated settings for Elastic Beanstalk Environments.
Because the update is done in the same API call, we need to split removals /
additions.

Fixes #6890

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
@@ -328,39 +328,50 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
 		os := o.(*schema.Set)
 		ns := n.(*schema.Set)
 
-                rm := extractOptionSettings(os.Difference(ns))
-                add := extractOptionSettings(ns.Difference(os))
+		rm := extractOptionSettings(os.Difference(ns))
+		add := extractOptionSettings(ns.Difference(os))
 
-                // Additions and removals of options are done in a single API call, so we
-                // can't do our normal "remove these" and then later "add these", re-adding
-                // any updated settings.
-                // Because of this, we need to remove any settings in the "removable"
-                // settings that are also found in the "add" settings, otherwise they
-                // conflict. Here we loop through all the initial removables from the set
-                // difference, and we build up a slice of settings not found in the "add"
-                // set
-                var remove []*elasticbeanstalk.ConfigurationOptionSetting
-                if len(add) > 0 {
-                        for _, r := range rm {
-                                for _, a := range add {
-                                        if *r.Namespace == *a.Namespace && *r.OptionName == *a.OptionName {
-                                                continue
-                                        }
-                                        remove = append(remove, r)
-                                }
-                        }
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -327,7 +326,39 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
...
+		// Because of this, we need to remove any settings in the "removable"
+		// settings that are also found in the "add" settings, otherwise they
+		// conflict. Here we loop through all the initial removables from the set
+		// difference, and we build up a slice of settings not found in the "add"
+		// set
+		var remove []*elasticbeanstalk.ConfigurationOptionSetting
+		if len(add) > 0 {
+			for _, r := range rm {
+				for _, a := range add {
+					if *r.Namespace == *a.Namespace && *r.OptionName == *a.OptionName {
```
dharrisio: @catsby do we need to take into account the ResourceName setting attribute here as well? I have a feeling this may not work as intended with more than one scheduled action (http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options-general.html#command-options-general-autoscalingscheduledaction)

This attribute can be a bit frustrating, because it is only documented for scheduled actions, but is returned by the api for some other settings. To avoid the  issues that caused I had to do this https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go#L717

catsby: In the statefile, we ignore `ResourceName` unless it's a scheduled action. I think here I'll check if the local setting has `ResourceName` and then check if they match, continuing if they do not match. If we don't have it set locally, I'll just continue on as we have here. Does that sound reasonable?

catsby: I added a check in https://github.com/hashicorp/terraform/pull/7777/commits/0cec643907e2c3f0ab30514422b5efdaeb2f6efe that I think covers this

catsby:
updated with acc test

builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
dharrisio: @catsby Sorry for the delayed response here, been out for a bit. This sounds good to me!

Thanks!

stack72:
LGTM!

PR #6956, opened 2016-05-31
provider/aws: Fixing IAM data source policy generation to prevent spurious diffs

vancluever:
Just re-submitting this as a new PR, this was "submitted" in #6881.

As mentioned there, there are a number of things AWS likes to do with a policy (namely resource policies versus IAM ones), that cause spurious diffs in Terraform even though nothing has changed, ie: sorting arrays, inserting an empty `Sid` attribute, and converting single-element lists to a string value.

Having the data source affords us the opportunity to ensure that these problems do not persist as we can now tweak the policy generation in one place.

Again, an example of the issue in action:

https://gist.github.com/vancluever/526e9c1d8153436e58244847b8fb20f1

The changes in this commit fixes things to allow a policy generated via the `aws_iam_policy_document` data source to properly pass a TF plan/apply without any changes when nothing has indeed changed.

References for this: I list a few here, however if you search in the issue backlog there are more:

S3: #6952
SNS: #6909
S3, discussing ordering: #5978
ElasticSearch: #5067
My original PR: #4278

PR #7785, opened 2016-07-23
provider/aws: Normalize IAM policy documents

### LINKED ISSUE #3634: aws_elasticsearch_domain access_policies keeps reapplying
I have 

```terraform
resource "aws_elasticsearch_domain" "es" {
  domain_name = "test"
  advanced_options {
    "rest.action.multi.allow_explicit_index" = true
  }
  access_policies = <<EOF
{
  "Statement": [
    {
      "Action": "es:*",
      "Condition": {
        "IpAddress": {
          "aws:SourceIp": [
            "${aws_instance.nat_a.public_ip}",
            "${aws_instance.nat_c.public_ip}"
          ]
        }
      },
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Resource": "arn:aws:es:${var.AWS_REGION}:${var.AWS_ACCOUNT_ID}:domain/test/*",
      "Sid": "ES"
    }
  ],
  "Version": "2012-10-17"
}
EOF
  ebs_options {
    ebs_enabled = true
    volume_type = "gp2"
    volume_size = 35
  }
  cluster_config {
    instance_type = "t2.small.elasticsearch"
    instance_count = 2
    dedicated_master_enabled = false
    zone_awareness_enabled = true
  }
  snapshot_options {
    automated_snapshot_start_hour = 23
  }
}
```

The access_policies section keeps reapplying on every run. I can verify in AWS that the correct policy is applied. I suspect it is this line https://github.com/hashicorp/terraform/compare/master...lancehudson:patch-1

### LINKED ISSUE #7495: aws_iam_policy_document data source diff with single element arrays
### Terraform Version

### Affected Resource(s)
- aws_iam_policy_document

```hcl
data "aws_iam_policy_document" "key_store_kms" {
  statement {
    actions = [
      "kms:*"
    ]
    resources = [
      "*"
    ]
    effect = "Allow"
    principals {
      type = "AWS"
      identifiers = ["arn:aws:iam::0958430598340:root"]
    }
  }
}

resource "aws_kms_key" "key_store" {
  description = "Protects secrets stored in S3"
  policy = "${data.aws_iam_policy_document.key_store_kms.json}"
}
```

### Expected Behavior
The policy shouldn't be marked to be modified in the plan phase. All the arguments that received a [] with only one element, should be rendered as a string.

### Actual Behavior
The policy is marked to be modified every time I run terraform plan and apply. 

### Steps to Reproduce
1. `terraform plan`
1. `terraform apply`
1. `terraform plan`

### References
- [GH-1234](https://github.com/hashicorp/terraform/pull/6881#issuecomment-222390832)
The is a mention about the problem but it's hidden in a merge PR.
> There needs to be some logic to ensure that array items like Actions get converted to strings instead of []strings if there is only one element (this was one of the things that was causing recurring diffs in unchanged policies). So those elements probably need to be interface{} and then handled accordingly.

dtolnay:
Fixes #3634 and fixes #7495.

Two normalizations are applied:

- Arrays of length 1 become single strings because this is how they are returned from AWS APIs.
- Arrays of length more than 1 are sorted because AWS stores them as sets and returns them in random order.


builtin/providers/aws/json.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func normalizePolicyDocument(policyString interface{}) string {
+	if policyString == nil || policyString == "" {
+		return ""
+	}
+	var policy IAMPolicyDoc
+	err := json.Unmarshal([]byte(policyString.(string)), &policy)
+	if err != nil {
+		return fmt.Sprintf("Error parsing IAM policy document: %s", err)
```
apparentlymart: This doesn't feel right... seems like if an error occurs here we'd end up storing this error message in the state, which I don't think is what you intended here.

Probably the best we could do here is to make this function silently change nothing in the event of an error (just return the provided string verbatim) and trust the backend to catch and return any syntax errors. I suppose we could couple this with a policy validation function that we could use as `ValidateFunc` on these attributes and thus catch certain errors locally during Terraform validation, and thus assume that the policy will always be valid by the time it gets in here.

Also wondering about the rationale for the `policyString` argument being `interface{}` rather than `string` here... it seems like this code would panic at runtime if the policy were anything other than a string, so could we just make the argument be of type `string` in the first place?

builtin/providers/aws/json.go
```diff
@@ -0,0 +1,38 @@
...
+		return fmt.Sprintf("Error parsing JSON: %s", err)
+	}
+	b, err := json.Marshal(j)
+	if err != nil {
+		panic(err) // unexpected
+	}
+	return string(b)
+}
+
+func normalizePolicyDocument(policyString interface{}) string {
```
apparentlymart: This is not super-important but I feel like this function may be more at home in the `iam_policy_model.go` file rather than the `json.go` file... that's where we're already doing all of the other work related to marshaling and unmarshaling IAM policy JSON documents in particular.

builtin/providers/aws/json.go
dtolnay: Re: returning an error. This matches the behavior of the existing `normalizeJson` function. I can file a separate ticket to fix both.

Re: `interface{}`. This allows `normalizePolicyDocument` to be used as a [`StateFunc`](https://github.com/hashicorp/terraform/blob/57cf9fd2953192f274731bfe4d38ea4e39e74270/helper/schema/schema.go#L194).

builtin/providers/aws/json.go
dtolnay: If in the future you make a change to IAMPolicyDoc, you almost certainly do not need to make a change to `normalizePolicyDocument`. If in the future you make a change to `normalizeJson`, you almost certainly also want to make a change to `normalizePolicyDocument`. That is why I organized the files this way.

radeksimko: > If in the future you make a change to IAMPolicyDoc, you almost certainly do not need to make a change to `normalizePolicyDocument`

Agreed.

> If in the future you make a change to `normalizeJson`, you almost certainly also want to make a change to `normalizePolicyDocument`.

Why do you think so? `normalizeJson` is a generic function unrelated to IAM, it can be used to normalize CloudFormation templates for example (JSON).

dtolnay: Well, one example is the change you suggested in your other comment - changing the error handling behavior.

dtolnay: *not you, @apparentlymart suggested it

apparentlymart:
Thanks for this @dtolnay! Apart from a couple minor comments this looks great to me.

builtin/providers/aws/json.go
apparentlymart: Ahh got it... I didn't make the connection that you were using this directly as a `StateFunc`. Sorry... should've woken up a bit more before asking that question. :grinning:

dtolnay:
Thanks for the review @apparentlymart. Can you take a look at my response to the comments? I don't think any of them should be addressed in this PR.

builtin/providers/aws/json.go
apparentlymart: Oh yeah... agreed that it's weird for `normalizeJson` to be doing it too, but we can certainly fix that separately.

ronaldtse:
This is an awesome fix that makes the iam_policy_document data source actually useable. Right now it's pretty much breaking all AWS Elasticsearch and S3 bucket policies.

As an avid terraform user, may I ask that this gets merged soon? 👍

aerickson:
@apparentlymart, can you please respond to @dtolnay's comments?

Sorry for the ping... this is a pretty nasty 'bug' that's affecting me (every apply now takes 15 minutes in my environment).

sstarcher:
@aerickson agreed it's very painful with AWS ES as it takes a very long time to apply

apparentlymart:
@dtolnay sorry for sitting on this one for so long. It looks like in the mean time a comparable change was submitted and merged in #6956.

Apparently @vancluever took the opposite approach there: rather than normalizing what we send to AWS, instead we "de-normalize" what we retrieve *from* AWS. This addresses the part of the problem where we get spurious diffs, but only addressed `aws_iam_policy_document` itself and didn't deal with all the various resource arguments where IAM policies can be specified.

I think perhaps the path of least resistance to get from where `master` is to what you wanted to achieve here would be for us to to split out the "de-normalization" part into a separate function that can be used as a `StateFunc`, and then we can use it in all the places you identified in your patch here... so it would have a similar result to your change but with Terraform's "normalized form" being arrays-always, rather than having the special case for single items as your patch would've done. I don't think it will matter which we we go with the normalization as long as we're always consistent.

I'm going to work on that now, using your changeset here as a prompt for which attributes need to be normalized.

dtolnay:
Another option is to revert #6956 and merge this. As far as I can tell, this is strictly better. Is there any advantage to the fix in #6956?

I am on vacation this week and won't be able to make changes until 8/27 so I will leave it to you.

dtolnay:
@apparentlymart I pushed a revert + merge.

```
git checkout -b revert origin/master
git revert 9872026235b094ada8ef017cae9db2a1c0cac399
git checkout dtolnay/normalize
git merge revert
```

vancluever:
@dtolnay it's disappointing to me to see you say that that work in #6956 should just be "reverted" - judging from other feedback that I've read in referenced issues, there does not appear to be any regressions - just that the only thing that I fixed was the data source. Which seemed fair enough in my opinion - I didn't really think that one could expect Terraform to be able to properly normalize all policy documents that were sent thru it and if a user wanted a guarantee, they could use the data source, now that it's available.

Regardless on whether or not yours (or #8350 for that matter) is a better approach, I think that we should not simply dismiss other people's hard work on an open source project (especially considering that I've been trying to get this fixed for a very long time - see #4278). #8350 does a much better job in explaining the outstanding issues with my approach without making it sound like my work should just be thrown out.

dtolnay:
Thanks for the feedback @vancluever. I agree with everything you wrote. Now that #8350 is available I am closing this PR in favor of that one. It is nice to see that after many months (counting from your first PR) we are finally making progress on resolving this.

PR #7787, opened 2016-07-24
provider/scaleway add default_security_group datasource

nicolai86:
Scaleway automatically creates one security group per organization. When you want to manage your security group rules for the entire org you have to know which security group is the default.

This PR adds a data source dedicated to that purpose:

```
# lookup default security group
data "scaleway_default_security_group" "default" {}

# modify rules within the default security group
resource "scaleway_security_group_rule" "http" {
  security_group = "${data.scaleway_default_security_group.default.id}"

  action = "drop"
  direction = "inbound"
  ip_range = "0.0.0.0/0"
  protocol = "TCP"
  port = 80
}
```

### DIFF
--- a/builtin/providers/scaleway/datasource_scaleway_default_security_group.go
+++ b/builtin/providers/scaleway/datasource_scaleway_default_security_group.go
@@ -1,0 +1,46 @@
+package scaleway
+
+import (
+	"fmt"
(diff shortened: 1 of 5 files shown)


nicolai86:
closing it for now as we might not need this…

PR #7789, opened 2016-07-24
Fix broken link to Consul demo

jonatanblue:
The links to the Key/value page on the Consul demo points to nyc1, which returns a 500.

Changing to nyc3 works.


stack72:
HI @jonatanblue 

Thanks so much for the PR here :) This is a great change to have made! Sorry it caught you out

Paul

PR #7802, opened 2016-07-25
Bug fix terraform 0.7.0 remote state data docs

mtougeron:
Terraform 0.7.0 data resource for remote state does not use the 'output' path


stack72:
Hi @mtougeron 

Thanks for the fix here :)

Paul

PR #7803, opened 2016-07-25
Add tf_vars to the variables sent in push

jbardin:
Add tf_vars to the data structures sent in `terraform push`. 

This takes any value of type `[]interface{}` or `map[string]interface{}` and marshals it as a string representation of the equivalent HCL. This prevents ambiguity in atlas between a string that looks like a json structure, and an actual json structure. 

For the time being we will need a way to serialize data as HCL, so the
command package has an internal `encodeHCL` function to do so. We can
remove this if we get complete package for marshaling HCL.

### DIFF
--- a/command/hcl_printer.go
+++ b/command/hcl_printer.go
@@ -1,0 +1,114 @@
+package command
+
+// Marshal an object as an hcl value.
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/hcl/hcl/printer"
+)
+
+// This will only work operate on []interface{}, map[string]interface{}, and
+// primitive types.
+func encodeHCL(i interface{}) ([]byte, error) {
+
+	state := &encodeState{}
+	err := state.encode(i)
+	if err != nil {
+		return nil, err
+	}
+
+	hcl := state.Bytes()
+	if len(hcl) == 0 {
+		return hcl, nil
+	}
+
+	// the HCL parser requires an assignment. Strip it off again later
+	fakeAssignment := append([]byte("X = "), hcl...)
+
+	// use the real hcl parser to verify our output, and format it canonically
+	hcl, err = printer.Format(fakeAssignment)
+
+	// now strip that first assignment off
+	eq := regexp.MustCompile(`=\s+`).FindIndex(hcl)
+	return hcl[eq[1]:], err
+}
+
+type encodeState struct {
+	bytes.Buffer
+}
+
+func (e *encodeState) encode(i interface{}) error {
+	switch v := i.(type) {
+	case []interface{}:
+		return e.encodeList(v)
+
+	case map[string]interface{}:
+		return e.encodeMap(v)
+
+	case int, int8, int32, int64, uint8, uint32, uint64:
+		return e.encodeInt(i)
+
+	case float32, float64:
+		return e.encodeFloat(i)
+
+	case string:
+		return e.encodeString(v)
+
+	case nil:
+		return nil
+
+	default:
+		return fmt.Errorf("invalid type %T", i)
+	}
+
+}
+
+func (e *encodeState) encodeList(l []interface{}) error {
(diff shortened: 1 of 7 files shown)


command/hcl_printer.go
```diff
@@ -0,0 +1,191 @@
...
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/hcl/hcl/printer"
+)
+
+// This will only work operate on []interface{}, map[string]interface{}, and
+// primitive types.
+func encodeHCL(i interface{}) ([]byte, error) {
+
```
phinze: whitespace nit - stray newline 😁

command/push.go
```diff
@@ -209,12 +210,23 @@ func (c *PushCommand) Run(args []string) int {
...
+			"An error has occurred while serializing the variables for uploading:\n"+
+				"%s", err))
+		return 1
+	}
+
 	// Upsert!
 	opts := &pushUpsertOptions{
 		Name:      name,
 		Archive:   archiveR,
 		Variables: ctx.Variables(),
```
phinze: the `TFVars` field supersedes `Variables` so we should drop this

phinze: Oh I see it's no longer used below - so this is more just a clean up then.

command/push.go
```diff
@@ -306,6 +371,7 @@ func (c *atlasPushClient) Get(name string) (map[string]interface{}, error) {
 
 	var variables map[string]interface{}
 	if version != nil {
+		// TODO: merge variables and TFVars
 		//variables = version.Variables
```
phinze: Leaving this for a follow on PR I presume? 👍

command/push_test.go
```diff
@@ -323,12 +322,11 @@ func TestPush_preferAtlas(t *testing.T) {
...
-	variables := map[string]interface{}{
-		"foo": "old",
-		"bar": "foo",
-	}
+	variables := pushTFVars()
+	variables["foo"] = "old"
+
 	if !reflect.DeepEqual(client.UpsertOptions.Variables, variables) {
-		t.Fatalf("bad: %#v", client.UpsertOptions)
+		t.Fatalf("bad: %#v", client.UpsertOptions.Variables)
```
phinze: This should be `TFVars` not `Variables` I think....

phinze:
Looking good! Just a few nits and Qs around the old Variables field and then this is good to go.

command/push_test.go
jbardin: Nope, this is another test, which we'll fix later. I only added `.Variables` for a shorter error message.

command/push.go
jbardin: Oh, I forgot about this. I'm not sure if this is supposed to still be commented out or not :/

phinze: So this is for the "subsequent pushes" use case, where the client fetches the existing variables and prints which ones are going to be overridden. Will chat you about this.

jbardin:
Fixed the pushes! (and that pesky busy-loop on input error)

terraform/context.go
```diff
@@ -345,7 +346,12 @@ func (c *Context) Input(mode InputMode) error {
 				}
 
 				if value == "" && v.Required() {
-					// Redo if it is required.
+					// Redo if it is required, but abort if we keep getting
+					// blank entries
+					if retry > 2 {
+						return fmt.Errorf("missing required value for %q", n)
+					}
+					retry++
```
phinze: 👏  great find!

command/hcl_printer.go
```diff
@@ -0,0 +1,190 @@
...
+			state.pop()
+		}
+
+		out = append(out, s[i])
+	}
+
+	out = append(out, '"')
+
+	return out
+}
```
phinze: lol if your daily assigned work doesn't require a stack-based parser to complete 😆 🚀

phinze:
Great work on this!

LGTM

PR #7804, opened 2016-07-25
provider/vSphere: Fixes the hasBootableVmdk flag when attaching multiple disks

dkalleg:
The hasBootableFlag logic had a bug where it would only be set properly
if the bootable disk was the last specified.  Adding some bool logic
resolves the issue.  Also cleaning up a redundant var.

### DIFF
--- a/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
+++ b/builtin/providers/vsphere/resource_vsphere_virtual_machine.go
@@ -789,7 +789,6 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 		if diskSet, ok := vL.(*schema.Set); ok {
 
 			disks := []hardDisk{}
-			hasBootableDisk := false
 			for _, value := range diskSet.List() {
 				disk := value.(map[string]interface{})
 				newDisk := hardDisk{}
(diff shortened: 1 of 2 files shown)


builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -799,10 +798,10 @@ func resourceVSphereVirtualMachineCreate(d *schema.ResourceData, meta interface{
 						return fmt.Errorf("Cannot specify name of a template")
 					}
 					vm.template = v
-					if hasBootableDisk {
+					if vm.hasBootableVmdk {
```
sputnik13: should this check include checking for multiple non-template disks with bootable?

dkalleg:
@stack72 Is there going to be an 0.7.0 rc4?  If so, any idea when it might be created?  If so, any chance this can make it in?  This bug has some pretty bad effects on multiple disk vms.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
dkalleg: Done.

stack72:
@dkalleg are we good with this one? Have you ran the tests?

P.

dkalleg:
Hey @stack72, they're passing on my end.

stack72:
ok, I'm going to merge this - if possible, please drop the test results in so people can see them if they look :)

dkalleg:
Thanks @stack72 !

PR #7805, opened 2016-07-26
Fix typo in aws_vpc resource docs (VPN -> VPC)

bradfeehan:
Just noticed this while browsing the docs, is this fix correct?


radeksimko:
👍

PR #7818, opened 2016-07-26
provider/aws: Restore Defaults to SQS Queues

### LINKED ISSUE #7505: aws_sqs_queue not restoring visibility_timeout_seconds to default value after removal of attribute
### Terraform Version
Terraform v0.6.16

### Affected Resource(s)
- aws_sqs_queue

```hcl
resource "aws_sqs_queue" "backend_queue" {
    name = "backend"
    visibility_timeout_seconds = 90
    policy = "{*snip*}"
}
```

### Expected Behavior
According to https://www.terraform.io/docs/providers/aws/r/sqs_queue.html, the default value for the visibility_timeout_seconds attribute is 30. Upon the removal of this attribute from the resource the value should be set to 30.

### Actual Behavior
Upon removal, the value of the attribute stayed on the previously set value, 90 in this case.

### Steps to Reproduce
1. Create an aws_sqs_queue resource without specifying visibility_timeout_seconds
2. Run `terraform apply` to create the resource
3. Add the visibility_timeout_seconds attribute and set it to a value other than 30
4. Run `terraform apply` to modify the resource
5. Remove the visibility_timeout_seconds attribute
6. Run `terraform apply` to see no modifications are done

catsby:
Fixes #7505 by removing `computed` and adding `default` values, to match with the SDK defaults.

I think this is somewhat new territory, as we typically just do nothing with the _removal_ of an optional, computed value, but here we restore the defaults because they are documented. Without this, you couldn't remove a setting once you set it, you'd have to find the default yourself and manually set it.

### DIFF
--- a/builtin/providers/aws/resource_aws_sqs_queue.go
+++ b/builtin/providers/aws/resource_aws_sqs_queue.go
@@ -70,7 +70,7 @@ func resourceAwsSqsQueue() *schema.Resource {
 			"visibility_timeout_seconds": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
-				Computed: true,
+				Default:  30,
 			},
 			"policy": &schema.Schema{
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


stack72:
@catsby LGTM! 

There was 1 small issue with a test:

```
=== RUN   TestAccAWSSQSQueue_importBasic
--- FAIL: TestAccAWSSQSQueue_importBasic (15.87s)
	testing.go:264: Step 1 error: Resource specified by ResourceName couldn't be found: aws_sqs_queue.queue-with-defaults
```

Fixed in 4e1d54a

PR #7819, opened 2016-07-26
provider/powerdns: Add support for PowerDNS 4 API

### LINKED ISSUE #6056: PowerDNS provider changed for version 4.0.0(Alpha2) ttl key moved outside of the record
For anyone trying to use PowerDNS provider against their release candidate (Alpha2 and beyond) please note that they've moved their ttl key out of the record json block to outer block. If you use the current powerdns provider with terraform >=0.6.14 you will get an error with Key 'ttl' is not an integer or not present because this key is required and must be set when sent payload against pdns api

smuth4:
This adds support for the non-experimental API present in PowerDNS 4, while retaining support for 3.X, with minimal changes.

Notes:
* I left the acceptance tests alone in this PR, but by add a trailing `.` to all DNS names (including in the records), they all succeeded against an actual 4.0.0 server
* I noted an edge case regarding case sensitivity, but refrained from trying to fix it, since it's not clear which records and/or situations it effects.

If the plan is to drop 3.X entirely, I don't mind fixing those two things up, but that's not a decision I wanted to make alone.

Fixes #6056

### DIFF
--- a/builtin/providers/powerdns/client.go
+++ b/builtin/providers/powerdns/client.go
@@ -126,7 +126,7 @@ func parseId(recId string) (string, string, error) {
 
 // Detects the API version in use on the server
 // Uses int to represent the API version: 0 is the legacy AKA version 3.4 API
-// Any other integer correlates with te same API version
+// Any other integer correlates with the same API version
 func (client *Client) detectApiVersion() (int, error) {
 	req, err := client.newRequest("GET", "/api/v1/servers", nil)
 	if err != nil {
--- a/builtin/providers/powerdns/client.go
+++ b/builtin/providers/powerdns/client.go
@@ -70,20 +70,21 @@ func (c *Client) newRequest(method string, endpoint string, body []byte) (*http.
 }
 
 type ZoneInfo struct {
-	Id      string   `json:"id"`
-	Name    string   `json:"name"`
-	URL     string   `json:"url"`
-	Kind    string   `json:"kind"`
-	DnsSec  bool     `json:"dnsssec"`
-	Serial  int64    `json:"serial"`
-	Records []Record `json:"records,omitempty"`
+	Id                 string              `json:"id"`
+	Name               string              `json:"name"`
+	URL                string              `json:"url"`
+	Kind               string              `json:"kind"`
+	DnsSec             bool                `json:"dnsssec"`
+	Serial             int64               `json:"serial"`
+	Records            []Record            `json:"records,omitempty"`
+	ResourceRecordSets []ResourceRecordSet `json:"rrsets,omitempty"`
 }
 
 type Record struct {
 	Name     string `json:"name"`
 	Type     string `json:"type"`
 	Content  string `json:"content"`
-	TTL      int    `json:"ttl"`
+	TTL      int    `json:"ttl"` // For API v0
 	Disabled bool   `json:"disabled"`
 }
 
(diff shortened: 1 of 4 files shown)


stack72:
Hi @smuth4 

Thanks for the work here - will this PR allow the PowerDNS provider to work with both 3 and 4 version servers?

Paul

smuth4:
@stack72 Yes, I ran the acceptance tests against a version 3 server to make sure. There are a few differences in the 4.X API that it doesn't handle automagically, but it is able API-compatible with both versions.

stack72:
Hi @smuth4 

Ok, I have also tested this ran against out current test framework for PowerDNS :)

```
% make testacc TEST=./builtin/providers/powerdns TESTARGS='-run=TestAccPDNS'
TF_ACC=1 go test ./builtin/providers/powerdns -v -run=TestAccPDNS -timeout 120m
=== RUN   TestAccPDNSRecord_A
--- PASS: TestAccPDNSRecord_A (0.06s)
=== RUN   TestAccPDNSRecord_WithCount
--- PASS: TestAccPDNSRecord_WithCount (0.08s)
=== RUN   TestAccPDNSRecord_AAAA
--- PASS: TestAccPDNSRecord_AAAA (0.05s)
=== RUN   TestAccPDNSRecord_CNAME
--- PASS: TestAccPDNSRecord_CNAME (0.06s)
=== RUN   TestAccPDNSRecord_HINFO
--- PASS: TestAccPDNSRecord_HINFO (0.06s)
=== RUN   TestAccPDNSRecord_LOC
--- PASS: TestAccPDNSRecord_LOC (0.05s)
=== RUN   TestAccPDNSRecord_MX
--- PASS: TestAccPDNSRecord_MX (0.09s)
=== RUN   TestAccPDNSRecord_NAPTR
--- PASS: TestAccPDNSRecord_NAPTR (0.05s)
=== RUN   TestAccPDNSRecord_NS
--- PASS: TestAccPDNSRecord_NS (0.06s)
=== RUN   TestAccPDNSRecord_SPF
--- PASS: TestAccPDNSRecord_SPF (0.18s)
=== RUN   TestAccPDNSRecord_SSHFP
--- PASS: TestAccPDNSRecord_SSHFP (0.06s)
=== RUN   TestAccPDNSRecord_SRV
--- PASS: TestAccPDNSRecord_SRV (0.06s)
=== RUN   TestAccPDNSRecord_TXT
--- PASS: TestAccPDNSRecord_TXT (0.06s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/powerdns	0.931s
```

Thanks so much for the work here ! LGTM so merging

linuxus:
Thanks @smuth4 I've been waiting for this PR for quite some time now and this will address a lot of my infra build with PDNS 4.x

Much appreciated.

builtin/providers/powerdns/client.go
```diff
@@ -111,6 +124,26 @@ func parseId(recId string) (string, string, error) {
 	}
 }
 
+// Detects the API version in use on the server
```
linuxus: This piece of code is breaking my .tf deployment as it tries to detect the API version before we launch the instance where PDNS is going to be hosted: here is my provider's config and you can see that my powerdns_server_url link is a place holder of a resource that will be created:

```
# Power DNS Provider
provider "powerdns" {
  api_key    = "${var.pdns_api_key}"
  #server_url = "http://${openstack_compute_instance_v2.admin_node.floating_ip}:8081"
}
```
obviously my terraform plan is failing with:
```
* Error setting up PowerDNS client: Error during parsing request URL: parse http://${openstack_compute_instance_v2.admin_node.floating_ip}:8081/api/v1/servers: invalid character "{" in host name
```
Wouldn't be better to check the server API version prior creating the records?

PR #7829, opened 2016-07-27
provider/aws: Don't delete Lambda function from state on initial call of the Read func

### LINKED ISSUE #7782: aws_lambda_function doesn't create lambda
Running aws_lambda_function reports successful resource creation, but resource is not shown in the "show" and arn is not available for subsequent resource creations.

### Terraform Version
v0.6.16

### Affected Resource(s)
aws_lambda_function (java8)

```hcl
variable "file" {
	default = "../target/test-0.0.1-SNAPSHOT.jar"
}

variable "region" {
	default = "ap-southeast-2"
}

provider "aws" {
	region = "${var.region}"
}

resource "aws_iam_role" "iam_for_lambda" {
    name = "iam_for_lambda"
    assume_role_policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "sts:AssumeRole",
      "Principal": {
        "Service": "lambda.amazonaws.com"
      },
      "Effect": "Allow",
      "Sid": ""
    }
  ]
}
EOF
}

resource "aws_lambda_function" "test_lambda" {
    filename = "${var.file}"
    function_name = "Hello"
    role = "${aws_iam_role.iam_for_lambda.arn}"
    handler = "LambdaFunctionHandler.handleRequest"
	runtime = "java8"
}
```

### Expected Behavior
EXpected lambda to be created and "terraform show" confirming that.

### Actual Behavior
Only aws_iam_role is created

### Steps to Reproduce
1. `terraform apply` - reports successful creation of the lambda
2. `terraform show` - only shows aws_iam_role being created

stack72:
Fixes #7782

Lambda functions are eventually consistent :( Therefore, when we move
from the Create func to the Read func, there is a chance that the Lambda
hasn't replicated yet and we could therefore find that it doesn't exist
and delete it as follows:

```
params := &lambda.GetFunctionInput{
        FunctionName: aws.String(d.Get("function_name").(string)),

}

    getFunctionOutput, err := conn.GetFunction(params)
    if err != nil {
            if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
                        d.SetId("")
                                    return nil

            }
                    return err

    }
```

This PR uses `d.IsNewResource()` to check if the Read is being called
after a Create and therefore, won't delete the lambda if not found. This
should allow the lambda to replicate

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSLambdaFunction_'
=> Checking that code complies with gofmt requirements...
-run=TestAccAWSLambdaFunction_ -timeout 120m
=== RUN   TestAccAWSLambdaFunction_importLocalFile
--- PASS: TestAccAWSLambdaFunction_importLocalFile (36.64s)
=== RUN   TestAccAWSLambdaFunction_importLocalFile_VPC
--- PASS: TestAccAWSLambdaFunction_importLocalFile_VPC (45.17s)
=== RUN   TestAccAWSLambdaFunction_importS3
--- PASS: TestAccAWSLambdaFunction_importS3 (40.88s)
=== RUN   TestAccAWSLambdaFunction_basic
--- PASS: TestAccAWSLambdaFunction_basic (44.77s)
=== RUN   TestAccAWSLambdaFunction_VPC
--- PASS: TestAccAWSLambdaFunction_VPC (44.13s)
=== RUN   TestAccAWSLambdaFunction_s3
--- PASS: TestAccAWSLambdaFunction_s3 (43.62s)
=== RUN   TestAccAWSLambdaFunction_localUpdate
--- PASS: TestAccAWSLambdaFunction_localUpdate (33.49s)
=== RUN   TestAccAWSLambdaFunction_localUpdate_nameOnly
--- PASS: TestAccAWSLambdaFunction_localUpdate_nameOnly (51.83s)
=== RUN   TestAccAWSLambdaFunction_s3Update
--- PASS: TestAccAWSLambdaFunction_s3Update (106.49s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    447.055s
```

Thanks to @radeksimko for pointing out `d.IsNewResource()`

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -237,7 +237,7 @@ func resourceAwsLambdaFunctionRead(d *schema.ResourceData, meta interface{}) err
 
 	getFunctionOutput, err := conn.GetFunction(params)
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" && !d.IsNewResource() {
 			d.SetId("")
 			return nil
 		}


builtin/providers/aws/resource_aws_lambda_function.go
```diff
@@ -237,7 +237,7 @@ func resourceAwsLambdaFunctionRead(d *schema.ResourceData, meta interface{}) err
 
 	getFunctionOutput, err := conn.GetFunction(params)
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" && !d.IsNewResource() {
```
radeksimko: I know it's unrelated to the original intention behind this PR, but it would be great to have a `[WARN]` log message here, like we have in most other resources. 😉

radeksimko:
This LGTM.

builtin/providers/aws/resource_aws_lambda_function.go
stack72: Funny you mention that, there will be a little more standardisation in this area.... soon ™️

PR #7831, opened 2016-07-27
Remove unused variables

jbardin:
These variables weren't used, but the compiler misses them since they
are captured in a closure.


stack72:
LGTM!

PR #7832, opened 2016-07-27
config: Add map() interpolation function

phinze:
* `map(key, value, ...)` - Returns a map consisting of the key/value pairs
  specified as arguments. Every odd argument must be a string key, and every
  even argument must have the same type as the other values specified. Examples:
  * `map("hello", "world")`
  * `map("us-east", list("a", "b", "c"), "us-west", list("b", "c", "d"))`

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -71,6 +71,7 @@ func Funcs() map[string]ast.Function {
 		"length":       interpolationFuncLength(),
 		"list":         interpolationFuncList(),
 		"lower":        interpolationFuncLower(),
+		"map":          interpolationFuncMap(),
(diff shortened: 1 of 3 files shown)


phinze:
cc @jen20 @jbardin - this is a sibling function to `list()` - helps for symmetry and legit use cases but also helps testing edge cases in the other interpolation function unit tests 😄

jen20:
This does solve the immediate problem rather nicely, but I still think we should support list and map literals in interpolation at some point in the future?

phinze:
Yes totally agreed.

phinze:
Updated to return an error when duplicate map keys are specified.

PR #7838, opened 2016-07-27
Website typo

tomgoren:
Documentation for AWS provider, specifically for RDS DB instances, was missing a single space, and the casing was wrong for the word "Snapshot".

See: https://www.terraform.io/docs/providers/aws/r/db_instance.html#skip_final_snapshot

This is my first attempt at contributing to a "real" project via GitHub, so I hope I'm doing it right.

Thanks for Terraform!


stack72:
Hi @tomgoren 

Thanks so much for the PR here - this is a great way to get started :) We really love docs PRs as these are essential to new users!

When the build goes green, I shall merge

Thanks again

Paul

PR #7842, opened 2016-07-28
Ensure kms_key_id docs indicate it is an ARN

borgstrom:
If you specify just a bare ID, then the initial application works but
subsequent applications may end up doing bad things, like:

```
-/+ aws_ebs_volume.vol_1
    availability_zone: "us-east-1a" => "us-east-1a"
    encrypted:         "true" => "true"
    iops:              "" => "<computed>"
    kms_key_id:        "arn:aws:kms:us-east-1:123456789:key/59faf88b-0912-4cca-8b6c-bd107a6ba8c4" => "59faf88b-0912-4cca-8b6c-bd107a6ba8c4" (forces new resource)
    size:              "100" => "100"
    snapshot_id:       "" => "<computed>"
```


stack72:
Hi @borgstrom 

Thanks so much for the PR here - this is a good addition!

Paul

PR #7844, opened 2016-07-28
Fix OpenStack documentation

dvorak:
References to security_groups attribute on network objects should
actually be security_group_ids.


stack72:
Hi @dvorak 

Thanks so much for this! This LGTM - will merge after the build passes

Paul

PR #7847, opened 2016-07-28
Fix subnet part of OS network example

dvorak:
This fixes the hard coded network id and also fixes the fixed_ip address
to be one on the network that is specified above.


jen20:
:+1: LGTM, thanks @dvorak!

PR #7851, opened 2016-07-28
website: Docs sweep for lists & maps

phinze:


phinze:
cc @catsby @jen20 @jbardin @stack72 for some english 👀

website/source/docs/commands/apply.html.markdown
```diff
@@ -53,10 +53,11 @@ The command-line flags are all optional. The list of available flags are:
   multiple times.
 
 * `-var 'foo=bar'` - Set a variable in the Terraform configuration. This
-  flag can be set multiple times.
+  flag can be set multiple times. Variable values are interpreted as HCL, so
```
stack72: is HCL a known thing to people outside of Terraform codebase?

phinze: Hm good point. I'll look for something better to link to for these references.

jbardin: Believe it or not, it is! It's not widely known, but I've definitely encountered people who use it. (some adopted it because it was an option through the viper config package)

stack72: excellent - making sure that acronym is known will then mean it LGTM!

jbardin:
LGMT2

website/source/docs/commands/apply.html.markdown
phinze: Reworked to mention HCL on the syntax page and link off to it every time it is mentioned.

stack72:
Looks great now :) 🚢

PR #7855, opened 2016-07-29
Tainted resource not recreated if ignore_changes used.

sl1pm4t:
**Problem:**

If a resource has `ignore_changes` attribute set, and becomes tainted (either manually or due to an error during creation), Terraform does not include the tainted resource in the next diff.

**Steps to Reproduce**

Use the following terraform config:

```hcl
resource "template_file" "taint-test" {
  template = <<EOT
Hello ${world}
EOT

  vars {
    world = "world"
  }

  lifecycle {
    ignore_changes = ["vars"]
  }

  provisioner "local-exec" {
    command = "echo ${self.rendered}"
  }

  // run "false" so provisioner fails and resource is tainted
  provisioner "local-exec" {
    command = "false"
  }
}
```

and run `terraform apply`.
The `local-exec` with `command = "false"` provisioner will cause the resource creation to error, and the resource will be tainted, e.g.:

```
MMORRISON-MBP:tf-taint-test-2 mmorrison$ terraform apply
template_file.taint-test: Creating...
  rendered:   "" => "<computed>"
  template:   "" => "Hello ${world}\n"
  vars.%:     "" => "1"
  vars.world: "" => "world"
template_file.taint-test: Provisioning with 'local-exec'...
template_file.taint-test (local-exec): Executing: /bin/sh -c "echo Hello world
template_file.taint-test (local-exec): "
template_file.taint-test (local-exec): Hello world
template_file.taint-test: Provisioning with 'local-exec'...
template_file.taint-test (local-exec): Executing: /bin/sh -c "false"
Error applying plan:

1 error(s) occurred:

* Error running command 'false': exit status 1. Output:

MMORRISON-MBP:tf-taint-test-2 mmorrison$
```

However, when `terraform plan` is run, there is no diff:

```
MMORRISON-MBP:tf-taint-test-2 mmorrison$ terraform plan

template_file.taint-test: Refreshing state... (ID: 1894a19c85ba153acbf743ac4e43fc004c891604b26f8c69e1e83ea2afc7c48f)

doesn't need to do anything.
```
This is tested with master branch (0.7) commit 70999b1a643223f58a557c5d9c5e7e8c0c61bfe9

**Solution**

After the committed change, the plan output is:

```
MMORRISON-MBP:tf-taint-test-2 mmorrison$ terraform plan

template_file.taint-test: Refreshing state... (ID: 1894a19c85ba153acbf743ac4e43fc004c891604b26f8c69e1e83ea2afc7c48f)

-/+ template_file.taint-test (tainted)
    rendered:   "Hello world\n" => "<computed>"
    template:   "Hello ${world}\n" => "Hello ${world}\n"
    vars.%:     "1" => "1"
    vars.world: "world" => "world"


```

### DIFF
--- a/terraform/eval_diff.go
+++ b/terraform/eval_diff.go
@@ -185,6 +185,11 @@ func (n *EvalDiff) processIgnoreChanges(diff *InstanceDiff) error {
 		return nil
 	}
 
+	// If the resource has been tainted we shouldn't alter the Diff
+	if diff.DestroyTainted {
+		return nil
+	}
+
 	ignorableAttrKeys := make(map[string]bool)
 	for _, ignoredKey := range ignoreChanges {
 		for k := range diff.Attributes {


sl1pm4t:
Ping @stack72 @mitchellh - any chance we can get eyes on this one. 

The bug causes some frustrating problems (even beyond what I've shown above, e.g. dependency graph is messed up if one tries to `destroy -target <resource using ignore_changes>`) and has a very simple fix.

mitchellh:
I'm taking a look now, on first look this looks right, but we need to add a test case for this but I agree its a pretty not great bug so doing this now.

mitchellh:
Got it, good fix: https://github.com/hashicorp/terraform/commit/9e9779ebaabee303dc49c244cd40462aa8714391

Thanks!

sl1pm4t:
Awesome, thanks @mitchellh !

PR #7857, opened 2016-07-29
provider/aws: Bump SDK package version to v1.2.10

stack72:


catsby:
Postponing this until after v0.7.0!  Which well be soon!

soonish

PR #7864, opened 2016-07-29
Add missing 'name' attribute to documentation.

carinadigital:
When adding network security group rules they are attached by name,
as opposed to attaching the security group which is done by id.


stack72:
Hi @carinadigital 

Apologies for the confusion here. Any parameters passed into the resource (e.g. name) is also already available as an output. The `## Attributes Reference` section is usually for outputs from the resource that are not covered in the inputs, if that makes sense?

Paul

carinadigital:
It does. Closing.

PR #7865, opened 2016-07-29
terraform: Filter untargeted variable nodes

phinze:
When targeting, only Addressable untargeted nodes were being removed
from the graph. Variable nodes are not directly Addressable, so they
were hanging around. This caused problems with module variables that
referred to Resource nodes. The Resource node would be filtered out of
the graph, but the module Variable node would not, so it would try to
interpolate during the graph walk and be unable to find it's referent.

This would present itself as strange "cannot find variable" errors for
variables that were uninvolved with the currently targeted set of
resources.

Here, we introduce a new interaface that can be implemented by graph
nodes to indicate they should be filtered out from targeting even though
they are not directly addressable themselves.

### DIFF
--- a/terraform/context_plan_test.go
+++ b/terraform/context_plan_test.go
@@ -2114,6 +2114,43 @@ module.child:
 	}
 }
 
+func TestContext2Plan_targetedModuleUntargetedVariable(t *testing.T) {
+	m := testModule(t, "plan-targeted-module-untargeted-variable")
+	p := testProvider("aws")
+	p.DiffFn = testDiffFn
+	ctx := testContext2(t, &ContextOpts{
+		Module: m,
+		Providers: map[string]ResourceProviderFactory{
+			"aws": testProviderFuncFixed(p),
+		},
+		Targets: []string{"aws_instance.blue", "module.blue_mod"},
+	})
+
+	plan, err := ctx.Plan()
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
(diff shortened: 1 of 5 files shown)


terraform/graph_config_node.go
```diff
@@ -39,3 +39,14 @@ type GraphNodeTargetable interface {
...
 }
+
+// GraphNodeRemoveIfNotTargetedable is a special interface for graph nodes that
+// aren't directly addressable, but need to be removed from the graph when they
+// are not targeted. (Nodes that are not directly targeted end up in the set of
+// targeted nodes because something that _is_ targeted depends on them.) The
+// initial use case for this interface is GraphNodeConfigVariable, which was
+// having trouble interpolating for module variables in targeted scenarios that
+// filtered out the resource node being referenced.
+type GraphNodeRemoveIfNotTargetedable interface {
```
jen20: This is a _really_ weird name - perhaps `GraphNodeRemoveIfNotTargetable`?

jen20: Although with more context that's not descriptive enough. Either way, this one isn't a word! 😀

jen20:
LGTM in principle, one minor gripe about the naming of the new interface though.

terraform/graph_config_node.go
phinze: Yes terrible name I agree. I hate naming interfaces. 😝 

`GraphNodeTargetFilterable` perhaps? Will play.

jen20:
Way nicer!

PR #7874, opened 2016-07-29
Skip IAM/STS validation and metadata check

renier:
Skip IAM/STS validation and metadata check

* Skip IAM/STS identity validation - For environments or other api
  implementations where there are no IAM/STS endpoints available, this
  option lets you opt out from that provider initialization step.
* Skip metdata api check - For environments in which you know ahead of
  time there isn't going to be a metadta api endpoint, this option lets
  you opt out from that check to save time.

Sample provider config:

```hcl
provider "aws" {
  region = "us-east-1"
  skip_iam_creds_validation = true
  skip_iam_account_id = true
  skip_metadata_api_check = true
}
```

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -70,15 +70,16 @@ type Config struct {
 	AllowedAccountIds   []interface{}
 	ForbiddenAccountIds []interface{}
 
-	DynamoDBEndpoint     string
-	KinesisEndpoint      string
-	Ec2Endpoint          string
-	IamEndpoint          string
-	ElbEndpoint          string
-	S3Endpoint           string
-	Insecure             bool
-	SkipIamValidation    bool
-	SkipMetadataApiCheck bool
+	DynamoDBEndpoint       string
+	KinesisEndpoint        string
+	Ec2Endpoint            string
+	IamEndpoint            string
+	ElbEndpoint            string
+	S3Endpoint             string
+	Insecure               bool
+	SkipIamCredsValidation bool
+	SkipIamAccountId       bool
+	SkipMetadataApiCheck   bool
 }
 
 type AWSClient struct {
@@ -202,12 +203,15 @@ func (c *Config) Client() (interface{}, error) {
 		client.iamconn = iam.New(awsIamSess)
 		client.stsconn = sts.New(sess)
 
-		if c.SkipIamValidation == false {
+		if c.SkipIamCredsValidation == false {
 			err = c.ValidateCredentials(client.stsconn)
 			if err != nil {
 				errs = append(errs, err)
 				return nil, &multierror.Error{Errors: errs}
 			}
+		}
+
+		if c.SkipIamAccountId == false {
 			accountId, err := GetAccountId(client.iamconn, client.stsconn, cp.ProviderName)
 			if err == nil {
 				client.accountid = accountId
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -198,11 +198,11 @@ func (c *Config) Client() (interface{}, error) {
 		dynamoSess := sess.Copy(&aws.Config{Endpoint: aws.String(c.DynamoDBEndpoint)})
 		kinesisSess := sess.Copy(&aws.Config{Endpoint: aws.String(c.KinesisEndpoint)})
 
-		if c.SkipIamValidation == false {
-			// These two services need to be set up early so we can check on AccountID
-			client.iamconn = iam.New(awsIamSess)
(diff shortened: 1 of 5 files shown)


iamatypeofwalrus:
Just saw this PR while looking at references to my own. 

Totally agree with the Boolean switch reasoning and happy to close my PR and move forward with this approach.

builtin/providers/aws/config.go
```diff
@@ -195,23 +198,25 @@ func (c *Config) Client() (interface{}, error) {
 		dynamoSess := sess.Copy(&aws.Config{Endpoint: aws.String(c.DynamoDBEndpoint)})
 		kinesisSess := sess.Copy(&aws.Config{Endpoint: aws.String(c.KinesisEndpoint)})
 
-		// These two services need to be set up early so we can check on AccountID
-		client.iamconn = iam.New(awsIamSess)
-		client.stsconn = sts.New(sess)
+		if c.SkipIamValidation == false {
+			// These two services need to be set up early so we can check on AccountID
+			client.iamconn = iam.New(awsIamSess)
+			client.stsconn = sts.New(sess)
```
radeksimko: This would cause Terraform to crash if you tried `apply`ing any `iam_*` resource with `SkipIamValidation` set to `true`. Is there any reason why not keep it outside of the condition?

builtin/providers/aws/config.go
```diff
@@ -195,23 +198,25 @@ func (c *Config) Client() (interface{}, error) {
...
+			}
+			accountId, err := GetAccountId(client.iamconn, client.stsconn, cp.ProviderName)
+			if err == nil {
+				client.accountid = accountId
+			}
 
-		authErr := c.ValidateAccountId(client.accountid)
-		if authErr != nil {
-			errs = append(errs, authErr)
+			authErr := c.ValidateAccountId(client.accountid)
```
radeksimko: I'm thinking that some people theoretically may have the metadata API still available, so they may still want to perform the account ID validation, but they may not want to use IAM/STS API at all.

Therefore in the following case
```hcl
provider "aws" {
  skip_iam_validation = true
  allowed_account_ids = ["1111111111"]
}
```
ID check will be silently ignored even though the Account ID may be available via Metadata API.

As you can see in the `ValidateAccountId` implementation, it only performs validation if `AllowedAccountIds` or `ForbiddenAccountIds` are non-empty, so it's totally safe to keep it outside of the conditional block.

Likewise `ValidateCredentials` would _gracefully degrade_ - i.e. return empty string if it was impossible to find account ID from all available sources.

That said I understand you may want to eliminate any extra API calls to APIs that you know are unavailable in your environment - so how about this?

```hcl
provider "aws" {
  skip_credentials_validation = true
  skip_account_id_validation = true
  skip_metadata_api_check = true
}
```
That way we can cater both groups of people - those that don't want Terraform to perform _any_ API calls and those who _only_ want to ignore IAM/STS, but still want to perform account ID validation.

We could also mark `skip_account_id_validation` as `ConflictsWith: []string{"allowed_account_ids", "forbidden_account_ids"}`, but it would fail early anyway, because `"" != "given-account-id"`.

radeksimko:
Thanks for the separation. This really helps us reviewing each feature thoroughly and discuss details and effects more easily.

I left you two comments there that may need addressing.
Otherwise this PR is looking pretty good functional/approach-wise.

------

I will probably add a few comments to the docs once we merge this - effects to be aware of. e.g.

 - `skip_credentials_validation = true` doesn't prevent you from managing `iam_*` resources, but Terraform will still try to reach to the **original AWS IAM endpoint** which **may not be what you want in non-AWS environments**
 - `skip_requesting_account_id = true` prevents you from managing any resource that requires Account ID to construct an ARN, namely
   - `aws_db_instance`
   - `aws_db_option_group`
   - `aws_db_parameter_group`
   - `aws_db_security_group`
   - `aws_db_subnet_group`
   - `aws_elasticache_cluster`
   - `aws_glacier_vault`
   - `aws_rds_cluster`
   - `aws_rds_cluster_instance`
   - `aws_rds_cluster_parameter_group`
   - `aws_redshift_cluster`
 - `skip_metadata_api_check = true` prevents Terraform from authenticating via Metadata API - i.e. you may need to use other auth methods (static credentials set as ENV vars or config)

renier:
@radeksimko Thanks for the review. I don't think this would ever be used against real AWS environments, but I think I addressed all your comments with the iam/sts initialization and increased skip choice resolution.

radeksimko:
👍 

Technically credentials or account ID may not always come from the IAM API (but from metadata API), but that's a really a naming nitpick I'm happy to address in a separate PR.

PR #7877, opened 2016-07-29
core: Fix race conditions, mostly around diffs

jbardin:
Add a mutex and some locked getters and setters to InstanceDiff, so that we can safely use them concurrently. It's not optimal that we expose these in the exported API, but they need to be accessed from the command and the schema packages too. 

Add a mutex to the MockHook, and lock all methods.

### DIFF
--- a/config/raw_config.go
+++ b/config/raw_config.go
@@ -93,6 +93,8 @@ func (r *RawConfig) Value() interface{} {
 // structure will always successfully decode into its ultimate
 // structure using something like mapstructure.
 func (r *RawConfig) Config() map[string]interface{} {
+	r.lock.Lock()
+	defer r.lock.Unlock()
 	return r.config
 }
 
--- a/terraform/diff.go
+++ b/terraform/diff.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
 )
 
 // DiffChangeType is an enum with the kind of changes a diff has planned.
@@ -216,9 +217,9 @@ func (d *ModuleDiff) String() string {
 
 		crud := "UPDATE"
 		switch {
-		case rdiff.RequiresNew() && (rdiff.Destroy || rdiff.DestroyTainted):
+		case rdiff.RequiresNew() && (rdiff.GetDestroy() || rdiff.GetDestroyTainted()):
 			crud = "DESTROY/CREATE"
-		case rdiff.Destroy:
+		case rdiff.GetDestroy():
 			crud = "DESTROY"
 		case rdiff.RequiresNew():
 			crud = "CREATE"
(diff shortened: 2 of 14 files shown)


config/raw_config.go
```diff
@@ -93,6 +93,8 @@ func (r *RawConfig) Value() interface{} {
 // structure will always successfully decode into its ultimate
 // structure using something like mapstructure.
 func (r *RawConfig) Config() map[string]interface{} {
+	r.lock.Lock()
+	defer r.lock.Unlock()
```
phinze: This one is sort of a cheat since we're returning the pointer to the map anyways, yeah?

phinze: Relevant questions:

 * Do callers require write access to the result returned?
 * If yes, should we do a copy here?

terraform/diff.go
```diff
@@ -366,6 +375,17 @@ func (d *InstanceDiff) RequiresNew() bool {
 		return false
 	}
 
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
phinze: Why lock the mutex here instead of inside `requiresNew()`?

terraform/diff.go
```diff
@@ -379,24 +399,103 @@ func (d *InstanceDiff) RequiresNew() bool {
...
+}
+
+// Safely copies the Attributes map
+func (d *InstanceDiff) CopyAttributes() map[string]*ResourceAttrDiff {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	attrs := make(map[string]*ResourceAttrDiff)
+	for k, v := range d.Attributes {
+		attrs[k] = v
```
phinze: Noting that `ResourceAttrDiff` is not deep copied here. Offhand, I don't believe there's an operative difference.

phinze:
Bless your concurrent gopher heart! This looks great. A few Qs inline.

config/raw_config.go
jbardin: Yeah, this was a "stop the failures" change. I'll check if this can return a copy.

terraform/diff.go
jbardin: `requiresNew` is called from code than already holds the lock. A pattern I often use is have an exported method which takes a mutex and delegates to an private method to do the work. This way locked methods can call one another through the private interface.

terraform/diff.go
jbardin: Nope. This is strictly to protect the map. (this was the root of the runtime panics)

terraform/diff.go
phinze: Yeah now that I've read down to the bottom I see the pattern at play. 👍

phinze:
https://www.youtube.com/watch?v=xoHECVnQC7A

LGTM 🐎

PR #7882, opened 2016-07-30
Update doc page for aws_elb

moorara:
I've the name arguments for `access_logs`, `listener`, and `health_check` to their corresponding paragraphs, so one won't need to scroll up and find the actual name of arguments.


stack72:
Hi @moorara 

Thanks for the work here :) This looks good to me!

Paul

PR #7899, opened 2016-08-01
core: Disallow strings as arguments to concat

jbardin:
The concat interpolation function now only accepts list arguments.
Strings are no longer supported, for concatenation or appending to
lists. All arguments must be a list, and single elements can be promoted
with the `list` interpolation function.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -311,31 +311,25 @@ func interpolationFuncCoalesce() ast.Function {
 // multiple lists.
 func interpolationFuncConcat() ast.Function {
 	return ast.Function{
-		ArgTypes:     []ast.Type{ast.TypeAny},
+		ArgTypes:     []ast.Type{ast.TypeList},
 		ReturnType:   ast.TypeList,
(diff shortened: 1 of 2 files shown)


config/interpolate_funcs.go
```diff
@@ -337,7 +335,7 @@ func interpolationFuncConcat() ast.Function {
 					}
 
 				default:
-					return nil, fmt.Errorf("concat() does not support %T", arg)
+					return nil, fmt.Errorf("concat() does not support type %T", arg)
```
phinze: I think we might be able to get HIL to do this for us by changing `VariadicType` to `ast.TypeList` above

jbardin: Ah, you're right. We're down to one type, so I'll let HIL take care of it

jbardin:
now using only `ast.TypeList`

jbardin:
oh wait, hold on. I forgot this was recursive :/ That needs a test too

phinze:
LGTM

PR #7900, opened 2016-08-01
website: v0.7 Upgrade Guide

phinze:


jbardin:
I'm admittedly not the best proofreader, but LGTM

website/source/upgrade-guides/0-7.html.markdown
```diff
@@ -0,0 +1,135 @@
...
+
+Before v0.7, Terraform's built-in plugins for providers and provisioners were each distributed as separate binaries.
+
+```
+terraform               # core binary
+terraform-provider-*    # provider plugins
+terraform-provisioner-* # provisioner plugins
+```
+
+These binaries needed to all be extracted to somewhere on your `$PATH` for Terraform to work.
```
jen20: "In order for Terraform to work, each of these binaries must have been available on your `$PATH`, or in the `~/.terraform.d` directory."?

website/source/upgrade-guides/0-7.html.markdown
```diff
@@ -0,0 +1,135 @@
...
+
+As of v0.7, these plugins all ship embedded in a single binary. This means that if you just extract the v0.7 archive into a path, you may still have the old separate binaries in your `$PATH`. You'll need to remove them manually.
```
jen20: "all built-in plugins ship embedded into a single binary"?

website/source/upgrade-guides/0-7.html.markdown
```diff
@@ -0,0 +1,135 @@
...
+
+For example, if you keep Terraform binaries in `/usr/local/bin` you can clear out the old external binaries like this:
+
+```
+rm /usr/local/bin/terraform-*
+```
+
+External plugin binaries continue to work using the same pattern, but due to updates to the RPC protocol, they will need to be recompiled to be compatible with Terraform v0.7.
```
jen20: Same pattern as 0.6.x

website/source/upgrade-guides/0-7.html.markdown
```diff
@@ -0,0 +1,135 @@
...
+The `concat()` interpolation function used to work for both lists and strings. It now only works for lists.
+
+```
+"${concat(var.foo, "-suffix")}"     # => Error! No longer supported.
+```
+
+Instead, you can use nested interpolation braces for string concatenation.
+
+```
+"${"${var.foo}-suffix"}"
```
jen20: This is kind of ugly - `attribute_name = "${var.foo}-suffix"` should be sufficient without need for nesting braces

website/source/upgrade-guides/0-7.html.markdown
phinze: Ah yes I always forget about good ole `~/.terraform.d`!

website/source/upgrade-guides/0-7.html.markdown
```diff
@@ -133,3 +133,72 @@ resource "aws_instance" "example" {
...
+    eu-west-1 = "ami-789123"
+  }
+}
+```
+
+The key "us-west-2" could be overriden using `-var "amis.us-west-2=overriden_value"` (or equivalent in an environment variable or `tfvars` file). The syntax for this has now changed - instead maps from the command line will be merged with the default value, with maps from flags taking precedence. The syntax for overriding individual values is now:
+
+```
+-var 'amis = { us-west-2 = "overriden_value" }'
+```
```
phinze: I think we should also mention this is possible via the `terraform.tfvars` file & `-var-file`s as well.

website/source/upgrade-guides/0-7.html.markdown
```diff
@@ -0,0 +1,204 @@
...
+
+Terraform v0.7 addresses this by changing the Refresh process that is run during `terraform plan` to always be an in-memory only refresh. New state information detected during this step will not be persisted to permanent state storage.
+
+If the `-out` flag is used to produce a Plan File, the updated state information _will_ be encoded into that file, so that the resulting `terraform apply` operation can detect if any changes occurred that might invalidate the plan.
+
+For most users, this change will not affect your day-to-day usage of Terraform. For users with automation that relies on the old side effect of `plan`, you can use the `terraform refresh` command, which will still persist any changes it discovers.
+
+## Migrating to Data Sources
+
+With the addition of [Data Sources](/docs/configuration/data-sources.html), there are several resources that were acting as Data Sources that are now deprecated.
```
rowleyaj: Is the old way deprecated and will be removed in a future version, or removed in 0.7?

jbardin: @rowleyaj They will print a deprecation warning in 0.7 if you use them a a resource rather than a data source.

jbardin:
Pushed some fixes for all comments so far

jen20:
This looks good to me now all the comments have been addressed!

PR #7903, opened 2016-08-01
Silence log output during tests when not verbose

### LINKED ISSUE #7898: core: modify log output for unit tests
The main terraform binary captures and filters all log output according to the log level, but when running unit tests the log output is unfiltered. 

It would be useful to be able to set output levels during tests in the same manner as the main program (even if that method were to change in the future).

jbardin:
Set the default log package output to iotuil.Discard during tests if the
`-v` flag isn't set. If we are verbose, then apply the filter according
to the TF_LOG env variable.

Fixes #7898

### DIFF
--- a/command/command_test.go
+++ b/command/command_test.go
@@ -1,7 +1,9 @@
 package command
 
 import (
+	"flag"
 	"io/ioutil"
+	"log"
 	"os"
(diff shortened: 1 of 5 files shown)


jen20:
Yesssssss.

PR #7905, opened 2016-08-01
core: Add merge interpolation function

jbardin:
Add a `merge` interpolation function, which merges any number of maps.
Duplicate keys are OK, with the last write winning.

### DIFF
--- a/config/interpolate_funcs_test.go
+++ b/config/interpolate_funcs_test.go
@@ -498,6 +498,103 @@ func TestInterpolateFuncConcat(t *testing.T) {
 	})
 }
 
+func TestInterpolateFuncMerge(t *testing.T) {
+	testFunction(t, testFunctionConfig{
+		Cases: []testFunctionCase{
+			// basic merge
+			{
+				`${merge(map("a", "b"), map("c", "d"))}`,
+				map[string]interface{}{"a": "b", "c": "d"},
+				false,
+			},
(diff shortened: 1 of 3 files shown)


config/interpolate_funcs_test.go
```diff
@@ -498,6 +498,68 @@ func TestInterpolateFuncConcat(t *testing.T) {
...
+				map[string]interface{}{"key1": "a", "key2": "b", "c": "d"},
+				false,
+			},
+
+			// only accept maps
+			{
+				`${merge(map("a", "b"), list("c", "d"))}`,
+				nil,
+				true,
+			},
```
phinze: I'd want to see a map of maps or a list of lists be an explicit test case - I think we need to type switch on the value and error unless it's a string, but maybe it just works. Either way tests will help decide.

jbardin: Oh yes, I'll add that. It does work, but we're not currently checking map value types. Is it OK to let different type maps through? (I guess it would be caught somewhere else if the types weren't allowed)

phinze: Yeah the only thing we can't do yet is an untyped return value. So if it works it's fine to leave.

phinze:
This LGTM but needs docs.

jen20:
LGTM!

PR #7909, opened 2016-08-02
WIP: Adds ability to copy public blob

zachgersh:
Still working on this - want to make sure that private blobs can be copied. Tests may not be great.

- can copy via a filepath up to azure
- can also copy from one storage account to another (if public)

Signed-off-by: Zachary Gershman <zgershman@pivotal.io>


zachgersh:
hey @stack72 would you mind giving this a cursory glance and letting us know if you see any improvements you'd like to see made.

zachgersh:
closing in favor of #7994

PR #7922, opened 2016-08-02
Use absolute paths in Makefile and gofmtcheck.sh

### LINKED ISSUE #7920: Usage of $(CURDIR) in Makefile makes shadow builds impossible
Continuing packaging terraform, trying to do a shadow build (a build in a different directory) for terraform breaks due to usage of `$(CURDIR)` in the Makefile.

Reproduce with `mkdir ~/temp-terraform; cd ~/temp-terraform; make bin -f ~/real-terraform-folder/Makefile`

I recommend setting the following at the top of the makefile and using `$(BASE_DIR)` instead of `$(CURDIR)` wherever it occurs:

```make
MAKEFILE := $(lastword $(MAKEFILE_LIST))
BASE_DIR := $(realpath $(dir $(MAKEFILE)))
```

jleclanche:
Closes #7920

This won't make shadow builds possible just yet but it's an improvement. Incidentally, it also means `gofmtcheck.sh` will always look at the root directory of the project regardless of where it's run from - I don't know if that is the behaviour you want; if it's not I can change it.


phinze:
Thanks for this, @jleclanche!

> This won't make shadow builds possible just yet
> it looks like shadow builds won't be possible at the moment due to other issues in the vendor dependencies

I'm curious - what are the outstanding issues preventing it from working?

jleclanche:
I hit issues in the vendor/ dir I couldn't debug as I'm not too familiar with go. Another issue is all the `go list ./...` calls; I tried looking into how to give `go list` an absolute path (in other words, prevent go list from assuming we care about cwd), but couldn't figure it out -- maybe I'm missing something in its usage.

jen20:
Hi @jleclanche! Unfortunately `realpath` is non-portable - on my OS X machine it does not exist by default. Consequently we can't really consider merging this as-is - I'll close for now. If you can come up with a portable way of achieving this, we can take another look. Thanks for opening a pull request - sorry we can't be of more help at this point!

jleclanche:
@jen20 Do you mean it's not available as part of OSX's Make? It's been a standard part of GNU Make since 3.81 (2006) according to the changelog. I'm looking at an OSX instance right now and it has GNU Make 3.81 installed out of the box.

In shell, I'm not using `realpath` exactly because it's not portable. I'm using `readlink` instead, which you should have.

PR #7933, opened 2016-08-03
website/docs: Adding a list of resources that are available to import now

stack72:


website/source/docs/import/importability.html.md
```diff
@@ -20,3 +20,133 @@ would be grateful.
 
 To make a resource importable, please see the
 [plugin documentation on writing a resource](/docs/plugins/provider.html).
+
+# Currently Available to Import
+
+# AWS
```
radeksimko: How would you feel about turning this and the other headers below into `##` - i.e. 2nd level header? I think it makes sense from semantic point of view.

radeksimko:
I understand the motivation - people want to know what they can import without having to go to the source code or walk through each resource -> hence I see the value, but we desperately need such docs to be (semi-)automatically generated, otherwise this will be painful for ongoing maintenance.

LGTM except the markdown nitpick.

jen20:
Hi @stack72 and @radeksimko! In the interests of getting this out for now while 0.7 is still fresh and people are looking at it I've modified the heading structure and clarified that the Azure resource which can be imported are for resource manager and not ASM. I'd agree that long term we should generate this, however!

PR #7938, opened 2016-08-03
provider/aws: Retry association of IAM Role & instance profile

radeksimko:
This is to fix https://github.com/hashicorp/terraform/issues/2660 which reappears from time to time.

It is obviously difficult to reproduce this issue as it's caused by IAM eventual consistency, but here's a snippet from a debug log from today (0.7.0): https://gist.github.com/radeksimko/1e0533e6c998c55fbb69da9851fcb779
and here's the config used: https://gist.github.com/radeksimko/3891ad1e3b3d295238b04a55dd869c09

### Side effects

Theoretically this may cause small delays for users who are trying to launch an EC2 instance w/ IAM instance profile that _actually_ has no IAM roles assigned to it. Users that use Terraform to manage instance profiles should never run into such situations though as [`aws_iam_instance_profile.roles`](https://www.terraform.io/docs/providers/aws/r/iam_instance_profile.html#roles) is **required** field.

### DIFF
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -366,13 +366,18 @@ func resourceAwsInstanceCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Run configuration: %s", runOpts)
 
 	var runResp *ec2.Reservation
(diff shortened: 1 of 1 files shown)


stack72:
LGTM! Nice one @radeksimko :)

CpuID:
:heart:

PR #5910, opened 2016-03-29
Elasticsearch times out. It takes a while to create.  

grayaii:
Creating Elasticsearch resources via Terraform can be time consuming.

The default timeout values are too low.

Increasing the timeout value greatly increases the success rate of terraform.

There is another discussion in Terraform revolving around having custom timeout for a resource.  This Pull Request is simply to increase the hard coded values until that is resolved.

PR #7944, opened 2016-08-03
Increase timeout yet again for aws elasticsearch resource

grayaii:
We've create hundreds of AWS Elasticsearch resources over the last few months and we get occasional timeout failures from AWS.  This PR is to increase the timeout once again. I did it before:
https://github.com/hashicorp/terraform/pull/5910/files
But we've seen enough timeouts from AWS on this resource that increasing the timeout seems like the only solution.


jen20:
Sad that we have to do this, but this is entirely congruent to my experience @grayaii :-1:

PR #7945, opened 2016-08-03
fix awkward wording of atlas token generation

kylewest:
Too many "to"s (and other prepositions) made this hard to skim.


jen20:
Thanks @kylewest - I agree this is an improvement.

PR #7964, opened 2016-08-04
Update AMI ID on documentation

szollo:
It looks like the AMI ID for Ubuntu 16.04 used on the official documentation is outdated and not working anymore, so I have changed it with the current AMI ID of a Ubuntu 16.04 image.

This is what I get when attempting to use the one currently on the website:

```
Error applying plan:

1 error(s) occurred:

* aws_instance.example: Error launching source instance: InvalidAMIID.NotFound: The image id '[ami-2106ed4c]' does not exist
	status code: 400, request id:
```

PS. Such an awesome tool, keep up the great work!

### DIFF
--- a/website/source/intro/getting-started/change.html.md
+++ b/website/source/intro/getting-started/change.html.md
@@ -33,7 +33,7 @@ resource "aws_instance" "example" {
 }
 ```
 
-~> **Note:** EC2 Classic users please use AMI `ami-2106ed4c` and type `t1.micro`
+~> **Note:** EC2 Classic users please use AMI `ami-656be372` and type `t1.micro`
 
 We've changed the AMI from being an Ubuntu 14.04 LTS AMI to being
 an Ubuntu 16.04 LTS AMI. Terraform configurations are meant to be


website/source/intro/getting-started/change.html.md
```diff
@@ -33,7 +33,7 @@ resource "aws_instance" "example" {
 }
 ```
 
-~> **Note:** EC2 Classic users please use AMI `ami-2106ed4c` and type `t1.micro`
+~> **Note:** EC2 Classic users please use AMI `ami-8ac04ff9` and type `t2.micro`
```
stack72: We have to leave t1 instances in the README for EC2 Classic Users. T2 instances can only be launched into an EC2-VPC environment

stack72:
Hi @szollo 

Thanks for the PR here. As I noted in the comment, it changes the note for t1 to t2 and would break EC2-Classic Users. From my own investigation, I can still get that AMI (it's for 15.04 it seems) 

If you want to update the PR to only change the AMI-ID for the correct region (us-west-2) then that would be great

Thanks

Paul

szollo:
Hi @stack72,

You're totally right - I didn't think about the fact that AMI IDs change depending on the region. 

I was in eu-west-1, and looking at us-east-1 the AMI ID does exist -- us-east-1 being the region configured at the beginning of the tutorial. That being said, the AMI is for Ubuntu 15.04, as you mentioned, rather than the 16.04 the page refers to.

I pushed a change to fix this, so that it gets a 16.04 AMI instead (`ami-656be372`), and left it as `t1.micro` due to it being EC2-Classic.

Cheers
Stevie

jen20:
Thanks @szollo!

Issue #7959, opened 2016-08-03
aws_availability_zones gets unavailable zones as well


dene14:
Hi there!

I've got the problem:
* aws_subnet.private_1: Error creating subnet: InvalidParameterValue: Value (us-east-1a) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-east-1c, us-east-1b, us-east-1d, us-east-1e.

AZ were taken by:
data.aws_availability_zones.available.names[0]
data.aws_availability_zones.available.names[1]

The problem is that Amazon names AZ individually and uniquely for every single AWS account. Also, Amazon marks some particular zones for retirement from time to time, that basically means that it's not possible to create any new instance in that zone. And new zones usually added to account. For long-lived (5 or more years old) accounts it's really common thing.

For me now it looks like "aws_availability_zones.available" doesn't really produce a list of available, thus should be considered as a bug.

Thanks!

kwilczynski:
This one is a bit tricky one. I haver never really found a reliable way to detect this without running into an issue first (every time I added a new region). Different accounts get different access (depends often on per/post VPC-enabled availability zones, etc.), as you noted, and to my knowledge there is no API (unless this has changed?) that allows you to get a list of zones your account has an access too.

The data source uses the "DescribeAvailabilityZones", which is the API call used, can have a filter set to select one of the possible states for the availability zones, for example: available, information, impaired and unavailable: where, for instance, the "impaired" is often set when a zone has na outage.

This sadly setting a filter to e.g. "available", does not really solve the problem that most of people run into, for example:

```
Value (us-east-1a) for parameter availabilityZone is invalid. Subnets can currently only be created in the following  availability zones: us-east-1c, us-east-1b, us-east-1d.
```

People came up with a clever ways to devise which one affect them, for example: http://stackoverflow.com/questions/22744467/vpc-capable-availability-zones-in-amazon/22812138#22812138

But it does not really solve the problem for most people.

stack72:
Closed via #7965

PR #7965, opened 2016-08-04
Add state filter to aws_availability_zones data source.

kwilczynski:
This commit adds an ability to filter Availability Zones based on state.

Be advised that this does not always works reliably for an older accounts which
have been created in the pre-VPC era of EC2. These accounts tends to retrieve
availability zones that are not VPC-enabled, thus creation of a custom subnet
within such Availability Zone would result in a failure.

### DIFF
--- a/builtin/providers/aws/data_source_availability_zones.go
+++ b/builtin/providers/aws/data_source_availability_zones.go
@@ -21,6 +21,12 @@ func dataSourceAwsAvailabilityZones() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"state": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "available",
+				ValidateFunc: validateStateType,
+			},
 		},
 	}
 }
@@ -28,25 +34,52 @@ func dataSourceAwsAvailabilityZones() *schema.Resource {
 func dataSourceAwsAvailabilityZonesRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ec2conn
 
-	log.Printf("[DEBUG] Reading availability zones")
+	log.Printf("[DEBUG] Reading Availability Zones.")
 	d.SetId(time.Now().UTC().String())
 
-	req := &ec2.DescribeAvailabilityZonesInput{DryRun: aws.Bool(false)}
-	azresp, err := conn.DescribeAvailabilityZones(req)
+	request := &ec2.DescribeAvailabilityZonesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name:   aws.String("state"),
+				Values: []*string{aws.String(d.Get("state").(string))},
+			},
+		},
+	}
+	log.Printf("[DEBUG] Availability Zones request options: %#v", *request)
+
+	resp, err := conn.DescribeAvailabilityZones(request)
 	if err != nil {
-		return fmt.Errorf("Error listing availability zones: %s", err)
+		return fmt.Errorf("Error fetching Availability Zones: %s", err)
 	}
 
-	raw := make([]string, len(azresp.AvailabilityZones))
-	for i, v := range azresp.AvailabilityZones {
+	raw := make([]string, len(resp.AvailabilityZones))
+	for i, v := range resp.AvailabilityZones {
 		raw[i] = *v.ZoneName
 	}
 
 	sort.Strings(raw)
 
 	if err := d.Set("names", raw); err != nil {
-		return fmt.Errorf("[WARN] Error setting availability zones")
+		return fmt.Errorf("[WARN] Error setting Availability Zones: %s", err)
 	}
 
 	return nil
+}
+
+func validateStateType(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+
+	validState := map[string]bool{
+		"available":   true,
+		"information": true,
+		"impaired":    true,
+		"unavailable": true,
+	}
+
+	if !validState[value] {
+		errors = append(errors, fmt.Errorf(
+			"%q contains an invalid Availability Zone state %q. Valid states are: %q, %q, %q and %q.",
+			k, value, "available", "information", "impaired", "unavailable"))
+	}
+	return
 }
--- a/builtin/providers/aws/data_source_availability_zones.go
+++ b/builtin/providers/aws/data_source_availability_zones.go
@@ -24,7 +24,6 @@ func dataSourceAwsAvailabilityZones() *schema.Resource {
 			"state": &schema.Schema{
 				Type:         schema.TypeString,
 				Optional:     true,
-				Default:      "available",
 				ValidateFunc: validateStateType,
 			},
 		},
@@ -37,14 +36,17 @@ func dataSourceAwsAvailabilityZonesRead(d *schema.ResourceData, meta interface{}
 	log.Printf("[DEBUG] Reading Availability Zones.")
 	d.SetId(time.Now().UTC().String())
 
-	request := &ec2.DescribeAvailabilityZonesInput{
-		Filters: []*ec2.Filter{
+	request := &ec2.DescribeAvailabilityZonesInput{}
+
+	if v, ok := d.GetOk("state"); ok {
+		request.Filters = []*ec2.Filter{
 			&ec2.Filter{
 				Name:   aws.String("state"),
-				Values: []*string{aws.String(d.Get("state").(string))},
+				Values: []*string{aws.String(v.(string))},
 			},
-		},
+		}
 	}
+
 	log.Printf("[DEBUG] Availability Zones request options: %#v", *request)
 
 	resp, err := conn.DescribeAvailabilityZones(request)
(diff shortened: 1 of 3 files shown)


kwilczynski:
Resolves #7959.

kwilczynski:
Test is passing:

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAvailabilityZones_basic'
2016/08/04 10:52:29 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAvailabilityZones_basic -timeout 120m
=== RUN   TestAccAWSAvailabilityZones_basic
--- PASS: TestAccAWSAvailabilityZones_basic (23.28s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	23.303s
```

kwilczynski:
The debug output confirms that state is set:

```
data.aws_availability_zones.availability_zones:
  ID = 2016-08-04 01:35:06.34062534 +0000 UTC
  names.# = 3
  names.0 = us-west-2a
  names.1 = us-west-2b
  names.2 = us-west-2c
  state = available
```

kwilczynski:
@stack72 over to you 🚀

builtin/providers/aws/data_source_availability_zones.go
```diff
@@ -21,32 +21,65 @@ func dataSourceAwsAvailabilityZones() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"state": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "available",
```
apparentlymart: I think setting this default would make it impossible to retrieve an unfiltered list, and is a backward-compatibility break.

Could we instead default to no filter, and have the user explicitly set `state = "available"` to get the filtered list? This does make the user do a little more work, but has the advantage of being consistent with how some of our other data sources are working (each new attribute acts as an additional filter constraint) and makes things a little more explicit.

kwilczynski: @apparentlymart makes sense! I will push an update shortly.

kwilczynski:
Test is passing:

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAvailabilityZones_basic'
2016/08/05 06:30:40 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAvailabilityZones_basic -timeout 120m

=== RUN   TestAccAWSAvailabilityZones_basic
--- PASS: TestAccAWSAvailabilityZones_basic (35.80s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	35.822s
```

kwilczynski:
@stack72 @apparentlymart over to you 🚀

website/source/docs/providers/aws/d/availability_zones.html.markdown
```diff
@@ -35,10 +35,15 @@ resource "aws_subnet" "secondary" {
...
+current state. Can be either `"available"`, `"information"`, `"impaired"` or
+`"unavailable"`. By default the list includes a complete set of Availability Zones
+to which the underlying AWS account has access, regardless of their state.
 
 ## Attributes Reference
 
 The following attributes are exported:
 
-* `names` - A list of the availability zone names available to the account.
+* `names` - A list of the Availability Zone names available to the account.
```
apparentlymart: It's a little weird to capitalize non-proper nouns in English... what was the motivation for this change here?

kwilczynski: Just to reflect what AWS does in their documentation, which might not be correct convention (since I am not an English language native speaker, etc., I have assumed that their style is somewhat acceptable). For example, as per: [Regions and Availability Zones](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html).

apparentlymart: Hmm interesting... I'd not noticed that before. I think we've generally treated "availability zones" as a general noun rather than a proper noun elsewhere in our docs, so it might be a bit late for us to switch to the AWS docs style now without a lot of updates on other doc pages. Not a big deal though, certainly.

kwilczynski: Sorry for causing troubles!

In terms of consistency in documentation we have, it does vary:

(not a very scientific comparison, though)

```
krzysztof@apple ~/Development/Projects/Go/src/github.com/hashicorp/terraform (master %)$ ack --no-smart-case availability website/source/ -l | wc -l
      45
krzysztof@apple ~/Development/Projects/Go/src/github.com/hashicorp/terraform (master %)$ ack --no-smart-case Availability website/source/ -l | wc -l
      11
```

Would you like ma to revert this back?

stack72: @apparentlymart we can merge this one for now and then follow up on the docs later :)

stack72:
```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAvailabilityZones_'
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAvailabilityZones_ -timeout 120m
=== RUN   TestAccAWSAvailabilityZones_basic
--- PASS: TestAccAWSAvailabilityZones_basic (109.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	109.018s
```

cmusser:
With 0.7.5, I tried using the following, hoping to use the second AZ in the us-west-1 region in a subnet:

```
data "aws_availability_zones" "available" {
  state = "available"
}

resource "aws_subnet" "subnet-public-2" {
  vpc_id = "${aws_vpc.vpc-standalone.id}"
  availability_zone = "${data.aws_availability_zones.available.names[1]}"
  cidr_block = "${var.vpc_public_subnet_2}"
  tags {
    "Name" = "${var.vpc_name} Public 2"
    "VPC" = "${var.vpc_name}"
  }
}
```

It tried to insert "us-west-1b" (index 1), which isn't available. Did I do this incorrectly or is this fix not available in the version I'm using?

kwilczynski:
@cmusser hi there! I am sorry that you are having issues! Let me have a look what is going on here.

kwilczynski:
@cmusser hi there again!

I have had a look, and for example, for my account Terraform returns the right set of results:

```
$ terraform show
data.aws_availability_zones.available:
  id = 2016-10-10 23:04:29.858510486 +0000 UTC
  names.# = 2
  names.0 = us-west-1b
  names.1 = us-west-1c

$ aws ec2 describe-availability-zones --region us-west-1
{
    "AvailabilityZones": [
        {
            "State": "available",
            "RegionName": "us-west-1",
            "Messages": [],
            "ZoneName": "us-west-1b"
        },
        {
            "State": "available",
            "RegionName": "us-west-1",
            "Messages": [],
            "ZoneName": "us-west-1c"
        }
    ]
}
```

Which is consistent with the AWSCLI output for the given (same) set of credentials. Basically, this is all my IAM user sees in terms of what my account has an access to.

Also, `us-west-1b` seem to be available.

Is there something I am missing? I am not sure if I got the gist of the problem right.

cmusser:
There is something funny going on, perhaps with AWS itself.  I see the same thing you did with AWS CLI: 1b appears available. So it's reasonable that Terraform places it in the list. The specific error when creating the subnet with the list of available AZs is:

```
* aws_subnet.subnet-public-2: Error creating subnet: InvalidParameterValue: Value (us-west-1b) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-west-1a, us-west-1c.
```

Maybe its somehow possible for them to return a zone as available, when it really isn't for subnets.

kwilczynski:
@cmusser the Availability Zones are very elusive. From the AWS documentation (excerpt):

```
As Availability Zones grow over time, our ability to expand them can become constrained. If this happens, we might restrict you from launching an instance in a constrained Availability Zone unless you already have an instance in that Availability Zone. Eventually, we might also remove the constrained Availability Zone from the list of Availability Zones for new customers. Therefore, your account might have a different number of available Availability Zones in a region than another account.
```

See: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html

I have to accounts and both have different availability zones reported. The issue with subnets is quite a common one. Sadly, there is no reliable way to detect this (or at least I am not familiar with a possible solution and/or workaround), especially as there is no API from Amazon which would provide such details e.g. which Availability Zones are VPC-enabled and allow subnets creation, etc.

Related reading material:

- https://forums.aws.amazon.com/thread.jspa?messageID=369862
- http://stackoverflow.com/questions/22744467/vpc-capable-availability-zones-in-amazon

I am sorry that I have no better solution for you yet.

We could, in theory, have a "VPC subnet probe" which would make an attempt to create a VPC then create a subnet in each of the Availability Zones, and catch failure/success in turn assembling a list of ones which support subnets creation. But this would be rather slow.

pmoust:
Hitting the same solution, and I am afraid the VPC subnet probe is the only way to actually determine if an AZ is available.

Is there an issue for this?

kwilczynski:
@pmoust hi there!

I agree. And it is rather annoying to stumble into. I proposed the VPC probe, but was seen as not necessarily the right thing to do (albeit, there isn't any other that works 100% successfully), and I can see why as it would require to attempt to provision a VPC with a subnet which might not necessarily be the best thing to do e.g. can be affected by account limits, etc.

What do you think, @stack72 and @radeksimko?

pmoust:
@kwilczynski I have contacted AWS via our TAMs, and can confirm there is no other sane way at the moment.
Perhaps adding another parameter like a boolean `ensure_usable` (naming is hard, heh) that does the dummy VPC probe and catches that specific exception only?

PR #7977, opened 2016-08-04
numeric variables aren't always interpreted as str

jbardin:
If we have a number value in our config variables, format it as a
string, and send it with the HCL=true flag just in case.

Also use %g for for float encoding, as the output is a generally a
little friendlier.


phinze:
Code LGTM - how much work would it be for a covering test to repro the prior badness?

jbardin:
Made a better test, and stripped off the extra `\n` from variables, which were being printed in the atlas UI.

phinze:
Solid - LGTM!

PR #7979, opened 2016-08-04
update github.com/mitchellh/copystructure

jbardin:
Patched a panic where copystructure tries to set an unexported struct
field.


stack72:
LGTM @jbardin :)

PR #7984, opened 2016-08-04
Datasource for AWS and Fastly IP ranges

yawn:
This will allow the referencing of IP ranges like so:

```
provider "aws" {
  region = "eu-west-1"
}

provider "fastly" {
  api_key = "test"
}

data "fastly_ip_ranges" "fastly" {
}

data "aws_ip_ranges" "cloudfront" {
  services = [ "cloudfront" ]
}

data "aws_ip_ranges" "healthchecks" {
  services = [ "route53_healthchecks" ]
}

data "aws_ip_ranges" "european_ec2" {
  services = [ "ec2" ]
  regions = [ "eu-west-1", "eu-central-1" ]
}

resource "aws_security_group" "from_cloudfront" {

  name = "from_cf"

  ingress {
    from_port = "443"
    to_port = "443"
    protocol = "tcp"
    cidr_blocks = [ "${data.aws_ip_ranges.cloudfront.cidr_blocks}" ]
  }

  tags {
    CreateDate = "${data.aws_ip_ranges.cloudfront.create_date}"
    SyncToken = "${data.aws_ip_ranges.cloudfront.sync_token}"
  }

}

resource "aws_security_group" "from_europe" {

  name = "from_europe"

  ingress {
    from_port = "443"
    to_port = "443"
    protocol = "tcp"
    cidr_blocks = [ "${data.aws_ip_ranges.european_ec2.cidr_blocks}" ]
  }

  tags {
    CreateDate = "${data.aws_ip_ranges.european_ec2.create_date}"
    SyncToken = "${data.aws_ip_ranges.european_ec2.sync_token}"
  }

}

resource "aws_security_group" "from_fastly" {

  name = "from_fastly"

  ingress {
    from_port = "443"
    to_port = "443"
    protocol = "tcp"
    cidr_blocks = [ "${data.fastly_ip_ranges.fastly.cidr_blocks}" ]
  }

}

resource "aws_security_group" "from_r53" {

  name = "from_r53"

  ingress {
    from_port = "443"
    to_port = "443"
    protocol = "tcp"
    cidr_blocks = [ "${data.aws_ip_ranges.healthchecks.cidr_blocks}" ]
  }

  tags {
    CreateDate = "${data.aws_ip_ranges.healthchecks.create_date}"
    SyncToken = "${data.aws_ip_ranges.healthchecks.sync_token}"
  }

}
```

If you like this contribution, I'll

* [X] add documentation
* [x] add tests
* [X] rename `blocks` to `cidr_blocks`

The `main.tf` referenced above should already be `apply`able.

### DIFF
--- a/builtin/providers/aws/data_source_aws_ip_ranges.go
+++ b/builtin/providers/aws/data_source_aws_ip_ranges.go
@@ -6,8 +6,8 @@ import (
 	"io/ioutil"
 	"log"
 	"sort"
+	"strconv"
 	"strings"
-	"time"
 
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -50,7 +50,7 @@ func dataSourceAwsIPRanges() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"sync_token": &schema.Schema{
-				Type:     schema.TypeString,
+				Type:     schema.TypeInt,
 				Computed: true,
 			},
 		},
@@ -62,7 +62,6 @@ func dataSourceAwsIPRangesRead(d *schema.ResourceData, meta interface{}) error {
 	conn := cleanhttp.DefaultClient()
 
 	log.Printf("[DEBUG] Reading IP ranges")
-	d.SetId(time.Now().UTC().String())
 
 	res, err := conn.Get("https://ip-ranges.amazonaws.com/ip-ranges.json")
 
@@ -88,7 +87,15 @@ func dataSourceAwsIPRangesRead(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error setting create date: %s", err)
 	}
 
-	if err := d.Set("sync_token", result.SyncToken); err != nil {
+	syncToken, err := strconv.Atoi(result.SyncToken)
+
+	if err != nil {
+		return fmt.Errorf("Error while converting sync token: %s", err)
+	}
+
+	d.SetId(result.SyncToken)
+
+	if err := d.Set("sync_token", syncToken); err != nil {
 		return fmt.Errorf("Error setting sync token: %s", err)
 	}
 
--- a/builtin/providers/aws/data_source_aws_ip_ranges.go
+++ b/builtin/providers/aws/data_source_aws_ip_ranges.go
@@ -30,7 +30,7 @@ func dataSourceAwsIPRanges() *schema.Resource {
 		Read: dataSourceAwsIPRangesRead,
 
 		Schema: map[string]*schema.Schema{
-			"blocks": &schema.Schema{
+			"cidr_blocks": &schema.Schema{
 				Type:     schema.TypeList,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
@@ -126,7 +126,7 @@ func dataSourceAwsIPRangesRead(d *schema.ResourceData, meta interface{}) error {
 
 	sort.Strings(prefixes)
 
-	if err := d.Set("blocks", prefixes); err != nil {
+	if err := d.Set("cidr_blocks", prefixes); err != nil {
 		return fmt.Errorf("Error setting ip ranges: %s", err)
 	}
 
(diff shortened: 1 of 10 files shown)


phinze:
@yawn this is _very cool!!_ We'll definitely pull this in when the docs and tests are good to go. Thanks for your work on this! 👍 💟

apparentlymart:
Agreed that this is awesome. Thanks so much for this!

Perhaps we could rename the `blocks` attribute on each of these to `cidr_blocks` for consistency with the way we've named this attribute elsewhere, and to make it super-obvious that these blocks are presented in CIDR notation.

yawn:
Not a problem (the renaming to `cidr_blocks`).

One more question: I'm leaning towards making it an [error](https://github.com/hashicorp/terraform/pull/7984/files#diff-7dc808e23a18e26db7a2c58340f15df4R116) if the selected filters yield *no* blocks in the case of the AWS IP ranges but I'm sure if that fits the principle of least surprise (in the context of Terraform). Any opinions regarding this?

apparentlymart:
@yawn for several data sources we've been making it an error if the filters are over- or under-constrained. In this case, where the result is a list anyway, that seems less obviously correct since an empty list is a valid result.

However, I think I agree that returning no blocks at all, and then in turn e.g. creating a security group that has no blocks, is likely an indication of user error rather than a useful result, so I'm leaning towards agreeing about that being an error at least until we discover there's a use-case for it.

yawn:
* Failure on empty results done in 2caae49a3b773217e8efff95fcf2345031e947dc
* Documentation added in 67bf13fccf21d7c4411b69761d58fd53c222631d

Test will come later.

Feel free to review, especially the `SetId` selection in 67b4b4cbfb453ceba45d39e9c43e187448843d98 and e21203a36973ef3dd7ec47b8202da6ea96a3e1f0 about which I'm unsure. What's the semantics of ID's for data sources?

builtin/providers/aws/data_source_aws_ip_ranges.go
```diff
@@ -0,0 +1,135 @@
...
+		return fmt.Errorf("Error setting create date: %s", err)
+	}
+
+	syncToken, err := strconv.Atoi(result.SyncToken)
+
+	if err != nil {
+		return fmt.Errorf("Error while converting sync token: %s", err)
+	}
+
+	d.SetId(result.SyncToken)
```
apparentlymart: Right now, due to some interactions with some core Terraform code that's common to both data and managed resources, it's required to set an id for a data resource but it doesn't really matter what it's set to... so this should be fine as long as `result.SyncToken` is guaranteed to never be the empty string.

builtin/providers/aws/data_source_aws_ip_ranges.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if err := d.Set("sync_token", syncToken); err != nil {
+		return fmt.Errorf("Error setting sync token: %s", err)
+	}
+
+	var (
+		regions         = d.Get("regions").(*schema.Set)
+		services        = d.Get("services").(*schema.Set)
```
apparentlymart: Should we normalize the strings in these sets to lowercase too, so that a user can safely use the uppercase values given in the AWS docs? Otherwise these presumably won't match in the `regions.Contains` and `services.Contains` checks below.

builtin/providers/aws/data_source_aws_ip_ranges.go
yawn: Should we perhaps declare some constant that manifests this fact (e.g. `DefaultDataSourceId`) and start to use that?

builtin/providers/aws/data_source_aws_ip_ranges.go
yawn: Unsure. I believe most users will expect the regions and services to be lower case (since that's the default in TF everywhere else) and will probably not know the AWS API for IP ranges in detail.

Anybody else has a stronger opinion here? Since it won't hurt I can easily add it ...

yawn: Note to self: no need to check length of required arguments.

yawn:
Ready for review.

apparentlymart:
Thanks for this, @yawn. I expect it will be really useful!

yawn:
Glad to be of service @apparentlymart!

Issue #7980, opened 2016-08-04
terraform crash with encrypted aws_ami_copy


rbuels:
Terraform 0.7.0 crashed for me just now while it was creating an `aws_ami_copy` with the `encrypted` and `kms_key_id` options set.

It crashed after exactly 5 minutes, the default connection timeout, which might be significant.

The console output was:
```
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (3m30s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (3m40s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (3m50s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (4m0s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (4m10s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (4m20s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (4m30s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (4m40s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (4m50s elapsed)
module.ecs_cluster.aws_ami_copy.ecs_ami: Still creating... (5m0s elapsed)
Error applying plan:

1 error(s) occurred:

* aws_ami_copy.ecs_ami: unexpected EOF

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
panic: runtime error: invalid memory address or nil pointer dereference
2016/08/04 11:35:24 [DEBUG] plugin: terraform: [signal 0xb code=0x1 addr=0x0 pc=0x370914]
2016/08/04 11:35:24 [DEBUG] plugin: terraform: 
2016/08/04 11:35:24 [DEBUG] plugin: terraform: goroutine 569 [running]:
2016/08/04 11:35:24 [DEBUG] plugin: terraform: panic(0x2fb1ba0, 0xc8200120f0)
2016/08/04 11:35:24 [DEBUG] plugin: terraform: 	/opt/go/src/runtime/panic.go:481 +0x3e6
[...]

PR #7988, opened 2016-08-04
docs/aws: Add  and  as docs to the AWS  resource

stack72:


rbuels:
Might have a look at #7980 in case it's related to the original change to ami.

PR #7989, opened 2016-08-04
Override atlas variables even if they aren't local

jbardin:
Some Atlas usage patterns expect to be able to override a variable set
in Atlas, even if it's not seen in the local context. This allows
overwriting a variable that is returned from atlas, and sends it back.


phinze:
LGTM

PR #7995, opened 2016-08-05
Fix. Handle lack of snapshot ID for a volume.

kwilczynski:
This commit resolves the issue where lack of snapshot ID in the device mapping
section of the API response to DescribeImagesResponse would cause Terraform to
crash due to a nil pointer dereference. Usually, the snapshot ID is included,
but in some unique cases (e.g. ECS-enabled AMI from Amazon available on the
Market Place) a volume that is attached might not have it.

The API documentation does not clearly define whether the snapshot ID either
should be or must be included for any volume in the response.

### DIFF
--- a/builtin/providers/aws/resource_aws_ami.go
+++ b/builtin/providers/aws/resource_aws_ami.go
@@ -174,12 +174,15 @@ func resourceAwsAmiRead(d *schema.ResourceData, meta interface{}) error {
 				"delete_on_termination": *blockDev.Ebs.DeleteOnTermination,
 				"encrypted":             *blockDev.Ebs.Encrypted,
 				"iops":                  0,
-				"snapshot_id":           *blockDev.Ebs.SnapshotId,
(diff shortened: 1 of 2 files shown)


kwilczynski:
Resolves #7980.

kwilczynski:
Test is passing:

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAMICopy'
2016/08/05 11:04:43 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAMICopy -timeout 120m
=== RUN   TestAccAWSAMICopy
--- PASS: TestAccAWSAMICopy (503.22s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	503.248s
```

kwilczynski:
@stack72 over to you 🚀

stack72:
Thanks for this :) This LGTM!

PR #8008, opened 2016-08-05
Don't send access_token in request params

jbardin:
Always send the access_token in the X-Atlas-Token header.


phinze:
This LGTM - there's also an Atlas client in the `terraform push` path and in the Atlas provider. Are we good in both those spots as well?

jbardin:
I think so ... this package is the only place we ever set `access_token` (my research consisting of `git grep`), let me double check

jbardin:
Look OK. I don't see it set anywhere else, and verified that the `tf push` request sends the header

phinze:
Nice thanks for checking. LGTM then!

Issue #8014, opened 2016-08-05
-state/-state-out options are ignored when remote state is used.


clstokes:
When my local environment is configured for Terraform remote state and I try to use `-state` or `-state-out` options for plans or applies, the `-state/state-out` options are ignored. Terraform should exit with an error if they're specified _and_ remote state is configured to prevent confusing behavior (or lack of behavior).

### Terraform Version
- `v0.7.0`

### Affected Resource(s)
- _Core functionality_

### Terraform Configuration Files
_Any config works..._
```hcl
resource "aws_vpc" "main" {
  cidr_block = "10.0.0.0/16"
  tags { Name = "tf-remote-and-local-state" }
}
```

### Output
https://gist.github.com/clstokes/8dcc9dde1248e7362ec052721e062e3d

### Expected Behavior
I expected the `-state/state-out` options to override my configured remote state.

### Actual Behavior
The `-state/state-out` options were ignored.

### Steps to Reproduce
1. `terraform remote config -backend-config=name=YOUR_ORG/YOUR_ENV`
1. `terraform apply -state-out=terraform apply -state-out=my-custom-state.tfstate`
1. # observe `terraform apply -state-out=my-custom-state.tfstate` is not created and `.terraform/terraform.tfstate` was used.

### Important Factoids
_none_

MiLk:
Instead of ignoring the `-state/state-out` options, I would like to be able to specify the destination of the local cache of my remote state (currently `.terraform/terraform.tfstate`).

I could run something like that:
```bash
terraform remote config $(REMOTE_CONFIG) \
	-backend-config="key=staging/project1.tfstate" \
	-state=states/staging/project1.tfvars
terraform apply \
	-var-file=variables/staging/project1.tfvars \
	projects/project1

terraform remote config $(REMOTE_CONFIG) \
	-backend-config="key=staging/project2.tfstate" \
	-state=states/staging/project2.tfvars
terraform apply \
	-var-file=variables/staging/project2.tfvars \
	projects/project2

terraform remote config $(REMOTE_CONFIG) \
	-backend-config="key=production/project1.tfstate" \
	-state=states/production/project1.tfvars
terraform apply \
	-var-file=variables/production/project1.tfvars \
	projects/project1
```

Currently, I have to clean the `.terraform` folder every time I change the active project/environment.

gerr1t:
+1

voroniys:
@MiLk for what you need you can use workspaces instaed of -state-out option. 

[...]

PR #8015, opened 2016-08-05
docs/commands: clarify -state w/ remote state

clstokes:
Clarifies usage of `-state` and `-state-out` when remote state is used. Not a fix for #8014, but related to it.


stack72:
Thanks @clstokes :)

PR #8016, opened 2016-08-05
New Swift Object Storage provider (with docs)

renier:
_Why not use OpenStack's provider's object storage resources?_

OpenStack's provider uses gophercloud, which currently
assumes that authentication will be managed by a separate
identity server (e.g. Keystone). This is a problem
for cases in which there is no such server. For example,
if you are running a Swift installation standalone.

A solution would be to make fundamental changes to
gophercloud's authentication flow that would allow
direct authentication to OpenStack services that support
it (e.g. Swift), get those changes upstream, then
integrate that new capability into the OpenStack
provider here.

*Or* simply build a new provider using a go library
(i.e. ncw/swift) that focuses on Swift only with support
for direct authentication already.


jtopjian:
@renier Thank you for submitting this!

I know we discussed this a little bit on the mailing list. I'm in complete support of accessing Swift environments that don't use Keystone, so I think this is a great addition.

One thing I want to understand more is if authentication is the only difference between this provider and the Gophercloud-based OpenStack provider. If Gophercloud ever added support for swift authentication, would that satisfy this provider's goal?

Just to do some due diligence, I'm going to ask the Gophercloud team if Swift authentication is ever a possibility. If it is a possibility, and if it's ever added, then the only difference between this provider and the existing OpenStack provider would be the underlying library used, right?

Another thing I'd like to understand more is how tricky it would be to integrate `ncw/swift` into the existing OpenStack provider. For example, if new swift-only fields were added to the provider, could the existing swift resources conditionally use either gophercloud or ncw? Or could the existing swift resources be rewritten to only use ncw (since it looks like ncw also supports Keystone v2 and v3)?

Please don't take these questions as trying to negate this provider - I really am in support of it! I just want to make sure it's added in the best way possible for long-term use.

renier:
@jtopjian Great questions, Joe.

>One thing I want to understand more is if authentication is the only difference between this provider and the Gophercloud-based OpenStack provider. If Gophercloud ever added support for swift authentication, would that satisfy this provider's goal?

Yes, it would satisfy what was needed and then we could use the `openstack_objectstroage_*` resources. Of course, the openstack provider would need to be adapted to that.

>Just to do some due diligence, I'm going to ask the Gophercloud team if Swift authentication is ever a possibility. If it is a possibility, and if it's ever added, then the only difference between this provider and the existing OpenStack provider would be the underlying library used, right?

Well, the possibility existing doesn't change anything (currently). Also, gophercloud does a lot more than just Swift. But if the possibility becomes reality, the only thing missing would be the openstack terraform provider adapting to this. Then this Swift provider would not be needed anymore.

>Another thing I'd like to understand more is how tricky it would be to integrate ncw/swift into the existing OpenStack provider. For example, if new swift-only fields were added to the provider, could the existing swift resources conditionally use either gophercloud or ncw? Or could the existing swift resources be rewritten to only use ncw (since it looks like ncw also supports Keystone v2 and v3)?

I do think that path would be more complex (selectively and complementary using ncw or sticking with gophercloud depending on the type of auth you want to use for the object storage resources...). Ideally, gophercloud just changes to allow that intrinsically.

If gophercloud ever enables this, like I said in the ML, I'm willing to work on the openstack provider changes.

jtopjian:
> Well, the possibility existing doesn't change anything (currently).

Yup, understood, and I don't intent to leave this unresolved indefinitely!

> But if the possibility becomes reality, the only thing missing would be the openstack terraform provider adapting to this.

I'd be more than happy to quickly adopt this into Terraform.

> I do think that path would be more complex

Agreed. 

To be honest, I'm trying to avoid the scenario where a separate Swift provider is merged into Terraform and then later becomes redundant. I think that would ultimately confuse users. 

I've done a cursory look through Gophercloud and I don't think this would be too difficult to implement, but before I went off building a standalone Swift environment and started hacking away, I've opened an Issue at `gophercloud/gophercloud` and asked how it could be done as well as if it's a suitable addition to Gophercloud.

If adding this functionality to Gophercloud ultimately won't work, another idea would be to write a Go library that acts as a dispatcher / proxy for the two libraries. The existing object and container resources would implement this and only have to use one library rather than two on a conditional basis.

As a side note, in the meantime, if you'd like to use this functionality in Terraform without having to use your own patched copy, are you familiar with making a standalone provider? You'd basically just be rearranging the files you've already created into a different structure. Doing a search on Github for "terraform provider" will yield a bunch of results you can use as an example.

renier:
>To be honest, I'm trying to avoid the scenario where a separate Swift provider is merged into Terraform and then later becomes redundant. I think that would ultimately confuse users.

The hard thing to do is to make a decision based on a prediction. You can always remove a provider from the tree later. Also, documentation should help to clarify any confusion.

>As a side note, in the meantime, if you'd like to use this functionality in Terraform without having to use your own patched copy, are you familiar with making a standalone provider?

I've been wondering about this. Hashi is currently slammed right now with over 180 PRs. Most of them about providers. I've just confirmed that you can override any existing internal provider with your own newer version through `~/.terraformrc`. So why not have a command like `terraform-provider-get` that installs/configures a any third-party provider you want? That way Hashi doesn't need to bother with a lot of PRs (like this one).

I think that will be my next contrib :-)

jtopjian:
@renier Just a quick update on this. You can see the Gophercloud conversation referenced above for complete detail.

The actual API calls to Swift are decoupled from an authenticated client. This is really good news because it means that Gophercloud is not completely reliant on a Keystone authenticated client. The tricky part is adding the `x-auth-key` and `x-auth-user` headers into the request. There's currently not a quick and easy way this is exposed to the user. I'm going to be investigating the ideas described in the Gophercloud issue.

jtopjian:
@renier Quick update on this: I have a pull request into Gophercloud to enable Swauth support: https://github.com/gophercloud/gophercloud/pull/65.

The only downside is that this will be included in `gophercloud/gophercloud` which Terraform still needs to move to. Probably another month for that to happen. That's not a _terribly_ long time to wait, though.

jtopjian:
And that's been merged, so Gophercloud now natively supports Swift authentication :)

Integrating this into the Terraform provider will be at the top of my list once I get Terraform moved over to `gophercloud/gophercloud`.

@renier Let me know if you have any comments or thoughts.

renier:
@jtopjian Thanks for doing this. We already have a separate plugin repo for the swift provider. Once you do the work in terraform to enable this new feature in gophercloud, we can think about moving to it (you said one month?).

Issue #7967, opened 2016-08-04
Terraform doesn't produce error if parsing of s3 bucket policy json failed


murat1985:
Spotted minor bug in behaviour. Terraform produces green success output even s3 bucket policy json parsing is failed:
```
make plan
...
+ aws_s3_bucket.bucket-name
    acceleration_status:      "<computed>"
    acl:                      "private"
    arn:                      "<computed>"
    bucket:                   "bucket-name"
    force_destroy:            "false"
    hosted_zone_id:           "<computed>"
    policy:                   "Error parsing JSON: invalid character '\"' after object key:value pair"
    region:                   "<computed>"
    website.#:                "1"
    website.0.index_document: "index.html"
    website_domain:           "<computed>"
    website_endpoint:         "<computed>"
```

### Terraform Version
```
 >> terraform -v
Terraform v0.7.0
```
### Affected Resource(s)
```
aws_s3_bucket
```

### Terraform Configuration Files

To reproduce make json output invalid, e.g. remove coma before "Condition" key.

```
  policy = <<EOF
{
  "Version": "2008-10-17",
  "Id": "Policy1234567",
  "Statement": [
    {
      "Sid": "AllowGetObject",
      "Effect": "Allow",
      "Principal": "*",
      "Action": "s3:GetObject",
      "Resource": "arn:aws:s3:::bucket_name"
      "Condition": {
        "IpAdress": {
          "aws:SourceIp": [ "${var.myvar}" ]
        }
      }
    }
  ]
}
EOF
```

### Expected Behavior
Fail terraform plan on json error
```
"Error parsing JSON: invalid character '\"' after object key:value pair"
```

### Actual Behavior
It doesn't fail and produces green output

kwilczynski:
Sorry about this!

This could definitely be improved, as per:

[resource_aws_s3_bucket.go#L1313-L1324](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_s3_bucket.go#L1313-L1324)

and

[resource_aws_s3_bucket.go#L438-L455](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_s3_bucket.go#L438-L455)

Which probably could benefit form returning both error and string allowing for marshalling errors to bubble outside of `normalizeJson` function.

Having said that, this might be expected behaviour and/or by design for plan part.

jen20:
Hi @murat1985! I definitely agree this is a bug, we'll get it taken care of.

[...]

PR #8028, opened 2016-08-07
[WIP] provider/aws: Add normalizeJsonString and validateJsonString functions.

kwilczynski:
This commit adds ValidateFunc to the policy attribute so that JSON parsing
errors can be caught early. Generally, when there is a ValidateFunc set for the
attribute, one can safely assume that before any of the creation and/or update
of the existing resource would happen it would have to succeed validation. Also
adds support for new helper function which is used to normalise JSON string.

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket.go
@@ -1399,6 +1399,7 @@ func removeNil(data map[string]interface{}) map[string]interface{} {
 	return withoutNil
 }
 
+// DEPRECATED. Please consider using `normalizeJsonString` function instead.
 func normalizeJson(jsonString interface{}) string {
 	if jsonString == nil || jsonString == "" {
 		return ""
--- a/builtin/providers/aws/structure.go
+++ b/builtin/providers/aws/structure.go
@@ -1567,3 +1567,24 @@ func flattenPolicyAttributes(list []*elb.PolicyAttributeDescription) []interface
 
 	return attributes
 }
+
+// Takes a value containing JSON string and passes it through
+// the JSON parser to normalize it, returns either a parsing
+// error or normalized JSON string.
+func normalizeJsonString(jsonString interface{}) (string, error) {
+	var j interface{}
+
+	if jsonString == nil || jsonString.(string) == "" {
+		return "", nil
+	}
+
+	s := jsonString.(string)
+
+	err := json.Unmarshal([]byte(s), &j)
+	if err != nil {
+		return s, err
+	}
+
+	bytes, _ := json.Marshal(j)
+	return string(bytes[:]), nil
+}
--- a/builtin/providers/aws/structure_test.go
+++ b/builtin/providers/aws/structure_test.go
@@ -1143,7 +1143,7 @@ func TestNormalizeJsonString(t *testing.T) {
 	}
 
 	if actual != expected {
-		t.Fatalf("Got:\n\n%#s\n\nExpected:\n\n%#s\n", actual, expected)
+		t.Fatalf("Got:\n\n%s\n\nExpected:\n\n%s\n", actual, expected)
 	}
 
 	// Well formatted but not valid,
@@ -1165,6 +1165,6 @@ func TestNormalizeJsonString(t *testing.T) {
 
 	// We expect the invalid JSON to be shown back to us again.
 	if actual != invalidJson {
-		t.Fatalf("Got:\n\n%#s\n\nExpected:\n\n%#s\n", expected, invalidJson)
+		t.Fatalf("Got:\n\n%s\n\nExpected:\n\n%s\n", expected, invalidJson)
 	}
 }
--- a/builtin/providers/aws/structure_test.go
+++ b/builtin/providers/aws/structure_test.go
@@ -1122,14 +1122,14 @@ func TestNormalizeJsonString(t *testing.T) {
 	var actual string
 
 	// Well formatted and valid.
-	validJson := `{  
-   "abc": {  
+	validJson := `{
+   "abc": {
       "def": 123,
-      "xyz": [  
-         {  
+      "xyz": [
+         {
             "a": "ホリネズミ"
          },
-         {  
+         {
             "b": "1\\n2"
          }
       ]
@@ -1148,11 +1148,11 @@ func TestNormalizeJsonString(t *testing.T) {
 
 	// Well formatted but not valid,
 	// missing closing squre bracket.
-	invalidJson := `{  
-   "abc": {  
+	invalidJson := `{
+   "abc": {
       "def": 123,
-      "xyz": [  
-         {  
+      "xyz": [
+         {
             "a": "1"
          }
       }
--- a/builtin/providers/aws/structure_test.go
+++ b/builtin/providers/aws/structure_test.go
@@ -1116,3 +1116,55 @@ func TestFlattenPolicyAttributes(t *testing.T) {
 		}
 	}
 }
+
+func TestNormalizeJsonString(t *testing.T) {
+	var err error
+	var actual string
+
+	// Well formatted and valid.
+	validJson := `{  
+   "abc": {  
+      "def": 123,
+      "xyz": [  
+         {  
+            "a": "ホリネズミ"
+         },
+         {  
+            "b": "1\\n2"
+         }
+      ]
+   }
+}`
+	expected := `{"abc":{"def":123,"xyz":[{"a":"ホリネズミ"},{"b":"1\\n2"}]}}`
+
+	actual, err = normalizeJsonString(validJson)
+	if err != nil {
+		t.Fatalf("Expected not to throw an error while parsing JSON, but got: %s", err)
+	}
+
+	if actual != expected {
+		t.Fatalf("Got:\n\n%#s\n\nExpected:\n\n%#s\n", actual, expected)
+	}
+
+	// Well formatted but not valid,
+	// missing closing squre bracket.
+	invalidJson := `{  
+   "abc": {  
+      "def": 123,
+      "xyz": [  
+         {  
+            "a": "1"
+         }
+      }
+   }
+}`
+	actual, err = normalizeJsonString(invalidJson)
+	if err == nil {
+		t.Fatalf("Expected to throw an error while parsing JSON, but got: %s", err)
+	}
+
+	// We expect the invalid JSON to be shown back to us again.
+	if actual != invalidJson {
+		t.Fatalf("Got:\n\n%#s\n\nExpected:\n\n%#s\n", expected, invalidJson)
+	}
+}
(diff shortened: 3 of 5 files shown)


kwilczynski:
Related to #7967.

kwilczynski:
@stack72 the idea here is to add validation of JSON to catch any potential issues at the planning stag by explicitly notifying user about an error.

At the moment, the `normalizeJson` function would only return a string, regardless whether it was an error or a genuine value (JSON). This causes an error message originating from the JSON parser to be shown as a value for the attribute, and since the `normalizeJson` is often used (almost always) as to provide `StateFunc`, the outcome is a "green" plan (appearing as having no errors), which is expected given what the `StateFunc` is designed to do - and an string with error message is still a string, thus no issues from the `StateFunc` mechanics point of view.

The changes here are to make the `normalizeJson` so that it would return an error. This allows for an appropriate `ValidateFunc` to be introduced, so that errors related to JSON parsing can be caught early in the process and reported accordingly.

Example usage:

(before)

```
+ aws_s3_bucket.test123
    arn:              "" => "<computed>"
    bucket:           "" => "kwilczynski-test-bucket-123"
    force_destroy:    "" => "false"
    policy:           "" => "Error parsing JSON: invalid character '[' looking for beginning of object key string"


```

(after)

```
fix these before continuing.

Errors:

  * aws_s3_bucket.test123: "policy" contains an invalid JSON: invalid character '[' looking for beginning of object key string
```

Let me know what do you think.

builtin/providers/aws/resource_aws_s3_bucket.go
```diff
@@ -1310,17 +1315,30 @@ func removeNil(data map[string]interface{}) map[string]interface{} {
...
+func normalizeJson(jsonString interface{}) (string, error) {
 	if jsonString == nil || jsonString == "" {
-		return ""
+		return "", nil
 	}
 	var j interface{}
 	err := json.Unmarshal([]byte(jsonString.(string)), &j)
 	if err != nil {
-		return fmt.Sprintf("Error parsing JSON: %s", err)
+		return "", err
```
apparentlymart: I think it's best if we return exactly what the user provided (`return jsonString.(string), err`) here so that `normalizeJsonStateFunc` will just return the value unchanged when it's invalid.

The reason for this is that `StateFunc` is called in some situations where `ValidateFunc` *isn't* called, such as normalizing a value set from the upstream API during a resource's `Read`, and so preserving the invalid value will allow it to show in a diff, rather than showing as the empty string.

This is definitely an edge case and mainly a cosmetic thing; I took a similar approach in #8350, which does something similar for normalizing/validating IAM policy JSON.

apparentlymart:
Hi @kwilczynski. Thanks for doing this! That behavior of returning the error string as the normalized value has been bugging me for a while now.

This is a bit of an intimidating patch to accept because it touches many resources all at once and so would require running a *lot* of acceptance tests before merge. Could you possibly split this up in to one commit that adds the necessary shared code, and then a separate commit to update the S3 bucket resource, something like what I did in #8350? Then we can review and merge that smaller increment and attack the rest of these as separate PRs that can each be tested and merged in isolation.

I'm happy to help do the split and re-submit as a new PR if you don't have time. I want to get this in. :grinning:

builtin/providers/aws/resource_aws_s3_bucket.go
kwilczynski: Makes sense!

kwilczynski:
@apparentlymart thank you! I did bother me too (the lack of correct error handling), which is why this attempt to fix this.

No issues with splitting the commits, I will get to it.

I had a look at #8350, which is amazing, and it gave me an idea - what if we would to design an interface that would allow for abstracting JSON reading (parsing), writing and validation of us? 

The reason why I thought about an interface that could provide a concrete implementation for a specific use case e.g. IAM role, S3 bucket, etc., is because what I did was simply improve upon the `normalizeJson` function which is being used in few places whereas your change in #8350 goes beyond that and actively concerns itself with how the IAM policy document is represented, which is far superior to simply changing the behaviour of a function (a low-hanging fruit I saw and decided to grab it).

After merging changes from #8350, it would make very little sense to add my changes, which is why I am thinking that perhaps there would be a better way and abstract universal elements out so that we could re-use these.

What do you think?

apparentlymart:
@kwilczynski it is true that #8350 overlaps with the parts of this change that concern IAM policies. What you did here does seem like it'd be useful for other JSON-ish things that *aren't* IAM policies, though.

Do you have some thoughts about what a common abstraction (like you propose here) would look like?

We already have one abstraction here in the idea of `StateFunc` and `ValidateFunc` being generic across all schemas, and thus we can re-use certain functions like `normalizeJson` between them. It sounds like you're proposing another layer above that that handles some aspect of the JSON parsing/reserialization. I'm not opposed to such a thing, but I'm not sure I can imagine yet what such a thing would look like.

It would still be nice to get your improvements to `normalizeJson` and the validation function merged, even if we stop using it for IAM policies after #8350 is merged... what do you think about continuing down the path you were already on for the moment, and then once we have both of these merged we can use a separate issue/PR to discuss any generalizations between the two?

kwilczynski:
@apparentlymart :thumbsup:

I fully agree. I will continue. We can revise and refractor things later, if you want to, exactly like you say. :)

kwilczynski:
Things to do:

- [ ] Modify `normalizeJson` so that it returns JSON parsing error appropriately
- [ ] Provide `normalizeJsonStateFunc` to wrap `normalizeJson` accordingly
- [ ] Provide `validateJsonFunc` to wrap `normalizeJson` and provide feedback early
- [ ] Manually review change making sure that the behaviour is as expected for each resource
- [ ] Update acceptance tests for reach resource (as needed)
- [ ] Run acceptance tests to verify that core functionality is intact for each resource
- [ ] Update documentation so that it describes new behaviour (as needed)

kwilczynski:
Acceptance tests to run:

* Resources affected (AWS):

      - [ ] aws_cloudformation_stack
      - [ ] aws_cloudwatch_event_rule
      - [ ] aws_elasticsearch_domain
      - [ ] aws_glacier_vault
      - [ ] aws_kms_key
      - [x] aws_s3_bucket
      - [ ] aws_sns_topic
      - [ ] aws_sqs_queue
      - [ ] aws_vpc_endpoint
      - [ ] opsworks_layers

* Resources affected (AzureRM):

      - [ ] arm_template_deployment

Results tacked here: https://gist.github.com/kwilczynski/f89e7d1f0cfc4f744a98bdac15567954

kwilczynski:
The `opsworks_layers` resources seem to error a lot, but for may reasons. I need to look into these.

kwilczynski:
Related to #8615, since it removes the `normalizeJson` function.

kwilczynski:
@apparentlymart @stack72 if you don't mind, I could use your help. I cannot run the AzureRM test.

kwilczynski:
I have rebased against master to accommodate recent changes.

kwilczynski:
After having a conversation with @radeksimko , who was very kind and provided me with an idea about making this more manageable, I will split this commit into a series of Pull Requests. This one will be revised and subsequent Pull Requests will be pushed.

kwilczynski:
Unit tests are passing:

```
$ make test TEST=./builtin/providers/aws TESTARGS='-v -run=TestNormalizeJsonString'
2016/09/14 09:07:36 Generated command/internal_plugin_list.go
TF_ACC= go test ./builtin/providers/aws -v -run=TestNormalizeJsonString -timeout=30s -parallel=4
=== RUN   TestNormalizeJsonString
--- PASS: TestNormalizeJsonString (0.00s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	0.016s
```

```
$ make test TEST=./builtin/providers/aws TESTARGS='-v -run=TestValidateJsonString'
2016/09/14 09:09:05 Generated command/internal_plugin_list.go
TF_ACC= go test ./builtin/providers/aws -v -run=TestValidateJsonString -timeout=30s -parallel=4
=== RUN   TestValidateJsonString
--- PASS: TestValidateJsonString (0.00s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	0.016s
```

kwilczynski:
I have a failure in the `aws_s3_bucket`, but it also fails on the `master` branch. Seem like a problem introduced somewhere else.

kwilczynski:
Finally, I was able to recover my Azure account and test `arm_template_deployment`. which works fine.

builtin/providers/aws/validators_test.go
```diff
@@ -553,3 +553,58 @@ func TestValidateDbEventSubscriptionName(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateJsonString(t *testing.T) {
+	type testCases struct {
+		Value    string
+		ErrCount int
```
radeksimko: Since the cases in each slice below are all either valid or invalid, the `ErrCount` seems unnecessary, especially because if someone actually decided to add `ErrCount: 0` to `invalidCases` the error message wouldn't make much sense. But after all it's a test which is testing what it should, so 👍

builtin/providers/aws/structure_test.go
```diff
@@ -1116,3 +1116,55 @@ func TestFlattenPolicyAttributes(t *testing.T) {
...
+	var err error
+	var actual string
+
+	// Well formatted and valid.
+	validJson := `{
+   "abc": {
+      "def": 123,
+      "xyz": [
+         {
+            "a": "ホリネズミ"
```
radeksimko: 😃

builtin/providers/aws/structure.go
```diff
@@ -1567,3 +1567,24 @@ func flattenPolicyAttributes(list []*elb.PolicyAttributeDescription) []interface
...
+	}
+
+	s := jsonString.(string)
+
+	err := json.Unmarshal([]byte(s), &j)
+	if err != nil {
+		return s, err
+	}
+
+	bytes, _ := json.Marshal(j)
```
radeksimko: I'm assuming that nothing can go wrong in marshalling an object which we were previously able to successfully unmarshal... so I guess 👍  

I am aware this is more or less copy-paste of the existing function and it's rather a rhetorical question.

builtin/providers/aws/validators_test.go
kwilczynski: Very true, albeit I kept it like that for sake of consistency with other table-based tests I've seen in the code.

builtin/providers/aws/structure_test.go
kwilczynski: 😄

builtin/providers/aws/structure.go
kwilczynski: Indeed, this is a copy-and-paste. I can add error checking, although the only thing that could upset things here would be some odd thing making memory backing this corrupted.

Having said that, things like e.g. (errcheck)[https://github.com/kisielk/errcheck] could mark this as a potential issue.

kwilczynski:
@radeksimko thank you for the merge! I will now issue pull requests on per-resource basis, so that we can benefit from having this added.

PR #8033, opened 2016-08-07
Lambda ENI cleanup added to security group delete

gposton:
This adds support for cleaning up ENIs that get automatically created by Lambda functions when dealing with security groups used in Lambda functions

### DIFF
--- a/builtin/providers/aws/resource_aws_security_group.go
+++ b/builtin/providers/aws/resource_aws_security_group.go
@@ -345,6 +345,8 @@ func resourceAwsSecurityGroupDelete(d *schema.ResourceData, meta interface{}) er
 
 	log.Printf("[DEBUG] Security Group destroy: %v", d.Id())
 
+	deleteLingeringLambdaENIs(conn, d)
+
 	return resource.Retry(5*time.Minute, func() *resource.RetryError {
 		_, err := conn.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
 			GroupId: aws.String(d.Id()),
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_security_group.go
```diff
@@ -876,3 +878,68 @@ func protocolForValue(v string) string {
...
+
+	lambdaENIs := make([]ec2.NetworkInterface, 0)
+	if err != nil {
+		log.Printf("[WARN] Unable to fetch ENIs: %s", err)
+		return lambdaENIs
+	}
+
+	v := networkInterfaceResp.NetworkInterfaces
+	for _, eni := range v {
+		if strings.Contains(*eni.Description, "AWS Lambda VPC ENI") {
```
radeksimko: Is there any reason why not put this into the initial filtering on the API level? I think that would possibly make the code slightly cleaner + save some bytes between AWS/user and CPU cycles.

When I'm looking at the real ENI created by Lambda I think we can be even more strict in filtering to be _really_ sure we won't delete any ENI by mistake. It would be a sad coincidence, but after all description is something users can specify (i.e. name it `AWS Lambda VPC ENI`). We want to avoid deleting ENIs that are not managed by the underlying Lambda API if possible.

How about something like this? 😺 

```sh
aws ec2 describe-network-interfaces \
  --filter \
    'Name=group-id,Values=sg-aaaaaaaa' \
    'Name=description,Values=AWS Lambda VPC ENI: *' \
    'Name=requester-id,Values=*:awslambda_*'
```

radeksimko: ^ Theoretically we could still delete ENI which was originally created via IAM role that was called `awslambda_...`, but that would happen less likely than if we did a simple SG & `description` match.

builtin/providers/aws/resource_aws_security_group.go
```diff
@@ -876,3 +878,68 @@ func protocolForValue(v string) string {
...
+			lambdaENIs = append(lambdaENIs, *eni)
+		}
+	}
+	return lambdaENIs
+}
+
+func deleteLingeringLambdaENIs(conn *ec2.EC2, d *schema.ResourceData) error {
+	lambdaENIs := getLambdaENIs(conn, d)
+
+	for i := 0; i < len(lambdaENIs); i++ {
```
radeksimko: I think this and the next for loop can be merged into a single one and shouldn't even become too messy if you leverage the existing retry helpers `resource.StateChangeConf` like [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_network_interface.go#L204-L214).

That will also allow us to error out if we _actually_ failed to detach the network interface for any weird reason (i.e. if the ENI fails to detach after given `timeout`). Such edge case is currently ignored in this implementation as you just _hope_ it will be detached after 1 second.

Potentially you may be able to squeeze the third loop w/ `DeleteNetworkInterface` here too as that would also reduce complexity (from `3n` to `n`)... but I don't see that as a big deal compared to the first two loops.

Either way in all of these loops there is no need to use the C-like approach with index-incrementation, instead you can easily [range](https://github.com/golang/go/wiki/Range) over the slice:

```go
for _, eni := range lambdaENIs {
  params := &ec2.DetachNetworkInterfaceInput{
    AttachmentId: eni.Attachment.AttachmentId,
  }
```

radeksimko:
Hi @gposton 
thank you for taking the time to prepare and submit this patch. I left you some comments there.
Let me know what you think.

gposton:
Great feedback @radeksimko thanks... take another look?

builtin/providers/aws/resource_aws_security_group.go
```diff
@@ -876,3 +878,82 @@ func protocolForValue(v string) string {
...
+	networkInterfaceResp, err := conn.DescribeNetworkInterfaces(params)
+
+	if err != nil {
+		return err
+	}
+
+	v := networkInterfaceResp.NetworkInterfaces
+
+	for _, eni := range v {
+		detachNetworkInterfaceParams := &ec2.DetachNetworkInterfaceInput{
```
radeksimko: Probably just rare edge case, but we may find ENIs that are not attached, so I think we should have a condition here

e.g.

```
if eni.Attachment != nil {
  // all the detach logic
}

deleteNetworkInterfaceParams := ...
```

Otherwise ENI that is already detached would cause terraform to crash here (reproducible easily by manually detaching the ENI via AWS Console before running `terraform destroy`).

radeksimko:
Thanks for the modifications @gposton 
this is looking good except that nitpick with detached ENIs.

It would be awesome to have an acceptance test for this - e.g. cfg to create VPC-enabled Lambda function, then `lambda:Invoke`, then destroy that cfg

but missing acceptance test is not a blocker for this particular PR.

gposton:
@radeksimko, apologies for the slow response.  Good catch... I made the recommended update to handle the scenario where the ENI is detached.  I tested by manually detaching the ENI before destroying the stack and verifying that there were no errors and that the ENI was successfully deleted.

radeksimko:
Great, thanks for the modification. This LGTM now.

Issue #6128, opened 2016-04-11
aws_api_gateway_integration_response should depend on aws_api_gateway_integration


mattx86:
Hello,

I'm running Terraform v0.6.14 on Ubuntu 14.04 LTS (trusty).

I'm experimenting with the new Terraform functionality for managing AWS Lambda and AWS API Gateway.  I've run into an order/dependency issue, whereby Terraform tries to create the `aws_api_gateway_integration_response` before it creates the `aws_api_gateway_integration`, resulting in an error.

Please see the snippet below:

```
$ terraform apply
[...snip...]
aws_api_gateway_integration_response.mcs-test-api-integration-response: Creating...
  http_method:                         "" => "GET"
  resource_id:                         "" => "vaswdm"
  response_templates.#:                "" => "1"
  response_templates.application/json: "" => ""
  rest_api_id:                         "" => "477egnmtwh"
  status_code:                         "" => "200"
aws_lambda_function.mcs-test-lambda: Creation complete
aws_api_gateway_integration.mcs-test-api-integration: Creating...
  credentials:             "" => "arn:aws:iam::{REDACTED}:role/lambda-iam-mcs-test-lambda"
  http_method:             "" => "GET"
  integration_http_method: "" => "POST"
  resource_id:             "" => "vaswdm"
  rest_api_id:             "" => "477egnmtwh"
  type:                    "" => "AWS"
  uri:                     "" => "arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:{REDACTED}:function:mcs-test-lambda/invocations"
aws_api_gateway_integration.mcs-test-api-integration: Creation complete
Error applying plan:

1 error(s) occurred:

* aws_api_gateway_integration_response.mcs-test-api-integration-response: Error creating API Gateway Integration Response: NotFoundException: No integrationdefined for method
    status code: 404, request id: {REDACTED}

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

At this point, `terraform apply` can be run a second time, and this time succeed, because the `aws_api_gateway_integration` was created at the end of the above `apply`:

```
$ terraform apply
aws_api_gateway_rest_api.mcs-test-api: Refreshing state... (ID: 477egnmtwh)
[...]

PR #8049, opened 2016-08-08
Updated API GW integration response documentation

Ninir:
### Reasoning for docs update
Recently, I've been using Terraform to manage AWS API GWs with Lambda backends.
It appears that an explicit `depends_on`dependency is required when adding an `aws_api_gateway_integration_response ` resource. Not setting it would lead to this error:

> [...] Error creating API Gateway Integration Response: NotFoundException: No integration defined for method

Checked against 0.6.16

### Thread Issue
https://github.com/hashicorp/terraform/issues/6128


jen20:
Thanks @Ninir!

Ninir:
Thank **you** for the quick review/merge @jen20 :-)

Issue #8051, opened 2016-08-08
provider/aws: aws_rds_cluster create may take > 15 minutes


kcrawley:
When restoring large db snapshots within a RDS cluster, I've witnessed create actions take > 30 minutes. I propose we bump the timeout to 40 minutes, which corresponds with the maximum timeout on aws_rds_cluster_instance create actions.

I have submitted a PR to address this issue.

catsby:
Closed by #8052, thanks!

PR #7158, opened 2016-06-14
provider/aws: Added support for `snapshot_identifier` parameter in aws_rds_cluster

kcrawley:
This PR adds support for `snapshot_identifier` use in `resource_aws_rds_cluster`. eg:

```
resource "aws_rds_cluster" "rds-snapshot-cluster" {
  cluster_identifier = "rds-snapshot-cluster"
  snapshot_identifier = "rds-snapshot"
  db_subnet_group_name = "${aws_db_subnet_group.snapshot-dbsg.name}"
  vpc_security_group_ids = ["${aws_security_group.snapshot-vpc-sg.id}"]
}
```
Once the cluster has been restored from the snapshot, an instance can be added to the cluster and it will contain the data stored in the cluster snapshot. eg:

```
resource "aws_rds_cluster_instance" "rds-cluster-instance" {
  instance_class = "db.r3.large"
  cluster_identifier = "${aws_rds_cluster.rds-snapshot-cluster.id}"
  db_subnet_group_name = "${aws_db_subnet_group.snapshot-dbsg.name}"
}
```

FWIW, the acceptance tests for `db_instance` and `rds_cluster_` are fairly basic. I did not see a pattern for several conditional behaviors in `db_instance` that I could replicate for `rds_cluster_` (specifically, restoration from snapshots). I'd be interested in any feedback to that regard.

PR #8052, opened 2016-08-08
providers/aws: #8051 :: bumped aws_rds_cluster timeout to 40 minutes

kcrawley:
This PR addresses this issue reported in #8051 ; the timeout for aws_rds_cluster create actions has been increased to 40 minutes from 15 minutes.


radeksimko:
Hi @kcrawley 
I have no objections against raising the timeout, just to clarify - did you experience timeout with 0.7.0 (i.e. after https://github.com/hashicorp/terraform/pull/7604 ) or is this from an older version of Terraform?

kcrawley:
@radeksimko 
i experienced it with 0.7.0. prior to 0.7.0, i believe support for restoring clusters from a snapshot wasn't even there (i introduced the support in https://github.com/hashicorp/terraform/pull/7158)

PR #8053, opened 2016-08-08
don't retry the atlas requests with the wrong cert

jbardin:
If the certs don't match on between the atlas client and server it probably won't ever recover without intervention, so abort immediately.


phinze:
LGTM!

Issue #8060, opened 2016-08-08
docs for state CLI link to rm command, which doesn't exist


mblakele:
### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. visit https://www.terraform.io/docs/commands/state/index.html
2. observe a link to the subcommand `rm` https://www.terraform.io/docs/commands/state/rm.html
3. select `rm`

### Expected Behavior
The link should work.

### Actual Behavior
404 Not Found

PR #5953, opened 2016-03-31
command/state rm

mitchellh:
This introduces `terraform state rm` for removing resources out of a Terraform state. This doesn't physically destroy the resource, it just removes it from Terraform management.

Note that this can also remove entire modules, tainted instances, and more. Anything that matches a resource address can be removed.

See the documentation for more information.

This PR contains:

  * The `terraform state rm` command and tests
  * A bug fix for `terraform.StateFilter` when filtering single resources
  * Website updates for `terraform state rm`

PR #8066, opened 2016-08-09
Fix. Remove reference to the `rm` state sub-command.

kwilczynski:
At the moment the `terraform state rm ...` invocation is not supported (there
is no `state_rm.go`, etc.). This commit removes link to the page detailing the
sub-command, which does not exist at the moment, thus removing unexpected 404
error and confusion around whether this is supported or not.

### DIFF
--- a/website/source/layouts/commands-state.erb
+++ b/website/source/layouts/commands-state.erb
@@ -25,10 +25,6 @@
 							<a href="/docs/commands/state/mv.html">mv</a>
 						</li>
 
-						<li<%= sidebar_current("docs-state-sub-rm") %>>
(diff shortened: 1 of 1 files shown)


kwilczynski:
Resolves #8060.

jen20:
Hi @kwilczynski! Thanks for looking at fixing this - I'm actually going to fix this in a slightly different manner - we did implement (and merge) `state rm` but apparently it got lost during one of the many large rebases that went into 0.7. It was implemented in #5953, so I'll pull over and fix up that implementation. Thanks for the PR though!

kwilczynski:
@jen20 ah I had no idea! (although, I suspected something was planned).

PR #8070, opened 2016-08-09
provider/openstack: Add support for updating the External GW assigned to a Neutron router

fatmcgav:
The `openstack_networking_router_v2` [doc's ](https://www.terraform.io/docs/providers/openstack/r/networking_router_v2.html) suggest that it should be possible to modify the `external_gateway` value of a router. 

However there is currently no functionality in the provider to support this. 

This PR adds that functionality. 

Comments welcome.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_router_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_router_v2.go
@@ -213,6 +213,15 @@ func resourceNetworkingRouterV2Update(d *schema.ResourceData, meta interface{}) 
 		asu := d.Get("admin_state_up").(bool)
 		updateOpts.AdminStateUp = &asu
 	}
+	if d.HasChange("external_gateway") {
+		externalGateway := d.Get("external_gateway").(string)
+		if externalGateway != "" {
+			gatewayInfo := routers.GatewayInfo{
+				NetworkID: externalGateway,
+			}
+			updateOpts.GatewayInfo = &gatewayInfo
+		}
+	}
 
 	log.Printf("[DEBUG] Updating Router %s with options: %+v", d.Id(), updateOpts)
 
--- a/builtin/providers/openstack/resource_openstack_networking_router_v2_test.go
+++ b/builtin/providers/openstack/resource_openstack_networking_router_v2_test.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"fmt"
+	"os"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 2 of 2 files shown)


jtopjian:
@fatmcgav Thanks for this!

Just to confirm: you've tested this and all works well? :)

Would you be able to add an acceptance test? No big deal if not - I can add one later.

fatmcgav:
@jtopjian Yeh, I've built and tested this locally and it all works as expected... 

Example log line:
>>2016/08/09 15:06:30 [DEBUG] plugin: terraform: openstack-provider (internal) 2016/08/09 15:06:30 [DEBUG] Updating Router 00fcb5ac-7c42-493c-92aa-b74a354ca22e with options: {Name: AdminStateUp:<nil> Distributed:<nil> GatewayInfo:0xc820427e90 Routes:[]}
2

Whereas previously `GatewayInfo` was always `<nil>`. 

WRT test's, I can give it a go... Is it just a case of extending the existing test here? https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_router_v2_test.go#L16-L35

jtopjian:
Yeah, the best way would be to duplicate this entire test: https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_router_v2_test.go#L13-L35

With new copies of the two configs found here: https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_router_v2_test.go#L90-L102

that include an updated gateway.

But if it becomes too cumbersome, no worries :)

Also looks like travis had a hiccup. Force pushing the commit should trigger it to run again.

fatmcgav:
OK, I've added what I think should be a valid test case... 

However I seem to be unable to get the test to fail locally, even if I comment out the additions I've made... But then the tests don't fail if I comment out the `updateOpts.Name` line aswell... 

So probably me not doing it right... 

For ref, I'm trying to test with:
`make test TEST=./builtin/providers/openstack`

And the result with my additions commented out is:
>openstack_router_update_external_gw ± make test TEST=./builtin/providers/openstack
/Users/gavinw/golang/bin/stringer
2016/08/09 17:26:04 Generated command/internal_plugin_list.go
TF_ACC= go test ./builtin/providers/openstack  -timeout=30s -parallel=4
ok      github.com/hashicorp/terraform/builtin/providers/openstack      0.025s

jtopjian:
Looks like you're just running the unit tests. To run the acceptance tests, you can do:

```
make testacc TEST=./builtin/providers/openstack
```

To run an individual test, do:

```
make testacc TEST=./builtin/providers/openstack TESTARGS="-run=TestAccNetworkingV2Router_update_external_gw"
```

But that requires certain environment variables to be set, which are described at the bottom of this page: https://www.terraform.io/docs/providers/openstack/index.html

It looks like you're hardcoding the gateway ID. You can store this in the environment variable `OS_EXTGW_ID` (just noticed it's missing from the above doc link - will get that fixed). But then you need to write out the test fixture within the test itself so the variable interpolation works! 😵  You can see an example of that here: https://github.com/hashicorp/terraform/blob/709e0bf9ad29f200b93dbe7cc3abe8571b4623f7/builtin/providers/openstack/resource_openstack_networking_floatingip_v2_test.go#L68

Looks like you hit a perfect storm of what should have been a simple test. Sorry about that. I'd be happy to fix this up this evening and submit a PR to your branch of this.

fatmcgav:
Hah, sounds like fun... ;) 

I'll see if I can spin up a local devstack and give it a go...

fatmcgav:
@jtopjian Cheers for the acceptance test PR. 

Merged in, and fixed'up my commits aswell.

jtopjian:
@fatmcgav Nice! Would you be able to squash everything into one commit?

fatmcgav:
@jtopjian Squashed all of mine into one commit, and then left yours as is... 

Unless you want your commit squashed aswell?

jtopjian:
@fatmcgav This is perfect. Thank you so much!

fatmcgav:
No worries, glad I could add some value :)

PR #8080, opened 2016-08-09
docs: Fix map key interpolation documentation

jen20:
Previously was recommending the now-invalid dot syntax for map keys, change to using HIL indexing.


stack72:
LGTM!

PR #8081, opened 2016-08-09
Allow the HCL input when prompted

jbardin:
We already accept HCL encoded input for -vars, and this expands that to
accept HCL when prompted for a value on the command line as well.

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -317,16 +317,18 @@ func (c *Context) Input(mode InputMode) error {
 				}
 			}
 
+			var valueType config.VariableType
+
 			v := m[n]
-			switch v.Type() {
+			switch valueType = v.Type(); valueType {
 			case config.VariableTypeUnknown:
 				continue
 			case config.VariableTypeMap:
-				continue
+				// OK
 			case config.VariableTypeList:
-				continue
+				// OK
 			case config.VariableTypeString:
-				// Good!
+				// OK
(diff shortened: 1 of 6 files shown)


terraform/context.go
```diff
@@ -340,6 +342,12 @@ func (c *Context) Input(mode InputMode) error {
 				}
 			}
 
+			// this should only happen during tests
+			if c.uiInput == nil {
+				log.Println("[WARN] Content.uiInput is nil")
+				continue
+			}
```
phinze: lol I am sorry for the loss of sanity required to figure this bit out

terraform/context.go
```diff
@@ -355,27 +363,33 @@ func (c *Context) Input(mode InputMode) error {
...
-					break
+					if v.Required() {
+						// Redo if it is required, but abort if we keep getting
+						// blank entries
+						if retry > 2 {
+							return fmt.Errorf("missing required value for %q", n)
+						}
+						retry++
+						continue
+					}
```
phinze: Any reason this got fanned out from `&&` before to nested `if` here?

phinze:
LGTM w/ one tiny Q

terraform/context.go
jbardin: I think there was a statement after the nested if during some iteration. I'll collapse this back down again before merging.

PR #8086, opened 2016-08-09
Adding support for sync_time_with_host flag

dkalleg:


builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1710,6 +1722,16 @@ func (vm *virtualMachine) setupVirtualMachine(c *govmomi.Client) error {
...
 		configSpec.GuestId = "otherLinux64Guest"
 	}
+	if vm.syncTimeWithHost {
+		if configSpec.Tools == nil {
+			toolsConfig := types.ToolsConfigInfo{
+				SyncTimeWithHost: &vm.syncTimeWithHost,
+			}
+			configSpec.Tools = &toolsConfig
+		} else {
+			*configSpec.Tools.SyncTimeWithHost = vm.syncTimeWithHost
```
dagnello: SyncTimeWithHost might also be null here, should check for that too

dkalleg: Good call, will push a patch shortly.

chrislovecnm:
Do we need docs and a test??

dkalleg:
@chrislovecnm This is a fairly granular feature and may cause more overhead than its worth to add an additional test for this.  I'm, of course, open to go with the popular decision on this.

builtin/providers/vsphere/resource_vsphere_virtual_machine.go
```diff
@@ -1103,6 +1114,7 @@ func resourceVSphereVirtualMachineRead(d *schema.ResourceData, meta interface{})
 	d.Set("cpu", mvm.Summary.Config.NumCpu)
 	d.Set("datastore", rootDatastore)
 	d.Set("uuid", mvm.Summary.Config.Uuid)
+	d.Set("sync_time_with_host", *mvm.Config.Tools.SyncTimeWithHost)
```
stack72: This doesn't need deferenced - d.Set() does this in a safe way :)

paddycarver:
Hi! This PR has been open for a while and waiting response from the community without any activity, so I'm going to close it out. We do this to try to keep our attention on issues that are actively impacting the community, it's not a comment on the PR itself. If there is still interest in this PR, please feel free to reply, and we'll re-open the PR. Thanks!

PR #8089, opened 2016-08-10
docs/powerdns: Fix invalid markdown syntax

f440:
[Power DNS record resource](https://www.terraform.io/docs/providers/powerdns/r/record.html) page has broken code block. This PR will resolve it.


stack72:
Hi @f440 

This LGTM! thanks :)

P.

Issue #1313, opened 2015-03-26
Security Groups do not allow ICMP


wazoo:
I was trying to create a security group with the following configuration:

```
ingress {
  protocol = "icmp"
  security_groups = ["${aws_security_group.sgr_test_nodes.id}"]
}
```
and I get the following error:
```
Errors:

  * 'aws_security_group.sgr_test_nodes' error: ingress.2.from_port: required field is not set
  * 'aws_security_group.sgr_test_nodes' error: ingress.2.to_port: required field is not set
```
While I understand why this error is happening, ICMP doesn't really have a port and if you try to specify one you get this error from amazon:

```
Error applying plan:

1 error(s) occurred:

* Error revoking security group ingress rules: ICMP code (65535) out of range (InvalidParameterValue)
```
Is there a specific way to configure ICMP?  From looking at [resource_aws_security_group.go](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_security_group.go) it didn't look that way so this may be a feature request, maybe a rule type or something so that way we could accomplish the ability to do "all tcp" or whatever without having to specify ports.

rhiann0n:
The Terraform Resource for security groups with the AWS provider does allow ICMP, but I'm researching if complex custom rules can be added without writing a plugin or modifying source code.  Haven't found a solution to that, but hopefully HashiCorp will consider this a feature request?

Please allow Custom ICMP rules? (Its even tricky with the AWS CLI.)

Here is an example I found randomly online that should help you, @wazoo:
https://github.com/cransom/tf_aws_sg/blob/master/sg_vpn/main.tf

Hope that helps.

Rhiannon Portwood
Senior Operations Engineer
www.messagesystems.com
twitter @MessageSystems

ltcdnunez:
I think I just sorted this out. As it happens, Amazon uses the same interface for CloudFormation and there is some documentation on how to use ICMP. The `from_port` is the ICMP type number and the `to_port` is the ICMP code (which does not exist for may types and should be either 0 or -1 in that case). See the [Amazon CloudFormation Documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html) for details. And here are the [ICMP types/codes](http://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml).

Here's an example of how to allow inbound echo requests:

```
ingress {
  protocol = 200
[...]

PR #8093, opened 2016-08-10
Documentation: explain the role of to_port in a security group rule when protocol is "icmp"

ldanz:
Security group rules overload "from_port" and "to_port" for ICMP.  The docs currently only specify what from_port does (the ICMP type) but not to_port. [This thread](https://github.com/hashicorp/terraform/issues/1313) explains that to_port is the ICMP code, so I've added that to the docs because it would have helped me set up my security group.

This is relevant to the current terraform version.


catsby:
Thanks you!

catsby:
Deployed! https://www.terraform.io/docs/providers/aws/r/security_group.html

PR #8102, opened 2016-08-10
Support setting datacenter when using consul remote state

### LINKED ISSUE #1614: Consul remote state backend to support setting a datacenter
It would be convenient for the Consul remote state backend to support setting the datacenter name on the client it creates. This would be consistent with the other places where Terraform interacts with Consul.

guidob:
Small change to allow setting datacenter when using consul remote state:

`terraform remote config -backend=consul -backend-config=path=some/path -backend-config=access_token=some-token -backend-config=datacenter=dc1`

This fixes #1614.

### DIFF
--- a/state/remote/consul.go
+++ b/state/remote/consul.go
@@ -24,6 +24,9 @@ func consulFactory(conf map[string]string) (Client, error) {
 	if scheme, ok := conf["scheme"]; ok && scheme != "" {
 		config.Scheme = scheme
 	}
+	if datacenter, ok := conf["datacenter"]; ok && datacenter != "" {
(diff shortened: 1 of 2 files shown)


stack72:
Hi @guidob 

The tests pass as expected here. Please could you add some documentation for this as part of the PR?

Thanks

Paul

guidob:
@stack72 ,

Good point, first time for me. Added a small piece.

stack72:
@guidob thanks for getting back to me so fast on this :) This LGTM so will merge when the build goes green!

Thanks for the work on this

PR #8103, opened 2016-08-10
helper: create validation package to provide common validation functions

pmcatominey:


pmcatominey:
Hi,

This is a suggestion on removing the need for writing multiple variations of the same logic when validating a schema.

I've used the traffic_manager_profile as an example of what the functions can replace.

Cheers,
Peter

jen20:
Hi @pmcatominey! This looks like a good idea to me. We're gearing up to release so I don't want to merge this immediately beforehand, I'd like to get at least one run of acceptance testing on it. We'll circle back around to this soon though.

stack72:
ok, let's get this merged now :) We have time between now and 0.7.6 to see if there are any issues we need to work out

@pmcatominey can you keep an eye on the nightly tests for AzureRM? https://travis-ci.org/hashicorp/terraform/builds

PR #7794, opened 2016-07-25
provider/aws: Re-implement api gateway parameter handling

nicolai86:
This PR is a follow up of my original API Gateway PR, #4295

Now that GH-2143 is finally closed this PR does away with the ugly `request_parameters_in_json` and `response_parameters_in_json` hack.

E.g. before:

```
resource "aws_api_gateway_method" "test" {
  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
  resource_id = "${aws_api_gateway_resource.test.id}"
  http_method = "GET"
  authorization = "NONE"

  request_models = {
    "application/json" = "Error"
  }

  request_parameters = <<PARAM 
  {
	  "method.request.querystring.page": false
  }
PARAM
}
```

after

```
resource "aws_api_gateway_method" "test" {
  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
  resource_id = "${aws_api_gateway_resource.test.id}"
  http_method = "GET"
  authorization = "NONE"

  request_models = {
    "application/json" = "Error"
  }

  request_parameters = {
	  "method.request.querystring.page" = false
  }
}
```

Issue #3980, opened 2015-11-19
aws_elasticsearch_domain Advanced options are not applied properly


robertfirek:
When tried to apply the following advanced option on **aws_elasticsearch_domain**

```
(...)
advanced_options {
    "rest.action.multi.allow_explicit_index" = true
}
(...)
```

it was not applied properly in amazon (**rest.action.multi.allow_explicit_index** option was set to **false**).

Changing this option doesn't change plan (I've changed true to false) and it is impossible to control options.

State file:

```
(...)
"advanced_options.#": "1",
"advanced_options.rest.action.multi.allow_explicit_index": "false", 
(...)
```

Terraform version:
v0.6.5

radeksimko:
I've got a hunch that this may be caused by the way we process the map for advanced_options:
https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/structure.go#L639

I'll try adding some extra logging and reproduce it.
Thanks for the report!

radeksimko:
@robertfirek You can try specifying options as strings in the meantime:

```
advanced_options {
    "rest.action.multi.allow_explicit_index" = "true"
}
```

robertfirek:
@radeksimko I've tried this approach before I wrote this issue. It doesn't work.

gmorpheme:
I'm having the same issue - have tried with both booleans and strings with no luck.

radeksimko:
After trying to reproduce this & running a few tests, I believe this is a core issue, described here https://github.com/hashicorp/terraform/issues/2143

Tags were something that a user could choose/change, but we cannot choose the map keys here, so the bug is affecting this resource more significantly. `advanced_options` are completely ignored at the moment due to this bug.

genofedanzo:
Ran into this today.  Would love a fix.

genofedanzo:
Also wanted to mention that from I testing I noticed that if you try to set advanced_options they end up being set to false  For instance, rest.action.multi.allow_explicit_index normally defaults to true but if you happen to manually set this to true it ends up being set as false.

chiefy:
@genofedanzo also I believe the advanced options for ES domains are required, so it will essentially always be set to false :cry:

genofedanzo:
[...]

PR #8104, opened 2016-08-10
core: Map values are always treated as strings (Elem is ignored)

radeksimko:
Attached is a failing test to demonstrate the issue.

I'm not sure if non-string maps were ever supposed to be supported, if not I think that we should be checking if `Elem` is specified for `TypeMap` and if it is, we should be just returning error early.
### Related PRs/issues:
- #7794 
- #3980

### DIFF
--- a/helper/schema/resource_data_test.go
+++ b/helper/schema/resource_data_test.go
@@ -2959,6 +2959,111 @@ func TestResourceDataState_schema(t *testing.T) {
 	}
 }
 
+func TestResourceData_nonStringValuesInMap(t *testing.T) {
+	cases := []struct {
+		Schema       map[string]*Schema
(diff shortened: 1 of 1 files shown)


radeksimko:
This was already implemented in https://github.com/hashicorp/terraform/pull/10203 and relevant validation was added in https://github.com/hashicorp/terraform/pull/12638 and https://github.com/hashicorp/terraform/pull/12722 respectively.

We still do not have any clear way to specify mixed maps (values that aren't of the same type) and there may be some decision & thought process needed to get that done - but that's for a separate issue/PR.

Maps with primitive values can now be specified like this:

```go
"mapInt": &Schema{
	Type: TypeMap,
	Elem: TypeInt,
},
```

PR #8108, opened 2016-08-10
provider/aws: Change CloudFormation outputs element to be optional

sigurdknippenberg:
With the changes made to Terraform 0.7.0 any outputs defined in a
CloudFormation template can no longer be referenced by other Terraform
resources. The reason for this is that Terraform does not read the
CloudFormation template and as a result is not aware of the output
variables until the CloudFormation template is executed and the variable
checks happen before this. As a result 'terraform plan' and 'terraform
apply' will fail if any CloudFormation output variables are referenced
within Terraform.

In this change the `outputs` variable is changed to be optional so that
it can be specified in the Terraform `aws_cloudformation_stack`
resource. This allows output variables to be predefined so that they
can be referenced by dependent resources. Their values will be updated
once the CloudFormation template executes and the dependent resources
will get the correct value from the variables.

For example, if the CloudFormation template has the following output
defined:

```
"Outputs" : {
    "VPCReference" : {
      "Description" : "VPC reference ID",
      "Value" : { "Ref" : "SampleVPC" }
    }
  }
```

Then the Terraform template can predefine the outputs like this:

```
resource "aws_cloudformation_stack" "main" {
  name = "test-sample-cloudformation"
  template_body = "${file(var.cloudformation_file)}"
  outputs = { VPCReference = "" }
```

And the variable can be referenced elsewhere like this:

```
output "vpc_reference" { value = "${aws_cloudformation_stack.main.outputs.VPCReference}" }
```


radeksimko:
Hi @sigurdknippenberg 
thank you for bringing this to our attention.

What you're describing here sounds like a core bug that would prevent users from referencing `Computed`-only `TypeMap` fields.

I'm going to try and reproduce this in isolation from the CloudFormation resource, because if that's true, then I'm pretty sure it's a core bug that needs fixing in the core, not on resource level.

User should not be forced to pre-populate a computed map, in fact these two options should behave as mutual exclusive options at runtime - i.e. either the user specified them - in which case user's value is taken - or it's not specified - in which case it's computed. I may be wrong, but I doubt we have any "merge" logic that would figure out differences between `Optional`/`Computed`.

^ @jen20 correct me if I'm wrong about the expected behaviour of `TypeMap` & `Optional`/`Computed`.

sigurdknippenberg:
I just found pull request https://github.com/hashicorp/terraform/pull/8080 which specifies an alternate syntax for referring to map variables. I just tested it out and it works fine for me. As a result this pull request is no longer needed.

apparentlymart:
Thanks for following up, @sigurdknippenberg!

PR #7617, opened 2016-07-13
provider/aws: Added the ability to import aws_iam_role's

bigkraig:
Not much to it, needed this to build a tf configuration for our core configs.

PR #8112, opened 2016-08-10
Revert "provider/aws: Added the ability to import aws_iam_role's"

catsby:
Reverts hashicorp/terraform#7617

This PR, as determined by `git bisect` and another review, has caused numerous failures in our nightly test suite. Specifically, any user using `aws_iam_role` and the [heredoc](https://en.wikipedia.org/wiki/Here_document) style for their policy statement will get a perpetual diff.


phinze:
LGTM - if we are seeing test failures +1 for revert-then-investigate strategy

PR #8114, opened 2016-08-10
aws: Change names of new skip_* fields + document those

radeksimko:
Follow up to https://github.com/hashicorp/terraform/pull/7874

cc @renier any feedback appreciated


renier:
👍

PR #8120, opened 2016-08-10
Fix panic from a null resources module in a state file

jbardin:
https://gist.github.com/davidneudorfer/39bfdd1854384e57537fc0a5e3a4fc65

Make sure all maps and slices are initialized when reading a state file to prevent nil panics like above.

If we read a state file with "null" objects in a module and they become
initialized to an empty map the state file may be written out with empty
objects rather than "null", changing the checksum. If we can detect
this, increment the serial number to prevent a conflict in atlas.

Our fakeAtlas test server now needs to decode the state directly rather
than using the ReadState function, so as to be able to read the state
unaltered.

The terraform.State data structures have initialization spread out
throughout the package. More thoroughly initialize State during
ReadState, and add a call to init() during WriteState as another
normalization safeguard.

Expose State.init through an exported Init() method, so that a new State
can be completely realized outside of the terraform package.
Additionally, the internal init now completely walks all internal state
structures ensuring that all maps and slices are initialized.  While it
was mentioned before that the `init()` methods are problematic with too
many call sites, expanding this out better exposes the entry points that
will need to be refactored later for improved concurrency handling.

The State structures had a mix of `omitempty` fields. Remove omitempty
for all maps and slices as part of this normalization process. Make
Lineage mandatory, which is now explicitly set in some tests.

### DIFF
--- a/command/refresh_test.go
+++ b/command/refresh_test.go
@@ -173,7 +173,7 @@ func TestRefresh_defaultState(t *testing.T) {
 	}
 
 	p.RefreshFn = nil
-	p.RefreshReturn = &terraform.InstanceState{ID: "yes"}
+	p.RefreshReturn = newInstanceState("yes")
 
 	args := []string{
 		testFixturePath("refresh"),
@@ -200,7 +200,8 @@ func TestRefresh_defaultState(t *testing.T) {
 	actual := newState.RootModule().Resources["test_instance.foo"].Primary
 	expected := p.RefreshReturn
 	if !reflect.DeepEqual(actual, expected) {
-		t.Fatalf("bad: %#v", actual)
+		t.Logf("expected:\n%#v", expected)
+		t.Fatalf("bad:\n%#v", actual)
 	}
 
 	f, err = os.Open(statePath + DefaultBackupExtension)
@@ -347,7 +348,7 @@ func TestRefresh_outPath(t *testing.T) {
 	}
 
 	p.RefreshFn = nil
-	p.RefreshReturn = &terraform.InstanceState{ID: "yes"}
+	p.RefreshReturn = newInstanceState("yes")
 
 	args := []string{
 		"-state", statePath,
@@ -577,7 +578,7 @@ func TestRefresh_backup(t *testing.T) {
 	}
 
 	p.RefreshFn = nil
-	p.RefreshReturn = &terraform.InstanceState{ID: "yes"}
+	p.RefreshReturn = newInstanceState("yes")
 
 	args := []string{
 		"-state", statePath,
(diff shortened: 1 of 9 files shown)


jbardin:
I ended up squashing all the commits down here, since the there were too many change done and undone by the series of commits.  This commit is a little large, but much easier to understand in isolation.

jbardin:
cc @phinze and/or @jen20 for review

terraform/state.go
```diff
@@ -713,7 +752,7 @@ type ModuleState struct {
 	// Terraform. If Terraform doesn't find a matching ID in the
 	// overall state, then it assumes it isn't managed and doesn't
 	// worry about it.
-	Dependencies []string `json:"depends_on,omitempty"`
+	Dependencies []string `json:"depends_on"`
```
phinze: The normalization-serial-bump below will also cover these changes as well, yes?

phinze:
Looking good on first pass!

Trying to structure a "rule of thumb" going forward: never create a `terraform.State`, `ResourceState`, or `InstanceState` without `Init()`. Does that sound right?

terraform/state.go
jbardin: yes, since every ReadState checks WriteState for changes, we'll always bump the serial on a change.

jbardin:
I don't want to commit yet to a global pattern for handling these, as I'd rather not have this api cruft leak out into provider-land. 

Right now it should be the case that `init()` is called everywhere it matters, and you only need to call it manually for changes within the terraform package or tests that build their own state. I was going to do further review when I get back, as the `init()` calls all over place are part of the concurrency issues we have, and need to be cleaned up. It's going to take a little planning to not have the affect providers.

phinze:
Cool that all makes sense. This LGTM, but let's also get @jen20 to do a triple check here before merge.

phinze:
cc @jen20 or @mitchellh for double check review then I'll merge it 👍

mitchellh:
LGTM.

One thing we should probably look into in the future is exposing test helpers via `testing.go` or `testing_state.go` (our convention is just _prefixing_ the file with `testing`) for public APIs to get variously formed state files that can be used in and out of the core package that handles this in various ways.

I see there are a lot of ways that that wouldn't have helped but at least for "I need a ready to go state" those cases should ideally be centralized in a single function.

This PR LGTM though!

command/refresh_test.go
```diff
@@ -742,6 +743,20 @@ func TestRefresh_displaysOutputs(t *testing.T) {
 	}
 }
 
+// When creating an InstaneState for direct comparison to one contained in
+// terraform.State, all fields must be inintialized (duplicating the
```
Ninir: Small typo here: `initialized`, awesome work otherwise! :-)

phinze: Good catch! Will fix on master.

Issue #8007, opened 2016-08-05
provider/azurerm: currently lacking load balancer resource


pmcatominey:
Hi,

I'm creating this issue as a feature request and central discussion point for load balancer support in the azurerm provider.

There are a number of PR's open, proposing solutions:
 - #4343
 - #6429
 - #6467

The latter two are attempts to providing the resource in different ways to the first. It would be good to hear thoughts on what solution is preferred and get the ball rolling again.

Personally I am in favour of having separated resources to model in particular the probes and rules, at least until it is possible to pass multiple sets into a module of the same type.

Regards

jen20:
Hi @pmcatominey! I agree that in the short term (until nested resources are implemented) that we should model each load balancer component as an independent resource. One of the team will pick up one of open proposals at some point soon now that 0.7 is out the door.

edevil:
Should we follow this ticket or #8130 for the status of load balancer support in the AzureRM module?

zachgersh:
hey @jen20 - could we just pour a tiny bit more time or even just merge #6467? It would seem that this is the split resource approach that you were wanting (and I agree is a better approach). It would be super disappointing if we had to wait even longer to get even rudimentary LB support for azure as a whole handful of these PRs have been open for quite a few months without a definitive direction. cc @stack72

stack72:
Hi @pmcatominey 

This is going to be seen through to completion by @jen20 and I in #9199 

Paul

PR #8130, opened 2016-08-11
[WIP] Azure Resource Manager: Load Balancers

tombuildsstuff:
This is a complete work in progress - opening this so I can keep track of tasks & get some feedback. Load Balancers are available in the Azure SDK for Go - so I'm trying to follow the convention of the other components from this module.

- [ ] Create Load Balancer
- [ ] Get Load Balancer
- [ ] Delete Load Balancer
- [ ] Update the Docs
- [ ] Tests
  - [ ] Internal Load Balancer
  - [ ] Public Load Balancer
- [ ] Linting / Whitespace (I didn't realise my Atom settings were so far out, I'll fix that in a bit..)

Questions:
 - Given that the internal and external LB's have completely different properties associated with them, is it worth making them separate resources, so we can strongly type them?

Just noticed #8007 after opening this - let me know if one of the team would rather pick this up instead?


stack72:
Hi @tombuildsstuff 

this is going to be seen through to completion by @jen20 and I in #9199

PR #7770, opened 2016-07-22
Provider: Packet - add ip and volume

ayudemura:
We have updated our [packngo](https://github.com/packethost/packngo), and plugged in ip service and volume service.

PR #8142, opened 2016-08-11
Provider: Packet - Add volume

ayudemura:
#7770 is not reopen-able...


ayudemura:
@stack72 
The original PR was unopenable so here is the new PR.
We have decided to drop ipaddress calls for now.
I have ran acceptance test and it passed.

stack72:
Hi @ayudemura 

Thanks so much for opening the PR to add the functionality again :) Please can you let me know what environment variables i should use when testing this PR? I see new ones have been added for the tests of volume
Are ENV VARs the only way this can be done?

P.

ayudemura:
@stack72 
You can use `PACKET_PROJECT_ID='e9f7eaad-09cf-44ad-8022-4d1322d9ce87' PACKET_FACILITY='ewr1'`as of the new env (pass as parameter as well).
I think we try not to hardcode them since the instance with given id could be deleted by the time of running test.  You were already successfully running the acceptance test before, so I am assuming that you already have appropriate `PACKET_AUTH_TOKEN` env.  If you don't, then please let me know so that I can generate and provide you the safe token.

stack72:
Hi @ayudemura 

Thanks for the help in understand the PR here :)

the tests look good to me!

```
% make testacc TEST=./builtin/providers/packet TESTARGS='-run=TestAccPacket'
2016/08/15 08:36:47 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/packet -v -run=TestAccPacket -timeout 120m
=== RUN   TestAccPacketProject_Basic
--- PASS: TestAccPacketProject_Basic (8.23s)
=== RUN   TestAccPacketSSHKey_Basic
--- PASS: TestAccPacketSSHKey_Basic (4.48s)
=== RUN   TestAccPacketVolume_Basic
--- PASS: TestAccPacketVolume_Basic (28.60s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/packet	41.330s
```

I made a small change which allows us to eliminate the need for environment variables. We now create a Project as part of the test, that way we can spin up and tear down the entire env each time ffd04882e7072d93ab194a306df6f2a0421d0125

Apart from that, it all looks good! Thanks so much

Paul

linuxus:
Hello and thanks for this new feature of packet. I was hoping that there will be an argument to pass into this resource for attachment. Once the volume is created we will need a way to attach it to a Packet device (aka server). I looked at the device resource argument and there isn't an attach to volume available either.
Thanks.

PR #8147, opened 2016-08-11
provider/google: Support Import of 'google_compute_instance_template'

nwwebb:
```
% make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccComputeInstanceTemplate_'
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccComputeInstanceTemplate_ -timeout 120m
=== RUN   TestAccComputeInstanceTemplate_importBasic
--- PASS: TestAccComputeInstanceTemplate_importBasic (27.36s)
=== RUN   TestAccComputeInstanceTemplate_importIp
--- PASS: TestAccComputeInstanceTemplate_importIp (52.18s)
=== RUN   TestAccComputeInstanceTemplate_importDisks
--- PASS: TestAccComputeInstanceTemplate_importDisks (27.42s)
=== RUN   TestAccComputeInstanceTemplate_importSubnetAuto
--- PASS: TestAccComputeInstanceTemplate_importSubnetAuto (140.49s)
=== RUN   TestAccComputeInstanceTemplate_importSubnetCustom
--- PASS: TestAccComputeInstanceTemplate_importSubnetCustom (95.10s)
=== RUN   TestAccComputeInstanceTemplate_basic
--- PASS: TestAccComputeInstanceTemplate_basic (27.06s)
=== RUN   TestAccComputeInstanceTemplate_IP
--- PASS: TestAccComputeInstanceTemplate_IP (52.22s)
=== RUN   TestAccComputeInstanceTemplate_disks
--- PASS: TestAccComputeInstanceTemplate_disks (27.33s)
=== RUN   TestAccComputeInstanceTemplate_subnet_auto
--- PASS: TestAccComputeInstanceTemplate_subnet_auto (129.91s)
=== RUN   TestAccComputeInstanceTemplate_subnet_custom
--- PASS: TestAccComputeInstanceTemplate_subnet_custom (86.55s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	
665.682s
```
@lwander

### DIFF
--- a/builtin/providers/google/resource_compute_instance_template.go
+++ b/builtin/providers/google/resource_compute_instance_template.go
@@ -3,6 +3,7 @@ package google
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -15,6 +16,9 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 		Create: resourceComputeInstanceTemplateCreate,
 		Read:   resourceComputeInstanceTemplateRead,
 		Delete: resourceComputeInstanceTemplateDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
@@ -66,6 +70,7 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 							Type:     schema.TypeBool,
 							Optional: true,
 							ForceNew: true,
+							Computed: true,
 						},
 
 						"device_name": &schema.Schema{
@@ -90,6 +95,7 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							ForceNew: true,
+							Computed: true,
 						},
 
 						"source_image": &schema.Schema{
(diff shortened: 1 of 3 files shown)


builtin/providers/google/resource_compute_instance_template.go
```diff
@@ -134,7 +143,6 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 			"automatic_restart": &schema.Schema{
 				Type:       schema.TypeBool,
 				Optional:   true,
-				Default:    true,
```
lwander: Have you verified that this works without `automatic_restart` set? Also removing this default could be dangerous to existing configs.

stack72: @lwander you are correct - if we are going to change a default, then we may possibly need a state_migration

nwwebb: I'm not sure a state_migration will be able to fix the issue. Previously a config without auto_restart or scheduler set will create a default scheduler with auto_restart set to true (which seems like odd behavior to me, especcially for a deprecated field). Taking away this default will make configs such as this destroy the scheduler. To keep the scheduler it seems like I would need to migrate configs rather than states somehow. Thoughts on this?

nwwebb: Actually I just went into the gce api and if scheduler is not specified it will still make a scheduler with the default attributes, so keeping the default should be fine. I'm going to make the other properties of the scheduler have defaults set to avoid confusion in the future.

lwander: Nice work! One last little change, since your adding a default to `preemptible`, could you make it clear that the default is `false` in the docs? https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/google/r/compute_instance_template.html.markdown

nwwebb: Changed the docs

PR #8151, opened 2016-08-12
docs: Fix invalid reference in iam_policy_document document page

kjmkznr:


stack72:
Hi @kjmkznr 

Thanks so much for this - this is a good fix

Paul

PR #8155, opened 2016-08-12
provider/openstack: Add support for 'value_specs' param on 'openstack…

fatmcgav:
…_networking_network_v2' provider.

This can be used to pass additional custom values to the network
resource upon creation.


jtopjian:
Thanks, @fatmcgav!

PR #8164, opened 2016-08-12
Fix panic when there's no state to print

jbardin:


command/state_show.go
```diff
@@ -45,22 +45,27 @@ func (c *StateShowCommand) Run(args []string) int {
...
 	instance, err := c.filterInstance(results)
 	if err != nil {
 		c.Ui.Error(err.Error())
 		return 1
 	}
+
 	is := instance.Value.(*terraform.InstanceState)
 
 	// Sort the keys
-	keys := make([]string, 0, len(is.Attributes))
```
phinze: It seems like you _hate_ performance!?!? 😆 

The allocs, think of the allocs!!! 😛

phinze:
LGTM, bonus snark inline 😄

command/state_show.go
jbardin: well, if there's enough allocs, `append` amortizes to `O(1)`! ;P

Issue #8032, opened 2016-08-07
terraform taint null_resource doesn't work anymore (v0.7.x).


lesaux:
It seems `terraform taint null_resource` doesn't work in v0.7.1 anymore (I'm using Terraform v0.7.1-dev).

```
resource "null_resource" "some_command" {

    provisioner "remote-exec" {
        connection {
            host = "somehost"
            user = "someUser"
        }
        inline = [
          "sudo docker run -d --restart=always -p 8080:8080 someimage
        ]
    }

}
```

```
$ terraform taint null_resource.some_command
The resource null_resource.some_command in the module root has been marked as tainted!
$ terraform plan
...
null_resource.some_command: Refreshing state...
...
No changes. Infrastructure is up-to-date. This means that Terraform
could not detect any differences between your configuration and
the real physical resources that exist. As a result, Terraform
doesn't need to do anything.
```

afritzler:
Same issue here. I am on terraform 0.7.0.

phinze:
Thanks for the report! This is fixed in #8167 which will come out with v0.7.1 due out next week. 👍

Issue #8072, opened 2016-08-09
tainted resource not enforced on plan or apply for resources using count.  v0.7


elliottucker-swrve:
### Terraform Version
Terraform v0.7.0
OSX 10.11.6
installed via homebrew

### Affected Resource(s)
- aws_instance

### Terraform Configuration Files
```hcl
variable "host_count" {
  default = {
    "web" = 2
    }
}

resource "aws_instance" "web" {
  count = "${var.host_count["web"]}"
  ami = "${lookup(var.images,var.region)}"
  instance_type = "c3.large"
  subnet_id = "${element(aws_subnet.internal.*.id, count.index)}"
  vpc_security_group_ids = ["${var.sg_default_id}","${aws_security_group.web.id}"]
  iam_instance_profile = "iam-web"
  key_name = "secret"
  ebs_block_device {device_name = "/dev/xvdf" volume_size = 50 volume_type = "standard"}
  lifecycle {
    ignore_changes = ["ami", "user-data"]
  }
  tags {
    Name = "web-${count.index}"
    Role = "Web"
    Stack = "${var.stackname}"
  }

  user_data = <<EOF
#!/bin/bash
## stuff
EOF

```

### Expected Behavior

I am running the command:
`terraform taint aws_instance.web.0`
This returns:

> The resource aws_instance.web.0 in the module root has been marked as tainted!

When I run plan and apply, i'd expect the instance to be destroyed and recreated.

### Actual Behavior
When I run plan and apply: 
> No changes. Infrastructure is up-to-date. This means that Terraform
> could not detect any differences between your configuration and
> the real physical resources that exist. As a result, Terraform
> doesn't need to do anything.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform taint aws_instance.web.0`
2. `terraform plan`
3. `terraform apply`

### Important Factoids
Using s3 remote state
`terraform remote config -backend s3 -backend-config="bucket=mybucket" -backend-config="key=terraform/stackname" -backend-config="region=us-east-1"`

Running `terraform show` shows the resource as tainted
> aws_instance.web.0: (tainted)

mengesb:
I think I'm seeing this as well; but count doesn't seem to matter either. I have a file generated by template that I dump to the filesystem, when I taint that resource it doesn't re-write the file.

I then converted to the new `data` source, and it shows my template items twice when using 'terraform state show data.template_file.knife-rb` resource

cbarbour:
[...]

PR #8167, opened 2016-08-12
terraform: diffs with only tainted set are non-empty

phinze:
Fixes issue where a resource marked as tainted with no other attribute
diffs would never show up in the plan or apply as needing to be
replaced.

One unrelated test needed updating due to a quirk in the testDiffFn
logic - it adds a "type" field diff if the diff is non-Empty. NBD


phinze:
Will fix #8032 and #8072

mitchellh:
LGTM. Looks like we'll be chasing down `Empty` and `Equal` bugs for the rest of our days.

PR #8172, opened 2016-08-13
provider/openstack: Volume Attachment and Detachment Fixes

### LINKED ISSUE #8040: Openstack volume does not detect detach
### Descirption
After creating an image with a root volume and an additional volume attached if the terraform configuration file is changed to remove the attachment (the `volume` object in the `openstack_compute_instance_v2` resource) the `terraform plan` command does not detect the change

### Terraform Version
```
$ terraform -v
Terraform v0.7.0
```

### Affected Resource(s)
The issue affects these resources:
- openstack_blockstorage_volume_v1
- openstack_compute_instance_v2

This issue occurs after modifying the configuration file. The initial file is the following

```hcl
resource "openstack_blockstorage_volume_v1" "root_volume_1" {
  name = "root_volume_1"
  size = 10
  image_id = "90ee49d8-1c18-45ca-8a72-14ffd120521e"
}

resource "openstack_blockstorage_volume_v1" "additional_volume_1" {
  name = "additional_volume_1"
  size = 5
}

resource "openstack_compute_instance_v2" "test_name" {
  name = "test_name"
  flavor_id = "3"
  key_pair = "test-keypair"
  security_groups = ["default"]

  block_device {
    uuid = "${openstack_blockstorage_volume_v1.root_volume_1.id}"
    source_type = "volume"
    boot_index = 0
    destination_type = "volume"
    delete_on_termination = false
  }

  volume {
    volume_id = "${openstack_blockstorage_volume_v1.additional_volume_1.id}"
  }

  metadata {
    application = "app1"
  }

  network {
    name = "network"
  }
}
```

After applying this configuration remove the `volume` object in the vm and you get the following file

```hcl
resource "openstack_blockstorage_volume_v1" "root_volume_1" {
  name = "root_volume_1"
  size = 10
  image_id = "90ee49d8-1c18-45ca-8a72-14ffd120521e"
}

resource "openstack_blockstorage_volume_v1" "additional_volume_1" {
  name = "additional_volume_1"
  size = 5
}

resource "openstack_compute_instance_v2" "test_name" {
  name = "test_name"
  flavor_id = "3"
  key_pair = "test-keypair"
  security_groups = ["default"]

  block_device {
    uuid = "${openstack_blockstorage_volume_v1.root_volume_1.id}"
    source_type = "volume"
    boot_index = 0
    destination_type = "volume"
    delete_on_termination = false
  }

  metadata {
    application = "app1"
  }

  network {
    name = "network"
  }
}
```

### Debug Output
The debug output is in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/61b7cea1bf96c8e74199b9ffb300d40c)

### Expected Behavior
The `terraform plan` command should detech the need to detach the `openstack_blockstorage_volume_v1.additional_volume_1`

### Actual Behavior
The `terraform plan` command did not detect any changes

### Steps to Reproduce
1. create a config.tf file as reported above
2. `terraform apply`
3. modify the config.tf file removing the volume element from the vm (see above)
4. `terraform plan`

### References
No known issues

### LINKED ISSUE #8042: Openstack volume does not move to another instance
### Description
When trying to detatch a volume from one instance and attaching it to another instance the `terraform plan` command does not detect the detach and the `terraform apply` command fails trying to attach the volume without detatching it first.

### Terraform Version
```
$ terraform -v
Terraform v0.7.0
```

### Affected Resource(s)
The issue affects these resources:
- openstack_blockstorage_volume_v1
- openstack_compute_instance_v2

This issue occurs after modifying the configuration file. 
The configuration file is a bit long so here it goes the [gist](https://gist.github.com/GiovanniPaoloGibilisco/cdc08b310e0a359b03860498a73f3e8e)

After applying this configuration move the `volume` object from the `test_name` instance to the `test_name2` instance. 
Still the configuration file is a bit long so here gors the [gist](https://gist.github.com/GiovanniPaoloGibilisco/9e8d433ca7a41001cbe63ddd16f2bc65)

### Debug Output
The issue is related to two commands. The `terraform plan` which does not detect the detachment of the volume from the first instance and the `terraform apply` which fails in attaching the volume to the new isntance
The debug output of the `terraform plan` command is in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/79f64227521e792185fb5337b0e88740)

The debug output of the `terraform apply` command is in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/f34dc9923b8883e3115e02ddb0e1cc03)

### Expected Behavior
The volume should be detached from the first instance and attached to the second one.

### Actual Behavior
The `terraform plan` command detects only a new attachment on the second instance
The `terraform apply` command fails with
```
Error applying plan:

1 error(s) occurred:

* openstack_compute_instance_v2.test_name_2: Expected HTTP response code [200] when accessing [POST http://xxx.xxx.xxx.xxx:8774/v2/c660af04fabe41f29787fa27ad3400e9/servers/1b823818-d2b6-4e58-ad6f-d2476507205a/os-volume_attachments], but got 400 instead
{"badRequest": {"message": "Invalid volume: volume 'f8871928-3712-4c7a-bd0b-5bb1e5488358' status must be 'available'. Currently in 'in-use'", "code": 400}}
```
### Steps to Reproduce
1. create a config.tf file as in the first  [gist](https://gist.github.com/GiovanniPaoloGibilisco/cdc08b310e0a359b03860498a73f3e8e)
2. `terraform apply`
3. modify the config.tf file moving the volume element from one vm top the other (see second [gist](https://gist.github.com/GiovanniPaoloGibilisco/f34dc9923b8883e3115e02ddb0e1cc03))
4. `terraform plan`
5. `terraform apply`

### References
- GH-8040

### LINKED ISSUE #8045: Openstack volume state not updated after VM removal
### Description
If a VM with external volumes is created and then destroyed, leaving the volumes intact, the state of the volumes is not updated after the vm deletion until a refresh is performed. 

### Terraform Version
```
$ terraform -v
Terraform v0.7.0
```

### Affected Resource(s)
The issue affects these resources:
- openstack_blockstorage_volume_v1
- openstack_compute_instance_v2

This issue occurs after modifying the configuration file.  The initial configuration is the following
```hcl
resource "openstack_blockstorage_volume_v1" "root_volume_1" {
  name = "root_volume_1"
  size = 10
  image_id = "90ee49d8-1c18-45ca-8a72-14ffd120521e"
}

resource "openstack_blockstorage_volume_v1" "additional_volume_1" {
  name = "additional_volume_1"
  size = 5
}


resource "openstack_compute_instance_v2" "test_name" {
  name = "test_name"
  flavor_id = "3"
  key_pair = "test-keypair"
  security_groups = ["default"]

  block_device {
    uuid = "${openstack_blockstorage_volume_v1.root_volume_1.id}"
    source_type = "volume"
    boot_index = 0
    destination_type = "volume"
    delete_on_termination = false
  }

  volume {
    volume_id = "${openstack_blockstorage_volume_v1.additional_volume_1.id}"
  }

  metadata {
    application = "app1"
  }

  network {
    name = "netwrok"
  }
}
```

after applying this configuration remove the VM obtaining: 

```hcl
resource "openstack_blockstorage_volume_v1" "root_volume_1" {
  name = "root_volume_1"
  size = 10
  image_id = "90ee49d8-1c18-45ca-8a72-14ffd120521e"
}

resource "openstack_blockstorage_volume_v1" "additional_volume_1" {
  name = "additional_volume_1"
  size = 5
}
```

### Debug Output
The debug output of the `terraform apply` command that removes the vm is available in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/743ad33fb888da99a53895d68429fa4d)
The terraform state file after the execution of this command is available in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/0ec03278197f370348e6bb173b130504)

The debug output of the `terraform refresh` command is available in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/72543ce7c6470a7c716469b4588d2776)
The (correct) terraform state file after the execution of this command is available in this [gist](https://gist.github.com/GiovanniPaoloGibilisco/fc0b5b6381cd297200f4e22b8859d64e)

### Expected Behavior
The command `terraform show` issued after the removal of the vm should show two volumes not attached to any machine. It should be consistent with the output of the same command after the `terraform refresh` (assuming no other change happened in the mean time)

### Actual Behavior
Issuing `terraform show` after the deletion of the vm shows two volumes still attached to the vm, which does not exist.

After issuing a `terraform refresh` the `terraform show` command shows correctly the two volumes not attached. 

### Steps to Reproduce
1. create a config.tf file as described above with 2 volumes and 1 vm
2. `terraform apply`
3. remove from the file the specification of the vm
4. `terraform apply`
5. `terraform show` 
6. `terraform refresh`
7. `terraform show`

### References
Might refer to:
- GH-8040 (it also deals with attachments of volumes to VMs)

jtopjian:
This commit is changing the `volumes` block from being computed to non-computed.
This change makes the Terraform configuration the source of truth about volumes
attached to the instance and therefore is able to correctly detect when a user
detaches a volume during an update.

One thing to be aware of is that if a user attached a volume out of band of an
instance controlled by Terraform, that volume will be detached upon the next
apply. The best thing to do is add a `volume` entry in the instance's
configuration of any volumes that were attached out of band.

This commit also explicitly detaches volumes from an instance before the
instance terminates. Most Block Storage volume drivers account for this
scenario internally, but there are a few that don't. This change is to support
those that don't.

In addition, when volumes are read by the instance, volumes configured in the
Terraform configuration are the source of truth. Previously, a call was being
made to OpenStack to provide the list of attached volumes.

It also adds a few new tests and fixes existing tests for various volume
attach-related scenarios.

Fixes #8040 
Fixes #8042 
Fixes #8045

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
@@ -24,7 +24,6 @@ import (
 	"github.com/rackspace/gophercloud/openstack/compute/v2/flavors"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/images"
 	"github.com/rackspace/gophercloud/openstack/compute/v2/servers"
-	"github.com/rackspace/gophercloud/pagination"
 )
 
 func resourceComputeInstanceV2() *schema.Resource {
@@ -238,17 +237,16 @@ func resourceComputeInstanceV2() *schema.Resource {
 			"volume": &schema.Schema{
 				Type:     schema.TypeSet,
 				Optional: true,
-				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"id": &schema.Schema{
 							Type:     schema.TypeString,
+							Optional: true,
 							Computed: true,
 						},
 						"volume_id": &schema.Schema{
 							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Required: true,
 						},
 						"device": &schema.Schema{
 							Type:     schema.TypeString,
@@ -348,12 +346,6 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
 
 	flavorId, err := getFlavorID(computeClient, d)
 	if err != nil {
-		return err
-	}
-
-	// determine if volume configuration is correct
-	// this includes ensuring volume_ids are set
-	if err := checkVolumeConfig(d); err != nil {
 		return err
 	}
 
@@ -712,16 +704,12 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
 	}
 
 	if d.HasChange("volume") {
-		// ensure the volume configuration is correct
-		if err := checkVolumeConfig(d); err != nil {
-			return err
-		}
-
 		// old attachments and new attachments
 		oldAttachments, newAttachments := d.GetChange("volume")
 		// for each old attachment, detach the volume
 		oldAttachmentSet := oldAttachments.(*schema.Set).List()
 
+		log.Printf("[DEBUG] Attempting to detach the following volumes: %#v", oldAttachmentSet)
 		if blockClient, err := config.blockStorageV1Client(d.Get("region").(string)); err != nil {
 			return err
 		} else {
@@ -812,6 +800,22 @@ func resourceComputeInstanceV2Delete(d *schema.ResourceData, meta interface{}) e
 	computeClient, err := config.computeV2Client(d.Get("region").(string))
 	if err != nil {
 		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
+	// Make sure all volumes are detached before deleting
+	volumes := d.Get("volume")
+	if volumeSet, ok := volumes.(*schema.Set); ok {
+		volumeList := volumeSet.List()
+		if len(volumeList) > 0 {
+			log.Printf("[DEBUG] Attempting to detach the following volumes: %#v", volumeList)
+			if blockClient, err := config.blockStorageV1Client(d.Get("region").(string)); err != nil {
+				return err
+			} else {
+				if err := detachVolumesFromInstance(computeClient, blockClient, d.Id(), volumeList); err != nil {
+					return err
+				}
+			}
+		}
 	}
 
 	if d.Get("stop_before_destroy").(bool) {
@@ -1397,6 +1401,7 @@ func detachVolumesFromInstance(computeClient *gophercloud.ServiceClient, blockCl
 		va := v.(map[string]interface{})
 		aId := va["id"].(string)
 
+		log.Printf("[INFO] Attempting to detach volume %s", va["volume_id"])
 		if err := volumeattach.Delete(computeClient, serverId, aId).ExtractErr(); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


jtopjian:
@GiovanniPaoloGibilisco thank you very much for your bug reports. I believe this PR should fix all of them. Would you be able to test and verify?

A note about #8042: if you look at the test I wrote, you'll see I had to add an explicit `depends_on` between the instances. This is to avoid a race condition of both instances being updated in parallel and thus acting on the volume at the same time. `depends_on` is a bit of a workaround... a better solution would be something like a `openstack_compute_volumeattach_v2` resource that was dedicated to the relationship between a volume and instance. That's on my list of things to look into.

Let me know if you have any questions!

jtopjian:
Some notes about volume attachment in general (both as a brain dump for myself and in case this is helpful to anyone looking into the OpenStack provider):

For those not familiar with OpenStack, keep in mind that OpenStack is a collection of projects that provide APIs to different services. Sometimes a project provides "proxy" support to another project. For example, the OpenStack Block Storage API has two major versions in the wild: v1 and v2. OpenStack Compute provides an `os-volumes_attach` API that will attach a block storage volume to an instance by only using the Compute API. The developer only needs to work with the Compute API. It will handle the work of determining whether to use Block Storage v1 or v2.

Other cloud providers support the attachment of volumes during instance creation. OpenStack doesn't have that kind of feature, so to mimic it in Terraform, a separate call is made to attach volumes after the instance has been successfully created.

Because volume attachment is not native to instance creation, it's arguably best moved to a new resource like:

```hcl
openstack_compute_volume_attach_v2 {
  instance_id = "${openstack_compute_instance_v2.name.id}"
  volume_id = "${openstack_blockstorage_volume_v1.name.id}"
}
```

As mentioned earlier, OpenStack Nova/Compute provides the `os-volumes_attach` API that can handle volume attachment... except that `os-volumes_attach` doesn't report the status of the attachment or the status of the volume. This means a call is made to attach a volume and if the volume is still in a `CREATING` state, an error will be returned. The best that can be done within the Compute API is to just retry when an error occurs and hopefully the errors eventually stop. I'm not a fan of that. The same goes for detaching as well.

The next option is to integrate the Block Storage API, so an attach is triggered by the Compute API and the attachment status is detected by the Block Storage API. This is what the `openstack_compute_instance_v2` resource currently does. There's a problem with this, though: It's hard-coded to use the Block Storage v1 API. This needs resolved at some point in time.

A third option is to just make volume attachment entirely within the Block Storage API. That would work well for Block Storage v2 since it has an attachment API, but Block Storage v1 actually relies on the Compute API for attachment. 😵 

So at the moment, this is what I'm thinking for a long-term goal:

1. Deprecate the `volumes` block in `openstack_compute_instance_v2`.
2. Create `openstack_blockstorage_volume_attach_v1` that will use both the Compute API and Block Storage v1 API to attach Block Storage v1 volumes.
2. Create `openstack_blockstorage_volume_attach_v2` that will use the Block Storage v2 api to attach Block Storage v2 volumes.

The downside of breaking out volume attachment into their own resources is that if a user wants to wait for all volumes to finish attaching before provisioning, they will need to use a null resource or something similar. I think that's acceptable as that type of workflow is usually inevitable with more complex Terraform configurations.

PR #8179, opened 2016-08-14
Fix typo in OpenStack LBaaSv2 pool resource

dgethx:
This PR fixes a typo which prevents creating an OpenStack LBaaSv2 TCP pool.

```
resource "openstack_lb_pool_v2" "lb1_pool_1" {
  protocol    = "TCP"
  lb_method   = "ROUND_ROBIN"
  listener_id =  "${openstack_lb_listener_v2.listener_1.id}"
}
```

```
Errors:
  * openstack_lb_pool_v2.lb1_pool_1: Only 'TCP', 'HTTP', and 'HTTPS' are supported values for 'protocol'
```


jtopjian:
Nice catch. Thank you!

PR #8181, opened 2016-08-15
provieder/openstack: Add 'value_specs' support for openstack_networki…

fatmcgav:
…ng_subnet_v2 provider.

Updated provider documentation to support.


jtopjian:
Thank you!

PR #8194, opened 2016-08-15
docs/import: Add a note to the Import section that only ENV VARs can be used atm

stack72:


website/source/docs/import/usage.html.md
```diff
@@ -22,6 +22,9 @@ $ terraform import aws_instance.bar i-abcd1234
 ...
 ```
 
+**Note** In order to import resources, the provider should be configured with environment variables.
```
radeksimko: Probably a nitpick, but I'd make this more visible, e.g. via `-> ` which would enclose this paragraph in a blue rectangle. 😉

radeksimko:
Except that small nitpick this LGTM.

PR #8202, opened 2016-08-15
testing: Add TestCheckResourceAttrSet helper

jen20:
This commit adds a `TestCheckFunc` which ensures that a value is set for a given name/key combination. It is primarily useful for ensuring that computed values are set where it is not possible to know the expected value ahead of time.


mitchellh:
LGTM

PR #8205, opened 2016-08-15
aws: Make EFS Mount Target creation fail for 2+ targets per AZ

radeksimko:
The current real limitation is 1 mount target per AZ, Terraform does not enforce that in any way - ain't saying it should - it just relies on API returning errors. However if the user happens to create 2 mount targets in two different subnets, both in the same AZ, the API simply returns the exact same mount target ID which causes Terraform to manage the same mount target as two resources.

See snippet from a debug log:
https://gist.github.com/radeksimko/f1e3dd1077f18495726528b175b0bb47
and full repro case:
https://gist.github.com/radeksimko/f09b8d8db2764d908b9a009d4df802e0

Acceptance tests:

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=AWSEFSMountTarget'
```
```
TF_ACC=1 go test ./builtin/providers/aws -v -run=AWSEFSMountTarget -timeout 120m
=== RUN   TestAccAWSEFSMountTarget_importBasic
--- PASS: TestAccAWSEFSMountTarget_importBasic (251.04s)
=== RUN   TestAccAWSEFSMountTarget_basic
--- PASS: TestAccAWSEFSMountTarget_basic (422.77s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	673.840s
```

The plan is (as usually for similar bugs) to report this to AWS, but I can imagine this being a non-trivial fix, so I would not expect it to be fixed any time soon.

### DIFF
--- a/builtin/providers/aws/resource_aws_efs_mount_target.go
+++ b/builtin/providers/aws/resource_aws_efs_mount_target.go
@@ -67,9 +67,24 @@ func resourceAwsEfsMountTarget() *schema.Resource {
 func resourceAwsEfsMountTargetCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).efsconn
 
+	fsId := d.Get("file_system_id").(string)
+	subnetId := d.Get("subnet_id").(string)
+
+	// CreateMountTarget would return the same Mount Target ID
+	// to parallel requests if they both include the same AZ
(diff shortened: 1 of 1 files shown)


mitchellh:
Not foolproof but I also think not a bad idea. Looks great to me. Merge at your leisure.

Issue #4390, opened 2015-12-18
Expose the discovered account_id via the aws provider


gozer:
My precise use-case is really simple ,I am using terraform to build an IAM policy.

However, I am building the ARN of a known resource (not created by terraform), and it needs the account id. For now, I am making do by providing the aws account-id as a variable, alongside the secret-key and access-key, but it's one more piece of data that could be incorrect.

Since, from my looking at the provider's code, it already determines the accountid (for whitelist/blacklist matching), so I suspect exposing it as a variable might be simple and allow me to do ${aws.account_id} or something similar

jen20:
Hi @gozer, this is definitely something we want to do, I was reasonably sure there was another issue opened for it but I can't find it right now.

bradfeehan:
I'd also love to grab the region which is set by `AWS_REGION` or `AWS_DEFAULT_REGION` in the environment.

enaeseth:
I can't find another issue open for this, either, and it would indeed be super useful for defining IAM policies. :smile:

diroussel:
I need this to be able to put the account_id in an S3 policy.

mdevs5531:
+1

egoldschmidt:
+1

vaidik:
+1

jmahowald:
Not very proud of this, as it's making a side effect in a user edited file, but this is what I did
```
cat output_account_id_var.sh
#!/bin/bash

#HT: http://stackoverflow.com/questions/33791069/quick-way-to-get-aws-account-number-from-the-cli-tools
account_id=`aws ec2 describe-security-groups --group-names 'Default' --query 'SecurityGroups[0].OwnerId' --output text`
echo "account_id =\"$account_id\""
```

Then in my makefile I insert into tfvars

```
get_vars:
        if grep 'account_id =' terraform.tfvars; \
        then echo "already has account id"; \
        else ./output_account_id_var.sh >> terraform.tfvars; \
        fi

plan: .terraform get_vars
```

ajcrites:
Note that this is acquired from `aws_ecr_repository` `registry_id` attribute. If you are creating an ECR and you can create it before you need the user ID to create ARNs, you can use that attribute. This is not ideal, but I think it is better than relying on a variable if you don't have to.

cahlbin:
+1

kaaloo:
+1

stahs:
+1

cswingler:
+1

[...]

PR #8206, opened 2016-08-15
provider/aws: Add aws_account_id data source

jen20:
This data source provides access during configuration to the ID of the AWS account for the connection to AWS. It is primarily useful for interpolating into policy documents, for example when creating the policy for an ELB or ALB access log bucket.

This will need revisiting and further testing once the work for AssumeRole is integrated.

### DIFF
--- a/builtin/providers/aws/data_source_aws_caller_identity.go
+++ b/builtin/providers/aws/data_source_aws_caller_identity.go
@@ -1,0 +1,40 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsCallerIdentity() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsCallerIdentityRead,
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsCallerIdentityRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*AWSClient)
+
+	log.Printf("[DEBUG] Reading Caller Identity.")
+	d.SetId(time.Now().UTC().String())
+
+	if client.accountid == "" {
+		log.Println("[DEBUG] No Account ID available, failing")
+		return fmt.Errorf("No AWS Account ID is available to the provider. Please ensure that\n" +
+			"skip_requesting_account_id is not set on the AWS provider.")
+	}
+
+	log.Printf("[DEBUG] Setting AWS Account ID to %s.", client.accountid)
+	d.Set("account_id", meta.(*AWSClient).accountid)
+
+	return nil
+}
--- a/website/source/docs/providers/aws/d/caller_identity.html.markdown
+++ b/website/source/docs/providers/aws/d/caller_identity.html.markdown
@@ -1,0 +1,35 @@
+---
+layout: "aws"
+page_title: "AWS: aws_caller_identity"
+sidebar_current: "docs-aws-datasource-caller-identity"
+description: |-
+  Get information about the identity of the caller for the provider
+  connection to AWS.
+---
+
+# aws\_caller\_identity
+
+Use this data source to get the access to the effective Account ID in
+which Terraform is working.
+
+~> **NOTE on `aws_caller_identity`:** - an Account ID is only available
+if `skip_requesting_account_id` is not set on the AWS provider. In such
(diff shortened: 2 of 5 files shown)


apparentlymart:
This seems to be an implementation for #4390. (Just noting for the benefit of those following the other one.)

jen20:
@apparentlymart I prefer your terminology as it allows for expansion later. I'll update this PR.

radeksimko:
Re terminology: keep in mind the account ID _may not_ always come from STS at this point, although we could probably ditch all other methods now when we have `STS:GetCallerIdentity`. It seems to be very reliable way of getting account ID. I understand your point about expansion though.

Although it may be obvious, would you mind adding a warning/notice to docs that account ID may be unavailable under certain circumstances?

https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/index.html.markdown => if `skip_requesting_account_id == true`

Also the data source should IMO be returning error in such case ^.

jen20:
@radeksimko, @apparentlymart - Thanks for the feedback here. I've adjusted the terminology to match the discussion in #4390 and added the error condition suggested by @radeksimko!

builtin/providers/aws/data_source_aws_caller_identity.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+func dataSourceAwsCallerIdentityRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*AWSClient)
+
+	log.Printf("[DEBUG] Reading Caller Identity.")
+	d.SetId(time.Now().UTC().String())
+
+	if client.accountid != "" {
```
apparentlymart: This looks inverted... don't we want to produce this error if the account id *is* empty?

website/source/layouts/aws.erb
```diff
@@ -13,12 +13,16 @@
...
                     <ul class="nav nav-visible">
+
                         <li<%= sidebar_current("docs-aws-datasource-ami") %>>
                             <a href="/docs/providers/aws/d/ami.html">aws_ami</a>
                         </li>
                         <li<%= sidebar_current("docs-aws-datasource-availability-zones") %>>
                             <a href="/docs/providers/aws/d/availability_zones.html">aws_availability_zones</a>
                         </li>
+                        <li<%= sidebar_current("docs-aws-datasource-caller-identity") %>>
+                          <a href="/docs/providers/aws/d/caller_identity.html">aws_account_id</a>
```
apparentlymart: The caption here is still `aws_account_id` rather than `aws_caller_identity`

website/source/docs/providers/aws/d/caller_identity.html.markdown
```diff
@@ -0,0 +1,35 @@
...
+if `skip_requesting_account_id` is not set on the AWS provider. In such
+cases, the data source will return an error.
+
+## Example Usage
+
+```
+data "aws_caller_identity" "current" { }
+
+output "account_id" {
+  value = "${data.aws_caller_identity.current.id}"
```
apparentlymart: I think this should be `${data.aws_caller_identity.current.account_id}`

builtin/providers/aws/data_source_aws_caller_identity.go
jen20: Hah yes. Apparently last minute alterations are what code review is intended to catch ;-)

stack72:
LGTM! Nice addition :)

builtin/providers/aws/data_source_aws_caller_identity.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsCallerIdentity() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsCallerIdentityRead,
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
```
radeksimko: This field is not actually set in `Read`, is it?

jen20:
@radeksimko fixed up now.

```
› envchain aws.terraform make testacc TEST=./builtin/providers/aws TESTARGS=" -run TestAccAWSCallerIdentity_basic"
2016/08/16 11:30:19 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run TestAccAWSCallerIdentity_basic -timeout 120m
=== RUN   TestAccAWSCallerIdentity_basic
--- PASS: TestAccAWSCallerIdentity_basic (11.55s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	11.575s
```

PR #8211, opened 2016-08-16
Update Vagrantfile to use Ubuntu 14.04 LTS

cblecker:
Update the Vagrantfile to use the slightly more up to date Ubuntu 14.04 LTS bento image. Tested and verified go script and build functions as expected.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -43,7 +43,7 @@ source /etc/profile.d/gopath.sh
 SCRIPT
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = "bento/ubuntu-12.04"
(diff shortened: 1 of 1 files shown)


stack72:
Hi @cblecker 

Should we use 16.04 now instead?

P.

jen20:
I think we need to have a more general overhaul of the Vagrantfile if we're going to keep it. It's still using Go 1.6 (3 point releases behind) also. This is good start though - IMO we should aim for Ubuntu 16.04 if we're going to stick with Linux - a better alternative would be FreeBSD so that tools like DTrace are available under Vagrant.

jen20:
Thanks @cblecker! We'll go with this for now, and then at some point in the future maybe consider upgrading again.

PR #8215, opened 2016-08-16
helper/signalwrapper and azurerm_storage_account listens for signals

mitchellh:
`azurerm_storage_account` creation can take a _long_ time (24+ hours). If you Ctrl-C, Terraform currently waits for this to complete before cancelling but for this particular resource it is too long.

This PR introduces a new package `helper/signalwrapper` for easily wrapping functions that can cancel based on signals and early exiting. This package is used with the `azurerm_storage_account` resource to implement cancellability.

I tested this and it definitely responds to Ctrl-C now. I'm however extremely skeptical in the Azure SDK that it is doing anything other than simply exiting. We don't receive an ID in the cancelled case and it cancels _immediately_. I'm really scared that we're actually just dangling a resource this way but I haven't gotten the storage account to show up _yet_. I just can't believe this is actually working, so I'm going to test it more. I'd love your thoughts @stack72 on the Azure stuff. Besides this, 95+% of the LoC in this PR are fine for review and well tested.

So please review but allow me to merge once I verify more behavior.

**WARNING: This is not a long term solution.** There are very subtle races here: if you Ctrl-C before it reaches the code to create the storage account you could end up waiting anyways. And its also just unclean that provider code is listening for OS signals. In the future, I'd like to introduce a Stop or Cancel API to the `ResourceProvider` interface that can bubble a stop request through, therefore avoiding the race conditions and also abstracting providers away from "signals" to an API.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_storage_account.go
+++ b/builtin/providers/azurerm/resource_arm_storage_account.go
@@ -143,6 +143,10 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 	// since we will likely also want to introduce a time-based timeout.
 	var createErr error
 	select {
+	case <-time.After(1 * time.Hour):
+		// An hour is way above the expected P99 for this API call so
+		// we premature cancel and error here.
+		createErr = wrap.Cancel()
 	case createErr = <-wrap.ErrCh:
 		// Successfully ran (but perhaps not successfully completed)
 		// the function.
--- a/builtin/providers/azurerm/resource_arm_storage_account.go
+++ b/builtin/providers/azurerm/resource_arm_storage_account.go
@@ -155,7 +155,7 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 	read, err := storageClient.GetProperties(resourceGroupName, storageAccountName)
 
 	// Set the ID right away if we have one
-	if read != nil && read.ID != nil {
+	if err == nil && read.ID != nil {
 		log.Printf("[INFO] storage account %q ID: %q", storageAccountName, *read.ID)
 		d.SetId(*read.ID)
 	}
--- a/builtin/providers/azurerm/resource_arm_storage_account.go
+++ b/builtin/providers/azurerm/resource_arm_storage_account.go
@@ -141,16 +141,11 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 
 	// Check the result of the wrapped function. I put this into a select
 	// since we will likely also want to introduce a time-based timeout.
-	var err error
+	var createErr error
 	select {
-	case err = <-wrap.ErrCh:
+	case createErr = <-wrap.ErrCh:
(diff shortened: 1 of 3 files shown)


mitchellh:
Confirmed a dangling resource. 😵 I'm going to look into this tomorrow but definitely do not merge. Please feel free to review my direction though.

mitchellh:
Fixed the dangling resource reliably and added a 1 hour timeout as well as requested by @phinze and @stack72

builtin/providers/azurerm/resource_arm_storage_account.go
```diff
@@ -131,16 +132,47 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
...
-		return fmt.Errorf("Error creating Azure Storage Account '%s': %s", storageAccountName, err)
+	// Create the storage account. We wrap this so that it is cancellable
+	// with a Ctrl-C since this can take a LONG time.
+	wrap := signalwrapper.Run(func(cancelCh <-chan struct{}) error {
+		_, err := storageClient.Create(resourceGroupName, storageAccountName, opts, cancelCh)
+		return err
+	})
+
+	// Check the result of the wrapped function. I put this into a select
+	// since we will likely also want to introduce a time-based timeout.
```
phinze: Second sentence no longer accurate, timeout in place.

builtin/providers/azurerm/resource_arm_storage_account.go
```diff
@@ -131,16 +132,47 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
...
+	var createErr error
+	select {
+	case <-time.After(1 * time.Hour):
+		// An hour is way above the expected P99 for this API call so
```
stack72: We may get to see this in action during nightly acceptance tests if they have any issues - that's when we tend to get a lot of failures

builtin/providers/azurerm/resource_arm_storage_account.go
```diff
@@ -131,16 +132,47 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
...
+		d.SetId(*read.ID)
+	}
+
+	// If we had a create error earlier then we return with that error now.
+	// We do this later here so that we can grab the ID above is possible.
+	if createErr != nil {
+		return fmt.Errorf(
+			"Error creating Azure Storage Account '%s': %s",
+			storageAccountName, createErr)
+	}
```
phinze: Should this check be before the follow-on read? Thinking that if the read 404s we mask the create error...

stack72:
This LGTM! The use of cancellation channel will be able to be rolled out to all of the AzureRM resources - that was introduced in their SDK 2.1-beta

builtin/providers/azurerm/resource_arm_storage_account.go
mitchellh: Good thought, I'll move the read error check down later.

mitchellh:
Addressed all feedback.

PR #8221, opened 2016-08-16
aws: Add elb_account_id data source

radeksimko:
I'm not sure if data source is the right place for such static lookup, but I can't think of a better place right now.

http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-access-logs.html#attach-bucket-policy

### DIFF
--- a/website/source/docs/providers/aws/d/elb_account_id.html.markdown
+++ b/website/source/docs/providers/aws/d/elb_account_id.html.markdown
@@ -1,0 +1,69 @@
+---
+layout: "aws"
+page_title: "AWS: aws_elb_account_id"
+sidebar_current: "docs-aws-datasource-elb-account-id"
+description: |-
+  Get AWS Elastic Load Balancing Account ID
+---
+
+# aws\_elb\_account\_id
+
+Use this data source to get the Account ID of the [AWS Elastic Load Balancing Account](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-access-logs.html#attach-bucket-policy)
+in a given region for the purpose of whitelisting in S3 bucket policy.
+
+## Example Usage
+
+```
+data "aws_elb_account_id" "main" { }
+
+resource "aws_s3_bucket" "elb_logs" {
+    bucket = "my-elb-tf-test-bucket"
+    acl = "private"
+    policy = <<POLICY
+{
+  "Id": "Policy",
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "s3:PutObject"
+      ],
+      "Effect": "Allow",
+      "Resource": "arn:aws:s3:::my-elb-tf-test-bucket/AWSLogs/*",
+      "Principal": {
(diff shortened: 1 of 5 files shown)


jen20:
I am generally :+1: on this - exposing as much information as possible via data sources to prevent hard coded values is a good idea IMO!

jen20:
LGTM pending Travis and running acctests - sorry about the merge conflict there!

jen20:
```
› envchain aws.terraform make testacc TEST=./builtin/providers/aws TESTARGS="-run TestAccAWSElbAccountId_basic"
2016/08/16 11:43:39 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run TestAccAWSElbAccountId_basic -timeout 120m
=== RUN   TestAccAWSElbAccountId_basic
--- PASS: TestAccAWSElbAccountId_basic (19.17s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	19.187s
```

apparentlymart:
I was initially kinda confused by the name of this, thinking to myself "What is an ELB account?"

After reading the linked docs I understand now that this is the account id of the ELB service itself, for each region.

So with that definition in mind, could we instead call this `aws_elb_service_account`? (The `_id` suffix is also redundant, since id is an attribute so it causes "stutter".)

So this would look like:

```hcl
data "aws_elb_service_account" "current_region" {
}
data "aws_iam_policy_document" "elb_access" {
  statement {
    principals {
        type        = "AWS"
        identifiers = "${data.aws_elb_service_account.current_region.id}"
    }
    # etc
  }
}
```

(Sorry this is a bit of a nit... just concerned that the naming as merged will be misleading and cause people to misunderstand what this is for.)

website/source/docs/providers/aws/d/elb_account_id.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+    instance_protocol = "http"
+    lb_port = 80
+    lb_protocol = "http"
+  }
+}
+```
+
+## Argument Reference
+
+* `region` - (Optional) Region of a given AWS ELB Account
```
apparentlymart: Maybe clearer to say: "Name of the region whose ELB account id is desired. If not specified, default's to the region from the AWS provider configuration."

website/source/docs/providers/aws/d/elb_account_id.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+
+
+## Attributes Reference
+
+* `id` - Account ID
```
apparentlymart: Maybe clearer to say: "The id of the ELB service account in the selected region."

radeksimko:
Thanks for all the comments, agreed to all points. I'll address that in a separate PR.

radeksimko:
^ @stack72 FYI - I know there's a decent portion of code you copied from here, so it would be good to make similar changes in https://github.com/hashicorp/terraform/pull/8224

PR #8233, opened 2016-08-16
testing: Add ComposeAggregateTestFunc

jen20:
This commit adds a function which composes a series of TestFuncs, but will run all tests before returning an error, unlike ComposeTestFunc. This is useful when verifying contents of state in acceptance tests and it is desirable to see all the failing cases in one run for slow resources.


phinze:
Solid, LGTM

PR #8235, opened 2016-08-16
Updated Plugins section

brianshumate:
- Minor update to remove some possibly confusing discussion about
  pre 0.7 separate binaries for plugins, and link to internal
  plugins section for more clarification of how plugins are
  handled in 0.7+.


catsby:
LGTM thanks!

Issue #8230, opened 2016-08-16
mysql_user resource not compatible with MySQL < 5.7.6


protomouse:
`mysql_user`, as introduced by GH-7656, [uses](https://github.com/hashicorp/terraform/blob/8623803dd71ec0b7eed2908f15a211adabb93b43/builtin/providers/mysql/resource_user.go#L71) an `ALTER USER` syntax [that was introduced](https://dev.mysql.com/doc/refman/5.7/en/alter-user.html) in MySQL 5.7.6. This leads to the following error when trying to update user passwords through Terraform against, in this case, AWS Aurora (5.6.10):

```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL 
server version for the right syntax to use near 'IDENTIFIED BY 'xxxxxxx' at line 1'
```

The documentation [states](https://dev.mysql.com/doc/refman/5.7/en/set-password.html) that the old `SET PASSWORD` syntax still works, but has been deprecated in favour of `ALTER USER` as of 5.7.6.

PR #8251, opened 2016-08-17
mysql_user pre-5.7.6 compat (fixes GH-8230)

protomouse:
As described in GH-8230, `mysql_user` isn't compatible with MySQL versions < 5.7.6. This PR adds a conditional based on the version string reported by MySQL, and a test that covers this scenario.

### DIFF
--- a/builtin/providers/mysql/provider.go
+++ b/builtin/providers/mysql/provider.go
@@ -2,13 +2,22 @@ package mysql
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


mitchellh:
Looks great. Thanks for fixing your own issue!

ronaldtse:
Hi @protomouse , thanks for the fix but just wanted to let you know that this patch broke support for AWS RDS (with TF 0.7.2).

RDS MySQL versions are like this "5.7.1-log", in which the last part "1-log" will fail to be parsed as uint.

ronaldtse:
Thanks @protomouse !

Issue #8137, opened 2016-08-11
Support AWS Application Load Balancers (ALB)


tpounds:
AWS just announced Layer 7 load balancing. According to the FAQ provisioning an Application Load Balancers requires calling new APIs so a new top-level resource will need to be created.

see:
https://aws.amazon.com/blogs/aws/new-aws-application-load-balancer/
https://aws.amazon.com/elasticloadbalancing/applicationloadbalancer/faqs/

catsby:
We'll see what we can do 😉

jen20:
I'm working on this at the moment - expecting to have it in time for 0.7.1.

kamoljan:
@jen20 where do you think you can make it?

anosulchik:
@jen20 THanks for update, looking forward to start using it will terraform.

catsby:
> @jen20 where do you think you can make it? 

It's in the works right now, though we don't promise any timelines on deliveries in general. As soon as we can!

robmorgan:
thanks guys! looking forward 👍

llarsson:
Just going to write the word "elb" here to make this issue show up in searches for that term. ;)

stekershaw:
It may be implied but just to be safely explicit: the _ecs_aws_service_ resource would ideally also support the new _Application Load Balancers_.

catsby:
@stekershaw that may not come in the initial Application Load Balancers PR, but it will follow shortly, same for integrating the new ALBs with AutoScaling Groups

morgante:
What's the status of this?

bronislav:
@morgante Terraform 0.7.1 added ALB support

radeksimko:
@bronislav correct, ALB is supported in `0.7.1`. [ECS integration](https://github.com/hashicorp/terraform/pull/8190) is coming, besides that Terraform should now support all the ALB related functionality.

Do let us know if you find any other missing features or bugs in ALB related resources. I'd close this issue once we merge that linked ECS PR.

kamoljan:
@bronislav and @radeksimko thank you guys!

kesor:
Notably missing from 0.7.1 is support for target groups in AutoScaling, this is quite essential.
https://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_AttachLoadBalancerTargetGroups.html

radeksimko:
[...]

PR #8254, opened 2016-08-17
provider/aws: Initial support for Application Load Balancers

jen20:
This pull request contains the first two resources for Application Load Balancers ("ALB" or "elbv2"), `aws_alb` and `aws_alb_target_group`.

The resources necessary will be split across a number of pull requests to facilitate review.

Part of #8137.


builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,356 @@
...
+			return nil
+		}
+
+		return errwrap.Wrapf("Error retrieving ALB: {{err}}", err)
+	}
+	if len(describeResp.LoadBalancers) != 1 {
+		return fmt.Errorf("Unable to find ALB: %#v", describeResp.LoadBalancers)
+	}
+
+	alb := describeResp.LoadBalancers[0]
```
stack72: Can we guarantee that our ALB is the first in the collection?

jen20: Yes - otherwise the error above would trigger.

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,356 @@
...
+
+		albArn = *resp.LoadBalancers[0].LoadBalancerArn
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] ALB ID: %s", d.Id())
```
stack72: You are using d.Id before setting it below

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		if isLoadBalancerNotFound(err) {
+			// The ALB is gone now, so just remove it from the state
+			d.SetId("")
+			return nil
+		}
+
+		return errwrap.Wrapf("Error retrieving ALB: {{err}}", err)
+	}
+	if len(describeResp.LoadBalancers) != 1 {
```
stack72: Can it ever be > 1?

jen20: Not if ARNs are unique.

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,346 @@
...
+
+			"internal": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+
+			"security_groups": {
+				Type:     schema.TypeSet,
```
stack72: If this cannot be modified, then this should be ForceNew: true

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,346 @@
...
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+				Computed: true,
+				Set:      schema.HashString,
+			},
+
+			"subnets": {
+				Type:     schema.TypeSet,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Required: true,
```
stack72: If this cannot be modified, then this should be ForceNew: true

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,346 @@
...
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+			},
+
+			"enable_deletion_protection": {
+				Type:     schema.TypeBool,
+				Optional: true,
```
stack72: I think we need to set a default here as d.GetOK on false means it won't get added to the reauest

builtin/providers/aws/resource_aws_alb.go
jen20: Good catch, corrected.

builtin/providers/aws/resource_aws_alb.go
jen20: Yup, good catch. Added.

builtin/providers/aws/resource_aws_alb_target_group.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+func validateAwsAlbTargetGroupHealthCheckMatcher(v interface{}, k string) (ws []string, errors []error) {
+	//TODO(jen20) implement this
+	return
+}
+
+func validateAwsAlbTargetGroupPort(v interface{}, k string) (ws []string, errors []error) {
+	port := v.(int)
+	if port < 0 || port > 65536 {
```
stack72: Should be <1 not <0

builtin/providers/aws/resource_aws_alb_target_group_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_alb_target_group.test",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckAWSALBTargetGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSALBTargetGroupConfig_basic(targetGroupName),
+				Check: resource.ComposeAggregateTestCheckFunc(
```
stack72: love the new Test func!

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,348 @@
...
+		elbOpts.SecurityGroups = expandStringList(v.(*schema.Set).List())
+	}
+
+	if v, ok := d.GetOk("subnets"); ok {
+		elbOpts.Subnets = expandStringList(v.(*schema.Set).List())
+	}
+
+	log.Printf("[DEBUG] ALB create configuration: %#v", elbOpts)
+	var albArn string
+	err := resource.Retry(1*time.Minute, func() *resource.RetryError {
```
radeksimko: Why exactly are we retrying this operation if there's no `RetryableError`?

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,348 @@
...
+
+	describeAlbOpts := &elbv2.DescribeLoadBalancersInput{
+		LoadBalancerArns: []*string{aws.String(albArn)},
+	}
+
+	describeResp, err := elbconn.DescribeLoadBalancers(describeAlbOpts)
+	if err != nil {
+		if isLoadBalancerNotFound(err) {
+			// The ALB is gone now, so just remove it from the state
+			d.SetId("")
```
radeksimko: Probably nitpick, but usually (in many other resources) we also send `[WARN]` log message to the log in such cases.

builtin/providers/aws/resource_aws_alb.go
```diff
@@ -0,0 +1,348 @@
...
+
+	attributesResp, err := elbconn.DescribeLoadBalancerAttributes(&elbv2.DescribeLoadBalancerAttributesInput{
+		LoadBalancerArn: aws.String(d.Id()),
+	})
+	if err != nil {
+		return errwrap.Wrapf("Error retrieving ALB Attributes: {{err}}", err)
+	}
+
+	accessLogMap := map[string]interface{}{}
+	for _, attr := range attributesResp.Attributes {
```
radeksimko: Wouldn't this be easier to maintain going forward if we just had a single `attributes` (`TypeMap`) field instead of trying to do this conversion?

Admittedly we'd probably still need to have some kind of key-filtering because of https://github.com/hashicorp/terraform/pull/8104

website/source/docs/providers/aws/r/alb.html.markdown
```diff
@@ -0,0 +1,70 @@
...
+
+```
+# Create a new load balancer
+resource "aws_alb" "test" {
+  name            = "test-alb-tf"
+  internal        = false
+  security_groups = ["${aws_security_group.alb_sg.id}"]
+  subnets         = ["${aws_subnet.public.*.id}"]
+
+  disable_api_protection = true
```
radeksimko: Where did this field come from? I was unable to find it anywhere in the whole diff.

radeksimko: I assume it was supposed to be `enable_deletion_protection`?

jen20: Yup, it's now `enable_deletion_protection`. Will update.

radeksimko:
@jen20 This is looking pretty good.

I tried running the attached acceptance tests and one failed:

```
TF_ACC=1 go test ./builtin/providers/aws -v -run=AccAWSALB -timeout 120m
=== RUN   TestAccAWSALBTargetGroup_basic
--- PASS: TestAccAWSALBTargetGroup_basic (34.37s)
=== RUN   TestAccAWSALBTargetGroup_updateHealthCheck
--- PASS: TestAccAWSALBTargetGroup_updateHealthCheck (53.44s)
=== RUN   TestAccAWSALB_basic
--- PASS: TestAccAWSALB_basic (46.35s)
=== RUN   TestAccAWSALB_accesslogs
--- FAIL: TestAccAWSALB_accesslogs (53.77s)
       	testing.go:265: Step 1 error: Error applying: 1 error(s) occurred:

       		* aws_s3_bucket.logs: Error creating S3 bucket: InvalidBucketName: The specified bucket is not valid.
       			status code: 400, request id: B37F0544044DE86D
FAIL
exit status 1
FAIL   	github.com/hashicorp/terraform/builtin/providers/aws   	187.957s
```

builtin/providers/aws/resource_aws_alb.go
jen20: I don't think we can easily do that and retain the same semantics? Here we construct the required structure to set on `access_log` - the other attributes work on different prefixes? We _could_ have a single `attributes` map in the schema, but I think this is more explicit and closer to expectations at the moment.

jen20:
@radeksimko Yup, fixing that up at the moment.

builtin/providers/aws/resource_aws_alb.go
radeksimko: > the other attributes work on different prefixes

yeah, it would mean making the map keys more verbose, which goes against UX, so forget what I said. This looks 👌

jen20:
```
› envchain aws.terraform make testacc TEST=./builtin/providers/aws TESTARGS="-run TestAccAWSALB"
2016/08/17 15:54:22 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run TestAccAWSALB -timeout 120m
=== RUN   TestAccAWSALBTargetGroup_basic
--- PASS: TestAccAWSALBTargetGroup_basic (36.57s)
=== RUN   TestAccAWSALBTargetGroup_updateHealthCheck
--- PASS: TestAccAWSALBTargetGroup_updateHealthCheck (59.23s)
=== RUN   TestAccAWSALB_basic
--- PASS: TestAccAWSALB_basic (51.25s)
=== RUN   TestAccAWSALB_accesslogs
--- PASS: TestAccAWSALB_accesslogs (93.51s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	240.584s
```

PR #5076, opened 2016-02-09
[WIP] add allowed_address_pairs

roobert:
Opening this PR (as requested by @jtopjian) as a work in progress to support Allowed Address Pairs[1] now that gophercloud has support has been merged[2].

[1] http://specs.openstack.org/openstack/neutron-specs/specs/api/allowed_address_pairs.html
[2] https://github.com/rackspace/gophercloud/pull/488#issuecomment-182002690

PR #8257, opened 2016-08-17
provider/openstack: Add allowed_address_pairs

FedericoCeratto:
Based on  #5076

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_port_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_port_v2.go
@@ -96,6 +96,25 @@ func resourceNetworkingPortV2() *schema.Resource {
 					},
 				},
 			},
+			"allowed_address_pairs": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: false,
+				Computed: true,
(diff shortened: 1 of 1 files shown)


jtopjian:
@FedericoCeratto Thanks for this. Let me know when this is ready for a review. :)

FedericoCeratto:
@jtopjian the PR is essentially ready as it is, unless acceptance tests are required (I'm having some trouble running the tests)

jtopjian:
It looks like you added a test for the `import` functionality, and I don't think that specific test is adding anything more to what was already being tested.

Acceptance tests are not required - I can always merge this and add one post-merge.

If this is ready to go, I'd say remove the import test and then squash everything into one commit.

FedericoCeratto:
@jtopjian the test was already in place, I populated the testAccNetworkingV2Port_basic variable.

jtopjian:
That's already done [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_port_v2_test.go#L112) so it's not required.

jtopjian:
@FedericoCeratto sorry for the delay. Thank you very much for this!

PR #8267, opened 2016-08-17
Document credentials variable and env var for gcs remote state

ChrisMcKenzie:
Just added a line documenting the environment variable for credentials as it wasn't immediately clear how this is set.


mitchellh:
LGTM. Thanks

PR #8268, opened 2016-08-17
provider/aws: Add aws_alb_listener resource

jen20:
This commit adds the `aws_alb_listener` resource and associated acceptance tests and documentation.

This is part of a series of pull requests in order to implement #8137.

### DIFF
--- a/builtin/providers/aws/resource_aws_alb_listener.go
+++ b/builtin/providers/aws/resource_aws_alb_listener.go
@@ -1,0 +1,261 @@
+package aws
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsAlbListener() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsAlbListenerCreate,
+		Read:   resourceAwsAlbListenerRead,
+		Update: resourceAwsAlbListenerUpdate,
+		Delete: resourceAwsAlbListenerDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"load_balancer_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"port": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validateAwsAlbListenerPort,
+			},
+
+			"protocol": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "HTTP",
+				StateFunc: func(v interface{}) string {
+					return strings.ToUpper(v.(string))
+				},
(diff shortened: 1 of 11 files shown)


website/source/docs/providers/aws/r/alb_listener.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+resource "aws_alb" "front_end" {
+  # Other parameters...
+}
+
+resource "aws_alb_target_group" "front_end" {
+  # Other parameters...
+}
+
+resource "aws_alb_listener" "front_end" {
+   load_balancer_arn = "${aws_alb.front_end.id}"
```
catsby: I realize that in our case, the ID and the ARN of the ALB are the same, but what do you think about exposing the ARN on the ALB resource, and then referencing that here? 

I think that, while it's kind of a minor detail for us, it makes more sense to a user use a config like this:

load_balancer_**arn** = "${aws_alb.front_end.**arn**}"

over this

load_balancer_**arn** = "${aws_alb.front_end.**id**}"

catsby: The same would be true for `TargetGroup` below

builtin/providers/aws/resource_aws_alb_listener.go
```diff
@@ -0,0 +1,255 @@
...
+			"default_action": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"target_group_arn": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"type": {
```
catsby: nitpick: if this only has 1 acceptable value, why is it required instead of a `Computed` or `Optional` with `Default: blah` ?

builtin/providers/aws/resource_aws_alb_listener.go
```diff
@@ -0,0 +1,255 @@
...
+		if isListenerNotFound(err) {
+			log.Printf("[WARN] DescribeListeners - removing %s from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return errwrap.Wrapf("Error retrieving Listener: {{err}}", err)
+	}
+
+	if len(resp.Listeners) != 1 {
+		return fmt.Errorf("Error retrieving Listener %q", d.Id())
```
catsby: I've noticed that when we hit an error, sometimes we return `errwrap.Wrapf`, one time it's `errors.New`, and another it's `fmt.Errorf`. It's hard for me to see if this discrepancy is intentional

catsby:
I think for a UX perspective we should add and export the ARNs of this and `aws_alb`. Other than that and minor nits, this looks great!

builtin/providers/aws/resource_aws_alb_listener.go
jen20: We have no idea what they will add in future, and probably shouldn't back ourselves into a corner of having a weird default or having to break backwards compatibility - I've basically mirrored the AWS API defaults here, which seems like a reasonable principle.

builtin/providers/aws/resource_aws_alb_listener.go
jen20: This is applying the pattern discussed with @phinze - if we are adding context to an error returned by someone else we wrap, but if constructing our own error we use fmt.Errorf (or errors.New if there are no format placeholders). This is a standard worth rolling out across the entire code base, but will take a very long time, hence introducing it with new resources and modifying existing ones over time.

website/source/docs/providers/aws/r/alb_listener.html.markdown
jen20: This seems fairly reasonable, I'll add it shortly.

PR #8273, opened 2016-08-17
Update dns_zone.html.markdown

ryanmoran:
Include documentation of `name_servers` output.


stack72:
Hi @ryanmoran

thanks so much for the PR here - this is a great addition!

Paul

Issue #4952, opened 2016-02-02
Amazon Simple Systems Manager (SSM) 


aminek01:
Hello,
Is there a way to create SSM document and join instance to directory service domain:

http://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2-join-aws-domain.html

skippy:
updated doc location: https://docs.aws.amazon.com/ssm/latest/APIReference/Welcome.html

liamjbennett:
I stumbled into this issue for the same use-case mentioned above. I should have a PR to support this shortly.

aminek01:
Thank you so much.

liamjbennett:
There is one more resource `ssm_activation` still to be added before this issue can be closed.

liamjbennett:
http://docs.aws.amazon.com/ssm/latest/APIReference/API_CreateActivation.html

stack72:
All 3 resources now added - thanks @liamjbennett :)

PR #8282, opened 2016-08-18
provider/aws: add aws_ssm_document resource

liamjbennett:
First PR in resolving #4952


builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsSsmDocument() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSsmDocumentCreate,
+		Read:   resourceAwsSsmDocumentRead,
+		Update: resourceAwsSsmDocumentUpdate,
```
stack72: If both are ForceNew then you can omit this Update method completely

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,107 @@
...
+	resp, err := ssmconn.DescribeDocument(params)
+
+	if err != nil {
+		return fmt.Errorf("Error describing SSM document: %s", err)
+	}
+
+	doc := resp.Document
+	d.Set("created_date", doc.CreatedDate)
+	d.Set("description", doc.Description)
+	d.Set("hash", doc.Hash)
```
stack72: These all need to be added to the schema as Computed: true :)

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,107 @@
...
+func resourceAwsSsmDocumentCreate(d *schema.ResourceData, meta interface{}) error {
+	ssmconn := meta.(*AWSClient).ssmconn
+
+	name := d.Get("name").(string)
+	content := d.Get("content").(string)
+
+	log.Printf("[DEBUG] SSM document create: %s", name)
+
+	params := &ssm.CreateDocumentInput{
+		Name:    aws.String(name),
```
stack72: you know you can just reference the d.Get("name").(string) in the aws.String( ) right?

liamjbennett:
TESTS:
```
2016/08/19 09:44:21 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSSSMDocument_basic -timeout 120m
=== RUN   TestAccAWSSSMDocument_basic
--- PASS: TestAccAWSSSMDocument_basic (23.50s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    23.513s
```

liamjbennett:
```
➜  terraform git:(aws_ssm_document) ✗ TF_LOG=TRACE TF_LOG_PATH=test_log.txt make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSSSMDocument_basic' 
/home/liam/development/go/bin/stringer
2016/08/22 13:48:17 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSSSMDocument_basic -timeout 120m
=== RUN   TestAccAWSSSMDocument_basic
--- PASS: TestAccAWSSSMDocument_basic (26.36s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    26.374s
```

liamjbennett:
@stack72

jen20:
Hi @liamjbennett! I'm doing a second review of this now - thanks for making the changes! I have a few more areas of feedback but they are mostly minor and I'll fix them up directly before merging, after leaving inline comments.

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,309 @@
...
+
+func resourceAwsSsmDocument() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSsmDocumentCreate,
+		Read:   resourceAwsSsmDocumentRead,
+		Update: resourceAwsSsmDocumentUpdate,
+		Delete: resourceAwsSsmDocumentDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
```
jen20: We don't need the `&schema.Schema` here - though that is currently being spread across the code base. For new resources it's preferable to use the Go 1.5+ style of not declaring the struct names in these lists as it results in cleaner syntax.

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,309 @@
...
+
+	docInput := &ssm.CreateDocumentInput{
+		Name:    aws.String(d.Get("name").(string)),
+		Content: aws.String(d.Get("content").(string)),
+	}
+
+	resp, err := ssmconn.CreateDocument(docInput)
+
+	if err != nil {
+		return fmt.Errorf("[ERROR] Error creating SSM document: %s", err)
```
jen20: Wherever we return an error which originates from upstream, the current best practice is to wrap it using the `errwrap` library. So in this case, this line would read:

```go
return errwrap.Wrapf("[ERROR] Error creating SSM document: {{err}}", err)
```

This preserves any context attached to the original error, as well as allowing tests for particular types of error having occurred in a given chain.

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	d.SetId(*resp.DocumentDescription.Name)
+
+	if v, ok := d.GetOk("permissions"); ok && v != nil {
+		setDocumentPermissions(d, meta)
+	} else {
+		log.Printf("[DEBUG] not setting document permissions")
```
jen20: These may be interleaved in the log, so it's best to mention the ID of the resource in question where possible.

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	return resourceAwsSsmDocumentRead(d, meta)
+}
+
+func resourceAwsSsmDocumentRead(d *schema.ResourceData, meta interface{}) error {
+	ssmconn := meta.(*AWSClient).ssmconn
+
+	log.Printf("[DEBUG] Reading SSM Document: %s", d.Get("name").(string))
```
jen20: Is this the same as the ID? Is there a reason to `d.Get("name")` to read rather than using `d.Id()`?

builtin/providers/aws/resource_aws_ssm_document.go
jen20: Looks like this one has been completed.

builtin/providers/aws/resource_aws_ssm_document.go
liamjbennett: I guess not. `name` just happens to be only unique element mentioned in the docs:
https://docs.aws.amazon.com/ssm/latest/APIReference/API_CreateDocument.html

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,309 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func setDocumentPermissions(d *schema.ResourceData, meta interface{}) error {
+	ssmconn := meta.(*AWSClient).ssmconn
+
+	log.Printf("[INFO] Setting permissions for document: %s", d.Get("name").(string))
```
jen20: I think we can use `d.Id()` here too.

builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -0,0 +1,309 @@
...
+		return fmt.Errorf("[ERROR] Error setting permissions for SSM document: %s", err)
+	}
+
+	return nil
+}
+
+func getDocumentPermissions(d *schema.ResourceData, meta interface{}) (map[string]interface{}, error) {
+	ssmconn := meta.(*AWSClient).ssmconn
+
+	log.Printf("[INFO] Getting permissions for document: %s", d.Get("name").(string))
```
jen20: Same as above with `d.Id()`

builtin/providers/aws/resource_aws_ssm_document.go
jen20: Looks like `permission` can be updated, so we probably do want the update function here.

jen20:
Hi @liamjbennett! I left some comments inline - however I'll fix these up and merge this. Thanks for contributing to Terraform! I notice there is a second resource PR from you for a follow up attachment resource to this one - it would be great if you could take a quick look through the feedback here and apply the same things?

liamjbennett:
@jen20: thanks for the comments and the quick review. Yes I will update my second PR based upon the comments made here.

liamjbennett:
@jen20 is there anything else you need me to do here?

stack72:
@liamjbennett - i opened a PR to add this resource and the changes that @jen20 asked for :)

#8460

thanks for the work here!

Paul

PR #8285, opened 2016-08-18
provider/aws: Bump the AWS SDK version to 1.4.2

stack72:
I'd like to leave this until after 0.7.1 is released so we can get a few sets of nightly acceptance tests out of it!


phinze:
LGTM now that we're released

PR #8286, opened 2016-08-18
provider/aws: Explain better why we retry IAM role creation

radeksimko:
As discussed in https://github.com/hashicorp/terraform/pull/7733#issuecomment-240759666

cc @dtolnay


stack72:
LGTM! Thanks @radeksimko :)

Issue #4222, opened 2015-12-08
Google CloudSQL authorized networks not populated


coreone:
I have been using the new Google CloudSQL resource in Terraform, which all seems to work great, except for the `authorized_networks` configuration.  When I fill in this section with one or more networks, it blanks out all the authorized networks on the Google side, even though the **plan** and **apply** steps both seem to suggest that the correct changes are being made.  I started this discussion at:

https://groups.google.com/d/msgid/terraform-tool/88db0e72-8b10-493e-9fa9-fb8ec66c3682%40googlegroups.com

My configuration is as follows (identifying information removed):
```
resource "google_sql_database_instance" "mysql" {
    name = "mysql${format("%02d", count.index+1)}"
    count = "${var.num_cloudsql_instances}"
    region = "${var.cloudsql_region}"
    database_version = "${var.cloudsql_version}"

    settings {
        activation_policy = "${var.cloudsql_activation_policy}"
        pricing_plan = "${var.cloudsql_pricing_plan}"
        replication_type = "${var.cloudsql_replication_type}"
        tier = "D16"

        backup_configuration {
            binary_log_enabled = false
            enabled = true
            start_time = "06:00"
        }

        ip_configuration {
            ipv4_enabled = true
            # require_ssl = true
            authorized_networks = {
                name = "The Network"
                value = "x.x.x.x/x"
            }
            authorized_networks = {
                name = "${google_compute_instance.computename.0.name}"
                value = "${google_compute_address.computename.0.address}/32"
            }
        }
    }
}
```
When I run the plan, I get:
```
~ google_sql_database_instance.mysql
    settings.0.ip_configuration.0.authorized_networks.#:       "0" => "2"
    settings.0.ip_configuration.0.authorized_networks.0.name:  "" => "The Network"
    settings.0.ip_configuration.0.authorized_networks.0.value: "" => "x.x.x.x/x"
    settings.0.ip_configuration.0.authorized_networks.1.name:  "" => "computename"
    settings.0.ip_configuration.0.authorized_networks.1.value: "" => "x.x.x.x/32"
```

And then apply:
```
google_sql_database_instance.mysql: Modifying...
  settings.0.ip_configuration.0.authorized_networks.#:       "0" => "2"
  settings.0.ip_configuration.0.authorized_networks.0.name:  "" => "The Network"
  settings.0.ip_configuration.0.authorized_networks.0.value: "" => "x.x.x.x/x"
[...]

PR #8290, opened 2016-08-18
provider/google: Correct update process for auth nets in sql instance

lwander:
This addresses #4222. It was a combination of three errors:

1. Networks were being read out of the wrong field.
2. Networks populated outside of terraform were being ignored.
3. Networks added during an update were being ignored.

@phinze


phinze:
LGTM, thanks @lwander!

PR #8278, opened 2016-08-18
provider/archive support folders in output_path

nicolai86:
The `archive` provider right now relies on `FileInfo`s `Name()` function, which only returns the base name of the`output_path`, ignoring any folders:

```
type FileInfo interface {
        Name() string       // base name of the file
        Size() int64        // length in bytes for regular files; system-dependent for others
        Mode() FileMode     // file mode bits
        ModTime() time.Time // modification time
        IsDir() bool        // abbreviation for Mode().IsDir()
        Sys() interface{}   // underlying data source (can return nil)
}
```

this PR changes to use the `output_path` directly, allowing archives to be placed in subfolders:

before:

```
resource "archive_file" "foo" {
  type                    = "zip"
  source_content          = "This is some content"
  source_content_filename = "content.txt"
  output_path             = "example/path/test.zip"
}
```

throws error that `test.zip` could not be found

after: creates test.zip in `example/path`.

PR #8291, opened 2016-08-18
provider/archiver fix test output path breaking other tests

nicolai86:
This PR addresses a bug which I introduced in #8278.

ping @stack72


stack72:
Thanks @nicolai86 

This works :)

nicolai86:
Yay! Thanks for the polite heads up,@stack72

PR #8292, opened 2016-08-18
provider/google: Changed instances to a computed property of 'google_compute_target_pool'

nwwebb:
Instances of a target pool must be computed in the case when an instance group manager is adding instances. This change makes the list of instances a computed property so that the instance group manager tests pass. @lwander


lwander:
Nice fix, LGTM

PR #8293, opened 2016-08-18
docs/remote_state: Update docs to use data instead of resource

protochron:
I noticed that a lot of the new `remote_state` backends were still using a resource instead of a data source.


radeksimko:
👍 Thanks for the catch!

PR #8299, opened 2016-08-18
provider/dns: a provider for DNS dynamic updates (RFC 2136) with optional support for TSIG (RFC 2845)

drebes:
This is an initial attempt at supporting standard DNS dynamic updates using transaction signatures and shared secrets.

It's particularly interesting for enterprises that do not have any form of DNS as a service.

Acceptance tests can be run based on the preconfigured bind docker image available [here](https://hub.docker.com/r/drebes/bind/). It can be launched with:

```
docker run -p 53:53/udp drebes/bind
```

and the acceptance test needs the following environment variables:

```
export DNS_KEY_ALGORITHM="hmac-md5"
export DNS_KEY_NAME="example.com."
export DNS_KEY_SECRET="c3VwZXJzZWNyZXQ="
export DNS_SERVER="127.0.0.1"
export DNS_ZONE="example.com."
export DNS_UPDATE_SERVER=127.0.0.1
$ make testacc TEST=./builtin/providers/dns 
/opt/local/bin/stringer
2016/08/18 21:27:27 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/dns -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccDnsRecord_basic
--- PASS: TestAccDnsRecord_basic (0.09s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/dns    0.111s
```

This is just an initial version for feedback. It has only a single resource type (dns_record) which can be used to create A, AAAA, CNAME, MX, NS, PTR and SRV records. But it may make sense to split these all into different resource types, to make it easier to check for their existence.

Also, a data provider could be included for existing records.

Depending on feedback I will adapt and plan to include documentation as well.


radeksimko:
Hi @drebes 
thanks for the PR.

I'm just looping in @BSick7 as https://github.com/hashicorp/terraform/pull/7270 may conflict with this PR in terms of provider name, even though both PRs are kind of solving slightly different problem.

Either we'll need to find a way for both PRs to coexist under the same provider in which case the provider configuration would have to be protocol-agnostic - i.e. it would be impossible to mark any TSIG specific options `Required` because user may only want to use data sources to query DNS records.

The other option would be to have 2 providers, e.g. `dns_tsig` and `dns` or something like that.

I personally don't have a strong opinion about this - as long as we have reasons documented/discussed to either keep both under the same provider or to separate them out.

drebes:
Thanks @radeksimko. I was not aware of #7270 but it goes in line with the ideas I considered for this provider as well. I'll check there and see if we can collaborate on a single provider. I'm not against making the credentials optional: it makes sense if we assume the data providers which are read only. And I think in theory one could even send updates without signatures in some isolated environments.

BSick7:
I would definitely break out dns record types into separate terraform resources.

In terms of merging the providers, I'm a little torn.

Ideally, dns datasources have no dependence on a third party library.
This makes problem discovery much easier for terraform consumers.
As of now, nearly every datasource has "deprecated" resource exposure.
It may be odd to some that `resource "dns_a"` is totally different from `data "dns"a"`.

That being said, it could be very confusing to users that there are 2 different providers to manage dns records.

This note from @radeksimko strikes a chord with me and makes me lean toward having 2 providers: `dns` and `dns_tsig`.
> ...it would be impossible to mark any TSIG specific options Required because user may only want to use data sources to query DNS records.

drebes:
Here's how I plan to proceed:

1. I'll remove the requirement of authentication, basically making the provider compliant with RFC 2136 by default and 2845 only when the user user provides the secret parameters.

2. I'll move the zone parameter into each DNS record. That way, the provider will not require any parameter by default and be compatible with #7270 if we decide to merge the two providers.

3. I'll break the DNS record type into different records based on the record type. I think that make sense.

At that point we can discuss if we merge the two providers to have the data providers, or rename this one. Personally I'd prefer to merge them.

apparentlymart:
@drebes your plan sounds good to me, and I also agree that merging the two makes sense.

Do you think it would make sense to model the resources as record *sets* rather than individual records? e.g.:

```hcl
resource "dns_a_record_set" "example" {
  zone = "example.com"
  name = "foo"
  records = ["127.0.0.1", "127.0.0.2"]
}
```

I can see a few advantages to this formulation:
* Generally when we query DNS we look for all of the records with a given name/type, rather than a specific one. This creates consistency between how we read and how we create.
* Currently your `Read` implementation fetches all of the records with the given name/type and then simply checks to see if the given value is among them. With this approach, Terraform would manage the entire recordset and we would avoid surprises where some matching record has been created outside of Terraform and some percentage of requests end up choosing this rogue record from the set.
* This is consistent with the level of abstraction on the existing `aws_route53_record` and `google_dns_record_set` (but not on `dyn_record`, `dme_record`, and `dnsimple_record`, I concede).

drebes:
I've updated the PR to remove any dependency on TSIG configuration for the provider, and also make the whole RR set be managed as a single resource. I broke out the A type in its own resource, and plan to add other resource types later (at least AAAA, PTR and CNAMEs which are quite straightforward).

drebes:
Yes I do. :) The docker image in the first comment makes it quite straightforward. I'll put that in a script,  running the regression tests with and without authentication, and update the PR. But I need to update the tests for the data sources as well. Will do that.

apparentlymart:
Oh my.. somehow I totally missed that paragraph in your comment. I guess I must've just dove right in to the code. Sorry about that! I will give this a try when I'm next hacking on Terraform.

drebes:
I added a script that can be used to run the acceptance tests, both without and with authentication. For now I removed the data sources as that was incomplete. As it is it supports A, AAAA, PTR and CNAME records.

catsby:
Hello everyone –

Thank you all for your work here! I'd like to move forward on pulling this and #7270 into Terraform, going forward with the plan discussed above to unify them into 1 `dns` provider. If for some reason(s) that is no longer a good idea or you have reservations, please let me know!

I have a question/request on the configuration of the provider in this PR; there is an attribute `server` that is required. I don't believe this attribute makes sense in the context of the data sources in #7270, would you agree? Please let me know if this is something that can also be made optional. 

Right now that's my only question! Once that's sorted I'll go ahead with the unification work. Thanks again!

drebes:
The server attribute for the resource is due to it using its own library for talking to the DNS servers, not the OS native resolver. The advantage to that is that the behavior is consistent accross different OSes, and DNS caching can be ignored. 

Getting rid of the server parameter would be tricky: we could take the IP address from /etc/resolv.conf when that is available, but that wouldn't work on Windows.

drebes:
We could, OTOH, rename it to "update_server" and make it required only when using resources, and have the data sources use the native OS resolver as #7270 does.

catsby:
> make it required only when using resources, and have the data sources use the native OS resolver as #7270 does.

This is what I was meaning to suggest, I apologize for not being clear.

I want to make `server` optional in the DNS _provider_ config. If it's required, then users have to supply _something_ in the provider configuration, just to use the data sources, which I didn't feel made much sense.

It makes complete sense for anyone using the resource. I'm hoping we can make it optional just like `port` or `key_name`, et. al

radeksimko:
Maybe we'd benefit from having something like `tsig` or `update` or whatevs config block (`TypeList`; `MaxItems: 1`) which would allow us to clearly separate out the configuration?

catsby:
Hey @drebes – I'd like to merge this today, is there anything remaining? I ask because it's still labeled as `WIP`. Thanks!

drebes:
It's missing docs for resource_dns_aaaa_record_set and resource_dns_cname_record. I'll add it now.

drebes:
Should now be good to go.

catsby:
Merged, thank you!

BSick7:
@catsby This was never added to the changelog.
It appears like it should have gone in 0.8.8.

bpoland:
For anyone wondering, I tried it and it does seem to be in 0.8.8.  As mentioned, it's not in the changelog however, and the included documentation was also not published to the terraform website.

drebes:
I think @catsby's [idea](https://github.com/hashicorp/terraform/pull/8299#issuecomment-274884331) was to merge this and #7270 and combine the two into a single provider.

catsby:
Correct, that's the idea, I've just fallen behind on it. v0.9 is about to come out, and I'll sort all this out. Sorry!

BSick7:
That was my understanding @drebes; however, we started getting warnings that we were overriding the built-in provider `dns` and it was quite a surprise.
@catsby I can merge the data sources in #7270 if you haven't started.

catsby:
@BSick7 I have not started, sorry! Feel free to take a stab at it. I think merging the provider merge should be straight forward, I believe the data sources should be as well.

PR #8301, opened 2016-08-18
provider/google: Ignore source image for 'google_compute_instance_template'

nwwebb:
Changed source_image so that it is read from the schema if it exists and otherwise it is read from the google api. The source image can be supplied by family rather than a specific image, while the api always returns the specific image. This caused terraform to plan a new resource because the image names differ, even though they refer to the same image. Ignoring source_image entirely would cause import to fail, so I only ignore it when it has a value in the schema (which will never be true if we are importing the resource).


nwwebb:
@lwander PTAL

lwander:
LGTM

PR #8303, opened 2016-08-18
provider/azurerm: Setting the AzureRM SDK to an explicit version

stack72:
We tended to use master which isn't the best idea. Also set Go Autorest
to the correct version (7.1.0)

This will happen after the release of Terraform 0.7.1 so we can get some acceptance test coverage


mitchellh:
LGTM when you're ready

PR #8304, opened 2016-08-18
Fix `state mv` with nested modules to work properly

### LINKED ISSUE #7797: State mv command is overly aggressive on name matching.
Terraform is overly aggressive on name matching when doing `state mv` commands resulting in resources being lost in my state. If I have `aws_vpc.main[0]` and `aws_vpc.main[1]` in my state, and `state mv aws_vpc.main` without the brackets my first vpc is moved and the rest are removed from my state.

### Terraform Version
```
$ ~/Downloads/terraform-0.7-rc3/terraform version
Terraform v0.7.0-rc3 (3f4857a07a24f3c9e2db6b4458fbf5be19a8b256)
```

### Affected Resource(s)
All resources I believe.

```hcl
resource "aws_vpc" "main" {
  cidr_block = "10.0.0.0/16"
  tags {
    Name = "import-${count.index}"
  }
  count = 2
}
```

https://gist.github.com/clstokes/d1b7cc64a310dd7d2e29d7176e613bf3#file-main-tf

### Debug Output
https://gist.github.com/clstokes/d1b7cc64a310dd7d2e29d7176e613bf3#file-log-log

### Expected Behavior
Terraform show have exited with an error that the resource was not found or that more than one resource was found.

### Actual Behavior
Terraform moved the first resource it found and removed the others.

### Steps to Reproduce
1. `terraform apply`
1. `terraform state list` <- See 2 VPCs
1. `terraform state mv aws_vpc.main aws_vpc.renamed` <- See _only_ 1 VPC remains

### LINKED ISSUE #8263: State files created with `tf state mv...` are invalid?
Up until now we have had a single state file holding 3 environments. This caused a bunch of problems in the past and we planned on separating them to different files for a while. With the release of TF 0.7, the `terraform state` command has been introduced so I thought I'll give it shot. Our old state file was `terraform.tfstate`, I issued the following commands:

`tf state mv -state-out=staging.tfstate module.staging module.staging`
`tf state mv -state-out=production.tfstate module.production module.production`
`tf state mv -state-out=dev.tfstate module.dev module.dev`

I was surprised by how simple it was, the state files themselves looked proper (to my untrained eye). The problem is that now terraform is trying to recreate all of our resources, as if it's not taking the state file into consideration, eg:

`tf plan -state=staging.tfstate -target=module.staging.module.foo`

will plan to create all of the resources defined in the submodule `foo`. The same happens for all the environments, when trying to plan and apply a submodule. I tested again with the old statefile just to make sure and I got the expected output (nothing to be changed).
I attempted refreshing the state, but that didn't change anything, more than that, it added state from other environments to the state file I was trying to refresh.

### Terraform Version
0.7

If any more information is required, I'll be glad to supply it.

mitchellh:
Fixes #8263 
Fixes #7797

We were only adding the first result of searching for what to move. This works for every result type except modules since nested modules aren't nested within the ModuleState but alongside each other with longer paths. Rather than do a BIG refactor on nesting module states (which probably isn't right anyways), I modified `State.Add` and `StateMvCommand` to understand this limitation.

### DIFF
--- a/command/state_mv.go
+++ b/command/state_mv.go
@@ -141,7 +141,8 @@ func (c *StateMvCommand) addableResult(results []*terraform.StateFilterResult) i
 		return result
 
 	case *terraform.ResourceState:
-		// If a module state then we should add the full list of modules
+		// If a resource state with more than one result, it has a multi-count
+		// and we need to add all of them.
 		result := []*terraform.ResourceState{v}
 		if len(results) > 1 {
 			for _, r := range results[1:] {
--- a/command/state_mv.go
+++ b/command/state_mv.go
@@ -156,6 +156,11 @@ func (c *StateMvCommand) addableResult(results []*terraform.StateFilterResult) i
 			}
 		}
 
+		// If we only have one item, add it directly
+		if len(result) == 1 {
+			return result[0]
+		}
+
 		return result
 
 	default:
--- a/command/state_mv.go
+++ b/command/state_mv.go
@@ -79,13 +79,16 @@ func (c *StateMvCommand) Run(args []string) int {
 		return 1
 	}
 
+	// Get the item to add to the state
+	add := c.addableResult(results)
+
 	// Do the actual move
 	if err := stateFromReal.Remove(args[0]); err != nil {
 		c.Ui.Error(fmt.Sprintf(errStateMv, err))
 		return 1
 	}
 
-	if err := stateToReal.Add(args[0], args[1], results[0].Value); err != nil {
+	if err := stateToReal.Add(args[0], args[1], add); err != nil {
 		c.Ui.Error(fmt.Sprintf(errStateMv, err))
 		return 1
 	}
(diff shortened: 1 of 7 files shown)


mitchellh:
Updated PR, also fixes #7797 now.

command/state_mv.go
```diff
@@ -119,6 +122,53 @@ func (c *StateMvCommand) Run(args []string) int {
...
+				if ms, ok := r.Value.(*terraform.ModuleState); ok {
+					result = append(result, ms)
+				}
+			}
+		}
+
+		return result
+
+	case *terraform.ResourceState:
+		// If a module state then we should add the full list of modules
```
phinze: Bad comment here I think.

terraform/state_add.go
```diff
@@ -111,6 +151,36 @@ func stateAddFunc_Resource_Module(
...
+			addrCopy := addr.Copy()
+			addrCopy.Index = i
+
+			if err := s.Add(fromAddr.String(), addrCopy.String(), rs); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}
```
phinze: This chunk is a bit confusing to me - why is the above "handle a list of resources" code so different than the below "handle one resource" code?

mitchellh: Its recursive! We're just going through each list and recursing to add just _one_ resource. :)

phinze:
LGTM generally! Two nits inline.

terraform/state_add.go
phinze: Ahhhh I was too deep in the `_Resource_Resource` helper to get that. Makes sense!

PR #8305, opened 2016-08-18
docs: Remove output from `terraform_remote_state`

### LINKED ISSUE #8296: Remote state variables present in root and not under output
So according to the documentation in here https://www.terraform.io/docs/providers/terraform/d/remote_state.html#output the variables in the remote state should be accessed under the `output` key but in reality they are present at the root level.

I was getting error about unknown attribute but running `terraform show` luckily showed me the real structure of the state and then I saw the issue.

jen20:
Fixes #8296.


mitchellh:
LGTM

PR #8307, opened 2016-08-19
provider/google: Update Acceptance Tests to use Debian 8

cblecker:
Debian 7 is no longer an official supported image in GCE, so updating Acceptance Tests and Documentation to use the most recent Debian 8 image.

While doing this, discovered a bug in the Autoscaler Test that was causing it to fail. Details [here](https://github.com/hashicorp/terraform/pull/8307#commitcomment-18698853).


mitchellh:
Looks good. Thanks!

PR #8311, opened 2016-08-19
aws/docs: Add example of aws_alb_target_group

radeksimko:
cc @jen20


stack72:
LGTM thanks @radeksimko

PR #8313, opened 2016-08-19
docs/aws: Fix example of aws_redshift_service_account

radeksimko:


stack72:
Nice catch @radeksimko

PR #8318, opened 2016-08-19
provider/aws: Add aws_alb_listener_rule resource

jen20:
This commit adds the aws_alb_listener_rule resource along with acceptance tests and documentation.

Part of #8137.

### DIFF
--- a/builtin/providers/aws/resource_aws_alb_listener_rule.go
+++ b/builtin/providers/aws/resource_aws_alb_listener_rule.go
@@ -1,0 +1,252 @@
+package aws
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsAlbListenerRule() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsAlbListenerRuleCreate,
+		Read:   resourceAwsAlbListenerRuleRead,
+		Update: resourceAwsAlbListenerRuleUpdate,
+		Delete: resourceAwsAlbListenerRuleDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"listener_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"priority": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validateAwsAlbListenerRulePriority,
+			},
+			"action": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"target_group_arn": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateAwsAlbListenerActionType,
+						},
+					},
+				},
+			},
+			"condition": {
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"field": {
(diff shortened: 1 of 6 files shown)


builtin/providers/aws/resource_aws_alb_listener_rule.go
```diff
@@ -0,0 +1,252 @@
...
+		return errwrap.Wrapf("Error creating ALB Listener Rule: {{err}}", err)
+	}
+
+	if len(resp.Rules) == 0 {
+		return errors.New("Error creating ALB Listener Rule: no rules returned in response")
+	}
+
+	d.SetId(*resp.Rules[0].RuleArn)
+
+	return nil
```
stack72: should call `resourceAwsAlbListenerRuleRead` func right?

builtin/providers/aws/resource_aws_alb_listener_rule.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	actions := make([]interface{}, len(rule.Actions))
+	for i, action := range rule.Actions {
+		actionMap := make(map[string]interface{})
+		actionMap["target_group_arn"] = *action.TargetGroupArn
+		actionMap["type"] = *action.Type
+		actions[i] = actionMap
+	}
+	d.Set("action", actions)
```
stack72: do we want to error check that the Set doesn't cause any issues here?

builtin/providers/aws/resource_aws_alb_listener_rule.go
```diff
@@ -0,0 +1,252 @@
...
+		conditionMap := make(map[string]interface{})
+		conditionMap["field"] = *condition.Field
+		conditionValues := make([]string, len(condition.Values))
+		for k, value := range condition.Values {
+			conditionValues[k] = *value
+		}
+		conditionMap["values"] = conditionValues
+		conditions[i] = conditionMap
+	}
+	d.Set("condition", conditions)
```
stack72: do we want to error check that the Set doesn't cause any issues here?

builtin/providers/aws/resource_aws_alb_listener_rule.go
```diff
@@ -0,0 +1,252 @@
...
+	resp, err := elbconn.ModifyRule(params)
+	if err != nil {
+		return errwrap.Wrapf("Error modifying ALB Listener Rule: {{err}}", err)
+	}
+
+	if len(resp.Rules) == 0 {
+		return errors.New("Error modifying creating ALB Listener Rule: no rules returned in response")
+	}
+
+	return resourceAwsAlbListenerRuleRead(d, meta)
```
stack72: the Read is calling the read - this should be `return nil`

jen20: This is in the create?

builtin/providers/aws/resource_aws_alb_listener_rule.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsAlbListenerRule() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsAlbListenerRuleCreate,
+		Read:   resourceAwsAlbListenerRuleRead,
+		Update: resourceAwsAlbListenerRuleUpdate,
+		Delete: resourceAwsAlbListenerRuleDelete,
+		Importer: &schema.ResourceImporter{
```
stack72: I think it'd be good to follow up with an Import test :)

stack72:
Tests pass :)

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSALBListenerRule_'
2016/08/19 13:12:46 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSALBListenerRule_ -timeout 120m
=== RUN   TestAccAWSALBListenerRule_basic
--- PASS: TestAccAWSALBListenerRule_basic (79.20s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	79.219s
```

stack72:
This LGTM! We can follow up with the nitpicks post merge :)

builtin/providers/aws/resource_aws_alb_listener_rule.go
jen20: This is somewhat following convention - I'm not too worried about it here. In general though I'd agree we should formalise more of the policy about where we check errors on `d.Set()`

catsby: I've formalized around checking the errors if `d.Set` is setting anything other than a scalar value

PR #8322, opened 2016-08-19
health_check's target explained

anapsix:
explaining the pattern for health_check's target


jen20:
Hi @anapsix! This makes things a lot clearer, thanks for improving the docs!

Issue #6093, opened 2016-04-08
Additional resources to manage default ACL, security group, and route tables.


clstokes:
AWS VPCs come with a default set of network resources - an ACL, a security group, and a main route table. There's no way to then manage these resources directly, because Terraform has no awareness of them (as of v0.6.14).

New resource types should be added to Terraform that allow these resources to be explicitly managed.

catsby:
I'm [WIP] here

kamaltherocky:
+1

This will be very helpful feature when you want to override the default configuration (SG, Route Table and NACL)

catsby:
#6165 added the default network acl resource
#8323 adds the default route table resource

catsby:
Default security group added in #8861

catsby:
@clstokes good to close this?

clstokes:
@catsby closed! Thanks! 🍰

PR #8323, opened 2016-08-19
provider/aws: Add aws_default_route_table resource

catsby:
Adds a resource to manage the default route table that comes with every VPC. 
This is an advanced resource with caveats mentioned in the documentation. 

Part of #6093

### DIFF
--- a/builtin/providers/aws/resource_aws_default_route_table.go
+++ b/builtin/providers/aws/resource_aws_default_route_table.go
@@ -1,0 +1,245 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsDefaultRouteTable() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsDefaultRouteTableCreate,
+		Read:   resourceAwsDefaultRouteTableRead,
+		Update: resourceAwsRouteTableUpdate,
+		Delete: resourceAwsDefaultRouteTableDelete,
+
+		Schema: map[string]*schema.Schema{
+			"default_route_table_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"vpc_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"propagating_vgws": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+
+			"route": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"cidr_block": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+
+						"gateway_id": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+
+						"instance_id": &schema.Schema{
+							Type:     schema.TypeString,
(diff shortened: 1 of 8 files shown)


builtin/providers/aws/resource_aws_default_route_table.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	rt := rtRaw.(*ec2.RouteTable)
+
+	d.Set("vpc_id", rt.VpcId)
+
+	// revoke all default and pre-existing routes on the default route table.
+	// In the UPDATE method, we'll apply only the rules in the configuration.
+	log.Printf("[DEBUG] Revoking default routes for Default Route Table for %s", d.Id())
+	if err := revokeAllRouteTableRules(d.Id(), meta); err != nil {
```
stack72: Is there a potentially for this to cause issues if the revoke happens and the update fails?

builtin/providers/aws/resource_aws_default_route_table.go
```diff
@@ -0,0 +1,204 @@
...
+	filter2 := &ec2.Filter{
+		Name:   aws.String("vpc-id"),
+		Values: []*string{aws.String(d.Get("vpc_id").(string))},
+	}
+
+	findOpts := &ec2.DescribeRouteTablesInput{
+		Filters: []*ec2.Filter{filter1, filter2},
+	}
+
+	resp, err := conn.DescribeRouteTables(findOpts)
```
stack72: Would it not be best to use d.Id here as part of the lookup?

builtin/providers/aws/resource_aws_default_route_table.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(resp.RouteTables) < 1 || resp.RouteTables[0] == nil {
+		return fmt.Errorf("Default Route table not found")
+	}
+
+	rt := resp.RouteTables[0]
```
stack72: Can we guarantee that the first RouteTable in the collection is the correct one?

stack72: I was being stupid here... comment revoked ;)

website/source/docs/providers/aws/r/vpc.html.markdown
```diff
@@ -61,6 +61,7 @@ The following attributes are exported:
      [`aws_main_route_table_association`](/docs/providers/aws/r/main_route_table_assoc.html).
 * `default_network_acl_id` - The ID of the network ACL created by default on VPC creation
 * `default_security_group_id` - The ID of the security group created by default on VPC creation
+* `default_security_group_id` - The ID of the route table created by default on VPC creation
```
stack72: this should be `default_route_table_id`

stack72:
Few small things but this is really good - loved the note about being wary! I can forsee people getting themselves into trouble here so it's good to point this out in advance

builtin/providers/aws/resource_aws_default_route_table.go
catsby: We want to revoke the rules regardless if there is any update to be made. The procedure is that during "creation" we wipe the slate clean. This is documented to warn people .

builtin/providers/aws/resource_aws_default_route_table.go
catsby: on read we need to find the default route table for this VPC. Unlike Default Network ACLs or Default Security Groups, the Default Route Table can be changed (you can make a new one and then make that "default"). So we need to find if we are the default, and then refresh and show a diff (if there is one) accordingly.

builtin/providers/aws/resource_aws_default_route_table.go
catsby: eh

website/source/docs/providers/aws/r/vpc.html.markdown
catsby: penguin

builtin/providers/aws/resource_aws_default_route_table.go
stack72: 👍

stack72:
The PR looks good to me - the questions were based more on my misunderstanding of how it would work. Explanations and the documentation help to clear this up

LGTM :)

Issue #5067, opened 2016-02-09
aws_elasticsearch_domain keeps reapplying 


thattommyhall:
Similar to #3634 

I am using your JSON format to create the aws_elasticsearch_domain

```
"aws_elasticsearch_domain": {
  "logs": {
    "domain_name": "logs",
    "advanced_options": {
      "rest.action.multi.allow_explicit_index": true
    },
    "access_policies": "{\"Version\":\"2012-10-17\",\"Statement\":{\"Action\":\"es:*\",\"Principal\":\"*\",\"Effect\":\"Allow\",\"Condition\":{\"IpAddress\":{\"aws:SourceIp\":[\"${aws_eip.public-a-nat.public_ip}\",\"${aws_eip.public-b-nat.public_ip}\",\"${aws_eip.public-c-nat.public_ip}\"]}}}]}",
    "snapshot_options": {
      "automated_snapshot_start_hour": 23
    }
  }
}
```

It successfully creates, but every time I run it it says it is changing 

```
{
    "Statement": [{
        "Action": "es:*",
        "Condition": {
            "IpAddress": {
                "aws:SourceIp": ["52.48.255.155", "52.48.254.180", "52.48.255.154"]
            }
        },
        "Effect": "Allow",
        "Principal": "*",
        "Resource": "arn:aws:es:eu-west-1:<ID>:domain/logs/*"
    }],
    "Version": "2012-10-17"
}
```

to 

```
{
    "Statement": [{
        "Action": "es:*",
        "Condition": {
            "IpAddress": {
                "aws:SourceIp": ["52.48.255.155", "52.48.254.180", "52.48.255.154"]
            }
        },
        "Effect": "Allow",
        "Principal": "*"
    }],
    "Version": "2012-10-17"
}
```

(ie adding the 'Resource')

If I add the Resource entry after it it created then it no longer re-applies, but obviously I can't do it for creation as I dont know the ARN, also if I lookup the ARN, I get `Self reference: aws_elasticsearch_domain.logs`

This is my first mild snag in ~6 months of using Terraform, it is honestly the best new tech ive used in the last 5y, thanks!

radeksimko:
> I can't do it for creation as I dont know the ARN

I think the easiest way we can get around this would be introducing a new resource, something like `aws_elasticsearch_domain_policy` which would depend on `aws_elasticsearch_domain` and would only be responsible to setting the policy.

[...]

PR #8350, opened 2016-08-21
provider/aws: IAM Policy Normalization Helpers

apparentlymart:
We have some long-standing issues where Terraform sends a user-supplied IAM policy document to an AWS API, and then on read AWS returns the same policy with a slightly different JSON structure due to some normalization done on the AWS side. Terraform then thinks there's a diff and will never converge.

In #6956 we started to make the `aws_iam_policy_document` data source generate documents in a shape that many AWS services consider to be normalized, which gave one path to addressing this issue. However, it doesn't help when users are writing their own IAM policy documents (or generating them with some other tool) and not using the data source.

This change is based on what was proposed in #7785, which turned the existing "IAM Policy Model" utility into something that can both read and write IAM policy documents (where before it could only write) and then used it to implement some general normalizations of IAM policies:

* For attributes that take sets of strings, represent a single-item set in JSON as a bare string rather than a one-element list containing that string.
* For attributes that take sets of strings, sort those strings lexicographically since ordering isn't significant in a set. 
* Don't include the `Sid` attribute in a statement when it's empty.

The primary goal here is to be internally consistent in normalization within Terraform. A secondary goal is to produce a normalization similar to what AWS itself would produce, but this is not *essential* 
because the our normalization code will be applied to both the values in the configuration and the values retrieved from the AWS APIs.

Building on the facility from #7785, this also includes a mechanism for callers to do additional *semantic* normalizations on the policy data, to handle cases where certain AWS services perform their own normalization on the data that assumes knowledge of that particular service. For example, in cases where a policy is defined as an attribute of a resource, some services (e.g. AWS ElasticSearch, as discussed in #5067) will populate the `Resource` attribute with the resource ARN on Read, even though that's redundant since this relationship is implied by it being specified on the resource.

The primary concern of this PR is to implement the utility functions for normalization. It also changes the `aws_cloudfront_stack` resource to use it, as a demonstration of how it can be used. Normalization can be added to other resources (including the several others that were attacked by #7785) in later PRs.

---

*Compatibility notes*: I originally tried to make this produce policies that are normalized in a way that would be byte-for-byte compatible with the previous behavior of `aws_iam_policy_document`, but discovered that it was inexplicably using reverse-lexicographic sorting for strings.

On further reflection it felt like such strict compatibility was not needed because the generated policy will remain semantically equivalent, and `aws_iam_policy_document` has only been around for a short time and so it's unlikely to be widely used.

Adding new normalization to existing policy attributes on other resources is also expected to be generally fine, since Terraform will re-normalize both the configuration and the value stored in the state next time it plans. However, there is a potential compatibility concern if there are any valid policy documents which the new validation function is unable to parse, whether due to a bug or due to undocumented AWS behavior; this is a general concern each time stricter validation is added, but in this case I expect the risk is minimal because it's unlikely that a policy document would be invalid enough to fail our parsing but yet still accepted by AWS APIs.

### DIFF
--- a/builtin/providers/aws/data_source_aws_iam_policy_document.go
+++ b/builtin/providers/aws/data_source_aws_iam_policy_document.go
@@ -112,12 +112,12 @@ func dataSourceAwsIamPolicyDocumentRead(d *schema.ResourceData, meta interface{}
 		}
 
 		if resources := cfgStmt["resources"].(*schema.Set).List(); len(resources) > 0 {
-			stmt.Resources = dataSourceAwsIamPolicyDocumentReplaceVarsInList(
+			stmt.Resources = dataSourceAwsIamPolicyDocumentReplaceVarsInSet(
 				iamPolicyDecodeConfigStringList(resources),
 			)
 		}
 		if resources := cfgStmt["not_resources"].(*schema.Set).List(); len(resources) > 0 {
-			stmt.NotResources = dataSourceAwsIamPolicyDocumentReplaceVarsInList(
+			stmt.NotResources = dataSourceAwsIamPolicyDocumentReplaceVarsInSet(
 				iamPolicyDecodeConfigStringList(resources),
 			)
 		}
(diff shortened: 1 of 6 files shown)


coen-hyde:
It should be noted that the `aws_iam_policy_document` doesn't help resolve conflicts arising from resorting of arrays on AWS's side. I'm still getting ordering conflicts using the `aws_iam_policy_document` on 0.7.2. I thought #6956 fixed this?

apparentlymart:
That change caused the string set items to be sorted in reverse lexical order. I expect it's not matching what AWS services are doing because they presumably sort in forward lexical order.

The change here should avoid such issues because it will (once applied across all of the policy attributes, which this patch doesn't yet do) normalize both the value given in the configuration *and* the value return from the API with the same logic, so they should both end up matching.

coen-hyde:
AWS sort in a methodology known as "AWS's school of sorting" aka not quite random. What you are proposing sounds like the only way to solve this problem.

apparentlymart:
Closing this since it sounds like a follow-up of #8585 is going to address this a different way.

PR #8352, opened 2016-08-21
Add note to use Server-Defined-Cipher-Order

### LINKED ISSUE #8340: Unable to set Server Order Preference in AWS Load Balancer SSL Negotiation Policy
Hello,

I'm running Terraform v0.7.1, and trying to set an SSL negotiation policy to load balancers in AWS using the `aws_lb_ssl_negotiation_policy` resource, as follows:

```
resource "aws_lb_ssl_negotiation_policy" "foo" {
  count = "${var.foo_count}"
  name = "${var.foo}-${var.bar}"
  load_balancer = "${aws_elb.foo.id}"
  lb_port = 443
  attribute {
    name = "Protocol-SSLv3"
    value = "false"
  }
  ...
  attribute {
    name = "Server Order Preference"
    value = "true"
  }
}
```

When I run a Terraform apply, I get:

> `aws_lb_ssl_negotiation_policy.foo: Error creating Load Balancer Policy: ValidationError: PolicyAttributeName may only contain non-whitespace printable ASCII characters`

I wondered whether this occurred because "Server Order Preference" contains spaces when it is used as a value to the `Name` property of the `attribute` parameter, but this appears not to be the case, as when I replace the spaces with hyphens (as in use in the rest of the documentation), the API returns:

> `aws_lb_ssl_negotiation_policy.foo: Error creating Load Balancer Policy: ValidationError: PolicyAttribute 'Server-Order-Preference' is not defined in the policy type`

Thought I'd raise this here. Going to keep trying to work out if I can get it working some way, and then look at the best way of resolving this fully for others in future when I know what the root of the problem is.

ajlanghorn:
The Terraform documentation, rather correctly, refers to a list of options
you can pass to an Elastic Load Balancer from the AWS documentation. All but
one of these options works; 'Server Order Preference' doesn't work, because
the API refers to it as 'Server-Defined-Cipher-Order'.

Add a note to explain this, at least as a temporary solution.

Fixes #8340.

### DIFF
--- a/website/source/docs/providers/aws/r/lb_ssl_negotiation_policy.html.markdown
+++ b/website/source/docs/providers/aws/r/lb_ssl_negotiation_policy.html.markdown
@@ -76,6 +76,8 @@ balancer.
 
 To set your attributes, please see the [AWS Elastic Load Balancing Developer Guide](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-security-policy-table.html) for a listing of the supported SSL protocols, SSL options, and SSL ciphers.
 
+~> **NOTE:** The AWS documentation references Server Order Preference, which the AWS Elastic Load Balancing API refers to as `Server-Defined-Cipher-Order`. If you wish to set Server Order Preference, use this value instead.
+
 ## Attributes Reference
 
 The following attributes are exported:


mitchellh:
LGTM. Thanks!

PR #8359, opened 2016-08-21
provider/aws: New Data Source: aws_acm_certificate

phinze:
Use this data source to get the ARN of a certificate in AWS Certificate
Manager (ACM). The process of requesting and verifying a certificate in ACM
requires some manual steps, which means that Terraform cannot automate the
creation of ACM certificates. But using this data source, you can reference
them by domain without having to hard code the ARNs as input.

The acceptance test included requires an ACM certificate be pre-created
in and information about it passed in via environment variables. It's a
bit sad but there's really no other way to do it.


builtin/providers/aws/data_source_aws_acm_certificate.go
```diff
@@ -0,0 +1,46 @@
...
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).acmconn
+	params := &acm.ListCertificatesInput{}
+	resp, err := conn.ListCertificates(params)
```
apparentlymart: It looks like this API returns results in pages... do we need to handle `NextToken` in the response in order to see the whole list once a user has lots of certificates?

(Not clear to me whether not setting `MaxItems` in the request means "give me all of them" or "use the default limit"...)

builtin/providers/aws/data_source_aws_acm_certificate.go
```diff
@@ -0,0 +1,46 @@
...
+
+func dataSourceAwsAcmCertificate() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsAcmCertificateRead,
+		Schema: map[string]*schema.Schema{
+			"domain": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"arn": &schema.Schema{
```
apparentlymart: Should we also have an optional `statuses` attribute here, so that users can filter for only certificates that are in a good state?

```hcl
data "aws_acm_certificate" "example" {
  domain   = "example.com"
  statuses = ["issued"]
}
```

builtin/providers/aws/data_source_aws_acm_certificate.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return errwrap.Wrapf("Error describing certificates: {{err}}", err)
+	}
+
+	target := d.Get("domain")
+	for _, cert := range resp.CertificateSummaryList {
+		if *cert.DomainName == target {
+			// Need to call SetId with a value or state won't be written.
+			d.SetId(time.Now().UTC().String())
+			return d.Set("arn", cert.CertificateArn)
```
apparentlymart: Can there be potentially multiple certificates for the same domain with different statuses?

A pattern used in some of our other AWS data sources is to say that the constraints given in the arguments must match exactly one record. Matching zero records or matching two or more is an error, to avoid inadvertently selecting the "wrong" record. I'm worried here that I might accidentally select a revoked/inactive/expired cert rather than the active one with the same name, assuming that such a situation is possible.

builtin/providers/aws/data_source_aws_acm_certificate.go
phinze: Ah yes I had meant to look into this... will dig.

builtin/providers/aws/data_source_aws_acm_certificate.go
phinze: Good point! "Match exactly one" seems like a good rule of thumb. And I hadn't thought about the fact that revoked+active pairs for a single domain could be common. I'll add a status field and rework this a bit.

mitchellh:
Cool! I'll leave you and @apparentlymart to the review but this is a cool data source.

jen20:
Test Run:

```
TF_ACC=1 go test ./builtin/providers/aws -v -run TestAccAwsAcmCertificateDataSource_basic -timeout 120m
=== RUN   TestAccAwsAcmCertificateDataSource_basic
--- PASS: TestAccAwsAcmCertificateDataSource_basic (21.51s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	21.534s
```

phinze:
LGTM @jen20 - I can't approve via reviews since I was the original author lol

PR #8370, opened 2016-08-22
Do not sleep between first attempt or between successful attempts

jen20:
Rebase of #7709.

### DIFF
--- a/helper/resource/state.go
+++ b/helper/resource/state.go
@@ -2,7 +2,6 @@ package resource
 
 import (
 	"log"
-	"math"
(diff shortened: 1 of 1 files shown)


jen20:
cc @dtolnay

stack72:
Took me a while to realise what this was for but after speaking to @jen20:

```
the first time we don't want to wait
because otherwise we get arbitrary delay at the start
```

in that case LGTM!

jen20:
Thanks for the work on this @dtolnay!

PR #8374, opened 2016-08-22
build: Exclude darwin/arm from OS/Arch targets

### LINKED ISSUE #8365: build: Exclude darwin/amd64 from targets
We don't need to build for `darwin/amd64` and it presents alarming errors when making release builds. Consequently we should modify the build script to opt in to supported OS/Architecture pairs.

jen20:
Fixes #8365


stack72:
LGTM!

PR #8381, opened 2016-08-22
provider/aws: `aws_elasticache_replication_group_id` validation change

### LINKED ISSUE #8379: Elasticache Replication Group allows replication_group_id to be longer than aws will accept
`replication_group_id` is limited to 20 characters by the provider but the AWS API only allows 16. These should really match to save users some frustration.

```
* aws_elasticache_replication_group.bar: Error creating Elasticache Replication Group: InvalidParameterValue: Replication group id should be no more than 16 characters.
	status code: 400, request id:
```

stack72:
Fixes #8379 

The AWS documentation tells us the following:

```
--replication-group-id (string)

The replication group identifier. This parameter is stored as a
lowercase string.
Constraints:
A name must contain from *1 to 20* alphanumeric characters or hyphens.
The first character must be a letter.
A name cannot end with a hyphen or contain two consecutive hyphens.
```

This is not correct and is causing users errors:

```
* aws_elasticache_replication_group.bar: Error creating Elasticache
* Replication Group: InvalidParameterValue: Replication group id should
* be no more than 16 characters.
    status code: 400, request id:
```

Tuning the Validation from 20 to 16 characters to avoid user issues

### DIFF
--- a/builtin/providers/aws/resource_aws_elasticache_replication_group.go
+++ b/builtin/providers/aws/resource_aws_elasticache_replication_group.go
@@ -404,9 +404,9 @@ func validateAwsElastiCacheReplicationGroupEngine(v interface{}, k string) (ws [
 
 func validateAwsElastiCacheReplicationGroupId(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
-	if (len(value) < 1) || (len(value) > 20) {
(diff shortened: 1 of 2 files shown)


jen20:
LGTM

devinsba:
Is there a way to report bugs back to the AWS folks on this for their docs?

PR #8390, opened 2016-08-22
provider/google: Fix panic and state migration for google_compute_firewall

### LINKED ISSUE #8341: "Unexpected EOF" and "connection is shut down" with google providers when using terraform 0.7.1
Hi there,

I just switched to `terraform 0.7.1`, my infra is a mix of `±200` ressources on gcp + openstack (so multi provider).

**Note**: Everything work's like a charm with `terraform 0.7.0`

### Terraform Version
* `terraform 0.7.1`

### Affected Resource(s)

- `google_*`

### Debug Output

* http://pub.sebian.fr/pub/crash-run.log.gpg
 * `size:8417`
 * `hash:b56ac0776faa3f638bf919ba130509fb`
 * crypted with `0x51852D87348FFC4C`

### Panic Output

* http://pub.sebian.fr/pub/crash.log.gpg
 * `size:279154`
 * `hash:90280650f20ba041310e84f247cc1b9a`
 * crypted with `0x51852D87348FFC4C`

### Expected Behavior

> What should have happened?

Terraform should report or execute the plan for `google_records`

### Actual Behavior

> What actually happened?

Any `google_record` throws either a "Unexpected EOF" and "connection is shut down" error.

**Note**: I tried with `-parallelism=1` as suggested in https://github.com/hashicorp/terraform/issues/8011#issuecomment-238953349 but no change…

### Steps to Reproduce
1. `terraform plan`

### Important Factoids

> Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

No

### References
Maybe this one (but for cloudfare ressources)
- GH-8011

Thanks in advance!

### LINKED ISSUE #8345: panic: interface conversion: interface is nil, not string
### Terraform Version

0.7.1. Under 0.7.0 it works just fine.

### Affected Resource(s)

- google_compute_firewall

The `google_compute_firewall` resource is simple enough and contains just 2 rules to allow incoming ports to specific instances marked by tags.

### Debug Output

http://sprunge.us/fCWA

All sensitive information has been removed.

### Steps to Reproduce

1. `terraform plan -target "module.firewall_gce"`

### References

This might be the cause or related:

- GH-8341

jen20:
This pull request fixes the issues reported against Terraform 0.7.1 whereby a state migration which never ran (because it wasn't hooked in) caused a panic because of uninitialized or incorrect structure use.

It also tidies up the code for `google_compute_firewall` to conform to the newer style without redundant type declarations.

Fixes #8345, Fixes #8341.

### DIFF
--- a/builtin/providers/google/resource_compute_firewall.go
+++ b/builtin/providers/google/resource_compute_firewall.go
@@ -23,6 +23,7 @@ func resourceComputeFirewall() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 		SchemaVersion: 1,
+		MigrateState:  resourceComputeFirewallMigrateState,
 
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
--- a/builtin/providers/google/resource_compute_firewall.go
+++ b/builtin/providers/google/resource_compute_firewall.go
@@ -26,29 +26,29 @@ func resourceComputeFirewall() *schema.Resource {
 		MigrateState:  resourceComputeFirewallMigrateState,
 
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"network": &schema.Schema{
+			"network": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"allow": &schema.Schema{
+			"allow": {
 				Type:     schema.TypeSet,
 				Required: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"protocol": &schema.Schema{
+						"protocol": {
 							Type:     schema.TypeString,
 							Required: true,
 						},
 
-						"ports": &schema.Schema{
+						"ports": {
 							Type:     schema.TypeList,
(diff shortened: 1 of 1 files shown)


phinze:
LGTM pending travis

jen20:
Whoops looks like some kind of issue there, I'll take a look shortly when I'm back.

builtin/providers/google/resource_compute_firewall.go
```diff
@@ -23,6 +23,7 @@ func resourceComputeFirewall() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 		SchemaVersion: 1,
+		MigrateState:  migrateFirewallStateV0toV1,
```
cblecker: I'm still learning go, so I apologize if I'm off here, but I think the reason this is breaking is the wrong function is being called. My guess is this should be [resourceComputeFirewallMigrateState](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_firewall_migrate.go#L13).

phinze: Yep that's it! We'll get this sorted soon. Thanks @cblecker!

nwwebb:
Thanks @jen20! My fault for forgetting to hook in the state migration. Appreciate the fix

PR #8394, opened 2016-08-22
command: use bufio.ReadString instead of scanning to get spaces

### LINKED ISSUE #2628: Interactive prompt only takes the first string in a multi string input
The interactive prompt only takes the first string in a multi string input.

For example : 

```
var.stream-name
  Default: stream
  Enter a value: sample stream from dev
```

only the word "sample" will only be parsed.

I manage to workaround by specifying over the CLI.
```sh
terraform apply -var 'stream-name=sample stream from dev'
```

mitchellh:
Fixes #2628 

Surprised this hasn't bit more people! Turns out fmt's Scan functions break on spaces and I don't see a way around that. Luckily `bufio` has some nice functions reading up to a certain character. Added another test to verify.

### DIFF
--- a/command/ui_input.go
+++ b/command/ui_input.go
@@ -11,6 +11,7 @@ import (
 	"os/signal"
 	"strings"
 	"sync"
+	"unicode"
 
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/mitchellh/colorstring"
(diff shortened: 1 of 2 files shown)


jbardin:
LGTM!

PR #8396, opened 2016-08-22
terraform: prefix destroy resources with module path [GH-2767]

### LINKED ISSUE #2767: UI inconsistency: "Destroying..." message not scoped by module
Snippet from an instance replacement:

```
aws_db_instance.mod: Destroying...
module.network.openvpn.aws_instance.mod: Creating...
```

mitchellh:
Fixes #2767 

On destroy nodes we weren't populating the path for the instance info so the output wasn't showing it. This adds that and a test.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -2677,6 +2677,63 @@ func TestContext2Apply_destroy(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


phinze:
60 lines of test, 1 line of change, LGTM 😄

PR #8397, opened 2016-08-22
remove copystructure from the terraform package

jbardin:
copystructure had a habit of walking over structs which required locks
for reading some of the field values. Change all copies to use DeepCopy
or deepcopy (should these get unified?) recursively for the various
parts of the State.

Implement nestedCopy as a replacement for copystructure for our stringy
maps and slices.

Add a few "belt & suspenders" locks to the state structs while
concurrency issues are sorted out.

### DIFF
--- a/terraform/nested_copy.go
+++ b/terraform/nested_copy.go
@@ -1,0 +1,51 @@
+package terraform
+
+import "fmt"
+
+// nestedCopy returns a deep copy of i, provided that all interface{} values
+// are string, []string, []interface{}, map[string]string, or
+// map[string]interface{}. This is an interim replacement for
+// copystructure.Copy where we want to copy simple nested data structures, but
+// don't want to risk walking over structs that may contain locks.
+func nestedCopy(src interface{}) interface{} {
+	switch v := src.(type) {
+	case string:
+		return v
+	case []interface{}:
+		var sliceCopy []interface{}
+		for _, i := range v {
+			sliceCopy = append(sliceCopy, i)
+		}
+		return sliceCopy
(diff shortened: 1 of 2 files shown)


jbardin:
Forgot to check vet. I forgot we copy some *State values around too (mostly for printing)

mitchellh:
Is there any way we can address this by wrapping the proper locks around copystructure? Or by perhaps extending copystructure to find Lockable resources somehow?

Copystructure was meant to be written so we wouldn't have to write custom deep copy functions and if we can make that work that'd be more ideal I think.

jbardin:
We can't lock _around_ copystructure.Copy, because we lose control as soon as it starts its walk through the structs.

I was tossing around the idea of modifying copystructure though, but I wasn't sure if we wanted to change that API (hence the "interim" comment in the nestedCopy). I'll submit a PR over there first.

jen20:
@jbardin we also use copystructure fairly extensively during state upgrading, but I'm fairly sure that is ok.

jbardin:
@mitchellh,

If we want to do this in copystructure, I've got a POC at https://github.com/mitchellh/copystructure/pull/5

We could use copystructure.Copiers, but after you implement the proper function for each type you haven't saved much over implementing the entire copy manually.

mitchellh:
Ah, hah, I'll take a look.

Yeah, Copiers was meant for more specific low-level types, and is recursive so the idea was you implement it for a basic struct and then nested items with Copiers would also go (e.x. time.Time).

jbardin:
closing this in favor of #8452

PR #1418, opened 2015-04-07
config/module: storage path should be independent of system

mitchellh:
This is a really important change to make copying entire `.terraform` directories work properly. Before, the hash in the `.terraform/modules` folder would be calculated based on the full URL of the source of the module. There are benefits to this:

  1. One copy per source
  2. When the source changes, `terraform` immediately notices.

The downside is that the paths aren't portable at all, which conflicts with `terraform push`. In this PR, I change the hash to be the hash of the "path" of the module: example "root.child.child2". This makes it portable, but makes it so that we have an exact 1:1 mapping of download of a module to folder. I don't think this optimization matters right now.

I think the proper long term solution is to have some sort of "module.lock" file that does this mapping for us so the hash doesn't matter at all. But for a 0.4.1 I wanted to minimize what is touched.

@phinze: Would appreciate some thoughts here and a review.

PR #8398, opened 2016-08-22
config/module: use source as part of key

### LINKED ISSUE #3070: terraform module symlink error
My module was pointing to a GH repository. When debugging, I changed it to point to the local checkout of that repository. This caused the following error:

```
$ terraform get -update
Get: file:///Users/Jean/code/terraform/terraform-dcos (update)
Error loading Terraform: Error downloading modules: error downloading module 'file:///Users/Jean/code/terraform/terraform-dcos': destination exists and is not a symlink
```

removing the `.terraform/modules` directory solved the problem.

Some possible solutions:

* automatically remove/override the module directory with a symlink
* ask yes/no
* clarify why the error happens and how to solve it

mitchellh:
Fixes #3070 

Ehhh, couldn't find a really great way to test this without embedding a git repo which I didn't want to do here. I view this as sort of a stop-gap solution anyways as we do a bigger plan around how modules should be stored on disk.

This is **kind of** backwards incompatible in that after updating you'll have to run `terraform get` again. I don't think that is very disruptive but happy to talk about it.

### DIFF
--- a/config/module/tree.go
+++ b/config/module/tree.go
@@ -170,7 +170,7 @@ func (t *Tree) Load(s getter.Storage, mode GetMode) error {
 
 		// Get the directory where this module is so we can load it
 		key := strings.Join(path, ".")
-		key = "root." + key
+		key = fmt.Sprintf("root.%s-%s", key, source)
 		dir, ok, err := getStorage(s, key, source, mode)
 		if err != nil {
 			return err


jbardin:
this LGTM, as long as we're OK with having to re-get.

rowleyaj:
@mitchellh, let me know if you'd prefer me to open a new issue rather than comment on this merged PR.

A while ago I looked at changing this in our fork (https://github.com/Ensighten/terraform/pull/9), to only use source for the key as it was previously. It looked like it was originally changed away from that (https://github.com/hashicorp/terraform/pull/1418) due to incompatibilities with Windows.

I don't use Windows so can't test, but would this reintroduce the same problems that existed prior to #1418?

I'm interested in improving the module get / update process, but not really sure where to start in the code beyond what I found previously. The reason for wanting to improve this is it takes quite some time for us to download all of our modules, and several are repeated. Whislt testing changing the key back to source helped improve things for us in terms of the initial get, but an update still wants to grab every module. I was wondering if there was a way to record if we had previously got a module with the same source as it built the tree, and then not try again if it had?

From #1418:
> I think the proper long term solution is to have some sort of "module.lock" file that does this mapping for us so the hash doesn't matter at all.

From above:
> as we do a bigger plan around how modules should be stored on disk

These two comments sound interesting, would be good to hear the plans for this if you're able to share.

PR #8405, opened 2016-08-23
provider/openstack: docs and tests for allowed_address_pairs

jtopjian:


stack72:
LGTM! I don't have an OpenStack setup to be able to run the tests though

jtopjian:
That's OK! I have access to a number of them 😉

PR #2375, opened 2015-06-16
Add ForceNew metadata_startup_script field

sparkprime:
This lacks tests & docs, but I wanted to discuss it before working further:

The idea is that if you use the metadata field startup-script to define the internal state of the instance, then you want changing that script to recreate the instance.  The same is true for aws_instance and the userdata field, but this PR is just for GCE right now.

This solution adds another attribute called metadata_startup_script that is ForceNew and cannot be specified at the same time as metadata.startup_script.

What do you think?

PR #8407, opened 2016-08-23
provider/google: instance template support for metadata_startup_script

### LINKED ISSUE #7827: GCE instance template support for metadata_startup_script
Currently the gce provider support metadata_startup_script for google_compute_instance but not google_compute_instance_template.

The GCE API support this for both
https://cloud.google.com/compute/docs/instance-templates

could we please have this feature supported.

cblecker:
Fixes #7827.

This pull updates the `google_compute_instance_template` resource to use the same metadata arguments and validation functions as the `google_compute_instance` resource. That being said, while this directly addresses the issue, I'm not sure it's the most graceful way of doing it for a couple of reasons.

For users with existing configuration, this will force them to migrate their:

``` hcl
metadata {
    startup-script = "echo hi > /test.txt"
}
```

to: 

``` hcl
metadata_startup_script = "echo hi > /test.txt"
```

This results in a plan that detects a change to the config and forces deletion and recreation of the template. If that template is in use by a `google_compute_instance_group_manager` resource, then the apply will fail as a template can't be deleted while in use. Then you'd need to taint the instance group which would delete and recreate all instances created by it.

Second, this change isn't strictly necessary. While it is good to have consistent configuration, functions, and validations where possible between the two resources, instance templates are more restrictive than instances. As an existing instance template can't be updated at all, any change of any metadata at all results in the above deletion and recreation scenario. The original reason as far as I can tell for splitting the main metadata apart from the startup script was that if the startup script changed, then the instance should be destroyed and recreated so that it can be re-run. This left the remaining metadata keys to be changed freely. As all metadata is treated the same here splitting the startup script out isn't really needed.

Perhaps a state migration would solve this? Or perhaps it's not worth the hassle and is simply a WONTFIX. Thoughts? :)

### DIFF
--- a/builtin/providers/google/resource_compute_instance_template.go
+++ b/builtin/providers/google/resource_compute_instance_template.go
@@ -168,7 +168,15 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 			},
 
 			"metadata": &schema.Schema{
-				Type:     schema.TypeMap,
+				Type:         schema.TypeMap,
+				Optional:     true,
+				ForceNew:     true,
+				Elem:         schema.TypeString,
+				ValidateFunc: validateInstanceMetadata,
+			},
+
+			"metadata_startup_script": &schema.Schema{
+				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
 			},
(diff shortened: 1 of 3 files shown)


cblecker:
@evandbrown @lwander -
If either of you Google ninjas have cycles, I'd love feedback on this direction 😀

baskaran-md:
Any updates on this feature?

stack72:
ping @evandbrown :)

evandbrown:
Hey @cblecker, sorry I missed this a few months ago! O_o

I'm actually not a fan of how `google_compute_instance` treats startup scripts as a special field (I'm guessing this was done to mimic how `gcloud` and the web UI level-up that feature).

I'd actually be in favor of reverting the `google_compute_instance` behavior with a migration, which shouldn't result in the need to taint or re-create resources.

How does that sound?

cblecker:
No worries, @evandbrown! I know you're a busy guy.

Yeah, I think the splitting out kind of over complicates things. I tried to dig and find where this was introduced, and it looks like all the way back in #2375.

I think either way is acceptable, but the big key for me is that there is consistent behaviour between the two resource types. It's extremely powerful to be able to take a working `google_compute_instance` resource, and then change it to a template by changing the resource type with little to no other changes required.

cblecker:
Closing in favour of #10537. Thanks @paddyforan!

PR #8413, opened 2016-08-23
Added example of how the Option settings works

arsdehnel:
It took me some googling to find this and thought it would be good to have an example of how that structure is meant to be right in the docs.


stack72:
Hi @arsdehnel 

thanks so much for the work here - I agree that this is a useful addition :)

Paul

PR #8424, opened 2016-08-23
docs/aws: Documentation Update for `aws_db_option_group`

### LINKED ISSUE #8421: aws_db_option_group apply_immediately argument
### Terraform Version
0.7.1

### Affected Resource(s)
aws_db_option_group

The documentation lists the `apply_immediately = true` argument for the `aws_db_option_group` resource but it throws an error when trying to run a `terraform plan` when it's added to a config file following the example in the documentation - https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/r/db_option_group.html.markdown#example-usage

Also tried adding the `apply_immediately = true` in the `option` block but that doesn't work either.

stack72:
Remove `apply_immediately` from the example and add a short note that we
apply_immediately by default

fixes #8421

### DIFF
--- a/website/source/docs/providers/aws/r/db_option_group.html.markdown
+++ b/website/source/docs/providers/aws/r/db_option_group.html.markdown
@@ -28,10 +28,10 @@ resource "aws_db_option_group" "bar" {
   option {
     option_name = "TDE"
   }
-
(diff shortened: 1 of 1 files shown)


catsby:
👍

PR #8445, opened 2016-08-24
provider/aws Fix import errors when aws_route has VPC endpoint

### LINKED ISSUE #8225: crash when importing aws_route_table
Terraform v0.7.0 crashes when i try to import aws_route_table resource:
```terraform import aws_route_table.private <my_route_table_id>```

crash.log:
https://gist.github.com/drscre/1b7994cadb588924ce627b2487e985a9

brianbianco:
Fixes #8225

If an imported aws_route contains a PrefixListId entry the RouteHashID function
no longer throws a nil pointer reference.

### DIFF
--- a/builtin/providers/aws/import_aws_route_table.go
+++ b/builtin/providers/aws/import_aws_route_table.go
@@ -43,9 +43,10 @@ func resourceAwsRouteTableImportState(
 			// Minimal data for route
 			d := subResource.Data(nil)
 			d.SetType("aws_route")
-			d.Set("route_table_id", id)
 			d.Set("destination_cidr_block", route.DestinationCidrBlock)
+			d.Set("destination_prefix_list_id", route.DestinationPrefixListId)
 			d.SetId(routeIDHash(d, route))
+			d.Set("route_table_id", id)
 			results = append(results, d)
 		}
 	}
--- a/builtin/providers/aws/resource_aws_route.go
+++ b/builtin/providers/aws/resource_aws_route.go
@@ -359,7 +359,11 @@ func resourceAwsRouteExists(d *schema.ResourceData, meta interface{}) (bool, err
 
 // Create an ID for a route
 func routeIDHash(d *schema.ResourceData, r *ec2.Route) string {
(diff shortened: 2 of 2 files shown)


brianbianco:
I was able to successfully import my AWS route table after making this code fix.

I was UNABLE to create acceptance tests because it seems, at least on my machine, the aws providers acceptance tests are failing on the master branch (and even from the 0.7.1 tag).  Very possible I am just doing something wrong and am open to suggestion as to how I should fix it and even how I should write a good acceptance test for this PR.

jonbender:
@brianbianco seems like this patch doesn't quite account for the logic to find this route in AWS (findResourceRoute currently takes a cidr which this route does not have), or the changes to the AWS route resource schema in terraform (destination_cidr_block is a `Required` attribute which isn't the case with VPEs).

In any case, I'd recommend closing and check in with hashicorp on whether they're working on a fix.  I'd love to see this implemented but I'm guessing since its a somewhat involved change they're going to take care of it?

brianbianco:
@jonbender It does seem like a fairly in-depth change.  I'll poke around and look into the resource searching and schema.  My biggest issue was that I ceased doing any further changes as I couldn't get the acceptance tests to pass (and so I could not add my own).  I have yet to see a response in the issue from hashicorp so I was hoping to hear if they were working on fixing this or not.

catsby:
Hello – I've opened #9686 to patch the crash found in #8225 – right now the _direct_ creation of routes, either in-line in `aws_route_table` or externally with `aws_route`, require the `DestinationCidrBlock`, as per the AWS API:

- http://docs.aws.amazon.com/fr_fr/AWSEC2/latest/APIReference/API_CreateRoute.html

Routes that do not contain `DestinationCidrBlock` like those created from `aws_vpc_endpoint` are managed externally, so in #9686 we exclude them from import.

If you feel this is a bad idea please let me know, otherwise I'm going to close this. Thanks!

brianbianco:
Great @catsby, Thanks for this.  This seems like a reasonable work around, especially if the AWS api itself requires you have a CIDR block.

PR #8451, opened 2016-08-24
Allow specifying ldflags via env vars

grubernaut:
Allows specifying `LD_FLAGS` during development builds.

```
$ LD_FLAGS="-linkmode=external" make dev
```

Since the version of DTrace on macOS Sierra (On the public beta's at least) has been updated, this allows DTrace to run on Terraform during development/debugging.

```
$ sudo dtrace -n 'pid$target::main.main:entry' -c "terraform apply"
dtrace: description 'pid$target::main.main:entry' matched 1 probe

Outputs:

foo = bar
dtrace: pid 23096 has exited
CPU     ID                    FUNCTION:NAME
  2 265673                  main.main:entry
```

Also redirects the `@which stringer` command inside the `generate` Makefile target to `/dev/null` so we stop echoing the path to the `stringer` binary on every call to `generate`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ vet:
 # generate runs `go generate` to build the dynamically generated
 # source files.
 generate:
-	@which stringer ; if [ $$? -ne 0 ]; then \
+	@which stringer > /dev/null; if [ $$? -ne 0 ]; then \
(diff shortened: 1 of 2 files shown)


jbardin:
LGTM

PR #8452, opened 2016-08-24
Add locks to state structs for copying

jbardin:
Add locks to the state structs to handle concurrency during the graph
walks. We can't embed the mutexes due to serialization constraints when
communicating with providers, so expose the Lock/Unlock methods
manually.

Use copystructure.LockedCopy to ensure locks are honored (Waiting on update to github.com/mitchellh/copystructure)


jbardin:
I'm not saying this fixes all races, but I've exhausted what I can currently trigger ;)

mitchellh:
LGTM! May need to update to use the copystructure config class but looks great.

mitchellh:
You already updated it. Well, LGTM then.

PR #8453, opened 2016-08-24
Output 'destroy complete' when it's destroy (was: apply)

radeksimko:
Previously the output from `terraform destroy` looked like this:

```
Apply complete! Resources: 0 added, 0 changed, 1 destroyed.
```
even though destroy is (under the hood) basically `apply`, it may IMO cause unnecessary confusions for common users, hence this PR is changing it to

```
Destroy complete! Resources: 0 added, 0 changed, 1 destroyed.
```

### DIFF
--- a/command/apply.go
+++ b/command/apply.go
@@ -232,12 +232,19 @@ func (c *ApplyCommand) Run(args []string) int {
 		return 1
 	}
 
-	c.Ui.Output(c.Colorize().Color(fmt.Sprintf(
-		"[reset][bold][green]\n"+
-			"Apply complete! Resources: %d added, %d changed, %d destroyed.",
-		countHook.Added,
-		countHook.Changed,
-		countHook.Removed)))
+	if c.Destroy {
+		c.Ui.Output(c.Colorize().Color(fmt.Sprintf(
+			"[reset][bold][green]\n"+
+				"Destroy complete! Resources: %d destroyed.",
+			countHook.Removed)))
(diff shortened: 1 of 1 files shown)


apparentlymart:
This seems like a reasonable change... should we also drop the "0 added, 0 changed" parts since on destroy they will presumably always be zero anyway?

radeksimko:
> since on destroy they will presumably always be zero anyway?

I don't know, admittedly I haven't checked, but are you 100% sure? 😃 
I _think_ we don't display any diff in case of errors (i.e. fail to destroy a resource), so practically this is true, but theoretically it isn't.

apparentlymart:
If we ever created or updated resources as part of running `terraform destroy` I think that would be a worse UX issue than writing "Apply complete!" when we are destroying, but certainly I don't intend to block moving ahead with this change on account of this. :cherries:

radeksimko:
Sorry I misread your first message! 🙈 

> should we also drop the "0 added, 0 changed" parts

Of course we can do that. I originally thought you were suggesting to remove the whole diff (including `x destroyed`).

I'm going to update the PR soon.

radeksimko:
@apparentlymart  PR updated. Sorry for the initial confusion.

The new output for `destroy` is:

> Destroy complete! Resources: 1 destroyed.

mitchellh:
Nice simple fix that I think greatly enhances UX. LGTM!

PR #8454, opened 2016-08-24
provider/aws: Skip detaching when aws_internet_gateway not found

Bowbaq:
Deleting an `aws_internet_gateway` intermittently fails because the gateway fails to detach, and this is considered an error.

The gateway should only be deleted by terraform after it is detached, so there's no reason for it to disappear at this stage in the normal flow. This makes `InvalidInternetGatewayID.NotFound` a retryable error.

### DIFF
--- a/builtin/providers/aws/resource_aws_internet_gateway.go
+++ b/builtin/providers/aws/resource_aws_internet_gateway.go
@@ -249,24 +249,28 @@ func resourceAwsInternetGatewayDetach(d *schema.ResourceData, meta interface{}) 
 
 // InstanceStateRefreshFunc returns a resource.StateRefreshFunc that is used to watch
 // an EC2 instance.
-func detachIGStateRefreshFunc(conn *ec2.EC2, instanceID, vpcID string) resource.StateRefreshFunc {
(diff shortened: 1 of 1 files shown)


stack72:
Hey @Bowbaq 

Thanks for the work here :) This is a valid use case and the tests are looking good!

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSInternetGateway_'
2016/09/04 00:48:48 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSInternetGateway_ -timeout 120m
=== RUN   TestAccAWSInternetGateway_importBasic
--- PASS: TestAccAWSInternetGateway_importBasic (53.18s)
=== RUN   TestAccAWSInternetGateway_basic
--- PASS: TestAccAWSInternetGateway_basic (100.03s)
=== RUN   TestAccAWSInternetGateway_delete
--- PASS: TestAccAWSInternetGateway_delete (75.74s)
=== RUN   TestAccAWSInternetGateway_tags
--- PASS: TestAccAWSInternetGateway_tags (83.93s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	312.902s
```

Paul

PR #8456, opened 2016-08-24
command/push: -upload=modules=false works 

mitchellh:
Fixes an issue we saw in customer support.

With "-upload-modules=false" the modules were actually being updated causing them to not be updated in Atlas unless they were updated locally. This was hard to track down because our tests claimed it was working but that was only because they didn't set their `pwd` properly. I added a test to do this.

All remaining tests continue to pass.


jbardin:
LGTM

PR #8463, opened 2016-08-25
Validate the state on load

### LINKED ISSUE #4942: Duplicate module entries added to statefile cause confusing failures
If for any reason the statefile JSON ends up with duplicate entries for a given module path, very confusing behavior can result - with the duplicates being preserved for some time, then resulting in unresolvable state conflicts when working w/ remote state.

Terraform should recognize duplicate module entries in a state file being loaded and immediate raise a validation error.

mitchellh:
Fixes #4942

This adds a new validation function that is run on every state file load. If validation fails, the load is blocked. This is a super hostile action that should only be done for situations where the state isn't reconcilable automatically safely.

The only thing that can block this for now is multiple duplicate modules.

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -1522,6 +1522,7 @@ func ReadState(src io.Reader) (*State, error) {
 		return nil, fmt.Errorf("Decoding state file version failed: %v", err)
 	}
 
+	var result *State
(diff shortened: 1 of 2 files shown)


jbardin:
LGTM

PR #8466, opened 2016-08-25
Fix segmentation fault in "aws_api_gateway_base_path_mapping" resource

richardc:
It is possible for the `mapping.BasePath`, `mapping.RestApiId`, and
`mapping.Stage` to be nil when they have not been set for the
mapping.[1]  When this occurs a nil pointer is dereferenced and terraform
segmentation faults.

Here we remove the blind derefrences and trust in the behaviour of
(*ResourceData).Set() to handle the nil pointer safely.

[1] https://github.com/hashicorp/terraform/blob/master/vendor/github.com/aws/aws-sdk-go/service/apigateway/api.go#L4892-L4904

### DIFF
--- a/builtin/providers/aws/resource_aws_api_gateway_base_path_mapping.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_base_path_mapping.go
@@ -101,9 +101,9 @@ func resourceAwsApiGatewayBasePathMappingRead(d *schema.ResourceData, meta inter
 		return fmt.Errorf("Error reading Gateway base path mapping: %s", err)
 	}
 
-	d.Set("base_path", *mapping.BasePath)
-	d.Set("api_id", *mapping.RestApiId)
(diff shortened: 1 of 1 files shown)


stack72:
Hi @richardc 

Thanks so much for the PR here - you are indeed correct. We should pass the pointer into the Set func - that takes care of dereferencing it in the safe way :)

Will merge this when green!

Paul

BerndWessels:
@stack72 Hi, is this resolved? I still get this error with Terraform v0.7.5 :
```
* aws_api_gateway_base_path_mapping.SeoWebsite: SerializationError: failed decoding REST JSON error response
caused by: invalid character '<' looking for beginning of value
```

richardc:
@BerndWessels  This change only targeted the segmentation fault that was possible with the code at the time.  That you are managing to generate invalid configurations still would be a new issue, so it's probably better to continue this discussion in #8953

PR #8479, opened 2016-08-25
provider/aws: Propagate errors from DetachVolume

dtolnay:


mitchellh:
Well done. Thanks!

dtolnay:
This class of bugs really should be caught by static analysis. I filed #8536 to follow up.

DSh3p4rd:
What if I want that error to be ignored? https://github.com/hashicorp/terraform/issues/9000 makes a pretty good use case for doing so. 

The problem I have with Terraform is that it doesn't seem to know how to, detach a volume from an instance, terminate that instance, bring up a new one, and attach the existing volume. In my case I perform this when updating AMIs in a running environment.

PR #8481, opened 2016-08-25
command/push: fix tests to be vcs=false GH-8478

### LINKED ISSUE #8478: Brew 0.7.2 Fails due to vcs flag missing
I was trying to create the new Brew formula for 0.7.2 and when building it it fails out at.

`illegal
--- FAIL: TestPush_noUploadModules (0.02s)
    push_test.go:115: bad: 1
        An error has occurred while archiving the module for uploading:
        error detecting VCS: no VCS found for path: /private/tmp/terraform-20160825-45182-op6i5r/terraform-0.7.2/src/github.com/hashicorp/terraform/command/test-fixtures/push-no-upload
FAIL
FAIL    github.com/hashicorp/terraform/command  2.088s`

This is due to a new test committed in [https://github.com/hashicorp/terraform/commit/e8267f4907510b50b8d5d0d92f129b141cfa7116](https://github.com/hashicorp/terraform/commit/e8267f4907510b50b8d5d0d92f129b141cfa7116)command/push_test.go
As you can see its missing the "-vcs=false", flag that is used to support running outside of VCS is not shown. As you can see in this commit. It fixed the problem [https://github.com/hashicorp/terraform/commit/a2cb77058c50e1d5196935024e5141addae70cb3](https://github.com/hashicorp/terraform/commit/a2cb77058c50e1d5196935024e5141addae70cb3)
I tried to fix it here. But I do not have enough go experience to understand what is happening
[https://github.com/Techcadia/terraform/commit/516bf4cdc5d2656b5bce993a3874026e0b54578d](https://github.com/Techcadia/terraform/commit/516bf4cdc5d2656b5bce993a3874026e0b54578d)

`ok  	github.com/hashicorp/terraform/builtin/provisioners/remote-exec	0.028s
illegal
--- FAIL: TestPush_noUploadModules (0.04s)
	push_test.go:128: bad: []string{".terraform/", ".terraform/", ".terraform/terraform.tfstate", ".terraform/terraform.tfstate", "child/", "child/main.tf", "main.tf"}
FAIL
FAIL	github.com/hashicorp/terraform/command	2.026s
`

Please let me know if you have questions.

Thanks

mitchellh:
Fixes #8478 

This introduces the `-vcs=false` flag to the command to disable VCS mode. This exposed some weird buggy behavior where double entries are inserted into the archive. This isn't an actual problem though since it'll still extract properly at the expense of disk space. We can fix that in the future.

### DIFF
--- a/command/push_test.go
+++ b/command/push_test.go
@@ -107,6 +107,7 @@ func TestPush_noUploadModules(t *testing.T) {
 	defer os.Remove(testStateFileRemote(t, s))
 
 	args := []string{
+		"-vcs=false",
 		"-name=mitchellh/tf-test",
 		"-upload-modules=false",
 		path,
@@ -115,9 +116,14 @@ func TestPush_noUploadModules(t *testing.T) {
 		t.Fatalf("bad: %d\n\n%s", code, ui.ErrorWriter.String())
 	}
 
+	// NOTE: The duplicates below are not ideal but are how things work
+	// currently due to how we manually add the files to the archive. This
+	// is definitely a "bug" we can fix in the future.
 	actual := testArchiveStr(t, archivePath)
 	expected := []string{
 		".terraform/",
+		".terraform/",
+		".terraform/terraform.tfstate",
 		".terraform/terraform.tfstate",
 		"child/",
 		"child/main.tf",


stack72:
LGTM!

PR #8482, opened 2016-08-25
config: variable names and outputs must be unique

### LINKED ISSUE #5052: config: Report duplicate `output` definitions as invalid
The following is currently considered as valid config:

```hcl
variable "var1" {
	default = "Yoda"
}
variable "var2" {
	default = "Pug"
}

output "yada" {
	value = "${var.var1}"
}
output "yada" {
	value = "${var.var2}"
}
```

```
$ terraform apply
```
```

Outputs:

  yada = Yoda
```

I believe we should be returning error in these cases as outputs are saved as key/value map - i.e. there will always be only 1 unique value returned for a single key.

### LINKED ISSUE #5144: Mixing variable definition and specification is order specific
From: https://www.terraform.io/docs/configuration/load.html

<blockquote>
The order of variables, resources, etc. defined within the configuration doesn't matter. Terraform configurations are declarative, so references to other resources and variables do not depend on the order they're defined.
</blockquote>

However, it appears that a empty (forward) declaration that occurs after a value declaration unsets the value.

To wit, this works:

````
variable "test" {}
variable "test" { default = "5" }
````

But swap the order, and it queries for a value for "test":

````
variable "test" { default = "5" }
variable "test" {}
````

Expected Behaviour: A late forward declaration should be silently ignored if there is already a value set for that variable in the symbol table.

mitchellh:
Fixes #5052 
Fixes #5144

This adds new config validation to verify output names and variable names are unique.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -586,43 +586,55 @@ func (c *Config) Validate() error {
 	}
 
 	// Check that all outputs are valid
-	for _, o := range c.Outputs {
-		var invalidKeys []string
-		valueKeyFound := false
-		for k := range o.RawConfig.Raw {
-			if k == "value" {
-				valueKeyFound = true
-				continue
-			}
-			if k == "sensitive" {
-				if sensitive, ok := o.RawConfig.config[k].(bool); ok {
-					if sensitive {
-						o.Sensitive = true
-					}
-					continue
-				}
-
+	{
+		found := make(map[string]struct{})
+		for _, o := range c.Outputs {
+			// Verify the output is new
+			if _, ok := found[o.Name]; ok {
 				errs = append(errs, fmt.Errorf(
-					"%s: value for 'sensitive' must be boolean",
+					"%s: duplicate output. output names must be unique.",
 					o.Name))
 				continue
 			}
-			invalidKeys = append(invalidKeys, k)
-		}
-		if len(invalidKeys) > 0 {
-			errs = append(errs, fmt.Errorf(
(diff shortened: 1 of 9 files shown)


darrensony:
Regarding #5144, and looking at your test code, this may or may not be the right solution.

Say I have two .tf files, A.tf and B.tf

*A.tf* sets the value of the variable
```variable "foo" { default = 1 }```

*B.tf* needs to use the value of "foo" but not define it.
```variable "foo" {}```

The core issue of #5144 was that the above worked fine, but if *A.tf* was renamed to *C.tf*, it no longer worked because Terraform reads files in alphabetical order and thus sees the reference before the declaration.

Throwing an error because there is both a declaration and a reference is not the correct behaviour.

mitchellh:
Unfortunately what you were doing before should've never worked. Terraform loads and merges all files together. Having a config in A.tf and B.tf is the same as concatenating B.tf to A.tf. Terraform currently happens to load files in alphabetical order which caused your behavior.

We never intended two variables to be defined, as they're all in the same namespace, so this is the correct behavior.

Actually, to fight against this in the future, we should probably load all the files in a _random_ order in each load. The declarative nature of Terraform means the load order shouldn't matter.

mitchellh:
@jen20 brought up a good point in our Slack channel that I need to test how this interacts with _override_ files. I'm going to add test cases for this since its currently an unknown.

jen20:
LGTM pending Travis with the new test.

PR #8485, opened 2016-08-25
config: JSON resource keys with only one item load properly GH-5140

### LINKED ISSUE #5140: Problem with aws_security_group in JSON format
Not working

```
{
  "resource" : {
    "aws_security_group" : {
      "allow_external_http_https" : {
        "tags" : {
          "Name" : "allow_external_http_https"
        }
      }
    }
  }
}
```

```
Error loading config: Error loading test/vpc.tf: position -: resource must be followed by exactly two strings, a type and a name
```

Whereas this works

```
{
  "resource" : {
    "aws_security_group" : {
      "allow_external_http_https" : {
        "name" : "allow_external_http_https",
        "tags" : {
          "Name" : "allow_external_http_https"
        }
      }
    }
  }
}
```

mitchellh:
Fixes #5140 

When a resource has only a single key set, the HCL parser treats that
key as part of the overall set of object keys. This isn't valid since
we expect resources to have exactly two keys. In this scenario, we have
to "unwrap" the keys back into a set of objects.

Details are in the comment in the changeset.

### DIFF
--- a/config/loader_hcl.go
+++ b/config/loader_hcl.go
@@ -555,6 +555,37 @@ func loadManagedResourcesHcl(list *ast.ObjectList) ([]*Resource, error) {
 				item.Pos())
 		}
 
+		// HCL special case: if we're parsing JSON then directly nested
+		// items will show up as additional "keys". We need to unwrap them
+		// since we expect only two keys. Example:
+		//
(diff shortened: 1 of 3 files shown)


jbardin:
LGTM

PR #8486, opened 2016-08-25
provider/aws: Cleanup the Lambda ENI deletion process a bit

radeksimko:
Just a little bit of cleanup after merging https://github.com/hashicorp/terraform/pull/8033 to explain why do we have the function there + to catch errors.

cc @gposton

### DIFF
--- a/builtin/providers/aws/resource_aws_security_group.go
+++ b/builtin/providers/aws/resource_aws_security_group.go
@@ -359,7 +359,9 @@ func resourceAwsSecurityGroupDelete(d *schema.ResourceData, meta interface{}) er
 
 	log.Printf("[DEBUG] Security Group destroy: %v", d.Id())
 
-	deleteLingeringLambdaENIs(conn, d)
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_security_group.go
```diff
@@ -359,7 +359,10 @@ func resourceAwsSecurityGroupDelete(d *schema.ResourceData, meta interface{}) er
 
 	log.Printf("[DEBUG] Security Group destroy: %v", d.Id())
 
-	deleteLingeringLambdaENIs(conn, d)
+	err := deleteLingeringLambdaENIs(conn, d)
+	if err != nil {
```
mitchellh: Since only the err is used here can we use the `if err := ...; err != nil {` style here?

mitchellh:
One nit, then LGTM

radeksimko:
Nit fixed, merging.

PR #8487, opened 2016-08-25
aws_iam_instance_profile: Restrict example perms

sethp-jive:
In this example, principal "AWS": "*" tells IAM that any user in the world can assume that role as long as they know the account ID + role name. 

In my testing, "Service": "ec2.amazonaws.com" is sufficiently permissive to allow the instance_profile to work without allowing global accesses.


mitchellh:
LGTM!

PR #8496, opened 2016-08-26
command: remove value restriction when overriding

### LINKED ISSUE #8477: Terraform v0.7.2 mishandles command line variables with leading slash
Hi there,

Starting with v0.7, Terraform has problems with assigning variables from the command line using the ```-var``` construct. Specifically, it mangles strings that start with a "/" such as a UNIX file path.

### Terraform Version
Terraform v0.7.2

### Affected Resource(s)
- Terraform core

```hcl
variable    test1           {}
variable    test2           {}
variable    test3           {}
variable    test4           {}

provider "aws"
    {
    region                  = "us-west-2"
    profile                 = "roma-provisioner"
    }

resource "aws_security_group" "sg"
    {
    name        = "Test"
    description = "Test"

    #   Allow inbound HTTPS access.
    ingress
        {
        from_port   = 443
        to_port     = 443
        protocol    = "tcp"
        cidr_blocks = [ ]
        }

    tags
        {
        Name        = "test"
        Test1       = "${var.test1}"
        Test2       = "${var.test2}"
        Test3       = "${var.test3}"
        Test4       = "${var.test4}"
        }
    }
```

### Expected Behavior
When the value of a variable specified via `-var` on the command line  contains a leading slash, the value is not assigned as expected inside Terraform.

### Actual Behavior
The variable's value inside Terraform consists of an empty string.

### Steps to Reproduce
I have created a relatively simple nonsense Terraform configuration to illustrate the behavior

This configuration builds an AWS security group and sets certain tags simply to illustrate the problem. Using the configuration shown above, I run the command:
`TF_VAR_test2=/path/2 ./terraform plan -input=true -var test3=/path/3 -var test4=\"/path/4\"`

The output (trimmed of extraneous details in the interest of brevity) shows:
```hcl
+ aws_security_group.sg
    owner_id:                            "<computed>"
    tags.%:                              "5"
    tags.Name:                           "test"
    tags.Test1:                          "/path/1"
    tags.Test2:                          "/path/2"
    tags.Test3:                          ""
    tags.Test4:                          "/path/4"
```

The actual behavior illustrated above exhibits the following behavior.
1. The variable *test1* is set in terraform.tfvars (the contents are a single line reading ```test1 = "/path/1"```). This is successfully read and assigned to the variable inside Terraform, populating the tag as expected.
2. The variable *test2* which is set using the ```TF_VAR_test2``` environment variable also successfully assigns the variable inside Terraform, also populating the tag as expected.
3. The variable *test3* is set using the ```-var``` construct on the command line. This does **not** work as expected, and the variable inside Terraform is set to an empty string. **This is the problem case.**
4. The variable *test4** is also set via the ```-var``` construct on the command line, but it is quoted (using backslash escapes to prevent the UNIX shell from eating the quotes). This **does** set the variable inside Terraform as expected.

While I have several available workarounds, this is undocumented behavior – and I believe it undesirable as well.

### Important Factoids
* If the leading slash in the variable specified for test3 on the command line is removed, the value **is** assigned in Terraform. 
* This problem did not exist at v0.6.16. It exists with v0.7, v0.7.1, and v0.7.2.

### References
*None known.*

monstermunchkin:
Remove the restriction of not accepting certain strings as values when
overriding variables using the `-var` flag. This allows values to be
accepted even if they do not start with a letter, e.g. paths.

This fixes #8477.

### DIFF
--- a/command/flag_kv.go
+++ b/command/flag_kv.go
@@ -158,9 +158,10 @@ func parseVarFlagAsHCL(input string) (string, interface{}, error) {
 		return "", nil, fmt.Errorf("Cannot parse value for variable %s (%q) as valid HCL: %s", probablyName, input, err)
 	}
 
-	// Cover cases such as key=
+	// Cover cases in which the value is either empty or does not start with a
+	// letter, e.g. paths.
 	if len(decoded) == 0 {
-		return probablyName, "", nil
+		return probablyName, input[idx+1:], nil
 	}
 
 	if len(decoded) > 1 {
--- a/command/flag_kv_test.go
+++ b/command/flag_kv_test.go
@@ -2,10 +2,11 @@ package command
 
 import (
 	"flag"
-	"github.com/davecgh/go-spew/spew"
 	"io/ioutil"
 	"reflect"
 	"testing"
+
+	"github.com/davecgh/go-spew/spew"
 )
 
 func TestFlagStringKV_impl(t *testing.T) {
@@ -46,6 +47,12 @@ func TestFlagStringKV(t *testing.T) {
 			"key",
 			nil,
 			true,
+		},
+
+		{
+			"key=/path",
+			map[string]string{"key": "/path"},
+			false,
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


mitchellh:
This was actually a bug in the HCL layer, fix here: https://github.com/hashicorp/hcl/pull/146

Thanks for your fix as well but we'd like to fix it at the root! :)

Issue #8500, opened 2016-08-26
aws_autoscaling_group has no arn value


TaiSHiNet:
Hi, 

I'm on terraform 0.7.1, and when trying to build an IAM role policy to allow CodeDeploy target an autoscaling group, I can't reference the ASG ARN.
When I run describe-auto-scaling-groups on awscli I do get the value.
Can this be added as an attribute to aws_autoscaling_group ?

Bests,
Juan

catsby:
Simple enough! I added that in #8503

TaiSHiNet:
@catsby you're the fairy tale of tickets!

stack72:
Closed via #8503

PR #8503, opened 2016-08-26
provider/aws: Get and export ASG ARN value

catsby:
Adds the ARN value for the AutoScaling group to its values.

Fixes #https://github.com/hashicorp/terraform/issues/8500


stack72:
LGTM!

PR #8507, opened 2016-08-27
Update installation guide

nmcalabroso:
Mention that there's only a single binary file in package as of 0.7.2


apparentlymart:
Thanks @nmcalabroso!

PR #8508, opened 2016-08-27
schema: Return all validation errors together from InternalValidate

radeksimko:
This can be helpful for any developers making changes in existing providers/resources/data sources or creating new ones.

Previously we were returning the first validation error which lead to the (IMO boring & slow) workflow _"fix, run tests again, rinse & repeat"_.

Now the developer should be able to fix all errors at once.


mitchellh:
Agreed. Thanks!

PR #8465, opened 2016-08-25
provider/aws: Increase route_table timeouts

carlossg:
It may take more than 1 minute

```
21:01:58 [vpc] aws_route_table.public: Refreshing state... (ID: rtb-d202deb4)
21:01:58 [vpc] aws_route_table_association.jce-public: Refreshing state... (ID: rtbassoc-248f4a5d)
21:02:00 [vpc] aws_route_table_association.jce-public: Destroying...
21:02:00 [vpc] aws_route_table_association.jce-public: Destruction complete
21:02:00 [vpc] aws_route_table.public: Destroying...
21:02:12 [vpc] aws_route_table.public: Still destroying... (10s elapsed)
21:02:22 [vpc] aws_route_table.public: Still destroying... (20s elapsed)
21:02:32 [vpc] aws_route_table.public: Still destroying... (30s elapsed)
21:02:40 [vpc] aws_route_table.public: Still destroying... (40s elapsed)
21:02:50 [vpc] aws_route_table.public: Still destroying... (50s elapsed)
21:03:00 [vpc] aws_route_table.public: Still destroying... (1m0s elapsed)
21:03:01 [vpc] Error applying plan:
21:03:01 [vpc] 
21:03:01 [vpc] 1 error(s) occurred:
21:03:01 [vpc] 
21:03:01 [vpc] * aws_route_table.public: Error waiting for route table (rtb-d202deb4) to become destroyed: timeout while waiting for state to become ''. last error: %!s(<nil>)
```

PR #8510, opened 2016-08-27
helper/resource: Show last state in timeout err message

radeksimko:
This PR is fixing a couple of existing tests that could easily report false positives and then adding a bit extra information into commonly seen error message to help the user to understand what the problem is.

Related: https://github.com/hashicorp/terraform/pull/8465

### DIFF
--- a/helper/resource/error.go
+++ b/helper/resource/error.go
@@ -40,15 +40,26 @@ func (e *UnexpectedStateError) Error() string {
 // TimeoutError is returned when WaitForState times out
 type TimeoutError struct {
 	LastError     error
+	LastState     string
 	ExpectedState []string
 }
 
 func (e *TimeoutError) Error() string {
-	if e.LastError != nil {
-		return fmt.Sprintf("timeout while waiting for state to become '%s': %s",
-			strings.Join(e.ExpectedState, ", "), e.LastError)
+	expectedState := "resource to be gone"
+	if len(e.ExpectedState) > 0 {
+		expectedState = fmt.Sprintf("state to become '%s'", strings.Join(e.ExpectedState, ", "))
 	}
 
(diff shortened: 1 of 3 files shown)


helper/resource/error.go
```diff
@@ -40,15 +40,26 @@ func (e *UnexpectedStateError) Error() string {
...
 	if e.LastError != nil {
-		return fmt.Sprintf("timeout while waiting for state to become '%s': %s",
-			strings.Join(e.ExpectedState, ", "), e.LastError)
+		return fmt.Sprintf("timeout while waiting for %s%s: %s",
+			expectedState, lastState, e.LastError)
 	}
 
-	return fmt.Sprintf("timeout while waiting for state to become '%s'",
-		strings.Join(e.ExpectedState, ", "))
+	return fmt.Sprintf("timeout while waiting for %s%s",
```
stack72: Nit pick - should be a space between the %s values

helper/resource/error.go
```diff
@@ -40,15 +40,26 @@ func (e *UnexpectedStateError) Error() string {
...
+
+	lastState := ""
+	if e.LastState != "" {
+		lastState = fmt.Sprintf(" (last state: '%s')", e.LastState)
+	}
+
 	if e.LastError != nil {
-		return fmt.Sprintf("timeout while waiting for state to become '%s': %s",
-			strings.Join(e.ExpectedState, ", "), e.LastError)
+		return fmt.Sprintf("timeout while waiting for %s%s: %s",
```
stack72: Nit pick - space required between %s values

radeksimko:
@stack72 The `%s%s` is actually an intention to prevent the trailing space. See line 55 where there's an extra leading space. 😉

stack72:
@radeksimko ah! in that case, ignore me :)

PR #8544, opened 2016-08-29
Fix error from copy/paste of another example

SeanKnight:
It looks like the Elasticsearch example was copied from the Redshift example, but the "destination" field wasn't updated.


apparentlymart:
Thanks, @SeanKnight!

Issue #2811, opened 2015-07-22
Wait for SSH Connectivity before local-exec (Ansible)


IanVS:
I am using Ansible for provisioning by kicking off `ansible-playbook ...` as a local-exec script.  

Similar to https://github.com/hashicorp/terraform/issues/2137, I find that the local-exec is executed as soon as an EC2 instance is created, but before SSH connectivity is established, and SSH times out.

What is the right way to handle this case?  Do I just need to put a long ssh timeout in Ansible?  That seems brittle.  Is it not possible to delay execution of `local-exec` provisioners until SSH is ready?  There must be some kind of logic for that already to allow `remote-exec` scripts to work.

mlrobinson:
I don't know of a terraform way of solving this problem, but there is an Ansible module to help you with it:

http://docs.ansible.com/ansible/wait_for_module.html

That (ran as a local action) would allow ansible to start immediately, but then wait until you saw the port become available. Not a perfect solution, but it should work for you for now.

IanVS:
Thanks, I'll give that a shot.  Still would be a little more elegant if there was a way to handle in Terraform...

IanVS:
Actually I have a question about this, since you seem to know what you're talking about, @mlrobinson.  I'm using roles to set up my boxes, and my top level playbook is empty except for specifying a role.  So, would this `wait_for` task need to go into my `common` role which is the base dependency of my roles?  I think if I put it in the playbook, it would be executed after all the roles are set up, correct?  Thanks again for the advice.

mlrobinson:
That depends on your role setup of course. If all your roles include the `common` role first, then it should run before everything else. Another option would be to make this it's own, very simple role, and just define that before even common. Either option would be viable, but I'd probably go with the easier of the two, depending on how your playbooks/roles are setup.

But, according to http://docs.ansible.com/ansible/playbooks_roles.html you are right that a "loose" task in the same playbook as roles is executed after the roles have been finished. Someone online created a "command role" that basically allowed you to pass a string into the role definition so it would occur in order, but still support defining one off commands at the playbook level. Here's one example of it: https://github.com/RoleModel/bash_command

mlrobinson:
Not to double post, but ignore me about the command role. That won't work exactly in your situation, though something like it is what I was thinking of: a role that allowed you to pass in the name of another module to run. Such a meta thing would probably not be worth it in the run, but it would be a neat solution to this and other things with Ansible.

IanVS:
[...]

Issue #2137, opened 2015-05-29
Wait for ssh connectivity in local-exec


pastafari:
Hi,

We're using chef-solo as our provisioning tool, and I wanted to know the best way to integrate it with a terraform provision block. Here's what I need to do:

1. Spin up a digital ocean droplet
2. Bootstrap it (lets say I want to call bootstrap.sh)
3. Provision it (lets say I want to call provision.sh)

I have local scripts that do 2 and 3 given a remote IP, key and run list. 
I added both scripts to a `local-exec` block like so:

```
provisioner 'local-exec' {
  command = "provision.sh && bootstrap.sh" 
}
```
Now, what happens is that the digital ocean droplet comes up, but OpenSSH takes some time to start, so when the provisioner gets triggered, my bootstrap script is unable to connect to the droplet. 

One way around this would be to add a sleep-until-i-can-connect loop in the bootstrap script. Another way would be to delegate that to terraform, like `remote-exec` does (correct me if I'm wrong!) via its connection block. 

What would you recommend?

stephenchu:
If you are not inside of a VPC (or bastion host-like environment, i.e. no ssh hopping), you could use a `remote-exec` (which has retry logic on connection refused) as the first provisioner that does nothing to do it.

```
provisioner "remote-exec" {
  inline = ["# Connected!"]
}
```

pastafari:
@stephenchu  that is a great idea. Thanks! :+1: 

Essentially, I'm doing a remote-exec via a local script. 

I was wondering if this is a common enough use case to warrant it being first class in local-exec.

pastafari:
Closing this as its not really an issue. Thanks again @stephenchu for the idea.

PR #8549, opened 2016-08-30
local-exec: the OS doesn't need to be ready

jakubholynet:
It is not obvious that the resource being created doesn't mean that the OS and system services such as sshd are ready (contrary to `remote-exec`). It is better to make that explicit and same developers like me some headache :-)

(See #2811, #2137)


jen20:
Hi @jakubholynet! This is a nice addition, thanks for contributing! I'm going reword it slightly as `local_exec` provisioners are often used on things other than compute resources, and merge this. Thanks again.

jen20:
Hi @jakubholynet! I've rebased this and merged it as 8d05408 and clarified the comment to scope to compute resources d31656a. Thanks for the contribution!

jakubholynet:
Thanks a lot!

PR #8553, opened 2016-08-30
Add note on cd'ing into cloned repo

manojlds:
Simple change, but making it clear where to run `make` from.


stack72:
Thanks @manojlds 

This LGTM!

P.

PR #8557, opened 2016-08-30
Fix inconsistent results with self interpolation

jbardin:
Due to a race in interpolateForce(), a reference to self could return
inconsistent results.

### DIFF
--- a/config/raw_config.go
+++ b/config/raw_config.go
@@ -48,6 +48,18 @@ func NewRawConfig(raw map[string]interface{}) (*RawConfig, error) {
 	return result, nil
 }
 
+// RawMap returns a copy of the RawConfig.Raw map.
+func (r *RawConfig) RawMap() map[string]interface{} {
+	r.lock.Lock()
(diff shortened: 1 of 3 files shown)


config/raw_config.go
```diff
@@ -48,6 +48,18 @@ func NewRawConfig(raw map[string]interface{}) (*RawConfig, error) {
...
+func (r *RawConfig) RawMap() map[string]interface{} {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	m := make(map[string]interface{})
+	for k, v := range r.Raw {
+		m[k] = v
+	}
+	return m
+}
```
mitchellh: As annoying as it is, I would add a test for `RawMap` that ensures:

  1. It points to a different map

  1. The contents (key/values) are `reflect.DeepEqual` to the map itself

mitchellh:
One test I'd add then LGTM

jbardin:
Added test

Issue #8527, opened 2016-08-29
count is not implemented in data sources


rchalumeau:
Hi,

### Terraform Version
Terraform v0.7.2

### Affected Resource(s)
Related to data sources implementation 
error noticed with  
- data.template_file
- data.null_data_source

### Terraform Configuration Files
```hcl
data "template_file" "check_connection_to_jboss" {
    count = "2"
    template = "${file("${path.root}/files/consul/check_connection_to_jboss.json.tpl")}"
    vars {
        aps = "${element(azurerm_virtual_machine.aps.*.name, count.index)}"
    }
}
```

### Expected Behavior
I need to generate specific consul check scripts for each created node. 
To do that, I have looped on the VMs count and pass them as vars to rendered templates. 
It was working fine with terraform 0.6.16. To migrate in 7.2, I have transformed 
```hcl
resource "template_file" "check_connection_to_jboss" {
```
into 
```hcl
data "template_file" "check_connection_to_jboss" {
```

### Actual Behavior
I got the message
  * data.template_file.check_connection_to_jboss #1: : invalid or unknown key: count
  * data.template_file.check_connection_to_jboss #0: : invalid or unknown key: count

I cannot see in the documentation reference to supported meta-parameters for data sources. 
I have found the same limitation with null_data_source. 
So, I would assume that this affects any data source. 

### Steps to Reproduce
terraform plan

rchalumeau:
Hi, 

I confirm it is solved. 
Thanks for the fix

Romain

PR #8566, opened 2016-08-31
Template are now data sources

MiLk:
Since Terraform 0.7.0 [Templates](https://www.terraform.io/docs/providers/template/) are now data sources.

The interpolation documentation page is using a deprecated sample of code.

### DIFF
--- a/website/source/docs/configuration/interpolation.html.md
+++ b/website/source/docs/configuration/interpolation.html.md
@@ -246,12 +246,12 @@ The supported built-in functions are:
 
 ## Templates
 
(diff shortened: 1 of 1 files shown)


apparentlymart:
Thanks for this, @MiLk!

The part of this about `count` won't actually work this way right now because of #8527, so I think we should pause on updating this until we get that fixed. The `resource` form of templates does still work right now, it's just deprecated... so I think having an example that works and generates deprecation warnings is better than one that doesn't work at all.

Hopefully #8527 can be fixed before too long and this won't be a problem anymore.

MiLk:
Right, let's wait for #8527.

stack72:
Folks, #8527 is now resolved! So merging this :)

Thanks for the work here @MiLk

PR #8567, opened 2016-08-31
provider/fastly: Change error text on findService

premist:
This PR changes error message on `findService` function on [resource_fastly_service_v1.go](https://github.com/hashicorp/terraform/blob/master/builtin/providers/fastly/resource_fastly_service_v1.go) file.

I was performing `terraform plan` locally and got this error, like this:

```
* fastly_service_v1.REDACTED: [WARN] Error listing services when deleting Fastly Service (REDACTED): 401 - Unauthorized
Message: Provided credentials are missing or invalid
```

I was misinformed by the message, because I thought Terraform actually tried to delete my (production) services. I looked into resource_fastly_service_v1.go and found out that it's an error message printed at `findService` function, which doesn't do any destructive action by itself.


mitchellh:
LGTM!

PR #8568, opened 2016-08-31
Update interpolation.html.md

zhuk-intetics:


stack72:
Hi @zhuk-intetics 

Thanks for this - this LGTM :)

Paul

Issue #8537, opened 2016-08-29
Run `errcheck` in Travis


dtolnay:
[`errcheck`](https://github.com/kisielk/errcheck) is a static analyzer that detects calls to functions that return an error but the error is not checked.

``` go
func f() error { /* ... */ }

f() // <- does not check the error
```

dtolnay:
This would have prevented the longstanding bug in #7448 as well as many many many other bugs.

catsby:
Thanks @dtolnay ! I'll look into integrating `errcheck` into the project

pkolyvas:
We no longer use Travis for better or worse. Closing.

PR #8569, opened 2016-08-31
Fix errcheck issues outside of tests and /builtin

dtolnay:
With this commit, the following command reports no issues:

```
errcheck -ignoretests ./... | grep -v ^vendor | grep -v ^builtin
```

Most of these are mundane but some are scary or embarrassing, like in flag_kv.go.

Progress toward #8537. @catsby


catsby:
Nice! The `bytesnoerror` package looks like a good patch to our current usage of `buf.WriteString` without any error checking :smile:

It's a big patch so please give us a bit to review it all, but at first cursory glance it looks good 👍

catsby:
Hey @dtolnay – I'm curious what you think about simply ignoring the errors from the `bytes` package, either entirely or specific methods, instead of introducing this new code?

Something like:

`$ errcheck -ignore "bytes:[rR]ead|[wW]rite|<other patterns>"`

I'm working on a branch to scan the AWS builtin provider in our TravisCI runs as a start to introducing `errcheck` more broadly in the repo, and at the moment I'm using the `-ignore` flag like so:

```
-ignoretests -ignore 'github.com/hashicorp/terraform/helper/schema:Set,bytes:.*,os:.*,io:.*'
```

jbardin:
Hi @dtolnay,

Thanks for the effort you put into this. I'm not sure yet if this is the full path we want to take. Adding named returns and extra closures to check for errors that either can't happen or we can't handle. 

Take os.File.Close for example: there's only 3 POSIX errors that it could return, `EBADF` (and Go inserts a possible `EINVAL` here too if the os.File wasn't valid to begin with) which we wouldn't see if the Open succeeded and we completed a successful Read/Write on the file, `EINTR` which we can't get because the runtime uses `SA_RESTART`, and `EIO` which could only be a remnant from an unchecked write error. 

Named returns and assignments in deferred closures can lead to hard to reason about bugs themselves, and I tend to try and avoid those whenever possible. 

While there are some errors we could do a better job checking, I think @catsby's suggestion of making a curated list of functions that are ok to skip is a good pragmatic choice.

dtolnay:
@catsby I am on board with ignoring specific errors from the `bytes` package in our invocation of errcheck, but I am not comfortable with your extremely broad whitelist. For example `bytes.Buffer.ReadFrom(io.Reader)` needs to be checked. I also don't see a reason to whitelist `os.*` and `io.*` other than as a temporary measure because we don't feel like dealing with them right away. I will update the PR to revert the specific `bytes` changes that I think don't need to be checked.

@jbardin I disagree that for functions that return errors, "I don't think it will happen" means "don't check." Instead "I don't think it will happen" means "panic if it happens." In the `EIO` case I would rather know I failed to handle a write error correctly, instead of pretending the rest of the codebase outside of my `f.Close()` call is flawless. I disagree that the solution is to ignore more errors. I think the solution is to use helper functions that reduce the amount of logic and visual overhead. For example:

```go
// bad
defer f.Close()

// bad
defer func() {
    err2 := f.Close()
    if err == nil {
        err = err2
    }
}()

// less bad
defer errsafe.Close(f, &err)

// package errsafe
func Close(f io.Closer, err *error) {
    err2 := f.Close()
    if *err == nil && err2 != nil {
        panic(err2) // unexpected
    }
}
```

@jbardin let me know what you think and I will update the PR.

jbardin:
@dtolnay, Yes, maybe File.Close wasn't the most clear-cut example I could have used, but it is generally accepted practice to ignore it's error unless there is a specific case you're looking for. 

I agree the "less bad" solution looks better, but has some of the same drawbacks as the closure, primarily that the `err` value is often shadowed throughout code, and named returns and deferred assignments are easily overlooked and often hard to quickly reason about in large functions. 

I personally don't want blanket ignore rules for errcheck if we intend to use it (for example, ignoring `fmt:.*` ignores `fmt.Fprintf`, which you usually _do_ want to check). What I would like is a curated list of matches, each with a minimal scope that we use for the tests. That list might be quite long, but then we're forced to make the decision about each function rather than having it fall under a blanket statement. This will let our code stay more idiomatic, while still making use of static analysis tools to detect programmer errors. 

I think there's a good middle ground we can find without resorting to workarounds for deferred errors that can be otherwise ignored.

dtolnay:
@jbardin I don't follow your arguments against `defer errsafe.Close(f, &err)`. It is not assigning to the named return so that is not an issue. Also there is practically always a variable called `err` in the top-level block of functions that return an error, so I don't see the big difference between declaring it at the top vs declaring it with `:=` the first time it is assigned.

jbardin:
@dtolnay due to the style of most Go code, in any function of sufficient length a `var err error` declared at the top (either in the body, or as a named return) will get shadowed. This means that the `err` seen in the `errsafe.Close` function will often be nil where it shouldn't causing a `Close()` error to panic. That may be fine in many cases, but it should then be done explicitly where we absolutely know we should crash, e.g.

```
    defer func() {
        if err := f.Close(); err != nil {
            panic(err)
        }
    }
```

Otherwise this risks a crash during a run by a user, possibly for a minor error that may not even be presented to the user outside of the terraform logs. It would take a very thorough auditing of the code to ensure this couldn't cause spurious crashes if these were dropped in for every deferred Close. Maybe a trio of helpers with one that explicitly panics, and one that logs errors would help, but I think that might be a discussion better suited for an isolated PR rather than trying to hash through all possibilities here.

dtolnay:
Makes sense, thanks for the explanation. I will update the PR to revert changes to `Close` calls.

dtolnay:
I have updated the PR to remove error checking for `os.File.Close()` as well as a few `bytes` methods. Here is the whitelist:

```
errcheck \
    -ignoretests \
    -ignore 'bytes:Write|WriteString|WriteRune' \
    -ignore 'os:Close' \
    ./... \
| grep -v ^vendor \
| grep -v ^builtin
```

josephholsten:
@dtolnay some really good stuff in here, but rebasing is a nightmare. I'm making an attempt though.

josephholsten:
Seriously, did you write all of this by hand? If there's a tool to do this, it may be faster.

josephholsten:
Well, I rebased it but there's still 69 warnings. So I guess there's still room for improvement!

teamterraform:
Hi folks! We're undertaking a general github issue/PR cleanup. Most of the code in this PR has changed in terraform 0.12, so I will close this PR.

PR #8579, opened 2016-08-31
provider/aws: Run errcheck in tests

catsby:
This is part of a greater strategy to integrate [errcheck](https://github.com/kisielk/errcheck) into Terraform (see https://github.com/hashicorp/terraform/issues/8537). Here we add a new Make task `errcheck`, which specifically tests for unchecked errors in the AWS package. 

Specifically, we do the following:

- run errcheck on each PR as part of `make test`
- Check the AWS builtin provider:
  - ignoring test files with `-ignoretests` 
  - Ignoring:
    - `github.com/hashicorp/terraform/helper/schema:Set`  
    - `ignore 'bytes:.*'`  
    - `ignore 'io:[cC]lose|[wWrite]'`  

We use `d.Set` through the codebase of providers heavily, rarely checking for errors, the exceptions typically for when we set complex structures (anything _not_ a string, int, etc). Other errors we don't check are things like `buf.WriteString`, etc.

This is meant to be a step in the right direction, as opposed to be a comprehensive addition of `errcheck`

------
Update: It remains to be seen if we want to rename the make task to be more clear that we're only testing AWS, and add more, similar tasks in the future, or leave as-is and gradually expand the scope of packages it runs the checks on.

### DIFF
--- a/scripts/errcheck.sh
+++ b/scripts/errcheck.sh
@@ -12,7 +12,7 @@ fi
 err_files=$(errcheck -ignoretests -ignore \
   'github.com/hashicorp/terraform/helper/schema:Set' \
   -ignore 'bytes:.*' \
-  -ignore 'io:.*' \
+  -ignore 'io:[cC]lose|[wWrite]' \
   ./builtin/providers/aws/...)
 
 if [[ -n ${err_files} ]]; then
--- a/scripts/errcheck.sh
+++ b/scripts/errcheck.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+# Check gofmt
+echo "==> Checking AWS provider for unchecked errors..."
+echo "==> NOTE: at this time we only look for uncheck errors in the AWS package"
+
+if ! which errcheck > /dev/null; then
+    echo "==> Installing errcheck..."
+    go get -u github.com/kisielk/errcheck
+fi
+
+err_files=$(errcheck -ignore 'github.com/hashicorp/terraform/helper/schema:Set,bytes:.*' ./builtin/providers/aws/...)
+if [[ -n ${err_files} ]]; then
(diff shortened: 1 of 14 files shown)


scripts/errcheck.sh
```diff
@@ -0,0 +1,20 @@
...
+# Check gofmt
+echo "==> Checking AWS provider for unchecked errors..."
+echo "==> NOTE: at this time we only look for uncheck errors in the AWS package"
+
+if ! which errcheck > /dev/null; then
+    echo "==> Installing errcheck..."
+    go get -u github.com/kisielk/errcheck
+fi
+
+err_files=$(errcheck -ignoretests -ignore 'github.com/hashicorp/terraform/helper/schema:Set,bytes:.*,os:.*,io:.*' ./builtin/providers/aws/...)
```
mitchellh: Can this be split across multiple lines (maybe with `\`?) so that its easy to modify this going forward?

Can we use multiple `-ignore` flags instead? (not sure if possible from errcheck)

scripts/errcheck.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if ! which errcheck > /dev/null; then
+    echo "==> Installing errcheck..."
+    go get -u github.com/kisielk/errcheck
+fi
+
+err_files=$(errcheck -ignoretests -ignore \
+  'github.com/hashicorp/terraform/helper/schema:Set' \
+  -ignore 'bytes:.*' \
+  -ignore 'io:[cC]lose|[wWrite]' \
```
jbardin: Do we need to match the lowercase `c` or `w` for anything? Also you have brackets around the entire `[wWrite]`, matching any single one of those characters.

jbardin:
I think this gives a good starting point to expand the checks as we go. 
LGTM

PR #8581, opened 2016-08-31
terraform: self.count works in interpolations [GH-5283]

### LINKED ISSUE #5283: Expanded resources do not have `count` attribute
Terraform v0.6.11

Steps the reproduce:

```hcl
  count = 3
  # ...

  provisioner "remote-exec" {
    inline = "echo ${self.count}"
  }
}
```

Note, this is different than `count.index`, I want to get the value supplied to the resource for `count`, which should be "3" in this example. This works for the first resource (0), but not 1 and 2. Terraform complains that `aws_instance.web.1` does not define a count attribute, but `aws_instance.web.0` does.

This could be that the pre-expanded resource has `count`, but the post-dynamicexpand resources do not.

mitchellh:
Fixes #5283 

This enables the following to work:

```
resource "null_resource" "web" {
  count = 3
  # ...

  provisioner "local-exec" {
    command = "echo ${self.count}"
  }
}
```

Potentially even cooler and as a side effect (that we now cover by tests) you can now reference counts of other resources:

```
resource "null_resource" "web" {
  count = 3
}

output "count" { value = "${null_resource.web.count}" }
```

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -149,6 +149,39 @@ module.test:
 	}
 }
 
+func TestContext2Apply_refCount(t *testing.T) {
+	m := testModule(t, "apply-ref-count")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+	ctx := testContext2(t, &ContextOpts{
+		Module: m,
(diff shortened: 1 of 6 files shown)


sethvargo:
LGTM

PR #8585, opened 2016-08-31
helper/schema: Add diff suppression callback

jen20:
This commit adds a new callback, `DiffSuppressFunc`, to  the `schema.Schema` structure. If set for a given schema, a callback to the user-supplied function will be made for each attribute for which the default type-based diff mechanism produces an attribute diff. Returning `true` from the callback will suppress the diff (i.e. pretend there was no diff), and returning `false` will retain it as part of the plan.

There are a number of motivating examples for this:

1. On SSH public keys, trailing whitespace does not matter in many cases - and in some cases it is added by provider APIs. For `digitalocean_ssh_key` resources we previously had a `StateFunc` that trimmed the whitespace - we now have a `DiffSuppressFunc` which verifies whether the trimmed strings are equivalent. This is included in the commit as an example usage.

2. IAM policy equivalence for AWS. A good proportion of AWS issues relate to IAM policies which have been "normalized" (used loosely) by the IAM API endpoints. This can make the JSON strings differ from those generated by iam_policy_document resources or template files, even though the semantics are the same (for example, reordering of `bucket-prefix/` and `bucket-prefix/*` in an S3 bucket policy. DiffSupressFunc can be used to test for semantic equivalence rather than pure text equivalence, but without having to deal with the complexity associated with a full "provider-land" diff implementation without helper/schema.

### DIFF
--- a/builtin/providers/digitalocean/resource_digitalocean_ssh_key.go
+++ b/builtin/providers/digitalocean/resource_digitalocean_ssh_key.go
@@ -21,31 +21,33 @@ func resourceDigitalOceanSSHKey() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"id": &schema.Schema{
+			"id": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
 
-			"public_key": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
-				StateFunc: func(val interface{}) string {
-					return strings.TrimSpace(val.(string))
-				},
+			"public_key": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: resourceDigitalOceanSSHKeyPublicKeyDiffSuppress,
 			},
 
-			"fingerprint": &schema.Schema{
+			"fingerprint": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 		},
 	}
+}
+
+func resourceDigitalOceanSSHKeyPublicKeyDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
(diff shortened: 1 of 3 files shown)


apparentlymart:
Does this solve problems that a normalizing `StateFunc` can't solve? The examples you gave here both seem solvable by `StateFunc`, with the latter one having a patch waiting in #8350.

jen20:
@apparentlymart The first is (and was) solvable by StateFunc - but for IAM policies this is not possible with StateFunc alone. The problem here is that some services normalise the policies differently to others - for example in S3 it seems to be a crapshoot as to whether your `prefix/` and `prefix/*` rules get reordered or not - and in some cases strings come back as single element arrays or vice versa.

Consequently you need a way of genuinely testing for equivalence (that is coming in a separate pull request because it involves a ton of helper code) such that:

```
...
"Resources": ["sts:AssumeRole"],
...
```
can be treated as equivalent to:

```
...
"Resources": "sts:AssumeRole",
...

in either direction - along with ordering (e.g. some endpoints sort with Version at the top, some at the bottom). Hopefully this explains the motivation better?

An improvement to the Digital Ocean SSH key resource here would be to parse the OpenSSH format and generate `ssh.PublicKey` values from it and compare the component values - that is left as a future improvement, however!

File: builtin/providers/digitalocean/resource_digitalocean_ssh_key.go [hunk 1]
```diff
@@ -21,33 +21,38 @@ func resourceDigitalOceanSSHKey() *schema.Resource {
...
 			},
 		},
 	}
 }
 
+func resourceDigitalOceanSSHKeyPublicKeyDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	if strings.TrimSpace(old) == strings.TrimSpace(new) {
+		return true
+	}
+	return false
```
mitchellh: can just `return strings.TrimSpace(old) == strings.TrimSpace(new)`

jen20: :+1: good point.

PR #8586, opened 2016-08-31
adding locks

jbardin:
I don't see an easy way out of the concurrency issues without a major refactor. This PR aims to be an intermediate step which will add locks around all concurrently accessed data structures used in the graph walks. This isn't necessarily going to make managing the current code easier, but it will hopefully make it more correct. 

The first round of changes will add locks to all exported methods of the various State and Diff structs. If these methods need to be called while holding a lock, we delegate the functionality to a matching private method that does not acquire the mutex. This will provide a pattern to follow, so that we can always ensure that all access is done through a mutex.

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -755,10 +755,19 @@ func (r *RemoteState) deepcopy() *RemoteState {
 }
 
 func (r *RemoteState) Empty() bool {
-	return r == nil || r.Type == ""
+	if r == nil {
+		return true
+	}
+	r.Lock()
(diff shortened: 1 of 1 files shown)


jbardin:
I think this is pretty safe to merge as-is now.
I don't want this PR to get too big, and the terraform state types won't really benefit from more setters and getters. 

This PR and #8889 complete the outstanding obvious race issues. More refactoring can come later.

mitchellh:
LGTM!

PR #8588, opened 2016-08-31
command/validate: Add flag to check that all variables are specified

dtolnay:
This PR adds a flag `terraform validate -check-vars` which checks that all variables declared in the configuration are specified in one of the possible ways:
- `-var foo=...`
- `-var-file=foo.vars`
- `TF_VAR_foo` environment variable
- `terraform.tfvars`
- default value

```
$ cat main.tf 
  resource "test_instance" "foo" {
      ami = "bar"

      network_interface {
        device_index = 0
        description = "Main network interface ${var.name}"
      }
  }

  variable "name" {}

$ terraform validate

$ terraform validate -check-vars
  fix these before continuing.

  Errors:

    * 1 error(s) occurred:

  * Required variable not set: name
```

### DIFF
--- a/command/validate.go
+++ b/command/validate.go
@@ -19,14 +19,23 @@ func (c *ValidateCommand) Help() string {
 	helpText := `
 Usage: terraform validate [options] [dir]
 
-  Validate the terraform files in a directory. If dir is not specified,
-  then the current directory will be used.
+  Validate the terraform files in a directory. Validation includes a
+  basic check of syntax as well as checking that all variables declared
+  in the configuration are specified in one of the possible ways:
+
+      -var foo=...
+      -var-file=foo.vars
+      TF_VAR_foo environment variable
+      terraform.tfvars
+      default value
+
+  If dir is not specified, then the current directory will be used.
 
 Options:
 
-  -check-vars         If specified, the command will check that all the
-                      variables without defaults in the configuration are
-                      specified.
+  -config-only        If specified, the command will check basic syntax of
+                      the config only. It will not check that required
+                      variables have been specified.
 
   -no-color           If specified, output won't contain any color.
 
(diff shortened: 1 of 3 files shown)


mitchellh:
I think this behavior should probably be inversed: this should be the default, and we should have a `-config-only` type that only does basic structure/syntax configuration on the config.

dtolnay:
I agree but was hesitant to make this a breaking change. I will update the PR to invert the behavior.

dtolnay:
I have updated the PR to invert the default behavior, replacing the opt-in flag `-check-vars` with an opt-out flag `-config-only`.

dtolnay:
@radeksimko @mitchellh is there anything I can do to help get this merged?

erickt:
@radeksimko @mitchellh: Has there been any progress on this? We're running into problems that this probably would solve. In our setup, we are using the s3 backend to store our state. We want to check that all our terraform scripts are correct, but we don't want to give our test runner access to our production s3 state, just to be safe. In terraform 0.8, we could use `terraform get` to check if we're missing variables, but this functionality no longer works in 0.9.

Consider this reduced situation. Say we have these two files:

`main.tf`:
```
module "foo" {
    source = "foo"
}
```

`foo/main.tf`:
```
variable "foo" {}
```

With terraform 0.8, we could discover the error with `terraform get`, which doesn't need access to the .tfstate:

```
➜  bug terraform -v
Terraform v0.8.8

Your version of Terraform is out of date! The latest version
is 0.9.3. You can update by downloading from www.terraform.io
➜  bug terraform validate
➜  bug terraform get
Get: file:///tmp/bug/foo
Error loading Terraform: module root: module foo: required variable foo not set
```

However in terraform 0.9, the only way to get an error is with a plan:

```
➜  bug terraform -v
Terraform v0.9.3

➜  bug terraform validate
➜  bug terraform get
Get: file:///tmp/bug/foo
➜  bug terraform plan
1 error(s) occurred:

* module root: module foo: required variable "foo" not set
```

It really seems like `terraform validate` is actually the proper tool to use here, but unfortunately it doesn't check for this condition. It'd be great if we could land a variation of this patch to check for this condition.

(hello @dtolnay! nice to run into you here, and potentially already have fixed the problem we're running into.)

PR #8590, opened 2016-09-01
providers/template: template_file supports floating point math

### LINKED ISSUE #7887: Variable interpolation doesn't work for floats
I'm testing Terraform at HEAD and it seems that variable interpolation inside a string assumes that arguments are integers. For example, the following snippet fails:

```
data "template_file" "test" {
  template = "${x + y}"

  vars {
    x = 0.001
    y = 0.002
  }
}
```

The error message is:

```
Error applying plan:

1 error(s) occurred:

* data.template_file.test: failed to render : __builtin_StringToInt: strconv.ParseInt: parsing "0.001": invalid syntax

```

It looks like `x` and `y` are being parsed as integers rather than floats because `${0.001 + 0.002}` works fine.

mitchellh:
Fixes #7887 

This adds a special case to detect numbers as floating point variables and converts them to floats. This allows floating point math to work.

### DIFF
--- a/builtin/providers/template/datasource_template_file.go
+++ b/builtin/providers/template/datasource_template_file.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/hil"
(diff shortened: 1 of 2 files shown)


jbardin:
LGTM

phinze:
This is unfortunately not a transparent change.

```tf
resource "template_file" "iam_policy" {
  template = "${file("${path.module}/templates/iam_policy.json.tpl")}"
  vars {
    region            = "${var.region}"
    owner_id          = "${aws_security_group.lab.owner_id}"
  }
}
```

This template is getting rendered as owner_id = "1.30490850807e+11"

sethvargo:
This breaks when the value is a bigint. It overflows and you get sadness like `1.30490850807e+11`, which is actually a bigint owner_id from AWS.

PR #8591, opened 2016-09-01
command/meta: validate config right away

### LINKED ISSUE #8524: Strange parse error when interpolating count using another resource
### Terraform Version

terraform 0.7.2

### Affected Resource(s)

Not specific to a resource

```hcl
provider "aws" {
  region = "us-west-1"
}

variable "azs" {
  type = "list"
  default = [ "us-west-1b", "us-west-1c" ]
}

variable "az_cidrs" {
  type = "list"
  default = [ "10.1.0.0/16", "10.2.0.0/16" ]
}

resource "aws_vpc" "default" {
  cidr_block = "10.0.0.0/8"
}

resource "aws_subnet" "private" {
  count = "${length(var.azs)}"

  vpc_id = "${aws_vpc.default.id}"
  availability_zone = "${var.azs[count.index]}"
  cidr_block = "${var.az_cidrs[count.index]}"
}

resource "aws_route_table_association" "private" {
  count = "${length(aws_subnet.private.*.id)}"
  subnet_id = "${element(aws_subnet.private.*.id, count.index)}"
  route_table_id = "${aws_route_table.private.id}"
}

resource "aws_route_table" "private" {
  vpc_id = "${aws_vpc.default.id}"
}
```

### Debug Output

https://gist.github.com/61c3dcbe1f3d64aedbf9b04da61d70de

### Expected Behavior

Either the count is correctly interpolated or a better error message is output if it is not possible to interpolate a count using the outputs of another resource.

### Actual Behavior

An incorrect error was output:

> * Error reading aws_subnet.private count: strconv.ParseInt: parsing "${length(var.azs)}": invalid syntax

### Steps to Reproduce
1. `terraform plan`

mitchellh:
Fixes #8524 

We were running `Input` before ever validating the configuration. This makes sense because `Context.Validate` require vars be set before running. However, we should always validate the configuration itself before doing anything else to catch issues like the above that is fixed.

This caught some errors in the tests I had to fix which should've never worked, you can't have interpolated defaults in variables for example.

### DIFF
--- a/command/test-fixtures/push-tfvars/main.tf
+++ b/command/test-fixtures/push-tfvars/main.tf
@@ -7,7 +7,6 @@ variable "baz" {
 
   default = {
     "A"    = "a"
-    interp = "${file("t.txt")}"
   }
 }
 
--- a/command/command.go
+++ b/command/command.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/mitchellh/cli"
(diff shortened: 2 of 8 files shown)


command/test-fixtures/push-tfvars/main.tf
```diff
@@ -7,7 +7,6 @@ variable "baz" {
 
   default = {
     "A"    = "a"
-    interp = "${file("t.txt")}"
```
mitchellh: @jen20 I noticed this was here. We've always had validation that this shouldn't work (interpolations not allowed as variable defaults). This was passing when it shouldn't have. Is this okay to remove? It doesn't work anyways.

jen20: Yes, this should be fine to remove.

jen20:
This LGTM - merged locally and verified the test runs etc.

PR #8595, opened 2016-09-01
#3449 Scope dependency to selected resource when indexing splat reference

### LINKED ISSUE #3449: Using element with splat reference should scope dependency to selected resource
I'm trying to setup a multi-node cluster with attached ebs volumes.  An example below:

```
resource "aws_instance" "nodes" {
    instance_type = "${var.model}"
    key_name = "${var.ec2_keypair}"
    ami = "${lookup(var.zk_amis, var.region)}"
    count = "${var.node_count}"
    vpc_security_group_ids = ["${aws_security_group.default.id}"]
    subnet_id = "${lookup(var.subnet_ids, element(keys(var.subnet_ids), count.index))}"
    user_data = "${file("cloud_init")}"
    tags {
        Name = "${var.cluster_name}-${count.index}"
    }
}

resource "aws_ebs_volume" "node-ebs" {
    count = "${var.node-count}"
    availability_zone = "${element(keys(var.subnet_ids), count.index)}"
    size = 100
    tags {
        Name = "${var.cluster_name}-ebs-${count.index}"
    }
}

resource "aws_volume_attachment" "node-attach" {
    count = "${var.node_count}"
    device_name = "/dev/xvdh"
    volume_id = "${element(aws_ebs_volume.node-ebs.*.id, count.index)}"
    instance_id = "${element(aws_instance.nodes.*.id, count.index)}"
}
```

If a change happens to a single node (for instance if a single ec2 instance is terminated) ALL of the aws_volume_attachments are recreated.  

Clearly we would not want volume attachments to be removed in a production environment.  Worse than that, in conjunction with #2957 you first must unmount these attachments before they can be recreated.  This has the effect of making volume attachments only viable on brand new clusters.

maxenglander:
(At least partially) fixes #3449.

This fix produces the expected output when using the new syntax `${list[idx]}`, but does not yet fix the issue when using the old syntax `${element(list,idx)}`.

As per @mitchellh advice on #3449, this is an attempt to add a test that reproduces the bug on v0.6. Mitchell advised creating a test for context_apply_test.go, but I started with context_plan_test.go since that seemed like a good place to test as well. Will add a test to context_apply_test.go if this isn't sufficient.

### DIFF
--- a/terraform/test-fixtures/plan-count-inc-index-list/main.tf
+++ b/terraform/test-fixtures/plan-count-inc-index-list/main.tf
@@ -19,6 +19,6 @@ resource "aws_ebs_volume" "foo" {
 
 resource "aws_volume_attachment" "foo" {
     count = "${var.count}"
-    instance_id = "${element(aws_instance.foo.*.id,count.index)}"
-    volume_id = "${element(aws_ebs_volume.foo.*.id,count.index)}"
+    instance_id = "${aws_instance.foo.*.id[count.index]}-baz"
+    volume_id = "${aws_ebs_volume.foo.*.id[count.index]}-baz"
 }
--- a/terraform/test-fixtures/plan-count-inc-index-list/main.tf
+++ b/terraform/test-fixtures/plan-count-inc-index-list/main.tf
@@ -1,0 +1,24 @@
+variable "count" {
+    default = 2
+}
+
+resource "aws_instance" "foo" {
+    count = "${var.count}"
+    foo = "foo"
+}
+
+resource "aws_instance" "bar" {
+    count = "${var.count}"
+    foo = "${element(aws_instance.foo.*.foo,count.index)}-bar"
+}
+
+resource "aws_ebs_volume" "foo" {
+    count = "${var.count}"
+    foo = "foo"
+}
+
+resource "aws_volume_attachment" "foo" {
+    count = "${var.count}"
+    instance_id = "${element(aws_instance.foo.*.id,count.index)}"
+    volume_id = "${element(aws_ebs_volume.foo.*.id,count.index)}"
+}
--- a/config/interpolate.go
+++ b/config/interpolate.go
@@ -305,53 +305,97 @@ func (v *UserVariable) GoString() string {
 // DetectVariables takes an AST root and returns all the interpolated
 // variables that are detected in the AST tree.
 func DetectVariables(root ast.Node) ([]InterpolatedVariable, error) {
-	var result []InterpolatedVariable
-	var resultErr error
+	collector := VariableCollector{}
+	VisitVariables(root, &collector)
 
-	// Visitor callback
-	fn := func(n ast.Node) ast.Node {
-		if resultErr != nil {
-			return n
-		}
+	if collector.HasErrors() {
+		errors := collector.GetErrors()
+		return nil, errors[len(errors)-1]
+	}
 
+	variables := collector.Collect()
+
+	return variables, nil
+}
+
+func VisitVariables(node ast.Node, visitor VariableVisitor) {
+	node.Accept(func(n ast.Node) ast.Node {
 		switch vn := n.(type) {
 		case *ast.VariableAccess:
 			v, err := NewInterpolatedVariable(vn.Name)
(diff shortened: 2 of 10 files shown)


terraform/test-fixtures/plan-count-inc-index-list/main.tf
```diff
@@ -0,0 +1,24 @@
...
+}
+
+resource "aws_instance" "foo" {
+    count = "${var.count}"
+    foo = "foo"
+}
+
+resource "aws_instance" "bar" {
+    count = "${var.count}"
+    foo = "${element(aws_instance.foo.*.foo,count.index)}-bar"
```
maxenglander: Interestingly, even though this resource is using the same element-splat-count-index technique as `aws_volume_attachment`, it is not marked for re-creation during the plan stage.

maxenglander: But if this is changed to `foo = "${element(aws_instance.foo.*.id,count.index)}-bar"`, then it _will_ be marked for update.

maxenglander:
This is ready for review! Looking forward to feedback 😃

terraform/test-fixtures/plan-count-inc-index-list/main.tf
maxenglander: Now that I've looked through the code a bit, naturally this makes sense: none of the `foo` attributes in `aws_instance.foo` are computed, and all are known.

ZeroPointEnergy:
Hi, I tried to merge your patch on top of 0.7.4, but I only get the following error when running "terraform plan":

`panic: interface conversion: interface is string, not []ast.Variable`

It is perfectly possible that I do something wrong here.

maxenglander:
@ZeroPointEnergy thanks for trying it out! I was able to reproduce - I wasn't properly handling an obvious use case. I've added a commit that should fix it..would appreciate if you gave it another shot. Thanks!

ZeroPointEnergy:
This looks a lot better now. Awesome work!

It now works partially for the plan I'm testing this with. I still have a problem with a server list which comes from a module and then gets used in resources in another module. There the variable still shows:

`instance: "somemachine.example.com" => "${var.servers[count.index]}" (forces new resource)`

For resources inside the same module this seams to work now and resources referencing an existing list entry do not get recreated.

maxenglander:
@ZeroPointEnergy thanks for testing again! I'll work on addressing this new use case and get back to you once it's fixed.

maxenglander:
@ZeroPointEnergy I was able to get it working with a configuration ([parent](https://github.com/maxenglander/terraform/blob/5e72b76cea3e8361306d61e7f425a26bd7502f7f/terraform/test-fixtures/plan-count-inc-index-list-cross-module/main.tf) and [child module](https://github.com/maxenglander/terraform/blob/5e72b76cea3e8361306d61e7f425a26bd7502f7f/terraform/test-fixtures/plan-count-inc-index-list-cross-module/foo/main.tf)) that looks like this. Not sure how similar this is to yours, but hopefully similar enough that it will work for you! Hope you'll give it another shot, thanks for your help.

ZeroPointEnergy:
@maxenglander This looks really good now. Applied your branch on top of 0.7.6 and it now calculates the dependencies correctly! Thanks a lot for your work, this is really awesome. I hope this will be merged into master soon.

oli-g:
Hi @maxenglander and @mitchellh, is this fix going to be merged to next 0.8 release? Is this PR still up to date? #3449 is here since a lot of time and it seems to be the main place where to track the issue with splat and counts.

Can you make a clear statement on that issue on when you plan to release the fix? I'm sure it will help a lot of people that maybe finds the bug for fist time.

Thank you and keep the good work!

mitchellh:
Hi @oli-g, we're currently targeting to attempt to fix this for 0.9.

We didn't merge the PR since it wasn't a complete fix and had some other unintended side effects. It was a great PR though in that it showed us one option and we're grateful for it being submitted. For those that it may fix their usage, they're welcome to apply the patch manually.

But hopefully we can get a longer term fix in before 0.9. :)

PR #8599, opened 2016-09-01
core: add wildcard (match all) support to ignore_changes

svanharmelen:
One use case for supporting a wildcard feature like this, is for when you want to completely lock down a resource in an efficient way.

You can already protect it from being destroyed or recreated by using `prevent_destroy`, but if you also want to prevent it from being altered in any way after it has been initially created, you'll have to manually make sure you have included all the possible attributes that can be used (so not only the ones you have initially configured, but all attributes).

Of course it is much easier if in those cases you can just add a single entry with a `*` to "lock" the resource for all changes.

### DIFF
--- a/website/source/docs/configuration/resources.html.md
+++ b/website/source/docs/configuration/resources.html.md
@@ -83,6 +83,10 @@ include `create_before_destroy`. Referencing a resource that does not include
 ~> **NOTE on ignore\_changes:** Ignored attribute names can be matched by their
 name, not state ID. For example, if an `aws_route_table` has two routes defined
 and the `ignore_changes` list contains "route", both routes will be ignored.
+Additionally you can also use a single entry with a wildcard (e.g. `"*"`)
+which will match all attribute names. Using a partial string together with a
+wildcard (e.g. `"rout*"`) is **not** supported.
+
 
 -------------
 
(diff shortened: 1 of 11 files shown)


website/source/docs/configuration/resources.html.md
```diff
@@ -83,6 +83,10 @@ include `create_before_destroy`. Referencing a resource that does not include
 ~> **NOTE on ignore\_changes:** Ignored attribute names can be matched by their
 name, not state ID. For example, if an `aws_route_table` has two routes defined
 and the `ignore_changes` list contains "route", both routes will be ignored.
+Additionally you can also use a single entry with a wildcard (e.g. `"*"`)
+which will match all attribute names. Using a partial string together with a
+wildcard (e.g. `"rout*"`) is **not** supported.
```
phinze: Can we add on a validation to this field to enforce that `*` is only a valid character when it appears alone?

svanharmelen: Yes, I guess it would be good to inform users to manage their expectations. So added a [validation](https://github.com/hashicorp/terraform/pull/8599/files#diff-3baf47c64847a8fb8aaa8cc2e088513b) for this one.

phinze:
LGTM!

svanharmelen:
Thanks for the review @phinze!

PR #8611, opened 2016-09-01
Fix multiple typos in current docs

alfonso-cabrera:
0.7.2 - kinesis firehose delivery stream docs


catsby:
😱 

Thanks for fixing all of these!

catsby:
Deployed! Thanks again

PR #8613, opened 2016-09-01
Waiting for 'terminated' on aws instance destroy is slow and unnecessary

knuckolls:
This changes the destroy behavior of the aws instance resource to shutting-down instead of terminated. It has sped up my development cycle significantly. Some terraform destroys were taking upwards of ten minutes before this patch. Please consider merging, as this should likely be the default behavior.


mitchellh:
I *think* the reason for this is when you have dependent things like security groups and so on, they don't let you destroy until their dependencies are fully terminated. I'll let @catsby comment though.

jen20:
Hi @knuckolls! As @mitchellh says, the reason for this is that we have to wait for instances to be destroyed in order to allow destruction of dependent resources - subnets, security groups and load balancers come to mind - so the current behaviour is correct even if it does sometimes take a while longer than desirable to wait for destruction. Thanks for opening a pull request, however!

Issue #4948, opened 2016-02-02
Terraform aws s3 bucket policy are planned for change every time


aldarund:
Every time i run plan and apply terraform says bucket policy is changed and needed to be applied. But its same policy all over again.
```
Path: /terraform/terraform.tfplan

~ aws_s3_bucket.rent_front_s3
    policy: "{\"Statement\":[{\"Action\":\"s3:GetObject\",\"Effect\":\"Allow\",\"Principal\":\"*\",\"Resource\":\"arn:aws:s3:::rent-front-static-staging/*\",\"Sid\":\"AddPerm\"}],\"Version\":\"2012-10-17\"}" => "{\"Statement\":[{\"Action\":[\"s3:GetObject\"],\"Effect\":\"Allow\",\"Principal\":\"*\",\"Resource\":[\"arn:aws:s3:::rent-front-static-staging/*\"],\"Sid\":\"AddPerm\"}],\"Version\":\"2012-10-17\"}"

```

radeksimko:
Hi @aldarund 
this is a known issue with IAM policies across all IAM/policy-based resources.

Short-term solution is to always use string instead of array for `Action` and `Resource` values if these only contain one item.

#### Long-term solution

In ideal case we could get IAM JSON mappings from the AWS SDK and do the (un)marshalling easily, that's being discussed in https://github.com/aws/aws-sdk-go/issues/127

Another, similar approach was proposed in https://github.com/hashicorp/terraform/pull/3124 where the mappings are part of the codebase.

Also there's https://github.com/hashicorp/terraform/pull/4278 which is following the same principal as the other linked PR.

brikis98:
I ran across the same issue with an `aws_s3_bucket` resource where it would show up changed every time I ran `terraform plan`. @radeksimko's fix to make all the `Action` and `Resource` values strings instead of arrays was not enough to fix it. I *also* had to get the `Principal` value just right.

Here's what the `aws_s3_bucket` looked like **before** the fix:

```hcl
resource "aws_s3_bucket" "my-bucket" {
  bucket = "my-bucket"

  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [{
    "Sid": "",
    "Action": ["s3:PutObject"],
    "Effect": "Allow",
    "Resource": ["my-resource"],
    "Principal": {
      "AWS": "1234567890"
    }
  }]
}
EOF
}
``` 

And here is the `aws_s3_bucket` **after** the fix:

```hcl
[...]

PR #8615, opened 2016-09-01
provider/aws: Add aws_s3_bucket_policy resource

jen20:
This commit adds a new "attachment" style resource for setting the policy of an AWS S3 bucket. This is desirable such that the ARN of the bucket can be referenced in an IAM Policy Document.

In addition, we now suppress diffs on the (now-computed) policy in the S3 bucket for structurally equivalent policies, which prevents flapping because of whitespace and map ordering changes made by the S3 endpoint.

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket_policy.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket_policy.go
@@ -1,0 +1,106 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsS3BucketPolicy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsS3BucketPolicyPut,
+		Read:   resourceAwsS3BucketPolicyRead,
+		Update: resourceAwsS3BucketPolicyPut,
+		Delete: resourceAwsS3BucketPolicyDelete,
+
+		Schema: map[string]*schema.Schema{
+			"bucket": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"policy": {
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: suppressEquivalentAwsPolicyDiffs,
+			},
+		},
+	}
+}
+
+func resourceAwsS3BucketPolicyPut(d *schema.ResourceData, meta interface{}) error {
+	s3conn := meta.(*AWSClient).s3conn
+
+	bucket := d.Get("bucket").(string)
+	policy := d.Get("policy").(string)
+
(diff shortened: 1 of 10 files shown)


jen20:
Just realised I forgot the docs, will add.

builtin/providers/aws/resource_aws_s3_bucket_policy.go
```diff
@@ -0,0 +1,112 @@
+package aws
+
+import (
+	//"encoding/json"
```
mitchellh: Worth removing?

builtin/providers/aws/resource_aws_s3_bucket_policy.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	} else {
+		if v := pol.Policy; v == nil {
+			if err := d.Set("policy", ""); err != nil {
+				return err
+			}
+		} else if err := d.Set("policy", v); err != nil {
+			return err
+		}
+	}
```
mitchellh: First, is it okay to be throwing away the first error? (`if err != nil`)

Second, I think this logic can be simplified quite a bit:

```go
v := ""
if err == nil && pol.Policy != nil {
  v = pol.Policy
}
if err := d.Set("policy", v); err != nil {
  return err
}
```

mitchellh:
A few reviews, and a meta question: I have no problem with that lib being under your account, but I also wouldn't be against it just being a sub-library within the Terraform repo if it is fairly specialized to us. Either way.

builtin/providers/aws/resource_aws_s3_bucket_policy.go
jen20: Yup, good catch, this was a hangover from extracting the same code out of `aws_s3_bucket`, we should probably clean it up there too.

jen20:
I think over time (once we have more confidence in effectiveness across a wider range of policy types) we should bring it to be a helper library in the AWS provider. For now I think it's better to keep it vendored so it can be worked on separately without risking regression in Terraform?

mitchellh:
LGTM!

kwilczynski:
@mitchellh the change here that introduces `Computed: true` to the `policy` in S3 bucket seem to be causing this (on master):

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSS3Bucket_Policy'
2016/09/14 14:47:14 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSS3Bucket_Policy -timeout 120m
=== RUN   TestAccAWSS3Bucket_Policy
--- FAIL: TestAccAWSS3Bucket_Policy (52.19s)
       	testing.go:265: Step 1 error: Check failed: Check 2/2 error: unexpected end of JSON input
FAIL
exit status 1
FAIL   	github.com/hashicorp/terraform/builtin/providers/aws   	52.206s
make: *** [testacc] Error 1
```

Either the test or the ReadFunc needs to be updated, I haven't looked into this yet.

SydOps:
@stack72 

In `Terraform aws s3 bucket policy are planned for change every time` (#4948), you mentioned the problem has been fixed by this PR. But I still have the problem with latest terraform version. 

In [a comment](https://github.com/hashicorp/terraform/issues/4948#issuecomment-270701684), @jurajseffer  provided a workaround. Will this be fixed in terraform finally?

PR #8619, opened 2016-09-02
Add missing json attribute to documentation for aws_iam_policy_document

BabakMN:
Fixes `data variables must be four parts: data.TYPE.NAME.ATTR` error.


stack72:
Hi @BabakMN 

Thanks so much for the update here :)

Pail

Issue #8044, opened 2016-08-08
Passing value via -var fails


hugoduncan:
The following command fails with any terraform `.tf` files.

    $ terraform -d plan  -var "image=1234.dkr.ecr.us-east-1.amazonaws.com/proj:abcdef" 
    invalid value "image='1234.dkr.ecr.us-east-1.amazonaws.com/proj:abcdef'" for flag -var: Cannot parse value for variable image ("image='1234.dkr.ecr.us-east-1.amazonaws.com/proj:abcdef'") as valid HCL: At 1:7: illegal char

I would have expected anything after the `=` to have been assigned to the var `image`.

    $ terraform -v
    Terraform v0.7.0

qnu:
Passing a map also fails:
`terraform plan -var 'somemap={ env = "env", service = "zoo" }'`
somemap["env"] is empty. terraform v0.7.0

Nomon:
-var git_sha=9af86eacf7656af1279557d0ab887cc52622f37f in terraform apply causes:
Cannot parse value for variable git_sha ("git_sha=9af86eacf7656af1279557d0ab887cc52622f37f") as valid HCL: key 'af86eacf7656af1279557d0ab887cc52622f37f'

thatarchguy:
I'm experiencing this as well. 
The problem is when it starts with a number for me.

Adding a letter to the start or single quoting like @Nomon described works.
```
$ terraform plan -var "revision=23d324" deploy
invalid value "revision=23d324" for flag -var: Cannot parse value for variable revision ("revision=23d324") as valid HCL: key 'd324' expected start of object ('{') or assignment ('=')

# adding a letter the the beginning
$ terraform plan -var "revision=test-23d324" deploy
[works]

# using single quote
$ terraform plan -var 'revision="23d324"' deploy
[works]

$ terraform version
Terraform v0.7.0

```

hugoduncan:
Adding a letter doesn't seem to be a workaround, as terraform doesn't have any substring function, iiiuc.

I'm having trouble coming up with something that works when interpolating in value from the environment, e.g. for:

    terraform plan -var revision="${revision}" deploy

I'm resorting to generating a script with the command that uses the single quotes, and then calling the script.

hugoduncan:
Possibly related, passing a number to a var requires single quoting too, otherwise an error results:

    * variable host_port should be type string, got number

Nomon:
[...]

PR #8620, opened 2016-09-02
command: more resilient HCL check for inputs

mitchellh:
Fixes https://github.com/hashicorp/terraform/issues/8044

This introduces a more resilient check when parsing variable values as HCL for determining if they were really HCL or not. Before, we simply checked error message (I kept this around although it may no longer be necessary). Now, I check for certain starting symbols ignoring whitespace:

  * `[` - List

  * `{` - Map

  * `"` - String

If it DOESN'T have one of those, we return as-is.

Existing tests pass. I want to double check with @jen20 if this is enough.

### DIFF
--- a/command/flag_kv.go
+++ b/command/flag_kv.go
@@ -129,6 +129,14 @@ func (v *FlagStringSlice) Set(raw string) error {
 	return nil
 }
 
+var (
+	// This regular expression is how we check if a value for a variable
+	// matches what we'd expect a rich HCL value to be. For example: {
+	// definitely signals a map. If a value DOESN'T match this, we return
+	// it as a raw string.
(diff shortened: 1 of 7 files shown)


command/flag_kv.go
```diff
@@ -129,6 +129,14 @@ func (v *FlagStringSlice) Set(raw string) error {
 	return nil
 }
 
+var (
+	// This regular expression is how we check if a value for a variable
+	// matches what we'd expect a rich HCL value to be. For example: {
+	// definitely signals a map. If a value DOESN'T match this, we return
+	// it as a raw string.
+	varFlagHCLRe = regexp.MustCompile(`^["\[\{]`)
```
jen20: I think previously we would have allowed whitespace at the start of the string, so this is somewhat different.

PR #8623, opened 2016-09-02
Increase aws_rds_cluster timeout to 40 minutes

qivers:
The timeout was increased to 40 minutes when creating a new cluster in https://github.com/hashicorp/terraform/pull/8052. However when creating a cluster from a snapshot the timeout is still 15 minutes. This has been increased to 40 minutes.


stack72:
Hi @qivers 

This LGTM! Thanks :)

Paul

PR #8625, opened 2016-09-02
provider/openstack: Fix documentation of openstack LBaaS resources

Fodoj:
For unknown reason all resource names are wrong and some pieces were missing like pool_id for lb_member_v2.


jtopjian:
During the development of the lbaas v2 resources, it was decided to change them from "lbaas" to "lb" to match the lbaas v1 resources. Looks like the docs were missed. Thank you for catching this. :)

PR #8626, opened 2016-09-02
provider/aws: `aws_cloudwatch_log_stream` resource

stack72:
This is a requirement for enabling CloudWatch Logging on Kinesis Firehose

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCloudWatchLogStream_'
2016/09/02 16:19:14 Generated command/internal_plugin_list.go
-run=TestAccAWSCloudWatchLogStream_ -timeout 120m
=== RUN   TestAccAWSCloudWatchLogStream_basic
--- PASS: TestAccAWSCloudWatchLogStream_basic (22.31s)
=== RUN   TestAccAWSCloudWatchLogStream_disappears
--- PASS: TestAccAWSCloudWatchLogStream_disappears (21.21s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    43.538s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudwatch_log_stream.go
+++ b/builtin/providers/aws/resource_aws_cloudwatch_log_stream.go
@@ -1,0 +1,133 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCloudWatchLogStream() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCloudWatchLogStreamCreate,
+		Read:   resourceAwsCloudWatchLogStreamRead,
+		Delete: resourceAwsCloudWatchLogStreamDelete,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateCloudWatchLogStreamName,
+			},
+
+			"log_group_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceAwsCloudWatchLogStreamCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchlogsconn
+
+	log.Printf("[DEBUG] Creating CloudWatch Log Stream: %s", d.Get("name").(string))
+	_, err := conn.CreateLogStream(&cloudwatchlogs.CreateLogStreamInput{
+		LogGroupName:  aws.String(d.Get("log_group_name").(string)),
+		LogStreamName: aws.String(d.Get("name").(string)),
+	})
+	if err != nil {
+		return errwrap.Wrapf("Creating CloudWatch Log Stream failed: %s", err)
+	}
+
+	d.SetId(d.Get("name").(string))
+
+	return resourceAwsCloudWatchLogStreamRead(d, meta)
+}
+
+func resourceAwsCloudWatchLogStreamRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchlogsconn
+
+	ls, exists, err := lookupCloudWatchLogStream(conn, d.Id(), d.Get("log_group_name").(string), nil)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


apparentlymart:
Hmm... it feels weird to me for Terraform to make log *streams*. The design for Cloudwatch Logs is that each distinct writer makes its own stream, because each stream has a single "cursor" that is used for writing and only the client that wrote the most recent record has the cursor token for the next write.

Thus I think it makes sense for Terraform to make log *groups*, but individual log emitters (whether they be EC2 instances, Lambda functions, Kinesis Firehose streams, or whatever) should create one more more log streams themselves, on the fly.

In the case of Lambda and Kinesis Firehose, you don't even get to create the log *group*... they just make their own based on a predictable naming scheme.

radeksimko:
I think I know where @apparentlymart is coming from and it may does feel _a bit_ weird based on how many AWS services deal with log streams (the services just manage those directly), but I don't see why we shouldn't allow the user to manage these themselves.

>  The design for Cloudwatch Logs is that each distinct writer makes its own stream

I'd argue this is design of many AWS services (including Lambda and API Gateway), but not necessarily CloudWatch itself.

CloudFormation [does support this](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-logstream.html) alongside [CloudWatch Log Destinations](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-destination.html).
I'm assuming there was a reason why Amazon CloudFormation engineers spent time on implementing these (?). 😉

builtin/providers/aws/resource_aws_cloudwatch_log_stream.go
```diff
@@ -0,0 +1,133 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
```
jen20: Import formatting is screwed up

builtin/providers/aws/resource_aws_cloudwatch_log_stream.go
```diff
@@ -0,0 +1,133 @@
...
+func resourceAwsCloudWatchLogStreamRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchlogsconn
+
+	ls, exists, err := lookupCloudWatchLogStream(conn, d.Id(), d.Get("log_group_name").(string), nil)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		log.Printf("[DEBUG] CloudWatch Stream %q Not Found", d.Id())
```
jen20: We normally put a "Removing from state" message here IIRC?

builtin/providers/aws/resource_aws_cloudwatch_log_stream.go
```diff
@@ -0,0 +1,133 @@
...
+	conn := meta.(*AWSClient).cloudwatchlogsconn
+
+	log.Printf("[INFO] Deleting CloudWatch Log Stream: %s", d.Id())
+	params := &cloudwatchlogs.DeleteLogStreamInput{
+		LogGroupName:  aws.String(d.Get("log_group_name").(string)),
+		LogStreamName: aws.String(d.Id()),
+	}
+	_, err := conn.DeleteLogStream(params)
+	if err != nil {
+		return fmt.Errorf("Error deleting CloudWatch Log Stream: %s", err)
```
jen20: This should use errwrap

builtin/providers/aws/resource_aws_cloudwatch_log_stream.go
```diff
@@ -0,0 +1,133 @@
...
+func resourceAwsCloudWatchLogStreamCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchlogsconn
+
+	log.Printf("[DEBUG] Creating CloudWatch Log Stream: %s", d.Get("name").(string))
+	_, err := conn.CreateLogStream(&cloudwatchlogs.CreateLogStreamInput{
+		LogGroupName:  aws.String(d.Get("log_group_name").(string)),
+		LogStreamName: aws.String(d.Get("name").(string)),
+	})
+	if err != nil {
+		return fmt.Errorf("Creating CloudWatch Log Stream failed: %s", err)
```
jen20: This should use errwrap

jen20:
Few minor nits but otherwise this LGTM. Looking at the design of it we do need to support this to allow people to configure CloudWatch/Firehose - and while I agree with @apparentlymart that it would be a nicer design for writers to create their own streams it doesn't seem to be the case, so I think we'll have to follow CloudFormation's lead here.

apparentlymart:
I'm certainly fine with this... was just describing how I've observed CloudWatch behave in case it was rooted in a misunderstanding of how something works. Certainly no harm in having this here, as long as it doesn't cause confusion for other people trying to figure out how to use Cloudwatch Logs and thinking that they always need to create these things with Terraform (since for many services they would not need to).

PR #8629, opened 2016-09-02
provider/scaleway: Document ports in security group rules properly

louis-paul:
The documentation appears to have been written following the [Scaleway API](https://developer.scaleway.com/#security-groups-manage-rules-post). However this is not how the resource is actually implemented in the current version ([code](https://github.com/hashicorp/terraform/blob/9f314a3c29944885a9d61e54eb62e363d9a64c23/builtin/providers/scaleway/resource_scaleway_security_group_rule.go#L75) and [tests](https://github.com/hashicorp/terraform/blob/9f314a3c29944885a9d61e54eb62e363d9a64c23/builtin/providers/scaleway/resource_scaleway_security_group_rule_test.go#L156)).


mitchellh:
Thanks!

PR #8634, opened 2016-09-02
Revert "providers/template: template_file supports floating point math"

phinze:
Reverts hashicorp/terraform#8590

See comments on that thread.

### DIFF
--- a/builtin/providers/template/datasource_template_file.go
+++ b/builtin/providers/template/datasource_template_file.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


phinze:
cc @sethvargo @mitchellh

sethvargo:
👍  - do we need to reopen the original issue #7887?

mitchellh:
Great catch before a release!

mitchellh:
LGTM, I think we do, I think helper/schema needs a new type to support typed maps.

sethvargo:
I don't always run master, but when I do it's to catch pre-release bugs 😄

Issue #7919, opened 2016-08-02
template_file data source doesn't allow count meta-parameter


laggyluke:
### Terraform Version
Terraform v0.7.0

### Affected Resource(s)
- template_file

### Terraform Configuration Files
The new data source fails:
```hcl
data "template_file" "example" {
  template = "Hello World!"
  count = 2
}
```

While the legacy resource works fine (although with a deprecation warning):
```hcl
resource "template_file" "example" {
  template = "Hello World!"
  count = 2
}
```

### Expected Behavior
Something similar to what legacy resource gives:
```
[...skipped...]

+ template_file.example.0
    rendered: "<computed>"
    template: "Hello World!"

+ template_file.example.1
    rendered: "<computed>"
    template: "Hello World!"


Plan: 2 to add, 0 to change, 0 to destroy.
```

### Actual Behavior
```
There are warnings and/or errors related to your configuration. Please
fix these before continuing.

Errors:

  * data.template_file.example #1: : invalid or unknown key: count
  * data.template_file.example #0: : invalid or unknown key: count
```

### Steps to Reproduce
Simply run `terraform plan` with the above configuration file.

jen20:
Hi @laggyluke - we're aware of the limitation of not being able to use `count` with data sources, and will address this in an upcoming version. Thanks for opening an issue!

vaygr:
Yes, I'm facing the same issue while trying to convert this to data sources usage after **0.7.0** release and drop the deprecation warning:

```
resource "template_file" "config" {
  count    = "${length(keys(var.hosts))}"
  template = "${file("${path.module}/config.tpl")}"

  vars {
    host = "${element(values(var.hosts), count.index)}"
  }
}

resource "google_compute_instance" "server" {
  count        = "${length(keys(var.hosts))}"
  ...
  provisioner "file" {
    content = "${element(template_file.config.*.rendered, count.index)}"
    destination = "/etc/default/server.conf"
  }
}
```

spkane:
@jen20 It would be useful to update the *Migrating to Data Sources* section of the upgrade docs (https://www.terraform.io/upgrade-guides/0-7.html) with a note about this, so that folks don't upgrade all their files before realizing that they simply need to revert it, since this important functionality is missing.

osterman:
Any clever work-arounds? :)

osterman:
[...]

PR #8635, opened 2016-09-02
Allow count in data sources

### LINKED ISSUE #8527: count is not implemented in data sources

jbardin:
Count worked in data sources, but the resource validation would fail due to the extra "count" key in the raw config. Make sure the HCL loader removes count.

Fixes #8527

### DIFF
--- a/builtin/providers/test/data_source.go
+++ b/builtin/providers/test/data_source.go
@@ -11,10 +11,20 @@ func testDataSource() *schema.Resource {
 		Read: testDataSourceRead,
 
 		Schema: map[string]*schema.Schema{
-			"list": &schema.Schema{
+			"list": {
 				Type:     schema.TypeList,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"input": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"output": {
+				Type:     schema.TypeString,
+				Computed: true,
 			},
 		},
 	}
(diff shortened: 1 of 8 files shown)


mitchellh:
A couple things:

  * Could we add a `context_apply_test` as well for this to make sure that the multiple query works properly? 

  * Can we add a test to verify `count.index` works properly within a data source? (e.x. `foo = "${count.index}"`)

jbardin:
@mitchellh Can you give me an example of what you mean by that first part? I couldn't come up with a good test for apply that didn't just repeat the plan. It's probably the same issue I had with assigning "${count.index}" -- I couldn't figure out how to expose them correctly in the tests (the data sources show up differently in the state than regular resources).

jen20:
@mitchellh I'm not so sure there is a distinct apply test here because of the refresh at plan to make the values known?

mitchellh:
You're probably right with refresh and I can add a count.index test later.

jen20:
Also fixes #7919 (and probably some others).

jen20:
I've added a test of count.index, but to avoid circular dependencies or large modifications to the null providers used in context tests it is implemented as a unit test in the `builtin/providers/test` package. It's possible that there is a better way to test this in future however!

PR #8506, opened 2016-08-27
[GH-1275] Support for AWS access via IAMs AssumeRole functionality

imduffy15:
This commit enables terraform to utilise the assume role functionality
of sts to execute commands with different privileges than the API
keys specified.

Feedback very much so welcome, this was thrown together as a quick hack.

Signed-off-by: Ian Duffy <ian@ianduffy.ie>

PR #8638, opened 2016-09-02
provider/aws: Add support for AssumeRole prior to operations

### LINKED ISSUE #1275: Support for AWS access via IAM - AssumeRole
I get it's probably a lot of complexity to add, but it'd be great if Terraform supported using IAM roles in its AWS provider authentication.

This adds a few steps to how API credentials are used – you use your normal AWS access key and secret key to call `sts assume-role` in the API, and this responds with a temporary access key, secret key and session token which you use for the rest of the session. An example with AWS's own CLI is [here](http://docs.aws.amazon.com/IAM/latest/UserGuide/roles-walkthrough-crossacct-3-userole.html#roles-walkthrough-crossacct-3-userole-cli).

The big use case for this is cross-account access. We have lots of AWS accounts to isolate billing to individual projects, and we link these to one parent account with AWS's consolidated billing. The cross-account roles let us have one set of IAM users in the parent account (eg one IAM user for each actual developer using AWS), and then just a single IAM role in each sub-account, which the developers can assume when needed. This saves having to create a new set of IAM users in each sub-account.

I imagine this would involve adding extra AWS provider config fields to describe the role to assume (using an AWS ARN like `arn:aws:iam::account-of-role-to-assume:role/name-of-role`), and would also involve changing the AWS API access layer to use the temporary role access/secret keys _and_ a session token.

This should be linked to the [AWS meta-issue](https://github.com/hashicorp/terraform/issues/28). I felt the explanation was long enough to warrant a full issue.

jen20:
This PR builds on the work of @imduffy15 in #8506, adding in support for Session Name and External ID, and changing around the configuration block a bit.

There is still a case which needs fixing regarding the error message when no valid source credentials are passed in, and also if the source credentials are valid but the role cannot be assumed.

Automated testing is left as a follow-up to this - we have a strategy in place but it requires some (more broadly useful) changes to the acceptance testing harness to be feasible.

Fixes #1275.

### DIFF
--- a/builtin/providers/aws/auth_helpers.go
+++ b/builtin/providers/aws/auth_helpers.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -18,7 +19,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
-	"github.com/hashicorp/go-multierror"
 )
 
 func GetAccountId(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (string, error) {
@@ -77,7 +77,7 @@ func GetAccountId(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (
 	}
 
 	if len(outRoles.Roles) < 1 {
-		return "", fmt.Errorf("Failed getting account ID via 'iam:ListRoles': No roles available")
+		return "", errors.New("Failed getting account ID via 'iam:ListRoles': No roles available")
 	}
 
 	return parseAccountIdFromArn(*outRoles.Roles[0].Arn)
@@ -95,8 +95,6 @@ func parseAccountIdFromArn(arn string) (string, error) {
 // environment in the case that they're not explicitly specified
 // in the Terraform configuration.
 func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
-	var errs []error
-
 	// build a chain provider, lazy-evaulated by aws-sdk
 	providers := []awsCredentials.Provider{
 		&awsCredentials.StaticProvider{Value: awsCredentials.Value{
@@ -130,7 +128,7 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 			providers = append(providers, &ec2rolecreds.EC2RoleProvider{
 				Client: metadataClient,
 			})
-			log.Printf("[INFO] AWS EC2 instance detected via default metadata" +
+			log.Print("[INFO] AWS EC2 instance detected via default metadata" +
 				" API endpoint, EC2RoleProvider added to the auth chain")
 		} else {
 			if usedEndpoint == "" {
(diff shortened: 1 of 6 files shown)


builtin/providers/aws/auth_helpers.go
```diff
@@ -75,7 +78,7 @@ func GetAccountId(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (
 	}
 
 	if len(outRoles.Roles) < 1 {
-		return "", fmt.Errorf("Failed getting account ID via 'iam:ListRoles': No roles available")
+		return "", errors.New("Failed getting account ID via 'iam:ListRoles': No roles available")
```
stack72: why errors.New and not errWrap?

builtin/providers/aws/auth_helpers.go
```diff
@@ -137,7 +142,49 @@ func GetCredentials(c *Config) *awsCredentials.Credentials {
...
-	return awsCredentials.NewChainCredentials(providers)
+	if c.AssumeRoleARN != "" {
+		log.Printf("[INFO] Attempting to AssumeRole %s (SessionName: %q, ExternalId: %q)",
+			c.AssumeRoleARN, c.AssumeRoleSessionName, c.AssumeRoleExternalID)
+
+		creds := awsCredentials.NewChainCredentials(providers)
+		cp, err := creds.Get()
+		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NoCredentialProviders" {
+				errs = append(errs, errors.New(`No valid credential sources found for AWS Provider.
```
stack72: I think this is the error message we spoke about - isn't this the same generic error message we throw for non-assume role? Shouldn't we mention something here about assume role not working due to credentials?

builtin/providers/aws/auth_helpers_test.go
```diff
@@ -251,10 +256,13 @@ func TestAWSGetCredentials_shouldBeStatic(t *testing.T) {
 			Token:     c.Token,
 		}
 
-		creds := GetCredentials(&cfg)
+		creds, err := GetCredentials(&cfg)
 		if creds == nil {
 			t.Fatalf("Expected a static creds provider to be returned")
 		}
+		if err != nil {
```
stack72: I think we should check the err before we check if the creds are nil

builtin/providers/aws/auth_helpers_test.go
```diff
@@ -286,11 +294,13 @@ func TestAWSGetCredentials_shouldIAM(t *testing.T) {
 	// An empty config, no key supplied
 	cfg := Config{}
 
-	creds := GetCredentials(&cfg)
+	creds, err := GetCredentials(&cfg)
 	if creds == nil {
 		t.Fatalf("Expected a static creds provider to be returned")
 	}
-
+	if err != nil {
```
stack72: Same - err check first before creds?

builtin/providers/aws/auth_helpers_test.go
```diff
@@ -362,7 +375,10 @@ func TestAWSGetCredentials_shouldErrorWithInvalidEndpoint(t *testing.T) {
 	ts := invalidAwsEnv(t)
 	defer ts()
 
-	creds := GetCredentials(&Config{})
+	creds, err := GetCredentials(&Config{})
+	if err != nil {
+		t.Fatalf("Error gettings creds: %s", err)
+	}
 	v, err := creds.Get()
```
stack72: Do we want to check the creds are not nil as elsewhere?

builtin/providers/aws/auth_helpers_test.go
```diff
@@ -335,10 +345,13 @@ func TestAWSGetCredentials_shouldIgnoreIAM(t *testing.T) {
 			Token:     c.Token,
 		}
 
-		creds := GetCredentials(&cfg)
+		creds, err := GetCredentials(&cfg)
 		if creds == nil {
 			t.Fatalf("Expected a static creds provider to be returned")
 		}
+		if err != nil {
```
stack72: Err before creds? :)

builtin/providers/aws/auth_helpers_test.go
```diff
@@ -380,7 +396,10 @@ func TestAWSGetCredentials_shouldIgnoreInvalidEndpoint(t *testing.T) {
 	ts := invalidAwsEnv(t)
 	defer ts()
 
-	creds := GetCredentials(&Config{AccessKey: "accessKey", SecretKey: "secretKey"})
+	creds, err := GetCredentials(&Config{AccessKey: "accessKey", SecretKey: "secretKey"})
+	if err != nil {
+		t.Fatalf("Error gettings creds: %s", err)
+	}
 	v, err := creds.Get()
```
stack72: check that creds are not nil ?

stack72:
@jen20 so pleased to see this included. I left a few nit picks about the ordering of error checking and about what way we throw errors. These are not show stoppers at all :) More nit picks

builtin/providers/aws/auth_helpers.go
jen20: We are generating a new error rather than wrapping one produced by an upstream API?

builtin/providers/aws/auth_helpers_test.go
jen20: Yup, added this now.

builtin/providers/aws/auth_helpers_test.go
jen20: :+1:

stack72:
LGTM! Nice work :)

imduffy15:
Thanks for making this happen @stack72 and @jen20

bentterp:
I totally don't understand how to use this, could somebody please explain? I've read through #1275 as well, without any light bulbs going off.

Until now, we've been using a script that does STS::AssumeRole with MFA, obtains the temporary credentials, exports them as environment variables. After that, we happily continue with terraform plan and apply.

Is there now a smarter way of doing this?

imduffy15:
Hi @benttrep,
There is indeed, there's no more need for assume role scripts. 

Authenticate to your account that has the assume role privilege using any of the standard authentication methods(env vars, profile, ec2 IAM role etc). 

Then just follow the docs at https://www.terraform.io/docs/providers/aws/

bentterp:
Aaah :-D This is awesome :exclamation: 

And the docs were updated as part of the release :+1: 

A suggestion: could we have a link to the updated docs included in the changelog? As the documentation grows, it becomes harder and harder to spot the changes in the docs.

E.g. https://www.terraform.io/docs/providers/aws/index.html#Assume_role - and yes, anchors in the docs would be great.

drdamour:
i'm probably missing something but the docs aren't really followable...i have the arn_role block in my assume role...but it's not working. always get

```

* The role "arn:aws:iam:::role/devops" cannot be assumed.

  There are a number of possible causes of this - the most common are:
    * The credentials used in order to assume the role are invalid
    * The credentials do not have appropriate permission to assume the role
    * The role ARN is not valid
```

how would i know what value to put in session_id and external_id?  i expected terraform to prompt me for my MFA value.

jen20:
Hi @drdamour! There are a few different issues in your example:

Firstly the role ARN is indeed not valid - they have an account ID in them. The IAM console will show you the correct ARN for a given role. This covers the third error.

Secondly, Assuming a role has two aspects:
- the first is acquiring a session token - the `sts:GetSessionToken` API operation. This is where an MFA code is typically specified. Terraform does _not_ do this for you (though if running as part of Terraform Enterprise it can be configured).
- the second is assuming a role - the `sts:AssumeRole` operation. Roles may specify as part of the specification that in order to be assumed, the session token provided must have an MFA token attached. This is the part that Terraform does for you if the `role_arn` option is specified.

In order to fix your example you'll need to do the following:

1. Fix the role ARN so it is valid
1. Make a call to sts:GetSessionToken via whatever mechanism you would use for the CLI (likely `aws sts get-session-token`) and use the credentials returned as the values to `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN`.

drdamour:
thanks, i was obfuscating my role id...i'm using a valid one.  the whole reason we had the script was to get the session token. i suspect that's what @bentterp 's script was doing as well...so alas you still need a wrapper script to do stuff with MFA..that sucks.

thanks for confirming.

Issue #5552, opened 2016-03-10
google_container_cluster hits 400 when using a network with self_link


devth:
```
google_container_cluster: {
  ...
  network: "${google_compute_network.test.self_link"
}
```

Trying to apply results in:

```
* google_container_cluster.us-central1-c-standard-test-private: googleapi: Error 400: network "https://www.googleapis.com/compute/v1/projects/.../global/networks/test-m" does not exist., badRequest
```

The `prod` network exists for sure. When I refresh tf state, it shows up. It also shows up in the list of networks from `google compute networks list`.

devth:
It works if I do this instead:

```
google_container_cluster: {
  ...
  network: "test-m"
}
```

`test` is the network name in terraform inventory, `test-m` is the actual resource name in GCP.

sl1pm4t:
The Google API expects the name of the network, not the resource URL.
Does it work if you change your resource definition to:
```
google_container_cluster: {
  ...
  network: "${google_compute_network.test.name"
}
```

paddyforan:
This should now work with either the `self_link` or the `name`, thanks to @sl1pm4t's work, so I'm going to close this issue out. If you encounter more problems, feel free to open another issue or comment in this issue again, and we'll reopen it. Thanks!

PR #8639, opened 2016-09-03
Google: Make network attribute more consistent across resources

### LINKED ISSUE #5584: google network attribute usage inconsistent
For different resources in the google cloud provider the usage of the network attribute is inconsistent.

For example the **google_compute_subnetwork** uses self_link:

_network = "${google_compute_network.default.self_link}"_

Whereas the **google_compute_firewall** uses name:

_network = "${google_compute_network.default.name}"_

This might be an inconsistency in the google apis, but makes it unpredictable how to use terraform.

sl1pm4t:
Subnetwork and VPN Gateway Google resources required the network to be supplied as the resource URL, but the rest of the Google resources accepted just the network name.
This makes the Subnetwork and VPN Gateway resources more consistent, by allowing the network name to be provided.
The network resource URL (aka SelfLink) can still be provided for backwards compatibility.

Fixes #5584

### DIFF
--- a/builtin/providers/google/resource_compute_subnetwork.go
+++ b/builtin/providers/google/resource_compute_subnetwork.go
@@ -91,12 +91,17 @@ func resourceComputeSubnetworkCreate(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
+	network, err := getNetworkLink(d, config, "network")
+	if err != nil {
+		return err
+	}
+
 	// Build the subnetwork parameters
 	subnetwork := &compute.Subnetwork{
 		Name:        d.Get("name").(string),
 		Description: d.Get("description").(string),
 		IpCidrRange: d.Get("ip_cidr_range").(string),
-		Network:     d.Get("network").(string),
+		Network:     network,
 	}
 
 	log.Printf("[DEBUG] Subnetwork insert request: %#v", subnetwork)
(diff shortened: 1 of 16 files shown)


sl1pm4t:
Ping @lwander - as discussed in #5584

builtin/providers/google/resource_compute_subnetwork.go
```diff
@@ -91,12 +91,24 @@ func resourceComputeSubnetworkCreate(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
+	// Network can be provided as just the name, or the full resource URL
+	network := d.Get("network").(string)
```
lwander: Looks like this code shows up twice - could you extract this into a shared file?

lwander:
Nice work! LGTM after the comment

sl1pm4t:
Hi @lwander - I DRY'd out the network lookup logic for the two resources I'd previously touched, and the others that were also doing a similar lookup (google_compute_instance, google_compute_instance_template, google_compute_route). 
I also added a fix for #5552, so the google_container_cluster resource could accept network name or self_link.
Thanks!

lwander:
Great stuff - could you squash and then I'll merge? Thanks!

sl1pm4t:
Hi @lwander - the commits have been squashed.
Thanks!

PR #6213, opened 2016-04-18
Officially support "local" remote state

joshuaspence:
See #3618. I'm not entirely sure if any other changes are required here, but would appreciate some feedback. Quite possibly maybe this PR should wait until after #4169.

PR #8647, opened 2016-09-03
state/remote: Officially Support local backend

### LINKED ISSUE #3618: Feature: file backend for remote state
It would be nice to have the same idea of remote state implemented with `file` as a backend.

Bringing `_local` in as a first-class form of remote state would probably suffice.

https://github.com/hashicorp/terraform/blob/master/state/remote/remote.go#L47
```
resource "terraform_remote_state" "vpc" {
    backend = "file"
    
    config {
        path = "./vpc/terraform.tfstate"
    }
}
```

jen20:
This is a rework of pull request #6213 submitted by @joshuaspence, adjusted to work with the remote state data source. We also add a deprecation warning for people using the unsupported API, and retain the ability to refer to "_local" as well as "local" for users in a mixed version environment.

Fixes #3618, reworks #6213.

### DIFF
--- a/builtin/providers/terraform/data_source_state.go
+++ b/builtin/providers/terraform/data_source_state.go
@@ -16,6 +16,16 @@ func dataSourceRemoteState() *schema.Resource {
 			"backend": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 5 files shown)


stack72:
LGTM! Backwards compatibility FTW :)

PR #8672, opened 2016-09-06
provider/aws: Bump `aws_elasticsearch_domain` timeout values

### LINKED ISSUE #8541: [provider/aws] aws_elasticsearch_domain delete timeout too "short" ?
### Terraform Version
0.7.2-dev (master)

### Affected Resource(s)
- aws_elasticsearch_domain

The current Terraform implement has a timeout set at 60 mins for create/update/delete.  It is enough for create and update.  However, I notice delete can take over an hour (it's never definied in AWS documentation).  I wonder if it makes sense to increase the delete timeout on Terraform.

stack72:
Fixes #8541

The Update timeout and delete timeouts were a little short. Bumped them
to 60 mins and 90 mins respectively. I have been on the receiving of the
timeout for the Delete function


catsby:
LGTM

PR #8678, opened 2016-09-06
Improvements for Scaleway machine creation

larsla:
- Add support for creating extra volumes during machine creation.
  Some machine types require extra volumes to be present when creating the machine.
- Add support for specifying a name of a bootscript instead of only id.
  Finding the ID of a bootscript is a big pain.

### DIFF
--- a/builtin/providers/scaleway/resource_scaleway_server.go
+++ b/builtin/providers/scaleway/resource_scaleway_server.go
@@ -99,20 +99,6 @@ func resourceScalewayServerCreate(d *schema.ResourceData, m interface{}) error {
 	server.DynamicIPRequired = Bool(d.Get("dynamic_ip_required").(bool))
 	server.CommercialType = d.Get("type").(string)
 
-	arch := ""
-	if arch == "" {
-		server.CommercialType = strings.ToUpper(server.CommercialType)
-		switch server.CommercialType[:2] {
-		case "C1":
-			arch = "arm"
-		case "C2", "VC":
-			arch = "x86_64"
-		default:
-			log.Printf("[ERROR] %s wrong commercial type", server.CommercialType)
-			return errors.New("Wrong commercial type")
-		}
-	}
-
 	if bootscript, ok := d.GetOk("bootscript"); ok {
 		bootscript_id := bootscript.(string)
 
--- a/builtin/providers/scaleway/resource_scaleway_server.go
+++ b/builtin/providers/scaleway/resource_scaleway_server.go
@@ -1,8 +1,11 @@
 package scaleway
 
 import (
+	"errors"
 	"fmt"
 	"log"
+	"strconv"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/scaleway/scaleway-cli/pkg/api"
@@ -70,6 +73,13 @@ func resourceScalewayServer() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"volumes": &schema.Schema{
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeInt,
+				},
+				Required: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


stack72:
Hi @larsla 

Thanks for the PR here - please can you update the associate documentation for this ? :)

Paul

larsla:
Hi @stack72 
I updated the documentation with the new options.

//Lars

builtin/providers/scaleway/resource_scaleway_server.go
```diff
@@ -70,6 +71,13 @@ func resourceScalewayServer() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"volumes": &schema.Schema{
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeInt,
+				},
+				Required: true,
```
stack72: Should this be required? Do we always have to create a volume?

builtin/providers/scaleway/resource_scaleway_server.go
```diff
@@ -99,6 +120,25 @@ func resourceScalewayServerCreate(d *schema.ResourceData, m interface{}) error {
 		}
 	}
 
+	if raw, ok := d.GetOk("volumes"); ok {
+		server.Volumes = make(map[string]string)
+		for i, vol := range raw.([]interface{}) {
+			var volume = api.ScalewayVolumeDefinition{
+				Name:         fmt.Sprintf("%s-%s", server.Name, strconv.Itoa(vol.(int))),
+				Size:         uint64(vol.(int)) * gb,
+				Type:         "l_ssd",
```
stack72: Is there only 1 type of Volume we can create?

stack72:
Hi @larsla 

Thanks for the PR here. I can see that we have already got `scaleway_volume` and `scaleway_volume_attachment`

Can you tell me what will happen if i specify a volume as part of a server and then also specify a volume externally and then attachment?

I think we are going to need a test to ensure that this works as expected :)

Is there only the option to specify the volume sizes right now?

Paul

builtin/providers/scaleway/resource_scaleway_server.go
nicolai86: When you don't specify volumes you'll just get the default root volume after invoking `api.PostServer()`

nicolai86:
I think that instead of overloading the `bootscript` attribute to lookup identifiers by name, we should have a data source for this. What do you think, @larsla ?

builtin/providers/scaleway/resource_scaleway_server.go
tboerger: So it's not required, it should be optional

builtin/providers/scaleway/resource_scaleway_server.go
```diff
@@ -90,7 +98,20 @@ func resourceScalewayServerCreate(d *schema.ResourceData, m interface{}) error {
 	server.CommercialType = d.Get("type").(string)
 
 	if bootscript, ok := d.GetOk("bootscript"); ok {
-		server.Bootscript = String(bootscript.(string))
+		bootscript_id := bootscript.(string)
+
+		bootscripts, err := scaleway.GetBootscripts()
```
nicolai86: terraform 0.7.7 added a data source so we don't have to overload the meaning here anymore. please revert this chunk

tboerger:
@nicolai86 this is now already implemented, right?

nicolai86:
@tboerger yes.

stack72:
@nicolai86 this can be closed?

nicolai86:
@stack72 yup

stack72:
@nicolai86 what PR / commit was it added in?

nicolai86:
@stack72 the server volume attribute is https://github.com/hashicorp/terraform/pull/9695 and the bootscript is now a datasource as of https://github.com/hashicorp/terraform/pull/9386

stack72:
Closing as per #9695 and #9386 - thanks again @larsla :)

PR #8685, opened 2016-09-06
Corrected stage_name argument

MarcusNoble:
**Reasoning for docs update:** api_gateway_base_path_mapping docs incorrectly state the argument as `stage` when it should be `stage_name`


stack72:
Hi @MarcusNoble 

thanks so much :) This LGTM

Paul

PR #8691, opened 2016-09-06
Cleanup inaccurate state structure comments.

tpounds:
Minor cleanup of some inaccurate comments I noticed regarding state file structure and version.


stack72:
Thanks for this @tpounds :)

PR #8700, opened 2016-09-07
provider/aws: Export AWS ELB service account ARN

steveh:
I'm new to Go so please bear with me.

This PR exports the `arn` attribute for the AWS ELB service account data source.

While `id` is helpful, AWS will re-write plain IDs in policies to use the full ARN, generating spurious diffs. Perhaps that is better solved by changing the policy parser/differ, but I don't know how to do that...

I was unable to get `make test TEST=./builtin/providers/aws/` to work so I'm relying on CI or someone to tell me what's wrong, or how to test just this data source.

### DIFF
--- a/builtin/providers/aws/data_source_aws_elb_service_account.go
+++ b/builtin/providers/aws/data_source_aws_elb_service_account.go
@@ -32,6 +32,10 @@ func dataSourceAwsElbServiceAccount() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -44,6 +48,9 @@ func dataSourceAwsElbServiceAccountRead(d *schema.ResourceData, meta interface{}
 
 	if accid, ok := elbAccountIdPerRegionMap[region]; ok {
 		d.SetId(accid)
+
+		d.Set("arn", "arn:aws:iam::"+accid+":root")
+
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


stack72:
Hi @steveh 

Thanks for the work here :) It's great to see someone new be able to work with the repository!

Can you give me a small usecase for this - i.e. what requires the ARN over the ID?

Paul

steveh:
@stack72: I updated the description:

--

While id is helpful, AWS will re-write plain IDs in policies to use the full ARN, generating spurious diffs. Perhaps that is better solved by changing the policy parser/differ, but I don't know how to do that...

radeksimko:
I think I understand what do you mean @steveh 

Even though the following policy is accepted:

```json
{
  "Id": "Policy",
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "s3:PutObject"
      ],
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::my-elb-tf-test-bucket/AWSLogs/*",
      "Principal": {
        "AWS": [
          "123456789012"
        ]
      }
    }
  ]
}
```
it is in reality understood and translated by S3/IAM as _"any users in that AWS account"_, therefore this

```
123456789012
```
becomes
```
arn:aws:iam::123456789012:root
```

So it is indeed helpful.

The policy [diff comparator](https://github.com/jen20/awspolicyequivalence) could be improved as well. I'm sure @jen20 would be happy to accept a PR for this. 😉 
The change would be probably somewhere [here](https://github.com/jen20/awspolicyequivalence/blob/master/aws_policy_equivalence.go#L206).

steveh:
@radeksimko Yes, pretty much. If you put `123456789012` in a policy, when AWS returns it, it will be expanded to the full ARN, so Terraform will try and change it back to `123456789012` again.

PR #8704, opened 2016-09-07
fix typo in local-exec provisioner documentation

sjourdan:
A recent edit to this doc added a minor typo.


stack72:
Thanks for the fix here @sjourdan :) 

Paul

PR #8722, opened 2016-09-08
typo consul catalog entry example

JohnDittmar:
In the consul catalog entry example `name` was specified in the root of the resource rather than the key `node` which is the actual required key-name.


stack72:
Hi @JohnDittmar 

Good catch here - :) Thanks for the PR! We love updates to the documentation

Paul

PR #8726, opened 2016-09-08
Reword "Multiple Provider Instances" documentation

### LINKED ISSUE #8702: Bad wording in website/source/docs/configuration/providers.html.md
>The primary use case for this for multiple cloud regions.

Should this be *The primary use case for this is multiple cloud regions* instead?

>Other use cases including targeting multiple Docker hosts, multiple Consul hosts, etc.

Should this be *Other use cases include targeting multiple Docker hosts, multiple Consul hosts, etc.* instead?

cblecker:
Fixes #8702.


stack72:
Hi @cblecker 

Thanks so much - I would agree with you that this is better :)

LGTM!

Paul

Issue #8729, opened 2016-09-08
Terraform 0.7.3 - crash with disallowed AWS account


elblivion:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
0.7.3

### Affected Resource(s)
AWS Provider: Terraform crashes when using an account which is not in `allowed_account_ids` attribute of the AWS provider.

### Terraform Configuration Files
Configuration is not an issue; my config works when I have the API keys for the correct AWS account in my environment.

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
https://gist.github.com/anonymous/188b284fe43c5747c1b1f13418e7a078

### Expected Behavior
Terraform should give an "Account not allowed" error message as it did until 0.7.2 (I think, 0.6.16 for sure).

### Actual Behavior
Crash

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. set AWS keys for an account not in `allowed_account_ids` in the AWS provider config
2. `terraform plan`

### Important Factoids
N/A

### References
Not to my knowledge.

radeksimko:
Thanks for the report @elblivion .

Just cutting out the important part of crash log:

```
2016/09/08 07:49:20 [DEBUG] plugin: terraform: panic: interface conversion: interface {} is nil, not *aws.AWSClient
2016/09/08 07:49:20 [DEBUG] plugin: terraform: 
2016/09/08 07:49:20 [DEBUG] plugin: terraform: goroutine 760 [running]:
2016/09/08 07:49:20 [DEBUG] plugin: terraform: panic(0x28d0ae0, 0xc424e2b340)
2016/09/08 07:49:20 [DEBUG] plugin: terraform: 	/opt/go/src/runtime/panic.go:500 +0x1a1
2016/09/08 07:49:20 [DEBUG] plugin: terraform: github.com/hashicorp/terraform/builtin/providers/aws.resourceAwsElasticacheParameterGroupRead(0xc424efe360, 0x0, 0x0, 0x0, 0x1f)
[...]

PR #8731, opened 2016-09-08
provider/aws: Prevent crash on account ID validation

radeksimko:
Fixes https://github.com/hashicorp/terraform/issues/8729

Introduced in https://github.com/hashicorp/terraform/pull/8638

### Test plan
```hcl
provider "aws" {
  region = "us-west-2"
  allowed_account_ids = ["invalid"]
}
```
expected output:
```

* Account ID not allowed (123456789012)
```


stack72:
Nice catch here @radeksimko :)

This LGTM!

Issue #8491, opened 2016-08-26
Add gzip support for Consul remote state


christoe:
After adopting terraform more and more, we've now come to a point when the tfstate is too big to be stored in Consul (that we use for remote state):

`Failed to save state: Unexpected response code: 413 (Value exceeds 524288 byte limit)`

An idea, since the state is quite verbose, would be to allow for the state to be stored gzip'ed in Consul. I'm willing to implement this myself, but I would just like to verify that this sounds sane?
### Terraform Version

Any

carlpett:
@stack72 I ran across this today, forcing me to now start making manual changes to our AWS resources since I can no longer use terraform to change anything. Is there something preventing this getting merged?

christoe:
Closing this as the PR that resolves this has been merged.

PR #8748, opened 2016-09-09
state/remote: Add gzip support to consul remote state (#8491)

christoe:
This allows for storing the remote state compressed in Consul (#8491): it will transparently gzip/un-gzip the state using the Go standard gzip library. It will not care in what format the current remote state is in, which allows you to change the gzip_state setting at any time.

### DIFF
--- a/backend/remote-state/consul/backend.go
+++ b/backend/remote-state/consul/backend.go
@@ -53,6 +53,13 @@ func New() backend.Backend {
 				Description: "HTTP Auth in the format of 'username:password'",
 				Default:     "", // To prevent input
 			},
+
+			"gzip": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Compress the state data using gzip",
+				Default:     false,
+			},
 		},
 	}
 
--- a/backend/remote-state/consul/backend_state.go
+++ b/backend/remote-state/consul/backend_state.go
@@ -85,11 +85,15 @@ func (b *Backend) State(name string) (state.State, error) {
 	// Determine the path of the data
 	path := b.path(name)
 
+	// Determine whether to gzip or not
+	gzip := b.configData.Get("gzip").(bool)
+
 	// Build the state client
 	stateMgr := &remote.State{
 		Client: &RemoteClient{
 			Client: client,
 			Path:   path,
+			GZip:   gzip,
 		},
 	}
 
--- a/backend/remote-state/consul/client.go
+++ b/backend/remote-state/consul/client.go
@@ -18,13 +18,13 @@ import (
 const (
 	lockSuffix     = "/.lock"
 	lockInfoSuffix = "/.lockinfo"
-	maxKVSize      = 512 * 1024
 )
 
 // RemoteClient is a remote client that stores data in Consul.
 type RemoteClient struct {
 	Client *consulapi.Client
 	Path   string
+	GZip   bool
 
 	consulLock *consulapi.Lock
 	lockCh     <-chan struct{}
@@ -58,8 +58,7 @@ func (c *RemoteClient) Get() (*remote.Payload, error) {
 
 func (c *RemoteClient) Put(data []byte) error {
 	payload := data
-	// If the payload to be written exceeds the Consul KV byte limit, compress
-	if len(data) > maxKVSize {
+	if c.GZip {
 		if compressedState, err := compressState(data); err == nil {
 			payload = compressedState
 		} else {
--- a/backend/remote-state/consul/client.go
+++ b/backend/remote-state/consul/client.go
@@ -1,6 +1,8 @@
 package consul
 
 import (
+	"bytes"
+	"compress/gzip"
 	"crypto/md5"
 	"encoding/json"
 	"errors"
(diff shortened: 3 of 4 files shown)


website/source/docs/state/remote/consul.html.md
```diff
@@ -47,3 +47,5 @@ The following configuration options / environment variables are supported:
  * `datacenter` - (Optional) The datacenter to use. Defaults to that of the agent.
  * `http_auth` / `CONSUL_HTTP_AUTH` - (Optional) HTTP Basic Authentication credentials to be used when
    communicating with Consul, in the format of either `user` or `user:pass`.
+ * `gzip_state` - (Optional) Specifies whether the state should be stored gzipped as opposed to
```
apparentlymart: Sorry that this is a bit of a nitpick... feels like this should be called just `gzip` since it's implied that we're talking about state by the context here.

Since I'm already nit-picking, here's another: since remote state properties don't natively support booleans (they're always strings) I think it would be helpful to be specific here about what strings are required. So maybe something like this:

```
* `gzip` - (Optional) `true` to compress the state data using gzip, or `false` (the default) to leave it uncompressed.
```

apparentlymart:
Thanks for this, @christoe! This looks good from an implementation standpoint, and so I just left a couple inline nits about naming/docs. I'm happy to help make those updates if you'd prefer.

christoe:
Hi @apparentlymart! I've updated the PR according to your comments.

carlpett:
@apparentlymart Any update on merging this? I've also hit the maximum size of the Consul remote state engine, and unless we start splitting everything into _very_ granular parts, we're stuck and can't make any further changes to our AWS infrastructure...

apparentlymart:
Hi @christoe! Sorry for the delay in getting back to you on this.

I was just trying this out in my local dev environment. It seems to work great except for an odd interaction I couldn't quite get to work:

I was trying the two "migration" steps, going *from* non-gzip *to* gzip, and then going back the other way. In short, it seems like this doesn't work quite right:

* Configuring remote state with `-backend-config="gzip=true"` (keeping in mind that there is an existing state present at the given key which is *not* gzipped)
* Terraform creates `.terraform/terraform.tfstate` *without* the `gzip` flag set in the remote config, because it's just bringing the stored remote state in verbatim.
* Do any operation that causes Terraform to update the remote state.
* The state file in Consul is still not gzipped.

The same seems to be true in reverse too. I expect this is because when Terraform discovers an already-existing remote state it wants to cache it locally without modifying it, and that state file has the original remote config embedded in it. This would also be problematic if e.g. the hostname of the Consul server changed, since after read the remote config would still refer to the old hostname.

The only workaround I could find here was to entirely disable remote state with `terraform remote config -disable`, then delete the underlying Consul key (!) before re-enabling it with the `gzip=true` flag set.

This bug in core Terraform ought not to be a blocker for merging this -- it's a bug that affects more than just this situation, after all -- but I think we'll need to add a note about this to the docs to explain since I'm pretty sure people are going to try to turn this on for existing remote states. I'll try to write something reasonable in there now and see about getting this merged.

apparentlymart:
@christoe, thinking about this some more:

What do you think about just automatically enabling gzip on write if we detect that the state file is larger than 512KB? That way it's handled automatically for the user and they won't need to pre-emptively enable it in case the state happens to get too large later.

This would mean we wouldn't need the `gzip=true` setting and all of the annoyance around setting it that I mentioned in my previous comment.

What do you think? Would this alternative approach suit your use-case?

jason-archidera:
Any ideas on when this will be merged?

stack72:
Hi @christoe 

This LGTM - @apparentlymart I am going to merge this in it's current state - we can take your thoughts as a followup if that's ok?

Paul

stack72:
ok, I take this back - this is going to need to be transferred to the new `backends` work for 0.9

Paul

christoe:
@stack72, I now transferred the code to the new `backends` structure, implementing @apparentlymart's suggestion in the same go (removed configuration and made terraform automatically select whether to gzip or not).

jbardin:
Thanks @christoe, this looks good.

My only concern is that the upgrade path where past versions of terraform will fail to read a compressed state. I'm fairly certain we allow reading state from future versions, which otherwise would force teams coordinating via remote state to upgrade in unison. 

I think we may have to go back to making this a config flag so it can be opt-in.

christoe:
@jbardin: Ok, made it configurable again.

jbardin:
@christoe,

Thanks for putting the work into port this! I want to fix up the consul unit tests to some some better verification first, but this LGTM, and I'll take care of merging it soon.

apparentlymart:
@jbardin Terraform already rejects state from future versions for safety reasons, though admittedly the error message would not be so helpful in this case since it would presumably be a JSON parse error rather than actually talking about Terraform versions.

Seems to me though that this temporary inconvenience is worth it for usability in the long run.

jbardin:
@apparentlymart, you can use a remote state as a datasource with an older tf version (we should probably have some restriction there though). Coordinating upgrades in disparate teams using remote states could be difficult otherwise.

I think this could follow a standard deprecation cycle, eventually enabling it by default and removing the option.

apparentlymart:
I guess a little late for this comment now, but my suggestion was to enable gzip automatically if the state gets bigger than what can be stored in Consul, so it isn't physically possible for there to be an uncompressed version created by an earlier Terraform version because it would've been rejected by Consul anyway.

PR #6980, opened 2016-06-02
[WIP] Added Codecommit Trigger

optimisticanshul:
Adds #6973

PR #8751, opened 2016-09-09
Feature CodeCommit Trigger

### LINKED ISSUE #6973: AWS: aws_codecommit_repository triggers
hi hi,

Currently, it is possible to create an AWS CodeCommit repository easily through terraform, but adding [triggers](https://docs.aws.amazon.com/codecommit/latest/userguide/how-to-notify.html) does not seem to be possible. Can these be added as AWS resources?

### Terraform Version
```
 bin/terraform -v
Terraform v0.6.16
```

### Affected Resource(s)
- aws_codecommit_repository

Spitballing an example: 

```hcl
resource "aws_lambda_function" "notify" {
  filename = "codecommit-lambda.zip"
  function_name = "codecommit_to_${var.repo_id}"
  role = "${aws_iam_role.iam_for_lambda.arn}"
  handler = "exports.call"
  source_code_hash = "${base64sha256(file("lambda_function_payload.zip"))}"
}

resource "aws_codecommit_repository" "test" {
  repository_name = "${var.repo_name}"
  description = "${var.repo_description}"
}

resource "aws_codecommit_trigger" "lambda" {
  lambda = "${lambda.arn}"
  "branches" = [ "master", "preprod"  ],
  "events": ["all"]
}
```

optimisticanshul:
Fixes #6973
I deleted my original repo so based on discussion #6980 

/cc @stack72

### DIFF
--- a/builtin/providers/aws/resource_aws_codecommit_trigger.go
+++ b/builtin/providers/aws/resource_aws_codecommit_trigger.go
@@ -1,0 +1,163 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/codecommit"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCodeCommitTrigger() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCodeCommitTriggerCreate,
+		Read:   resourceAwsCodeCommitTriggerRead,
+		Delete: resourceAwsCodeCommitTriggerDelete,
+
+		Schema: map[string]*schema.Schema{
+			"repository_name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"configuration_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"trigger": &schema.Schema{
+				Type:     schema.TypeSet,
+				ForceNew: true,
+				Required: true,
+				MaxItems: 10,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"destination_arn": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"custom_data": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+							ForceNew: true,
+						},
+
+						"branches": &schema.Schema{
+							Type:     schema.TypeList,
+							Optional: true,
+							ForceNew: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+
+						"events": &schema.Schema{
+							Type:     schema.TypeList,
+							Required: true,
+							ForceNew: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceAwsCodeCommitTriggerCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codecommitconn
+
+	// Expand the "trigger" set to aws-sdk-go compat []*codecommit.RepositoryTrigger
+	triggers := expandAwsCodeCommitTriggers(d.Get("trigger").(*schema.Set).List())
+
+	input := &codecommit.PutRepositoryTriggersInput{
+		RepositoryName: aws.String(d.Get("repository_name").(string)),
+		Triggers:       triggers,
+	}
+
+	resp, err := conn.PutRepositoryTriggers(input)
+	if err != nil {
+		return fmt.Errorf("Error creating CodeCommit Trigger: %s", err)
+	}
+
+	log.Printf("[INFO] Code Commit Trigger Created %s input %s", resp, input)
+
+	d.SetId(d.Get("repository_name").(string))
+	d.Set("configuration_id", resp.ConfigurationId)
+
+	return resourceAwsCodeCommitTriggerRead(d, meta)
+}
+
+func resourceAwsCodeCommitTriggerRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codecommitconn
+
+	input := &codecommit.GetRepositoryTriggersInput{
+		RepositoryName: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetRepositoryTriggers(input)
+	if err != nil {
+		return fmt.Errorf("Error reading CodeCommit Trigger: %s", err.Error())
+	}
+
+	log.Printf("[DEBUG] CodeCommit Trigger: %s", resp)
+
+	return nil
+}
+
+func resourceAwsCodeCommitTriggerDelete(d *schema.ResourceData, meta interface{}) error {
+
+	conn := meta.(*AWSClient).codecommitconn
+
+	log.Printf("[DEBUG] Deleting Trigger: %q", d.Id())
+
+	input := &codecommit.PutRepositoryTriggersInput{
+		RepositoryName: aws.String(d.Get("repository_name").(string)),
+		Triggers:       []*codecommit.RepositoryTrigger{},
+	}
+
+	_, err := conn.PutRepositoryTriggers(input)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+		Create: resourceAwsCodeCommitTriggerCreate,
+		Read:   resourceAwsCodeCommitTriggerRead,
+		Delete: resourceAwsCodeCommitTriggerDelete,
+
+		Schema: map[string]*schema.Schema{
+			"repository_name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
```
stack72: As this is just a reference to a repository, should we validate here? Isn't the validation in the `aws_code_commit` resource? :)

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+					}
+					return
+				},
+			},
+
+			"trigger": &schema.Schema{
+				Type:     schema.TypeSet,
+				ForceNew: true,
+				Required: true,
+				MaxItems: 10,
```
stack72: There can be 10 triggers?

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+func resourceAwsCodeCommitTriggerCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codecommitconn
+	region := meta.(*AWSClient).region
+
+	//	This is a temporary thing - we need to ensure that CodeCommit is only being run against us-east-1
+	//	As this is the only place that AWS currently supports it
+	if region != "us-east-1" {
```
stack72: Let's remove this now - we should add a note to the docs pointing to where it is used. This will mean that when it is rolled to another region, users can't use it. Please also remove from `aws_code_commit` as well :)

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+		return fmt.Errorf("CodeCommit can only be used with us-east-1. You are trying to use it on %s", region)
+	}
+
+	// Expand the "trigger" set to aws-sdk-go compat []*codecommit.RepositoryTrigger
+	rawTriggers := d.Get("trigger").(*schema.Set).List()
```
stack72: can we wrap this in a func? something like expandAwsCodeCommitTriggers would be a good name here

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+	triggers := make([]*codecommit.RepositoryTrigger, 0, len(rawTriggers))
+	for _, cors := range rawTriggers {
+		corsMap := cors.(map[string]interface{})
+		r := &codecommit.RepositoryTrigger{}
+		for k, v := range corsMap {
+			if k == "custom_data" {
```
stack72: would a case statement be better than if /else?

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+
+func resourceAwsCodeCommitTriggerRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codecommitconn
+
+	input := &codecommit.GetRepositoryTriggersInput{
+		RepositoryName: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetRepositoryTriggers(input)
+	if err != nil {
```
stack72: Is there a chance that the triggers can be manually deleted from the console? IF so, what happens here?

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+		return fmt.Errorf("Error reading CodeCommit Trigger: %s", err.Error())
+	}
+
+	log.Printf("[INFO] CodeCommit Trigger READ %s", resp)
+
+	return nil
```
stack72: We need to set the details back to state if possible

builtin/providers/aws/resource_aws_codecommit_trigger.go
```diff
@@ -0,0 +1,170 @@
...
+
+func resourceAwsCodeCommitTriggerDelete(d *schema.ResourceData, meta interface{}) error {
+
+	conn := meta.(*AWSClient).codecommitconn
+
+	log.Printf("[DEBUG] Deleting Trigger: %q", d.Id())
+
+	input := &codecommit.PutRepositoryTriggersInput{
+		RepositoryName: aws.String(d.Get("repository_name").(string)),
+		Triggers:       []*codecommit.RepositoryTrigger{},
```
stack72: ? what is this for?

stack72: Oh i see what you are doing - to remove the triggers, you need to Put an empty list

builtin/providers/aws/resource_aws_codecommit_trigger_test.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_codecommit_trigger" {
+			continue
+		}
+
+		_, err := conn.GetRepositoryTriggers(&codecommit.GetRepositoryTriggersInput{
+			RepositoryName: aws.String(rs.Primary.ID),
+		})
+
+		if ae, ok := err.(awserr.Error); ok && ae.Code() == "RepositoryDoesNotExistException" {
```
stack72: Can we add the same check to the REad func in the main resource? This seems a valid error to guard against

builtin/providers/aws/resource_aws_codecommit_trigger_test.go
```diff
@@ -0,0 +1,160 @@
...
+			RepositoryName: aws.String(rs.Primary.ID),
+		})
+
+		if err != nil {
+			return err
+		}
+
+		log.Printf("[INFO] TriggerMetadata %s", rs.Primary.ID)
+
+		if *out.Triggers[0].Name != rs.Primary.ID {
```
stack72: Is out.Triggers[0].Name the name of the repository? Or is it the name of a trigger?

stack72:
Hi @optimisticanshul 

This is looking good - i have left a few comments inline - nothing bad at all

This will also need some doumentation as well please :)

Paul

builtin/providers/aws/resource_aws_codecommit_trigger.go
optimisticanshul: You can create up to 10 unique triggers in response to repository events, such as creating or deleting a branch.

builtin/providers/aws/resource_aws_codecommit_trigger_test.go
optimisticanshul: repository_name

builtin/providers/aws/resource_aws_codecommit_trigger_test.go
optimisticanshul: There is an already err condition that checks for error

builtin/providers/aws/resource_aws_codecommit_trigger.go
optimisticanshul: If there is no trigger it doesn't return any error. Using Api we can't look for exact trigger it only returns list of triggers.

optimisticanshul:
@stack72 Hi i think i took care of most of your comments, so i thinks it's ready for another review :)

radeksimko:
@optimisticanshul there's a forgotten import which is why Travis is reporting 🔴 :
```
builtin/providers/aws/resource_aws_codecommit_trigger_test.go:5: imported and not used: "log"
```

optimisticanshul:
@radeksimko Updated thanks.

stack72:
Hi @optimisticanshul 

This LGTM! I made 1 change - I removed the test for changes. All it does is just force a new resource so it's fine to be covered by the basic

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCodeCommitTrigger_'                                                                                                                  ✹
2016/09/12 10:52:39 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCodeCommitTrigger_ -timeout 120m
=== RUN   TestAccAWSCodeCommitTrigger_basic
--- PASS: TestAccAWSCodeCommitTrigger_basic (22.56s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	22.578s
```

PR #8762, opened 2016-09-10
Vagrantfile improvement (apt-get)

visit1985:
This pull does the following:

- runs `apt-get` in non-interactive mode, to prevent errors with `dpkg-preconfigure` while `apt-get upgrade`
- do not install *recommended* packages automatically, to prevents the unnecessary install of X11

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,7 @@ SRCROOT="/opt/go"
 SRCPATH="/opt/gopath"
 
 # Get the ARCH
-ARCH=`uname -m | sed 's|i686|386|' | sed 's|x86_64|amd64|'`
+ARCH="$(uname -m | sed 's|i686|386|' | sed 's|x86_64|amd64|')"
 
 # Install Prereq Packages
 export DEBIAN_PRIORITY=critical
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -13,9 +13,9 @@ SRCPATH="/opt/gopath"
 ARCH=`uname -m | sed 's|i686|386|' | sed 's|x86_64|amd64|'`
 
 # Install Prereq Packages
-sudo apt-get update
-sudo apt-get upgrade -y
-sudo apt-get install -y build-essential curl git-core libpcre3-dev mercurial pkg-config zip
+sudo DEBIAN_FRONTEND=noninteractive apt-get update
```
kwilczynski: You could export `DEBIAN_FRONTEND=noninteractive` this once, I believe.

In fact, for ultimate unattended experience, you might want this:

```
export DEBIAN_PRIORITY=critical
export DEBIAN_FRONTEND=noninteractive
export DEBCONF_NONINTERACTIVE_SEEN=true
```

kwilczynski:
@visit1985 a very good point, although it would keep environment if you add `-E` there where needed.

I was actually wondering why this is even needed to have the `sudo` there in the first place, but then I noticed the `privileged: false` set for the shell provisioner which is a bit odd choice. Not like anything that is executed there really matters after the provisioner finishes, plus it needs to be root anyway to move files around. I personally would re-factor to avoid setting `privileged: false` to false.

Anyhow. You can add all the things I suggested and then add `-E` where needed, or just go with your version, as this would be very much the apple vs oranges thing here.

visit1985:
Let me propose a version without sudo.

visit1985:
Please review again.

cblecker:
Tested and LGTM 👍

kwilczynski:
@visit1985 thank you so much!

visit1985:
@cblecker @kwilczynski may I ask when this is merged? I'd like to propose another PR on top of that :grin:

cblecker:
@visit1985 Hopefully a maintainer can review and merge -- I'm just a public contributor :)

visit1985:
@radeksimko can you please review/merge?

cblecker:
@jen20 --
Is it possible to get this PR reviewed? I think this is a great overall improvement to the Vagrantfile

Thanks so much!

stack72:
I am going to go ahead and merge this - sorry it has taken so long :)

visit1985:
thanks @stack72 :grin:

Issue #8406, opened 2016-08-23
No way to use data "template_file" on a file generated on the fly by local-exec as it fails during planning


bloo:
### Terraform Version

Terraform v0.7.1

### Affected Resource(s)

- null_resource
- data "template_file"

### Terraform Configuration Files
```hcl
variable etcd_cluster_size {
  default = 3
}

resource "null_resource" "discovery_url_get" {
  provisioner "local-exec" {
    command = "curl -s 'https://discovery.etcd.io/new?size=${var.etcd_cluster_size}' > /tmp/discovery.url"
  }
}

data "template_file" "discovery_url" {
  template = "${file("/tmp/discovery.url")}"
  depends_on = ["null_resource.discovery_url_get"]
}
```

### Debug Output
```sh
Errors:

  * file: open /tmp/discovery.url: no such file or directory in:

${file("/tmp/discovery.url")}
```

### Expected Behavior

I (perhaps incorrectly) assumed that `terraform plan` wouldn't require that a file referenced by `file()` needed to exist during the plan stage, but would fail on it being missing during `terraform apply`.

I suppose this could be considered intended behavior, and that a better solution would be to have something like `tmp_file()` as a `file()` alternative, or a core way to capture a `local-exec` output into a variable instead of a temp file?

### Actual Behavior

`terraform plan` fails on not finding a file that had yet to be generated by a dependent `null_resource`.

### Steps to Reproduce

1. `terraform plan`

mitchellh:
This is _kind of_ working as intended but there are definite buggy behavior in here.

Data sources cannot depend on resources (we should've given you an error, we didn't, that's a bug) since data sources must be able to run prior to any resources. This is the point of data source: to give you data that is external to Terraform that isn't dependent on the ordering of Terraform's execution.

At least, that's how they work in their current form. 

I think the primary issue here is that we didn't give you a config validation error. There are other issues like #8144 (referenced) that aim to solve your use case.

apparentlymart:
@mitchellh what you described there isn't *entirely* correct: a data source *can* depend on a resource, but that causes its resolution to be deferred until the apply phase. In this case they appear in the plan output as "read" actions, with the `<=` icon.

I think the issue here is more that interpolation functions are dealt with during the validation pass as well as other passes, but during validation the file will of course not exist yet.

Using the `filename` argument to `template_file` was the traditional workaround for this problem, though of course we've deprecated it. :grinning:

[...]

PR #8768, opened 2016-09-10
"external" data source, for integrating with external programs

apparentlymart:
This is an implementation of the data source part of was proposed in #8144, with some modifications.

After playing with the design some more I made the following adjustments compared to what I originally proposed:
- The data source and resource will be called just `external`, since `data "external_data_source"` (etc) felt redundant. Now you can simply say `data "external" "foo"`, though this required a small tweak to core to allow a provider to export a resource whose name exactly matches the provider name.
- Rather than having separate `program` and `interpreter` arguments, it instead expects `program` to be a list whose first element is the program (which might be an interpreter) and whose subsequent elements are arguments. An array is used so that expressions can be safely interpolated into arguments without needing to worry about the complicated business of escaping spaces and shell metacharacters.

Implementation Progress:
- [x] Provider
  - [x] Documentation
- [x] Core tweak to allow the resources to be called just `"external"`
- [x] Data Source
  - [x] Tests
  - [x] Documentation

Since the resource is more complex, I'd like to land the data source first.


apparentlymart:
Sorry for leaving this sitting here unfinished for so long.

@mitchellh I'm curious as to what you think of this idea philosophically. @radeksimko made some good points over in my longer proposal issue #8144 and I attempted to address them over here. While there is some risk here that this will be used as a crutch, I think it's more likely to just give people a more robust alternative to existing weird patterns using the `remote_exec`provisioner, as was discussed over in #8406.

---

Out of curiosity I implemented the example over in #8406 to be an external data source glue program written in `bash` with `jq`:

```bash
#!/bin/sh
set -ue
eval "$(jq -r '@sh "CLUSTER_SIZE=\(.cluster_size)"')"
DISCOVERY_URL="$(curl -s "https://discovery.etcd.io/new?size=$CLUSTER_SIZE")"
jq -n --arg disco_url "$DISCOVERY_URL" '{"discovery_url":$disco_url}'
```

Associated Terraform config:

```hcl
variable "etcd_cluster_size" {
  default = 3
}

data "external" "test" {
  program = ["${path.module}/etcd_discovery.sh"]

  query {
    cluster_size = "${var.etcd_cluster_size}"
  }
}

output "discovery_url" {
  value = "${data.external.test.result["discovery_url"]}"
}
```

Running it:

```
$ terraform apply
data.external.test: Refreshing state...

Outputs:

discovery_url = https://discovery.etcd.io/c88691558304d9f9171ca0c98e238637
```

mitchellh:
@apparentlymart My personal opinion is that this is both worth having and extremely dangerous. I'm all about having "escape hatches" so you're not 100% forced to work within the confines of the tool (I see JSON configs as one of those for example). 

I think we just need to do our part to very actively put a yellow-border warning on the docs for this provider/resource that explains the cost of using this: lack of portability, may not work with Terraform Enterprise without additional work, etc. It puts a lot of burden on the user but at the same time could let them do things w/o being beholden to the core team here on Terraform.

Conclusion:  I'm :+1: with heavy docs/warnings.

apparentlymart:
Here's what I have in a yellow box in the docs right now:

> **Warning** This mechanism is provided as an "escape hatch" for exceptional situations where a first-class Terraform provider is not more appropriate. Its capabilities are limited in comparison to a true data source, and implementing a data source via an external program is likely to hurt the portability of your Terraform configuration by creating dependencies on external programs and libraries that may not be available (or may need to be used differently) on different operating systems.

I didn't mention Terraform Enterprise here, but that's definitely worth noting. Are there any guarantees for things other than Terraform being available in the environment where Terraform Enterprise runs these things, or is it the case that this is basically useless in Enterprise unless you want to ship a statically-linked language runtime up along with the configs?

stack72:
Hey @apparentlymart 

I like this and agree with both your's and @mitchellh's points above. The code itself is sound and the tests pass

```
% make testacc TEST=./builtin/providers/external
2016/11/30 14:07:51 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/external -v  -timeout 120m
=== RUN   TestDataSource_basic
--- PASS: TestDataSource_basic (0.29s)
=== RUN   TestDataSource_error
--- PASS: TestDataSource_error (0.07s)
=== RUN   TestProvider
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/external	0.377s
```

So I am going to leave the final yes / no to the man above //cc @mitchellh 

P.

apparentlymart:
Sorry when I was skimming this again I missed my own "note-to-self" about adding the statement about Terraform Enterprise to the docs. I'll work on that now, just to get this finished up.

apparentlymart:
I've now rebased and added some additional yellow warning boxes to the docs to talk about Terraform Enterprise specifically, though of course I would encourage reading what I wrote there since I don't really know what guarantees (if any) the Terraform Enterprise environment provides and I may be inadvertently using terminology inconsistent with the language within that product.

stack72:
@apparentlymart we just spoke about this and it LGTG now :)

Thanks!

Paul

Issue #7930, opened 2016-08-03
terraform updates aws_cloudfront_distribution every time over and over again


BerndWessels:
Hi there,

### Terraform Version
Terraform v0.6.16

### Affected Resource(s)
- aws_cloudfront_distribution

It may be an issue with Terraform's core.

### Terraform Configuration Files
```hcl
/**
 * AWS region and credentials.
 */
provider "aws" {
  region = "${var.aws_region}"
  access_key = "${var.aws_access_key}"
  secret_key = "${var.aws_secret_key}"
}
/**
 * Set of nameservers used for the platform domain.
 */
resource "aws_route53_delegation_set" "PlatformDelegationSet" {
  reference_name = "${var.environment}"
}
/**
 * Hosted zone used for the platform domain.
 */
resource "aws_route53_zone" "PlatformHostedZone" {
  name = "${var.platform_domain}"
  delegation_set_id = "${aws_route53_delegation_set.PlatformDelegationSet.id}"
}
/**
 * CNAME record used by the email provider to verfify the domain ownership.
 */
resource "aws_route53_record" "PlatformMailVerificationRecord" {
  zone_id = "${aws_route53_zone.PlatformHostedZone.zone_id}"
  name = "${var.email_domain_verification_cname_name}"
  type = "CNAME"
  ttl = "300"
  records = [
    "${var.email_domain_verification_cname_value}"]
}
/**
 * MX record pointing to the email provider.
 */
resource "aws_route53_record" "PlatformMailMxRecord" {
  zone_id = "${aws_route53_zone.PlatformHostedZone.zone_id}"
  name = "${aws_route53_zone.PlatformHostedZone.name}"
  type = "MX"
  ttl = "300"
  records = [
    "${var.email_mx_record_values.1}",
    "${var.email_mx_record_values.2}"]
}
/**
 * S3 bucket hosting the platform website.
 */
resource "aws_s3_bucket" "PlatformWebsiteBucket" {
  bucket = "${var.platform_domain}"
  acl = "public-read"
  policy = "${file("./Websites/Marketing/BucketPolicy.json")}"
  website {
    index_document = "index.html"
    routing_rules = "${file("./Websites/Marketing/RoutingRules.json")}"
  }
}
/**
 * CloudFront distribution of the platform website.
 */
resource "aws_cloudfront_distribution" "PlatformWebsiteDistribution" {
  aliases = [
    "${var.platform_domain}",
    "www.${var.platform_domain}"]
[...]

PR #8770, opened 2016-09-11
provider/aws: query_string_cache_keys for CloudFront and require origin_access_identity in s3_origin_config complex structure

vancluever:
I've combined these two commits because the tests won't pass without the first commit now (this is new, guessing in the last few days as I was messing with CloudFront a few days ago).

The second commit addresses #7930 by requiring `origin_access_identity` in `s3_origin_config`, if it's specified when creating a distribution. If it's not, then an origin access identity won't be used, but if the origin is S3 it will be configured as an S3 origin.


vancluever:
Closing this in favour of #8777.

PR #8778, opened 2016-09-12
Remove Region Condition From CodeCommit Repo Resource

optimisticanshul:
As discussed in #8751 removing region from codecommit repo
/cc @stack72


radeksimko:
LGTM.

PR #8705, opened 2016-09-07
[WIP] Feature AWS WAF

optimisticanshul:
#6646 

list of resources that need to be added:

- [x] vendor waf
- [ ] aws_waf_byte_match_set
- [x] aws_waf_ip_set
- [x] aws_waf_rule
- [ ] aws_waf_size_constraint_set
- [ ] aws_waf_sql_injection_match_set (could also be aws_waf_sqli_match_set)
- [x] aws_waf_web_acl (which plugs into the CloudFront distribution)
- [ ] aws_waf_xss_match_set

PR #8785, opened 2016-09-12
Vendored WAF

optimisticanshul:
@radeksimko As discussed in #8705 This PR vendor WAF of aws sdk.
I will close the #8705 once we complete all the checkpoints.


stack72:
Hi @optimisticanshul 

Can you add the vendored WAF as of the 1.4.7 release of the SDK? This means we are going to be following the same release as the other pieces of the SDK

Thanks

Paul

optimisticanshul:
Hi Paul sorry for stupid question but can you please elaborate what needs
to be done here.

apparentlymart:
Hi @optimisticanshul,

I believe @stack72 was suggesting to upgrade the existing vendoring of the whole of the AWS SDK to 1.4.7, rather than just adding the WAF service package within the version we were already running.

I'm not sure if there are some specific steps to deal with this for the AWS provider, but usually when upgrading packages we'd do something like this:

```
govendor fetch github.com/aws/aws-sdk-go
```

We like to upgrade upstream libraries in whole so that we don't get any unexpected friction between the parts and so we can more easily reason about what version we're currently running compared to the upstream.

stack72:
Hi @optimisticanshul (and @apparentlymart)

The command I run for a specific tag is as follows:

```
 govendor fetch github.com/aws/aws-sdk-go/...@v1.4.7
```

If you can, please ensure the WAF vendor updates are of the same tag 

Paul

optimisticanshul:
@stack72 Updated ready for another review 👍

radeksimko:
👍

PR #8797, opened 2016-09-12
Make "Edit this page" purple.

paddyforan:
Make the "Edit this page" link at the bottom of the documentation pages
the purple highlight colour, to better draw attention to it.

### DIFF
--- a/website/source/assets/stylesheets/_footer.scss
+++ b/website/source/assets/stylesheets/_footer.scss
@@ -53,7 +53,7 @@ body.page-sub{
 
   a{
     text-transform: uppercase;
-    color: $black;
+    color: $purple;
     font-size: 13px;
   }
 }


paddyforan:
Also, to follow guidelines:

* This came up when talking to @stack72 today, and was a small enough PR I felt comfortable making it. The idea is to make the link a little bit more noticeable. It could also serve as a partial fix for #8798 
* This probably isn't worth an immediate deploy, and deploying with the next version would be OK.

jen20:
This seems reasonable to me - perhaps @captainill has some additional feedback however?

captainill:
It was a deliberate decision to make the 'EDIT THIS PAGE' link not stand out because it's on every page. It could detract a user's attention from the content and become a visual nuisance. I think the content of the page is most important and the edit link is there as a functional addition which is probably never going to be clicked by the majority of visitors. Those users who do spend a lot of time in the docs and are making changes would probably be aware of it presence. I'm making assumptions...

Open to discussion of course as this is the first I've given my opinion on it. Our other projects use the same thinking on the style.

paddyforan:
I think that's a valid take on it. I, personally, missed it until it was pointed out to me. #8798 may have contributed to or been the cause of that. If it's deemed worthwhile, I suppose it's something that could be A/B tested pretty easily, but I'm happy to shelve the PR in favour of a higher priority target.

stack72:
@paddyforan I am going to close this out :) Now that you are part of the team, you can talk to @captainill directly about this

paddyforan:
:+1:

PR #8809, opened 2016-09-13
aws_instance doc: fix attachment resource name

moio:
Correct name is `aws_volume_attachment`.


stack72:
Hi @moio 

Thanks for the work here :) This LGTM!

Paul

PR #8817, opened 2016-09-13
Warn when plugins are locally overridden

Jonty:
It is currently very easy to miss that a provider/provisioner has a local override in your `.terraformrc`, leading to unexpected behaviour that can be difficult to debug as no information is output about the overrides.

### DIFF
--- a/config.go
+++ b/config.go
@@ -169,12 +169,18 @@ func (c1 *Config) Merge(c2 *Config) *Config {
 		result.Providers[k] = v
 	}
 	for k, v := range c2.Providers {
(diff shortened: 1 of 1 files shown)


jen20:
Hi @Jonty! In principle I agree with this - we already warn if a binary on PATH overrides a provider. However, I'd argue in this case since a user has specified this explicitly we should log at INFO rather than WARN. I agree it's definitely of value during debugging though, so happy to land this.

Jonty:
@jen20 That seems reasonable. I wasn't sure how noisy TF is about things like this. Thanks for merging & switching it to INFO!

PR #8830, opened 2016-09-13
Add manta remote state backend

watters:
I tried to match what the other backends offered in terms of documentation and testing, but any/all feedback is welcome.

I've marked this [WIP] for now while I continue to exercise it, but I wanted to invite any feedback early on.

### DIFF
--- a/state/remote/manta.go
+++ b/state/remote/manta.go
@@ -2,14 +2,16 @@ package remote
 
 import (
 	"crypto/md5"
+	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 
 	joyentclient "github.com/joyent/gocommon/client"
 	joyenterrors "github.com/joyent/gocommon/errors"
-	"github.com/joyent/gocommon/jpc"
 	"github.com/joyent/gomanta/manta"
+	joyentauth "github.com/joyent/gosign/auth"
 )
 
 const DEFAULT_OBJECT_NAME = "terraform.tfstate"
@@ -25,12 +27,8 @@ func mantaFactory(conf map[string]string) (Client, error) {
 		objectName = DEFAULT_OBJECT_NAME
 	}
 
-	keyName, ok := conf["keyName"]
-	if !ok {
-		keyName = ""
-	}
+	creds, err := getCredentialsFromEnvironment()
 
-	creds, err := jpc.CompleteCredentialsFromEnv(keyName)
 	if err != nil {
 		return nil, fmt.Errorf("Error getting Manta credentials: %s", err.Error())
 	}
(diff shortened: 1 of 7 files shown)


jen20:
Hi @watters! I strongly approve of this PR, obviously! I'll look at getting it tested and merged into Terraform shortly.

jen20:
Hi @watters - I've made a quick review of this, but can't figure out via the various layers of indirection which environment variables need to be set to make the tests run here? Can you shed some light on this (and maybe fix up the "MANTA_USER and friends" comment to reflect that 😆)? Thanks!

watters:
@jen20 thanks so much for taking a look at this so soon!

I've [updated the skip message](https://github.com/hashicorp/terraform/pull/8830/commits/7bedd2b263ffcf032455eea2eaec6d74bba87e2b#diff-5d74442920598cc5e7c23e6365281202R24) in the tests to be explicit about all [three Manta environment variables](https://apidocs.joyent.com/manta/#setting-up-your-environment) that must be set.

I also added a [link to the Manta documentation for setting up the environment](https://github.com/hashicorp/terraform/pull/8830/commits/7bedd2b263ffcf032455eea2eaec6d74bba87e2b#diff-c81220b64c12626593b550dd6812c26fR42) in the website documentation.

When running `make test TEST=./state/remote TESTARGS=-v` with the `MANTA_USER, MANTA_KEY_ID and MANTA_URL` environment variables set, the tests pass and I can see the calls to Manta being made when I inspect the traffic.

Let me know if there are any other changes you'd like to see.

jen20:
Hi @watters - those changes look good to me. Running the tests I'm still seeing "Error getting Manta credentials: An error occurred while parsing the key: asn1: structure error: length too large" however -  the key is in the agent and available unencrypted on disk, though it's unclear where it would be loaded from?

watters:
I'll see if I can repro.

That error is coming from [the underlying auth bits in `github.com/joyent/gosign`] (https://github.com/joyent/gosign/blob/master/auth/auth.go#L61), which I didn't mess with—yet (as discussed elsewhere, the underlying libraries in play can be unwieldy).

`keyName`, which defaults to `$HOME/.ssh/id_rsa` currently cannot be overridden for the tests, which is probably something that needs solving.

state/remote/manta.go
```diff
@@ -80,3 +78,47 @@ func (c *MantaClient) Put(data []byte) error {
...
+	keyMaterial := os.Getenv("MANTA_KEY_MATERIAL")
+
+	if _, err := os.Stat(keyMaterial); err == nil {
+		// key material is a file path; try to read it
+		keyBytes, err := ioutil.ReadFile(keyMaterial)
+		if err != nil {
+			return nil, fmt.Errorf("Error reading key material from %s: %s",
+				keyMaterial, err)
+		} else {
+			block, _ := pem.Decode(keyBytes)
```
watters: This was cribbed directly from https://github.com/mitchellh/packer/blob/master/common/ssh/key.go#L27

watters:
@jen20 I was able to reproduce the error message you described when using a passphrase-protected 4096-bit SSH key. The underlying auth code chokes on passphrase-protected keys and currently only supports keys specified by file path or decrypted into an environment variable.

I've exposed the key path via environment variable `MANTA_KEY_MATERIAL`, which will allow you to specify a key that can be used during the tests (previously it defaulted to `~/.ssh/id_rsa`), but it must align with the `MANTA_KEY_ID` and, of course, be associated with your account.

Hopefully, this gets the tests passing for you. I'd already determined before submitting this PR that Joyent's Go library for Manta probably needed a re-write, and now I'm only even more convinced.

jen20:
That explains a lot - I have a separate unencrypted key for testing with HashiCorp's Joyent account rather than my personal (encrypted) key, so `~/.ssh/id_rsa` producing that error message make sense. I'll test it with `MANTA_KEY_MATERIAL` set now. The Go library for Manta is a bit janky right now - I suspect it was written before the idioms for Go had become apparent. I'll look at merging this now though and we can sort that out in future.

state/remote/manta.go
jen20: Yup, this is working now.

state/remote/manta.go
```diff
@@ -0,0 +1,124 @@
...
+			if block == nil {
+				return nil, fmt.Errorf(
+					"Failed to read key material '%s': no key found", keyMaterial)
+			}
+
+			if block.Headers["Proc-Type"] == "4,ENCRYPTED" {
```
jen20: This is great - we should roll it out into the Triton provider as well.

jen20:
Example test run:

```
=== RUN   TestMantaClient_impl
--- PASS: TestMantaClient_impl (0.00s)
=== RUN   TestMantaClient
2016/09/14 20:37:13 [DEBUG] New state was assigned lineage "aace983a-9729-4997-9ca0-121b4e2f698c"
2016/09/14 20:37:13 [TRACE] Preserving existing state lineage "aace983a-9729-4997-9ca0-121b4e2f698c"
--- PASS: TestMantaClient (7.28s)
```

jen20:
Hi @watters! I've run the tests and reviewed this and it seems fine. I've rebased to master and squashed some of the commits (leaving the vendoring separate). Finally, I've fixed up the documentation sidebar for remote state to include the Manta option. This has landed in 72a341b and b4eb63d, though GitHub sadly doesn't reflect that here! Thanks for your work!

watters:
Thank you!

PR #8833, opened 2016-09-14
provider/aws: Expose ARN suffix on ALB

### LINKED ISSUE #8808: Ease using AWS ALBs in CloudWatch metrics
Fantastic to see ALBs available in Terraform so quickly. Thank you! 👏 

Part of cutting over to ALBs for us involves creating equivalent CloudWatch alarms around our upstream health. The dimensions aren't just the `LoadBalancerName` set to `aws_elb.*.name` like ELBs however. For ALBs the dimension is `LoadBalancer` with a partial suffix from the ALB ARN:

```hcl
# Alarm if this alb's targets have more than 50 5xx errors within two minutes
resource "aws_cloudwatch_metric_alarm" "my-alb-too-many-server-errors" {
  alarm_name  = "my-alb-too-many-server-errors"
  namespace   = "AWS/ApplicationELB"
  metric_name = "HTTPCode_Target_5XX_Count"

  dimensions = {
    # CloudWatch wants part of the arn. It's kinda silly.
    # arn:aws:elasticloadbalancing:us-east-1:123456:loadbalancer/app/my-alb/abc123 => app/my-alb/abc123
    LoadBalancer = "${replace("${aws_alb.my-alb.arn}", "/arn:.*?:loadbalancer\\/(.*)/", "$1")}"
  }

  statistic           = "Sum"
  period              = 60
  comparison_operator = "GreaterThanThreshold"
  threshold           = "50"
  evaluation_periods  = 2
}
```

This definitely works, but it's pretty clumsy. Perhaps this ARN suffix could be exposed as a derived attribute of `aws_alb` resources?

jen20:
When creating a CloudWatch Metric for an Application Load Balancer it is neccessary to use the suffix of the ARN as the reference to the load balancer. This commit exposes that as an attribute on the `aws_alb` resource to prevent the need to use regular expression substitution to make the reference.

Fixes #8808.

### DIFF
--- a/builtin/providers/aws/resource_aws_alb.go
+++ b/builtin/providers/aws/resource_aws_alb.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -24,6 +25,11 @@ func resourceAwsAlb() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"arn_suffix": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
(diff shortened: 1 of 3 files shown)


radeksimko:
Functionally this looks good, I'm just thinking that the suffix _may_ be useful for something else in the future - so I'm not sure I'd make the assumption it's only useful for CloudWatch.

TL;DR I'd simply call it `arn_suffix` instead of `cloudwatch_ref` and keep the note in docs about the use case (CloudWatch).

jen20:
That was the other name I was considering - happy to change it. I'll push a new commit up here.

jen20:
@radeksimko I've changed the name and pushed a fix.

catsby:
LGTM

sj26:
Thanks so much!

PR #8671, opened 2016-09-06
provider/aws: Add support for `cloudwatch_logging_options` to AWS Kinesis Firehose Delivery Streams

stack72:
Fixes #7152

Adding support for CloudWatch Logging to Firehose as per the
instructions here -
http://docs.aws.amazon.com/firehose/latest/dev/monitoring-with-cloudwatch-logs.html

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSKinesisFirehoseDeliveryStream_'
==> Checking that code complies with gofmt requirements...
go generate $(go list ./... | grep -v /terraform/vendor/)
2016/09/07 23:02:35 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v
-run=TestAccAWSKinesisFirehoseDeliveryStream_ -timeout 120m
=== RUN   TestAccAWSKinesisFirehoseDeliveryStream_s3basic
--- PASS: TestAccAWSKinesisFirehoseDeliveryStream_s3basic (273.48s)
=== RUN   TestAccAWSKinesisFirehoseDeliveryStream_s3WithCloudwatchLogging
--- PASS: TestAccAWSKinesisFirehoseDeliveryStream_s3WithCloudwatchLogging (200.48s)
=== RUN   TestAccAWSKinesisFirehoseDeliveryStream_s3ConfigUpdates
--- PASS: TestAccAWSKinesisFirehoseDeliveryStream_s3ConfigUpdates (338.16s)
=== RUN   TestAccAWSKinesisFirehoseDeliveryStream_RedshiftConfigUpdates
--- PASS: TestAccAWSKinesisFirehoseDeliveryStream_RedshiftConfigUpdates (835.44s)
=== RUN   TestAccAWSKinesisFirehoseDeliveryStream_ElasticsearchConfigUpdates
--- PASS: TestAccAWSKinesisFirehoseDeliveryStream_ElasticsearchConfigUpdates (2289.74s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws
3937.351s
```

PR #8837, opened 2016-09-14
Support CloudWatch logging with Kinesis Firehose

bvingerh:
Add support for configuring CloudWatch logging with Kinesis Firehose as per http://docs.aws.amazon.com/firehose/latest/APIReference/API_CreateDeliveryStream.html

### DIFF
--- a/builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
+++ b/builtin/providers/aws/resource_aws_kinesis_firehose_delivery_stream.go
@@ -135,6 +135,22 @@ func resourceAwsKinesisFirehoseDeliveryStream() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+
+						"log_enabled": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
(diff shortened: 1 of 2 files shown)


bvingerh:
See also issue https://github.com/hashicorp/terraform/issues/8838

cheesefactory:
Hey @bvingerh, FYI I was also looking into the feature this morning and noticed there is already a PR for this: #8671

bvingerh:
Good to know I'm not the only one in need of the functionality :-)
I don't much care for which pull request gets approved, but I'd like the functionality in an official release sooner rather than later; (and the other guy has even written tests; go for it).

stack72:
Hi @bvingerh 

Thanks for the PR here. If possible, please can you have a look and see if my PR covers your use-case for functionality :) https://github.com/hashicorp/terraform/pull/8671

Paul

bvingerh:
@stack72 your PR seems to cover my use-case, yes.

stack72:
Awesome - closing this in favour of that PR (#8671) so we can get that work taken care of

Paul

stack72:
Thank you so much for all the work put in here!

Issue #8819, opened 2016-09-13
Openstack creating secgroup timeout


ChiefAlexander:
When creating multiple security groups with the Openstack provider I ran into a issue where terraform would fail after 30 seconds.
### Affected Resource(s)

Please list the resources as a list, for example:
- openstack_networking_secgroup_v2
### Terraform Configuration Files

I have 5 created security groups in a test file, just increased the number

``` hcl
resource "openstack_compute_secgroup_v2" "security_group" {
  name = "sec_group"
  description = "This is a security group"
  rule {
    from_port = 22
    to_port = 22
    ip_protocol = "tcp"
    cidr = "0.0.0.0/0"
  }
  rule {
    from_port = 1
    to_port = 65535
    ip_protocol = "udp"
    cidr = "${var.openstack_cidr}"
  }
  rule {
    from_port = 1
    to_port = 65535
    ip_protocol = "tcp"
    cidr = "${var.openstack_cidr}"
  }
  rule {
    from_port = 0
    to_port = 0
    ip_protocol = "icmp"
    cidr = "0.0.0.0/0"
  }
  rule {
    from_port = 8
    to_port = 0
    ip_protocol = "icmp"
    cidr = "0.0.0.0/0"
  }
}

resource "openstack_compute_secgroup_v2" "security_group2" {
  name = "sec_group2"
  description = "This is a security group"
  rule {
    from_port = 22
    to_port = 22
    ip_protocol = "tcp"
    cidr = "0.0.0.0/0"
  }
  rule {
    from_port = 1
    to_port = 65535
    ip_protocol = "udp"
    cidr = "${var.openstack_cidr}"
  }
  rule {
    from_port = 1
    to_port = 65535
    ip_protocol = "tcp"
    cidr = "${var.openstack_cidr}"
  }
  rule {
    from_port = 0
    to_port = 0
    ip_protocol = "icmp"
    cidr = "0.0.0.0/0"
  }
  rule {
    from_port = 8
    to_port = 0
    ip_protocol = "icmp"
    cidr = "0.0.0.0/0"
  }
}
```
### Debug Output

https://gist.github.com/ChiefAlexander/87875d431c5eaeedee699c8340ba47cc
### Expected Behavior

Terraform should not have exited its build
### Actual Behavior

Terraform quits after 30s of trying to create the security groups
### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
### Important Factoids

[...]

PR #8845, opened 2016-09-14
Fixed typo in aws_elb_attachment website docs

lcrisci:
The instance argument is a string and not a list.
It will give you an error if you try to define it as a list.


jen20:
Force merged this as GitHub's checks were misbehaving.

jen20:
Looks like this fixes #8819.

PR #8854, opened 2016-09-15
Clarify how to notify multiple lambdas

a-h:
Clarified that to send notifications to more than one Lambda function requires a single `s3_bucket_notification` element with multiple `lambda_function` configuration elements.

If multiple `s3_bucket_notification` elements, each with a single `lambda_function` element are used, one of the configurations is applied, seemingly at random.


catsby:
Looks good, thanks!

PR #8861, opened 2016-09-15
provider/aws: Add Default Security Group Resource

catsby:
Part of https://github.com/hashicorp/terraform/issues/6093 , this PR adds a new resource `aws_default_security_group` for managing the Default Security Group.

- each VPC has a default security group you cannot destroy (unless you destroy the VPC itself)
- in EC2 Classic, each _region_ has a default security group that you cannot destroy


builtin/providers/aws/resource_aws_default_security_group_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+resource "aws_default_security_group" "web" {
+  ingress {
+    protocol = "6"
+    from_port = 80
+    to_port = 8000
+    cidr_blocks = ["10.0.0.0/8"]
+  }
+
+	tags {
```
jbardin: Just some mixed tabs and spaces in the test HCL here and in the previous const.

catsby: Done!

PR #8862, opened 2016-09-15
Make terraform get to download an URL only once

sarguru:
We have use cases where the same terraform module via git is invoked n - number of times (sometimes in the range of 100). git-cloning them 100s of times makes terraform get too slow. This PR attempts to clone once and link to the downloaded file URL other times.

Building with this and testing seems to make terraform get work fine but there seems to be a problem with plan. I guess work around for this would be to store this info in a file and read that as well when module is loaded.

But before that any advice on how to solve this better or if there are any alternative approaches to this which might work better would be great.

### DIFF
--- a/command/module_storage.go
+++ b/command/module_storage.go
@@ -2,10 +2,14 @@ package command
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/go-getter"
 	"github.com/mitchellh/cli"
 )
+
+// map to store all already downloaded URLs
+var storedModules = make(map[string]string)
 
 // uiModuleStorage implements module.Storage and is just a proxy to output
 // to the UI any Get operations.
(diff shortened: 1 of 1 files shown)


sarguru:
Update: Since the change is in the storage level, plan seems to work fine without the need of writing to file etc.

sarguru:
A gentle ping to see if anyone has looked at it yet.

sarguru:
Just checking back again.

matfra:
That would be nice and greatly reduce Terraform plan time !

jen20:
@mitchellh was the last person to touch this area, it would be good for him to give this a review prior to merge.

mitchellh:
Hi! Yes, this was a explicit decision, I knew it would it make `get` download multiple times but the changes made to do that were fixing Terraform-breaking bugs. It was a tradeoff. I plan to make it download only once again...

Thanks for your work on this PR but unfortunately while this is functional it isn't done in the right place. The `ui_module_storage.go` file contains an implementation of ModuleStorage that should only be used to communicate to the UI the status of getting modules. While it is possible to implement this sort of single-download logic there, it conflates the encapsulation and isn't the right place for the logic.

The proper logic would be somewhere in `config/module` which is where the actual downloading is triggered. If you could modify that to do single download that'd be great. Other feedback: please avoid global variables as well, you should be able to implement this as a struct variable.

Sorry, I hate to close PRs like this. I agree with what it is trying to solve, and hopefully my tips above help point you in the right direction.

sarguru:
thanks @mitchellh . Agreed 👍  . The intention was more to find out the future path for solving the problem rather than just getting this merged.

mitchellh:
Thanks for undestanding. Yes, this needs to get fixed for sure. Like I said, the reason for the change was without it, there were scenarios Terraform didn't work at all (versus taking a long time to work). So the options were: 

  1. Terraform downloads efficiently but breaks in edge cases completely
  2. Terraform works in all known cases, but is really inefficient in some

So the answer was clearly 2, but we'll address that...

sarguru:
@mitchellh @jbardin Was curious on how this works with the Module registries now! Will the modules will still be downloaded from registries repeatedly or are there any new optimizations that prevent this now?

jbardin:
@sarguru, The registry doesn't change much in this case, as far as terraform is concerned it essentially only adds the ability to query module versions. Any optimizations have to be in how terraform stores the modules, but the good news is that we have all new config loader code coming in the next major release which will make this much easier.

sarguru:
@jbardin thanks for the update here!

PR #8891, opened 2016-09-17
Unchanged NestedSets are not returned by DiffFieldReader.

gotascii:
Given a schema such as:
```
schema := map[string]*Schema{
  "list_of_sets_1": &Schema{
    Type: TypeList,
    Elem: &Resource{
      Schema: map[string]*Schema{
        "nested_set": &Schema{
          Type: TypeSet,
          Elem: &Resource{
            Schema: map[string]*Schema{
              "val": &Schema{
                Type: TypeInt,
              },
            },
          },
          Set: func(a interface{}) int {
            m := a.(map[string]interface{})
            return m["val"].(int)
          },
        },
      },
    },
  },
  "list_of_sets_2": &Schema{
    Type: TypeList,
    Elem: &Resource{
      Schema: map[string]*Schema{
        "nested_set": &Schema{
          Type: TypeSet,
          Elem: &Resource{
            Schema: map[string]*Schema{
              "val": &Schema{
                Type: TypeInt,
              },
            },
          },
          Set: func(a interface{}) int {
            m := a.(map[string]interface{})
            return m["val"].(int)
          },
        },
      },
    },
  },
}
```
And a configuration such as:
```
resource "provider_name_resource_name" "test" {
  list_of_sets_1 {
    nested_set {
      val = 1
    }
  }
  list_of_sets_2 {
    nested_set {
      val = 1
    }
  }
}
```
Updating the `val` of the `nested_set` in`list_of_sets_1` from `1` to `2`:
```
resource "provider_name_resource_name" "test" {
  list_of_sets_1 {
    nested_set {
      val = 2
    }
  }
  list_of_sets_2 {
    nested_set {
      val = 1
    }
  }
}
```
Results in `resourceData.Get("list_of_sets_2")` returning an empty `*schema.Set`.

This PR provides a unit test along with a fix. The fix is to check if the state has a value for the field after determining that the value does not exist in the diff. This matches the behavior of the primitive and other field types.

I ran all of the other core tests and they continue to pass, so I think the fix is not inadvertently breaking anything. Would love feedback on how to get this PR to a state where it is mergable! Thanks!

### DIFF
--- a/helper/schema/field_reader_diff.go
+++ b/helper/schema/field_reader_diff.go
@@ -186,6 +186,16 @@ func (r *DiffFieldReader) readSet(
 		}
 	}
 
+	if !exists {
+		result, err := r.Source.ReadField(address)
+		if err != nil {
+			return FieldReadResult{}, err
+		}
+		if result.Exists {
+			return result, nil
+		}
+	}
+
 	return FieldReadResult{
 		Value:  set,
 		Exists: exists,
--- a/helper/schema/field_reader_diff_test.go
+++ b/helper/schema/field_reader_diff_test.go
@@ -12,6 +12,11 @@ func TestDiffFieldReader_impl(t *testing.T) {
 }
 
 func TestDiffFieldReader_NestedSetUpdate(t *testing.T) {
+	hashFn := func(a interface{}) int {
+		m := a.(map[string]interface{})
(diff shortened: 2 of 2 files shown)


phinze:
This looks great - thanks @gotascii!

PR #8895, opened 2016-09-17
Update index.html.markdown

baboune:
Use default docker server remote API port 2376 instead of 1234.


baboune:
While cert_path is indicated as optional, it seems that terraform might require this attribute to be set for on most deployments since docker-machine, as per Docker Remote API documentation, instantiates a Docker daemon that uses an encrypted TCP socket using TLS.  Maybe worth adding this in the doc too? I am unsure please advise.

jen20:
Thanks @baboune! With regards to the certificate path - if you think it is something people commonly encounter then it's probably worth adding a note in the documentation. I'm not familiar enough with the Docker provider to know whether that is the case or not.

PR #8896, opened 2016-09-17
provider/aws: Add JSON validation to the aws_cloudformation_stack resource.

kwilczynski:
This commit adds support for new helper function which is used to
normalise and validate JSON string.

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudformation_stack.go
+++ b/builtin/providers/aws/resource_aws_cloudformation_stack.go
@@ -28,10 +28,14 @@ func resourceAwsCloudFormationStack() *schema.Resource {
 				ForceNew: true,
 			},
 			"template_body": &schema.Schema{
-				Type:      schema.TypeString,
-				Optional:  true,
-				Computed:  true,
-				StateFunc: normalizeJson,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateJsonString,
+				StateFunc: func(v interface{}) string {
+					json, _ := normalizeJsonString(v)
+					return json
+				},
 			},
 			"template_url": &schema.Schema{
 				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


kwilczynski:
Part of work which began in #8028.

kwilczynski:
Acceptance test is passing:

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCloudFormation_'
2016/09/14 23:55:53 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCloudFormation_ -timeout 120m
=== RUN   TestAccAWSCloudFormation_basic
--- PASS: TestAccAWSCloudFormation_basic (82.33s)
=== RUN   TestAccAWSCloudFormation_defaultParams
--- PASS: TestAccAWSCloudFormation_defaultParams (83.21s)
=== RUN   TestAccAWSCloudFormation_allAttributes
--- PASS: TestAccAWSCloudFormation_allAttributes (178.29s)
=== RUN   TestAccAWSCloudFormation_withParams
--- PASS: TestAccAWSCloudFormation_withParams (151.38s)
=== RUN   TestAccAWSCloudFormation_withUrl_withParams
--- PASS: TestAccAWSCloudFormation_withUrl_withParams (227.25s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	722.481s
```

builtin/providers/aws/resource_aws_cloudformation_stack.go
```diff
@@ -69,10 +73,14 @@ func resourceAwsCloudFormationStack() *schema.Resource {
...
-				Optional:  true,
-				Computed:  true,
-				StateFunc: normalizeJson,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateJsonString,
+				StateFunc: func(v interface{}) string {
+					json, _ := normalizeJsonString(v)
+					return json
```
radeksimko: Probably a nit pick - but looking at the code makes me think we should have a tiny function for this which we could simply reference as `StateFunc` to avoid the repetition, but I frankly can't think of a good name for such function 💭 
and maybe the repetition of 2-3 lines isn't that bad after all.

kwilczynski: Good point. I thought about factoring this out, but then I have a wish to replace this with an appropriate `DiffSuppressFunc` function. Therefore, I would consider these not-there-to-stay long term. Also, I would hope that the compiler is going to inline these... :)

PR #8909, opened 2016-09-17
provider/azurerm: Add normalizeJsonString and validateJsonString functions.

kwilczynski:
This commit adds ValidateFunc to the policy attribute so that JSON parsing
errors can be caught early. Generally, when there is a ValidateFunc set for the
attribute, one can safely assume that before any of the creation and/or update
of the existing resource would happen it would have to succeed validation. Also
adds support for new helper function which is used to normalise JSON string.

### DIFF
--- a/builtin/providers/azurerm/structure.go
+++ b/builtin/providers/azurerm/structure.go
@@ -1,0 +1,24 @@
+package azurerm
+
+import "encoding/json"
+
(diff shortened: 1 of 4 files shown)


kwilczynski:
Part of work which began in #8028.

stack72:
LGTM! Thanks for the work here @kwilczynski :)

PR #8913, opened 2016-09-18
provider/rundeck) enable validation for multiple values in an array

BobVanB:
Enable ValidateFunc, copied from another provider:

``` bash
            "rank_order": &schema.Schema{
                Type:         schema.TypeString,
                Optional:     true,
                Default:      "ascending",
                ValidateFunc: validateValueFunc([]string{"ascending", "descending"}),
            },
```

This will be used when rank_order is implemented, but can be used for other schema validations.
This one is handy to have.

### DIFF
--- a/builtin/providers/rundeck/util.go
+++ b/builtin/providers/rundeck/util.go
@@ -1,0 +1,25 @@
+package rundeck
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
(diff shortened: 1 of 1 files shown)


BobVanB:
@radeksimko Is this standing still?

stack72:
Hi @BobVanB 

Thanks so much for the work here - sorry this took a while to get back to! This LGTM!

Paul

PR #8915, opened 2016-09-18
provider/aws: Make it possible to remove S3 bucket policy

radeksimko:
Due to a bug introduced in #8615 it is impossible to remove a policy as the `policy` field is marked as `Computed` and `"existing-policy" => ""` is always a no-op if the field is `Computed`.

There's actually already an acceptance test covering this and [it was failing](https://travis-ci.org/hashicorp/terraform/jobs/157258835#L324) for the past 15 days after merging #8615 , just returning a slightly misleading error message - so I fixed that too in this PR.

I understand the reason for making this field `Computed` was:

1. to expose policy which would be added/managed via `aws_s3_bucket_policy` after refresh
2. to prevent `aws_s3_bucket` from deleting the policy which was set in the context of `aws_s3_bucket_policy` - i.e. prevent those resources from fighting with each other
3. to allow import of bucket with policy

(2) can be still achieved without `Computed` - see my third commit.
(3) can be still achieved without `Computed` - see my last commit of this PR

I remember @stack72 and I discussed recently that there may be some plans in regards to overlapping/nested resources like these - but I don't know what these are - if there are any.

Either way this is now broken, so I think it needs fixing.

### Test plan

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=AWSS3Bucket'
```

### DIFF
--- a/builtin/providers/aws/import_aws_s3_bucket.go
+++ b/builtin/providers/aws/import_aws_s3_bucket.go
@@ -1,0 +1,39 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsS3BucketImportState(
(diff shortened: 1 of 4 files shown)


jen20:
Hi @radeksimko! At first blush this looks like a good set of fixes, I'll review the this morning and get them merged.

PR #8928, opened 2016-09-20
Add heading markdown

arthurbarr:
It looks like the text is written with the intention of it being a heading, but it wasn't marked as such, which meant it read strangely in HTML.


jen20:
Thanks @arthurbarr!

PR #8930, opened 2016-09-20
Added missed link to sphere virtual disk

partamonov:
The link on web site for vsphere virtual disk was missed (https://www.terraform.io/docs/providers/vsphere/index.html)


jen20:
Thanks @partamonov!

PR #8935, opened 2016-09-20
provider/aws: Added a cluster_address attribute to aws elasticache.

A-Gordon:
Added the cluster address as a separate attribute to the configuration endpoint. When using the configuration endpoint in conjunction with route 53 it was appending the cluster address with the port and invalidating the route 53 record.

When using the AWS elasticache client you can specify the cluster address and use Auto discovery, Terraform currently doesn't support this as you have to either specify the cache nodes or manually edit the route 53 record. 
https://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/ClientConfig.AutoDiscovery.html

There is a requirement of the aws elasticache client when using a CNAME to reference the elasticache cluster address, it requires the CNAME to include .cfg. which informs the client that it is a cluster and to treat it as such.

### DIFF
--- a/vendor/github.com/aws/aws-sdk-go/service/elasticache/api.go
+++ b/vendor/github.com/aws/aws-sdk-go/service/elasticache/api.go
@@ -2572,6 +2572,9 @@ type CacheCluster struct {
 	// node.
 	ConfigurationEndpoint *Endpoint `type:"structure"`
 
+	// The DNS name of the cache cluster without the port appended.
+	ClusterAddress *string `type:"string"`
+
 	// The name of the cache engine (memcached or redis) to be used for this cache
 	// cluster.
 	Engine *string `type:"string"`
--- a/builtin/providers/aws/resource_aws_elasticache_cluster.go
+++ b/builtin/providers/aws/resource_aws_elasticache_cluster.go
@@ -170,6 +170,11 @@ func resourceAwsElasticacheCluster() *schema.Resource {
 		Computed: true,
 	}
 
+	resourceSchema["cluster_address"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: true,
+	}
+
 	resourceSchema["replication_group_id"] = &schema.Schema{
 		Type:     schema.TypeString,
 		Computed: true,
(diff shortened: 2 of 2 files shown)


vendor/github.com/aws/aws-sdk-go/service/elasticache/api.go
```diff
@@ -2572,6 +2572,9 @@ type CacheCluster struct {
 	// node.
 	ConfigurationEndpoint *Endpoint `type:"structure"`
 
+	// The DNS name of the cache cluster without the port appended.
+	ClusterAddress *string `type:"string"`
```
stack72: Did you add this manually?

stack72:
Hi @A-Gordon 

Thanks for the work here - I have left a comment about the vendor folder

Paul

vendor/github.com/aws/aws-sdk-go/service/elasticache/api.go
A-Gordon: I did, I'm not sure if that is the correct procedure for changes within the vendor directory.

stack72: Ah! That change will be over written the next time that we updat ethe AWS SDK - I will manually merge your PR here and make the necessary changes :)

stack72:
Hi @A-Gordon 

Thanks again for the work here - I manually merged this work after adding 87f632cb92b6e2132b9f196d6c7f0b5dadd37fe0

This commit removes the additions to the vendor directory. I also updated the memcached test to make sure this new computed value was being set :)

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSElasticacheCluster_basic'
2016/09/20 17:20:00 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSElasticacheCluster_basic -timeout 120m
=== RUN   TestAccAWSElasticacheCluster_basic
--- PASS: TestAccAWSElasticacheCluster_basic (409.06s)
PASS
ok     	github.com/hashicorp/terraform/builtin/providers/aws   	409.078s
```

PR #8940, opened 2016-09-20
Fix typo in chef provisioner

cblecker:
Typo in example


stack72:
Hi @cblecker 

Thanks for the PR here :) This is a great find

Paul

PR #8946, opened 2016-09-20
Update index.html.markdown

baboune:
Add note to reinforce the need for cert_path on any docker-machine based Docker environment.


jen20:
Hi @baboune! Thanks for this. I've merged it and added another note explaining that you can also use `eval $(docker-machine env default)` or similar to avoid having to set these in a Docker Machine environment.

jen20:
These landed as  1b6c1cc (your original) and c942707 (my update) respectively.

Issue #8735, opened 2016-09-08
Terrafrom not working behind a proxy using http(s)_proxy envvariable


jfpucheu:
Hello, 

Currently in terraform 0.7.3  in not working behind a proxy using openstack provider.

Steps to reproduce:

export http_proxy=127.0.0.1:3128
export https_proxy=127.0.0.1:3128

terraform plan

-> impossible to connect.

cblecker:
@jfpucheu Have you tried a different provider? Terraform appears to work correctly when using the Google Cloud provider, passing traffic through a proxy. I personally don't have an openstack environment to test using that provider. This may be a bug specific to that provider.

jfpucheu:
@cblecker Hello I don't have other provider to test but i change the line 86 in 

https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/config.go

(old)      transport := &http.Transport{ TLSClientConfig: config}
 
(new)         transport := &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: config}
 
And now i can use proxy but i dont' know if it is the best place to set the proxy .

Jeff

cricketlong:
this help me a lot! This fix muss be merged in the new version.

jtopjian:
Closed by #8948

eskarthi:
for Windows environment , set the proxy property like below in dos prompt or permanently in windows environment variable.

set HTTP_PROXY=http://1.2.3.4:8080
set HTTP_USER=abcd1234
set HTTP_PASSWORD=abcsd12312

then run terraform script, it should work.

PR #8948, opened 2016-09-20
Correct Proxy setting for openstack provider: issue 8735

jfpucheu:
Correction of the Issue #8735 to use openstack provider trought a proxy.


jtopjian:
LGTM 👍 

Sorry for the delay.

PR #8960, opened 2016-09-21
Correct the data source configuration example to be valid

MiLk:
According to https://www.terraform.io/docs/providers/aws/d/ami.html, the example given on https://www.terraform.io/docs/configuration/data-sources.html is incorrect.

I've edited it, please confirm that it's the correct way to do what is expected.


stack72:
Hi @MiLk 

You are indeed correct updating this example! Thanks for the work here :)

Paul

PR #8970, opened 2016-09-21
provider/aws: Fix importing of EIP by IP address

### LINKED ISSUE #8962: Importing EIP by IP address appears to work, but doesn't really
I'd like to import an existing NAT gateway with its EIP intact. No matter what I do, the EIP is not imported and the nat gateway is recreated. This is important as the EIP is whitelisted at other sites.

### Terraform Version
Terraform v0.7.4-dev (968472a63e4542a2c5e2693c4489da4776dac996)

### Affected Resource(s)
aws_nat_gateway

```hcl
resource "aws_eip" "nat-eip-test" {
  vpc   = true
}

resource "aws_nat_gateway" "nat-gw" {
  subnet_id     = "subnet-12345678"
  allocation_id = "${aws_eip.nat-eip-test.id}"
}
```

### Expected Behavior
The NAT Gatway and EIP are imported. A terraform plan or apply does nothing as state matches the reality of AWS.

### Actual Behavior
Terraform destroys the imported NAT and creates a new NAT with a new EIP with the wrong address

### Steps to Reproduce
1) Terraform apply the above file
2) Remove the tfstate file
3) Import the EIP: terraform import aws_eip.nat-eip-test 52.1.1.1
4) Import the NAT Gatewat :  terraform import aws_nat_gateway.nat-gw nat-aaaaaaaaaa
4) terraform plan gives the below output

```hcl
+ aws_eip.nat-eip-test
    allocation_id:     "<computed>"
    association_id:    "<computed>"
    domain:            "<computed>"
    instance:          "<computed>"
    network_interface: "<computed>"
    private_ip:        "<computed>"
    public_ip:         "<computed>"
    vpc:               "true"

-/+ aws_nat_gateway.nat-gw
    allocation_id:        "eipalloc-abcdefg" => "${aws_eip.nat-eip-test.id}" (forces new resource)
    network_interface_id: "eni-abcabcab" => "<computed>"
    private_ip:           "10.0.0.1" => "<computed>"
    public_ip:            "52.1.1.1" => "<computed>"
    subnet_id:            "subnet-aaaaaa" => "subnet-aaaaaa"
```
5) terraform apply - This destroys the imported NAT, creates a new one and this new NAT has the wrong EIP

catsby:
Allow users to import EIP resources by IP even if they are in a VPC. Users should use the allocation id to import when using a VPC, but this is possible either way.

~~EIPs are meant to be imported by their allocation id, however, importing
by their EIP *appears* to work because this API actually accepts IP
lookup, despite the [documentation asking for the allocation id](http://docs.aws.amazon.com/fr_fr/AWSEC2/latest/APIReference/API_DescribeAddresses.html).~~

Here we:

- update docs on how to import EIPs and provide an example
- fix case if user imported by IP, to switch to using the alloc id for
the resource id **when in a VPC**

~~I chose not to document that looking up by IP is a method of import,
because the AWS  API docs do not explicitly say that looking up by IP is
OK, so I'd rather not encourage users to do so, so long as the AWS API doesn't explicitly say it's OK.~~

~~Alternatively, we could choose to not support IP imports and parse the resource ID and reject it (remove from
state with error/warning) if it doesn't match the `eipalloc-*` format. I thought this was a bit better UX, as some users may just think the IP is OK to import on, and maybe it should be.~~

**Update:** Importing by IP is the only way EC2 Classic users can import, ignore the strike through content above

Fixes #8962 

```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEIP -timeout 120m
=== RUN   TestAccAWSEIPAssociation_basic
--- PASS: TestAccAWSEIPAssociation_basic (140.31s)
=== RUN   TestAccAWSEIP_basic
--- PASS: TestAccAWSEIP_basic (8.17s)
--- PASS: TestAccAWSEIP_instance (169.95s)
--- PASS: TestAccAWSEIP_network_interface (32.44s)
--- PASS: TestAccAWSEIP_twoEIPsOneNetworkInterface (31.58s)
--- PASS: TestAccAWSEIP_associated_user_private_ip (124.87s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    507.351s
Test:
```

### DIFF
--- a/builtin/providers/aws/resource_aws_eip.go
+++ b/builtin/providers/aws/resource_aws_eip.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"net"
 	"strings"
 	"time"
 
@@ -172,6 +173,13 @@ func resourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.Set("domain", address.Domain)
+
+	// confirm we have an Allocation ID for our ID
+	// Allows users to import with IP address and not just allocation id
+	if net.ParseIP(id) != nil {
+		log.Printf("[DEBUG] Re-assigning EIP ID (%s) to it's Allocation ID (%s)", d.Id(), *address.AllocationId)
+		d.SetId(*address.AllocationId)
+	}
 
 	return nil
 }
--- a/builtin/providers/aws/resource_aws_eip.go
+++ b/builtin/providers/aws/resource_aws_eip.go
@@ -174,8 +174,8 @@ func resourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
 
 	d.Set("domain", address.Domain)
 
-	// confirm we have an Allocation ID for our ID
-	// Allows users to import with IP address and not just allocation id
+	// Force ID to be an Allocation ID if we're on a VPC
+	// This allows users to import the EIP based on the IP if they are in a VPC
 	if *address.Domain == "vpc" && net.ParseIP(id) != nil {
 		log.Printf("[DEBUG] Re-assigning EIP ID (%s) to it's Allocation ID (%s)", d.Id(), *address.AllocationId)
 		d.SetId(*address.AllocationId)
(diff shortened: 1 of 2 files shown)


stack72:
This LGTM! One thought )not related to this PR) is that we currently have no tests for the import of EIPs

I think this should happen as a follow up - I can take care of this in the morning UK time

PR #8971, opened 2016-09-21
provider/azurerm: update Azure SDK

pmcatominey:
Based off master v4


jen20:
Hi @pmcatominey! Looks like there are CI failures here about missing packages. Regrettably I suspect this upgrade will not be as easy as one would hope as it looks like we have resource code which needs adjustment.

pmcatominey:
Hi @jen20, I've just fixed the error, was due to the import still hanging around in my GOPATH but not in the vendor directory. The only code changes I've made so far are to remove the import on the `github.com/Azure/azure-sdk-for-go/core/http` and replace with `net/http`, they previously had this and the `crypto/tls` package vendored. They have since been removed as TLS Renegotiation support was added in Go 1.7.

I was originally updating the SDK in the hopes that it would fix an issue that I've been encountering with creating ServiceBus Topics, however it still appears to be an issue at this version. I opened the PR anyway to stop us from falling too far behind.

I think a full acceptance test run is needed before this can be considered for merging.

Cheers

pmcatominey:
It's also worth noting that version 5 is on the way Azure/azure-sdk-for-go#394

stack72:
@pmcatominey ok, going to merge this to let it soak in CI for a couple of days :) Thanks for the work here

P.

pmcatominey:
@maryamklabib what version of Go are you using, a similar error was [reported](https://github.com/Azure/azure-sdk-for-go/issues/393) and was solved by using Go 1.7.

maryamklabib:
@pmcatominey Thanks! So Terraform now requires Go 1.7 going forward. We are using Go 1.6. We'll try upgrading and see if that fixes it.

stack72:
Hi @maryamklabib 

Apologies for this - this is an upstream change that affected us i'm afraid

P.

maryamklabib:
@stack72 @pmcatominey No problem. Upgrading to Go 1.7 did fix it for us. Might be helpful to update the README to note that Go 1.7 is now required.

PR #8974, opened 2016-09-21
Add a way to avoid panicwrap when debugging

### LINKED ISSUE #8795: core: add an environment variable to prevent forking
terraform currently re-execs itself in the background to capture panics and logs, but this can be inconvenient for debugging core issues. 

Add a new `TF_FORK=0` environment variable to disable this behavior.

jbardin:
When trying to use a debugger, you don't want the terraform to run in a
new child process. Setting TF_FORK=0 will skip panicwrap and continue
running the program in the current process. This will also give direct
access to log output, and stdout.

Fixes #8795

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,11 @@ func main() {
 func realMain() int {
 	var wrapConfig panicwrap.WrapConfig
 
+	// don't re-exec terraform as a child process for easier debugging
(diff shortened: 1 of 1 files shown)


jen20:
This is great, I was meaning to do it for a long time! LGTM pending Travis!

jbardin:
It's the little things ;)

PR #8975, opened 2016-09-21
Allow use of protocol numbers for ah and esp

mootpt:
It appears the aws sdk does not accept the protocol name for esp or ah. This fix allows the user to simply specify the [protocol number](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) in their security group. 

This fix covers: https://github.com/hashicorp/terraform/issues/6888

### DIFF
--- a/builtin/providers/aws/network_acl_entry.go
+++ b/builtin/providers/aws/network_acl_entry.go
@@ -82,8 +82,6 @@ func protocolIntegers() map[string]int {
 	var protocolIntegers = make(map[string]int)
 	protocolIntegers = map[string]int{
 		// defined at https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
-		"ah":   51,
-		"esp":  50,
 		"udp":  17,
 		"tcp":  6,
 		"icmp": 1,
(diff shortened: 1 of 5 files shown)


jrnt30:
The same thing will apply I believe to my comment here: https://github.com/hashicorp/terraform/pull/6915#issuecomment-223364486

The `network_acl_entry`'s call to AWS-SDK does support these two port mappings (the sdk clearly is not consistent).  As a heads up, I believe this will then be a breaking change for anyone using them.

catsby:
Thanks for the ping @jrnt30 , I'll take a look at that comment/PR thread.
Re: breaking change, it's my understanding that if a user is using `esp` that it isn't working at all (e.g. the rule fails to create at all and errors), as opposed to silently not working or something. So, I don't feel it's breaking anything if it was broken and never worked before 😄  Am I wrong there? Is there some way / scenario that users are using `ah` or `esp` successfully, that I'm missing?

jrnt30:
That is correct in the `security_group_rule` resource which this fixes.  I need to take a look at the code again, however before that list that was modified in the PR was also used by a few other resources, which under the covers the AWS SDK *did* support the logical names for.  

I can take a look tonight if it's not cleared up, just really wanted to give a public service announcement :)

catsby:
@jrnt30 hey, sorry, I think I misread your first comment, so you are likely correct that any `aws_network_acl` user will have a bad time now. Let me do some thinking and digging to see what I can come up with.

catsby:
@jrnt30 this should be sorted in https://github.com/hashicorp/terraform/pull/8999

PR #8983, opened 2016-09-22
[WIP] provider/aws: fix option updates to beanstalk

987poiuytrewq:
Before upgrading to v0.7.3 I was seeing the issue where, when you update elasticbeanstalk settings, the result is that the setting gets removed rather than updated. I was seeing this only with environment variables, but the issue may apply to all types of settings.

After upgrading to v0.7.4, this fix that was released https://github.com/hashicorp/terraform/pull/8848 that addresses the issue. However, because it is modifying the slice it is iterating over, you get a `slice bounds out of range` panic.

I've modified the logic to track which "removals" are actually updates, and not adding them to the set of options to remove. This seems to work in my manual testing. 

I tried to modify the test for this but I couldn't figure out how to run it. Never programmed in go before so not really sure how to run it, any help welcome! Running `make test` appears to pass but I'm not confident it is actually running.

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/builtin/providers/aws/resource_aws_elastic_beanstalk_environment.go
@@ -334,13 +334,15 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
 		// Additions and removals of options are done in a single API call, so we
 		// can't do our normal "remove these" and then later "add these", re-adding
 		// any updated settings.
-		// Because of this, we need to remove any settings in the "removable"
+		// Because of this, we need to exclude any settings in the "removable"
 		// settings that are also found in the "add" settings, otherwise they
 		// conflict. Here we loop through all the initial removables from the set
-		// difference, and delete from the slice any items found in both `add` and
-		// `rm` above
+		// difference, and create a new slice `remove` that contains those settings
+		// found in `rm` but not in `add`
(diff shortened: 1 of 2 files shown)


Dchamard:
I have built a special build with this commit and so far so good. It does not crash when I tried to update multiple options at the same time.

Dchamard:
Any chances that this will be merged soon?

myoung34:
I'm experiencing this as well and it's giving many headaches

catsby:
Thanks @987poiuytrewq ! Is there anything stopping me merging this PR? It's labeled "WIP" so I wanted to make sure. I ran the tests myself and they passed

```
[pr-8983](2)$ git log -1
commit 826ff3f0f02a0814aa8304557a03be0d8953f986
Author: Duncan <duncancraigwilliams@gmail.com>
Date:   Thu Sep 22 00:05:59 2016 +0100

    provider/aws: fix option updates to beanstalk

$ TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSBeanstalkEnv_basic_settings_update -timeout 120m
=== RUN   TestAccAWSBeanstalkEnv_basic_settings_update
--- PASS: TestAccAWSBeanstalkEnv_basic_settings_update (591.94s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	591.965s
Test:
[pr-8983](2)$
```

987poiuytrewq:
Oh thanks - yeah I wasn't confident the tests were running which is why it was WIP

PR #8985, opened 2016-09-22
Fmt all the config files

w3st3ry:
Lint and reformatting using fmt command all the terraform config files.

Signed-off-by: Valentin Pichard <valentin.pichard@corp.ovh.com>


stack72:
Hi @w3st3ry 

thanks so much for the work here :) This LGTM!

Paul

PR #8992, opened 2016-09-22
provider/aws: Handle JSON parsing error in the ReadFunc for various JSON documents.

kwilczynski:


kwilczynski:
Work initially completed in #8896, but as per @stack72, an error handling needs to be added.

PR #3858, opened 2015-11-11
implements gzip interpolation func

glerchundi:
fixes #3407 

Signed-off-by: Gorka Lerchundi Osa glertxundi@gmail.com

PR #8994, opened 2016-09-22
provider/template: Added single part support to datasource cloudinit template.

cgroschupp:
Some implementations from cloud-init does not support multi parts (e.g. coreos-cloudinit).

### DIFF
--- a/builtin/providers/template/datasource_cloudinit_config.go
+++ b/builtin/providers/template/datasource_cloudinit_config.go
@@ -59,6 +59,11 @@ func dataSourceCloudinitConfig() *schema.Resource {
 				Computed:    true,
 				Description: "rendered cloudinit configuration",
 			},
+			"single_part": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


apparentlymart:
Hi @cgroschupp! Thanks for the work here.

I'd like to understand a little more about the use-case here, since as far as I was aware the multipart MIME encoding was the primary reason for this data source to exist... if you're just generating a single part, could you not just use directly whatever value you would've passed to `content` in the single part you'd provide?

cgroschupp:
Hi @apparentlymart 
coreos-cloudinit supports only one plain part, i don't have find a way to create the template without mime.

See also the "cloud_datasource_init_config" tests.

kwilczynski:
@cgroschupp hi there! I believe "single part" is just a flat file, and using `template_file` should suffice.

cgroschupp:
@kwilczynski but template_file can't creating gzip content.
See also comment from @paultyng in pr #3858 and issue #3407

apparentlymart:
Ahh, I see. So really this is just a gzip function implemented as a data source...

I raised some concerns in that other issue about whether Terraform could losslessly persist gzip binary data in JSON. Given that we already do this for this resource, can I assume that it is actually working properly?

One interesting thing about data resources, which didn't exist at the time I wrote those comments, is that in most cases we don't use the values cached in state because they get rebuild during the refresh phase. So if it were to break in the way I was expecting, it would be only if -refresh=false were specified to force Terraform to use the cached values in state.

cgroschupp:
For me gzip+base64 works fine, i have no experiences with only gzip. But i have added tests for both cases.

teamterraform:
Hi folks! We are working on general github issue/PR cleanup. I am going to close this PR. If this is still something that needs to be done, the terraform template provider is in [its own repository](https://github.com/terraform-providers/terraform-provider-template).

PR #8999, opened 2016-09-22
provider/aws: Restore ah/esp protocol support for Network ACLs

catsby:
https://github.com/hashicorp/terraform/pull/8975 removed support for `ah` and `esp` string support for ingress/egress protocols in Security Groups. It seems it was too far reaching though, as while the SecurityGroup SDK/API does not support using strings for these protocols, the Network ACL API does.

This PR restores support for using the string version of those protocols with Network ACLs, while still removing support for them in Security Groups. 

Thanks to @jrnt30 for pointing this out, and @sarkis as well in #6915

A custom TravisCI run for all `TestAccAWSNet*` and `TestAccAWSSe*` tests:

- https://travis-ci.org/hashicorp/terraform/builds/163436738


stack72:
LGTM! sorry it took so long to get back to this :)

catsby:
The failing travis test(s) are flaky and unrelated; pulling this in

PR #9005, opened 2016-09-22
Fixed typo - no such thing as `aws_target_group`

tomgoren:
It seems like there was a typo and a reference was made to
`aws_target_group`, when the actual resource name is
`aws_alb_target_group`.


catsby:
thanks!

PR #9009, opened 2016-09-23
provider/aws: Add some tests for the Import for aws_eip

stack72:
```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEIP_'
2016/09/23 09:28:32 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEIP_ -timeout
120m
=== RUN   TestAccAWSEIP_importEc2Classic
--- PASS: TestAccAWSEIP_importEc2Classic (116.16s)
=== RUN   TestAccAWSEIP_importVpc
--- PASS: TestAccAWSEIP_importVpc (61.89s)
=== RUN   TestAccAWSEIP_basic
--- PASS: TestAccAWSEIP_basic (18.86s)
--- PASS: TestAccAWSEIP_instance (185.95s)
--- PASS: TestAccAWSEIP_network_interface (63.20s)
--- PASS: TestAccAWSEIP_twoEIPsOneNetworkInterface (65.64s)
--- PASS: TestAccAWSEIP_associated_user_private_ip (201.34s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    713.072s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_eip_test.go
+++ b/builtin/providers/aws/resource_aws_eip_test.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 
@@ -11,6 +12,50 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 )
+
+func TestAccAWSEIP_importEc2Classic(t *testing.T) {
+	oldvar := os.Getenv("AWS_DEFAULT_REGION")
+	os.Setenv("AWS_DEFAULT_REGION", "us-east-1")
+	defer os.Setenv("AWS_DEFAULT_REGION", oldvar)
+
+	resourceName := "aws_eip.bar"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEIPDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEIPInstanceEc2Classic,
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccAWSEIP_importVpc(t *testing.T) {
+	resourceName := "aws_eip.bar"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEIPDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEIPNetworkInterfaceConfig,
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
 
 func TestAccAWSEIP_basic(t *testing.T) {
 	var conf ec2.Address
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_eip_test.go
```diff
@@ -12,6 +12,52 @@ import (
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEIPInstanceConfig,
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"vpc",
```
catsby: I'm curious why we're ignoring the import of `vpc`, I believe it's setting has an impact to how the resource behaves.

stack72:
@catsby for the purposes of import, the ignoring of `vpc` isn't important. When we import, we determine if we are in a VPC using the `domain` parameter. Setting of the `vpc` to the bool just happens at the end of the func

this is the actual code that does the importing:

```
req := &ec2.DescribeAddressesInput{}

	if domain == "vpc" {
		req.AllocationIds = []*string{aws.String(id)}
	} else {
		req.PublicIps = []*string{aws.String(id)}
	}
```

catsby:
> vpc is just something we use internally - it is not exposed from the API

which is why I was questioning why we omit it in the import; it can possibly affect how the resource behaves (by way of the `domain` attribute), so I feared that not caring how it gets imported could have unintended consequences.

I tried this out, and the `domain` attribute seems to be getting set correctly, which I believe is ultimately the thing that controls the behavior of `aws_eip`, so that part seems OK. 

That said, I applied the below config and then removed and imported the `aws_eip`, and got a diff on the next plan:

```
#config 
  # us-west-2
  ami = "ami-4fccb37f"
  # us-east-1
  #ami = "ami-8c6ea9e4"

  instance_type = "m1.small"
}

resource "aws_eip" "bar" {
  instance = "${aws_instance.foo.id}"
}
```

import works fine, but a follow up plan shows a diff:

```
~ aws_eip.bar
    vpc: "true" => "false"
```

`vpc` is getting set to `true` on import (correctly, I think), however `vpc` being an `Optional` field in the schema, we're now picking up a diff because it's not specified in the config. This does not seem to be a bug in the import itself, but I think that by ignoring the `vpc` here in the import we've unintentionally covered up a bug with import (or the schema, I suppose). 

Thoughts?

(fwiw the above config works as expected in classic  👌)

stack72:
```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEIP_'
2016/10/07 12:32:10 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEIP_ -timeout 120m
=== RUN   TestAccAWSEIP_importEc2Classic
--- PASS: TestAccAWSEIP_importEc2Classic (113.29s)
=== RUN   TestAccAWSEIP_importVpc
--- PASS: TestAccAWSEIP_importVpc (71.80s)
=== RUN   TestAccAWSEIP_basic
--- PASS: TestAccAWSEIP_basic (21.25s)
--- PASS: TestAccAWSEIP_instance (198.77s)
--- PASS: TestAccAWSEIP_network_interface (71.90s)
--- PASS: TestAccAWSEIP_twoEIPsOneNetworkInterface (75.54s)
2016/10/07 12:44:30 Request body type has been overwritten. May cause race conditions
2016/10/07 12:44:34 Request body type has been overwritten. May cause race conditions
2016/10/07 12:44:36 Request body type has been overwritten. May cause race conditions
2016/10/07 12:44:39 Request body type has been overwritten. May cause race conditions
2016/10/07 12:44:43 Request body type has been overwritten. May cause race conditions
--- PASS: TestAccAWSEIP_associated_user_private_ip (209.33s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	761.909s
```

Issue #8146, opened 2016-08-11
core: "unknown variable accessed" when doing destroy with target


kyhavlov:
I'm getting the following error when trying to run `terraform destroy -target=module.thing1` on the config below:

```
Error configuring: 1 error(s) occurred:

* 1:3: unknown variable accessed: var.count in:

${var.count}
```

This doesn't happen without the -target, so it seems like that's the issue. It could be related to #8098 but I've only seen this error with -target when doing destroy, not apply or plan.
### Terraform Version

```
$ terraform -v
Terraform v0.7.0
```
### Terraform Configuration Files

./main.tf:

``` hcl
module "thing1" {
  source = "./thing"
  count = 3
}
```

./thing/null.tf:

``` hcl
variable "count" {}

resource "null_resource" "thing" {
  count        = "${var.count}"
}
```

vaygr:
Same issue here.

vaygr:
Same issue here, but it got fixed when I set `default` value for the variable.

roboll:
Setting a default for the variable doesn't fix the underlying issue, that the interpolated variable is not being respected properly in the destroy dependency graph. When the destroy executes, it uses the default, which in many cases will cause incorrect behavior.

roboll:
This broke sometime between `v0.7.0-rc3` and `v0.7.0`.

roboll:
I suspect #6599 may have something to do with this.

roboll:
Confirmed removing [this change](https://github.com/hashicorp/terraform/blob/559f017ebb7649d4db425f9f2412c1463a36f97d/terraform/graph_config_node_variable.go#L93-L101) from #6599 fixes the issue from `v0.7.0`.

RyanBowlby-Reflektion:
Thanks @vaygr for the temporary hack. After filling in several defaults (that would never be real defaults) I was able to destroy.

Scary to think that the non-default values are not being honored.

hazim1093:
Facing the same issue on 0.7.2 and 0.7.3. 
The workaround of adding default value did the trick

kyhavlov:
I put up a PR for this: #9021

mitchellh:
Since #9021 merged the repro above no longer occurs. Thanks @kyhavlov!

vaygr:
Looks like the same issue exists for `apply` in at least **0.8.1**:

```shell
$ terraform plan -target='module.ca_central_1'
```

```
Error running plan: 8 error(s) occurred:

[...]

PR #9021, opened 2016-09-23
core: Fixed variables not being in scope for destroy -target on modules

kyhavlov:
Addresses the bug in #8146. The call to `selectTargetedNodes()` was only returning the resource and not the variable that it depended on to interpolate its count, so the variable got removed from the graph and wasn't in scope for later interpolations, causing the `unknown variable accessed` error.

@phinze

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -3119,6 +3119,68 @@ module.child:
 	}
 }
 
+func TestContext2Apply_destroyTargetWithModuleVariableAndCount(t *testing.T) {
+	m := testModule(t, "apply-destroy-mod-var-and-count")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+
+	var state *State
+	var err error
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Module: m,
+			Providers: map[string]ResourceProviderFactory{
+				"aws": testProviderFuncFixed(p),
+			},
+		})
+
+		// First plan and apply a create operation
+		if _, err := ctx.Plan(); err != nil {
+			t.Fatalf("plan err: %s", err)
+		}
+
+		state, err = ctx.Apply()
+		if err != nil {
+			t.Fatalf("apply err: %s", err)
+		}
+	}
+
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Destroy: true,
+			Module:  m,
+			State:   state,
+			Providers: map[string]ResourceProviderFactory{
+				"aws": testProviderFuncFixed(p),
+			},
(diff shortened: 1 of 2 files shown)


terraform/context_apply_test.go
```diff
@@ -3119,6 +3119,91 @@ module.child:
...
+		}
+
+		state, err = ctx.Apply()
+		if err != nil {
+			t.Fatalf("apply err: %s", err)
+		}
+	}
+
+	h := new(HookRecordApplyOrder)
+	h.Active = true
```
phinze: Looks like this Hook is not used in this test.

terraform/context_apply_test.go
```diff
@@ -3119,6 +3119,91 @@ module.child:
...
+			Module:  m,
+			State:   state,
+			Hooks:   []Hook{h},
+			Providers: map[string]ResourceProviderFactory{
+				"aws": testProviderFuncFixed(p),
+			},
+			Targets: []string{"module.child"},
+		})
+
+		// First plan and apply a create operation
```
phinze: This comment isn't applicable here anymore, since this is the destroy.

terraform/context_apply_test.go
```diff
@@ -3119,6 +3119,91 @@ module.child:
...
+		planFromFile, err := ReadPlan(&buf)
+		if err != nil {
+			t.Fatalf("plan read err: %s", err)
+		}
+
+		ctx, err = planFromFile.Context(&ContextOpts{
+			Providers: map[string]ResourceProviderFactory{
+				"aws": testProviderFuncFixed(p),
+			},
+		})
```
phinze: The above `WritePlan` / `ReadPlan` dance is meant to emulate writing out to a `tfplan` file. It's unnecessary here, since the bug reproduces w/o it.

terraform/transform_targets.go
```diff
@@ -86,6 +86,14 @@ func (t *TargetsTransformer) selectTargetedNodes(
 			var err error
 			if t.Destroy {
 				deps, err = g.Descendents(v)
+
+				// Select any variables that we depend on in case we need them later for interpolating in count
```
phinze: Style nit: wrap comment at 80 columns.

kyhavlov:
@phinze addressed the comments and cleaned up the test. I was going add a test for the nested module case, but I hit the issue with Terraform not including child modules of a module when using -target, so that will have to be another PR.

phinze:
LGTM

matthughes:
Got the error this was supposed to fix on 0.7.4.  Upgraded to 0.7.6.  Now I get:

```

* Error reading data.template_file.master_config count: strconv.ParseInt: parsing "${var.master_count}": invalid syntax
```

PR #9023, opened 2016-09-23
AWS: Skip DetachVolume if volume is not attached

Jonnymcc:
I manually terminated an instance thus putting the EBS volume in an "available" state. When running `terraform apply` I got this error...
```
* aws_volume_attachment.jira_data_volume: Failed to detach Volume (vol-8237ff0d) from Instance (i-8d44ef87): IncorrectState: Volume 'vol-1537ff1d'is in the 'available' state.
    status code: 400, request id: e0fc1368-c2c5-4a9d-83c9-e958c8ee3ca2
```

This change appears to have fixed the problem. I ran `terraform apply` and the aws_volume_attachment resource was deleted as expected.

### DIFF
--- a/builtin/providers/aws/resource_aws_volume_attachment.go
+++ b/builtin/providers/aws/resource_aws_volume_attachment.go
@@ -136,7 +136,7 @@ func resourceAwsVolumeAttachmentRead(d *schema.ResourceData, meta interface{}) e
 		},
 	}
 
-	_, err := conn.DescribeVolumes(request)
+	vols, err := conn.DescribeVolumes(request)
 	if err != nil {
 		if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == "InvalidVolume.NotFound" {
 			d.SetId("")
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_volume_attachment.go
```diff
@@ -153,6 +153,20 @@ func resourceAwsVolumeAttachmentDelete(d *schema.ResourceData, meta interface{})
...
+	desc_opts := &ec2.DescribeVolumesInput{
+		VolumeIds: []*string{aws.String(vID)},
+	}
+
+	attr, desc_err := conn.DescribeVolumes(desc_opts)
+	if desc_err != nil {
+		return fmt.Errorf("Error reading EC2 volume %s: %s", vID, desc_err)
+	}
+
+	if *attr.Volumes[0].State == "available" {
```
catsby: can we put a guard here to protect us from the off chance that `attr.Volumes` is nil?

Jonnymcc: Done here https://github.com/hashicorp/terraform/pull/9023/commits/dc4a7b7d25f56f3ae3e30c889521827de68dfcbd#diff-4248c02c9fbfe452aff773530e14032dR148

catsby:
Thanks!

PR #7049, opened 2016-06-07
provider/softlayer: New Resources - DNS Domain/Resource Records, VPX Load Balancer

danielcbright:
### This PR will add 5 new resources to the SoftLayer provider:

**DNS:**
- `softlayer_dns_domain`
- `softlayer_dns_domain_resourcerecord`

**Load Balancing (Citrix VPX):**
- `softlayer_lb_vpx`
- `softlayer_lb_vpx_vip`
- `softlayer_lb_vpx_service`
#### DNS:

```
resource "softlayer_dns_domain" "main" {
    name = "main.example.com"
}

resource "softlayer_dns_domain_record" "www" {
    record_data = "123.123.123.123"
    domain_id = "${softlayer_dns_domain.main.id}"
    host = "www.example.com"
    contact_email = "user@softlayer.com"
    ttl = 900
    record_type = "a"
}
```
##### Load Balancing:

```
resource "softlayer_lb_vpx" "testacc_foobar_vpx" {
    datacenter = "DALLAS06"
    speed = 10
    version = "10.1"
    plan = "Standard"
    ip_count = 2
}

resource "softlayer_lb_vpx_vip" "testacc_vip" {
    name = "test_load_balancer_vip"
    nad_controller_id = "${softlayer_lb_vpx.testacc_foobar_vpx.id}"
    load_balancing_method = "lc"
    source_port = 81
    type = "HTTP"
    virtual_ip_address = "${softlayer_virtual_guest.terraform-acceptance-test-1.ipv4_address}"
}

resource "softlayer_lb_vpx_service" "testacc_service" {
  name = "test_load_balancer_service"
  vip_id = "${softlayer_lb_vpx_vip.testacc_vip.id}"
  destination_ip_address = "${softlayer_virtual_guest.terraform-acceptance-test-2.ipv4_address}"
  destination_port = 89
  weight = 55
  connection_limit = 5000
  health_check = "HTTP"
}
```
- This PR also fixes some type-o's for the `softlayer_ssh_key` and `softlayer_virtual_guest` resources that prevented the examples from working properly and bring some more consistency to the documentation.

PR #9024, opened 2016-09-23
Point users to latest version of terraform-softlayer provider plugin

renier:


renier:
@stack72 ^

radeksimko:
Hi @renier 
thank you for the effort in offering a solution to users that are waiting for https://github.com/hashicorp/terraform/pull/7049 to be merged and generally for the internal provider to catch-up, but I'm not sure if this is the right solution.

While I agree the Terraform website/docs should have a place where we promote 3rd-party providers/provisioners and generally point out projects that are worth pointing out (like this provider) I think it should be a separate, dedicated place - or at least it should be clear what is internal and what is external.

Sorry for the delay in reviewing https://github.com/hashicorp/terraform/pull/7049
I'm open for suggestions in regards to place for promoting community/3rd-party efforts - perhaps in a separate issue/PR?

renier:
@radeksimko I think the website docs already help promote 3rd-party providers/provisioners just by the fact that they exist. I believe those are done in large part by 3rd parties are they not? Still, it is your prerogative to decide the kinds of things you allow to be there. That is ok.

You mention that _at least it should be clear what is internal and what is external_. Can we do that with a PR like this? Is it problem with the wording then? Or is it that by putting anything remotely like a link to another repo in the docs it would irrevocably muddy the waters?

But the root problem is getting timely updates to the SoftLayer provider to begin with. Is there light at the end of the tunnel [here](https://github.com/hashicorp/terraform/pull/7049)?

Otherwise, I'm afraid the internal SoftLayer provider will remain so far behind. Without being able to give users any kind of pointers about this problem (e.g. Google shows the terraform website as the first hit when you search for _terraform softlayer_), then what is the value to continue having the provider within the terraform tree?

Would it be better to discuss this on the mailing list?

radeksimko:
> I think the website docs already help promote 3rd-party providers/provisioners just by the fact that there are docs for them.

The docs at https://www.terraform.io/docs/providers/softlayer/index.html and descendant pages are covering provider code that is bundled as _internal_ provider in each release & available from https://www.terraform.io/downloads.html - correct me if I'm wrong. If the docs match with functionality of https://github.com/softlayer/terraform-provider-softlayer then it's rather a coincidence than intention from either side (docs authors / code authors). The official docs can be indeed also managed by community - via PRs - but they still cover functionality of any code in https://github.com/hashicorp/terraform/tree/master/builtin/providers/softlayer

> I'm afraid the internal SoftLayer provider will remain so far behind

I feel your frustration while waiting for the progress, I remember there was a case when a provider was removed before more stable version was developed, but I don't think softlayer is such candidate. Such action would definitely require input from other maintainers and perhaps users that may be affected.

If we're talking _merely_ about missing features, rather than strongly buggy behaviour, then I don't think there's reason for such action.

> then what is the value to continue having the provider within the terraform tree?

Existing resources that work and allow people to automate their softlayer infrastructure at least partially.

> But the root problem is getting timely updates to the SoftLayer provider to begin with.

Please bear with us, having a proper acceptance testing setup for softlayer will help **ensuring the stability** of the provider code (which can be perceived as terraform stability too) - which is (IMO) important for everyone.

I cannot provide any timelines in regards to https://github.com/hashicorp/terraform/pull/7049, I can only advise you to:

1. Keep using https://github.com/softlayer/terraform-provider-softlayer in the meantime and take part in PRs (reviews and own submissions) and other efforts in getting this merged as internal provider eventually. [I assume](https://github.com/softlayer/terraform-provider-softlayer/graphs/contributors) you would be the best person that can help us with this.
2. Open a (separate) discussion about how can we promote 3rd party provider and generally community effort on the website - mailing list is fine, a separate Github issue may is good too.

Thank you and @danielcbright for the overall effort you invested so far into getting this merged.

renier:
>>I think the website docs already help promote 3rd-party providers/provisioners just by the fact that there are docs for them.

>The docs at https://www.terraform.io/docs/providers/softlayer/index.html and descendant pages are covering provider code that is bundled as internal provider in each release & available from https://www.terraform.io/downloads.html - correct me if I'm wrong.

What I mean is that providers like DigitalOcean, MySQL, Rundeck, SoftLayer, and others are contributed by the community (3rd-parties, invdividuals; not Hashi) into the tree. Yes, they are bundled as internal providers, just as the docs that are contributed for those are bundled in the published website. In allowing that integration and packaging, cloud platform users promote Terraform as a tool to use on their platform, but also Terraform promotes the cloud platforms to terraform users by showing it has support for those as well on the website.

>>I'm afraid the internal SoftLayer provider will remain so far behind

>I feel your frustration while waiting for the progress. I remember there was a case when a provider was removed before more stable version was developed, but I don't think softlayer is such candidate. Such action would definitely require input from other maintainers and perhaps users that may be affected.

So, we are the creators and maintainers of the SoftLayer provider that currently exists in the Terraform tree. And we have kept going as a [separate project](https://github.com/softlayer/terraform-provider-softlayer) due to this lag. We are also the best qualified group to review and test PRs for this provider. Would that change the dynamics here?

>If we're talking merely about missing features, rather than strongly buggy behaviour, then I don't think there's reason for such action.

It's not just missing features. It is a gross lack of them. For example, currently, the SoftLayer provider in the Terraform tree has 2 resources. SSH Keys and Virtual Guest. Enough to do simple experiments, but that's it. The SoftLayer platform supports hundreds of resources. Any serious user would probably have moved on already. At [terraform-provider-softlayer](https://github.com/softlayer/terraform-provider-softlayer) we are already up to 24.

The difference is so great, that it doesn't make sense anymore to keep both. Either we merge the other 22 resources into the Terraform tree (or at least have some hope that can happen in the near future) or we remove it from the tree. The former would be preferable, but if not possible, then the latter would cause less confusion for potential users. With this PR, of course, I was just trying to pursue a happy medium, where we notify users finding the in-tree provider that there is a newer one they can use, while the tree gets up-to-date.

Keep in mind that this is us, the creators and maintainers of both the in-tree SoftLayer provider, and the out-of-tree SoftLayer provider, speaking.

renier:
@radeksimko ^

renier:
Opening a separate issue to see if we can draw attention for updating the IBM SoftLayer provider.

PR #9027, opened 2016-09-23
AWS S3 bucket life_cycle rule IDs must be unique.

sean-:
Noticed while browsing the docs.

CC @stack72 or @catsby for review.


stack72:
LGTM! Nice spot :)

PR #9039, opened 2016-09-25
Automatic Rollback of CodeDeploy deployments and CloudWatch Alarms for a Deployment Group

niclic:
Added in [AWS SDK for Go v1.4.11](https://aws.amazon.com/releasenotes/3404190260331438)

This PR has a dependency on _AWS SDK for Go v1.4.11_

> AWS CodeDeploy also now supports automatically rolling back a deployment if certain conditions are met, such as a deployment failure or an activated alarm.
> 
> AWS CodeDeploy now integrates with Amazon CloudWatch alarms, making it possible to stop a deployment if there is a change in the state of a specified alarm for a number of consecutive periods, as specified in the alarm threshold. 

Although related, you could argue that these changes to `aws_codedeploy_deloyment_group` could be made separately. After all, you can configure automatic rollbacks without specifying a CloudWatch alarm. However, it might make sense to include them both in the same PR to avoid any delays (they were released together for a reason).

TODO:
- [x] Implement `auto_rollback_configuration`
- [x] Update docs for `auto_rollback_configuration`
- [x] Implement `alarm_configuration`
- [x] Update docs for `alarm_configuration`
- [x] Cleanup, refactoring, better test coverage?
- [x] Resolve issue of skipped tests (`TypeList` vs `TypeSet`)

### DIFF
--- a/builtin/providers/aws/resource_aws_codedeploy_deployment_group.go
+++ b/builtin/providers/aws/resource_aws_codedeploy_deployment_group.go
@@ -495,7 +495,7 @@ func buildAutoRollbackConfig(configured []interface{}) *codedeploy.AutoRollbackC
 		config := configured[0].(map[string]interface{})
 		result.Enabled = aws.Bool(config["enabled"].(bool))
 		result.Events = expandStringSet(config["events"].(*schema.Set))
-	} else {
+	} else { // delete the configuration
 		result.Enabled = aws.Bool(false)
 		result.Events = make([]*string, 0)
 	}
@@ -522,9 +522,10 @@ func buildAlarmConfig(configured []interface{}) *codedeploy.AlarmConfiguration {
 		result.Alarms = alarms
 		result.Enabled = aws.Bool(config["enabled"].(bool))
 		result.IgnorePollAlarmFailure = aws.Bool(config["ignore_poll_alarm_failure"].(bool))
-	} else {
+	} else { // delete the configuration
 		result.Alarms = make([]*codedeploy.Alarm, 0)
 		result.Enabled = aws.Bool(false)
+		result.IgnorePollAlarmFailure = aws.Bool(false)
 	}
 	return result
 }
@@ -584,12 +585,16 @@ func triggerConfigsToMap(list []*codedeploy.TriggerConfig) []map[string]interfac
 // into a []map[string]interface{} list containing a single item
 func autoRollbackConfigToMap(config *codedeploy.AutoRollbackConfiguration) []map[string]interface{} {
 	result := make([]map[string]interface{}, 0, 1)
+
+	// only create configurations that are enabled or temporarily disabled (retaining events)
+	// otherwise empty configurations will be created
 	if config != nil && (*config.Enabled == true || len(config.Events) > 0) {
 		item := make(map[string]interface{})
 		item["enabled"] = *config.Enabled
 		item["events"] = schema.NewSet(schema.HashString, flattenStringList(config.Events))
 		result = append(result, item)
 	}
+
 	return result
 }
 
@@ -598,21 +603,22 @@ func autoRollbackConfigToMap(config *codedeploy.AutoRollbackConfiguration) []map
 func alarmConfigToMap(config *codedeploy.AlarmConfiguration) []map[string]interface{} {
 	result := make([]map[string]interface{}, 0, 1)
 
-	if config == nil {
-		return result
+	// only create configurations that are enabled or temporarily disabled (retaining alarms)
+	// otherwise empty configurations will be created
+	if config != nil && (*config.Enabled == true || len(config.Alarms) > 0) {
+		names := make([]*string, 0, len(config.Alarms))
+		for _, alarm := range config.Alarms {
+			names = append(names, alarm.Name)
+		}
+
+		item := make(map[string]interface{})
+		item["alarms"] = schema.NewSet(schema.HashString, flattenStringList(names))
+		item["enabled"] = *config.Enabled
+		item["ignore_poll_alarm_failure"] = *config.IgnorePollAlarmFailure
+
+		result = append(result, item)
 	}
 
-	names := make([]*string, 0, len(config.Alarms))
-	for _, alarm := range config.Alarms {
-		names = append(names, alarm.Name)
-	}
-
-	item := make(map[string]interface{})
-	item["alarms"] = schema.NewSet(schema.HashString, flattenStringList(names))
-	item["enabled"] = *config.Enabled
-	item["ignore_poll_alarm_failure"] = *config.IgnorePollAlarmFailure
-
-	result = append(result, item)
 	return result
 }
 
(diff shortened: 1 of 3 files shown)


niclic:
I think this is almost complete, so a review would be helpful. There is _one_ outstanding issue to resolve.

There are two tests that are deliberately marked as _SKIP_:

 * skipTestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_disable
 * skipTestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_disable

These fail but they should really pass.

In `skipTestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_disable`, the list of `alarms` is updated to `[]` when it should still be `["foo"]`.

In `skipTestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_disable`, the list of `events` is updated to `[]` when it should still be `["DEPLOYMENT_FAILURE"]`.

I don't know why this is. Perhaps it is a bug in my code. However, if I change both resources from `TypeList` to `TypeSet`, both tests pass. I think it makes more sense for both `alarm_configuration` and `auto_rollback_configuration` to be of *TypeList* because they can only contain one item anyway, so making them *TypeSet* seems redundant. I'm not familiar with the inner workings of terraform to know what is going on here and I am unable to gleam anything from the DEBUG logs.

QUESTION: Is there a _known_ issue here? Should these be changed to `TypeSet`?

niclic:
### Notes

Some other rambling thoughts...

#### Auto Rollback Configuration

 - I don't think it is necessary to add validation to `auto_rollback_configuration.events`, like I did for `trigger_configuration.trigger_events`. The first commit of this PR is only necessary because the list of valid events in `validateTriggerEvent` needs to be updated. I suppose validation adds _some_ value, but it comes at the cost of having to keep these lists up to date. I'm not sure if that is worth it.

 - By default, the AWS SDK returns `"autoRollbackConfiguration":{"enabled":false}`, even if no `auto_rollback_configuration` is present in the terraform config/state. Since this can lead to config/state mismatch errors, the `auto_rollback_configuration` resource is only populated when:
  * `auto_rollback_configuration.enabled = true`, or
  * `len(auto_rollback_configuration.events) > 0` (i.e. previously disabled).

  This seems kinda smelly and can mean it is not possible to detect some unusual configuration updates, i.e.

  ```
  auto_rollback_configuration {
    enabled = false
  }
  ```

 - If I change `Optional: true` on any attributes to `Required: false`, I get some nasty errors. Since there are not many `Required: false` in the code base, I'm guessing there is a good reason for it.


#### Alarm Configuration

 -  The `alarm_configuration.enabled` attribute is designed to temporarily deactivate alarm monitoring for a deployment group without having to add the same alarms again later. By design, adding an alarm is enough to enable this feature. However, you will have to explicitly set `alarm_configuration.enable = true` to enable this feature.  I suppose additional logic could be added to enforce this. Perhaps there is an existing schema feature that does just this. Otherwise it's probably fine to leave it as is.

 - In the AWS SDK, the list of CloudWatch alarms is modeled as a `struct` called [`Alarm`](http://docs.aws.amazon.com/sdk-for-go/api/service/codedeploy/#Alarm), with just a single `Name` property and _NOT_ as a simple `[]*string` which is more common. I modeled this as a simple set of strings: `alarm_configuration.alarms`. I think this is OK, but perhaps there is a another approach.

 - When deleting the `alarm_configuration`, AWS requires that you send an object with an empty `alarms` list to avoid an `InvalidAlarmConfigException` error. After deletion, however, the response contains the following _empty_ configuration:

 ```
 "alarmConfiguration":{"alarms":[],"enabled":false,"ignorePollAlarmFailure":false}
 ```

 In order to avoid config/state mismatch errors, the `alarm_configuration` resource is only populated when:
  * `alarm_configuration.enabled = true`, or
  * `len(alarm_configuration.alarms) > 0`.

niclic:
After rebasing the latest master branch, the previously skipped tests now pass as they should. There must have been a bug fixed since this branch was created almost two months ago. I have enabled these tests again.

All tests pass. I think this PR is good to go.

stack72:
Hi @niclic 

Thanks for the work here - tests are good and it's working as expected

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCodeDeployDeploymentGroup_'
2016/12/01 16:00:10 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCodeDeployDeploymentGroup_ -timeout 120m
=== RUN   TestAccAWSCodeDeployDeploymentGroup_basic
--- PASS: TestAccAWSCodeDeployDeploymentGroup_basic (47.88s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_onPremiseTag
--- PASS: TestAccAWSCodeDeployDeploymentGroup_onPremiseTag (29.86s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_disappears
--- PASS: TestAccAWSCodeDeployDeploymentGroup_disappears (32.18s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_triggerConfiguration_basic
--- PASS: TestAccAWSCodeDeployDeploymentGroup_triggerConfiguration_basic (48.79s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_triggerConfiguration_multiple
--- PASS: TestAccAWSCodeDeployDeploymentGroup_triggerConfiguration_multiple (52.12s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_create
--- PASS: TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_create (57.07s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_update
--- PASS: TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_update (57.58s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_delete
--- PASS: TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_delete (39.52s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_disable
--- PASS: TestAccAWSCodeDeployDeploymentGroup_autoRollbackConfiguration_disable (44.98s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_create
--- PASS: TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_create (40.66s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_update
--- PASS: TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_update (44.07s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_delete
--- PASS: TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_delete (44.45s)
=== RUN   TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_disable
--- PASS: TestAccAWSCodeDeployDeploymentGroup_alarmConfiguration_disable (50.62s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	589.796s
```

Issue #9044, opened 2016-09-26
aws_default_security_group description key error


hanks:
terraform version: 0.7.4

code piece:

```
resource "aws_default_security_group" "default" {
  description = "${var.description}"

  vpc_id = "${var.vpc_id}"

  ingress {
    protocol  = -1
    self      = true
    from_port = 0
    to_port   = 0
  }

  egress {
    from_port   = 0
    to_port     = 0
    protocol    = "-1"
    cidr_blocks = ["0.0.0.0/0"]
  }
}
```

command: `terraform plan`

output: 
aws_default_security_group.default: : invalid or unknown key: description

refer docs:
[https://www.terraform.io/docs/providers/aws/r/default_security_group.html#description](https://www.terraform.io/docs/providers/aws/r/default_security_group.html#description)

hanks:
I checked the source code of [https://github.com/hashicorp/terraform/pull/8861/files](https://github.com/hashicorp/terraform/pull/8861/files), code as below:

```
func resourceAwsDefaultSecurityGroup() *schema.Resource {
	// reuse aws_security_group_rule schema, and methods for READ, UPDATE
	dsg := resourceAwsSecurityGroup()
	dsg.Create = resourceAwsDefaultSecurityGroupCreate
	dsg.Delete = resourceAwsDefaultSecurityGroupDelete

	// Descriptions cannot be updated
	delete(dsg.Schema, "description")

       ...
}
```

It seems do not support `description` argument, so the docs is error, and I create a pull request to update aws_default_security_group doc, #9046, would you check it?

stack72:
Fixed by #9046

PR #9046, opened 2016-09-26
Fix aws_default_security_group docs

hanks:
remove `description` section in Argument Reference of AWS_DEFAULT_SECURITY_GROUP,
because do not support description field according to the source code.

refer to issue #9044


catsby:
😱 thanks!

PR #9050, opened 2016-09-26
provider/aws: Add retry logic to the aws_ecr_repository delete func

### LINKED ISSUE #8597: aws_ecr_repository apply delete race condition
### Terraform Version
terraform v0.7.1

### Affected Resource(s)
- aws_ecr_repository

```hcl
resource "aws_ecr_repository" "repository" {
  name = "repository-name"
}
```

### Expected Behavior
On an apply and a following destroy the resource should always be created and destroyed

### Actual Behavior
sometimes the resource is not destroyed but still removed from the state file.
(Conjecture: Could be a race on Amazon side when creating a resource in that case the create should make sure the resource is up and usable).

### Steps to Reproduce
1. `terraform apply`
2. `terraform destroy`
3. repeat until error occurs (tried with 27 repositories up to 40 runs may be necessary)

We are running into this issue on our testing environment.  Our clean up step doesn't always clean up the freshly created repositories.

stack72:
Fixes #8597

There was sometimes an issue where Terraform was deleting the ECR
repository from the statefile before the reposity was actually deleted.

Added retry logic for Terraform to wait for the repository to be deleted
before proceeding with the statefile update

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEcrRepository_'
2016/09/26 12:46:57 Generated command/internal_plugin_list.go
-run=TestAccAWSEcrRepository_ -timeout 120m
=== RUN   TestAccAWSEcrRepository_importBasic
--- PASS: TestAccAWSEcrRepository_importBasic (17.86s)
=== RUN   TestAccAWSEcrRepository_basic
--- PASS: TestAccAWSEcrRepository_basic (16.40s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    34.288s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_ecr_repository.go
+++ b/builtin/providers/aws/resource_aws_ecr_repository.go
@@ -2,12 +2,14 @@ package aws
 
 import (
 	"log"
+	"time"
 
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_ecr_repository.go
```diff
@@ -117,7 +119,31 @@ func resourceAwsEcrRepositoryDelete(d *schema.ResourceData, meta interface{}) er
...
+			if awsErr.Code() == "RepositoryNotFoundException" {
+				return nil
+			}
+
+			return resource.NonRetryableError(err)
+		}
+
+		return resource.RetryableError(
+			fmt.Errorf("%q: Timeout while waiting for the ECR Repository to be deleted", d.Id()))
+	})
```
catsby: We don't check the `err` from `resource.Retry` here, so if the time limit hits we still remove from state, is that correct behavior?  If you hit any of the `resource.NonRetryableError` paths then we'll still remove from state because we don't check the error

catsby:
#8597 mentions a possible race condition. The description sounds as if we're attempting to delete this repo before it's actually created. Should we look into polling in the `CREATE` method to make sure the repo is available before moving on?

catsby:
The err check looks good; any thoughts on my question here? 

> Should we look into polling in the CREATE method to make sure the repo is available before moving on?

stack72:
After discussing with @catsby, we believe we are good with this solution and can merge it now :)

PR #9060, opened 2016-09-26
provider/aws: Make encryption in Aurora instances computed-only

radeksimko:
This is fixing two bugs:

1. The KMS Key ID was never read back from the API
2. It is only possible to set encryption (and hence KMS key ID too) on the cluster level, it is impossible to have a mixture of encrypted & unencrypted instances in the same cluster, as described [in the documentation](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html#Overview.Encryption.Limitations). Even though these fields have been marked as `Optional`, they were not actually used anywhere in any API calls when specified by the user.

## Test plan

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSRDSClusterInstance'
```
```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSRDSClusterInstance -timeout 120m
=== RUN   TestAccAWSRDSClusterInstance_importBasic
--- PASS: TestAccAWSRDSClusterInstance_importBasic (803.05s)
=== RUN   TestAccAWSRDSClusterInstance_basic
--- PASS: TestAccAWSRDSClusterInstance_basic (959.59s)
=== RUN   TestAccAWSRDSClusterInstance_kmsKey
--- PASS: TestAccAWSRDSClusterInstance_kmsKey (844.50s)
=== RUN   TestAccAWSRDSClusterInstance_disappears
2016/09/26 18:22:50 Request body type has been overwritten. May cause race conditions
--- PASS: TestAccAWSRDSClusterInstance_disappears (979.28s)
=== RUN   TestAccAWSRDSClusterInstance_withInstanceEnhancedMonitor
--- PASS: TestAccAWSRDSClusterInstance_withInstanceEnhancedMonitor (997.34s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	4583.792s
```

PR #9063, opened 2016-09-26
aws: Fix docs after changing Aurora instance schema

radeksimko:
I just realised the docs need changing too.

I really should've submitted this in https://github.com/hashicorp/terraform/pull/9060 😱  Sorry for the extra roundtrip.


stack72:
LGTM!

PR #9069, opened 2016-09-27
Allow new aws api gateway http_method and integration type [WIP]

gsfjohnson:
AWS API Gateway has been updated to permit an "ANY" http_method.  And integration type of AWS_PROXY (helpfully reported in aws web interface as LAMBDA_PROXY).

Pointing me toward existing PRs, like this one, that I could use as guidance would be appreciated.  Or next steps that are needed?

When using ANY http_method with AWS_PROXY integration type, the integration response resource is not created correctly.  Not sure why yet, so I have marked this [WIP].


kwilczynski:
@gsfjohnson hi there! Thank you for this! Any chance you could also update acceptance test?

stack72:
Hi @gsfjohnson 

Thanks for the work here - it seems another PR was merged in place of this so i am going to go ahead and close this now

Thanks again

Paul

PR #9088, opened 2016-09-27
ip_address is an array and requires the instance number

michelleheh:


paddyforan:
Agreed with @cblecker here. If you want to go ahead and update that, that'd be great. If not, and you don't mind enabling [edits for maintainers](https://github.com/blog/2247-improving-collaboration-with-forks), I can go ahead and make the change. However you'd like to move forward (including closing this PR and having someone else do it!) is totally fine. :)

PR #9102, opened 2016-09-28
network_id is computed

mcanevet:


svanharmelen:
@mcanevet nice one! Thanks 👍

PR #9116, opened 2016-09-28
provider/vsphere: Override to use adapterType if it is specified

king-jam:


king-jam:
Been open for months. Closing because probably not relevant anymore.

PR #9118, opened 2016-09-28
[WIP] core: Log diff mismatch using spew instead of %#v

jen20:
This commit improves the error logging for "Diffs do not match" errors by using the go-spew library to ensure that the structures are presented fully and in a consistent order. This allows use of the command line diff tool to analyse what is wrong.

Note that this is not necessarily intended to be merged in current form - we should instead log the message to a separate file and reference it in the error message so as not to present large walls of error message to users.

### DIFF
--- a/terraform/eval_diff.go
+++ b/terraform/eval_diff.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"strings"
 
+	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


mitchellh:
Dang it! I hit merge :( I reverted it but can't reopen this PR. I'll look at this though. I agree that single line output is better for users to copy and paste. We should put the spew output in the ogs as you said.

Issue #1058, opened 2015-02-26
DNSimple: Allow to set the record priority


sobrinho:
Hi,

For MX records we must be able to set the priority, example:

``` tf
resource "dnsimple_record" "aspmx" {
    domain = "${var.app_domain}"
    name = ""
    value = "ASPMX.L.GOOGLE.COM."
    type = "MX"
    ttl = 3600
    priority = 10
}
```

Makes sense?

sobrinho:
What can I tell for now is that attribute is set as computed: https://github.com/hashicorp/terraform/blob/master/builtin/providers/dnsimple/resource_dnsimple_record.go#L56

Probably changing this line to `Optional: true` would fix the problem but I have absolutely no idea how to boot the terraform development environment here.

phinze:
I'm doing a sweep of old issues without recent activity. This would be a welcome addition and we'd happily review any PR adding this feature. I took a quick look and unfortunately it's not as trivial as adding `Optional: true` to the field - it also needs to be added as a field to the API call.

For now I'm going to close this issue. Anybody who ends up here and is interested in picking up the work feel free to chime in! :+1:

fnmunhoz:
I missed this feature today. It would be great to have it in the future. Anyway, thanks :grinning:

PR #9128, opened 2016-09-29
Allow dnsimple_record.priority attribute to be set

yob:
Some DNS record types (like MX) allow a priority to specified, and the ability to do so is important in many environments.

This diff will change `dnsimple_record.priority` from computed to optional, allowing it to be used in terraform configs like so:

```
    resource "dnsimple_record" "mx1" {
        domain = "example.com"
        name = ""
        value = "mx1.example.com"
        type = "MX"
        priority = "1"
    }

    resource "dnsimple_record" "mx2" {
        domain = "example.com"
        name = ""
        value = "mx2.example.com"
        type = "MX"
        priority = "2"
    }
```

This requires an accompanying PR in github.com/pearkes/dnsimple (https://github.com/pearkes/dnsimple/pull/8) to be merged and then vendored in terraform.

An earlier issue (https://github.com/hashicorp/terraform/issues/1058) proposed this feature , but was eventually closed due to inactivity.

### DIFF
--- a/builtin/providers/dnsimple/resource_dnsimple_record.go
+++ b/builtin/providers/dnsimple/resource_dnsimple_record.go
@@ -57,6 +57,7 @@ func resourceDNSimpleRecord() *schema.Resource {
 			"priority": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
 			},
 		},
 	}
@@ -74,6 +75,10 @@ func resourceDNSimpleRecordCreate(d *schema.ResourceData, meta interface{}) erro
 
 	if ttl, ok := d.GetOk("ttl"); ok {
 		newRecord.Ttl = ttl.(string)
+	}
+
+	if priority, ok := d.GetOk("priority"); ok {
+		newRecord.Prio = priority.(string)
 	}
 
 	log.Printf("[DEBUG] DNSimple Record create configuration: %#v", newRecord)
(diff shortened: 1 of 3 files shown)


stack72:
Hi @yob 

thanks so much for the work here - as there are still some records that will create the computed value for the attribute, I believe the attribute will need to be computed and optional. This will stop perpetual diffs from occuring

If possible, please could you include an acceptance test for this change?

Thanks again

Paul

yob:
>  I believe the attribute will need to be computed and optional.

Ah, I'd wrongly assumed these two options were mutually exclusive. Can you point me to any docs that help explain the effect of both these options?

> If possible, please could you include an acceptance test for this change?

Sure, I'll have a crack at it

yob:
I've addressed the `Computed` feedback (but would still love to understand the underlying behaviour better), and added acceptance specs  for creating/updating an MX record with a priority.

They're the first terraform acceptance specs I've written, so all feedback is welcome!

I also assume they'll fail until we can get https://github.com/pearkes/dnsimple/pull/8 merged ad vendored

stack72:
Hi @yob 

Thanks so much for the extended work here - this is great - i will get this reviewed and then merged asap (when the upstream is merged)

With regards to Computed and Optional, there isn't really any external documentation on this I am afraid to say. This is a big of a failure on our part. The helper schema code is well documented though https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go

This may help understand what is and isn't allowed. In a nutshell:

* Required - mandatory field that must be set
* Optional - as expected

When these values are sent to the API and then the value set to state in the Read func, if there are any differences then we will get a perpetual diff on plan and apply. Therefore, if we know the values may be different (e.g. trailing dots on DNS names or uppercase on IPV6 addresses) we would add Computed to the parameter. This would tell terraform that changes may be allowed to the variable when it comes back from the API

Hope this helps

P.

skizz:
Any update on this PR? We are facing this same issue

yob:
I suspect this may not be needed now that terraform 0.9 has switched from @pearkes dnsimple client to the official client (See #10760).

PR #9136, opened 2016-09-29
Docs: Add import example for cloudfront_distribution

AMeng:


stack72:
Hi @AMeng 

Thanks for the work here - this LGTM! :)

Paul

PR #9139, opened 2016-09-29
deps: Update github.com/jen20/awspolicyequivalence

jen20:
This fixes a bug in the upstream library where different casing of `effect` values were not treated as equivalent.


stack72:
LGTM!

PR #9144, opened 2016-09-30
config/module: use the raw source as part of the key

mitchellh:
Fixes customer ticket 1806.

This changes the key for the storage to be the _raw_ source from the
module, not the fully expanded source. Example: it'll be a relative path
instead of an absolute path.

This allows the ".terraform/modules" directory to be portable when
moving to other machines. This was a behavior that existed in <= 0.7.2
and was broken with #8398. This amends that and adds a test to verify.

### DIFF
--- a/helper/copy/copy.go
+++ b/helper/copy/copy.go
@@ -1,0 +1,121 @@
+package copy
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// From: https://gist.github.com/m4ng0squ4sh/92462b38df26839a3ca324697c8cba04
+
+// CopyFile copies the contents of the file named src to the file named
+// by dst. The file will be created if it does not already exist. If the
+// destination file exists, all it's contents will be replaced by the contents
+// of the source file. The file mode will be copied from the source and
+// the copied data is synced/flushed to stable storage.
+func CopyFile(src, dst string) (err error) {
+	in, err := os.Open(src)
+	if err != nil {
+		return
+	}
+	defer in.Close()
+
+	out, err := os.Create(dst)
+	if err != nil {
+		return
+	}
+	defer func() {
+		if e := out.Close(); e != nil {
+			err = e
+		}
+	}()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return
+	}
+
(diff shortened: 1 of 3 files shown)


helper/copy/copy.go
```diff
@@ -0,0 +1,121 @@
...
+		// If the entry is a symlink, we copy the contents
+		for entry.Mode()&os.ModeSymlink != 0 {
+			target, err := os.Readlink(srcPath)
+			if err != nil {
+				return err
+			}
+
+			entry, err = os.Stat(target)
+			if err != nil {
+				return err
```
jbardin: Here's a prime example of having to be very careful with naked returns and name result parameters, since `err` was shadowed in this for loop. This is handled correctly here, but the mix of return values and naked returns in this code looks prime for introducing bugs later on. Can we drop the named result value for this function?

helper/copy/copy.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	defer in.Close()
+
+	out, err := os.Create(dst)
+	if err != nil {
+		return
+	}
+	defer func() {
+		if e := out.Close(); e != nil {
+			err = e
```
jbardin: You should check if `err == nil` before assigning `e` here, since an existing error is probably more important that the Close error.

helper/copy/copy.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	if !si.IsDir() {
+		return fmt.Errorf("source is not a directory")
+	}
+
+	_, err = os.Stat(dst)
+	if err != nil && !os.IsNotExist(err) {
+		return
+	}
+	if err == nil {
```
jbardin: passing a nil error to `os.IsNotExist` returns false, so you could just do a single check. Any other errors will be caught later anyway

```
	if !os.IsNotExist(err) {
 		return fmt.Errorf("destination already exists")
 	}

```

(only mention it because the `err == nil` check threw me for a second)

helper/copy/copy.go
mitchellh: I just copied this code from the gist at the top of the file. It isn't tested since it itself is only used in tests and the original source didn't come with it. I can add it if we want but I agree there is a LOT stylistically I disagree with in this code but it was a copy pasta.

Issue #3851, opened 2015-11-10
Problem with aws_iam_instance_profile roles


davidblewett:
The documentation for **aws_iam_instance_profile** suggests that you can attach multiple roles to a given instance profile. However, when I did so I got the following error:
- aws_iam_instance_profile.default: Error adding role dev_eyrie to IAM instance profile dev_eyrie: LimitExceeded: Cannot exceed quota for InstanceSessionsPerInstanceProfile: 1

I then looked up the documentation ( http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html ): 

```
The following are the default maximums for IAM entities:
Roles in an instance profile: 1 (each instance profile can contain only 1 role)
```

What I needed to do instead was create multiple **aws_iam_policy** and attach them to a single role, which is then assigned in the **aws_iam_instance_profile**.

davidblewett:
Hmm, but there doesn't appear to be a way to apply an indeterminate number of policies to a given role.  **iam_policy_attachment** is for a single policy, to multiple other users/groups/roles. I would like to be able to pass a list of policy IDs and attach them all to a single role.

davidblewett:
This could possibly be solved by #953. If the **iam_policy_attachment** resource doesn't support **count**, I can wrap it in a module and push in each policy ID via calls to **element**. It seems that **iam_policy_attachment** should support the count argument (maybe it does and there's just a bug in how it handles variable input?) .

I tried to invert the dependency chain, and attach policies to the instance role after the entire instance is created but ran into more problems. Different policies grant access to keys in S3 that are necessary during Chef provisioning. Since they wouldn't be granted until after the instance is created, Chef bails out with  an HTTP 403 Forbidden when trying to download the secrets.

jen20:
I've hit this in the past as well - I think the documentation needs updating in the first instance to reflect the Amazon API docs, and then we should consider further whether to do something to make this easier on users.

davidblewett:
If the **iam_policy_attachment** supported variable-based counts, this issue could be resolved fairly cleanly. Failing that, the module-level count would work as well. As it is, I've had to resort to explicitly defining all expected policy IDs as independent variables. This works for my use case for now, as it appears there's a relatively small number that I need to bubble through. If I get into a spot where only a subset of applications should have some IDs, then I would be truly stuck and have to resort to something like a Jinja or Mako pre-processor.

davidblewett:
[...]

PR #9146, opened 2016-09-30
provider/aws: aws_autoscaling_attachment resource

### LINKED ISSUE #8755: Enhance ELB Attachment statement with autoscaling groups
It's not possible to attach/detach an ELB to/from an existing Autoscaling Group without modifying the ASG. At the moment Terraform only provides an attachment statement for specific instances:

https://www.terraform.io/docs/providers/aws/r/elb_attachment.html

As it is possible to attach a loadbalancer to an existing Autoscaling Group (https://aws.amazon.com/es/blogs/aws/attach-and-detach-elastic-load-balancers-from-auto-scaling-groups/) Terraform should provide this option, too.

Don't know if it makes probably more sense to introduce a new command as the attachment is more a parameter in the autoscaling group data.

Btw. Here is my use case: I defined an autoscaling group with a few nodes in a core module and as the system evolves I deploy services on those instances. For each service I have a loadbalancer that should be attached to this running autoscaling group without having to modify the core module. Therefore it would be great to have something like:

```
# Create a new load balancer attachment
resource "aws_elb_attachment" "baz" {
  elb      = "${aws_elb.bar.id}"
  autoscaling_group = "${aws_autoscaling_group.server_group.id}"
}
```

jrnt30:
Fixes #8755 
- Introduces a new resource for attaching an ASG to an ELB.  

**Context:**
Adds support for associating an ASG with an ELB without having to declare this when the ASG is created.  The primary use case we wanted this for is to allow dynamically allocating ASGs to an ELB for Blue/Green deployments.  With the current process, there is not a clean way for us to do this without some hacking of variables. 

**NOTE:** I really struggled with the decision of keeping this as a unique resource or trying to branch the logic in the `aws_elb_attachment` resource.  In the end, I opted for the later because it seemed cleaner, as the internals of the two resources are fairly different and it made the implementation and testing cleaner. That being said, if you would prefer it merged with the other resource, I can take a stab at that. 

@kwilczynski This is what we were discussing in slack the other day.  Would love your 2 cents

### DIFF
--- a/builtin/providers/aws/resource_aws_autoscaling_attachment.go
+++ b/builtin/providers/aws/resource_aws_autoscaling_attachment.go
@@ -1,0 +1,110 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsAutoscalingAttachment() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsAutoscalingAttachmentCreate,
+		Read:   resourceAwsAutoscalingAttachmentRead,
+		Delete: resourceAwsAutoscalingAttachmentDelete,
+
+		Schema: map[string]*schema.Schema{
+			"group_name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+
+			"elb": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceAwsAutoscalingAttachmentCreate(d *schema.ResourceData, meta interface{}) error {
+	asgconn := meta.(*AWSClient).autoscalingconn
+	asgName := d.Get("group_name").(string)
+
+	elbName := d.Get("elb").(string)
+
+	attachElbInput := &autoscaling.AttachLoadBalancersInput{
+		AutoScalingGroupName: aws.String(asgName),
+		LoadBalancerNames:    []*string{aws.String(elbName)},
+	}
+
+	log.Printf("[INFO] registering asg %s with ELBs %s", asgName, elbName)
+
+	_, err := asgconn.AttachLoadBalancers(attachElbInput)
+	if err != nil {
+		return fmt.Errorf("Failure registering asg with ELBs: %s", err)
+	}
+
+	d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", asgName)))
+
+	return nil
+}
+
+func resourceAwsAutoscalingAttachmentRead(d *schema.ResourceData, meta interface{}) error {
+	asgconn := meta.(*AWSClient).autoscalingconn
+	asgName := d.Get("group_name").(string)
+	elbName := d.Get("elb").(string)
+
+	// Retrieve the ASG properites to get list of associated ELBs
+	asg, err := getAwsAutoscalingGroup(asgName, asgconn)
+
+	if err != nil {
+		return err
+	}
+	if asg == nil {
+		log.Printf("[INFO] Autoscaling Group %q not found", asgName)
+		d.SetId("")
+		return nil
+	}
+
+	found := false
+	for _, i := range asg.LoadBalancerNames {
+		if elbName == *i {
+			d.Set("elb", elbName)
+			found = true
+		}
+	}
+
+	if !found {
+		log.Printf("[WARN] Association for %s was not found in ASG assocation", elbName)
+		d.SetId("")
+	}
+
+	return nil
+}
+
+func resourceAwsAutoscalingAttachmentDelete(d *schema.ResourceData, meta interface{}) error {
+	asgconn := meta.(*AWSClient).autoscalingconn
+	asgName := d.Get("group_name").(string)
+
+	elbName := d.Get("elb").(string)
+
+	log.Printf("[INFO] Deleting ELB %s association from: %s", elbName, asgName)
+
+	detachOpts := &autoscaling.DetachLoadBalancersInput{
+		AutoScalingGroupName: aws.String(asgName),
+		LoadBalancerNames:    []*string{aws.String(elbName)},
+	}
+
+	_, err := asgconn.DetachLoadBalancers(detachOpts)
+	if err != nil {
+		return fmt.Errorf("Failure detaching ELB from ASG: %s", err)
+	}
+
+	return nil
+}
--- a/builtin/providers/aws/resource_aws_autoscaling_attachment.go
+++ b/builtin/providers/aws/resource_aws_autoscaling_attachment.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
@@ -17,7 +18,7 @@ func resourceAwsAutoscalingAttachment() *schema.Resource {
 		Delete: resourceAwsAutoscalingAttachmentDelete,
 
 		Schema: map[string]*schema.Schema{
-			"group_name": &schema.Schema{
+			"autoscaling_group_name": &schema.Schema{
 				Type:     schema.TypeString,
 				ForceNew: true,
 				Required: true,
@@ -34,8 +35,7 @@ func resourceAwsAutoscalingAttachment() *schema.Resource {
 
 func resourceAwsAutoscalingAttachmentCreate(d *schema.ResourceData, meta interface{}) error {
 	asgconn := meta.(*AWSClient).autoscalingconn
-	asgName := d.Get("group_name").(string)
-
+	asgName := d.Get("autoscaling_group_name").(string)
 	elbName := d.Get("elb").(string)
 
 	attachElbInput := &autoscaling.AttachLoadBalancersInput{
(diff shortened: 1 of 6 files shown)


kwilczynski:
@jrnt30 hi there! Thank you for this! Looks really good.

Few little comments.

Then, we would need @stack72 to look it through for us.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+func resourceAwsAutoscalingAttachment() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsAutoscalingAttachmentCreate,
+		Read:   resourceAwsAutoscalingAttachmentRead,
+		Delete: resourceAwsAutoscalingAttachmentDelete,
+
+		Schema: map[string]*schema.Schema{
+			"group_name": &schema.Schema{
```
kwilczynski: I would call the attribute `autoscaling_group_name` to be consistent with other resources in the same group.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceAwsAutoscalingAttachmentCreate(d *schema.ResourceData, meta interface{}) error {
+	asgconn := meta.(*AWSClient).autoscalingconn
+	asgName := d.Get("group_name").(string)
+
```
kwilczynski: Extra new line, I think.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+	log.Printf("[INFO] registering asg %s with ELBs %s", asgName, elbName)
+
+	_, err := asgconn.AttachLoadBalancers(attachElbInput)
+	if err != nil {
+		return fmt.Errorf("Failure registering asg with ELBs: %s", err)
+	}
+
+	d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", asgName)))
+
+	return nil
```
kwilczynski: Often, as a customary thing, we would invoke the `ReadFunc` so for example it would be:

```go
return resourceAwsAutoscalingAttachmentRead(d, meta)
```
This is so that it would re-read the resource back ensuring that AWS already created it, etc. Primarily as a precaution (or guard, if you wish) against the eventually consistent nature of the AWS.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+		log.Printf("[INFO] Autoscaling Group %q not found", asgName)
+		d.SetId("")
+		return nil
+	}
+
+	found := false
+	for _, i := range asg.LoadBalancerNames {
+		if elbName == *i {
+			d.Set("elb", elbName)
+			found = true
```
kwilczynski: I would break here, so that you don't check rest of them, as there will be no need at this point, I believe.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+		d.SetId("")
+	}
+
+	return nil
+}
+
+func resourceAwsAutoscalingAttachmentDelete(d *schema.ResourceData, meta interface{}) error {
+	asgconn := meta.(*AWSClient).autoscalingconn
+	asgName := d.Get("group_name").(string)
+
```
kwilczynski: Extra new line, I think.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+	log.Printf("[INFO] Deleting ELB %s association from: %s", elbName, asgName)
+
+	detachOpts := &autoscaling.DetachLoadBalancersInput{
+		AutoScalingGroupName: aws.String(asgName),
+		LoadBalancerNames:    []*string{aws.String(elbName)},
+	}
+
+	_, err := asgconn.DetachLoadBalancers(detachOpts)
+	if err != nil {
+		return fmt.Errorf("Failure detaching ELB from ASG: %s", err)
```
kwilczynski: See comment above about the `errwrap.Wrapf`.

builtin/providers/aws/resource_aws_autoscaling_attachment.go
```diff
@@ -0,0 +1,110 @@
...
+	elbName := d.Get("elb").(string)
+
+	log.Printf("[INFO] Deleting ELB %s association from: %s", elbName, asgName)
+
+	detachOpts := &autoscaling.DetachLoadBalancersInput{
+		AutoScalingGroupName: aws.String(asgName),
+		LoadBalancerNames:    []*string{aws.String(elbName)},
+	}
+
+	_, err := asgconn.DetachLoadBalancers(detachOpts)
```
kwilczynski: You can inline this to:

```go
if _, err := asgconn.DetachLoadBalancers(detachOpts); err != nil {
```

Since the `err` variable is not used below this line.

kwilczynski:
The good thing is, there isn't much in terms of needing to pool for state changes, which makes a lot of this much easier.

kwilczynski:
@jrnt30 as per the conversation on Slack, we might need to update the name of the resource to be something like e.g. `aws_autoscaling_elb_attachment`, or something similar that would allow people to get an idea what it does at a first glance.

jrnt30:
I held off on the renaming of the resource due to the conversation and potential expansion to also support an AutoScale group's attachment to the ALB Target Group resource.  Before I make those changes, I need to read up on that resource more so I am going to leave the name and resource attributes alone for right now. 

If it's decided that we would prefer to use two distinct resources for the attachment (which as a user I'm not sure I would expect) I can make the name more explicit.

jrnt30:
Also @kwilczynski I believe you may have commented on my explicit commit instead of the diff in the PR.  I believe all of your inline comments have been addressed, however the comments do not disappear as I would expect because the underlying file sha that you commented on (the previous commit) has not changed.

simonvanderveldt:
I do think `aws_autoscaling_attachment` is is a bit confusing, especially in relation to `aws_elb_attachment`.
Would it be problematic to use `aws_autoscaling_elb_attachment`?

jrnt30:
@simonvanderveldt I was a bit on the fence about this and am open to change.  There are several resources that also provide the ability to attach to an Application Load Balancer.  The idea was, since the resource's "api" footprint is essentially the same we may want to have those be the resource.  

Looking back at the other resources which support both of these (the autoscaling resource itself as an example), one common thread I see between them is that they seem to all call the same AWS API call. That might be a good enough litmus test for me to say "ALB and ELB resource associations should be two distinct resources, since they require different underlying AWS API calls for the attachment and detachment.

Honestly I don't feel very strongly one way or another, but would appreciate if there were some guidelines.  In general I think the question is, is it more explicit/maintainable/easier to approach if we break out the resources (alb vs. elb attachments being different in this case) or to keep them as a single resource?

simonvanderveldt:
@jrnt30 afaik Terraform's basic way of working is to mimic the API's it wraps as closely as possible. I guess that should be documented somewhere, but can't really find it.
I do know one example offhand, which is https://github.com/hashicorp/terraform/issues/3851#issuecomment-250618169

stack72:
So I have been looking through this. I believe this is the resource name we want. This will allow us the flexibility to use ELB and ALB target groups (going forward!)

Tests are looking good:

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAwsAutoscaling'
2016/11/03 12:29:03 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAwsAutoscaling -timeout 120m
=== RUN   TestAccAwsAutoscalingAttachment_basic
--- PASS: TestAccAwsAutoscalingAttachment_basic (111.94s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	111.954s
```

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAutoScalingGroup_'
2016/11/03 12:31:51 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAutoScalingGroup_ -timeout 120m
=== RUN   TestAccAWSAutoScalingGroup_importBasic
--- PASS: TestAccAWSAutoScalingGroup_importBasic (174.43s)
=== RUN   TestAccAWSAutoScalingGroup_basic
--- PASS: TestAccAWSAutoScalingGroup_basic (301.94s)
=== RUN   TestAccAWSAutoScalingGroup_autoGeneratedName
--- PASS: TestAccAWSAutoScalingGroup_autoGeneratedName (56.33s)
=== RUN   TestAccAWSAutoScalingGroup_terminationPolicies
--- PASS: TestAccAWSAutoScalingGroup_terminationPolicies (83.79s)
=== RUN   TestAccAWSAutoScalingGroup_tags
--- PASS: TestAccAWSAutoScalingGroup_tags (268.13s)
=== RUN   TestAccAWSAutoScalingGroup_VpcUpdates
--- PASS: TestAccAWSAutoScalingGroup_VpcUpdates (106.32s)
=== RUN   TestAccAWSAutoScalingGroup_WithLoadBalancer
--- PASS: TestAccAWSAutoScalingGroup_WithLoadBalancer (381.09s)
=== RUN   TestAccAWSAutoScalingGroup_withPlacementGroup
--- PASS: TestAccAWSAutoScalingGroup_withPlacementGroup (154.24s)
=== RUN   TestAccAWSAutoScalingGroup_enablingMetrics
--- PASS: TestAccAWSAutoScalingGroup_enablingMetrics (169.05s)
=== RUN   TestAccAWSAutoScalingGroup_withMetrics
--- PASS: TestAccAWSAutoScalingGroup_withMetrics (46.26s)
--- PASS: TestAccAWSAutoScalingGroup_ALB_TargetGroups (181.88s)
=== RUN   TestAccAWSAutoScalingGroup_initialLifecycleHook
--- PASS: TestAccAWSAutoScalingGroup_initialLifecycleHook (252.94s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	2176.410s
```

Thanks for all the work here @jrnt30 :)

P.

PR #9155, opened 2016-09-30
mostly typos in documens and aws autoscaling id instead of name

dnABic:
Documentation for aws resource aws_autoscaling_group in the section Attributes Reference it is written:

id - The autoscaling group name.

but it should be 

`id` - The autoscaling group id.

Also, I changed few more doc files replacing "Whethe" with "Whether" and "the the" with "the".


stack72:
Hi @dnABic 

Thanks so much for the updates here

Paul

PR #9159, opened 2016-10-01
provisioner/chef: make docs a little more explicit

svanharmelen:


svanharmelen:
Just a small doc update, so merging it right away...

PR #9163, opened 2016-10-01
provider/azurerm: fix resource ID parsing for subscriptions resources

pmcatominey:
If a resource ID has more than one subscriptions key, as is the case for Service
Bus subscriptions the Azure SubscriptionID field was overwritten by the second
value.

TF_ACC= go test ./builtin/providers/azurerm -run TestParseAzureResourceID -timeout=30s -parallel=4
ok  	github.com/hashicorp/terraform/builtin/providers/azurerm	0.060s


stack72:
LGTM! Thanks @pmcatominey :)

stack72:
I am going to let this soak in the acceptance tests tonight to make sure all is well before the 0.7.5 release

PR #9172, opened 2016-10-02
Documentation Update: Changing virtualization type in documentation for aws_instance

dnABic:
In current Terraform documentation for resource aws_instance and aws_launch_configuration, virtualization type is still "paravirtual" although AWS documentation suggests "hvm":

"For the best performance, we recommend that you use current generation instance types and HVM AMIs when you launch your instances."

Source: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html

Therefore, I though it would make sense to update Terraform documentation for aws_instance and aws_launch_configuration and have virtualization type "hvm" instead of "paravirtual".

What do you think?

### DIFF
--- a/website/source/docs/providers/aws/r/instance.html.markdown
+++ b/website/source/docs/providers/aws/r/instance.html.markdown
@@ -24,11 +24,11 @@ data "aws_ami" "ubuntu" {
   most_recent = true
   filter {
     name = "name"
-    values = ["ubuntu/images/ebs/ubuntu-trusty-14.04-amd64-server-*"]
+    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
   }
   filter {
     name = "virtualization-type"
-    values = ["paravirtual"]
+    values = ["hvm"]
   }
   owners = ["099720109477"] # Canonical
 }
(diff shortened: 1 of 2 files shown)


catsby:
Hey @dnABic – I'm not strictly opposed to this, however, that `hvm-ssd` type doesn't work with `t1.micro`. Switching to a `t2.micro` doesn't work either in EC2 Classic (it requires a VPC), so we'd need to make those changes 

- [ ] `t1`->`t2`
- [ ] `us-east-1` -> `us-west-2`

dnABic:
Hi @catsby - I have made changes as requested. Still, one of tests at Travis failed. I am not sure why. 

But also, in file launch_configuration.html.markdown I couldn't put t2.micro for the section which describes spot instances. Seems that t2 instance type can not be requested as spot instance. So, I put m4.large instead because that instance type is available as spot instance in all regions. Couldn't find any other cheaper instance type that is available across all regions.  But in that case value:

spot_price = "0.001"

should be changed into something more realistic?

catsby:
I think the changes you've made are fine for now. Is this OK to merge, or still a WIP ?

dnABic:
This is not WIP anymore. Please merge.

catsby:
Thanks!

PR #9173, opened 2016-10-03
[Provider] Rancher

johnrengelman:
This integrates the rancher provider from https://github.com/platanus/terraform-provider-rancher. 
I've tried to retain the commit history for the interesting parts here in order to give credit to the original authors.

/cc @raphink @blackjid


johnrengelman:
@stack72 how do we go about providing the external system for the acceptance tests? Is there some scripting that we can add stuff to spin up docker containers or something?

raphink:
@johnrengelman did you also take note of https://github.com/rancher/go-rancher/pull/121 ?

johnrengelman:
@raphink - yeah, this is still using your patched catalog client.

blackjid:
nice thanks @johnrengelman , I had the same question you ask @stack72 for the tests...

what do you think about version compatibility?? this provider should be compatible with rancher api v1 and v2 that is coming later this month?? (also I don't really know if rancher 1.2.0 will support both api versions)

johnrengelman:
I'm not sure on compatibility...there were definitely some significant changes (i.e. `Environment` -> `Stack`, etc.), Not sure if the code can logically decide between the two based on a property, might need to have different implementations.

raphink:
https://github.com/rancher/go-rancher/pull/124 seems to suggest v2 should be used as soon as it is available.

johnrengelman:
There are 2 different clients though: `github.com/rancher/go-rancher/client` (the original), and `github.com/rancher/go-rancher/v2` (the v2 client).
Ultimately the TF provider needs to handle operating with both the 1.1.x line of Rancher and the 1.2.x when it's released.
Perhaps @ibuildthecloud could provide some direction.

stack72:
@johnrengelman 

https://github.com/hashicorp/terraform/pull/9173#issuecomment-251104579

Is Rancher PaaS or do we have to build infra to run these tests?

johnrengelman:
@stack72 there is not a hosted solution that I know of, but it's a simple as doing `docker run -d rancher/server:latest` to get a running version.

stack72:
If that's the case, then I can test the provider and add it to our nightly tests easily enough

johnrengelman:
@stack72 - great!

@raphink - I'm running the acceptance tests now and I'm having issues with them. It appears the code is configured to require access control to be enabled on Rancher, but I'm not sure that should be a requirement. I could conceive someone running an authenticated cluster on a private network (or in CI tests). Thoughts?

blackjid:
@johnrengelman I don't think you need the access control enabled in rancher.. wWe are in fact running ci on travis against a newly created rancher server without any access control.

builtin/providers/rancher/resource_rancher_registration_token.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
```
mcanevet: We should use `Optinal: true` here so that we can terraformize what exists without having to update to token name in the database using SQL queries because this attribute is not updatable using the API.

stack72:
Hi @johnrengelman 

Please ping me when you feel this is worthy of final review / testing :)

P.

johnrengelman:
@stack72 will do. sorry been out sick the last few days. There's still some issues in here that I am addressing.

stack72:
You take as long as you need sir! :)

blackjid:
@johnrengelman I see you added in the docs how to import these resources. The fact is the code base code for import is there, but it's only tested for the `rancher_environment` resource. It would be good to add the test for the other resources if we are going to maintain the import feature, which is awesome by the way.

johnrengelman:
Yup. I'm still working on this PR. When it's ready for review, I'll remove
the WIP and comment in here.

johnrengelman:
@raphink @blackjid Ok, I've got everything in working except 1 thing: Creating a stack from a catalog entry. 
I get the following error: 
```
--- FAIL: TestAccRancherStack (6.60s)
	testing.go:265: Step 3 error: Error applying: 1 error(s) occurred:

		* rancher_stack.catalog: Failed to get catalog template: Unknown schema type [template]
```

Looks like, rancher doesn't respond to `/v1-catalog/template`. It wants `/v1-catalog/templates`.

johnrengelman:
Nevermind, when I vendor'd your fork of `go-rancher` I didn't get the correct branch.

raphink:
Great, thank you!

builtin/providers/rancher/provider.go
```diff
@@ -18,14 +17,14 @@ func Provider() terraform.ResourceProvider {
...
-				DefaultFunc: schema.EnvDefaultFunc("RANCHER_ACCESS_KEY", nil),
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("RANCHER_ACCESS_KEY", ""),
 				Description: descriptions["access_key"],
 			},
 			"secret_key": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("RANCHER_SECRET_KEY", nil),
+				Optional:    true,
```
blackjid: Awesome, makes total sense if we can use them from the environment.

blackjid:
What would be the next steps to get this merged?? integrating with CI on the terraform side?

builtin/providers/rancher/provider.go
johnrengelman: The point here is that Rancher doesn't require API keys. You could have a rancher server that does not have access control enabled, so it wouldn't make sense to force the user to provide these values then.

johnrengelman:
I'm finishing up the stack upgrade piece today and then we should be set.

johnrengelman:
@blackjid @raphink @stack72 this is ready for review.

For acceptance tests this is what I've done locally:
1. `docker run -d -p 8080:8080 --name rancher rancher/server:stable` (runs the rancher server)
2. `curl -X POST http://localhost:8080/v1/projects/1a5/registrationTokens -d '{"name": "test"}'` (create a registration token in the "Default" environment)
3. Then run this script to add a local host to Rancher (needed for the stack acc tests)
```
TOKEN=`curl --silent localhost:8080/v1/projects/1a5/registrationTokens | jq -r '.data[] | select(.accountId=="1a5")'`
IMAGE=`echo $TOKEN | jq -r '.image'`
URL=`echo $TOKEN | jq -r '.registrationUrl'`
IP=`ifconfig en0 | grep "inet " | awk '{print $2}'`
MOD_URL=`echo $URL | sed 's/localhost/'${IP}'/'`

docker run -d --privileged -e CATTLE_AGENT_IP="${IP}" -v /var/run/docker.sock:/var/run/docker.sock $IMAGE $MOD_URL
```

builtin/providers/rancher/resource_rancher_stack.go
```diff
@@ -148,24 +151,101 @@ func resourceRancherStackUpdate(d *schema.ResourceData, meta interface{}) error
...
+
+		d.SetPartial("name")
+		d.SetPartial("description")
+	}
+
+	if d.HasChange("docker_compose") ||
+		d.HasChange("rancher_compose") ||
+		d.HasChange("environment") ||
+		d.HasChange("catalog_id") ||
+		d.HasChange("scope") {
```
raphink: Does a change in scope really require an upgrade?

builtin/providers/rancher/resource_rancher_stack.go
```diff
@@ -148,24 +151,101 @@ func resourceRancherStackUpdate(d *schema.ResourceData, meta interface{}) error
...
+
+		stateConf := &resource.StateChangeConf{
+			Pending:    []string{"active", "upgrading"},
+			Target:     []string{"active"},
+			Refresh:    StackStateRefreshFunc(client, stack.Id),
+			Timeout:    10 * time.Minute,
+			Delay:      1 * time.Second,
+			MinTimeout: 3 * time.Second,
+		}
+		s, waitErr := stateConf.WaitForState()
```
raphink: Should Terraform finish the upgrade when there was no error, and rollback it otherwise?

johnrengelman: I don't know about auto-rolling back. Is there any precedent for doing that in other providers?
As for finishing the upgrade, it's an config option you can set on the resource: https://github.com/hashicorp/terraform/pull/9173/commits/fd856e51b20d442720b04749e0ac3bd75f752bbb#diff-52a80e386a0d139d6f657ff76d1a692cR70

builtin/providers/rancher/resource_rancher_stack.go
johnrengelman: It doesn't. I'll fix this.

raphink: Ah great, sorry.

stack72:
Hi @johnrengelman 

Please can you rebase this from master? :)

P.

johnrengelman:
@raphink @blackjid - I'd love to keep your're guys' commits in here but it's problematic when rebasing and such, it makes for a ton of work....so, I'm going to squash them down into single commit. I'll mention you both int he commit message though.

blackjid:
hey @johnrengelman now problem, thanks for that.. 👍

johnrengelman:
@stack72 branch is updated and i squashed down the commits.

stack72:
@raphink @blackjid are you both happy here with the provider and that it is all ok?

blackjid:
LGTM

raphink:
@stack72 fine with me.

raphink:
Any news on that?

johnrengelman:
@stack72 anything I can do to help shepherd this along?

raphink:
@stack72 any plan to merge this?

raphink:
@johnrengelman @blackjid shall we get back to the platanus project for new features/bugfix until this gets merged?

blackjid:
that make sense, but I have the feeling the merging the further changes back to this PR is not going to be easy. at least is going to be super manual.. right @johnrengelman???

stack72:
Hi all

Sorry I haven't been back to you all on this one - I am trying to work out how to get tests in place for this. We currently have no infrastructure running Rancher so I am trying to work out how to script it that it comes up as part of our test harnsess

I promise I will get back to this asap

P.

blackjid:
@stack72 just as a reminder, I have some tests running in the plugin version of this provider... if that helps..
https://github.com/platanus/terraform-provider-rancher/blob/master/bin/cibuild

mcanevet:
@stack72 FYI we are using this branch in production for a few weeks now (since the go for merging actually) and it works pretty well.

stack72:
hi @johnrengelman / @raphink 

This LGTM!  I am going to come back and setup the testing for this to our CI server later in the week

Thanks for all the work here

Paul

PR #9174, opened 2016-10-03
provider/vsphere: Allow spaces in datastore name

matthewpucc:
I ran into an issue where when the datastore contained a space in the name, the original split function would error out because it was splitting on " ". This now uses regex and parses based on the default vmware resource format.

I didn't see any tests covering this feature set so no tests were changed. If you want me to figure out a way to write additional tests to cover this, let me know.


matthewpucc:
Sorry. I had the wrong go func for regexp. I updated the code and retested.

andreabedini:
@matthewpucc does this address the same problem as in  #10089?

matthewpucc:
@andreabedini yes. That's how the error manifested itself.

markpeek:
@andreabedini thanks for pointing me to this PR. @matthewpucc another way to fix this issue is to use a new helper added to govmomi (https://github.com/vmware/govmomi/commit/e152c3553e1101445babdb94082a36df9fd3ccf3) which will parse a datastore path. This would require a new release of govmomi and vendoring it. The upside would be pushing the logic into that library which also has tests included. I could help make this happen if you think this is a better approach.

matthewpucc:
@markpeek I like that approach too. Being able to cover this with tests make me happier.

andreabedini:
The helper seems to have made it to govmomi v0.12.1. What else has to be done to make this happen?

j00p34:
@markpeek @matthewpucc Any update on this? I am also hitting this problem and would really like getting this resolved.

alexholliz:
I've just run into the same problem:

`vsphere_virtual_machine.win2012_test: [ERROR] Failed trying to parse disk path: [Compellent VMWare 9] win2012-test/win2012-test.vmdk`

matthewpucc:
@markpeek I went ahead and updated this based on your preferred method since there was a lack of communication. I updated the vendored govmomi module in the vendor folder and ported the disk searching code to use that.

matthewpucc:
please let me know how you guys want to coordinate testing beyond a passing unit test suite

alexholliz:
Just a quick note, I was able to merge this PR with the current Terraform master and build it, and it solved my problem with Datastores. Looking forward to it being merged!

PR #9175, opened 2016-10-03
Provider/aws/fix acceptance test log bucket

PayscaleNateW:
* Factored out the s3 buckets used in the various test configs
* Made sure to define the logs bucket as part of the config.  Since it wasn't defined, it was giving me errors for not being found
* Also fixed up the bucket naming.  I don't think the bucket should be named <name>.s3.amazonaws.com. I think that adding the domain there is a convention for accessing the bucket, but the name should just be <name>. So for example, the name might be mylogs.73178316, and then in the logging config bucket you list mylogs.73178316.s3.amazonaws.com

### DIFF
--- a/builtin/providers/aws/import_aws_cloudfront_distribution_test.go
+++ b/builtin/providers/aws/import_aws_cloudfront_distribution_test.go
@@ -1,12 +1,17 @@
 package aws
 
 import (
+	"fmt"
 	"testing"
 
+	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
 func TestAccAWSCloudFrontDistribution_importBasic(t *testing.T) {
+	ri := acctest.RandInt()
+	testConfig := fmt.Sprintf(testAccAWSCloudFrontDistributionS3Config, ri, originBucket, logBucket, testAccAWSCloudFrontDistributionRetainConfig())
+
 	resourceName := "aws_cloudfront_distribution.s3_distribution"
 
 	resource.Test(t, resource.TestCase{
@@ -15,7 +20,7 @@ func TestAccAWSCloudFrontDistribution_importBasic(t *testing.T) {
 		CheckDestroy: testAccCheckCloudFrontDistributionDestroy,
 		Steps: []resource.TestStep{
 			resource.TestStep{
-				Config: testAccAWSCloudFrontDistributionS3Config,
+				Config: testConfig,
 			},
 			resource.TestStep{
 				ResourceName:      resourceName,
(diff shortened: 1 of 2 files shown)


stack72:
Hi @PayscaleNateW 

Thanks so much for the work here. I just ran the tests and all worked except the first:

```
=== RUN   TestAccAWSCloudFrontDistribution_importBasic
--- PASS: TestAccAWSCloudFrontDistribution_importBasic (1146.30s)
=== RUN   TestAccAWSCloudFrontDistribution_S3Origin
--- FAIL: TestAccAWSCloudFrontDistribution_S3Origin (29.99s)

		* aws_s3_bucket.s3_bucket_origin: NoSuchBucket: The specified bucket does not exist
			status code: 404, request id: C263F98949EC55D2
=== RUN   TestAccAWSCloudFrontDistribution_S3OriginWithTags
--- PASS: TestAccAWSCloudFrontDistribution_S3OriginWithTags (1182.16s)
=== RUN   TestAccAWSCloudFrontDistribution_customOrigin
--- PASS: TestAccAWSCloudFrontDistribution_customOrigin (1192.55s)
=== RUN   TestAccAWSCloudFrontDistribution_multiOrigin
--- PASS: TestAccAWSCloudFrontDistribution_multiOrigin (1161.27s)
=== RUN   TestAccAWSCloudFrontDistribution_noOptionalItemsConfig
--- PASS: TestAccAWSCloudFrontDistribution_noOptionalItemsConfig (1071.63s)
```

I believe the issue here is that you changed the config that the first test depends on but didn't pass in both buckets to that test config

Please can you update the test and then we can merge this? :)

Thanks for all the work here

Paul

PayscaleNateW:
Ah I noticed that as well... Wasn't sure if it was a separate issue to be fixed next, because the error there only happens in the sequence of tests, not when you run TestAccAWSCloudFrontDistribution_S3Origin by itself. I think I figured out the issue:

Both TestAccAWSCloudFrontDistribution_importBasic and TestAccAWSCloudFrontDistribution_S3Origin use the *same* config, with the same random int for their aws buckets. The random int doesn't get regenerated. Probably want to regenerate it so each test is independent.

stack72:
YEs, now that the test has been updated to take the strings, the config must also be done to do that in the test :)

PayscaleNateW:
I just pushed another commit that fixes that issue, by supplying the import test and the s3config test with different random ints, so they don't collide.

Acceptance tests seem to be passing for me!

stack72:
Nice work @PayscaleNateW 

Thanks for the work here! Will get this tested now and then merged asap

P.

stack72:
New test looks good :)

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCloudFrontDistribution_S3Origin'
2016/10/03 23:15:38 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCloudFrontDistribution_S3Origin -timeout 120m
=== RUN   TestAccAWSCloudFrontDistribution_S3Origin
--- PASS: TestAccAWSCloudFrontDistribution_S3Origin (1007.72s)
=== RUN   TestAccAWSCloudFrontDistribution_S3OriginWithTags
--- PASS: TestAccAWSCloudFrontDistribution_S3OriginWithTags (1176.16s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	2183.897s
```

PayscaleNateW:
Awesome, thanks so much @stack72!

PR #9178, opened 2016-10-03
Use new Datadog GraphDefinitionRequest struct

saliceti:
We want to add features to the dashboard datadog terraform provider, that relies on the [go-datadog-api](https://github.com/zorkian/go-datadog-api). But at the moment any change would break the terraform code.
[This new struct](https://github.com/alphagov/paas-terraform/blob/datadog_dashboard/builtin/providers/datadog/resource_datadog_timeboard.go#L122-L124) avoids requiring to repeat the struct definition in this code [here](https://github.com/hashicorp/terraform/blob/b79aea491bd8d3c2e83236b6f265938469114530/builtin/providers/datadog/resource_datadog_timeboard.go#L122-L127). It avoids duplication and makes it more flexible so more options can be added to the struct without breaking the code here.

This PR must be merged alongside [this PR on go-datadog-api](https://github.com/zorkian/go-datadog-api/pull/58).


saliceti:
Will recreate with the required vendored dependency when available.

PR #9187, opened 2016-10-03
Edits to remote-config documentation

jamtur01:
Minor grammar and style edits.

Also updated Consul example to not use `demo.consul.io` - seems like a bad idea to hard-code something folks might cut-n-paste to a host that's public. Principle of least risk, etc.


stack72:
LGTM! Thanks @jamtur01 :)

PR #9188, opened 2016-10-03
provider/aws: Add retry to the `aws_ssm_document` delete func

stack72:
As noticed in the acceptance tests, we were expecting the document to be
deleted but it was still found

```
=== RUN   TestAccAWSSSMDocument_permission
--- FAIL: TestAccAWSSSMDocument_permission (5.60s)
    testing.go:329: Error destroying resource! WARNING: Dangling
    resources

                            Error: Check failed: Expected AWS SSM
                            Document to be gone, but was still found

                                            State: <no state>
```

### DIFF
--- a/builtin/providers/aws/resource_aws_ssm_document.go
+++ b/builtin/providers/aws/resource_aws_ssm_document.go
@@ -1,12 +1,16 @@
 package aws
 
 import (
+	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_ssm_document.go
```diff
@@ -215,11 +219,38 @@ func resourceAwsSsmDocumentDelete(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	_, err := ssmconn.DeleteDocument(params)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Waiting for SSM Document %q to be deleted", d.Get("name").(string))
```
catsby: `%q` here?

Issue #9170, opened 2016-10-02
Add note about using US East 1 ACM certs for CloudFront


robert-stuttaford:
Just a gotcha that could help some folks in the future.

According to https://forums.aws.amazon.com/thread.jspa?messageID=722324, ACM certs should be generated in US East 1 if they are to be used with CloudFront.

It'd save folks time if https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html#acm_certificate_arn had a note to this effect.

stack72:
Closed via 9190

PR #9190, opened 2016-10-03
DOCS: Added note about us-east-1 ACM certs.

nwalke:
Should note that ACM certs must be in US-EAST-1 in order to be used with CloudFront.  Relevant to current version.

#9170


stack72:
Thanks for the addition here @nwalke :) This LGTM!

P.

Issue #6427, opened 2016-04-30
Consul module requires a VPC


chalford:
Hi there,

### Terraform Version
Terraform v0.6.15

### Affected Resource(s)
- module.consul.aws_security_group.consul
- https://github.com/hashicorp/terraform/blob/master/website/source/intro/getting-started/modules.html.md

### Terraform Configuration Files
```hcl
resource "aws_key_pair" "ssh_admin_key" {
  key_name = "ssh_admin_key"
  public_key = "[MY PUBLIC KEY]"
}

module "consul" {
    source = "github.com/hashicorp/consul/terraform/aws"

    key_name = "${aws_key_pair.ssh_admin_key.key_name}"
    key_path = "[MY PRIVATE KEY PATH]"
    region = "eu-west-1"
    servers = "3"
}

output "consul_address" {
    value = "${module.consul.server_address}"
}
```

### Debug Output
https://gist.github.com/chalford/7da82451af645f5d7a2f920374de40b6

### Expected Behavior
3 consul servers and a security group should have been created

### Actual Behavior
The security group failed to create, as there was an egress parameter set. The consul module is used as an example in the Getting Started section of the web site, but the neither that section or the consul terraform readme tells me to create a VPC first.

### Steps to Reproduce
1. Add consul module to local terraform configs (see above)
2. `terraform get`
3. `terraform apply`

### Important Factoids
- I have not setup a VPC in my AWS account, it is almost empty of resources.

stack72:
Closed via #9191

PR #9191, opened 2016-10-03
DOCS: Added note about requiring a default VPC.

nwalke:
The module being used requires a default VPC set up in the account.  This is relevant to the current version of Terraform.

#6427


stack72:
LGTM! thanks @nwalke :)

PR #9203, opened 2016-10-04
Fixup some SQL DB docs

robrankin:
* Parameter is Objective, not Object
* Clarify what resource group needs to be used
* Name parameter is the database name
* Link to Azure docs for SQL Database Tiers/Editions/Size limits


stack72:
Hi @robrankin 

Thanks so much for these changes :)

Paul

PR #9073, opened 2016-09-27
provider/azurerm: set resource_group_name on resource import

carinadigital:
Fix for GH-9072

I will add more resources as I test them.

PR #9208, opened 2016-10-04
provider/azurerm: Ensure availability set location is lowercase

### LINKED ISSUE #8183: AzureRM - issue with Location and Availability sets
**Version**
 v 0.7.0 

**Affected Resource**
azurerm_availability_set

**Issue Details**

I have an availability set in Azure (Resource Manager) with 2 VM's. On every run my location is seen as different (as appears case sensitive) - this results in the forced creation of new resources within the availability set - resulting in the deletion and re-creation of all vm's in the availability set. It doesn't seem to matter what case I specify the location as - on subsequent runs it always returns as a change and wants to recreate the resource group and the VM's within the group.

Results from a terraform plan below:-

```
-/+ module.availability_set_test.azurerm_availability_set.availability_set
location: "UKSouth2" => "uksouth2" (forces new resource)
name: "availability-set-test" => "availability-set-test"
platform_fault_domain_count: "3" => "3"
platform_update_domain_count: "5" => "5"
resource_group_name: "test-prod-rg" => "test-prod-rg"
tags.%: "1" => "1"
tags.environment: "production" => "production"
```

Using a proxy between Terraform and Azure i can see that Azure responds with location in lowercase for most resources:

```
    "kind": "Storage",
    "location": "uksouth2",
    "name": "testdisk007",

    "type": "Microsoft.Network/networkSecurityGroups",
    "location": "uksouth2",
```

however for Availability sets it returns the location as:-

```
    "name": "availability-set-test",
    "type": "Microsoft.Compute/availabilitySets",
    "location": "UKSouth2",
```

This causes Terraform to recreate the Availability set and also the VM's within the availability set on every plan/apply.

I did not see the same behavior on version 0.6.16

titanous:
For some regions it is returned from the API in camel case.

Before:

```
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMAvailabilitySet_basic -timeout 120m
=== RUN   TestAccAzureRMAvailabilitySet_basic
--- FAIL: TestAccAzureRMAvailabilitySet_basic (64.70s)
        testing.go:265: Step 0 error: Check failed: Check 4/4 error: azurerm_availability_set.test: Attribute 'location' expected "canadacentral", got "CanadaCentral"
FAIL
exit status 1
FAIL    github.com/hashicorp/terraform/builtin/providers/azurerm    64.715s
```

After:

```
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMAvailabilitySet_basic -timeout 120m
=== RUN   TestAccAzureRMAvailabilitySet_basic
--- PASS: TestAccAzureRMAvailabilitySet_basic (65.01s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/azurerm    65.022s
```

Closes #8183

### DIFF
--- a/builtin/providers/azurerm/resource_arm_availability_set.go
+++ b/builtin/providers/azurerm/resource_arm_availability_set.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/arm/compute"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -129,7 +130,7 @@ func resourceArmAvailabilitySetRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("platform_update_domain_count", availSet.PlatformUpdateDomainCount)
 	d.Set("platform_fault_domain_count", availSet.PlatformFaultDomainCount)
 	d.Set("name", resp.Name)
-	d.Set("location", resp.Location)
+	d.Set("location", strings.ToLower(*resp.Location))
 
 	flattenAndSetTags(d, resp.Tags)
 
--- a/builtin/providers/azurerm/resource_arm_availability_set_test.go
+++ b/builtin/providers/azurerm/resource_arm_availability_set_test.go
@@ -28,6 +28,8 @@ func TestAccAzureRMAvailabilitySet_basic(t *testing.T) {
 						"azurerm_availability_set.test", "platform_update_domain_count", "5"),
 					resource.TestCheckResourceAttr(
 						"azurerm_availability_set.test", "platform_fault_domain_count", "3"),
+					resource.TestCheckResourceAttr(
+						"azurerm_availability_set.test", "location", "canadacentral"),
 				),
 			},
 		},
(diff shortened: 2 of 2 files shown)


carinadigital:
In fixing importing of state resource in #9073 I found 4 resources that suffer this problem and suggested a fix. 
See https://github.com/hashicorp/terraform/pull/9073/commits/82cc9d3e560dfedc8f5a67517a6122978cfa6b3c

carinadigital:
Strangely TestAccAzureRMAvailabilitySet_importBasic wasn't one of them. 
Some region's API's must behave differently.

ms1111:
I applied this patch against master locally and it helped me out. Thanks! Would love to see it in the next release as it's not possible to manage availability sets in the CanadaCentral region without it.

pmcatominey:
Hi @titanous,

I've noticed the same behaviour on some of the resources I've written, I used the following to resolve it:
```go
d.Set("location", azureRMNormalizeLocation(*resp.Location))
```

The normalize function handles the casing too, I'd say this would be more preferable than just `strings.ToLower`.

Cheers

carinadigital:
@pmcatominey You could create a simple diff suppress function that could be added via the schema. It might be a neater way to solve this across the different resource types as I'm sure `availability set` isn't the only one affected.

stack72:
FWIW, I love that idea of a DiffSuppress func - it would be a good way of being able to work for lots of params

pmcatominey:
@stack72 could roll this into a `locationSchema()` similar for tags for all resources?

stack72:
@pmcatominey I am not against it, we would have to do the same for Resource Groups as well though right?

pmcatominey:
@stack72 yep, schema func and diff suppress for both sounds good, can let them soak in acceptance tests

pmcatominey:
This should be resolved by #10409

stack72:
Closed via #10409

PR #9217, opened 2016-10-04
Add scaleway server ipv6 address attribute

lennart:
adds the `ipv6` attribute to the scaleway provisioner.

e.g.:

```
resource "scaleway_server" "demov6" {
  ...
  enable_ipv6 = true
  ...
  connection {
    host = "${self.ipv6}"
  }
}

```

allows to use ipv6 address of a scaleway server in e.g. ssh connections for remote provisioning of hosts without public ip.

needs `enable_ipv6` to be `true` to work.

no tests, yet

### DIFF
--- a/builtin/providers/scaleway/resource_scaleway_server.go
+++ b/builtin/providers/scaleway/resource_scaleway_server.go
@@ -61,6 +61,10 @@ func resourceScalewayServer() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"ipv6": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"state": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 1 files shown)


stack72:
Morning @lennart 

Thanks for the PR here - you said "no acceptance tests yet" - does this mean you are working on some to go with the PR?

Please can you also update the documentation

Thanks

Paul

lennart:
I haven't looked into how you test terraform, are there docs on how to add some?

regarding the docs, I'll add them later today

…

stack72:
Thanks for planning to add the docs. There is a short guide on the acceptance tests:

https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#writing-acceptance-tests

Let me know if there are any issues

Paul

nicolai86:
@lennart do you think it would make sense to name the attribute `public_ipv6` ? Scaleway does not expose a private IPv6 address but this would make it very clear.

lennart:
@nicolai86 yes might be wise.

@stack72 actually I am not yet sure if the implementation is enough like this.

a conceptual question to be answered first is: how does `server.public_ipv6` behave when `enable_ipv6` is `false`?

I haven't yet fully answered this and the current implementation will in this case lead to a runtime failure if you use the ipv6 in e.g. interpolations, things like this should be caught on validation no? Not sure how this works however need to read more of the source...

nicolai86:
I'll close this for now due to inactivity. Feel free to re-open when you've made progress, @lennart .

bkleef:
@lennart did you made any progress? Looks like you were very close!

lennart:
@bkleef I tried using terraform for our infrastructure but it didn't take off yet ;) code worked but I didn't have the time to write tests and there still may be corner cases where it might fail.

PR #9263, opened 2016-10-06
Don't parse a TestFlagTypedKV value as a number

### LINKED ISSUE #7962: Variable should be type string, got number
In Terraform 0.7 I have observed an issue with handling numeric variables. To replicate this issue, you need to create two files, `main.tf`:

    variable "some_number" {
      type    = "string"
      default = 0
    }

And `terraform.tfvars.json`:

    {
      "some_number": 1
    }

Running `terraform apply` produces the following error:

    Errors:
    
      * 1 error(s) occurred:
    
    * variable some_number should be type string, got number

jbardin:
Don't try to parse a varibale as HCL if the value can be parse as a
single number. HCL will always attempt to convert the value to a number,
even if we later find the configured variable's type to be a string.

Fixes #7962

### DIFF
--- a/command/flag_kv.go
+++ b/command/flag_kv.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/hcl"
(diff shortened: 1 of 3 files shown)


mitchellh:
LGTM

Issue #9241, opened 2016-10-05
Terraform prints error 500 instead of showing actual error


bondyk:
Hi,

Terraform prints generic error message instead of the actual message from SoftLayer.
In my example underscore char is forbidden to be used in host name. In such case SoftLayer API returns response (code 500) with following payload:

```
<?xml version="1.0" encoding="iso-8859-1"?>
<methodResponse>
<fault>
 <value>
  <struct>
   <member>
    <name>faultCode</name>
    <value>
     <string>SoftLayer_Exception_Public</string>
    </value>
   </member>
   <member>
    <name>faultString</name>
    <value>
     <string>The hostname and domain must be alphanumeric strings that may be separated by periods '.'.  The only other allowable special character is the dash '-'  However the special characters '.' and '-' may not be consecutive.  Each alphanumeric string separated by a period is considered a label.  Labels must begin and end with an alphanumeric character and must be between 1-63 characters in length.  The last label, the TLD (top level domain) must be between 2-24 alphabetic characters. The TLD label and hostname cannot be solely comprised of digits. The domain portion must consist of least one label followed by a period '.' then ending with the TLD label.  Combining the hostname, followed by a period '.', followed by the domain gives the FQDN (fully qualified domain name), which may not exceed 253 characters in total length.</string>
    </value>
   </member>
  </struct>
 </value>
</fault>
</methodResponse>
```

If use terraform with the same input data then it prints generic error 500 message.
### Terraform Version

v0.7.5-dev
### Affected Resource(s)

-softlayer_virtual_guest
### Terraform Configuration

```
resource "softlayer_virtual_guest" "server1" {
    name = "server1_name"
    block_device_template_group_gid = "${var.block_device_template_group_gid}"
    domain = "${var.domain}"
    region = "${var.region}"
    hourly_billing = "${var.hourly_billing}"
    private_network_only = "${var.private_network_only}"
    public_network_speed = "${var.public_network_speed}"
    cpu = "${var.cpu}"
    ram = "${var.ram}"
    local_disk = "${var.local_disk}"
}
```
### Expected Behavior

Terraform should print actual error message instead of showing generic Error 500:

```
[...]

PR #9264, opened 2016-10-06
#9241 Show actual SoftLayer error instead of generic message

bondyk:
Fix of issue https://github.com/hashicorp/terraform/issues/9241

### DIFF
--- a/vendor/github.com/maximilien/softlayer-go/client/http_client.go
+++ b/vendor/github.com/maximilien/softlayer-go/client/http_client.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"text/template"
+	"github.com/maximilien/softlayer-go/common"
 )
 
 const NON_VERBOSE = "NON_VERBOSE"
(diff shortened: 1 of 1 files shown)


TheKevJames:
I don't think we want to be manually modifying vendor'd code; rather we should update [the vendor file](https://github.com/hashicorp/terraform/blob/master/vendor/vendor.json#L1295) to pull from the correct commit.

I'm not sure See [the terraform readme](https://github.com/hashicorp/terraform#updating-a-dependency) for more info and the process by which we do this.

bondyk:
Actually it was fixed there as well. See https://github.com/maximilien/softlayer-go/pull/136

TheKevJames:
Yup, I saw that! What I mean is that we shouldn't change this repo by applying the same patch as we do to a vendor library, rather we should update the pinned version of that library with `govendor` to a version that includes that patch already. Please see [the readme](https://github.com/hashicorp/terraform#updating-a-dependency) for more info.

TheKevJames:
As of today, maximilien/softlayer-go has been updated to [v0.7.0](https://github.com/maximilien/softlayer-go/releases/tag/v0.7.0) which includes the fix relevant to this PR. Let's pin [v0.7.0's SHA: 1d4217550](https://github.com/maximilien/softlayer-go/commit/1d4217550aef2c9207aeae5a93c7ce5c3c91ef21) with `govendor`.

teamterraform:
Hi folks! We are undertaking a general GH issue/PR cleanup. The [softlayer provider](https://github.com/terraform-providers/terraform-provider-softlayer) now has its own repository, and any PRs and issues should be opened there.

Issue #8384, opened 2016-08-22
outputting maps not working


rbuels:
### Terraform Version
0.7.0

### Affected Resource(s)
Please list the resources as a list, for example:
- output

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
output "test_map" {
    value = {
        foo = "bar"
    }
}
```

### Debug Output
https://gist.github.com/rbuels/2832e49a256199f857037fda3a2b58ed

### Expected Behavior
Terraform plan should at least have run without error

### Actual Behavior
Terraform plan emitted the error: `output test_map is not a valid type ([]map[string]interface {})`'

### Steps to Reproduce
1.  Make a file with a map output, as shown above.
2. Run terraform plan on it.

### References
Wasn't support for this supposed to be added by #5841?
https://github.com/hashicorp/terraform/pull/5841

rcha86:
Also seeing this problem. What is the correct syntax/formatting for this?

cornfeedhobo:
Same problem here

rcha86:
After digging around some more, I figured out how to create maps in output. You have to use the map function. See [here](https://www.terraform.io/docs/configuration/interpolation.html#map_key_value_)

```
output "test-map" { 
  value = "${map(
    "test-value-1", "map-value-1"
    "test-value-2", "map-value-2"
  )}"
}
```

However, terraform still does not support nested maps. [Open issue](https://github.com/hashicorp/terraform/issues/2114)

jbardin:
Closed via #9268

PR #9268, opened 2016-10-06
Get rid of the list when parsing HCL maps for vars

### LINKED ISSUE #8057: Not accepting maps from a var file!
Terraform Version: **v0.7.0**
Affected Resource: **Terraform Core**

###### terraform.tvfars
```hcl
somemap = {
  foo = "bar"
  baz = "qux"
}
```

###### main.tf
```hcl
variable "somemap" {
  type = "map"
}

output "somemap" {
  value = "${var.somemap}"
}
```

This should simply output the map. But instead, I get this error:
`variable somemap should be type map, got list`

It works as expected if I run `TF_VAR_somemap='{foo = "bar", baz = "qux"}' terraform apply`. Another interesting tidbit--if I change the var type to `variable "somemap" { type = "list" }`, I get the error: `variable somemap should be type list, got list`.

According to the docs [here](https://www.terraform.io/docs/configuration/variables.html), this should work.

jbardin:
When we parse a map from HCL, it's decoded into a list of maps because
HCL allows declaring a key multiple times to implicitly add it to a
list.  Since there's no terraform variable configuration that supports
this structure, we can always flatten a `[]map[string]interface{}` value
to a `map[string]interface{}` and remove any type errors trying to apply
that value.

Fixes #8057

### DIFF
--- a/command/flag_kv.go
+++ b/command/flag_kv.go
@@ -111,6 +111,10 @@ func loadKVFile(rawPath string) (map[string]interface{}, error) {
 				"Decoding errors are usually caused by an invalid format.",
 			err)
 	}
+	err = flattenMultiMaps(result)
+	if err != nil {
+		return nil, err
+	}
 
 	return result, nil
 }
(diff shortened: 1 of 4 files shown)


jbardin:
Added a commit that does the same check on the output side, which now fixes #8384

mitchellh:
LGTM but I'd like @jen20 to review this one as well so please poke at him to do so. :)

PR #9286, opened 2016-10-07
DOCS: Add reference about AWS IAM policy requirements

### LINKED ISSUE #9002: Terraform documentation lists IAM policy resources key as optional, but without it AWS rejects the document
### Terraform Version

> Terraform v0.7.4

### Affected Resource(s)

- aws_iam_policy data source

### Expected Behavior

Either Terraform should have created a valid aws_iam_policy without me specifying resources, because [the documentation](https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html) indicates that the parameter is optional (see below), or the documentation should be updated to reflect that this parameter should not be optional.

screen shot 2016-09-22 at 4 13 08 pm

### Actual Behavior

I got the following error:

```
MalformedPolicyDocument: Policy statement must contain resources.
```

### Steps to Reproduce

Create an IAM policy like so:

```
data "aws_iam_policy_document" "bad" {
    statement = {
        actions = [
            "cloudwatch:PutMetricData"
        ]
    }
}

resource "aws_iam_policy" "bad_policy" {
    name = "this-wont-work"
    description = "A policy that blows up."
    policy = "${data.aws_iam_policy_document.bad.json}"
}
```

And attempt to create this policy in AWS with Terraform.

nwalke:
Relevant to the current version of Terraform and website.

This adds a note in the `aws_iam_policy_document` documentation that
`resources` is required by AWS if used on an IAM policy.  Also added a
note on `aws_iam_policy` that `aws_iam_policy_document` is a good thing
to use when configuring.

Closes #9002

### DIFF
--- a/website/source/docs/providers/aws/d/iam_policy_document.html.markdown
+++ b/website/source/docs/providers/aws/d/iam_policy_document.html.markdown
@@ -88,7 +88,7 @@ each accept the following arguments:
   apply to. Used to apply a policy statement to all actions *except* those
   listed.
 * `resources` (Optional) - A list of resource ARNs that this statement applies
-  to.
+  to. This is required by AWS if used for an IAM policy.
 * `not_resources` (Optional) - A list of resource ARNs that this statement
   does *not* apply to. Used to apply a policy statement to all resources
   *except* those listed.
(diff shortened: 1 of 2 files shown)


stack72:
Thanks for the work here @nwalke :)

Issue #9279, opened 2016-10-07
provider/aws: Add Ability To Enable / Disable ALB AccessLogs


stack72:
As per #8438 but for ALB

jvasallo:
I created a PR for this, feel free to review whenever! :)

https://github.com/hashicorp/terraform/pull/9290

stack72:
Closed via #9290

PR #9290, opened 2016-10-08
provider/aws: Add Ability To Enable / Disable ALB AccessLogs

jvasallo:
Hey all,

This is my first major swing at committing to Terraform, saw the issue below and I think this fixes it. Going to do some more testing on my end, but please provide feedback?

Related to #9279

### DIFF
--- a/builtin/providers/aws/resource_aws_alb.go
+++ b/builtin/providers/aws/resource_aws_alb.go
@@ -88,6 +88,11 @@ func resourceAwsAlb() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+						"enabled": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  true,
+						},
 					},
 				},
 			},
@@ -276,7 +281,7 @@ func resourceAwsAlbUpdate(d *schema.ResourceData, meta interface{}) error {
 			attributes = append(attributes,
 				&elbv2.LoadBalancerAttribute{
 					Key:   aws.String("access_logs.s3.enabled"),
-					Value: aws.String("true"),
+					Value: aws.String(log["enabled"].(bool)),
 				},
 				&elbv2.LoadBalancerAttribute{
 					Key:   aws.String("access_logs.s3.bucket"),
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_alb.go
```diff
@@ -292,7 +297,7 @@ func resourceAwsAlbUpdate(d *schema.ResourceData, meta interface{}) error {
 		} else if len(logs) == 0 {
 			attributes = append(attributes, &elbv2.LoadBalancerAttribute{
 				Key:   aws.String("access_logs.s3.enabled"),
-				Value: aws.String("false"),
+				Value: aws.String(strconv.FormatBool(false)),
```
stack72: why change this to the formatting of a bool?

jvasallo: Good catch, I think I was initially going to convert everything to be a `aws.Bool` until I looked at the AWS SDK and noticed they wanted String types. Since a user will never be setting this, i'll go ahead and change this back!

jvasallo:
I have finished the code and I have tested it out in my local TF/AWS environment. It works for me! 

I did have a question about the default value for `access_logs.s3.enabled` that stems from `resource_aws_elb.go`. Is there a reason it was set to `true` by default? I figured it would make more sense for this to be false by default since not everyone uses access_logs in AWS.

Working on updating the docs now.

jvasallo:
Added docs, and I just thought about it a bit more. By specifying `access_logs` you are essentially wanting this to be true. If you want to disable them  in the future, that is where `enabled` comes into play.

jen20:
Hi @jvasallo! Could we accomplish this without an explicit `enabled` and detect removal instead?

jvasallo:
@jen20 do you mean if I were to remove the `access_logs` block? I think that functionality is already built in before this request? Correct me if I am wrong though! :)

Issue #9279 I think was asking for an `enabled` type functionality. I could see it being useful when you want to have a bucket on standby, but not always writing to it for cost reasons.

jvasallo:
Anything holding this up? Just want to know before I jump onto the next issue! Thanks!

stack72:
Hi @jvasallo 

Thanks for the work here - I have just pulled and merged this locally. I had to make 1 small change to it to allow the tests to work as expected - you can see the change here b3a0145d8c8082f1479a6fbfa5f3a9d91b4393f6

Thanks for the work here

Paul

PR #9292, opened 2016-10-08
Update reflectwalk and copystructure

mitchellh:
This includes a number of changes that @jbardin and I have worked on in
order to improve the robustness of both of these libraries. There are
some state copy cases currently that can trigger failures that are fixed
here, and future changes in branches coming into Terraform require these
changes.

Throughout this process we've vetted the Terraform tests to make sure
that everything still passes and in the branches the new behavior works
as well.


mitchellh:
Just kidding @jbardin already did this.

PR #9294, opened 2016-10-08
provider/scaleway: improve first setup experience

nicolai86:
Assuming you're using the scaleway CLI already, we can improve the dev XP for scaleway & terraform a little by not forcing people to export their credentials to the environment, or hard code them for that matter.

the [scaleway CLI](https://github.com/scaleway/scaleway-cli#scw-login) stores all credentials in
`$HOME/.scwrc`. This file is JSON encoded and holds the necessary information
at `token` and `organization`.

Note that the scaleway CLI does not support multi-account setups, which is why
this is a viable way to get these information, besides the environment of course.

### DIFF
--- a/builtin/providers/scaleway/provider.go
+++ b/builtin/providers/scaleway/provider.go
@@ -1,9 +1,47 @@
 package scaleway
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+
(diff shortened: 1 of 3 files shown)


nicolai86:
I'll close this PR until multi-account support has landed on the scaleway SDK side.

PR #9302, opened 2016-10-09
No, for AWS Route53 zones the ID is not the domain

chbrown:
For AWS Route53 zones the ID is not the domain name, it is the Zone ID.

When I took your "If you're unsure, feel free to just try an ID." advice to heart I figured it out pretty quickly, but still, getting it right the first time would be nice.


chbrown:
Oh, turns out the correct documentation is at the bottom of <https://www.terraform.io/docs/providers/aws/r/route53_zone.html> so `Z1D633PJN98FT9` is probably a better example ID.

PR #9199, opened 2016-10-03
provider/azurerm: Load Balancer resources

stack72:
* [x] `azurerm_lb`
* [x] `azurerm_lb_backend_address_pool`
* [x] `azurerm_lb_rule`
* [x] `azurerm_lb_nat_rule` 
* [x] `azurerm_lb_probe`
* [x] `azurerm_lb_nat_pool`

Current test status:

```
% make testacc TEST=./builtin/providers/azurerm TESTARGS='-run=TestAccAzureRMLoadbalancer'                           ✹
==> Checking that code complies with gofmt requirements...
go generate $(go list ./... | grep -v /terraform/vendor/)
2016/10/05 12:52:28 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMLoadbalancer -timeout 120m
=== RUN   TestAccAzureRMLoadbalancerBackEndAddressPool_basic
--- PASS: TestAccAzureRMLoadbalancerBackEndAddressPool_basic (207.26s)
=== RUN   TestAccAzureRMLoadbalancerBackEndAddressPool_removal
--- PASS: TestAccAzureRMLoadbalancerBackEndAddressPool_removal (165.89s)
=== RUN   TestAccAzureRMLoadbalancerNatRule_basic
--- PASS: TestAccAzureRMLoadbalancerNatRule_basic (179.30s)
=== RUN   TestAccAzureRMLoadbalancerNatRule_removal
--- PASS: TestAccAzureRMLoadbalancerNatRule_removal (180.73s)
=== RUN   TestAccAzureRMLoadbalancerRule_basic
--- PASS: TestAccAzureRMLoadbalancerRule_basic (170.40s)
=== RUN   TestAccAzureRMLoadbalancerRule_removal
--- PASS: TestAccAzureRMLoadbalancerRule_removal (204.23s)
=== RUN   TestAccAzureRMLoadbalancer_basic
--- PASS: TestAccAzureRMLoadbalancer_basic (136.03s)
=== RUN   TestAccAzureRMLoadbalancer_frontEndConfig
--- PASS: TestAccAzureRMLoadbalancer_frontEndConfig (214.47s)
=== RUN   TestAccAzureRMLoadbalancer_tags
--- PASS: TestAccAzureRMLoadbalancer_tags (215.52s)
=== RUN   TestAccAzureRMLoadbalancerProbe_basic
[...]

PR #9307, opened 2016-10-10
provider/azurerm: fix computed tags causing non-empty plan

pmcatominey:
The tag schema was changed in #9199, setting the Computed flag, this was causing
the plan to not be empty for resources which support tags but none were set, as
no value would be set by flattenAndSetTags.

Setting an empty map instead fixes the issue, ran original failing test and an
update tags test to ensure nothing else was broken.

Depends on #9305.

```
TF_ACC=1 go test ./builtin/providers/azurerm -v -run TestAccAzureRMCdnProfile -timeout 120m
=== RUN   TestAccAzureRMCdnProfile_importWithTags
--- PASS: TestAccAzureRMCdnProfile_importWithTags (171.64s)
=== RUN   TestAccAzureRMCdnProfile_basic
--- PASS: TestAccAzureRMCdnProfile_basic (162.70s)
=== RUN   TestAccAzureRMCdnProfile_withTags
--- PASS: TestAccAzureRMCdnProfile_withTags (203.12s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/azurerm	537.538s
```

### DIFF
--- a/builtin/providers/azurerm/tags.go
+++ b/builtin/providers/azurerm/tags.go
@@ -64,6 +64,7 @@ func expandTags(tagsMap map[string]interface{}) *map[string]*string {
 
 func flattenAndSetTags(d *schema.ResourceData, tagsMap *map[string]*string) {
 	if tagsMap == nil {
+		d.Set("tags", make(map[string]interface{}))
 		return
 	}
 


pmcatominey:
Theres no unit test covering `flattenAndSetTags`, the reason for which appears to be that we can't get a working instance of ResourceData from outside the schema package.

This could be solved by changing all usages of this function to remove the dependency of that argument, from:

```go
flattenAndSetTags(d, resp.Tags)
```
To:
```go
d.Set("tags", flattenTags(resp.Tags))
```

stack72:
LGTM thanks :)

PR #9313, opened 2016-10-10
Cleanup main help message to look cleaner

iamthemuffinman:
This brings the main help more in line with how commands are done. Plus, it looks much cleaner than continually writing to a buffer.


iamthemuffinman:
@jen20 Thanks!

PR #9314, opened 2016-10-10
provider/aws: Add extra logging on AuthFailure

jen20:
This pull request is intended to add a temporary control to Terraform to output more verbose logging in the case of an AuthFailure error being returned by the AWS API.

cc @phinze, @sean-


catsby:
LGTM

Issue #8680, opened 2016-09-06
terraform prompts several times to enter a provider.aws.region when a certain number of nested modules is reached.


martin-flaregames:
**EDIT**: in order to reproduce this issue check out https://github.com/martin-flaregames/terraform-aws-region-prompt and follow the instructions provided in README.md

**EDIT**: (!) Please notice that the prompt for `provider.aws.region` will only occur if your `~./aws` credential configuration is missing the entry for the default region. If not, Terraform may silently just read and use those credentials instead of the ones declared in the AWS provider.

Terraform is prompting me to manually specify the `provider.aws.region` although it is already configured and properly working. This occurs when my projects uses more or less deep or complex module structures, like: module uses module uses module, etc.

If I do enter the region then there is a second problem, the provider fails with an error telling me that a specific VPC id cannot be found, although it actually exists and was created by the provider itself some seconds ago.

Funnily the problem disappears if I reduce the number of nested modules. See below.
### Terraform Version

`Terraform v0.7.2`
`Terraform v0.7.3`
`Terraform v0.7.4`
`Terraform v0.7.5`
`Terraform v0.7.6`
### Affected Resource(s)
- provider aws?
### Expected Behavior

The provider should be configured once and work in all childs
### Actual Behavior

It fails after some few nested modules including other modules
### Steps to Reproduce
1. `rm -rf .terraform/modules`
2. `terraform get`
3. `terraform plan`
4. Here I get the prompt to enter the region names, if I do then the plan works allright
5. `terraform apply`
6. Again the prompt to enter the region names, if I do then the terraform starts applying
7. see terraform tell me that a VPC id cannot be found 
### Steps to Reproduce (detailed)

The relevant part of the project folder structure looks like this

``` hcl
/
    security/
        group/
        ingress/
        allow/
        realm/
    opsworks/
        stack/
        app/
        layer/
            mongo/
    main/
```

The main test module is located in `/main/` and the nested module calls occur in this order

from `main/`
- `module "mongolayer" { source = "../opsworks/layer/mongo" }`

from `opsworks/layer/mongo/`
- `module "realm" { source = "../../../security/realm" }`

from `security/realm/`
- `module "securitygroup" { source = "../group" }`
- `module "allow" { source = "../allow" }`
- `module "allowself" { source = "../allow" }`

_the number of times the region input is requested is in accordance with the number of modules used in the "realm" module._

[...]

PR #9318, opened 2016-10-11
terraform: orphaned grandchild module inherits provider config

mitchellh:
This fixes an issue where orphaned grandchild modules don't properly
inherit their provider configurations from grandparents. I found this
while working on shadow graphs (the shadow graph actually caught an
inconsistency between runs and exposed this bug!), so I'm unsure if this
affects any pre-existing issues.

To better explain the issue, I'll diagram things.

Here is a hierarchy that _works_ (w/o this PR):

```
root
|-- child1 (orphan)
|-- child2
    |-- grandchild
```

All modules in this case will successfully inherit provider
configurations from "root".

Here is a hierarchy that _doesn't work without this PR_:

```
root
|-- child1 (orphan)
    |-- grandchild (orphan)
```

In this case, `child1` does successfully inherit the provider from root,
but `grandchild` _will not_ unless `child1` had resources. If `child1`
has no resources, it wouldn't inherit anything. This PR fixes that.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -1390,6 +1390,54 @@ func TestContext2Apply_moduleOrphanProvider(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_moduleOrphanGrandchildProvider(t *testing.T) {
+	m := testModule(t, "apply-module-orphan-provider-inherit")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+
+	p.ConfigureFn = func(c *ResourceConfig) error {
+		if _, ok := c.Get("value"); !ok {
+			return fmt.Errorf("value is not found")
+		}
+
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


jbardin:
You mentioned that this only "periodically" caught the inconsistency. Aren't the transform walk done without any concurrency, via a ReverseDepthFirstWalk?  Do you have any idea why the output wasn't deterministic?

mitchellh:
Yep, because it didn't create a dependece so it'd walk the two providers in paralllel. 

The transforms are sequential but the race was in the graph walk (since the edge wasn't made).

martin-flaregames:
I think this is related, see ticket [8680](https://github.com/hashicorp/terraform/issues/8680). There is an example in GitHub which is still failing: [terraform-aws-region-prompt](https://github.com/martin-flaregames/terraform-aws-region-prompt). In this example child modules are not properly inheriting the AWS provider, instead terraform tries to use my default profile in `~./aws` (!). Running terraform 0.7.6 under Linux 4.7.6-1-ARCH  64 bits (ArchLinux).

Issue #9235, opened 2016-10-05
scaleway provider


webroboteu:
Hello.

to access the the scaleway apis now ewe need to send the user token and the organization id and not the access token.
I'm doing something wrong or is in fact the new scenario

Roger

nicolai86:
Hey @webroboteu,

you need two things for the provider to work:

- the user access token. Once logged in to the scaleway UI you can manage access tokens in your settings under credentials: https://cloud.scaleway.com/#/credentials
- your organization ID. Since this is not listed on the scaleway UI, the most reliable way is to use the [scaleway CLI](https://github.com/scaleway/scaleway-cli). Once you login via the CLI (`scw login`) your organization ID is written to `~/.scwrc`. E.g. `cat ~/.scwrc | jq -r '.organization'`

I guess we should add this information to the provider page.

far-blue:
I've just started playing with the scaleway provider and found the following that might help you:
* The `organisation` value needed by the provider is now listed in the account credentials section (https://cloud.scaleway.com/#/credentials) as `access key` just above the table of access tokens.
* The `access_key` value needed by the provider is actually a token so just create a new token in the credentials section of your account and give it a helpful label for future reference.
* Server image values are not the same as the name given to the image when you create a server via the web control panel. I've yet to find a way to show the image id values via the web interface so I had to resort to using the Rest API. For this I used Postman (a great Rest API client), set a header of X-Auth-Token to the token I mentioned in the previous bullet point, set a header of Content-Type to `application/json` and made a GET request to https://api.scaleway.com/images which listed all the images available to me. Hunt around in there for the image you want - noting the architecture is correct - and then use the `id` value in the `image` key for the server.

stack72:
Closed via #9321

PR #9321, opened 2016-10-11
provider/scaleway: add missing information

Zhebr:
It's important to note this env var needs to be organization ID, at least for individual users (organization name doesn't work).

### DIFF
--- a/website/source/docs/providers/scaleway/index.html.markdown
+++ b/website/source/docs/providers/scaleway/index.html.markdown
@@ -91,6 +91,6 @@ provider "scaleway" {
 
 ...and instead set these environment variables:
 
-- **SCALEWAY_ORGANIZATION**: Your Scaleway organization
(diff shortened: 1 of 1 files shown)


nicolai86:
Hey @Zhebr 

thanks for the clarification!

What do you think about adding a hint about where to find it in the UI? (Credentials > Tokens > Access Key) ? I think this would help as well as different part of Scaleway call this differently (scaleway cli: organization, scaleway ui: access key)

Zhebr:
@nicolai86 I think it'll help. `access_key` name is misleading and the information is not exposed enough in Scaleway panel (especially with link about `access_token` generation). I'll add it in docs.

nicolai86:
Great, thank you. This should also solve getting started issues as discussed in #9235

stack72:
Thanks for the work here @Zhebr and the review @nicolai86 :)

LGTM!

P.

PR #9334, opened 2016-10-12
terraform: Shadow Graph

mitchellh:
**EDIT: Overview video here (in addition to below): https://www.youtube.com/watch?v=WXJRh--vw-U&feature=youtu.be**

This PR adds an internal (minimally user-facing) feature called the **shadow graph**.

The purpose of the shadow graph is to ensure future stability of Terraform by allowing us to test new core graph changes safely in a production environment. This is to support the upcoming operation-specific graph builder rewrites for Terraform 0.8.

The inspiration for the shadow graph is _shadow traffic_ in web applications: the act of mirroring traffic to non-user-facing versions of an application to ensure the same behavior comes back. 

Likewise, the shadow graph in Terraform records and duplicates the state of the world and re-runs Terraform to ensure identical behavior given the same state of the world. The purpose of this is to both 1.) ensure correct behavior of the **active graph** as well as 2.) ensure identical behavior of any **experimental graph**.

## This PR

Design and implementation follows in sections below and should be read but I also want to talk about what this PR does and doesn’t do:

* Implements the shadow graph

* Enables the shadow graph to verify the current graph works. This will surface any issues in tests.

* **Does not** expose shadow errors to the user in any way yet. Therefore, any errors will be suppressed for now. 

* Has complete (100%) unit test coverage on all behaviors described below, including more complex ones such as "interrupts abort the shadow graph and don't report errors."

The goal of this PR is to get some of these massive changes into master as soon as possible and get any core changes running against the shadow graphs as well to find any bugs (in Terraform itself or in the shadow graph implementation). 

Likely in a following patch release (Z in X.Y.Z) we’ll start enabling shadow graph error showing to users once we agree on a UX for that. This will let us test features for 0.8 without affecting real production state. 

## Design

The shadow graph was designed with the following decisions:

1. The shadow graph must **never** alter the behavior of the real run.

1. The shadow graph must **never** panic, and therefore cause potentially data loss from the real run. All errors must be checked and stored safely.

1. The shadow graph must **never** modify real state. All values must be deeply copied. 

1. The shadow graph **should** reuse as many components from the real run as possible in order to mimic the behavior as much as possible. 

## Implementation

The shadow graph was implemented and meets all above design goals. 

### Terminology

New terminology introduced by this feature:

* **Actual** - The implementation that would be run if there was no shadow graph. The implementation that affects real world state, has side effects.

* **Real, Real Side** - The implementation that records parameters/results for a shadow to compare to by calling the actual implementation. When a shadow is used, the real side is run rather than the actual implementation.

* **Shadow, Shadow Side** - The implementation that verifies parameters/results with the real side, but must never modify actual state or call the actual implementation in any way.

### “helper/shadow”

`helper/shadow` is a new helper library introduced that contains structs for safely coordinating setting and reading values concurrently for the use of shadowing behavior.

The idea is that any shared data is represented by a `shadow.Value`. The real side _sets_ values while the shadow side _reads_ values, blocking if the value is not yet available.

The blocking behavior makes it easy to implement a shadow since coordination with the real side is minimal: just read a value that the real side sets, and if it isn’t set it will block.

An important feature is **closing shadow values**: when a value is closed, it will return `ErrClosed` instead of blocking for any unknown value. This can be caught by the shadow and recorded to avoid waiting on a value that will never appear. Closing happens when the real side completes running, thus notifying the shadow that no further new data will appear.

### `terraform.Shadow` (Interface)

A new interface is in the `terraform` package: `Shadow`. This interface is implemented by all “shadows” and encourages a standard API + workflow for working with shadows:

1. `CloseShadow` is called when the _real_ side is complete. This tells the shadow that it shouldn’t block/wait for any real behavior, and should error if unknown calls are made. This is not `Close` because some implementations must actually implement `io.Closer` for other reasons.

1. `ShadowError` returns the actual errors accumulated during the run. These are stored separately rather than being returned during the calls themselves (i.e. `ResourceProvider.Apply`) to help ensure identical behavior: no shadow-specific errors are returned to the graph.

### ResourceProvider, ResourceProvisioner

New implementations of the `ResourceProvider` and `ResourceProvisioner` were made in order to use the `helper/shadow` library to mirror calls. This wraps existing real implementations.

In both cases, a shadow is initialized via a constructor that takes the actual value. This returns both a real and shadow implementation. Both implement the desired interface (such as ResourceProvider), with the shadow side also implementing the `Shadow` interface (covered above). 

The real value records all inputs and outputs from the actual implementation as its called, and uses the “helper/shadow” library to store deep copies of these so that when the identical function is called on the shadow side, parameters can be compared and identical results returned.

The real value is used with the real run. After it completes, the shadow is “closed” to tell it no new data will come, and the shadow can be run.

Once that completes, errors can be checked for inconsistencies. 

### Shadow Context Creation

A function `newShadowContext` was made that takes an actual context and handles the data duplication and initialization of a shadow side. 

This returns a `*Context`, so we’re reusing the identical logic. However the state of the context is fully identical and yet duplicated from the actual side. 

This also returns a `terraform.Shadow` that is used to close any graph runs and check for errors. This shadow compares the final end state/diff as well to ensure the final state is identical.

**As a benefit:** no new `EvalContext` implementation, `GraphBuilder`, `GraphWalker`, etc. is required. We’re able to reuse the _actual_ code used for actual runs! 

### Shadow Walk

When a graph walk is requested with `terraform.*Context.walk()`, it now takes a new parameter for the shadow graph. The behavior is as follows:

1. If there is a shadow graph, initialize a shadow context.

1. Walk the real graph

1. Walk the shadow graph (after the real graph completes)

1. Compare the end state, record any shadow errors. These must be retrieved by `ShadowError` so that Apply, Refresh, so on don’t return shadow errors to the user.

Some specific implementation notes:

* If Terraform is interrupted (i.e. Ctrl-C), then the shadow graph is aborted and any errors are ignored. Interrupts can’t be easily or safely mirrored on the shadow graph because graph timings are too complicated to match.

* Shadow errors must be retrieved manually via `ShadowError`. This is so that `terraform apply` will show a success even if the shadow graph recorded inconsistencies. We don’t want users thinking that shadow errors are real Terraform bugs or failures.

* The `walk()` function can take any `*Graph`, so it can be an identical graph (to verify identical behavior) or a completely new graph to verify the same end state. This will be used to both test the current graph is sane as well as test experimental new features. 

### Unit Tests

Modifications were made to `TestMain` to enable special behavior during unit tests: any shadow errors cause the test to fail.

This allows our existing massive Context test base against to Apply, Refresh, Plan, etc. to test both the actual implementations as well as the shadow logic. 

Any failures are raised via a **panic** rather than an error. This is unfortunately necessary since some tests expect an error to be returned, therefore masking any shadow failures. If an additional API was made to test shadow errors, we’d have to modify the hundreds of tests we have to call it. Instead, I chose to panic immediately. 

And this is a major part of this PR: by putting this in master, our CI and any future core fixes immediately also stress the new shadow implementation. 

## Next Steps

Following this PR, the following still needs to be done:

* Agree on a UX to order to surface the shadow errors to the end user in a way that conveys that Terraform itself is working fine but some experimental feature failed and should be reported. 

* Begin merging new graph builder work (i.e. the f-apply-builder branch) to use the shadow logic and release this into a patch release to test the new logic before promoting it to “production”.


mitchellh:
We did a video overview here: https://www.youtube.com/watch?v=WXJRh--vw-U&feature=youtu.be

terraform/diff.go
```diff
@@ -279,6 +311,10 @@ func (d *ModuleDiff) String() string {
 	return buf.String()
 }
 
+func (d *ModuleDiff) GoString() string {
```
jbardin: I don't think we need these GoString methods.
It's only replacing the `&` in the standard GoString format with a `*`, but causes vet to fail if we need to add a sync.* field at some point. I think it would be better to be consistent with the other structures that will have the `&` prepended with the default format. (I've already removed/fixed the ones that caused failures, but can clean them all up at some point)

mitchellh: Ah yeah, that is a real sadness. Perhaps we just use copystructure in GoString as well to get a non-pointer value. Realistically it should only happen for debugging... I'll remove it for now.

svanharmelen:
@mitchellh I must admit this is a really cool feature! Nice work!

But it also raises a question that I would like to ask... I was wondering if this is a `always on` feature? After a quick browse through the code it seems this new functionality is added is such a way that it is always enabled and so will always walk shadow graphs along side of the real graphs that are walked.

Did I read/understand that correctly? As I'm wondering how this will impact performance (CPU and memory usage). I would guess (but maybe I have to read this PR a little closer when I have a little more time) the shadow graphs take just about the same CPU and memory as the real graph does, which would mean that this will definitely impact performance if you run very large TF plans (which we have a lot).

Curious what your take is on this and if you see any opening in making the behaviour optional (maybe by a cmdline switch, but since we use TF as a library other solutions would be fine too)?

mitchellh:
@svanharmelen This is super true, actually there is a PR right now to make it disable-able...

You're right, this should take about the same amount of memory and CPU as a real graph run. They're done in sequence so it shouldn't balloon memory too much (though its certainly more since its storing a copy of the results).

I'd be really curious if you have the chance to try it on your giant plans and see how it impacts it. The general plan however is to make this disable-able. But perhaps we could also add a node limit to it (if your graph is over some arbitrary number of nodes then just... don't.)

svanharmelen:
@mitchellh ah, that is good news (that it can be disabled in the near future). I didn't get to run master to any of our bigger plans yet, but will see if I can set something up.

As for your suggestion about a node limit; nodes are not be our biggest problem. We have some custom resources that have sets (schema.Set) with 18 fields (some of which are again (smaller) sets) that contain 500 to a 1000 items. And that times a few (10-20 atm) for a plan makes it pretty heavy on the server it runs on.

Other plans do have higher node counts, but are much simpler (standard) resources and they are still not superbig I think. But will see if I can get around to doing some testing...

PR #9345, opened 2016-10-12
Update crypto/ssh to support ED25519 SSH keys

### LINKED ISSUE #5156: unsupported key type "OPENSSH PRIVATE KEY"
Remote execution fails if you have a "new-format private key". This means you're not able to protect your private key with a KDF, which is a Good Thing to do.

Output received:

```
Error applying plan:

1 error(s) occurred:

* Failed to parse key file "[XXX PRIVATE KEY DATA XXX]": ssh: unsupported key type "OPENSSH PRIVATE KEY"
```

The OpenSSH man page section about KDF support:

```
     -a rounds
             When saving a new-format private key (i.e. an ed25519 key or any
             SSH protocol 2 key when the -o flag is set), this option speci-
             fies the number of KDF (key derivation function) rounds used.
             Higher numbers result in slower passphrase verification and
             increased resistance to brute-force password cracking (should the
             keys be stolen).
```

passcod:
ED25519 key support was added in golang/crypto@1e61df8. This is
needed to parse and support "new format" SSH keys prefixed with:

    -----BEGIN OPENSSH PRIVATE KEY-----

Fixes #5156


jen20:
Hi @passcod! Thanks for opening a pull request. I agree this does need to be done - however I think there is an issue with the way `govendor` has been used to generate the diff - we have multiple commit SHAs for the same git repository. Could you take a look at rectifying this prior to merging? Thanks!

passcod:
Ah, originally I'd done a

    govendor fetch golang.org/x/crypto

but that vendored the entire crypto repo, which, looking at the existing vendor.json, I thought you were trying to avoid (you're only pulling the SSH subpath). So instead I ran

    govendor fetch golang.org/x/crypto/ssh

which updated the ssh subpath and then pulled in the ed25519 subpaths along because they're dependencies I guess.

Should I remove crypto/ssh and fetch in the entire crypto library?

stack72:
Hi @passcod 

Thanks for the work here - I have just updated the dependency in #9661

Thanks

Paul

passcod:
Thanks

PR #9347, opened 2016-10-12
[WIP] aws/provider: OpsWorks instance registration (and layer assignments)

hydroxide:
Unfortunately, AWS OpsWorks instances don't support all of the features that regular EC2 instances do. Therefore it is sometimes necessary to first create regular EC2 instances and then register them to an OpsWorks stack.

For this I have created a new "aws_opsworks_registration" resource. The problem, however, is that instances cannot be registered to an OpsWorks stack using just the SDK. The "aws opsworks register" command needs to be run locally from the instance. I would very much appreciate **any insight or suggestions** regarding building this logic into my resource (somehow referencing the logic for remote-exec provisioners - I have dug around the code a bit but didn't see any clear way to do this).

The following is a working example of using this resource. I would like to encapsulate the provisioner within the resource.

```
provider "aws" {
    region = "us-east-1"
}

resource "aws_instance" "foobaz" {
    ami = "ami-c481fad3"
    instance_type = "t2.micro"

    provisioner "remote-exec" {
        inline = ["aws opsworks register --infrastructure-class ec2 --stack-id ${var.opsworks_stack_id} --local --region us-east-1 && sleep 200"] // Sleep to make sure registration and setup complete..

        connection {
            ...
        }
    }
}

resource "aws_opsworks_registration" "foobaz" {
    instance_id = "${aws_instance.foobaz.id}"
    stack_id = "${var.opsworks_stack_id}"
    layer_ids = ["${var.opsworks_layer_id}"]
}
```

### DIFF
--- a/builtin/providers/aws/resource_aws_directory_service_directory.go
+++ b/builtin/providers/aws/resource_aws_directory_service_directory.go
@@ -336,7 +336,7 @@ func resourceAwsDirectoryServiceDirectoryCreate(d *schema.ResourceData, meta int
 				d.Id(), *ds.Stage)
 			return ds, *ds.Stage, nil
 		},
-		Timeout: 30 * time.Minute,
+		Timeout: 45 * time.Minute,
 	}
 	if _, err := stateConf.WaitForState(); err != nil {
 		return fmt.Errorf(


harveyxia:
Was there any follow up discussion for this feature?

PR #9356, opened 2016-10-13
Filter nil Deposed values during State init

### LINKED ISSUE #9337: crash with remote state file
In one of our projects the remote state file (stored in s3) was renamed. The remote config command was re-run to point to the new tfstate file but now every terraform plan results in the following crash

https://gist.github.com/liamjbennett/21a67fe4d260c485767307ca419ef292

TF version: v0.7.5

jbardin:
The Deposed slice wasn't being normalized and nil values could be read
in from a state file. Filter out the nils during init. There is
still a bug in copystructure, but that will be addressed separately.

Fixes #9337

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -1403,9 +1403,18 @@ func (s *ResourceState) init() {
 		s.Deposed = make([]*InstanceState, 0)
 	}
 
-	for _, dep := range s.Deposed {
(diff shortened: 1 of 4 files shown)


jbardin:
I'll merge this once a copystructure fix that get rid of the panic lands.

PR #9244, opened 2016-10-05
[WIP] provider/aws: Add new aws_vpc_endpoint_route_table_association resource.

kwilczynski:
This commit adds a new resource which allows to a list of route tables to be
either added and/or removed from an existing VPC Endpoint. This resource would
also be complimentary to the existing `aws_vpc_endpoint` resource where the
route tables might not be specified (not a requirement for a VPC Endpoint to
be created successfully) during creation, especially where the workflow is
such where the route tables are not immediately known.

Signed-off-by: Krzysztof Wilczynski krzysztof.wilczynski@linux.com

PR #9357, opened 2016-10-13
Allow empty route_table_ids list in aws_vpc_endpoint resources

mrwacky42:
Without this change, TF submits a request with empty RouteTableIDs and AWS responds `400 bad request`.
With this change, I can create a VPC Endpoint without a route table.

I am almost completely ignorant of Go, so let me know if there's a better to solve this.

### DIFF
--- a/builtin/providers/aws/resource_aws_vpc_endpoint.go
+++ b/builtin/providers/aws/resource_aws_vpc_endpoint.go
@@ -59,9 +59,15 @@ func resourceAwsVpcEndpoint() *schema.Resource {
 func resourceAwsVPCEndpointCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ec2conn
 	input := &ec2.CreateVpcEndpointInput{
-		VpcId:         aws.String(d.Get("vpc_id").(string)),
-		RouteTableIds: expandStringList(d.Get("route_table_ids").(*schema.Set).List()),
-		ServiceName:   aws.String(d.Get("service_name").(string)),
+		VpcId:       aws.String(d.Get("vpc_id").(string)),
+		ServiceName: aws.String(d.Get("service_name").(string)),
+	}
+
+	if v, ok := d.GetOk("route_table_ids"); ok {
+		list := v.(*schema.Set).List()
+		if len(list) > 0 {
+			input.RouteTableIds = expandStringList(list)
+		}
 	}
 
 	if v, ok := d.GetOk("policy"); ok {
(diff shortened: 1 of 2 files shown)


mrwacky42:
Of course, now I discovered https://github.com/hashicorp/terraform/issues/8225, so ...

catsby:
This looks good to me, @mrwacky42 , what is your concern / question with regards to #8225?

mrwacky42:
@catsby  -- My concern is that this is only partially useful, since I can't actually add RouteTableID to the VPC endpoint after the fact. I had to create my VPC endpoint with non-empty `route_table_ids`, since there is currently no way to add them after the fact with Terraform.

I think Terraform will need a `aws_vpce_endpoint_route_table_association` resource to fully fix #8225.

mrwacky42:
@catsby *ping*

catsby:
> I can't actually add RouteTableID to the VPC endpoint after the fact. I had to create my VPC endpoint with non-empty route_table_ids, since there is currently no way to add them after the fact with Terraform.

Sorry for my confusion; the resource allows you to add/modify the RouteTable IDs:

- https://www.terraform.io/docs/providers/aws/r/vpc_endpoint.html

What am I missing? 

This PR as it stands seems useful and I'd like to pull it in, unless you object

mrwacky42:
Yes, this is good to merge.

What you are missing is that I'm trying to create the `vpc_endpoint` resource without specifying `route_table_ids`, and associate it to route tables in another step. For better or worse, I don't have the `route_table_ids` available to me when I'm creating this endpoint (different modules essentially).

The original task I was trying to achieve is:
* Create `vpc_endpoint` resource with empty `route_table_ids` (fixed by this PR)
* Create `route_table`s, and add the `vpc_endpoint` prefix list as a route at a later time.

But... the `aws_route_table` resource requires CIDR blocks presently, and barfs if I give it `pl-c0ffee1234` as a `destination_cidr_block`.

From AWS web console, the only way I am able to add the route is from the VPC endpoints page, and not on the route tables themselves. In this, the web console seems to have the same limitation as Terraform -- only accepting CIDR blocks.

So that's why I'm suggesting `aws_vpc_endpoint_route_table_association` as a complete fix for #8225.

Hopefully I'm being clear enough, and yes please merge!

catsby:
Thank you for the explanation! I'll go ahead and merge this then

coen-hyde:
Linking #9244

mrwacky42:
@coen-hyde - I hadn't. Thanks.

PR #9360, opened 2016-10-13
What should be specified is not obvious for Azure Network Security Groups

jcrowthe:
Specified that the security group needs to be addressed by it's id and not by its name. This will assist when debugging the error with message "{name} is invalid. Expect fully qualified resource Id that start with azure .properties.networkSecurityGroup.id"


stack72:
Hi @jcrowthe 

Thanks for the PR here - you are indeed correct :) This LGTM1

P.

Issue #9352, opened 2016-10-13
providers/consul: support setting ID on consul_service


dmportella:
Hi there,

Right now the resource `consul_service` *does not* support the resource count features very well because since the consul record identifier is created from the name you give to the record ([Reference](https://www.terraform.io/docs/providers/consul/r/service.html#id)) you can't create record with the same `name` but different `identifiers` which allows you to use `consul` dns load balancer feature.

### Terraform Version
> Terraform v0.7.5

### Affected Resource(s)
- consul_service

### Terraform Configuration Files
```
# Docker provider
provider "docker" {
	host = "unix:///var/run/docker.sock"
}

# Consul Provider
provider "consul" {
	address	= "http://localhost:8500"
	datacenter = "dc1"
	scheme	 = "http"
}

variable "elastic_count" {
	type = "string"
	default = "4"
}

resource "docker_image" "elastic" {
	name = "elasticsearch:2.4.1"
}

# Create a container
resource "docker_container" "elastic" {
	count = "${var.elastic_count}"
	image = "${docker_image.elastic.name}"
	name = "elastic-${format("%02d", count.index+1)}"
	
	hostname = "elastic-${format("%02d", count.index+1)}"

	restart = "always"
	memory = 512

	dns = ["172.17.0.1"]

	labels {
		type = "elastic"
	}

	volumes {
		container_path  = "/usr/share/elasticsearch/data"
		host_path = "/home/dmportella/_workspaces/terraform/consul/data/elastic/data.${format("%d", count.index+1)}"
		read_only = false
	}

	volumes {
		container_path  = "/usr/share/elasticsearch/config"
		host_path = "/home/dmportella/_workspaces/terraform/consul/configs/elastic"
		read_only = false
	}
}

resource "consul_service" "elastic-single" {
    depends_on = ["null_resource.consul_provisioned"]
    count = "${var.elastic_count}"

    address = "${element(docker_container.elastic.*.ip_address, count.index)}"
    name = "elastic-${format("%02d", count.index+1)}"
    port = 9200
    tags = ["elastic"]
}

resource "consul_service" "elastic-cluster" {
    depends_on = ["null_resource.consul_provisioned"]
    count = "${var.elastic_count}"

    id = "elastic-cluster-${format("%02d", count.index+1)}"
    address = "${element(docker_container.elastic.*.ip_address, count.index)}"

    name = "elastic"
    port = 9200
    tags = ["elastic", "cluster"]
}
```
[...]

PR #9366, opened 2016-10-14
provider/consul: Fixes for consul_service resource

dmportella:
Added `service_id` in place of `id` for resource.
modified created, read, update to use `service_id`
modified tests to include `service_id`.
modified documentation for consul_service to include new value.

Tests results

CONSUL_HTTP_ADDR=localhost:8500 make testacc TEST=./builtin/providers/consul/

2016/10/14 14:43:05 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/consul/ -v  -timeout 120m
--- PASS: TestAccDataConsulKeys_basic (0.05s)
=== RUN   TestAccConsulAgentService_basic
--- PASS: TestAccConsulAgentService_basic (0.05s)
=== RUN   TestAccConsulCatalogEntry_basic
--- PASS: TestAccConsulCatalogEntry_basic (0.06s)
=== RUN   TestAccConsulKeyPrefix_basic
--- PASS: TestAccConsulKeyPrefix_basic (0.19s)
=== RUN   TestConsulKeysMigrateState
--- PASS: TestConsulKeysMigrateState (0.00s)
=== RUN   TestConsulKeysMigrateState_empty
--- PASS: TestConsulKeysMigrateState_empty (0.00s)
=== RUN   TestAccConsulKeys_basic
--- PASS: TestAccConsulKeys_basic (0.13s)
=== RUN   TestAccConsulNode_basic
--- PASS: TestAccConsulNode_basic (0.05s)
=== RUN   TestAccConsulPreparedQuery_basic
--- PASS: TestAccConsulPreparedQuery_basic (0.12s)
=== RUN   TestAccConsulService_basic
--- PASS: TestAccConsulService_basic (0.05s)
=== RUN   TestResourceProvider
--- PASS: TestResourceProvider (0.00s)
=== RUN   TestResourceProvider_impl
--- PASS: TestResourceProvider_impl (0.00s)
=== RUN   TestResourceProvider_Configure
--- PASS: TestResourceProvider_Configure (0.00s)
=== RUN   TestResourceProvider_ConfigureTLS
--- PASS: TestResourceProvider_ConfigureTLS (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/consul	0.708s

Refs: #9352

### DIFF
--- a/builtin/providers/consul/resource_consul_service.go
+++ b/builtin/providers/consul/resource_consul_service.go
@@ -22,9 +22,11 @@ func resourceConsulService() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"id": &schema.Schema{
+			"service_id": &schema.Schema{
 				Type:     schema.TypeString,
+				Optional: true,
 				Computed: true,
+				ForceNew: true,
 			},
 
 			"name": &schema.Schema{
@@ -53,7 +55,13 @@ func resourceConsulServiceCreate(d *schema.ResourceData, meta interface{}) error
 	agent := client.Agent()
 
 	name := d.Get("name").(string)
-	registration := consulapi.AgentServiceRegistration{Name: name}
+	identifier := name
+
+	if serviceId, ok := d.GetOk("service_id"); ok {
+		identifier = serviceId.(string)
+	}
+
+	registration := consulapi.AgentServiceRegistration{Name: name, ID: identifier}
 
 	if address, ok := d.GetOk("address"); ok {
 		registration.Address = address.(string)
@@ -79,12 +87,13 @@ func resourceConsulServiceCreate(d *schema.ResourceData, meta interface{}) error
 	// Update the resource
 	if serviceMap, err := agent.Services(); err != nil {
 		return fmt.Errorf("Failed to read services from Consul agent: %v", err)
-	} else if service, ok := serviceMap[name]; !ok {
-		return fmt.Errorf("Failed to read service '%s' from Consul agent: %v", name, err)
+	} else if service, ok := serviceMap[identifier]; !ok {
+		return fmt.Errorf("Failed to read service '%s' from Consul agent: %v", identifier, err)
 	} else {
-		d.Set("address", service.Address)
-		d.Set("id", service.ID)
 		d.SetId(service.ID)
+
+		d.Set("address", service.Address)
+		d.Set("service_id", service.ID)
 		d.Set("name", service.Service)
 		d.Set("port", service.Port)
 		tags := make([]string, 0, len(service.Tags))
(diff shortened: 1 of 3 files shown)


stack72:
Hi @dmportella 

Thanks for the work here - unfortunately, we can't just rename a parameter. This causes backward incompatibilities for people using the resource right now. What we have to do is:

* introduce a new parameter
* add a conflicts_with attribute to each to make sure both are not set
* add a deprecation to the old parameter to make sure users know it will be removed in a newer version
* have the ability that both parameters can be used to send the value to the cloud provider

Paul

apparentlymart:
@stack72 I think in this case things are a little more nuanced because Terraform treats the "id" attribute as a special-case alias for the instance id (as in `d.SetId`).

I remember while working on data sources that I had to make a special effort to support user-settable `id` attributes for looking up things by id, because there is an implicit schema on all resources that sets `id` as computed. I expect that's the same trouble that is being addressed here.

All of this ramble is to say: I don't think it is necessary to preserve the explicit `id` attribute here because it is handled automatically by `helper/schema` as long as `SetId` is called.

dmportella:
@stack72 I know I did consider the backwards compatibility, @apparentlymart is correct I noticed the `Id` is already used by terraform itself for the framework.

Set ID is definitely called on create and update so someone can still extract it since previously this property was never `settable` it could only be exported and that should still work since I set both service_id and call `r.SetID` with the same value. Notice that the tests still check for `ID` and they still pass.

@stack72 I agree with the the idea but i think this might be a special case because `ID` is like a `protected` keyword for terraform. 

> have the ability that both parameters can be used to send the value to the cloud provider

**^ This still be true always**

> add a deprecation to the old parameter to make sure users know it will be removed in a newer version

**^ The parameter cant be removed since it is an integral part of terraform.**

> introduce a new parameter

**^ this is done :+1: ;)**

> add a conflicts_with attribute to each to make sure both are not set

**^ I am happy to add this but I am not sure what to do here if you could provide examples...**

stack72:
Hi @dmportella 

After reading what @apparentlymart said, this PR is now indeed correct :)

```
% make testacc TEST=./builtin/providers/consul                                                                                            1 ↵
2016/10/26 13:05:52 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/consul -v  -timeout 120m
--- PASS: TestAccDataConsulKeys_basic (0.04s)
=== RUN   TestAccConsulAgentService_basic
--- PASS: TestAccConsulAgentService_basic (0.04s)
=== RUN   TestAccConsulCatalogEntry_basic
--- PASS: TestAccConsulCatalogEntry_basic (0.05s)
=== RUN   TestAccConsulKeyPrefix_basic
--- PASS: TestAccConsulKeyPrefix_basic (0.09s)
=== RUN   TestConsulKeysMigrateState
--- PASS: TestConsulKeysMigrateState (0.00s)
=== RUN   TestConsulKeysMigrateState_empty
--- PASS: TestConsulKeysMigrateState_empty (0.00s)
=== RUN   TestAccConsulKeys_basic
--- PASS: TestAccConsulKeys_basic (0.11s)
=== RUN   TestAccConsulNode_basic
--- PASS: TestAccConsulNode_basic (0.02s)
=== RUN   TestAccConsulPreparedQuery_basic
--- PASS: TestAccConsulPreparedQuery_basic (0.07s)
=== RUN   TestAccConsulService_basic
--- PASS: TestAccConsulService_basic (0.07s)
=== RUN   TestResourceProvider
--- PASS: TestResourceProvider (0.00s)
=== RUN   TestResourceProvider_impl
--- PASS: TestResourceProvider_impl (0.00s)
=== RUN   TestResourceProvider_Configure
--- PASS: TestResourceProvider_Configure (0.00s)
=== RUN   TestResourceProvider_ConfigureTLS
--- PASS: TestResourceProvider_ConfigureTLS (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/consul	0.507s
```

Thanks for the work here!

Paul

dmportella:
No worries glad to help!

PR #9370, opened 2016-10-14
provider/google: Ensure we don't assert on nil

### LINKED ISSUE #8609: Terraform panics when running plan after upgrading from 0.6.16 -> 0.7.2
I am attempting to upgrade terraform for a Google project from 0.6.16 -> 0.7.2. So far I have been unsuccessful in doing so. After making the necessary syntax updates, I consistently receive a panic when running plan or apply on the project. If I destroy the project and then rebuild the whole thing with 0.7.2 there is no issue. The panic only occurs when project is built with 0.6.16, the state file still exists, and then a terraform command is run with version 0.7.2.

### Terraform Version
Updating from 0.6.16 -> 0.7.2. Issues happen when running 0.7.2 with a state file from 0.6.16.

### Affected Resource(s)
Crashes

### Panic Output
https://gist.github.com/conord33/ed26056e5165b126ffa564af2e85f579

### Expected Behavior
When running plan with the upgraded version, I should see a plan with no actions to take.

### Actual Behavior
Crash

### LINKED ISSUE #9359: Terraform panics when running plan
### Terraform Version
0.6.16
Unable to upgrade because of a similar issue in #8609 

### Affected Resource(s)
- google_compute_target_pool

### Panic Output
https://gist.github.com/caseylang/d9278e9551db9d080155effa449e858c

### Expected Behavior
The plan should have been applied

### Actual Behavior
The plan panicked.

### References
- #8609

It's possible this issue is as well fixed by 6ee98a72e850803e8597b01111db2b9a3a15d33a but I thought this this is the first indication of this problem occurring with a 0.6.16 plan during an apply, it was worth creating a ticket. 

I should also bring up that, as opposed to `terraform plan` failing when using `0.7.2`, `terraform plan` runs without issue here. Only during `apply` does Terraform panic.

jen20:
This commit tests whether an interface is nil before type asserting it to string - this should fix the panic reported in #8609 and #9359.

We also clean up the schema definition to the newer style without redundant type declarations.

Fixes #8609, fixes #9359.

### DIFF
--- a/builtin/providers/google/resource_compute_target_pool.go
+++ b/builtin/providers/google/resource_compute_target_pool.go
@@ -21,38 +21,38 @@ func resourceComputeTargetPool() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"backup_pool": &schema.Schema{
+			"backup_pool": {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: false,
 			},
 
-			"description": &schema.Schema{
+			"description": {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
 			},
 
-			"failover_ratio": &schema.Schema{
+			"failover_ratio": {
 				Type:     schema.TypeFloat,
 				Optional: true,
 				ForceNew: true,
 			},
 
-			"health_checks": &schema.Schema{
+			"health_checks": {
 				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: false,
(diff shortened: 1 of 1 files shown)


builtin/providers/google/resource_compute_target_pool.go
```diff
@@ -91,7 +91,12 @@ func resourceComputeTargetPool() *schema.Resource {
 func convertStringArr(ifaceArr []interface{}) []string {
 	arr := make([]string, len(ifaceArr))
 	for i, v := range ifaceArr {
-		arr[i] = v.(string)
```
jbardin: this is a rare case where ignoring the `ok` is useful: `arr[i], _ = v.(string)` replaces the if-else

jen20: :+1: that's much nicer.

jbardin:
LGTM

caseylang:
Thanks for the work @jen20!

PR #9383, opened 2016-10-15
Small change to terraform import docs

georgespatton:


jamtur01:
This feels like it might work better in a dedicated "third party" tools page perhaps?

georgespatton:
Can you point me to a more appropriate place for this perhaps?  I didn't write the terraforming tool, but it was useful for initial use of terraform, so users may find value in knowing about the tool.

jamtur01:
@georgespatton Something like https://github.com/hashicorp/terraform/pull/9384.

georgespatton:
Looks good in that new location.

PR #9406, opened 2016-10-17
Update variables.html.md regarding variable files

jbruettcva:
Add information under Variable Files to explain how terraform.tfvars file is used.  currently that documenation only exists in the getting started guide.  The added information was taken directly from the getting started guide.


jen20:
Thanks @jbruettcva! I've merged this, and made another couple of tweaks to the precedence section below.

PR #9408, opened 2016-10-17
provider/azurerm: add account_kind and access_tier to storage_account

pmcatominey:
```
TF_ACC=1 go test ./builtin/providers/azurerm -v -run TestAccAzureRMStorageAccount -timeout 120m
=== RUN   TestAccAzureRMStorageAccount_importBasic
--- PASS: TestAccAzureRMStorageAccount_importBasic (140.06s)
=== RUN   TestAccAzureRMStorageAccount_basic
--- PASS: TestAccAzureRMStorageAccount_basic (155.43s)
=== RUN   TestAccAzureRMStorageAccount_disappears
--- PASS: TestAccAzureRMStorageAccount_disappears (134.99s)
=== RUN   TestAccAzureRMStorageAccount_blobEncryption
--- PASS: TestAccAzureRMStorageAccount_blobEncryption (161.59s)
=== RUN   TestAccAzureRMStorageAccount_blobStorageWithUpdate
--- PASS: TestAccAzureRMStorageAccount_blobStorageWithUpdate (131.49s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/azurerm	723.694s
```

### DIFF
--- a/builtin/providers/azurerm/resource_arm_storage_account.go
+++ b/builtin/providers/azurerm/resource_arm_storage_account.go
@@ -12,11 +12,14 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/signalwrapper"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 // The KeySource of storage.Encryption appears to require this value
 // for Encryption services to work
 var storageAccountEncryptionSource = "Microsoft.Storage"
+
+const blobStorageAccountDefaultAccessTier = "Hot"
 
 func resourceArmStorageAccount() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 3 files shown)


builtin/providers/azurerm/resource_arm_storage_account.go
```diff
@@ -153,6 +181,17 @@ func resourceArmStorageAccountCreate(d *schema.ResourceData, meta interface{}) e
 		},
 	}
 
+	// AccessTier is only valid for BlobStorage accounts
+	if accountKind == string(storage.BlobStorage) {
+		accessTier, ok := d.GetOk("access_tier")
```
stack72: I think we should add the default as part of the scheme declaration rather than defaulting in code

builtin/providers/azurerm/resource_arm_storage_account.go
```diff
@@ -317,10 +372,15 @@ func resourceArmStorageAccountRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("primary_access_key", accessKeys[0].Value)
 	d.Set("secondary_access_key", accessKeys[1].Value)
 	d.Set("location", resp.Location)
+	d.Set("account_kind", resp.Kind)
 	d.Set("account_type", resp.Sku.Name)
 	d.Set("primary_location", resp.Properties.PrimaryLocation)
 	d.Set("secondary_location", resp.Properties.SecondaryLocation)
 
+	if resp.Properties.AccessTier != "" {
```
stack72: If access tier is computed: true, is there ever any way this can be nil?

builtin/providers/azurerm/resource_arm_storage_account.go
pmcatominey: The issue I faced is that accessTier can only be specified on requests for BlobStorage accounts, using the Default parameter meant it would be set for all types

stack72:
Thanks for the responses - LGTM :)

PR #9420, opened 2016-10-17
Make TestApply_parallelism more reliable

jbardin:
Wait for our expected number of goroutines to be staged and ready, then
allow Run() to complete. Use a high-water-mark counter to ensure we
never exceeded the max expected concurrent goroutines at any point
during the run.


mitchellh:
This LGTM. I also fixed the shadow graph race in that PR.

PR #9442, opened 2016-10-18
provider/google: add scope aliases

obsh:
Hi,
this PR adds following gcloud scope aliases:
- cloud-source-repos
- monitoring-write
- service-control
- service-management

It is based on scope aliases list from this page: :https://cloud.google.com/sdk/gcloud/reference/compute/instances/create


builtin/providers/google/service_scope.go
```diff
@@ -4,23 +4,27 @@ func canonicalizeServiceScope(scope string) string {
...
+		"bigquery":           "https://www.googleapis.com/auth/bigquery",
+		"cloud-platform":     "https://www.googleapis.com/auth/cloud-platform",
+		"cloud-source-repos": "https://www.googleapis.com/auth/source.full_control",
+		"compute-ro":         "https://www.googleapis.com/auth/compute.readonly",
+		"compute-rw":         "https://www.googleapis.com/auth/compute",
+		"datastore":          "https://www.googleapis.com/auth/datastore",
+		"logging-write":      "https://www.googleapis.com/auth/logging.write",
+		"monitoring":         "https://www.googleapis.com/auth/monitoring",
+		"monitoring-write":   "https://www.googleapis.com/auth/monitoring.write",
+		"pubsub":             "https://www.googleapis.com/auth/pubsub",
```
cblecker: I don't see pubsub in the list you linked. Was it deprecated?

obsh: I've checked pubsub alias with gcloud, it doesn't work:

```
gcloud compute instances create test-scopes \
  --image-project=debian-cloud \
  --image-family=debian-8 \
  --machine-type n1-standard-1 \
  --scopes pubsub

 ERROR: (gcloud.compute.instances.create) Some requests did not succeed:
 - One or more of the service account scopes are invalid: 'pubsub'
 ```

obsh:
Also I see "cloud-source-repos-ro" in the list, I'm pretty sure it was added today.
Added to PR.

builtin/providers/google/service_scope.go
cblecker: So I guess we should remove this from the list in Terraform too, while we're updating the scopes?

obsh: Wouldn't it break configurations, that uses this alias?
Because full scope "https://www.googleapis.com/auth/pubsub" works, so alias is workable now in Terraform.

cblecker: Ah. Yeah, I just wanted to make sure that it was still a valid scope. Was able to confirm here: https://developers.google.com/identity/protocols/googlescopes#pubsubv1

stack72:
Thanks for the work here @obsh :) LGTM!

PR #9462, opened 2016-10-19
provider/aws: Added aws_api_gateway_api_key value attribute

### LINKED ISSUE #11105: provider/aws: Add the ability to set the API Gateway Key value
Hi everyone,

#### Description
This PR is made post-PR to keep track of the request.

When using the `aws_api_gateway_api_key` resource, the value for the key is computed and generated by AWS. 
However from a SDK point-of-view it is possible to set it, thus allowing user-defined tokens, and bringing even more flexibility to resources and use-cases.

#### Use-cases
As such, the main use-case (for me at the moment) is for demo/training purposes. 
Let's say you have 5 users, you want to have domain `api1` with key `user1token`, you want to have domain `api2` with key `user2token`, etc.

### Terraform Version
Checked against: Terraform v0.7.8-dev (bdb6089+CHANGES) & newest ones

### Affected Resource(s)
- aws_api_gateway_api_key

N/A

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
N/A

### Actual Behavior
N/A

### Steps to Reproduce
N/A

### Important Factoids
N/A

### References
Related PR: https://github.com/hashicorp/terraform/pull/9462

Ninir:
Hi there!

This closes #11105 

Checked against: Terraform v0.7.8-dev (bdb60893d5c48e5ad8bec9a74832abd4ff58fabd+CHANGES)
### Acceptance Tests

``` bash
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAPIGatewayApiKey_basic'                                                
2017/01/09 11:04:08 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAPIGatewayApiKey_basic -timeout 120m
=== RUN   TestAccAWSAPIGatewayApiKey_basic
--- PASS: TestAccAWSAPIGatewayApiKey_basic (20.87s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	20.898s
```
### References

[API Reference](https://github.com/aws/aws-sdk-go/blob/db5d68bbb6761e5800a4c4e646642eb18b08a2a6/models/apis/apigateway/2015-07-09/api-2.json#L2158-L2161) from the SDK


builtin/providers/aws/resource_aws_api_gateway_api_key.go
```diff
@@ -71,6 +76,7 @@ func resourceAwsApiGatewayApiKeyCreate(d *schema.ResourceData, meta interface{})
 		Description: aws.String(d.Get("description").(string)),
 		Enabled:     aws.Bool(d.Get("enabled").(bool)),
 		StageKeys:   expandApiGatewayStageKeys(d),
+		Value:       aws.String(d.Get("value").(string)),
```
stack72: IF this value is Computed: true and then nothing else, a user cannot set it

therefore this will fail - does the `value` need to be optional as well?

builtin/providers/aws/resource_aws_api_gateway_api_key.go
```diff
@@ -86,7 +92,8 @@ func resourceAwsApiGatewayApiKeyRead(d *schema.ResourceData, meta interface{}) e
 	log.Printf("[DEBUG] Reading API Gateway API Key: %s", d.Id())
 
 	apiKey, err := conn.GetApiKey(&apigateway.GetApiKeyInput{
-		ApiKey: aws.String(d.Id()),
+		ApiKey:       aws.String(d.Id()),
+		IncludeValue: aws.Bool(true),
```
stack72: I think this should be IncludeValue true *only if* a value is set - thoughts?

Ninir:
Hi @stack72 

Well, due to the current logic, the api_key is meant to be retrieved through the API, when getting the key.
Adding the `IncludeValue: aws.Bool(true)` is just passing a new query parameter as `includeValue=true` when getting the related api key.

After double checking the API, there are many ways to improve this resource.
I would prefer to throw this PR in, and then make the allowance of custom key (user-defined) in another PR. What do you think?

(first contributions, so not sure how to get into it 😄)

builtin/providers/aws/resource_aws_api_gateway_api_key.go
Ninir: This has been fixed :) got the point about the api key creation!

Ninir: You're right, it has been fixed for this work :)

Ninir:
Hi @stack72 !
I reworked this to include the retrieval of the token but also allowing users to set their own token.

Hope everything is fine and I didn't forget anything... first contributions here 😄

Ninir:
Hi @stack72,

Since this has been reworked, do you think this could go into 0.7.8?

Thank you.

stack72:
Hi @Ninir

If you can, please rebase this and I can get this tested and merged finally :)

Thanks

Paul

Ninir:
@stack72 Here it is! Currently running the acceptance test (just in case).

stack72:
Hi @Ninir 

Is this a sensitive value? i.e. is this something we want added to state? 

Paul

Ninir:
@stack72 

We may want to set is as sensitive or hide it from the state, since it acts as an authentication token yes. 

Will set the attribute in the schema. Thanks for pointing this out!

builtin/providers/aws/resource_aws_api_gateway_api_key.go
```diff
@@ -68,6 +68,15 @@ func resourceAwsApiGatewayApiKey() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"value": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				Sensitive:    true,
```
Ninir: @stack72 Are there any other things to do more than this? Not sure how to deal with the state yet

Ninir:
Hi @stack72, 

The sensitive attribute has been set.
is there something else you would want me to do?

Thanks!
Gauthier

stack72:
Hi @Ninir 

If you can rebase this, then we can get this merged - sorry this has taken so long!

Paul

Ninir:
Hi Paul,

Just rebased it :) 

As a note, the "API Gateway" API changed so even if this work will still help previous version of the API Gateways, new API Gateways will require Usage Plans (work already started on my own, PR exposed and soon to be ended) and the related API Keys.

Thanks!

stack72:
LGTM now thanks :)

PR #9373, opened 2016-10-14
Fix #9197: datadog_monitor spurious diffs

nyanshak:

PR #9466, opened 2016-10-20
provider/datadog: Reduce diff on thresholds

### LINKED ISSUE #9197: datadog_monitor always displays changes for thresholds
### Terraform Version

0.7.4

### Affected Resource(s)

datadog_monitor

Almost identical to configuration in [datadog_monitor documentation](https://www.terraform.io/docs/providers/datadog/r/monitor.html)
```hcl
resource "datadog_monitor" "foo" {
  name = "Name for monitor foo"
  type = "metric alert"
  message = "Monitor triggered. Notify: @none"

  query = "avg(last_1h):avg:aws.ec2.cpu{environment:foo,host:foo} by {host} > 2"

  thresholds {
    ok = 0
    warning = 1
    critical = 2
  }
}
```

### Debug Output

https://gist.github.com/billputer/83622712603e127b61407df05cbc2929

### Expected Behavior

`terraform plan` shows no changes

### Actual Behavior
`terraform plan` shows the following changes:
```
~ datadog_monitor.foo
    thresholds.critical: "2.0" => "2"
    thresholds.ok:       "0.0" => "0"
    thresholds.warning:  "1.0" => "1"
```

### Steps to Reproduce

1. `terraform apply`
2. `terraform plan`

### Important Factoids

Didn't happen on Terraform pre-0.7.

catsby:
Add a `DiffSuppressFunc` function to the `threshold` attribute of DataDog Monitor resource. 

I believe the DataDog API changed here recently, and these numbers are being returned as floats now. Instead of changing things in the state, or asking users to change their config files, we simply compare the float value of each value for equality, and suppress the diff if they are equal. 

Fixes #9197
Alternative to #9373

### DIFF
--- a/builtin/providers/datadog/resource_datadog_monitor.go
+++ b/builtin/providers/datadog/resource_datadog_monitor.go
@@ -73,6 +73,7 @@ func resourceDatadogMonitor() *schema.Resource {
 						},
 					},
 				},
+				DiffSuppressFunc: supressDataDogFloatIntDiff,
 			},
 			"notify_no_data": &schema.Schema{
 				Type:     schema.TypeBool,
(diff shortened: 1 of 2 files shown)


builtin/providers/datadog/resource_datadog_monitor.go
```diff
@@ -398,3 +399,26 @@ func resourceDatadogImport(d *schema.ResourceData, meta interface{}) ([]*schema.
...
+
+	nF, err := strconv.ParseFloat(new, 64)
+	if err != nil {
+		log.Printf("Error parsing float of new value (%s): %s", new, err)
+		return false
+	}
+
+	// if the float values of these attributes are equivalent, ignore this
+	// diff
+	if oF == nF {
```
jbardin: Do we ever need to worry comparing these within a delta, or do they only come from static string values?

PR #9475, opened 2016-10-20
Fix Scaleway documentation typo/error

IceBear2k:
This is definitely scaleway_server instead of scaleway_volume, wouldn't make any sense otherwise.


stack72:
LGTM - thanks @IceBear2k :)

PR #9480, opened 2016-10-20
add missing reference to `iops` property in `aws_ami` data source documentation

patdowney:
In the documentation the reference data within the `aws_ami` datasource was referring to `*.ebs.encrypted` twice, when one instance should have been `*.ebs.iops`. The descriptions for each of these keys is correct.

AFAIK - this affects version 0.7


catsby:
Thanks!

catsby:
Deployed! https://www.terraform.io/docs/providers/aws/d/ami.html

PR #9494, opened 2016-10-20
provider/google: Add google resources to importability docs

cblecker:
The google_compute_* objects that were previously had import scripts added, were not added to the website list of importable objects.


stack72:
Thanks for adding this @cblecker :) LGTM!

PR #9496, opened 2016-10-21
terraform: new apply resource node supports data sources

mitchellh:
This enables the new apply graph's resource node to apply data sources.
    Data sources appear to only be tested for "refresh" which is likely
    where they're set but they've also been implemented (not my code, not
    trying to edit code) within the "apply" operation as well.

    This adds an apply test to ensure data sources work, and then modifies
    the new apply node to support data sources.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -508,6 +508,38 @@ func TestContext2Apply_destroyComputed(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_dataBasic(t *testing.T) {
+	m := testModule(t, "apply-data-basic")
+	p := testProvider("null")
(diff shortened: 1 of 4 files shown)


mitchellh:
Merging since it doesn't touch old code, its blocking another branch, and it passes tests. 

Please review as you can though!

apparentlymart:
I'm not 100% sure of the context here but I assume what you're doing is shifting the existing apply code into the new structure required for the separate diff-based apply graph. In which case, this looks plausible and seems to still do all the same stuff the old codepath did.

mitchellh:
Responding via email since the DNS DDOS on the internet today won't let me
log in to GitHub...

Martin, that's exactly right.

PR #9497, opened 2016-10-21
I'd like to be able to anchor to functions

jamtur01:


stack72:
LGTM - thanks @jamtur01 :)

PR #9500, opened 2016-10-21
Bugfix in example

niclasnilsson:
Shouldn't the aws_route53_record in the example should use the "dev" zone_id?


stack72:
LGTM - thanks @niclasnilsson :)

PR #9517, opened 2016-10-21
command/meta: add -shadow flag to disable shadow graph

mitchellh:
Since it is still very much possible for this to cause problems, this
can be used to disable the shadow graph. We'll purposely not document
this since the goal is to remove this flag as we become more confident
with it.

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -63,10 +63,13 @@ type Meta struct {
 	//
 	// parallelism is used to control the number of concurrent operations
 	// allowed when walking the graph
+	//
+	// shadow is used to enable/disable the shadow graph
 	statePath    string
 	stateOutPath string
 	backupPath   string
 	parallelism  int
+	shadow       bool
 }
 
 // initStatePaths is used to initialize the default values for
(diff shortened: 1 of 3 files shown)


svanharmelen:
@mitchellh please consider **not** removing this flag somewhere in the future.

Or at least, if you do remove the flag, please keep the `shadow` field on the `Meta` and `Context` structs and keep the logic using these field to determine if you want to disable using the shadow graphs. That way people using Terraform in other ways than on the cmdline can still disable it in their code.

Thanks!

mitchellh:
@svanharmelen It exists `-Xshadow=false`. The shadow graph will start taking a back seat in 0.8 so I still don't think it'll be fully necessary but I will document it.

svanharmelen:
@mitchellh I noticed the `-shadow` flag added in this PR, but cannot seem to find the `-Xshadow` flag.

I'm now settings `Shadow: false` directly in `terraform.ContextOpts`, so if that will remain a valid field/option on the context that works for me. Thx!

PR #9523, opened 2016-10-22
adds missing `language` argument

ksatirli:
Best I can tell from https://github.com/hashicorp/terraform/blob/HEAD/builtin/providers/bitbucket/resource_repository.go#L23, the `language` argument is missing.

Adding it in a `bitbucket_repository` resource is supported:

```
resource "bitbucket_repository" "test-repository" {
  owner = "${var.bitbucket_username}"
  name = "test-repository"
  scm = "${var.bitbucket_settings["scm"]}"
  language = "markdown"
}
```


mitchellh:
Thanks!

PR #9528, opened 2016-10-23
Added a Parameter example

jrstarke:
**Reasoning for doc update**
Usage of parameters wasn't very obvious in the docs, and referred to a list of `Parameter structures` without describing what those were. After looking into the code, this is what I determined to be be how to use parameters. I felt adding an example into the documentation would help others in the future.

### DIFF
--- a/website/source/docs/providers/aws/r/cloudformation_stack.html.markdown
+++ b/website/source/docs/providers/aws/r/cloudformation_stack.html.markdown
@@ -15,13 +15,23 @@ Provides a CloudFormation Stack resource.
 ```
 resource "aws_cloudformation_stack" "network" {
   name = "networking-stack"
+  parameters = {
+    VPCCidrParameter = "10.0.0.0/16"
+  }
   template_body = <<STACK
 {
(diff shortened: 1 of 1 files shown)


website/source/docs/providers/aws/r/cloudformation_stack.html.markdown
```diff
@@ -15,13 +15,23 @@ Provides a CloudFormation Stack resource.
 ```
 resource "aws_cloudformation_stack" "network" {
   name = "networking-stack"
+  parameters = {
```
stack72: This can just be updated to:

```
  parameters {
    VpcCIDR = "10.0.0.0/16"
  }
```

Notice the lack of need for the `=` in the parameter declaration block and then the name of the parameter

jrstarke: Thanks for the review. Happy to make the change. :)

stack72:
LGTM now thanks @jrstarke :)

PR #9533, opened 2016-10-23
provider/aws: Exposed aws_api_gw_domain_name.certificate_upload_date attribute

Ninir:
### Description
This allows to get the API key certificate_upload_date when using the `aws_api_gateway_domain_name` resource.

Checked against: Terraform v0.7.8-dev (7cb2e69457b03c38332c4edfc849022c505852c5+CHANGES)

### DIFF
--- a/builtin/providers/aws/resource_aws_api_gateway_domain_name.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_domain_name.go
@@ -52,6 +52,11 @@ func resourceAwsApiGatewayDomainName() *schema.Resource {
 				Computed: true,
 			},
 
+			"certificate_upload_date": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"cloudfront_zone_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_api_gateway_domain_name.go
```diff
@@ -52,6 +52,11 @@ func resourceAwsApiGatewayDomainName() *schema.Resource {
 				Computed: true,
 			},
 
+			"certificate_upload_date": &schema.Schema{
```
stack72: We declare this parameter in the schema but never actually set the value to state...

Ninir: Hey @stack72 
thanks for the review. Not sure what you meant by this... 😞 is it about the `StateFunc`?

stack72: Hey - what i mean is, you have declared the parameter but  never set the parameter with a d.Set func call

Ninir: Oh got it! sorry, first contributions here... new to Go & Terraform development... 😅

Ninir: Well, it is already done looking at `https://github.com/Ninir/terraform/blob/6655e0f628e9d9b877aefe438a8ee63c17c45dfa/builtin/providers/aws/resource_aws_api_gateway_domain_name.go#L108` 

thank you again :)

stack72:
oh i see, you are just actually adding the value to the schema as that set was already in on master! In that case, LGTM!

P.

PR #9536, opened 2016-10-24
terraform: ResourceProvider.Stop

mitchellh:
This introduces a new `ResourceProvider` API: `Stop`.

This is called when Terraform is interrupted (Ctrl-C). It is mostly advisory, since the previous Stop logic is identical: Terraform core will wait until all currently executing graph nodes complete and then early exit from the graph walk.

The `Stop` API, however, notifies providers in some way that an interrupt has happened. This can be used by longer-running operations to exit early, thus expediting the "clean exit" scenario of interrupting Terraform. For example, waiting for an RDS instance to become ready can take many minutes. If a user ctrl-Cs, this API can be used to exit the wait loop waiting for the DB to become ready, and simply returning a partial state with the DB ID that will be updated in `refresh` in the future.

If a provider doesn't implement `Stop` (just returns nil), everything continues to operate as normal: Terraform just waits until the graph nodes complete and then exit as usual. 

**To reiterate: this is aimed primarily to be a UX improvement, although there have been some long (hours long) remote operations that should implement this (such as Azure storage deletion) to make Terraform actually function in those scenarios.**

In this PR:

  * Terraform core updated to use `Stop`
  * `helper/schema.Provider` updated with Stop-related functions
  * `helper/resource` has a new helper function for working with Stop
  * `providers/azurerm` updated as an example of using Stop to cancel an operation

:shipit:


jbardin:
Normally a [`Context`](https://golang.org/pkg/context/#Context) is used for cancellations, deadlines and timeouts. If we're codifying this in the API, could we upgrade to a more flexible/canonical implementation?

mitchellh:
@jbardin It'll be difficult to pass a `Context` across the RPC barrier, and I can't think of a use case for deadlines or timeouts from a Terraform core perspective that we can't solve by simply calling `Stop` on the API. I think what I want here is just a really basic notification that you're being stopped, not any rich additional metadata, deadlines, or timeouts.

Timeouts/deadlines WITHIN a resource I believe can be solved via `helper/schema` changes and shouldn't be a Terraform core feature.

mitchellh:
Talked out of band: agreed that on the provider side a Go `Context` makes a lot of sense. I've reworked the helper/schema interface and Azure example to use the context.

jbardin:
LGTM (pending travis)

Issue #8769, opened 2016-09-11
Proposal: Custom Diff Logic in Providers (WIP)


apparentlymart:
Currently the `Diff` step for all existing resources is implemented inside the `helper/schema` logic, focusing only on diffing the config against the state. There is currently no possibility for the provider's own code to do any of the following:
- veto particular changes in the diff (e.g. if it knows a particular change is impossible)
- customize the diff (e.g. if it knows that one change implies another that Terraform can't infer)
- replace the diff behavior altogether (e.g. if the upstream service is diff/apply based rather than resource-action-based)

This idea is in the general theme of "Terraform should detect as much as possible during plan". Making Terraform more trustworthy and robust requires us to be able to detect problems and conflicts while producing the plan, rather than bailing out while we're halfway through applying.
# Use-cases
## Vetoing particular changes in the diff

Some AWS operations (not the RDS ones, notably) support a "dry run" flag which allows a client to verify that the given credentials have suitable access and the arguments are syntactically valid. If we gave resources veto power over their diffs, they could potentially do a dry run of certain operations to catch some problems during the plan phase.

Amazon RDS imposes some restrictions on what changes are possible to the allocated storage on database instances, and these restrictions depend on the storage engine in use.

For example, at the time of writing the size value for all of the storage engines must either remain constant or be at least 10% larger than the present value, except Microsoft SQL server which does not permit changes at all. If the `aws_db_instance` resource were to have "veto" power on the diff then it could check these constraints during plan.

(This RDS use-case could potentially be thought of as another example for the following section, if we were to decide that asking RDS to make the storage smaller should act as if `ForceNew` were set.)
## Customizing the diff

Sometimes changing one attribute causes others to change. For example, updating an `aws_s3_bucket_object` when versioning is enabled will cause `version_id` to get updated.

Neither Terraform core nor `helper/schema` have enough information to infer this automatically. We have previously discussed a `ComputedWhen` attribute that can declare simplistic dependencies between attributes, but we can handle this more robustly across more cases by allowing a resource implementation to directly modify a diff produced by `helper/schema`, where it could add additional attribute diffs to represent changes it knows about due to its specific knowledge of a particular service.
## Replacing the diff behavior altogether

[...]

PR #9538, opened 2016-10-24
provider/nomad: Nomad provider for managing jobs

mitchellh:
This adds a new provider `nomad` with a single resource `nomad_job` that can be used for managing jobs within a [Nomad](https://www.nomadproject.io) cluster. 

The primary purpose for this provider is to manage system tasks or always-running service tasks that are used as part of cluster bootstrapping. For example, monitoring systems, core services, etc. Other jobs are usually submitted directly to Nomad (such as basic app deployment).

This is documented and tested (unit + acceptance). I rebased the commits so you can look at the `provider/nomad` commit to just see the provider separate from the vendoring.

Finally, this provider is built to work with Nomad 0.4.1. It likely won't work for earlier versions.


builtin/providers/nomad/provider_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+// How to run the acceptance tests for this provider:
+//
+// - Obtain an official Nomad release from https://nomadproject.io
+//   and extra "nomad" somewhere
```
jamtur01: extract? `extract the nomad binary` perhaps?

builtin/providers/nomad/resource_job.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if prevId != "" && prevId != jobspecStruct.ID {
+		log.Printf(
+			"[INFO] Deregistering %q before registering %q",
+			prevId, jobspecStruct.ID)
+
+		log.Printf("[DEBUG] Deregistering job: %q", prevId)
+		_, _, err := client.Jobs().Deregister(prevId, nil)
+		if err != nil {
+			return fmt.Errorf("error deregistering previous job %q: %s", prevId, err)
```
jamtur01: previous feels odd here. `error deregistering job` would work fine?

website/source/docs/providers/nomad/index.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+resource "nomad_job" "monitoring" {
+    jobspec = "${file("${path.module}/jobspec.hcl")}"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `address` - (Optional) The HTTP(S) API address of the agent to use. Defaults to "127.0.0.1:4646". `NOMAD_ADDR` can also be used.
```
jamtur01: `The HTTP(S) API address of the Nomad agent to use` - make it explicit to make it consistent. 

I'd probably also: `127.0.0.1:4646`.

jamtur01: `The "NOMAD_ADDR" environment can also be used.`? WIth `NOMAD_ADDR`.

builtin/providers/nomad/resource_job.go
mitchellh: I don't want the user to be confused why we may be deregistering a job "foo" when the jobspec perhaps has job "bar". I've updated the wording to be a bit more explicit about that, pushing soon.

website/source/docs/providers/nomad/index.html.markdown
jamtur01: Ditto with the `NOMAD_REGION` line.

website/source/docs/providers/nomad/r/job.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+description: |-
+  Manages a job registered in Nomad.
+---
+
+# nomad\_job
+
+Manages a job registered in Nomad.
+
+This can be used to initialize your cluster with system jobs, common services,
+and more. In day to day Nomad use it is common for developers to submit
```
jamtur01: The second sentence doesn't make sense to me? I think I'm either missing a Nomad concept or it's a little convoluted?

website/source/docs/providers/nomad/r/job.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+jobs to Nomad directly, but jobs that must always be running or system-level
+jobs can be registered with Terraform.
+
+## Example Usage
+
+```
+resource "nomad_job" "app" {
+    jobspec = "${file("${path.module}/job.hcl")}"
```
jamtur01: I doubt anyone is ever going to write a whole jobspec inline but maybe an example of a jobspec somewhere so folks know that it is they are looking for?

builtin/providers/nomad/provider_test.go
```diff
@@ -11,7 +11,7 @@ import (
 // How to run the acceptance tests for this provider:
 //
 // - Obtain an official Nomad release from https://nomadproject.io
-//   and extra "nomad" somewhere
+//   and extra the "nomad" binary
```
jamtur01: Still missing the `t` :)

mitchellh:
@jamtur01 Thanks! Addressed feedback (and fixed that typo), while adding more examples.

jamtur01:
@mitchellh Very neat BTW! Does << EOT blah EOT work in HCL? I had no idea. I wonder if that's documented anywhere? I'll go look. :)

mitchellh:
@jamtur01 Yes it does! You can even see it in "action" in the acceptance tests for this provider. It may not be documented actually... but I can do that separate from this PR.

jamtur01:
@mitchellh It's actually in the syntax docs already. Had just missed it! Thanks.

jen20:
@jamtur01 Yes - so does `<<-` to remove hanging indent (à la Ruby)

apparentlymart:
I was looking at #8769 primarily so that Terraform could interface with Nomad in a way that doesn't defeat Nomad's own diffing mechanism.

Certainly not meaning to suggest that this has to be blocked by #8769, but just wanted to point it out in case it helps to think about 1) whether such behavior is desirable anyway, and 2) assuming yes, whether adding it later would require a breaking change here that might make it worth tweaking the design now. (I think probably not, but I'm not 100% sure...)

mitchellh:
@apparentlymart I think in Nomad's cause that would be nice but doesn't block it since Nomad is idempotent. If you make a stylistic change that causes a plan to update the job, Nomad won't actually do anything once the job is resubmitted. I think that makes this okay.

jamtur01:
@jen20 Do you mean `<<~` (that removes hanging indent in Ruby - `<<-` preserves whitespace.

PR #9547, opened 2016-10-24
fix reading of environment attribute in resource_aws_opsworks_application

fmasuhr:
Reading the `environment` attribute for a `resource_aws_opsworks_application` currently does not update the state correctly.

Due to a wrong value passed to the `Set` method it will return an error:

```
environment.0.secure: '' expected type 'bool', got unconvertible type '*string'
```

Sadly there is no error handling implemented which caused the method to just ignore it and continue.

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_application.go
+++ b/builtin/providers/aws/resource_aws_opsworks_application.go
@@ -400,12 +400,7 @@ func resourceAwsOpsworksSetApplicationEnvironmentVariable(d *schema.ResourceData
 			data["value"] = *config.Value
 		}
 		if config.Secure != nil {
-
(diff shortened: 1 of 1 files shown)


stack72:
Hi @fmasuhr 

Thanks for the work here - please can you give me some thoughts on how I can test this functionality? Can you update one of the acceptance tests with the correct expected result?

Thanks

Paul

mitchellh:
Hello! I'm going through `waiting-response` labeled issues and closing issues that haven't been responded to with a meaningful response in at least 2+ weeks. If this is still an issue, please open a new issue so we can start anew. Feel free to reference this existing issue to make a link between the two. Thanks!

PR #9558, opened 2016-10-24
Add "gox" to the Makefile's tools section.

kwilczynski:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ tools:
 	go get -u github.com/kardianos/govendor
 	go get -u golang.org/x/tools/cmd/stringer
 	go get -u golang.org/x/tools/cmd/cover
+	go get -u github.com/mitchellh/gox
 
 # bin generates the releaseable binaries for Terraform
 bin: fmtcheck generate


kwilczynski:
@stack72 over to you 🚀 . I added it as it trips me every time. I hope it's fine with you.

catsby:
Fwiw, `gox` should be installed when `scripts/build.sh` is invoked... is that not happening for you?

kwilczynski:
@catsby oh! I haven't used this script before. My workflow usually as:

- git clone ...
- make tools
- make test / acctest
- make dev (or for build a specific provider)

Should I move over to the `build.sh`?

catsby:
Sorry for not being clear–

`$ make dev` invokes `$(CURDIR)/scripts/build.sh`, so you don't directly use it.

catsby:
```
$ which gox
/Users/clint/Projects/Go/bin/gox

$ rm $(which gox)

$ which gox
gox not found

$ make dev
2016/10/24 12:27:47 Generated command/internal_plugin_list.go
==> Removing old directory...
==> Installing gox...
==> Building...
Number of parallel builds: 3

-->    darwin/amd64: github.com/hashicorp/terraform

==> Results:
total 255216
-rwxr-xr-x  1 clint  staff   125M Oct 24 12:28 terraform
```

You could argue that it should be in `make tools`, though I don't think it's really necessary. I could say better if I knew the impetus of this pr, were you hitting errors?

kwilczynski:
@catsby Blimey. I must have done something wrong, as this clearly works as it says on the tin. 

Apologies for the unnecessary commotion :(

catsby:
Well, it's only unnecessary if there is no actual problem. If you'r e hitting something please let us know!

PR #9561, opened 2016-10-24
provider/aws: Remove VPC Endpoint from state if it's not found

catsby:
If a VPC Endpoint is removed outside of Terraform, we need to remove it from state when it's not found in the `READ` method


stack72:
LGTM!

PR #9573, opened 2016-10-25
Allow resizing DigitalOcean Droplets without increasing disk size.

### LINKED ISSUE #9402: Feature Request: resize DigitalOcean droplet without increasing disk size
- **Affected Resource:** `digitalocean_droplet`
- **Feature Request:** Add new argument to be able resize DigitalOcean droplet without increasing disk size

DigitalOcean supports two ways to resize a droplet:

- only increase or decrease the **CPU** and **RAM** of your Droplet, **not disk size**. This can be reversed.
- increase the disk size, CPU and RAM of your Droplet. This is a permanent change and cannot be reversed.

Terraform v0.7.6 [uses the second method](https://github.com/hashicorp/terraform/blob/eac6546e33c30034e5507ddc21747459eaa62003/builtin/providers/digitalocean/resource_digitalocean_droplet.go#L324
) to handle such changes.

andrewsomething:
This adds a new Boolean argument for the DigitalOcean Droplet resource, `resize_disk` It defaults to `true` to match the current behavior. When set to `false`, resizing a Droplet will only resize the RAM and CPU making it a reversible operation.

https://developers.digitalocean.com/documentation/v2/#resize-a-droplet

It also exposes the `disk` and `vcpus` attributes.

It resolves #9402.

### DIFF
--- a/builtin/providers/digitalocean/resource_digitalocean_droplet.go
+++ b/builtin/providers/digitalocean/resource_digitalocean_droplet.go
@@ -53,6 +53,22 @@ func resourceDigitalOceanDroplet() *schema.Resource {
 				},
 			},
 
+			"disk": &schema.Schema{
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"vcpus": &schema.Schema{
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"resize_disk": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
 			"status": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
@@ -243,6 +259,8 @@ func resourceDigitalOceanDropletRead(d *schema.ResourceData, meta interface{}) e
 	d.Set("name", droplet.Name)
 	d.Set("region", droplet.Region.Slug)
 	d.Set("size", droplet.Size.Slug)
+	d.Set("disk", droplet.Disk)
+	d.Set("vcpus", droplet.Vcpus)
 	d.Set("status", droplet.Status)
 	d.Set("locked", strconv.FormatBool(droplet.Locked))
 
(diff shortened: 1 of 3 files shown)


stack72:
hi @andrewsomething 

Thanks for the PR here - i had to make 1 change. Your code caused a failure in the import test :)

```
=== RUN   TestAccDigitalOceanDroplet_importBasic
--- FAIL: TestAccDigitalOceanDroplet_importBasic (44.58s)
	testing.go:265: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) {
		}


		(map[string]string) (len=1) {
		 (string) (len=11) "resize_disk": (string) (len=4) "true"
		}

```

This was because we never actually set `resize_disk` to state. So I simply ignored that field in the test - as it isn't part of an import :)

```
% make testacc TEST=./builtin/providers/digitalocean TESTARGS='-run=TestAccDigitalOceanDroplet_importBasic'                               2 ↵
2016/10/25 11:39:26 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/digitalocean -v -run=TestAccDigitalOceanDroplet_importBasic -timeout 120m
=== RUN   TestAccDigitalOceanDroplet_importBasic
--- PASS: TestAccDigitalOceanDroplet_importBasic (56.04s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/digitalocean	56.049s
```

I have manually merged this PR to get this changed added

Thanks for all the work here

Paul

andrewsomething:
Good catch. Thanks @stack72!

nikoren:
hi @andrewsomething , I'm trying to upgrade terraform from 0.7 to 0.9.3 , and terraform identifies that there is not `resize_disk` setting in my current state:

```hcl
~ digitalocean_droplet.stage-share-xxxx.1
    resize_disk: "" => "true"
```

When I try to apply the change , the node is  restarted, and in digital_ocean console I can see the following history:


|Event |	Initiated	Execution |Time|
|------|----------------------|-----|
|Power On |1 hour ago               |	6 seconds|
|Resize|	2 hours ago	|17 minutes 19 seconds|
|Power Off|	2 hours ago|	11 seconds|
|Create|	1 month ago|	20 seconds|



Nothing gets resized but as you can see node goes down , which makes upgrading terraform version very complicated  because all nodes should be restarted .

The question is , if there any workaround exist for this issue , or maybe I'm doing something wrong?

PR #9581, opened 2016-10-25
Added provider to meta-param resource docs

jamtur01:
I am not sure `provider` IS a meta-param but it looks like one...


mitchellh:
Thanks, looks great.

PR #9591, opened 2016-10-25
[WIP] provider/gitlab: Add a gitlab provider

richardc:
Opening as WIP for early review - 
I plan on adding an additional provider for `gitlab_project_member` but that can be a followup PR if people like the way this works already.

This adds a gitlab provider and `gitlab_project` resource.  It's copied heavily from the `github` provider and `github_repository`  resource, as there is some overlap in those types/apis.

Sample usage:

```
  name = "test1"
  visibility_level = "public"
}
```

Its implemented in terms of the [go-gitlab](https://github.com/xanzy/go-gitlab) library, which provides a wrapping of the [gitlab api](https://docs.gitlab.com/ee/api/)

We have been a little selective in the properties we surface for the project resource, as not all properties are very instructive.
Notable is the removal of the `public` bool as the `visibility_level`  will take precedent if both are supplied which leads to confusing interactions if they disagree.

### DIFF
--- a/builtin/providers/gitlab/provider.go
+++ b/builtin/providers/gitlab/provider.go
@@ -1,0 +1,52 @@
+package gitlab
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+// Provider returns a terraform.ResourceProvider.
+func Provider() terraform.ResourceProvider {
+
+	// The actual provider
+	return &schema.Provider{
+		Schema: map[string]*schema.Schema{
+			"token": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_TOKEN", nil),
+				Description: descriptions["token"],
+			},
+			"base_url": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_BASE_URL", ""),
+				Description: descriptions["base_url"],
+			},
+		},
+		ResourcesMap: map[string]*schema.Resource{
+			"gitlab_project": resourceGitlabProject(),
+		},
+
+		ConfigureFunc: providerConfigure,
+	}
+}
+
+var descriptions map[string]string
+
+func init() {
+	descriptions = map[string]string{
+		"token": "The OAuth token used to connect to GitLab.",
+
+		"base_url": "The GitLab Base API URL",
+	}
+}
+
+func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	config := Config{
+		Token:   d.Get("token").(string),
+		BaseURL: d.Get("base_url").(string),
+	}
+
+	return config.Client()
+}
--- a/builtin/providers/gitlab/resource_gitlab_project.go
+++ b/builtin/providers/gitlab/resource_gitlab_project.go
@@ -31,18 +31,22 @@ func resourceGitlabProject() *schema.Resource {
 			"issues_enabled": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
+				Default:  true,
 			},
 			"merge_requests_enabled": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
+				Default:  true,
 			},
 			"wiki_enabled": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
+				Default:  true,
 			},
 			"snippets_enabled": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
+				Default:  true,
 			},
 			"visibility_level": &schema.Schema{
 				Type:         schema.TypeString,
--- a/builtin/providers/gitlab/resource_gitlab_project.go
+++ b/builtin/providers/gitlab/resource_gitlab_project.go
@@ -2,6 +2,7 @@ package gitlab
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/xanzy/go-gitlab"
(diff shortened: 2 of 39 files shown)


richardc:
I have been running the acceptance tests and most development over the docker image of gitlab-ce, it needed some manual setup like so:

~~~
# start.sh
docker run --detach \
    --hostname gitlab.example.com \
    --publish-all \
    --name gitlab \
    --restart always \
    gitlab/gitlab-ce:latest
~~~

Login for the first time, set a password, then set up an api token.

~~~
# env.sh
export GITLAB_TOKEN="XXXXXX"
export GITLAB_BASE_URL=$(docker inspect gitlab | jq -r '.[0].NetworkSettings.Ports."80/tcp"[0] | "http://" + .HostIp + ":" + .HostPort + "/api/v3"')
~~~

Then I can run the acceptance tests/sample manifests with `. env.sh` and `TF_ACC=1 go test ./builtin/provider/gitlab`

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,138 @@
...
+	d.SetId(fmt.Sprintf("%d", project.ID))
+
+	return resourceGitlabProjectRead(d, meta)
+}
+
+func resourceGitlabProjectRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project, _, err := client.Projects.GetProject(d.Id())
+	if err != nil {
+		return err
```
stack72: I think we need to check for 404 from the API here - if someone deletes this manually from the UI, then this will error - we should refresh from state if that happens

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,138 @@
...
+	d.Set("ssh_url_to_repo", project.SSHURLToRepo)
+	d.Set("http_url_to_repo", project.HTTPURLToRepo)
+	d.Set("web_url", project.WebURL)
+
+	return nil
+}
+
+func resourceGitlabProjectUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	edit := &gitlab.EditProjectOptions{
```
stack72: i think we need to follow the same pattern above - and pass only the values that are changing rather than empty strings

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,138 @@
...
+			},
+		},
+	}
+}
+
+func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	options := &gitlab.CreateProjectOptions{
+		Name:                 gitlab.String(d.Get("name").(string)),
+		Description:          gitlab.String(d.Get("description").(string)),
```
richardc: Cool; will make those changes.

builtin/providers/gitlab/provider.go
```diff
@@ -0,0 +1,52 @@
...
+		"token": "The OAuth token used to connect to GitLab.",
+
+		"base_url": "The GitLab Base API URL",
+	}
+}
+
+func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	config := Config{
+		Token:   d.Get("token").(string),
+		BaseURL: d.Get("base_url").(string),
```
richardc: I'll test that more thoroughly;  I think the sane default here is to target the public gitlab.com install, but I haven't tested if this is the behaviour.

builtin/providers/gitlab/provider.go
```diff
@@ -0,0 +1,52 @@
...
+		Schema: map[string]*schema.Schema{
+			"token": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_TOKEN", nil),
+				Description: descriptions["token"],
+			},
+			"base_url": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
```
richardc: I'm not sure, I think I may strip this property it or mark it as Computed, as I don't think it's particularly a meaningful property to set, but is possibly interesting to inspect.

builtin/providers/gitlab/resource_gitlab_project.go
richardc: :+1: good catch

richardc:
Thanks for the comments, I'll address them soon.  Is it more convenient to add changes as distinct commits to squash later, or squash as I go?

builtin/providers/gitlab/provider.go
richardc: It would default to the empty string, but as per the comment on provider.go#48 I'm not sure if that will correctly map to the default gitlab.com api.  It should I think.

builtin/providers/gitlab/provider.go
richardc: Tested and it fails as we called SetBaseURL with the empty string, 7cefd8c fixes this for now by only calling it when a base_url is set.

stack72:
@richardc adding commits for fixes (like you have done) makes it easy to review - we can then rebase at the end :)

richardc:
@stack72 is there a pattern we should use to signal that the resource/provider is not fully-functional in a given configuration?   I've been bitten by issue in the gitlab api that is targeted for fixing in a later release, so would like to handle this gracefully by surfacing to the user that some attributes can not be managed correctly when running a certain version of gitlab.

stack72:
Hi @richardc 

How is this progressing? In a situation where you need it reviewed and tested yet?

Paul

richardc:
Sorry @stack72 this dropped off my radar so I didn't clean up a few things that were bothering me, and calling this phase 'done'.

I'll try and get  it cleaned up and to a point where it's reviewable in the next couple of weeks between turkeys and merriment.  Thanks for the reminder.

stack72:
👍

stack72:
ping @richardc :)

stack72:
Hi @richardc 

This PR hasn't had a lot of activity in a long time - I am going to close this out for now - if you feel you can come back to it at some point in the future then please do 

Thanks for all the work you have put in here

Paul

jugatsu:
@richardc can you please share some info on what is the progress of this PR?

richardc:
Well the short version is prior to the long weekend just gone I found myself running critically low on both tuits and damns, so I let this, and other software projects, fall by the wayside.  On Monday I started blowing the dust off this again, and here's my rescoped TODO list for this PR

- [ ] Rebase to current master for ease of comparison/review
- [ ] Yak shave into govendor to understand why it's so slow on my MacBook
- [ ] Update the go-gitlab vendoring to the latest stable release
- [ ] Revert the workarounds for gitlab issue [22831](https://gitlab.com/gitlab-org/gitlab-ce/issues/22831) 
- [ ] Extend acceptance tests to cover all attributes of gitlab_project
- [ ] Squash code down, and then re-open/re-submit the PR for review

I'm making no promises about timescales right now, but I am actively kicking this down the street.

jugatsu:
@richardc thanks for reply

builtin/providers/gitlab/resource_gitlab_project.go
richardc: Addressed in af6f37cb93

richardc:
This has been re-submitted as PR #13898 as I couldn't see how to re-open this PR.  

@stack72 sorry for all the delays, please accept this belated ping.

PR #9603, opened 2016-10-25
Fix the misuse of the word 'comprised'

relaxdiego:
The proper use of "comprise" is "Array1 comprises item1, item2, and item3" 
which is equivalent to saying "Array1 is composed of item1, item2, and item3." 
That is, "comprises" is equivalent to "is composed of." Therefore, to say 
"Array1 is comprised of item1, item2, and item3" is equivalent to saying 
"Array1 IS IS COMPOSED OF OF item1, item2, and item3" which makes no
sense and is like "The La Trattoria" from Mickey Blue Eyes! This change fixes
the misuse of the word.


stack72:
Hi @relaxdiego 

Thanks for the work here! LGTM!

P.

PR #9605, opened 2016-10-25
provider/aws: aws_iam_user_login_profile resource

jen20:
This commit introduces an `aws_iam_user_login_profile` resource which creates a password for an IAM user, and encrypts it using a PGP key specified in the configuration or obtained from Keybase.

For example:

```
resource "aws_iam_user" "u" {
       name = "auser"
       path = "/"
       force_destroy = true
}

resource "aws_iam_user_login_profile" "u" {
       user = "${aws_iam_user.u.name}"
       pgp_key = "keybase:some_person_that_exists"
}

output "password" {
value = "${aws_iam_user_login_profile.u.encrypted_password}"
}
```

The resulting attribute "encrypted_password" can be decrypted using PGP or Keybase - for example:

```
terraform output password | base64 --decode | keybase pgp decrypt
```

Optionally the user can retain the password rather than the default of being forced to change it at first login. Generated passwords are currently 20 characters long by default, though this can be overriden.

It is unclear how to test this correctly - we should discuss!


mitchellh:
Looks great, but I think the following should be done:

**Destroy:** `DeleteLoginProfile` exists and I think we should support it. It looks simple enough. If its useful, you can copy the `delete_on_destroy` type setting from Nomad provider so the user can choose to disable destroying it. But I think for testing reasons this is important.

**Exists:** `GetLoginProfile` exists and should be used to verify a login profile exists.

**Testing:** I think the easiest way to test this is just a basic create/destory test that just verifies it succeeds. We could go more hardcore and actually use the resulting password to try to run more API calls. I don't think its necessary but if you want. 

We can probably skip **update** for now.

jen20:
Hi @mitchellh! I added some acceptance tests that verify the behaviour for good and bad keys. For the non-keybase keys we verify that the password is usable (by changing it to something else random).

I don't think we should implement destroy (and thus exists is not necessary) - there is too much risk of removing a user's login credentials, and `aws_iam_user` will remove it at any rate, if `force_destroy` is specified (in the same manner as access keys).

Acceptance test run:

```
make testacc TEST=./builtin/providers/aws TESTARGS="-run TestAccAWSUserLoginProfile"
2016/10/25 16:28:13 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run TestAccAWSUserLoginProfile -timeout 120m
=== RUN   TestAccAWSUserLoginProfile_basic
--- PASS: TestAccAWSUserLoginProfile_basic (22.80s)
=== RUN   TestAccAWSUserLoginProfile_keybase
--- PASS: TestAccAWSUserLoginProfile_keybase (40.11s)
=== RUN   TestAccAWSUserLoginProfile_keybaseDoesntExist
--- PASS: TestAccAWSUserLoginProfile_keybaseDoesntExist (11.65s)
=== RUN   TestAccAWSUserLoginProfile_notAKey
--- PASS: TestAccAWSUserLoginProfile_notAKey (7.19s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	81.773s
```

mitchellh:
Awesome, LGTM!

Ninir:
Hi @jen20! small question from a newbie here (trying to get into Go & Terraform :smile:): as the password is generated, it's missing the ability to generate a custom password right?
Hoped I read the code well... 😅 

Use-case can be demos or handson for instance.

If you think it is doable, I can tro to send a PR for this :)
Still, awesome work!

PR #9613, opened 2016-10-26
helper/schema: only mark "ForceNew" on resources that cause the ForceNew

### LINKED ISSUE #2748: "Forces new resource" inaccurate
```
-/+ aws_launch_configuration.foo
    ...
    security_groups.#:           "3" => "4" (forces new resource)
    security_groups.1751326148:  "" => "sg-10bb7377" (forces new resource)
    security_groups.2708180956:  "sg-ed26ef8a" => "sg-ed26ef8a" (forces new resource)
    security_groups.4110005945:  "sg-e026ef87" => "sg-e026ef87" (forces new resource)
    security_groups.629771675:   "sg-e126ef86" => "sg-e126ef86" (forces new resource)
```

Three of the security groups remain the same, so it's misleading to mark the unchanged groups as forcing a new resource. Only the new SG should be labeled as forcing.

mitchellh:
Fixes #2748

This changes the diff to only mark "forces new resource" on the fields
that actually caused the new resource, not every field that changed.
This makes diffs much more accurate.

I'd like to request a review but I'm going to defer merging until
Terraform 0.8. Changes like this are very possible to cause "diffs
didn't match" errors and I want some real world testing in a beta before
we hit prod with this.

### DIFF
--- a/helper/schema/schema_test.go
+++ b/helper/schema/schema_test.go
@@ -2341,9 +2341,8 @@ func TestSchemaMap_Diff(t *testing.T) {
 						RequiresNew: true,
 					},
 					"instances.3": &terraform.ResourceAttrDiff{
-						Old:         "333",
-						New:         "333",
-						RequiresNew: true,
+						Old: "333",
+						New: "333",
 					},
 					"instances.4": &terraform.ResourceAttrDiff{
 						Old:         "",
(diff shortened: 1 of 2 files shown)


helper/schema/schema_test.go
```diff
@@ -2420,29 +2419,87 @@ func TestSchemaMap_Diff(t *testing.T) {
...
+						NewRemoved: true,
+					},
+				},
+			},
+
+			Err: false,
+		},
 	}
 
 	for tn, tc := range cases {
```
mitchellh: All the changes in this section here (below) are me just converting this to use subtests so you can run one at a time if needed.

mitchellh:
Thanks, going to hold for a TF 0.8 merge for beta.

PR #9617, opened 2016-10-26
provider/openstack: Allow any protocol in openstack_fw_rule_v1

### LINKED ISSUE #7507: provider/openstack: Can not create firewall rule with "any" protocol
### Terraform Version
Terraform v0.7.0-rc2

### Affected Resource(s)
- openstack_fw_rule_v1

```hcl
resource "openstack_fw_rule_v1" "test-case" {
  name = "test-case"
  action = "allow"
  protocol = "any"
  destination_ip_address = "1.1.1.2"
  source_ip_address = "1.1.1.1"
  enabled = "true"
}
```

### Expected Behavior
Create rule with protocol `any`

### Actual Behavior
`Expected HTTP response code [201 202] when accessing [POST https://OPENSTACK-PROVIDER:9696/v2.0/fw/firewall_rules], but got 400 instead
{"NeutronError": {"message": "Firewall Rule protocol any is not supported. Only protocol values [None, 'tcp', 'udp', 'icmp'] and their integer representation (0 to 255) are supported.", "type": "FirewallRuleInvalidProtocol", "detail": ""}}`

Setting `protocol = ""` gives `A protocol is required (tcp, udp, icmp or any)`, and `protocol = "0"` (since the error message indicated that numerals would be accepted) gives `Invalid input for protocol. Reason: '0' is not in [None, 'tcp', 'udp', 'icmp']`

jtopjian:
This commit allows a protocol of "any" to be used in the firewall
rule resource, which will allow any protocol.

Fixes #7507 

/cc @carlpett

### DIFF
--- a/builtin/providers/openstack/resource_openstack_fw_rule_v1.go
+++ b/builtin/providers/openstack/resource_openstack_fw_rule_v1.go
@@ -88,11 +88,12 @@ func resourceFWRuleV1Create(d *schema.ResourceData, meta interface{}) error {
 
 	enabled := d.Get("enabled").(bool)
 	ipVersion := resourceFWRuleV1DetermineIPVersion(d.Get("ip_version").(int))
+	protocol := resourceFWRuleV1DetermineProtocol(d.Get("protocol").(string))
 
 	ruleConfiguration := rules.CreateOpts{
 		Name:                 d.Get("name").(string),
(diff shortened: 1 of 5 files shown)


carlpett:
Looks great, thanks @jtopjian!

PR #9618, opened 2016-10-26
helper/schema,terraform: handle computed primitives in diffs

### LINKED ISSUE #3309: Bug: creating instances with EIPs based on a count variable, applying, then increasing that count variable causes an error
We have set up a repo at https://github.com/Ensighten/terraform-eip-bug which demonstrates the issue.

When trying to create instances with EIPs based on a count variable, the following error is produced after applying, increasing the count, and applying again:

Terraform version: Terraform v0.6.3

```
Error applying plan:

1 error(s) occurred:

* aws_eip.test-instance.0: diffs didn't match during apply. This is a bug with Terraform and should be reported.

```

The instances were created and EIPs were successfully bound despite the error. On running further plans and applies, no
changes are detected.

# Steps to reproduce:

1. Run `terraform apply --var subnet=$SOME_SUBNET_ID`
2. Modify main.tf to change the "count" variable for the "test" module to be 2
3. Run `terraform plan --var subnet=$SOME_SUBNET_ID` - observe that EIP instance attributes will be changed (from
   a concrete instance ID to an element function call into `aws_instance.test-instance.*.id)`:

   ```
        ~ module.test.aws_eip.test-instance.0
          instance: "<INSTANCE ID>" => "${element(aws_instance.test-instance.*.id, count.index)}"
   ```
4. Run `terraform apply --var subnet=$SOME_SUBNET_ID`, see error message given above.

mitchellh:
Fixes #3309

There are two primary changes, one to how helper/schema creates diffs
and one to how Terraform compares diffs. Both require careful
understanding.

## 1. helper/schema Changes

helper/schema, given any primitive field (string, int, bool, etc.)
_used to_ create a basic diff when given a computed new value (i.e. from
an unkown interpolation). This would put in the plan that the old value
is whatever the old value was, and the new value was the actual
interpolation. For example, from #3309, the diff showed the following:

```
~ module.test.aws_eip.test-instance.0
    instance: "<INSTANCE ID>" => "${element(aws_instance.test-instance.*.id, count.index)}"
```

Then, when running `apply`, the diff would be realized and you would get
a diff mismatch error because it would realize the final value is the
same and remove it from the diff.

**The change:** `helper/schema` now marks unknown primitive values with
`NewComputed` set to true. Semantically this is correct for the diff to
have this information.

## 2. Terraform Diff.Same Changes

Next, the way Terraform compares diffs needed to be updated

Specifically, the case where the diff from the plan had a NewComputed
primitive and the diff from the apply _no longer has that value_. This
is possible if the computed value ended up being the same as the old
value. This is allowed to pass through.

Together, these fix #3309.

### DIFF
--- a/helper/schema/schema_test.go
+++ b/helper/schema/schema_test.go
@@ -439,8 +439,9 @@ func TestSchemaMap_Diff(t *testing.T) {
 			Diff: &terraform.InstanceDiff{
 				Attributes: map[string]*terraform.ResourceAttrDiff{
 					"availability_zone": &terraform.ResourceAttrDiff{
-						Old: "",
-						New: "${var.foo}",
+						Old:         "",
+						New:         "${var.foo}",
+						NewComputed: true,
 					},
 				},
 			},
@@ -1675,8 +1676,9 @@ func TestSchemaMap_Diff(t *testing.T) {
 						New: "1",
 					},
 					"route.~1.gateway": &terraform.ResourceAttrDiff{
-						Old: "",
-						New: "${var.foo}",
+						Old:         "",
+						New:         "${var.foo}",
+						NewComputed: true,
 					},
 				},
 			},
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -995,7 +995,7 @@ func (m schemaMap) diffString(
 	all bool) error {
 	var originalN interface{}
 	var os, ns string
-	o, n, _, _ := d.diffChange(k)
+	o, n, _, computed := d.diffChange(k)
 	if schema.StateFunc != nil && n != nil {
 		originalN = n
 		n = schema.StateFunc(n)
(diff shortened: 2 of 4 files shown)


helper/schema/schema_test.go
```diff
@@ -439,8 +439,9 @@ func TestSchemaMap_Diff(t *testing.T) {
 			Diff: &terraform.InstanceDiff{
 				Attributes: map[string]*terraform.ResourceAttrDiff{
 					"availability_zone": &terraform.ResourceAttrDiff{
-						Old: "",
-						New: "${var.foo}",
+						Old:         "",
+						New:         "${var.foo}",
+						NewComputed: true,
```
mitchellh: This here is an actual test CHANGE versus a new change. The old behavior I believe to be incorrect since there was no way to tell that when applying, the EXACT value of `${var.foo}` wasn't expected... the diff comparison would simply never work.

mitchellh:
Thanks!

Issue #9169, opened 2016-10-02
Inconsistent operators precedence


bpineau:
Hi,

### Terraform Version
Terraform v0.7.4

### Terraform Configuration Files
```hcl
data "template_file" "example" {
  template = <<EOF

var1 = ${var1}
var2 = ${var2}
EOF

  vars {
    var1 = "${1 + 3 % 2}"
    var2 = "${3 % 2 + 1}"
  }
}

output "rendered" {
  value = "${data.template_file.example.rendered}"
}
```

### Debug Output
rendered = 
var1 = 0
var2 = 2

### Expected Behavior

If "%" has precedence over "+", var1 should output "2"
If "+" has precedence over "%", var2 should output "0"

### Actual Behavior
In the "1 + 3 % 2" case, the addition takes precedence (outputs 0).
In the "3 % 2 + 1" case, the modulo takes precedence (outputs 2)

Same thing with a simple division ("${2 + 2 / 2}" => 2 , "${2 / 2 + 2}" => 3).

As if the values were computed left to right, without precedence considerations ?
If that so, maybe should this be stated in the interpolation doc's "Math" section (since it could be counter-intuitive) ?

### Steps to Reproduce
1. `terraform apply`

rottenbytes:
upvote for this one. even if using parenthesis allows to force order, this is pretty annoying

mitchellh:
The precedence isn't inconsistent, they're equal: `+` is equal to `%` currently. This is primarily due to legacy reasons and changing this now would actually likely break a large number of configurations.

We can mark this for 0.8 since it'll break configurations. :(

rottenbytes:
Hi @mitchellh,

I had the behaviour with `-` and `*` operators.

Example : in `var.myvar * 1024.0 - var.myvar * 1024.0 * 0.9` : I had to use parenthesis and come up with `(var.myvar * 1024.0) - (var.myvar * 1024.0 * 0.9)` to get the expected result.

I think inconsistent is used here to mark what is usually expected from mathematic operator precedence.

mitchellh:
The expectation is fair and is why this issue is still open, but the fact is that this will still be a hugely backwards incompatible change and will likely break existing configs so we need to be careful about introducing it only in a major release, which we'll plan.

rottenbytes:
I totally get that and understand not introducing breaking changes 🙇  Maybe add a hint in the docs as @bpineau stated ?

mitchellh:
[...]

PR #9624, opened 2016-10-26
Add a note about operator precedence. #9169

rottenbytes:
This little note should help users until the issue #9169 is addressed.


mitchellh:
Thanks!

PR #9627, opened 2016-10-26
core: Add zipmap interpolation function

jen20:
This commit adds a new interpolation function, `zipmap`, which produces a map given a list of string keys and a list of values of the same length as the list of keys.

### DIFF
--- a/website/source/docs/configuration/interpolation.html.md
+++ b/website/source/docs/configuration/interpolation.html.md
@@ -252,6 +252,13 @@ The supported built-in functions are:
     returned by the `keys` function. This function only works on flat maps and
     will return an error for maps that include nested lists or maps.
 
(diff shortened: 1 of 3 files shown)


website/source/docs/configuration/interpolation.html.md
```diff
@@ -153,6 +153,13 @@ The supported built-in functions are:
       `formatlist("instance %v has private ip %v", aws_instance.foo.*.id, aws_instance.foo.*.private_ip)`.
       Passing lists with different lengths to formatlist results in an error.
 
+  * `formatmap(list, list)` - Creates a map from a list of keys and a list of
```
phinze: Docs need updating to `zipmap`

jen20: :+1: good catch, thanks.

PR #9633, opened 2016-10-26
helper/experiment: a helper for setting, making experiments

mitchellh:
This creates a standard package and interface for defining, querying,
setting experiments (`-X` flags).

I expect we'll want to continue to introduce various features behind
experimental flags. I want to make doing this as easy as possible and I
want to make _removing_ experiments as easy as possible as well.

The goal with this packge has been to rely on the compiler enforcing our
experiment references as much as possible. This means that every
experiment is a global variable that must be referenced directly, so
when it is removed you'll get compiler errors where the experiment is
referenced.

This also unifies and makes it easy to grab CLI flags to enable/disable
experiments as well as env vars! This way defining an experiment is just
a couple lines of code (documented on the package).

### DIFF
--- a/command/apply.go
+++ b/command/apply.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/go-getter"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform/config"
+	"github.com/hashicorp/terraform/helper/experiment"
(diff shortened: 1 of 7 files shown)


jen20:
:+1: This might be a good candidate for extracting into a library for use by other HC projects too at such time as they need experimental features.

mitchellh:
Good idea. Will look at that next :D

Issue #9549, opened 2016-10-24
Outputted map does not contain all expected keys


moio:
### Terraform Version

Latest 0.7.8
### Affected Resource(s)

Core issue. Reproduced with either:
- aws_instance
- [libvirt_volume and libvirt_domain](https://github.com/dmacvicar/terraform-provider-libvirt)
### Terraform Configuration Files
- https://github.com/moio/terraform_output_map_issue_example

Please add proper AWS credentials in https://github.com/moio/terraform_output_map_issue_example/blob/master/main.tf#L2
### Debug Output

https://gist.github.com/moio/2a980e348bccde29be8d13e47821e218
### Expected Behavior

```
terraform_output_map_issue_example$ terraform plan
Refreshing Terraform state in-memory prior to plan...
...
+ module.base.aws_instance.example
...
+ module.extension.aws_instance.example

Plan: 2 to add, 0 to change, 0 to destroy
```
### Actual Behavior

```
terraform_output_map_issue_example$ terraform plan
There are warnings and/or errors related to your configuration. Please
fix these before continuing.

Errors:

  * key "base_instance_id" does not exist in map var.configuration in:

Extension of ${var.configuration["base_instance_id"]}
```
### Steps to Reproduce
- clone repo above
- add AWS credentials
- `terraform plan`
### Important Factoids

I am trying to use a variable from an output of a module in another module.

Note that this does work in case of literals ([`param1` in the sample](https://github.com/moio/terraform_output_map_issue_example/blob/master/main.tf#L9)).

Also note that this does work using the [hacky interpolated map literal syntax](https://github.com/moio/terraform_output_map_issue_example/blob/master/base/main.tf#L15).
### References
- https://github.com/hashicorp/terraform/issues/8384
- https://github.com/hashicorp/terraform/pull/9268

flypenguin:
this is a major blocker in my current efforts to automate our whole vpc setup and I went pretty much crazy debugging my system until I found this. any ideas of when this is fixed?

flypenguin:
BTW #6524 is also applicable to this - the map contents get indented every time I (in that case my editor) uses fmt before saving. pretty annoying :)

mitchellh:
Fixed in master!

moio:
Thanks everybody, great job!

moio:
@mitchellh, @jbardin: JFYI while testing this I found several possibly related failures reported in https://github.com/hashicorp/terraform/issues/10187. Thanks again!

PR #9634, opened 2016-10-26
[WIP]: Add failing test for missing computed map entries

jbardin:
Failing test fixture for #9549 

The map output from the module "mod" loses the computed value from the
template when we validate. If the "extra" field is removed from the map,
the validation fails earlier with map "does not have any elements so
cannot determine type".

Apply will work, because the computed value will exist in the map.


config/raw_config.go
```diff
@@ -127,27 +127,6 @@ func (r *RawConfig) Interpolate(vs map[string]ast.Variable) error {
...
-		if err != nil {
-			return "", err
-		}
-		for _, v := range vars {
-			varVal, ok := vs[v.FullKey()]
-			if ok && varVal.Value == UnknownVariableValue {
-				return UnknownVariableValue, nil
-			}
-		}
-
```
mitchellh: This is the critical section. Because HIL (on `f-unknown`, so travis will fail) handles all this automatically for us now, we don't have to do any special handling. We just evaluate HIL like normal below...

jbardin:
This looks great! I thought I could get this without going into HIL, but ended up going in circles.

Is it possible now that we no longer need the `config.UnknownVariableValue` uuid with TypeUnknown?

If you want to do the HIL PR, I can take care of cleaning up this PR and fixing the vendoring.

mitchellh:
Perfect, made the HIL PR here: https://github.com/hashicorp/hil/pull/31

jbardin:
We might just to do the work to pull the `UnknownVariableValue` out now, as it's leaking into a few tests.

Interestingly, some helper schema tests are failing non-deterministically, with varying diffs and arguments to Schema.Set.

mitchellh:
Picking this up for 0.8, looking into test failures.

PR #9641, opened 2016-10-26
provider/aws: Suceed deleting bucket policy on err

jen20:
If there is no bucket, a bucket policy can be counted as successfully deleted.


stack72:
LGTM!

PR #9642, opened 2016-10-27
command: FlagTypedKV parses bool as string

mitchellh:
When passing a bool type to a variable such as `-var foo=true`, the CLI
would parse this as a `bool` type which Terraform core cannot handle.
It would then error with an invalid type error.

This changes the handling to convert the bool to its literally string
value given on the command-line.

From chat:

2016-10-26 at 9 36 pm


kwilczynski:
I wonder is this would be related to #9647.

mitchellh:
Unfortunately I don't think so :)

PR #9644, opened 2016-10-27
FIX: Private enterprise github references in module source

johnthedev97:
Added a few unit test cases for github detector
Removed the hardcoded domain ‘github.com’ in detectSSH() and set it from
the source
Modified the initial testing logic for HTTPS in Detect() such that
handles domain names where github is a subdomain

PS: The test cases was added by looking at the code and behavior before
my code changes and not from knowledge of use cases/expected values. My
main objective for the test cases was to ensure I don’t break the
behavior regular github.com URLs.

### DIFF
--- a/vendor/github.com/hashicorp/go-getter/detect_github.go
+++ b/vendor/github.com/hashicorp/go-getter/detect_github.go
@@ -15,9 +15,11 @@ func (d *GitHubDetector) Detect(src, _ string) (string, bool, error) {
 		return "", false, nil
 	}
 
-	if strings.HasPrefix(src, "github.com/") {
+	parts := strings.Split(strings.Split(src, "/")[0], ".")
+
+	if parts[0] == "github" && parts[len(parts)-1] == "com" {
 		return d.detectHTTP(src)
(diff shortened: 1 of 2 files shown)


johnthedev97:
I did functional testing only using SSH url for a private enterprise repository as module source. The repository I used has HTTPS access disabled so I couldn't do any functional testing on enterprise github using HTTPS

stack72:
Hi @johnthedev97 

Thanks for the work here! So this is an upstream library that this comes from. We currently use govendor to vendor the versions of libraries. As you have fixed the go files directly under vendor, the next govendor update will replace these changes

IF you can, please can you open the PR on the upstream library - github.com/hashicorp/go-getter

We would then be able to vendor the changes in the normal way :)

Thanks again!

Paul

johnthedev97:
sure, I will do so.

stack72:
thanks so much :)

I will close this one out for now and then we will revendor your changes when they are merged to the upstream lib - which we own btw :)

johnthedev97:
here is the pull request on upstream library https://github.com/hashicorp/go-getter/pull/38

PR #9645, opened 2016-10-27
provider/cloudstack: enhance security groups and rules

svanharmelen:
Based on the earlier work of @mcanevet, so thanks again for that!


svanharmelen:
As both myself and @mcanevet tested this PR and the acceptance test are also all green, I'll go ahead and merge this one in...

PR #9651, opened 2016-10-27
Fixed links to github issues in CHANGELOG.md

antonbabenko:
Run this command:

```bash
sed -i -r -e "s,GH-*([0-9]+),[#\1](https://github.com/hashicorp/terraform/issues/\1),g" CHANGELOG.md
```


catsby:
Hey @antonbabenko – we run  [scripts/changelog-links.sh](https://github.com/hashicorp/terraform/blob/master/scripts/changelog-links.sh) when we release a new version, which is very similar to what you've done here.  It does seem like there are a few in the v0.7.7 release that did not make it. If you were to modify this PR to only apply there I would merge this, but if not then we're going to pass here, as we do so on release.

Thanks!

antonbabenko:
Ok, thanks for letting me know. There were some things missing for the past releases.

PR #9655, opened 2016-10-27
command/show: show state empty if it's empty with empty modules

### LINKED ISSUE #3576: no stdout displayed - terraform show --module-depth=1 with empty statefile
On empty state file (no resources) I get helpful/useful output from terraform show:

```
$ terraform show

module.batch
  0 resource(s)
module.liverecord
  0 resource(s)
module.upload
  0 resource(s)
```

Same situation but with module-depth=1 I get zero useful/helpful info:
```
$ terraform show --module-depth=1

```

Perhaps default to module-depth=0 output when there's a lack of output?

mitchellh:
Fixes #3576

This changes the state.Empty() logic to be a little more robust and
check if modules are empty rather than just checking if there are zero
in the state.

There were some tests in Terraform testing the opposite of this in some
cases and I think they're just wrong and changed them since it caused no
other Context tests to fail. If acc tests pass tonight I think its safe
to say that it doesn't make sense.

### DIFF
--- a/command/format_state.go
+++ b/command/format_state.go
@@ -30,7 +30,7 @@ func FormatState(opts *FormatStateOpts) string {
 	}
 
 	s := opts.State
-	if len(s.Modules) == 0 {
+	if s.Empty() {
 		return "The state file is empty. No resources are represented."
 	}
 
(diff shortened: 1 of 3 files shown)


jbardin:
The tests are failing here, so I'm not sure if you were meaning to check that or not.

mitchellh:
Yeah this has some problems. :) Not merging yet.

mitchellh:
So much has changed. Going to revisit this.

PR #9661, opened 2016-10-27
Bump crypto/ssh package

stack72:
This is the fix up of #9345


jen20:
This gets applied after #9345?

jen20:
Fixes #9345

PR #2562, opened 2015-06-29
Fixed void Azure network config bug.

aznashwan:
Fixes https://github.com/hashicorp/terraform/issues/2560.

Now; creation of all new network elements is safe regardless of a void netconfig on Azure.

PR #2530, opened 2015-06-26
provider/azure: Added SQL Database Server Firewall Rule resource.

aznashwan:
Added the possibility of applying firewall rules to database servers on Azure.

This patch contains all the afferent feature code alongside tests and documentation.

PS: also contains two minor fixes on the secgroup rules and databse service resources.

PPS: Special thanks to @svanharmelen for the good spot.

PR #2515, opened 2015-06-26
provider/azure: Made instances deployable on already existing services.

aznashwan:
Now; instances are deployable on already existing hosted services; else, they default to the creation of a new one with their own name.
The functionality has been added alongside afferent tests and the documentation has been updated.

Also; fixed the indentation in some of the Azure resources documentation.

PR #2394, opened 2015-06-18
provider/azure: Added affinity group resource.

aznashwan:
Added support, tests and docs for Azure affinity groups.

PR #9664, opened 2016-10-27
Fixed comma-separated github issues in CHANGELOG.md

antonbabenko:
This is small follow-up for https://github.com/hashicorp/terraform/pull/9651

It will handle comma-separated links like these:
`[GH-2052, GH-2053, GH-2372, GH-2380, GH-2394, GH-2515, GH-2530, GH-2562]`

/cc @catsby


apparentlymart:
Haha... we had a couple of those in there when I first wrote this but I think I remember just rewriting the changelog to match my regexp rather than the other way around. Nice to be more flexible. :grinning:

PR #9666, opened 2016-10-27
preliminary debug output

jbardin:
This is the first step to get more detailed, and manageable debug info from terraform.  This will write the debug logs and graphs to a tar.gz file in the `DefaultDataDir` (./terraform). This preliminary output is organized by writing a separate file for each step taken, prefixed by a monotonic counter so that each file can be ordered.  Each step always writes a file, even if empty. A dot format graph will be written to the `graphs/` subdirectory for every transformation and walk of the graph. 

To write the debug info, set the `TF_DEBUG` environment variable to a non-empty value.

This also removes the need for the "Graph after ..." text graph representation in the terraform logs, greatly reducing the log output.

DebugInfo will be a global variable through which graph debug
information can be written to a compressed archive. The DebugInfo
methods are all safe for concurrent use, and noop with a nil receiver.
The API outside of the terraform package will be to call SetDebugInfo
to create the archive, and CloseDebugInfo() to properly close the file.
Each write to the archive will be flushed and sync'ed individually, so
in the event of a crash or a missing call to Close, the archive can
still be recovered.

The DebugGraph is a representation of a terraform Graph to be written to
the debug archive, currently in dot format. The DebugGraph also contains
an internal buffer with Printf and Write methods to add to this buffer.
The buffer will be written to an accompanying file in the debug archive
along with the graph.

The DebugHook implements all hooks, and will serialize the arguments and
write them to the debug archive.

This also adds a GraphNodeDebugger interface. Any node implementing
`NodeDebug() string` can output information to annotate the debug graph
node, and add the data to the log. This interface may change or be
removed to provide richer options for debugging graph nodes.

### DIFF
--- a/command/apply.go
+++ b/command/apply.go
@@ -96,6 +96,8 @@ func (c *ApplyCommand) Run(args []string) int {
 		}
 	}
 
+	terraform.SetDebugInfo(DefaultDataDir)
+
 	// Check for the new apply
 	if experiment.Enabled(experiment.X_newApply) && !experiment.Force() {
 		desc := "Experimental new apply graph has been enabled. This may still\n" +
--- a/command/meta.go
+++ b/command/meta.go
@@ -302,12 +302,10 @@ func (m *Meta) Input() bool {
 // context with the settings from this Meta.
 func (m *Meta) contextOpts() *terraform.ContextOpts {
 	var opts terraform.ContextOpts = *m.ContextOpts
-	opts.Hooks = make(
-		[]terraform.Hook,
-		len(m.ContextOpts.Hooks)+len(m.extraHooks)+1)
-	opts.Hooks[0] = m.uiHook()
-	copy(opts.Hooks[1:], m.ContextOpts.Hooks)
-	copy(opts.Hooks[len(m.ContextOpts.Hooks)+1:], m.extraHooks)
+
+	opts.Hooks = []terraform.Hook{m.uiHook(), &terraform.DebugHook{}}
+	opts.Hooks = append(opts.Hooks, m.ContextOpts.Hooks...)
+	opts.Hooks = append(opts.Hooks, m.extraHooks...)
 
 	vs := make(map[string]interface{})
 	for k, v := range opts.Variables {
--- a/command/plan.go
+++ b/command/plan.go
@@ -68,6 +68,12 @@ func (c *PlanCommand) Run(args []string) int {
 		return 1
 	}
 
+	err = terraform.SetDebugInfo(DefaultDataDir)
+	if err != nil {
+		c.Ui.Error(err.Error())
+		return 1
+	}
+
 	if err := ctx.Input(c.InputMode()); err != nil {
 		c.Ui.Error(fmt.Sprintf("Error configuring: %s", err))
 		return 1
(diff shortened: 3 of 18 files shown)


mitchellh:
High level comments:

  * I think the `DebugInfo` and `DebugGraph` should move into a subpackage of `terraform`. They're both fairly TF specific so don't make sense to me to put into `helper` but they also don't need to access any Terraform structs as far as I can tell which is usually a good indicator that they can be extracted. `github.com/hashicorp/terraform/terraform/debug` seems to make sense here.

  * Before we get too invested, could we consider a way for graph nodes to return structured data? (at least `map[string]string`, but ideally `map[string]interface{}` with a contract that it can only be primitives and containers or JSON encode-able or something) Right now it can only return strings and I can see a use case for returning more, potentially. 

  * All exported methods on `DebugInfo` and `DebugGraph` should have documentation on what they do. I don't know what a "phase" is or why I'd set one. What effect does that have? (Don't answer here, answer in the code).

  * Even unexported but significant structs such as `debugInfo` should likely have fairly significant comments. I don't know what the purpose of that is until I read the code more closely.

  * I don't think the `GraphWalker` needs the `Debug` func, it doesn't seem to use it. The function seems to be used to string the DebugGraph into the `Graph.Walk()` function (may be missing something). In this case, I propose we add a `DebugGraph` directly to the `Graph` struct and access it directly. We can do this from the context easily since that is also where we're building the DebugGraph.

  * Can we do the `SetPhase` and defer within `acquireRun` and `releaseRun`? (have acquireRun take an operation name string param). Those must be acquired/deferred for ANY graph operation that Terraform will execute from Context so I think we can remove a lot of duplication and centralize the setup/teardown within there.

Those are some big items for now. I have some bike shedding but we can talk about that maybe just online. If you have any questions ask here or we can talk about it afterwards!

jbardin:
Goods points, I'll try to hit each individually here:

- debug subpackage
Some of my debug experiments were more tightly coupled to terraform. I agree that this looks separable, and if we can keep it that way I do plan to split it out into another package.

- `NodeDebug() string` interface
This is something I think will change. The current signature is simply because any annotation done right now only uses the "comment" attribute in the dot nodes.

- documentation 1, 2
Looking at this I think I lost some in the cherrypicking and rebasing, but it definitely needs a once over with some thorough documentation. 

- GraphWalker.Debug()
I originally had the DebugGraph as a field in Graph, and moved them into the ContextGraphWalker. The primary reason was that it made it easy to differentiate between the shadow operations and the real operations, as there are separate walkers. The second reason was that I though the graph wakers could possibly provide more context later on for added debug information. 

- Yes aquireRun/releaseRun looks like a better place to call SetPhase, missing it was just an oversight.

There's definitely a lot of bike-shedding to do, and lots of this is subject to change ;) I'll let you know when I get the documentation and minor fixes in, and we can go over any other details you think we need to cover to get this first step in place. 

Thanks!

jbardin:
I added documentation, and made some changes based on the comments here, but left the overall structure the same for now.

It was mentioned elsewhere that we would have this include all logging output as well, but that is complicated by the fact that logging is handled out of band by the parent process. We could have the logger look for the TF_DEBUG env var and write the logs alongside the debug archive, and possibly even append those to the archive as part of a cleanup operation on exit, but I think we can address that in a future PR.

PR #9673, opened 2016-10-27
provider/google Support MySQL 5.7 instances

wendorf:
Google Cloud SQL has support for MySQL 5.7, but the option is not documented in the `google_sql_database_instance` resource, and the resource times out when attempting to create. These commits document creating 5.7 and allow it to succeed.


stack72:
Hi @wendorf 

Thanks for the change here - please can you rebase your PR - I just merged your other and it has caused conflicts

thanks

Paul

wendorf:
Rebased. Thanks for addressing these so quickly!

stack72:
LGTM! Thanks :)

PR #9677, opened 2016-10-28
Added s3 and unarchiving docs

jamtur01:
The underlying [go-getter](https://github.com/hashicorp/go-getter)
supports S3 Buckets and unarchiving. Adding mentions of this to the
module sources documentation.


stack72:
LGTM thanks @jamtur01

PR #9683, opened 2016-10-28
AWS EMR Cluster - core_instance_count doesn't actually refer to core instances

ddcprg:
The API will always create 1 master and n-1 core nodes so keep backwards compatibility the logic was changed to submit 1 + core_instance_count nodes instances in the API call.

### DIFF
--- a/builtin/providers/aws/resource_aws_emr_cluster.go
+++ b/builtin/providers/aws/resource_aws_emr_cluster.go
@@ -48,7 +48,7 @@ func resourceAwsEMRCluster() *schema.Resource {
 			"core_instance_count": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  0,
+				Default:  1,
 			},
 			"cluster_state": &schema.Schema{
 				Type:     schema.TypeString,
--- a/website/source/docs/providers/aws/r/emr_cluster.html.md
+++ b/website/source/docs/providers/aws/r/emr_cluster.html.md
@@ -64,16 +64,16 @@ The following arguments are supported:
 * `release_label` - (Required) The release label for the Amazon EMR release
 * `master_instance_type` - (Required) The EC2 instance type of the master node
 * `core_instance_type` - (Optional) The EC2 instance type of the slave nodes
-* `core_instance_count` - (Optional) number of Amazon EC2 instances used to execute the job flow. Default `0`
+* `core_instance_count` - (Optional) Number of Amazon EC2 instances used to execute the job flow. EMR will use one node as the cluster's master node and use the remainder of the nodes (`core_instance_count`-1) as core nodes. Default `1`
 * `log_uri` - (Optional) S3 bucket to write the log files of the job flow. If a value
 	is not provided, logs are not created
 * `applications` - (Optional) A list of applications for the cluster. Valid values are: `Hadoop`, `Hive`,
 	`Mahout`, `Pig`, and `Spark.` Case insensitive
-* `ec2_attributes` - (Optional) attributes for the EC2 instances running the job
+* `ec2_attributes` - (Optional) Attributes for the EC2 instances running the job
 flow. Defined below
-* `bootstrap_action` - (Optional) list of bootstrap actions that will be run before Hadoop is started on
+* `bootstrap_action` - (Optional) List of bootstrap actions that will be run before Hadoop is started on
 	the cluster nodes. Defined below
-* `configurations` - (Optional) list of configurations supplied for the EMR cluster you are creating
+* `configurations` - (Optional) List of configurations supplied for the EMR cluster you are creating
 * `service_role` - (Optional) IAM role that will be assumed by the Amazon EMR service to access AWS resources
 * `visible_to_all_users` - (Optional) Whether the job flow is visible to all IAM users of the AWS account associated with the job flow. Default `true`
 * `tags` - (Optional) list of tags to apply to the EMR Cluster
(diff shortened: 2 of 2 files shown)


ddcprg:
I was wondering whether there is a way to share a set of resources for multiple tests. It seems like a lot of configuration to test an error only.

catsby:
Hey @ddcprg can you help me understand some here? The default of `core_instance_count` right now is zero, and it is not a required attribute. As far as I can tell it's valid to create an EMR cluster like this, though I admit not very useful 😄  

Can you elaborate why this change is needed? 

> I was wondering whether there is a way to share a set of resources for multiple tests. It seems like a lot of configuration to test an error only.

For the sake of testing in isolation we prefer to not share resources, lest a change not actually be detected. We don't mind the redundancy

ddcprg:
Hi @catsby I was trying to create cluster with 0 core instances and got this error:

_* aws_emr_cluster.hive-metastore-service.0: ValidationException: Instance count must be greater than 0._
_	status code: 400, request id: 8911fca5-976f-11e6-a7c2-93ab7141cbe2_

Then I set it to 1 and created a cluster with only the master node.

The [CLI docs state](https://docs.aws.amazon.com/cli/latest/reference/emr/create-cluster.html):

_--instance-count (string)_

_Shortcut option for --instance-groups. A specification of the number of Amazon EC2 instances used together with --instance-type to create instance groups in a cluster. EMR will use one node as the cluster's master node and use the remainder of the nodes as core nodes. Specifying the --instance-type argument without also specifying --instance-count launches a single-node cluster._

Thanks for clarifying the on the tests.

catsby:
Hey @ddcprg – It took me a week or so, but I've reviewed this again! So, I'm thinking this isn't the solution we went, but maybe we can change it to be that. 

I think `core_instance_count` needs to be required and default 1, and we need to change our documentation to be more explicit on what's happening here, specifically that from the `core_instance_count`, AWS will take 1 as the master and create the other (n-1) and make them type `core` instances in the cluster. You then use they `emr_cluster_instance_group` resource to make additional instances of type `task`. 

Does this sound correct and reasonable ?

ddcprg:
Hi @catsby that's fine, the name may still cause confusion but as long as it's well documented it shouldn't be a problem. I'll change the PR to remove my code changes, which are not required going this way, and modify the docs to reflect the behavior in our comment.

ddcprg:
@catsby I've just updated the PR as per your request

catsby:
Excellent @ddcprg , thanks so much! I appreciate the compromise, I'll pull this in 😄

ddcprg:
Cool @catsby ! Thank you!

ddcprg:
Hi @catsby I'm sorry for coming back to this issue. Following this path of thinking, there is still a unexpected behavior in the code: is `core_instance_count` is modified then the resizing will set the CORE instance group to have `core_instance_count` nodes instead of `core_instance_count-1`. This is again a simple change but I'd like to check with you before opening a PR.

The code that needs changing is in `resource_aws_emr_cluster.go`:

```
func resourceAwsEMRClusterUpdate(d *schema.ResourceData, meta interface{}) error {
	conn := meta.(*AWSClient).emrconn

	if d.HasChange("core_instance_count") {
		log.Printf("[DEBUG] Modify EMR cluster")
		groups, err := fetchAllEMRInstanceGroups(meta, d.Id())
		if err != nil {
			log.Printf("[DEBUG] Error finding all instance groups: %s", err)
			return err
		}

		coreInstanceCount := d.Get("core_instance_count").(int)
		coreGroup := findGroup(groups, "CORE")
		if coreGroup == nil {
			return fmt.Errorf("[ERR] Error finding core group")
		}

		params := &emr.ModifyInstanceGroupsInput{
			InstanceGroups: []*emr.InstanceGroupModifyConfig{
				{
					InstanceGroupId: coreGroup.Id,
					InstanceCount:   aws.Int64(int64(coreInstanceCount)),
				},
			},
		}
		_, errModify := conn.ModifyInstanceGroups(params)
		if errModify != nil {
			log.Printf("[ERROR] %s", errModify)
			return errModify
		}

		log.Printf("[DEBUG] Modify EMR Cluster done...")
	}

	log.Println(
		"[INFO] Waiting for EMR Cluster to be available")

	stateConf := &resource.StateChangeConf{
		Pending:    []string{"STARTING", "BOOTSTRAPPING"},
		Target:     []string{"WAITING", "RUNNING"},
		Refresh:    resourceAwsEMRClusterStateRefreshFunc(d, meta),
		Timeout:    40 * time.Minute,
		MinTimeout: 10 * time.Second,
		Delay:      5 * time.Second,
	}

	_, err := stateConf.WaitForState()
	if err != nil {
		return fmt.Errorf("[WARN] Error waiting for EMR Cluster state to be \"WAITING\" or \"RUNNING\" after modification: %s", err)
	}

	return resourceAwsEMRClusterRead(d, meta)
}
```

The code block at line 376 would be:

```
		params := &emr.ModifyInstanceGroupsInput{
			InstanceGroups: []*emr.InstanceGroupModifyConfig{
				{
					InstanceGroupId: coreGroup.Id,
					InstanceCount:   aws.Int64(int64(coreInstanceCount)-1),
				},
			},
		}
```

Please let me know if you'd be happy with this change. Thanks!

catsby:
Hey @ddcprg , that's probably fine, could you submit a PR for this?

massdosage:
@catsby Daniel is away on holiday for the next 2 weeks but I'm sure he'll take a look when he gets back.

ajlanghorn:
Ohhhh! So, this was confusing.

ddcprg:
Hi @catsby I've uploaded the change to this PR: https://github.com/hashicorp/terraform/pull/10529

Issue #9679, opened 2016-10-28
IAM user and group names should allow underscore 


bandesz:
The https://github.com/hashicorp/terraform/pull/9584 changeset adds validation for the IAM user and group name, but the AWS official documentation is not correct and you can create users/groups with underscores in the name. This changeset will probably force some users to change names or not upgrade.

### Terraform Version
Terraform v0.7.8-dev (3500630243a167bf0b88b18bf27a221e2f7c6707)

### Affected Resource(s)
aws_iam_user
aws_iam_group

### Terraform Configuration Files
```hcl
resource "aws_iam_user" "some_user" {
    name = "some_user"
}
resource "aws_iam_group" "some_group" {
    name = "some_group"
}
```

### Expected Behavior
Names with underscores should be allowed. (Even the username "_" is allowed)

### Actual Behavior
Name validation fails.

### Steps to Reproduce
terraform apply

### References
- GH-9584

bandesz:
I submitted a PR for this: https://github.com/hashicorp/terraform/pull/9684

stack72:
Closed via #9684

PR #9684, opened 2016-10-28
Allow underscores in IAM user and group names

bandesz:
This PR solves https://github.com/hashicorp/terraform/issues/9679

My editor also cleaned up a few whitespaces in the docs. If you prefer them removed, please ping me and I update the PR.


kwilczynski:
@bandesz hi there! Thank you for this.

I wonder if we could also mention in the documentation that the name is actually case-insensitive. It might help some people avoid a possible brown-bag moment.

bandesz:
@kwilczynski sure, I added a second commit and copied the relevant parts from the aws cli documentation.

stack72:
LGTM! Thanks for the work here @bandesz :)

PR #9686, opened 2016-10-28
provider/aws: Fix import of RouteTable with destination prefixes 

### LINKED ISSUE #8225: crash when importing aws_route_table

catsby:
Route Tables can get routes that do not have a `DestinationCidrBlock` by way of `aws_vpc_endpoint`. These Routes are ignored in [reading `aws_route_table`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_route_table.go#L158-L162) and so we should ignore them here as well. 

Fixes #8225 , adds test

### DIFF
--- a/builtin/providers/aws/import_aws_route_table.go
+++ b/builtin/providers/aws/import_aws_route_table.go
@@ -40,6 +40,12 @@ func resourceAwsRouteTableImportState(
 				continue
 			}
 
+			if route.DestinationPrefixListId != nil {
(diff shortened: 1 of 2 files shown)


stack72:
LGTM!

PR #9696, opened 2016-10-28
provider/aws: Fixed the SNS password-protected HTTPS endpoints autoconfirm

Ninir:
Hi everyone,

Just suggesting a fix for the SNS autoconfirm for HTTPs endpoints having a Basic authentication. 
The suggested change makes the user-defined endpoint obfuscating the password. 

Related issue:
- https://github.com/hashicorp/terraform/issues/11108

### DIFF
--- a/builtin/providers/aws/resource_aws_sns_topic_subscription.go
+++ b/builtin/providers/aws/resource_aws_sns_topic_subscription.go
@@ -31,40 +31,40 @@ func resourceAwsSnsTopicSubscription() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"protocol": &schema.Schema{
+			"protocol": {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     false,
 				ValidateFunc: validateSNSSubscriptionProtocol,
 			},
-			"endpoint": &schema.Schema{
+			"endpoint": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"endpoint_auto_confirms": &schema.Schema{
+			"endpoint_auto_confirms": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
 			},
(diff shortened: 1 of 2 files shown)


Ninir:
Hi @apparentlymart @stack72 !

Do you think this could go into 0.7.8?

Thank you for your time, hope this is correct and will fit 😊

Ninir:
@catsby Could I get your eyes on this please? Thank you!

Ninir:
Hey @apparentlymart @stack72 

Do you think this could be merged? 😄 

Thanks :)

radeksimko:
Thanks for raising this PR @Ninir 

I think you uncovered a few more issues which we'll need to address, namely

1. If we can't find the subscription we don't error out -> we should.
2. Even when we manage to find the right subscription via ARN I think we'll have another issue with spurious diff because `Read()` will be setting `endpoint` to "obfuscated" version of the URL which won't match with the config/state. I think this is a candidate for `DiffSuppressFunc`. I don't have a valid URL on hand to test and verify this, but I'm fairly confident it's a problem.

For the detection/parsing I think we should use [`net/url`](https://golang.org/pkg/net/url/#Parse) instead of regular expressions. It should provide much cleaner API to work with. The [`url.Userinfo`](https://golang.org/pkg/net/url/#Userinfo) is immutable, but I think it should be easy to reconstruct it and with obfuscated password, modify the `url.URL` and turn it back into string for comparison purposes.

It would be awesome if we had acceptance tests for a confirming http(s) endpoint - I'll see if I can put something together with API Gateway & Lambda.

Feel free to ask if there's anything unclear in the meantime.

Ninir:
Hi @radeksimko 

This was probably my very first contribution... a lot to improve :(
Need to learn about DiffSuppressFunc in details and Go Packages in general.

I think there is a lot to refactor to make this resource even better.

Will work on this... thanks for your inputs, again :) 👍

Issue #8871, opened 2016-09-15
Unnecessary template_file recreation when no template variables have changed


lflux:
Hi there, I'm seeing an issue where a data.template_file thinks it's output has changed and the resource will be recreated on subsequent terraform runs, despite no variable inputs to the template_file resource changing.

In our case, we're using the template output as input to an S3 bucket name. Since `aws_s3_bucket` is using `data.template_file.aws_config_bucket_name.rendered` as the bucket name, and terraform believes the template resource has changed, terraform wants to destroy and recreate the bucket even though the name is exactly the same.

This only happens if template variables are interpolated - i e:

```
data "template_file" "aws_config_bucket_name" {
	template = "aws_config-${var.region}-4711"
}
``` 
will not force recreation, but 

```
data "template_file" "aws_config_bucket_name" {
	template = "aws_config-${region}-4711"
	vars {
		region = "${var.region}"
	}
}
```
will force recreation.

### Terraform Version
0.7.3

### Affected Resource(s)
- data.template_file
- aws_s3_bucket

### Terraform Configuration Files
```hcl
variable "region" {
	default = "us-west-2"
}

variable "account_id" {
	default = "4711"
}

data "template_file" "aws_config_bucket_name" {
	template = "${purpose}-${region}-${account_id}"

	vars {
		region = "${var.region}"
		account_id = "${var.account_id}"
		purpose = "aws-config"
	}
}

resource "aws_s3_bucket" "aws_config" {
	bucket = "${data.template_file.aws_config_bucket_name.rendered}"
	acl = "private"
}
```

### Expected Behavior
Subsequent plans/applies after first application don't want to recreate the bucket 

### Actual Behavior
Terraform plan output:
```
$ terraform plan -out plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but
will not be persisted to local or remote state storage.

aws_s3_bucket.aws_config: Refreshing state... (ID: aws-config-us-west-2-4711)

The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed. Cyan entries are data sources to be read.

Your plan was also saved to the path below. Call the "apply" subcommand
with this plan file and Terraform will exactly execute this execution
plan.

Path: plan

-/+ aws_s3_bucket.aws_config
[...]

Issue #8316, opened 2016-08-19
Data Source template_file causes resource to be recreated each execution.


kristjanelias:
Hi,

I have upgraded to Terraform v0.7.0 and now refactoring some of my code.
Changed template_file resource to data source but now i am seeing that my resource get regenerated each execution. I will try to illustrate my example the best i can for my concearn to come through.
Use-case: Create a reverse lookup zone in aws route53 using a domain name generated within a template.

- Firstly same use-case using **resource** template_file to create a string for the name of a route53 domain:

**Code:**
```
    variable "vpc_cidr_base" { default = "172.30" }

    resource "template_file" "private_reverse_zone_name" {
      template = "${second}.${first}.in-addr.arpa"
      vars {
        first = "${element(split(".", var.vpc_cidr_base), 0)}"
        second = "${element(split(".", var.vpc_cidr_base), 1)}"
      }
    }

    resource "aws_route53_zone" "reverse_private" {
      name = "${template_file.private_reverse_zone_name.rendered}"
      comment = "${var.vpc_cidr_base} - Reverse Zone"
      vpc_id = "${var.aws_default_vpc_id}"
      tags { Name = "${template_file.private_reverse_zone_name.rendered}" }
    }
```
**Statefile after apply:**
```
    "template_file.private_reverse_zone_name": {
        "type": "template_file",
        "primary": {
            "id": "a25c5ce7e9be18bd4e16d6e1b0915542c441a477b309031aab2fc2c8bc00719d",
            "attributes": {
                "id": "a25c5ce7e9be18bd4e16d6e1b0915542c441a477b309031aab2fc2c8bc00719d",
                "rendered": "30.172.in-addr.arpa",
                "template": "${second}.${first}.in-addr.arpa",
                "vars.%": "2",
                "vars.first": "172",
                "vars.second": "30"
            }
        }
    }
```
**Plan after initial apply:**
```
    My-MacBook-Pro:global kristjan$ ./terraform_run.sh plan
    There are warnings and/or errors related to your configuration. Please
    fix these before continuing.

    Warnings:

      * template_file.private_reverse_zone_name: using template_file as a resource is deprecated; consider using the data source instead

    No errors found. Continuing with 1 warning(s).

    Refreshing Terraform state in-memory prior to plan...

    ...

    No changes. Infrastructure is up-to-date. This means that Terraform
    could not detect any differences between your configuration and
    the real physical resources that exist. As a result, Terraform
    doesn't need to do anything.
```

[...]

PR #9698, opened 2016-10-28
terraform: disallow simple variables ("foo")

### LINKED ISSUE #5338: Terraform crash - reporting as instructed by the crash
The console output seems to indicate that it crashed on a syntax error while parsing "${var.apply_immediately}"

Attached is the console output and the crash log

[crash.log.txt](https://github.com/hashicorp/terraform/files/147716/crash.log.txt)
[console.txt](https://github.com/hashicorp/terraform/files/147718/console.txt)

mitchellh:
Fixes #5338 (and I'm sure many others)

There is no use case for "simple" variables in Terraform at all so
anytime one is found it should be an error.

There is a _huge_ backwards incompatibility here that was not supposed
to be by design but I'm sure a lot of people are relying on: in the
`template_file` datasource, this bug allowed you to not escape your
interpolations and have the work. For example:

```
data "template_file" "foo" {
  template = "${a}"
  vars { a = 12 }
}
```

The above would work, but it shouldn't. The template should have to be
`"$${a}"` (to escape the interpolation).

Because of this BC, I recommend holding this until Terraform 0.8.0 and
documenting it carefully. As part of this PR, I've added some special
error message notes.

### DIFF
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -284,12 +284,14 @@ func (i *Interpolater) valueSimpleVar(
 	n string,
 	v *config.SimpleVariable,
 	result map[string]ast.Variable) error {
-	// SimpleVars are never handled by Terraform's interpolator
-	result[n] = ast.Variable{
-		Value: config.UnknownVariableValue,
-		Type:  ast.TypeString,
-	}
-	return nil
+	// This error message includes some information for people who
+	// relied on this for their template_file data sources. We should
+	// remove this at some point but there isn't any rush.
+	return fmt.Errorf(
+		"invalid variable syntax: %q. If this is part of inline `template` parameter\n" +
+			"then you must escape the interpolation with two dollar signs. For\n" +
+			"example: ${a} becomes $${a}." +
+			n)
 }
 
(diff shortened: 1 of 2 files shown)


apparentlymart:
I've seen questions about this particular instance of the bug with `template_file` lots of times on here, so glad to see a fix for it... it has in particular been causing confusion with people using `template_file` as a data resource, since it confuses the logic for deciding if the data source read needs to be deferred until apply time.

This fix looks plausible, but agreed about the BC concerns given how many people reported phantom bugs in the data sources code due to using `template_file` interpolations in this way.

mitchellh:
@apparentlymart Yeah this is going to be paired with: https://github.com/hashicorp/hcl/pull/150

That makes the HCL/HIL interaction a lot more sense at the expense of BC

apparentlymart:
Some more issue references, just to make sure there are link trails from these other issues to here:

* I apparently filed a specific bug report about this and then forgot about it: #8077.
* Various questions and misunderstandings with the `template_file` antipattern shown in this issue's summary: #8021, #8316, #8871.

Issue #1964, opened 2015-05-14
Allow configuring remote storage via config file


radeksimko:
https://www.terraform.io/docs/commands/remote-config.html

It's ok to configure storage via CLI once as one member of the team, but it's pain to do this repeatedly and to share this with team members in a form of a Readme file suggesting "do the following":

```
terraform remote config -backend=S3 -backend-config="bucket=terraform-state" -backend-config="key=tf-state" -backend-config="region=us-east-1"
```

apparentlymart:
I have almost exactly that sprinkled in README files all over our codebases. :)

Just wanted to add that it'd also be useful to be able to use interpolations in a remote config specified inline, so that variables can affect the final location.

Interpolating actual resource attributes (rather than just input variables) would create the interesting situation where the remote state itself depends on some of the resources, but I think supporting interpolation of variables only would be sufficient for my use case. (which is: re-using the same terraform config for a number development environments that differ only in name, while maintaining a separate state for each)

johnrengelman:
Perhaps even a `.terraform.rc` file would work. Then you could override on the command line which rc file to use so your directory could look like:

```
dir/
  .dev.rc
  .prod.rc
  env.tf
```

radeksimko:
@apparentlymart

> Just wanted to add that it'd also be useful to be able to use interpolations in a remote config specified inline, so that variables can affect the final location.

I was thinking about this as well, mostly because of a single variable for the _environment name_, is there any other use-case, that you'd have here?

I'm being more and more convinced that Terraform just needs something like `terraform apply -env=env_name`, `terraform destroy -env=env_name` etc., then we could have a limited interpolation only to this special core variable that you wouldn't be able to define in the code, but only as a CLI flag or ENV variable (i.e. `export TF_env=...`).

@johnrengelman 

> Perhaps even a .terraform.rc file would work.

I believe these are mostly project-level settings, not user-level settings. That said, it's no harm to support this.

johnrengelman:
Lots of tools support an rc type file at the project level: NPM, Bower are just 2 off the top of my head. Call it what you like but I was just thinking a configs type file at the project base dir

radeksimko:
@johnrengelman Ah, right, I see what do you mean. Sure, that could be a solution as well.

apparentlymart:
I actually have two use-cases, which are related and could probably both be forced into the single-variable solution:

[...]

PR #9708, opened 2016-10-29
Cleanup and expansion of the S3 remote state documentation

jamtur01:
In response to comments in #1964.


stack72:
Thanks again! :) Loving these docs updates

PR #9713, opened 2016-10-29
provider/scaleway: govendor integration

nicolai86:
Up until now the Scaleway go SDK was not part of the govendor file. 

This will ease the upgrade path since the provider now follows established practice within terraform.


stack72:
👍 :)

PR #9714, opened 2016-10-29
terraform/core: typo in RawConfig documentation

nicolai86:
Nothing impacting functionality, but documentation should be correct, too.


stack72:
LGTM thanks!

PR #9716, opened 2016-10-29
Update doc: aws_iam_user with force_destroy deletes IAM User Login Profile

minamijoyo:
The aws_iam_user with force_destroy option now deletes IAM User Login Profile.

refs: https://github.com/hashicorp/terraform/pull/9583

But this behavior is not documented.  So I updated the doc.


stack72:
Thanks so much - great addition that I missed

Issue #6991, opened 2016-06-02
Resource does not have attribute 'id' for variable


carlossg:
We are getting this error from time to time, seems a race condition when AWS is slower than usual

```
19:29:48 [security-groups] aws_security_group.marker: Creating...
19:29:48 [security-groups] description:                "" => "Tiger security group"
19:29:48 [security-groups] egress.#:                   "" => "<computed>"
19:29:48 [security-groups] ingress.#:                  "" => "<computed>"
19:29:48 [security-groups] name:                       "" => "pse-integration-marker"
19:29:48 [security-groups] owner_id:                   "" => "<computed>"
19:29:48 [security-groups] tags.#:                     "" => "3"
19:29:48 [security-groups] tags.Name:                  "" => "pse-integration-marker"
19:29:48 [security-groups] tags.cloudbees:pse:cluster: "" => "pse-integration"
19:29:48 [security-groups] tags.tiger:cluster:         "" => "pse-integration"
19:29:48 [security-groups] vpc_id:                     "" => "vpc-9a974bfd"
19:29:49 [security-groups] aws_security_group.marker: Creation complete
19:29:49 [security-groups] Error applying plan:
19:29:49 [security-groups] 
19:29:49 [security-groups] 1 error(s) occurred:
19:29:49 [security-groups] 
19:29:49 [security-groups] * Resource 'aws_security_group.marker' does not have attribute 'id' for variable 'aws_security_group.marker.id'
```

the terraform.tfstate file seems corrupt, with no info about the security group created

```
{
    "version": 1,
    "serial": 0,
    "modules": [
        {
            "path": [
                "root"
            ],
            "outputs": {},
            "resources": {}
        }
    ]
}
```
### Terraform Version

0.6.15
### Affected Resource(s)
- aws_security_group
### Terraform Configuration Files

``` hcl
resource "aws_security_group" "marker" {
    name = "pse-integration-marker"
    description = "Tiger security group"
    tags = {
        Name = "pse-integration-marker"
        "tiger:cluster" = "pse-integration"
        "cloudbees:pse:cluster" = "pse-integration"
    }
    vpc_id = "vpc-9a974bfd"
}
output "marker_security_group" {
    value = "${aws_security_group.marker.id}"
}
```
### References
- Related to #6775, happening when AWS has those eventual consistency issues

carlossg:
Debug log https://gist.github.com/carlossg/a51f2fbbf200efbccc01af2d62f8e739

[...]

PR #9719, opened 2016-10-29
provider/aws: More noisy about aws_security_group consistency issues

apparentlymart:
It appears, based on the report in #6991, that the EC2 API is being inconsistent in reporting that a security group exists shortly after it has been created; we've seen Terraform get past the "Waiting for
Security Group to exist" step but then apparently detect that it's gone again once we get into the Update function.

This is not a fix for the issue, but it does make Terraform complain more loudly should the situation arise:
- If it happens in the `Update` function, as I suspect is the case for bug #6991, we will treat this as a fatal error and bail out with partial state, whereas before Terraform would lose track of the security group altogether. This seems more correct in any case, since it's not the `Update` function's job to refresh the state.
- If it happens in the `Read` function, we will still drop it from the state but we'll first generate a debug log acknowledging that this is happening, to give us more to work with when debugging should this issue
  continue to arise for those who have seen it so far.

### DIFF
--- a/builtin/providers/aws/resource_aws_security_group.go
+++ b/builtin/providers/aws/resource_aws_security_group.go
@@ -284,6 +284,10 @@ func resourceAwsSecurityGroupRead(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 	if sgRaw == nil {
+		log.Printf(
+			"[DEBUG] Security group %q seems to have been deleted; dropping it from state",
+			d.Id(),
+		)
 		d.SetId("")
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


mitchellh:
Even better: we should probably pull the update logic out into its own function that can take a `*ec2.SecurityGroup*` as an arg and have Create/Update call that after querying it themselves.

This will cause the `Create` path to just never have this failure scenario chance, and `Update` should probably continue to drop the security group (after a few, short retries).

apparentlymart:
I'm down with all of that but it still seems weird to me for `Update` to drop the security group... to my mind, in an ideal world it should be a fatal error if `Update` finds *anything* that is inconsistent with what the plan was expecting, since otherwise Terraform isn't being true to what the plan said.

We can't always guarantee that, but I'd consider it better to catch it if we have an opportunity to do so.

mitchellh:
I agree with that!

carlossg:
I have created #11361 based on Mitchell feedback, wdyt @apparentlymart ?

apparentlymart:
Closing this out in favor of #11361.

PR #9720, opened 2016-10-29
provider/aws: IAM policy document: normalize wildcard principals

### LINKED ISSUE #9335: IAM Policy document and * Principals
When trying to specify Principals with `aws_iam_policy_document` I'd do something like

```
principals {
  type = "*"
  identifiers = ["*"]
}
```

This would result as:

```
"Principal": {
  "*": "*"
},
```

In the terraform state.
But whenever you re-run plan/apply it'll show a mismatch and require a change due to AWS converting the `"*": "*"` simply to `"Principal": "*"`
If I were to try with empty principal, the apply would fail with the following message:
`Error putting S3 policy: MalformedPolicy: Missing required field Principal`

apparentlymart:
There are three equivalent forms for expressing "everyone" (including anonymous) in IAM policies:
- `"Principals": "*"`
- `"Principals": {"AWS": "*"}`
- `"Principals": {"*": "*"}`

The more-constrained syntax used by our `aws_iam_policy_document` data source means that the user can only express the latter two of these directly. However, when returning IAM policies from the API AWS likes to normalize to the first form, causing unresolvable diffs.

This fixes #9335 by handling the "everyone" case as a special case, serializing it in JSON as the `"*"`shorthand form.

This change does _not_ address the normalization of hand-written policies containing such elements. A similar change would need to be made in the external package [github.com/jen20/awspolicyequivalence](https://github.com/jen20/awspolicyequivalence) in order to avoid the issue for hand-written policies.

### DIFF
--- a/builtin/providers/aws/data_source_aws_iam_policy_document_test.go
+++ b/builtin/providers/aws/data_source_aws_iam_policy_document_test.go
@@ -106,6 +106,24 @@ data "aws_iam_policy_document" "test" {
         not_resources = ["arn:aws:s3:::*"]
     }
 
+    # Normalization of wildcard principals
+    statement {
+        effect = "Allow"
+        actions = ["kinesis:*"]
+        principals {
+            type = "AWS"
+            identifiers = ["*"]
+        }
+    }
+    statement {
+        effect = "Allow"
+        actions = ["firehose:*"]
+        principals {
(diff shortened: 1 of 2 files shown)


jen20:
:+1: - I will also make the change in AWS Policy Equivalence.

PR #9721, opened 2016-10-29
provider/aws: aws_ami: handle deletion of AMIs

apparentlymart:
Previously this resource (and, by extension, the `aws_ami_copy` and `aws_ami_from_instance` resources that share much of its implementation) was handling correctly the case where an AMI had been recently deregistered, and was thus still returned from the API, but not correctly dealing with the situation where the AMI has been removed altogether.

Now we additionally handle the `NotFound` error returned by the API when we request a non-existent AMI, and remove the AMI from the state in the same way we do for deregistered AMIs.

(I'm sorry to say that I've been sitting on this one for a while and just `-target`ing around the issues this was causing in my environment. :disappointed:)

### DIFF
--- a/builtin/providers/aws/resource_aws_ami.go
+++ b/builtin/providers/aws/resource_aws_ami.go
@@ -121,6 +121,12 @@ func resourceAwsAmiRead(d *schema.ResourceData, meta interface{}) error {
 
 	res, err := client.DescribeImages(req)
 	if err != nil {
+		if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == "InvalidAMIID.NotFound" {
+			log.Printf("[DEBUG] %s no longer exists, so we'll drop it from the state", id)
+			d.SetId("")
+			return nil
+		}
+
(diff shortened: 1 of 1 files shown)


stack72:
His looks good - would the addition of a _disappears style test prove that this works?

apparentlymart:
@stack72 It's not easy to test this in the *usual* way a _disappears test works, because an AMI doesn't get into this state (not existing at all) for some time after it's deregistered; it hangs around in the "deregistered" state for a while first. (This was what caused this bug to not be noticed in the first place.)

I tested this myself here by creating a Terraform state with an intentionally-incorrect AMI id in it and running `terraform refresh`. I could do a similar thing here but I'm not sure there's precedent for an acceptance test to create a "fake state" and then try to do operations on it, and not sure how that would fit in to the `resource.Test` framework. Are there any existing examples of this out there?

stack72:
@apparentlymart makes sense :) In that case, LGTM!

Issue #9678, opened 2016-10-28
Outputs should support descriptions


jamtur01:
### Terraform Version

Feature request.
### Affected Resource(s)

Please list the resources as a list, for example:
- outputs
### Terraform Configuration Files

It'd be great if I could provide a `description` for an output, in the same way I can provide one for a `variable`. This would make module documentation easier and make it easier to treat outputs like outgoing API documentation.

``` hcl
output "public_subnet_id" {
  description = "The AWS subnet ID"
  value = "${aws_subnet.public.id}"
}
```

mgagliardo:
I don't know if this works out for you but my workaround for this is to do the following:
```
//My comments/description here
output "public_subnet_id" {
  value = "${aws_subnet.public.id}"
}
```

And using terraform-docs [Terraform docs](https://github.com/segmentio/terraform-docs):

```
$ terraform-docs md .

## Outputs

| Name | Description |
|------|-------------|
| public_subnet_id |  |
```

It will auto-generate a nice and consistent output for my README.md file. Maybe this will work for you as well.

Regards!

jamtur01:
I use the same work around. But figured I'd ask at the source. :)

mitchellh:
Absolutely, we just talked about this in a meeting yesterday. Tagged.

Ninir:
As per https://github.com/hashicorp/terraform/pull/9722/files, this can be closed now I think.

Did you get a chance to try it @jamtur01?

jamtur01:
Yes - can be closed!

PR #9722, opened 2016-10-30
core: allow outputs to have descriptions

apparentlymart:
We allow variables to have descriptions specified, as additional context for a module user as to what should be provided for a given variable.

We previously lacked a similar mechanism for outputs. Since they too are part of a module's public interface, it makes sense to be able to add descriptions for these for symmetry's sake.

This change makes a `description` attribute valid within an `output` configuration block and stores it within the configuration data structure, but doesn't yet do anything further with it. For now this is useful only for third-party tools that might parse a module's config to generate user documentation; later we could expose the descriptions as part of the `apply` output, but that is left for a separate change.

This could be considered all or part of resolving #9678, depending on whether that issue is interpreted as just allowing descriptions or actually doing something with them in the `apply` output. Either way this is a straightforward change to Terraform's configuration parsing, and we could then have a separate discussion about how (and whether) to present these descriptions in the UI.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -160,10 +160,11 @@ type Variable struct {
 // output marked Sensitive will be output in a masked form following
 // application, but will still be available in state.
 type Output struct {
-	Name      string
-	Sensitive bool
-	DependsOn []string
-	RawConfig *RawConfig
+	Name        string
+	DependsOn   []string
+	Description string
+	Sensitive   bool
+	RawConfig   *RawConfig
 }
 
 // VariableType is the type of value a variable is holding, and returned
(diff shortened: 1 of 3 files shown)


jamtur01:
Awesome. Thanks.

stack72:
@apparentlymart please can you rebase this :)

apparentlymart:
@stack72 was just the classic issue where `go fmt` forced another line into conflict due to alignment spacing. Should be good now, assuming these tests pass.

stack72:
Build is green now! Thanks :) Good addition to 0.8

PR #9726, opened 2016-10-30
Removed redundant type declarations in composite literals

Ninir:
Since Go 1.7+, redundant type declaration in composite literal can be removed.

Thus, this aims to clean every composite literals I could find (using a regex).

Didn't find any similar PR. Sorry for the big diff, hope it is reviewable. Can still make it in multiple commits :smile:.


stack72:
Hi @Ninir 

Thanks for all of the work here - I don't think this is going to be merge-able in it's current format. Merging a PR that touches 1094 files is way too risky. I think the way we were approaching this was to make the changes piece by piece when we were working in an area of the codebase

If you want this work in place, I think a sensible option would be resource by resource so that acceptance tests could be ran and added to the PR with it. FWIW, there is no real gain from the removal of the `&schema.Schema` other than code cleanliness :) So we appreciate the work here

Paul

Ninir:
Hi Paul!
Sure, can understand easily :)

Would you prefer 1 commit per resource or 1 pr for each modified resource?

Thanks!

stack72:
Hi @Ninir 

I think the best way we can handle this type of change - is that as we (as a community) work on a resource, then we can include the removal as part of that :)

As you can see above, the repository changes so often that we will get conflicts

Hope this sounds ok?

Thanks

Paul

Ninir:
Hey Paul.

Can understand, hoped it could fit in another way. 🙂

Thanks for your time.

PR #7542, opened 2016-07-07
Adding result for loadbalancer v2 VipPortID

dkalleg:
This patch adds the VipPortID as a result of creating a lbaas v2
loadbalancer. This can be used to create a FIP.  This has a dependency
on a gophercloud PR: https://github.com/rackspace/gophercloud/pull/606

PR #9727, opened 2016-10-30
provider/openstack: LoadBalancer v2 VIP Port ID

jtopjian:
This commit adds vip_port_id as an exported attribute to the
lb_loadbalancer_v2 resource.

Fixes #7542

/cc @dkalleg

### DIFF
--- a/builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
+++ b/builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
@@ -57,6 +57,11 @@ func resourceLoadBalancerV2() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"vip_port_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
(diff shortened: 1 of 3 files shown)


builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
```diff
@@ -57,6 +57,12 @@ func resourceLoadBalancerV2() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"vip_port_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
stack72: IF this is exported only, then we shouldn't have the `Optional: true` - this allows the user to set it. But we never use the value

jtopjian: Thanks, Paul. I was going to double-check this on Slack :)

So just remove the `Optional`?

stack72:
LGTM! Thanks for the work @jtopjian :)

PR #9734, opened 2016-10-31
provider/aws: Expose ARN suffix on ALB Target Group

### LINKED ISSUE #9593: Create an output for aws_alb_target_group for the arn suffix
Hi,

It is possible to use the arn_suffix output on the aws_alb resource for cloudwatch alarms.  In some of the alarms, I need to use the target group as well, and for this there is no output for the arn_suffix.

Could this be added?

Thanks,
Andrew.
### Terraform Version

Terraform v0.7.7
### Affected Resource(s)
- aws_alb_target_group
- aws_cloudwatch_metric_alarm (requires the arn suffix of the target group).

stack72:
Fixes #9593 

When creating a CloudWatch Metric for an Application Load Balancer Target Group  it is
neccessary to use the suffix of the ARN as the reference to the load
balancer TG . This commit exposes that as an attribute on the `aws_alb_target_group`
resource to prevent the need to use regular expression substitution to
make the reference.

### DIFF
--- a/builtin/providers/aws/resource_aws_alb_target_group.go
+++ b/builtin/providers/aws/resource_aws_alb_target_group.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


firthh:
It looks like `arn_suffix` isn't the correct for what cloudwatch needs

From the tests I can see that the arn_suffix for `arn:aws:elasticloadbalancing:us-east-1:123456:targetgroup/my-targets/73e2d6bc24d8a067` will return `my-targets/73e2d6bc24d8a067`

However, Cloudwatch needs it to return `targetgroup/my-targets/73e2d6bc24d8a067`

PR #9735, opened 2016-10-31
Align output command -help and documentation

carlpett:
`terraform output -help` is missing one paragraph from the web documentation: "_With no additional arguments, output will display all the outputs for the root module._"

That means it says only "_If NAME is not specified, all outputs are printed._", which made me think it would give _all_ outputs, for all modules, which it does not, instead giving me an error message about no outputs being defined (my root module does not have any outputs).


stack72:
LGTM! Thanks :)

PR #9749, opened 2016-10-31
Adding note on remote state

manojlds:


manojlds:
I am using docs as training reference for internal use - it seemed appropriate to mention remote state here, and not just recommend version control for tfstate.

jamtur01:
Ping @jen20. Actually that version control recommendation should be removed.

stack72:
yeah, i'd agree. I think this should be reworded to not include Version Control at all - and suggest remote state from the outset

jen20:
:+1: on not recommending this be version controlled.

manojlds:
Should I reword it? :) I like mentioning that people can start off by version controlling though.

ajlanghorn:
One problem I have with version controlling state is when secrets are stored in it; for instance, the master passphrase associated with an RDS cluster is (or, at least, was!) stored in the state file. Version controlling it is then similar to checking secrets in to a repository, which is dangerous especially due to Git's ability to track changes to a file over time.

jen20:
It's not similar @ajlanghorn - it literally is the same :-) Thats why I'm generally against the suggestion that people version control .tfstate. A better solution now does - remote state - which didn't exist at the time of the original recommendation.

stack72:
Hi @manojlds 

I think, unless there are any objections, that I am going to make the appropriate changes to this PR locally and then merge it

It's been open a while now and it's not fair to go back and forward to ask you to fix it 

Paul

stack72:
Manually merged with the new contents:

```
Terraform also puts some state into the terraform.tfstate file by default. This state file is extremely important; it maps various resource metadata to actual resource IDs so that Terraform knows what it is managing. This file must be saved and distributed to anyone who might run Terraform. It is generally recommended to setup remote state when working with Terraform. This will mean that any potential secrets stored in the state file, will not be checked into version control
```

PR #9753, opened 2016-10-31
provider/aws: aws_ebs_volume datasource

### LINKED ISSUE #9620: new data source "aws_ebs_volume"
For the management of volume attachments to hosts a new data source would help me greatly. My idea is something like an `aws_ebs_volume` with which I could select one or many volumes by tags, e.g. like this:

```
data "aws_ebs_volume" "confluence" {
  single = true  # fail if more than one found

  filter {
    Name = "app-confluence"
  }
}
```

A thing that would also be useful maybe is this:

```
data "aws_ebs_volume" "confluence" {
  single = true     # fail if more than one found
  create = true     # create if not there

  creation_info {
    # basically same information as for aws_ebs_volume
    size = 20

    tags {
      Name = "app-confluence"
    }
  }

  filter {
    Name = "app-confluence"
  }
}
```
### Rationale

This creates a way for selecting a volume based on (for example) tags, so the volume can be used in an automated way, but **can never be deleted** automatically by accident (if you remove the definition from a *.tf file for example).

The auto-create case is just a convenience which saves a manual step, albeit one I would find highly useful. 

The reason behind is that I have stateful applications which are deployed to hosts. My hosts are ephemeral though, and can come and go all the time. I also re-distribute applications across hosts pretty often because our infrastructure is not final yet. So the data of those applications resides on external EBS volumes, which are _highly_ sensitive and should _never_ be deleted. Selecting those data volumes is right now a manual process, although they are heavily tagged.

Managing them with terraform I would like to maybe not do the classical way, because if I accidently delete them from terraform they will be gone in the cloud as well, and we will have data loss. The way with the auto-create would be perfect, but the data source itself would already help greatly.
### Terraform Version

0.7.7 +
### Affected Resource(s)

new resource

see example above

stack72:
This will allows us to filter a specific ebs_volume for attachment to an
aws_instance

fixes #9620

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEbsVolumeDataSource_'✹
2016/11/01 12:39:19 Generated command/internal_plugin_list.go
-run=TestAccAWSEbsVolumeDataSource_ -timeout 120m
=== RUN   TestAccAWSEbsVolumeDataSource_basic
--- PASS: TestAccAWSEbsVolumeDataSource_basic (28.74s)
=== RUN   TestAccAWSEbsVolumeDataSource_multipleFilters
--- PASS: TestAccAWSEbsVolumeDataSource_multipleFilters (28.37s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws57.145s
```

### DIFF
--- a/builtin/providers/aws/data_source_aws_ebs_volume.go
+++ b/builtin/providers/aws/data_source_aws_ebs_volume.go
@@ -1,0 +1,211 @@
+package aws
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"sort"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/davecgh/go-spew/spew"
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsEbsVolume() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsEbsVolumeRead,
+
+		Schema: map[string]*schema.Schema{
+			"filter": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+
+						"values": {
+							Type:     schema.TypeList,
+							Required: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+					},
+				},
+			},
+			"most_recent": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+			},
+			"availability_zone": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"encrypted": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"iops": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"volume_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"size": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"snapshot_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"kms_key_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"volume_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"tags": {
+				Type:     schema.TypeSet,
+				Computed: true,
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/provider.go
```diff
@@ -159,6 +159,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_s3_bucket_object":         dataSourceAwsS3BucketObject(),
 			"aws_subnet":                   dataSourceAwsSubnet(),
 			"aws_security_group":           dataSourceAwsSecurityGroup(),
+			"aws_volume":                   dataSourceAwsVolumes(),
```
apparentlymart: I'm not sure I can really explain why, but it feels weird to me not to name this `aws_ebs_volume`... I guess because it doesn't really feel "specific enough" without it, even though I'd concede that the naming scheme until now has generally been to use generic names for EC2 stuff.

A slightly stronger (but still rather weak) argument would be that when paired with `aws_ebs_snapshot` these would group together nicely in an alphabetical list.

stack72: Agreed - this was only during WIP to make sure it was viable :)

builtin/providers/aws/data_source_aws_ebs_volume.go
```diff
@@ -0,0 +1,216 @@
...
+		} else {
+			return fmt.Errorf("Your query returned more than one result. Please try a more " +
+				"specific search criteria, or set `most_recent` attribute to true.")
+		}
+	} else {
+		// Query returned single result.
+		volume = filteredVolumes[0]
+	}
+
+	log.Printf("[DEBUG] aws_volume - Single Volume found: %s", *volume.VolumeId)
```
jen20: aws_ebs_volume

builtin/providers/aws/data_source_aws_ebs_volume.go
```diff
@@ -0,0 +1,216 @@
...
+	return itime.Unix() < jtime.Unix()
+}
+
+func mostRecentVolume(volumes []*ec2.Volume) *ec2.Volume {
+	sortedVolumes := volumes
+	sort.Sort(volumeSort(sortedVolumes))
+	return sortedVolumes[len(sortedVolumes)-1]
+}
+
+// populate the numerous fields that the image description returns.
```
jen20: Comments no longer relevant?

website/source/docs/providers/aws/d/ebs_volume.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+layout: "aws"
+page_title: "AWS: aws_ebs_volume"
+sidebar_current: "docs-aws-datasource-ebs-volume"
+description: |-
+  Get information on an EBS volume.
+---
+
+# aws\_ebs\_volume
+
+Use this data source to get the information of an EBS volume for use in other
```
jen20: "to get information about"?

PR #9759, opened 2016-10-31
Use Docker-based deploys

sethvargo:


apparentlymart:
This is surprising but fine by me since testing the Terraform website is literally the only thing I use Virtualbox for. :grinning:

mitchellh:
We've been doing it on other projects with great success. LGTM but I'm going to wait until after tomorrow's release so we can continue with the existing process for now, then will merge.

mitchellh:
Okay jk this is an identical release process and way way faster. Thanks!

@apparentlymart Seth has been actually converting a lot of our sites to this process and Terraform is the one that is up. :) Consul, Serf, HashiCorp, and HashiConf already use this same process.

PR #9762, opened 2016-11-01
provider/digitalocean: Update resource docs for website

NicholasPCole:
Specifically:

- User data is available in all regions, so remove the sentence saying to check for supported features in each region.
- Clarify domain vs. DNS record resources.
- Explain which DNS record types the `weight`, `port`, and `priority` arguments are applicable for.


mitchellh:
Looks great. Thanks!

Issue #9590, opened 2016-10-25
allow aws_route53_record to use zone_name instead of zone_id


flypenguin:
The AWS interface seems to require only one of either the zone NAME or the zone ID to create a record ([see here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-hostedzonename))

It would be nice if the AWS resource would have the same choice. I imagine something like described below under "Terraform configuration files"
### Terraform Version

up to 0.7.7
### Affected Resource(s)
- aws_route53_record
### Terraform Configuration Files

``` hcl
# this works
resource "aws_route53_record" "jump" {
  name    = "jump.my-domain.com"
  zone_id = "ABCDEF123"
  type    = "CNAME"
  ttl     = "600"
  records = ["${aws_eip.my_eip.public_ip}"]
}

# this would be nice
resource "aws_route53_record" "jump" {
  name      = "jump.my-domain.com"
  zone_name = "my-domain.com"
  type      = "CNAME"
  ttl       = "600"
  records   = ["${aws_eip.my_eip.public_ip}"]
}
```

kwilczynski:
@flypenguin hi there!

This is a good idea, but a bit tricky to implement. It would require us to list all the hosted zones and then filter out these matching the particular name which would possibly return both the public and private zone IDs should the hosted zone have the same name (which is valid in Route53; and you can have multiple private hosted zones with the same domain name). In other words, it might fail should multiple hosted zones have the same name. Might not be the most ideal behaviour.

I would suggest a data source which allows one to query for a Route53 resources, plus apply filters like e.g. domain name, type (public or private), has a VPC ID assigned, etc. It would then return details about a matching hosted zone where one of them would be the hosted zone ID that could be passed down to the `aws_route53_record` resource.

Would that be something of interest to you?

flypenguin:
sure, basically I want to be able to just use the zone name to set things up. 

but I think AWS *does* all that for you, because the API allows for you to use just the zone name, not the zone ID. (at least that's how I understand the descriptions)

kwilczynski:
@flypenguin hi there again. I had a look at http://docs.aws.amazon.com/sdk-for-go/api/service/route53/ and I could not find a way to pass the zone name only.

CloudFormation, which is what you liked to, most certainly implements this in a bespoke way what is specific to its internal workings, the API (which is what we use via an official SDK), does not have such functionality, unfortunately.

[...]

PR #9766, opened 2016-11-01
provider/aws: data source for AWS Hosted Zone

mathieuherbert:
To add Hosted Zone data source for AWS provider.
```
data "aws_route53_zone" "selected" {
  name = "test.com."
}
```
For example we can do that with this data source 

```
resource "aws_route53_record" "www" {
  zone_id = "${data.aws_route53_zone.selected.zone_id}"
  name = "www.${data.aws_route53_zone.selected.name}"
  type = "A"
  ttl = "300"
  records = ["10.0.0.1"]
}
```

**The acceptance test** 
```
TF_ACC=true go test -v github.com/hashicorp/terraform/builtin/providers/aws -run ^TestAccDataSourceAwsRoute53Zone$
=== RUN   TestAccDataSourceAwsRoute53Zone
--- PASS: TestAccDataSourceAwsRoute53Zone (48.92s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	48.932s
```

### DIFF
--- a/builtin/providers/aws/data_source_aws_route53_zone.go
+++ b/builtin/providers/aws/data_source_aws_route53_zone.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/route53"
@@ -61,7 +62,7 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 		req.DNSName = aws.String(name.(string))
 		req.MaxItems = aws.String("1")
 		resp, err := conn.ListHostedZonesByName(req)
-		name := HostedZoneName(name.(string))
+		name := hostedZoneName(name.(string))
 		if err != nil {
 			errwrap.Wrapf("Error finding Route 53 Hosted Zone: {{err}}", err)
 		}
@@ -71,7 +72,7 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 		}
 		// We test that the first HZ is private or not, if it's not match the field private_zone, we test the second one
 		index := -1
-		if *resp.HostedZones[0].Config.PrivateZone == d.Get("private_zone") {
+		if *resp.HostedZones[0].Config.PrivateZone == d.Get("private_zone").(bool) {
 			index = 0
 		} else if len(resp.HostedZones) >= 2 && *resp.HostedZones[1].Name != name {
 			index = 1
@@ -119,9 +120,8 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 }
 
 // used to manage trailing .
-func HostedZoneName(name string) string {
-	n := len(name)
-	if n == 0 || name[n-1] == '.' {
+func hostedZoneName(name string) string {
+	if strings.HasSuffix(name, ".") {
 		return name
 	} else {
 		return name + "."
--- a/builtin/providers/aws/data_source_aws_route53_zone.go
+++ b/builtin/providers/aws/data_source_aws_route53_zone.go
@@ -40,6 +40,11 @@ func dataSourceAwsRoute53Zone() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"vpc_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
 			"resource_record_set_count": {
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/data_source_aws_route53_zone.go
```diff
@@ -0,0 +1,82 @@
+package aws
+
+import (
+	"fmt"
```
jen20: This should have `goimports` run on it for correct formatting.

jen20: This should have `goimports` run on it for correct formatting.

builtin/providers/aws/data_source_aws_route53_zone.go
```diff
@@ -0,0 +1,82 @@
...
+			return fmt.Errorf("no matching Route53Zone found")
+		}
+		hostedZone := resp.HostedZone
+		id := cleanZoneID(*resp.HostedZone.Id)
+		d.SetId(id)
+		d.Set("zone_id", id)
+		d.Set("name", hostedZone.Name)
+		d.Set("comment", hostedZone.Config.Comment)
+	} else {
+		return fmt.Errorf("name or zone_id have to be setted")
```
jen20: This should read `either name or zone_id must be set`, I think.

builtin/providers/aws/data_source_aws_route53_zone.go
```diff
@@ -0,0 +1,82 @@
...
+		d.Set("comment", hostedZone.Config.Comment)
+
+	} else if idExists {
+		req := &route53.GetHostedZoneInput{}
+		req.Id = aws.String(id.(string))
+
+		resp, err := conn.GetHostedZone(req)
+
+		if err != nil {
+			return err
```
jen20: We should wrap this in `errwrap.Wrapf("Error finding Route 53 Hosted Zone: {{err}}", err)`.

builtin/providers/aws/data_source_aws_route53_zone.go
```diff
@@ -0,0 +1,82 @@
...
+
+		if resp == nil {
+			return fmt.Errorf("no matching Route53Zone found")
+		}
+		hostedZone := resp.HostedZone
+		id := cleanZoneID(*resp.HostedZone.Id)
+		d.SetId(id)
+		d.Set("zone_id", id)
+		d.Set("name", hostedZone.Name)
+		d.Set("comment", hostedZone.Config.Comment)
```
jen20: This will fail - since there is no `comment` field defined in the schema. This should be declared as Computed since it is set here.

kwilczynski:
@mathieuherbert hi there! Thank you for this! This is a highly desirable new data source.

I mean to send a Pull Request which adds such data source, but since you did it already, I will focus on helping you make yours amazing!

The output we would be aiming at is something along the lines of the following:

```
data.aws_route53_hosted_zone.example:
  id = hostedzone-475ac21f
  caller_reference = 2016-10-18T11:43:41.139220387+01:00
  comment = Managed by Terraform
  private_zone = true
  resource_record_set_count = 2
  zone_id = Z24ZQHYYGFIKHW
  zone_name = example.com
  vpc_id = vpc-47ba0223
```

Above shows a Hosted Zone which was found based on zone name and VPC ID, as per:

```
data "aws_route53_hosted_zone" "example" {
  zone_name = "example.com"
  vpc_id = "${aws_vpc.example.id}"
}
```

In other words, for this resource to be very useful, it should support:

- Ability to pass either Hosted Zone ID or name
- Ability to pass VPC ID, to allow for searches involving private zones
- Ability to specify whether we are looking for a Private Zone by setting `private_zone`
attribute accordingly (implied by specifying VPC ID)
- Handle name with and without the trailing dot (e.g. `example.com.`)
- Refresh data should attributes have changed
- Handle the case where the list of Hosted Zones is truncated (and request needs to continue)
- Return meaningful errors to the user

I will be more than happy to help you to get there.

kwilczynski:
Related to #9590.

builtin/providers/aws/data_source_aws_route53_zone.go
```diff
@@ -0,0 +1,129 @@
...
+
+	} else {
+		return fmt.Errorf("Either name or zone_id must be set")
+	}
+
+	return nil
+}
+
+// used to manage trailing .
+func HostedZoneName(name string) string {
```
kwilczynski: This is a good idea to have a helper, but few things to consider:

Have a look at the https://golang.org/pkg/strings/ package which is part of the Go's standard library, it should have the tools you need to add and/or remove a trailing dot.

Also, if you are not planning on making this method accessible outside of this package, it should by a convention (and also language mechanics) start with a lower-case letter, so that the symbol is not going to be exported. Albeit, the `aws` package (namespace) is large at the moment, nevertheless we should follow some common idioms.

website/source/docs/providers/aws/d/hosted_zone.html.markdown
```diff
@@ -0,0 +1,58 @@
...
+sidebar_current: "docs-aws-datasource-hosted-zone"
+description: |-
+    Provides details about a specific Hosted Zone
+---
+
+# aws\_vpc
+
+`aws_hosted_zone` provides details about a specific Hosted Zone.
+
+This resource can be useful when a module accepts a Hosted Zone name as
```
kwilczynski: Should probably say "data source".

website/source/docs/providers/aws/d/hosted_zone.html.markdown
```diff
@@ -0,0 +1,58 @@
...
+an input variable and needs to, for example, add a Record Set in Route 53
```
kwilczynski: I would mention that it allows to find Hosted Zone ID, which is often needed, given a Hosted Zone name and certain search criteria.

mathieuherbert:
Hi,
Thank you for you help, I really appreciate that.
I take your comments and I will put an other commit soon.
I just have an issue with the VPC ID. A private Hosted zone can be associate with multiple VPCs. So do we accept a list in input of the data source and a list also in the output? So the request will match if every VPCs are in hosted zone.
Thank you!

kwilczynski:
@mathieuherbert hi there! You are welcome!

I had a look at the changes and it looks good. I would have few suggestions to consider which might help with the VPC ID logic implementation - especially, since the `ListHostedZones` and `ListHostedZonesByName` do not carry the VPC details in the response data.

The search criteria are either:

- Find Hosted Zone by name or by ID; and
- Filter by whether it is private or not; and
- Filter if associated with a particular VPC.

Also, only a Private Zone can have an association with a VPC, and it can only be associated once with a Hosted Zone that has unique name (domain name, if you wish). It can be associated with multiple Hosted Zones, but they need to have different name (which implies different Hosted Zone altogether; different ID).

Given the above, you could search for:

1) A Public Hosted Zone based on a name or ID; or
2) A Public Hosted Zone based on a name or ID with `private_zone` attribute set; or
3) A Private Hosted Zone based on a name or ID; or
4) A Private Hosted Zone based on a name or ID with `private_zone` attribute set; or
5) A Private Hosted Zone based on a name or ID which is attached to a particular VPC.

Where the (2) and (3) probably makes little sense (especially (2) would yield empty results), rest of them are useful, especially concerning a scenario when people have multiple Private Hosted Zones that have the same name. One could also add region there as the filter to be able to only consider VPCs from current or given regions when looking them up.

This allows for removal of duplication around `nameExists` and `idExists` since a lot of the logic would result in getting a list of the Hosted Zones and then performing search on it to reduce the list to a single zone (the one we want to find).

Note, that you need only to fetch details about each and every zone only when match against a given VPC has to be performed.

I hope this helps! Let me know if you anything else - I am here to help.

kwilczynski:
Adding an example of what happens when you try to associate VPC with a different private hosted zone that has the same name as the one that already has VPC associated with it (since you can have multiple zones having the same name):

screen shot 2016-11-02 at 10 10 23

Thus, for a private zone, its name together with a VPC ID would be enough to identify it.

radeksimko:
Good work so far @mathieuherbert 

I agree with @kwilczynski about the ambiguity of the zone - it may look like a 🐰  hole by now, but we do need to support VPC ID as a field. `private_zone` isn't sufficient.

Also I think we should not just simply take the 1st matching zone - as it may not be what the user wants. We should probably error out if we get `>1` zones from the API and suggest the user to narrow down the search (e.g. by specifying VPC ID or other attributes).

> A private Hosted zone can be associate with multiple VPCs. So do we accept a list in input of the data source and a list also in the output?

As @kwilczynski mentioned above you can have multiple private zones w/ the same name, but each zone would have 1-1 mapping between `zone name - VPC ID` - so there is hopefully no need to accept/output a list.

kwilczynski:
@mathieuherbert @radeksimko just to add, support for searching tags would also be nice (and might help resolve some of the issues around multiple zone with the same name).

mathieuherbert:
Thank you for the help.
I added the support of `vpc_id`. I don't have the time to add tags support today, but I hope to do that tomorrow.
I have a question, do you know why my tests are red (I think it's really simple but I don't understand) ?
```
Unchecked errors found in the following places:
builtin/providers/aws/data_source_aws_route53_zone.go:81:17:	errwrap.Wrapf("Error finding Route 53 Hosted Zone: {{err}}", err)
```
Thank you!

mathieuherbert:
In my last commit I added the support of tags

mathieuherbert:
Hi,
Any comments on my PR ?
Thank you :)

devinsba:
Would be nice to output the NS records for the zone. We would use this for delegating zones from other zones and dns providers

stack72:
Hi @mathieuherbert 

Thanks for all the work here - this LGTM! now :)

Paul

PR #9771, opened 2016-11-01
Update the PowerShell example to set the PATH

alexis-turpin:
`set PATH=%PATH%;C:\terraform` is the old fashioned CMD ways to do which doesn't work in a PowerShell command line.
Moreover, the change made in the CMD console would not be permanent.

The solution proposed here uses .NET Framework’s System.Environment class in PowerShell to properly edit the path.
Source : https://technet.microsoft.com/en-us/library/ff730964.aspx


stack72:
LGTM! thanks :)

PR #9775, opened 2016-11-01
Revert "Bump crypto/ssh package"

stack72:
Reverts hashicorp/terraform#9661

Reverting until after 0.7.8


passcod:
...given it _is_ now after 0.7.8, can that be unreverted?

PR #9776, opened 2016-11-01
terraform: NewComputed doesn't quit Same logic

mitchellh:
For #9618, we added the ability to ignore old diffs that were computed
and removed (because the ultimate value ended up being the same). This
ended up breaking computed list/set logic.

The correct behavior, as is evident by how the other "skip" logics work,
is to set `ok = true` so that the remainder of the logic can run which
handles stuff such as computed lists and sets.


mitchellh:
(Merged after test pass since this was a regression that was blocking release)

PR #9769, opened 2016-11-01
state/remote/swift: Updates

fatmcgav:
Updates to `swift` remote state support:

- Use Gophercloud/gophercloud
- Add support for 'insecure' TLS
- Add Keystone v3 auth values
- Documentation updates

PR #9725, opened 2016-10-30
provider/openstack: Openstack Provider Updates

jtopjian:
This series of commits makes the following changes to the OpenStack Provider configuration:
- Updates provider environment variables
- Makes the password field sensitive
- Adds Descriptions for all provider options
- Cleans up the Provider documentation
- Cleans up the EndpointType check

Fixes #8757
Fixes #8756
Fixes #7041

PR #9777, opened 2016-11-01
state/remote/swift: Enhancements to support Openstack config

fatmcgav:
Enhancements to support full set of Openstack configuration options, plus SSL certs. Documentation updated to support.

Requires #9769 to be merged first.

### DIFF
--- a/state/remote/swift.go
+++ b/state/remote/swift.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"crypto/md5"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"os"
@@ -20,8 +22,21 @@ const TFSTATE_NAME = "tfstate.tf"
 
 // SwiftClient implements the Client interface for an Openstack Swift server.
 type SwiftClient struct {
-	client *gophercloud.ServiceClient
-	path   string
+	client     *gophercloud.ServiceClient
+	authurl    string
+	cacert     string
+	cert       string
+	domainid   string
+	domainname string
+	insecure   bool
+	key        string
+	password   string
+	path       string
+	region     string
+	tenantid   string
+	tenantname string
+	userid     string
+	username   string
 }
 
 func swiftFactory(conf map[string]string) (Client, error) {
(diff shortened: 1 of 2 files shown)


fatmcgav:
@jtopjian This PR includes the 'enhancements' to Swift remote state (5038a57) to bring in-line with latest changes to Openstack provider config (#9725)

state/remote/swift.go
```diff
@@ -32,50 +50,201 @@ func swiftFactory(conf map[string]string) (Client, error) {
...
 	}
-	if val := os.Getenv("OS_USERNAME"); val == "" {
-		return fmt.Errorf("missing OS_USERNAME environment variable")
+	c.password = password
+
+	region, ok := conf["region_name"]
+	if !ok {
+		region = os.Getenv("OS_REGION_NAME")
+		if region == "" {
+			return fmt.Errorf("missing 'region_name' configuration or OS_REGION_NAME environment variable")
```
jtopjian: Region can be optional in some clouds.

state/remote/swift.go
```diff
@@ -32,50 +50,201 @@ func swiftFactory(conf map[string]string) (Client, error) {
...
+	}
+	c.region = region
+
+	tenantID, ok := conf["tenant_id"]
+	if !ok {
+		tenantID = os.Getenv("OS_TENANT_ID")
+		if tenantID == "" {
+			tenantID = os.Getenv("OS_PROJECT_ID")
+		}
+	}
```
jtopjian: Check out [MultiEnvDefaultFunc](https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L195). It might make sense to implement a modified local function in `swift.go` to handle cases where multiple environment variables can be used. Something like:

```go
func multiEnv(ks []string) string {
  for _, k := range ks {           
    if v := os.Getenv(k); v != "" {
      return v                     
    }                              
  }                                
  return ""                        
}                                  
```

jtopjian:
@fatmcgav This looks good! I made a few code comments. I'm not a Go expert, but just some ideas that might tighten things up.

I don't see an issue with breaking backwards compatibility, either. Given Swift remote state was broke for Keystone v3, I think this will just add more functionality. :smile: 

Nice work -- thank you again!

state/remote/swift.go
fatmcgav: Yeh, I did think the code was a bit... verbose there... 

Implementing a `multiEnv` func makes a lot of sense...

state/remote/swift.go
```diff
@@ -32,50 +50,201 @@ func swiftFactory(conf map[string]string) (Client, error) {
...
+		if insecure == "" {
+			// Default if no conf or Env variable set
+			insecure = "false"
+		}
+	}
+	insecureBool, err := strconv.ParseBool(insecure)
+	if err != nil {
+		return fmt.Errorf("non-boolean 'insecure' value")
+	}
+	c.insecure = insecureBool
```
fatmcgav: Yeh, I did look at the S3 bool parsing, but have the added complication here of wanting to pull from either a `conf` value or an env var...

Will see if I can tighten up a bit...

fatmcgav:
@jtopjian OK, I've added a `multiEnv` function, and restructured the `insecure` bool parsing... 

LMK your thoughts...

jtopjian:
@fatmcgav All looks good! I was hoping Github would recognize the previous PR, but that didn't happen... Can you rebase the the previous 3 commits off of master? I'm really sorry about that...

fatmcgav:
@jtopjian heh, no worries. I'll squash the last 2 commits down aswell...

fatmcgav:
@jtopjian All rebased and squashed down :)

PR #9780, opened 2016-11-01
provider/azurerm: Update loadbalancer docs with private ip allocation type.

carinadigital:
`private_ip_address_allocation` was missing from the docs (https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_loadbalancer.go#L71)
There's a few other options not documented there also.


stack72:
LGTM! Thanks :)

PR #9781, opened 2016-11-01
vendor: update HCL for printing updates

mitchellh:


mitchellh:
As long as tests pass here (they should, don't think anything tests the printer directly), going to merge to get it into latest release since it shouldn't affect the active workflow.

PR #9788, opened 2016-11-01
ultradns providers and improvements

josephholsten:
Please note: this was originally developed and tested in an 0.6 branch. Please point out any 0.7isms that should be used.


builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+			},
+			"type": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				// 0-255 char
```
stack72: if this is 0-255 char then we should have a validateFunc to provider faster feedback

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+			},
+			"rdata": &schema.Schema{
+				Type:     schema.TypeSet,
```
stack72: Does this have to be a set or can it be a list? Is there a max size of the items that can be supplied?

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+												},
+												"end": &schema.Schema{
+													Type:     schema.TypeString,
+													Optional: true,
+													// ConflictsWith: []string{"cidr", "address"},
+												},
+												"cidr": &schema.Schema{
+													Type:     schema.TypeString,
+													Optional: true,
+													// ConflictsWith: []string{"start", "end", "address"},
```
stack72: Intentional comments?

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"is_account_level": &schema.Schema{
+										Type:     schema.TypeBool,
+										Optional: true,
+										Default:  false,
+									},
+									"ips": &schema.Schema{
+										Type:     schema.TypeSet,
```
stack72: it may be worth splitting this block out - it's 4 levels deep in the schema and it could be "interesting" in getting / setting the values

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+			"ttl": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  3600,
+			},
+			"conflict_resolve": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "GEO",
+				// Valid: "GEO", "IP"
```
stack72: if there are 2 valid types, it would be worth checking to make sure we only accept those with a validateFunc

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+									"name": &schema.Schema{
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"is_account_level": &schema.Schema{
+										Type:     schema.TypeBool,
+										Optional: true,
+										Default:  false,
+									},
+									"ips": &schema.Schema{
```
stack72: is this the same schema as the `ips` above? If so, this can be extracted to a common schema :) 

See [replication groups](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_elasticache_replication_group.go#L19) for an example - it calls a common schema and then overrides specific values when needed

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+	r, err := makeDirpoolRRSetResource(d)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] ultradns_dirpool create: %#v", r)
+	_, err = client.RRSets.Create(r.RRSetKey(), r.RRSet())
+	if err != nil {
+		// FIXME: remove the json from log
+		marshalled, _ := json.Marshal(r)
```
stack72: Do we care about the json here? We didn't in the log message above

Also, if the unmarshal fails then we have an issue

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+	log.Printf("[INFO] ultradns_dirpool.id: %v", d.Id())
+
+	return resourceUltradnsDirpoolRead(d, meta)
+}
+
+func resourceUltradnsDirpoolRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*udnssdk.Client)
+
+	rr, err := makeDirpoolRRSetResource(d)
+	if err != nil {
```
stack72: Is there a chance that someone can delete the dirpool from the UI? If so, then we will get the equivalent of a not found and fail here. What we do in that case can be seen [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_virtual_machine.go#L555) as an example

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+					d.SetId("")
+					return nil
+				}
+				return fmt.Errorf("resource not found: %v", err)
+			}
+		}
+		return fmt.Errorf("resource not found: %v", err)
+	}
+
+	r := rrsets[0]
```
stack72: Are we sure that the record at index 0 is the one we want? Shouldn't we include some check in the range above to make sure?

builtin/providers/ultradns/resource_ultradns_dirpool.go
```diff
@@ -0,0 +1,612 @@
...
+		res = append(res, m)
+	}
+	return res
+}
+
+// hashIPInfoIPs generates a hashcode for an ip_info.ips block
+func hashIPInfoIPs(v interface{}) int {
+	var buf bytes.Buffer
+	m := v.(map[string]interface{})
+	buf.WriteString(fmt.Sprintf("%s-", m["start"].(string)))
```
stack72: if these are not all required then we should check to make sure there is a value before adding to the hash IMO

builtin/providers/ultradns/resource_ultradns_dirpool.go
josephholsten: this must be a set because we can't trust UltraDNS to respect our ordering

josephholsten: shall I add a comment re this?

josephholsten: resolved

builtin/providers/ultradns/resource_ultradns_dirpool.go
josephholsten: resolved

builtin/providers/ultradns/resource_ultradns_dirpool.go
josephholsten: resolved

builtin/providers/ultradns/resource_ultradns_dirpool.go
josephholsten: resolved, removed the marshalled formatting

stack72:
So @josephholsten and I had a call about this - we are happy that this PR is functional :) Over time, @josephholsten and I will refactor some of the code to use the new constructs like `MaxItems` and `MinItems` and `ValidateFuncs`

We may also change this over time to start support imports - this is further down the line and will require some thought around how the resource IDs :)

PR #9789, opened 2016-11-01
`instance_profile` is a required parameter when creating an EMR instance group

ajlanghorn:
When creating an EC2 instance group to process data in an EMR cluster,
specifying `instance_profile` as an attribute inside `ec2_attributes` is
mandatory, it would appear.

Not setting it results in an error from Terraform complaining that it isn't
set, and the AWS SDK for Go appears - at least in my interpretation - not to
default it anywhere:
https://github.com/aws/aws-sdk-go/blob/master/service/emr/api.go

This PR also includes some basic style and grammatical fixes.

### DIFF
--- a/website/source/docs/providers/aws/r/emr_cluster.html.md
+++ b/website/source/docs/providers/aws/r/emr_cluster.html.md
@@ -64,19 +64,19 @@ The following arguments are supported:
 * `release_label` - (Required) The release label for the Amazon EMR release
 * `master_instance_type` - (Required) The EC2 instance type of the master node
 * `core_instance_type` - (Optional) The EC2 instance type of the slave nodes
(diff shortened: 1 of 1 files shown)


ajlanghorn:
If I rebase this, any chance of a merge? /cc @catsby

catsby:
Please rebase, yes

stack72:
Hi @ajlanghorn 

Please can you rebase this PR and then we can get it merged for you :)

Thanks

Paul

stack72:
Hi @ajlanghorn 

Thanks for the work here - there was already a similar change merged in #10578 that means I can close this - you can see this live on the [site now](https://www.terraform.io/docs/providers/aws/r/emr_cluster.html#instance_profile)

Thanks again

Paul

PR #9800, opened 2016-11-02
Document about ElastiCache Replication Group port

dtan4:
## WHY

ElastiCache Replication Group `port` is _required_ argument, but there is no description about this argument in website.

[AWS: aws_elasticache_replication_group - Terraform by HashiCorp](https://www.terraform.io/docs/providers/aws/r/elasticache_replication_group.html)

## WHAT

Add description of `port` argument in aws_elasticache_replication_group document website.


stack72:
Hi @dtan4 

Thanks for catching this error - you are indeed correct - sorry I missed this when I wrote the resource! I will get this pushed directly to the site now

Paul

PR #9802, opened 2016-11-02
provider/aws: Setting static_routes_only on import of vpn_connection

### LINKED ISSUE #9110: import aws_vpn_connection not setting static_routes_only
Terraform v0.7.4
- aws_vpn_connection

When importing an existing aws_vpn_connection resource with dynamic routing, the static_routes_only attribute is not set. As this attribute is required, the following plan attempts to recreate the resource.

Import successful

```
$ terraform import aws_vpn_connection.vpn_conn vpn-7988fe32
aws_vpn_connection.vpn_conn: Importing from ID "vpn-7988fe32"...
aws_vpn_connection.vpn_conn: Import complete!
  Imported aws_vpn_connection (ID: vpn-7988fe32)
aws_vpn_connection.vpn_conn: Refreshing state... (ID: vpn-7988fe32)

Import success! The resources imported are shown above. These are
now in your Terraform state. Import does not currently generate
configuration, so you must do this next. If you do not create configuration
for the above resources, then the next `terraform plan` will mark
them for destruction.
```

Configuration for the resource

``` hcl
resource "aws_vpn_connection" "vpn_conn" {
    vpn_gateway_id = "${aws_vpn_gateway.dx_gw.id}"
    customer_gateway_id = "${aws_customer_gateway.vpn_gw.id}"
    type = "ipsec.1"
    static_routes_only = false

    tags {
        Name = "backup-vpn${null_resource.interpol_vars.triggers.name_suffix}"
        Terraform = "true" 
    }
}
```

Excerpt from plan

```
-/+ aws_vpn_connection.vpn_conn
...
    static_routes_only:             "" => "false" (forces new resource)
...
```

Worked around by manually adding static_routes_only attribute to terraform state and re-running.

Would expect import to set this attribute.

stack72:
fixes #9110

An error was found where, static_routes_only was not set on a vpn
connection import. This commit introduces setting the static_routes_only
to false when no Options are found. This follows the AWS convention as follows:

```
- options (structure)

Indicates whether the VPN connection requires static routes. If you are creating a VPN connection for a device that does not support BGP, you must specify true .
Default: false

```

So we take it that `static_options_only` is false by default

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSVpnConnection_'
2016/11/02 10:38:18 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSVpnConnection_ -timeout 120m
=== RUN   TestAccAWSVpnConnection_importBasic
--- PASS: TestAccAWSVpnConnection_importBasic (178.29s)
=== RUN   TestAccAWSVpnConnection_basic
--- PASS: TestAccAWSVpnConnection_basic (336.81s)
=== RUN   TestAccAWSVpnConnection_withoutStaticRoutes
--- PASS: TestAccAWSVpnConnection_withoutStaticRoutes (195.45s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	710.572s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_vpn_connection.go
+++ b/builtin/providers/aws/resource_aws_vpn_connection.go
@@ -57,78 +57,79 @@ func resourceAwsVpnConnection() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"vpn_gateway_id": &schema.Schema{
+			"vpn_gateway_id": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"customer_gateway_id": &schema.Schema{
+			"customer_gateway_id": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"type": &schema.Schema{
+			"type": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"static_routes_only": &schema.Schema{
+			"static_routes_only": {
 				Type:     schema.TypeBool,
-				Required: true,
+				Optional: true,
+				Computed: true,
 				ForceNew: true,
 			},
 
 			"tags": tagsSchema(),
 
 			// Begin read only attributes
-			"customer_gateway_configuration": &schema.Schema{
+			"customer_gateway_configuration": {
 				Type:     schema.TypeString,
 				Computed: true,
 				Optional: true,
 			},
 
-			"tunnel1_address": &schema.Schema{
+			"tunnel1_address": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"tunnel1_preshared_key": &schema.Schema{
+			"tunnel1_preshared_key": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"tunnel2_address": &schema.Schema{
+			"tunnel2_address": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"tunnel2_preshared_key": &schema.Schema{
+			"tunnel2_preshared_key": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"routes": &schema.Schema{
+			"routes": {
 				Type:     schema.TypeSet,
 				Computed: true,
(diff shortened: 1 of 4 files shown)


builtin/providers/aws/resource_aws_vpn_connection.go
```diff
@@ -57,78 +58,79 @@ func resourceAwsVpnConnection() *schema.Resource {
...
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"static_routes_only": &schema.Schema{
+			"static_routes_only": {
 				Type:     schema.TypeBool,
-				Required: true,
+				Optional: true,
```
stack72: The reason this changes from Required to Optional is that if we d.Get("") on a bool and it doesn't exist then it falls back to the default value - which is false

builtin/providers/aws/resource_aws_vpn_connection.go
```diff
@@ -300,10 +302,14 @@ func resourceAwsVpnConnectionRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("type", vpnConnection.Type)
 	d.Set("tags", tagsToMap(vpnConnection.Tags))
 
+	log.Printf("[INFO] VPN Connection Options: %s", spew.Sdump(vpnConnection.Options))
```
catsby: `spew`

builtin/providers/aws/resource_aws_vpn_connection.go
catsby: > Required to Optional is that if we d.Get("") on a bool and it doesn't exist 

Why wouldn't it exist if it's a `Required` attribute?

Can you elaborate why this change is needed? It seems that the below change:

```
+	} else {
+		//If there no Options on the connection then we do not support *static_routes*
+		d.Set("static_routes_only", false)
+    }

```

is what is needed here. #9110 states that on import it wasn't being set; with the above, it is. What am I missing such that we now make a once Required attribute to be optional and computed?

catsby: We cleared up my confusion in chat

PR #5621, opened 2016-03-14
provider/aws: send Terraform version in User-Agent

phinze:

Issue #5020, opened 2016-02-05
provider/aws: Use Terraform User-Agent for API requests


radeksimko:
This would be helpful for tracking down API requests and/or even doing some stats from CloudTrail data.

Here's an example of current CloudTrail event coming from Terraform:

```json
{
    "eventVersion": "1.03",
    "userIdentity": {
        "type": "IAMUser",
        "principalId": "AAAAAAAAAAAAAAAA",
        "arn": "arn:aws:iam::111111111111:user/username",
        "accountId": "111111111111",
        "userName": "username"
    },
    "eventTime": "2016-02-05T18:37:55Z",
    "eventSource": "s3.amazonaws.com",
    "eventName": "CreateBucket",
    "awsRegion": "us-east-1",
    "sourceIPAddress": "8.8.8.8",
    "userAgent": "[aws-sdk-go/1.1.0 (go1.5.2; darwin; amd64)]",
    "requestParameters": {
        "x-amz-acl": [
            "private"
        ],
        "bucketName": "my-own-bucket"
    },
    "responseElements": null,
    "requestID": "*REDACTED*",
    "eventID": "*REDACTED*",
    "eventType": "AwsApiCall",
    "recipientAccountId": "111111111111"
}
```

I believe we should replace `aws-sdk-go/1.1.0 (go1.5.2; darwin; amd64)` with Terraform version. I don't mind keeping the AWS-SDK version there, but I think that as long as users use released versions of Terraform, it should be implicitly possible to track down version of the SDK from the Terraform version.

stack72:
I think this is a great idea :)

kapilt:
looks like its just a matter of setting the variable / alternatively setting up a named handler on the session

session construction with
```golang
session.handlers.Build.PushBackNamed(TerraformAgentHandler)

var TerraformAgentHandler = request.NamedHandler{
        Name: "core.SDKVersionUserAgentHandler",
        Fn: request.MakeAddToUserAgentHandler("terraform", "0.6.3")
}
```

phinze:
Done and merged in #5621 :+1:

PR #9806, opened 2016-11-02
provider/aws: Modify AWS User-Agent to new format

catsby:
Modifies the `UserAgent` to match a newly recommended format for tracking Terraform usage in AWS. We do this by changing the string format and removing the SDK version information

Current format:

```
---[ REQUEST POST-SIGN ]-----------------------------
POST / HTTP/1.1
Host: ec2.us-west-2.amazonaws.com
User-Agent: terraform/0.7.9-dev aws-sdk-go/1.4.17 (go1.7.3; darwin; amd64)
```

New format:

```
---[ REQUEST POST-SIGN ]-----------------------------
POST / HTTP/1.1
Host: ec2.us-west-2.amazonaws.com
User-Agent: APN/1.0 HashiCorp/1.0 Terraform/0.7.9-dev
```

With this we should be able to get better information on Terraform usage.

I feel we'll need to note this in the CHANGELOG as a backwards incompatible change however, in the event that someone somewhere is relying on it (see #5020 , #5621 , cc @radeksimko and @phinze  as participants there)

It wasn't clear to me that the SDK version needed to be removed to match this new format, so I did for now. We can re-add it if needed 

------
Update: above I show `Terraform/0.7.9-dev`; this will only apply to custom complied versions of Terraform, say from a branch or `master`. Release versions do not have the `-dev`

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -201,6 +201,10 @@ func (c *Config) Client() (interface{}, error) {
 	if err != nil {
 		return nil, errwrap.Wrapf("Error creating AWS session: {{err}}", err)
 	}
+
+	// Removes the SDK Version handler, so we only have the provider User-Agent
+	// Ex: "User-Agent: APN/1.0 HashiCorp/1.0 Terraform/0.7.9-dev"
+	sess.Handlers.Build.Remove(request.NamedHandler{Name: "core.SDKVersionUserAgentHandler"})
 	sess.Handlers.Build.PushFrontNamed(addTerraformVersionToUserAgent)
 
 	if extraDebug := os.Getenv("TERRAFORM_AWS_AUTHFAILURE_DEBUG"); extraDebug != "" {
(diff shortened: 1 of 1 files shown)


stack72:
LGTM!

mitchellh:
One issue appears to be that the regexp that Amazon sent us noted that the version number can only be two fields, so we may have to change `0.7.9-dev` to `7.9`. I'm going to ask for clarification.

catsby:
We received clarification that the regex used is fine with `0.0.0` so we're going to merge as is.

radeksimko:
This change in general makes sense to me.

Would anything [RFC2616](https://www.ietf.org/rfc/rfc2616.txt)-compatible (`3.8 Product Tokens`) be ok here? If so, then could we mention that RFC in a comment somewhere - just for reference?

What does the `APN` stands for btw.? My Google-fu allowed me to find just _Access Point Name_ and I'm struggling to understand the meaning of it in this context. Is it supposed to help grouping tools like Terraform and allow easy separation from webapps that talk to the AWS API too or something like that?

mitchellh:
@radeksimko Stands for "Amazon Partner Network" afaik :)

mitchellh:
By using this UA, Amazon is able to determine/track how much AWS usage is actually driven by Terraform.

PR #9815, opened 2016-11-02
provider/aws: Fix TestAccAWSAPIGatewayDomainName_basic test

catsby:
The test `TestAccAWSAPIGatewayDomainName_basic` is failing because we're setting a `*time.Time` value to a `String` and it's failing to convert automatically. This PR formats the time to `RFC3339` (see [here for reference](https://golang.org/pkg/time/#pkg-constants))

Refs #9533


stack72:
LGTM!

Ninir:
good catch, thank you @catsby 😳 . Updated my other PRs as well.

PR #9817, opened 2016-11-02
Updated the documentation to reflect that subnets_ids is required

Ninir:
### Reasonning for this change
Redshift Subnet Group Group subnet_ids is a [required argument](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_redshift_subnet_group.go#L42), but it is showed as optional.


stack72:
Nice catch @ninir - LGTM!

Issue #8502, opened 2016-08-26
aws_route53_health_check does not support turning on SNI support via EnableSNI


ziggythehamster:
### Terraform Version

0.7.1 (0.7.2 is also affected if I look at the source code)
### Affected Resource(s)
- aws_route53_health_check
### Expected Behavior

The `aws_route53_health_check` resource should support an `enable_sni` attribute (with possible values `true` and `false`, where `false` is the default). 
### Actual Behavior

The `aws_route53_health_check` resource does not support `enable_sni`, which means that HTTPS/SSL endpoints that require SNI cannot be health checked (they will always return unhealthy) unless you make a change using the AWS API or console yourself.
### Steps to Reproduce
1. Create a health check pointing at a HTTPS endpoint that uses SNI.
2. Observe the health check is always unhealthy.
3. In the console, toggle the "Enable SNI" checkbox to checked.
4. Observe the health check works now.
### Important Factoids

I don't know Go, but this looks fairly simple to add. [The docs](http://docs.aws.amazon.com/Route53/latest/APIReference/API_CreateHealthCheck.html) show that it's `EnableSNI` with values `true|false`.

stack72:
Closed via #9822 :)

PR #9822, opened 2016-11-02
provider/aws: Add the enable_sni attribute for Route53 health checks.

paddyforan:
In #8502 it was requested that we add support for the EnableSNI
parameter of Route53's health checks; this enables customers to
manually specify whether or not the health check will use SNI when
communicating with the endpoint.

The reporter originally requested we default to `false`. While
implementing the issue, I discovered that when creating health
checks with a Type set to HTTP, Amazon's default value for EnableSNI
is `false`. However, when creating health checks with a Type set to
HTTPS, Amazon's default value is `true`. So rather than setting a
default value, I made the attribute computed.


jen20:
Great work @paddyforan!!

PR #9830, opened 2016-11-03
provider/openstack: Don't default 'shared' value, instead only set if…

### LINKED ISSUE #9829: provider/openstack: `openstack_fw_policy_v1` defaulting `shared` results in 400 response
The `openstack_fw_policy_v1` resource defaults `shared` to `false`. 
However `shared` is an 'admin-only' field, as described here: [openstack-api-ref/networking/v2/fwaas#create-firewall-policy](http://developer.openstack.org/api-ref/networking/v2/index.html?expanded=create-firewall-policy-detail) 

Attempting to specify `shared` with a non-admin accounts results in a HTTP 400 response from Neutron. 
> {"NeutronError": "Attribute 'shared' not allowed in POST"}

### Terraform Version
0.7.8-dev

### Affected Resource(s)
-`openstack_fw_policy_v1`

```hcl
resource "openstack_fw_policy_v1" "policy_1" {
  region = "uk-1"
  name = "policy_1"
  rules = [
    "${openstack_fw_rule_v1.allow_icmp.id}",
    "${openstack_fw_rule_v1.allow_ssh.id}"
  ]
}
```

Pull request incoming to resolve.

fatmcgav:
… specified. 
Fixes #9829

### DIFF
--- a/builtin/providers/openstack/resource_openstack_fw_policy_v1.go
+++ b/builtin/providers/openstack/resource_openstack_fw_policy_v1.go
@@ -44,7 +44,6 @@ func resourceFWPolicyV1() *schema.Resource {
 			"shared": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
 			},
 			"tenant_id": &schema.Schema{
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


fatmcgav:
@jtopjian Quick and easy one to fix... :)

PR #9831, opened 2016-11-03
provider/aws: Improve Redshift cluster documentation

Ninir:
This adds missing information in the [Redshift Cluster documentation](https://www.terraform.io/docs/providers/aws/r/redshift_cluster.html), such as:
* cluster_type
* snapshot_identifier
* snapshot_cluster_identifier


stack72:
Hi @Ninir 

Thanks for the work here - why are you changing these values to required? This will break the fact that we can restore from snapshot

Thanks

Paul

Ninir:
Hi Paul,

Sorry it is missing some inputs there. I have updated the PR description. Do you think it should be handled in another way?

Ninir:
@stack72 ,

Reworked it so that it is only about documentation. 
I also found some issues with the Redshift Cluster. Will send a PR for that only.

Tell me if it's missing anything.
Thank you.

stack72:
This is awesome :) Thanks!

PR #9833, opened 2016-11-03
provider/aws: Add aws_eip data source

rhoegg:
This allows users to look up the allocation ID of an Elastic IP given the public IP, or vice versa.


builtin/providers/aws/data_source_aws_eip.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func dataSourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	req := &ec2.DescribeAddressesInput{}
+
+	if id := d.Get("id"); id != "" {
+		req.AllocationIds = []*string{aws.String(id.(string))}
+	}
```
catsby: It seems we only do the lookup if given the `id`, is that correct? I don't see where we attempt to read the `public_ip` and query AWS on that

rhoegg: You're right, and my acceptance test passed.  Looking at it.

rhoegg: I think commit 5490cc8ec0be2ff0243cc6375fa6eab3d5a7d72f took care of this, and made sure the acceptance test would catch it as well.

stack72:
Hi @rhoegg 

Thanks for making the suggested changes! This LGTM now :)

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccDataSourceAwsEip'
2016/12/13 12:06:42 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccDataSourceAwsEip -timeout 120m
=== RUN   TestAccDataSourceAwsEip
--- PASS: TestAccDataSourceAwsEip (19.15s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	19.171s
```

Paul

PR #9834, opened 2016-11-03
provider/openstack: Add 'value_specs' option to 'openstack_fw_rule_v1' resource

fatmcgav:
Refactor to use common 'types.go' and 'MapValueSpecs' function.
Website docs updated.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_fw_rule_v1.go
+++ b/builtin/providers/openstack/resource_openstack_fw_rule_v1.go
@@ -74,6 +74,11 @@ func resourceFWRuleV1() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+			"value_specs": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
+			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


fatmcgav:
@jtopjian First of the `value_spec` additions to the `openstack_fw_*` resources :)

jtopjian:
@fatmcgav The logic for "any" protocol needs to be added to the `ToRuleCreateMap` function:

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/fwaas/rules/requests.go#L113-L124

I wonder if there's a way to trigger the "parent" function rather than copying custom logic like this.

fatmcgav:
@jtopjian Ah, didn't spot that... 

I'll get it added... 
Calling the parent function would be nice, but I'd guess challenging, as I'm already doing a lot of the work in `BuildRequest`...

fatmcgav:
@jtopjian OK, "any" protocol added to `ToRuleCreateMap`. 

I've also checked the other `openstack_fw` resources, and there wasn't any others with custom `CreateMap` functionality... :)

PR #9837, opened 2016-11-03
provider/aws: Redshift Cluster: Fix for master_username & password

Ninir:
This forces the `master_username` & `master_password` to be defined when not in a restore mode. 
If `snapshot_identifier` is not set by the user, those are required. Currently, AWS returns errors about those being not set.

### DIFF
--- a/website/source/docs/providers/aws/r/redshift_cluster.html.markdown
+++ b/website/source/docs/providers/aws/r/redshift_cluster.html.markdown
@@ -34,10 +34,11 @@ string.
  If you do not provide a name, Amazon Redshift will create a default database called `dev`.
 * `node_type` - (Required) The node type to be provisioned for the cluster.
 * `cluster_type` - (Optional) The cluster type to use. Either `single-node` or `multi-node`. 
-* `master_password` - (Required) Password for the master DB user. Note that this may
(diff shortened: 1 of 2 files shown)


website/source/docs/providers/aws/r/redshift_cluster.html.markdown
```diff
@@ -34,10 +34,11 @@ string.
  If you do not provide a name, Amazon Redshift will create a default database called `dev`.
 * `node_type` - (Required) The node type to be provisioned for the cluster.
 * `cluster_type` - (Optional) The cluster type to use. Either `single-node` or `multi-node`. 
-* `master_password` - (Required) Password for the master DB user. Note that this may
-    show up in logs, and it will be stored in the state file. Password must contain at least 8 chars and 
+* `master_password` - (Required unless a `snapshot_identifier` is provided) Password for the master DB user. 
+    Note that this may show up in logs, and it will be stored in the state file. Password must contain at least 8 chars and
```
catsby: `master_password` has `Sensitive: true` set, which should keep it's contents out of the logs. We should optionally pgp encrypt it in the state, like we been doing for some other resources

Issue #5973, opened 2016-04-01
SNS to SQS incompetency and incomplete documentation


jianhuawuchn:
following the URL of 
https://www.terraform.io/docs/providers/aws/r/sns_topic_subscription.html

does not get the job "sending message from SNS to SQS done". As up to this stage, there is no permission set aside in SQS queue which allows to receive message from SNS.

see this link for reference http://docs.aws.amazon.com/sns/latest/dg/SendMessageToSQS.html#SendMessageToSQS.arn

we could use "policy" argument in ws_sqs_queue to achieve our intention, however, since we could not use variable reference in JSON format policy, after setting up SNS and SQS via terraform tool, we had to login to AWS SQS console, and do permission changes on the permission tab.

jianhuawuchn:
the terraform documentation has to be changed, as it doesn't support email protocol in SNS, so we are bound to use SQS. or we need to hard code ARN of working SNS and SQS in to terraform variable file.

stack72:
Hi @jianhuawuchn 

Thanks for reporting the issue here. Apologies that the documentation isn't up to par in this area. Do you know what the changes to the docs should be? I haven't worked with these resources much and would love to be able to get them fixed up. We would love for you to open a PR with the suggested changes to help make it better

Thanks

Paul

jianhuawuchn:
the most easy fix is to ask users to login to their AWS SQS console, and make a subscribe from within SQS console page (this is different from the subscribe on AWS SNS console), after running "terraform apply" command.

If there is an SNS topic applied prior defining SQS queue, you can change 
resource "aws_sqs_queue" "user_updates_queue" {
    name = "user-updates-queue"
}
to
resource "aws_sqs_queue" "user_updates_queue" {
    name = "user-updates-queue"
    policy = {
  "Version": "2012-10-17",
  "Id": "ID_of_SQS",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": "*",
      "Action": "SQS:SendMessage",
      "Resource": "arn_of_sqs,
      "Condition": {
        "ArnEquals": {
          "aws:SourceArn": "arn_of_sns_topic"
        }
      }
    
  ]
}
}
}

however, in reality, most of time, we just use SNS with email protocol. we SQS with other applications like ELK.

Terraform doesn't support SNS with email protocol directly, but we can use aws_cloudformation_stack resource to set up SNS via AWS cloudformation.

jedi4ever:
[...]

PR #9838, opened 2016-11-03
"aws_sns_topic_subscription" add notes and examples for cross account…

AndHei:
… / region SNS topic to SQS queue subscription

Based on #6909 and #5973 I suggest adding information to the documentation about cross account / region SNS topic to SQS queue subscription.


stack72:
LGTM! Nice addition to the docs :)

stack72:
@AndHei does this PR solve those other issues? It feels like they were not a code issue but a documentation issue - thoughts?

AndHei:
@stack72 my understanding is that #6909 and #5973 are rather documentation issues and not code issues.

Issue #9757, opened 2016-10-31
Rollback when using rule resource w/ AWS default network ACL


tylerFowler:
I've found that when using the `aws_default_network_acl` resource alongside the `aws_network_acl_rule` there is a bug that causes the plan to succeed and then roll itself back on the next subsequent `plan`.

Since that's super confusing here's a step by step of what happens:
1. Write rules, link with a default network ACL resource (as opposed to inline rules), run plan & apply and all the rules are correctly written to the default ACL.
2. Run plan again and this time you will get a plan that reverses all the rules you just applied like so:
```bash
~ module.vpc.aws_default_network_acl.default
    egress.#:                     "1" => "0"
    egress.247735553.action:      "deny" => ""
    egress.247735553.cidr_block:  "0.0.0.0/0" => ""
    egress.247735553.from_port:   "2375" => "0"
    egress.247735553.icmp_code:   "0" => "0"
    egress.247735553.icmp_type:   "0" => "0"
    egress.247735553.protocol:    "6" => ""
    egress.247735553.rule_no:     "100" => "0"
    egress.247735553.to_port:     "2376" => "0"
    ingress.#:                    "1" => "0"
    ingress.247735553.action:     "deny" => ""
    ingress.247735553.cidr_block: "0.0.0.0/0" => ""
    ingress.247735553.from_port:  "2375" => "0"
    ingress.247735553.icmp_code:  "0" => "0"
    ingress.247735553.icmp_type:  "0" => "0"
    ingress.247735553.protocol:   "6" => ""
    ingress.247735553.rule_no:    "100" => "0"
    ingress.247735553.to_port:    "2376" => "0"


Plan: 0 to add, 1 to change, 0 to destroy.
```
3. Apply this plan, which succeeds and effectively resets the default ACL to its initial state
4. Run a refresh, now because your tfstate file *does have* those rules but they are *not* applied in AWS you will receive this error:
```bash
Error refreshing state: 2 error(s) occurred:

* aws_network_acl_rule.test_ingress: Expected to find one Network ACL, got: []*ec2.NetworkAcl(nil)
* aws_network_acl_rule.test_egress: Expected to find one Network ACL, got: []*ec2.NetworkAcl(nil)
```

I am able to repeat this behavior reliably, and am assuming that it's not the desired behavior. Also, using inline `ingress` & `egress` blocks fixes this problem.

Terraform Version: `v0.7.7`

catsby:
[...]

PR #9839, opened 2016-11-03
provider/aws: Fix documenation regarding default resources

catsby:
Updates our documentation to note that `aws_default_security_group` and `aws_default_network_acl` are incompatible with their respective `_rule` resources.

refs #9757


stack72:
LGTM! Nice addition to the docs about the diff

PR #9842, opened 2016-11-03
"aws_sns_topic_subscription" add further clarification for cross acco…

AndHei:
…unt SNS topic to SQS queue subscription

based on additional comments in #6909 further clarification was in order.


jen20:
Thanks @AndHei!

PR #9853, opened 2016-11-03
fix CreateBeforeDestroy with datasources

jbardin:
The graph transformation we implement around create_before_destroy
need to re-order all resources that depend on the create_before_destroy
resource. Up until now, we've requires that users mark all of these
resources as create_before_destroy. Data soruces however don't have a
lifecycle block for create_before_destroy, and could not be marked this
way.

This PR checks each DestroyNode that doesn't implement CreateBeforeDestroy
for any ancestors that do implement CreateBeforeDestroy. If there are
any, we inherit the behavior and re-order the graph as such.

### DIFF
--- a/terraform/context_plan_test.go
+++ b/terraform/context_plan_test.go
@@ -2703,3 +2703,39 @@ func TestContext2Plan_moduleVariableFromSplat(t *testing.T) {
 		t.Fatalf("bad:\n%s\n\nexpected\n\n%s", actual, expected)
 	}
 }
+
+func TestContext2Plan_createBeforeDestroy_depends_datasource(t *testing.T) {
+	m := testModule(t, "plan-cdb-depends-datasource")
+	p := testProvider("aws")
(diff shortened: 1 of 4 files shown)


apparentlymart:
Awesome to see this fixed! 😀

It looks like this would also address managed resources that depend on CBS resources, right? There is at least one issue out there for the cycles that causes, and some warnings about it in the docs for CBS.

jbardin:
@apparentlymart,

Yes, I think this should fix many (most, all??!) issues around cycles created from CBD resources. If this does work generally, we won't need the extra documentation and warnings about setting CBD for all dependent resources when you use it in a config, as TF should now properly re-order the the entire dependent destroy sequence. 

I think there's probably a more precise way to do this after the CreateBeforeDestroy transform, by analyzing the subgraph cycles and only re-ordering the minimal set necessary, but this seems like a good first step and a more precise ordering may not be necessary.

PR #9855, opened 2016-11-03
command: show shadow errors to the user

mitchellh:
This builds on #9334 and now exposes shadow errors to the end user. They look like this:

2016-11-03 at 2 47 pm

**A primary goal:** Don't block what would've otherwise been a successful operation on the fact that experimental features failed, and don't confuse the normal UX with experimental failures.

To that end, there are a few important points:

  * The shadow error message is verbose because it is meant to be sufficiently rare that we don't expect the user to know what happened. We assure the user quite a lot in it that everything is okay.

  * Shadow errors only happen if the TF operation succeeds. The user will never be overloaded with a failure and a shadow error.

  * Shadow errors show up currently in `refresh`, `plan`, `apply`, and `destroy`. 

  * Shadow errors can crop up at multiple points since `apply` for example can potentially run a Refresh, Plan, _and_ Apply operation. Shadow errors are collected and stored until the end and only shown if everything succeeds. This improves UX by not interrupted user flow with experimental feature errors when it would've otherwise succeeded.

### DIFF
--- a/command/apply.go
+++ b/command/apply.go
@@ -136,6 +136,9 @@ func (c *ApplyCommand) Run(args []string) int {
 		}
 	}
 
+	// This is going to keep track of shadow errors
+	var shadowErr error
+
 	// Build the context based on the arguments given
 	ctx, planned, err := c.Context(contextOpts{
 		Destroy:     c.Destroy,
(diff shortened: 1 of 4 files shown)


pearkes:
For those reading back to front, can I suggest a final – "Reminder, nothing went wrong!" Type message?

mitchellh:
@pearkes Yep! Fixed

PR #9866, opened 2016-11-04
Service role is not updated on AWS for a CodeDeploy deployment group

bandesz:
Bug description:
I tried to change the service role for a CodeDeploy deployment group. The plan showed the change but it was never applied on AWS.

Affected resources:
 * aws_codedeploy_deployment_group

Affected version: Terraform v0.7.9-dev (24e5ea89521c8af3930c5a87a76c0bdef0037627)

I applied the fix in this PR and it works, but the tests  doesn't seem to check the real changes. Even if I remove my fix the updated acceptance tests still pass. Is it possible these tests only check for internal state changes and don't validate the AWS changes?

I ran the following acceptance tests:
```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCodeDeployDeploymentGroup_basic'
```

P.S.: sorry for the whitespace updates, if requested I remove them.


stack72:
Hi @bandesz 

Is this still a WIP?

bandesz:
@stack72 according to the guideline I was expecting someone to review it properly and respond to my question about the integration tests

stack72:
FYI, I just tested this and it's working as expected :)

I am going to go ahead with the merge

Thanks for all the work here

Paul

Issue #9419, opened 2016-10-17
Feature request: Support for ElastiCache Redis cluster mode


bsiegel:
Amazon recently added support for Redis 3.2 and with it, support for the native Redis Cluster capability, allowing you to create a Redis ElastiCache cluster with 1-15 shards and 0-5 replicas per shard. The AWS documentation now makes the distinction between Memcache, Redis (cluster mode disabled), and Redis (cluster mode enabled). I would enjoy being able to create such a cluster via Terraform.

The announcement is here: https://aws.amazon.com/blogs/aws/amazon-elasticache-for-redis-update-sharded-clusters-engine-improvements-and-more/

Further reading here: 
https://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Replication.Redis-RedisCluster.html
https://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/SelectEngine.Uses.html

jaygorrell:
Is this still being worked on? Would love to see this feature available!

vamsirede:
Is this feature "ElastiCache Redis cluster mode" is available on terraform??

JarenGlover:
+1

carhartl:
According to the cli reference one is able to control the cluster mode via `--cache-parameter-group`, it seems this is already supported?
```
resource "aws_elasticache_cluster" "bar" {
    parameter_group_name = "default.memcached1.4"
}
```
=>
```
resource "aws_elasticache_cluster" "bar" {
    parameter_group_name = "default.redis3.2.cluster.on"
}
```

mblakele:
> According to the cli reference* one is able to control the cluster mode via --cache-parameter-group, it seems this is already supported?

Not quite, but see
https://github.com/hashicorp/terraform/pull/9892 for a candidate implementation.

Would be great to see this as part of terraform 0.9.

rafaelfc-olx:
Any news on this? Any intention of enabling this into terraform 0.8 as well?

deepybee:
It's taken me while to poke around but this is actually already supported in 0.8.8 (possibly earlier):

```
resource "aws_elasticache_replication_group" "foo-replication-group" {
  replication_group_id          = "foo"
  replication_group_description = "Foo replication group"
  node_type                     = "cache.m1.small"
  number_cache_clusters         = 2
  port                          = 6379
  parameter_group_name          = "default.redis3.2.cluster.on"
[...]

PR #9892, opened 2016-11-04
 #9419 - Feature request: Support for ElastiCache Redis cluster mode

dpetzold:
Added support for a new resource aws_elasticache_redis_cluster it will provision
 an ElastiCache Redis cluster with cluster mode enabled. See following for more information:

 https://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Replication.Redis-RedisCluster.html

 It reuses much of aws_elasticache_replication_group resource with the following changes:

 Added support for two new attributes:

   - replicas_per_node_group - (required) the number of replica nodes in each node group
   - num_node_groups - (required) the number of node groups for this Redis replication group

 These attributes have different defaults or functionality:

   - automatic_failover_enabled defaults to true. AWS will fail if its set to false.
   - number_cache_clusters is now a computed field. Its value is:
     ```num_node_groups + num_node_groups * replicas_per_node_group```

Notes:

I added this as a new resource however I would prefer to have it provision a cluster based on a flag:

```cluster_mode_enabled = true``` 

Then replicas_per_node_group and num_node_groups would become required parameters but I did not see how to change the schema after the resource has been called. If you have any ideas on how to do this better please let me know.

@stack72 @catsby

Thank you,

### DIFF
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -233,6 +233,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_eip_association":                          resourceAwsEipAssociation(),
 			"aws_elasticache_cluster":                      resourceAwsElasticacheCluster(),
 			"aws_elasticache_parameter_group":              resourceAwsElasticacheParameterGroup(),
+			"aws_elasticache_redis_cluster":                resourceAwsElasticacheRedisCluster(),
 			"aws_elasticache_replication_group":            resourceAwsElasticacheReplicationGroup(),
 			"aws_elasticache_security_group":               resourceAwsElasticacheSecurityGroup(),
 			"aws_elasticache_subnet_group":                 resourceAwsElasticacheSubnetGroup(),
--- a/builtin/providers/aws/resource_aws_elasticache_redis_cluster.go
+++ b/builtin/providers/aws/resource_aws_elasticache_redis_cluster.go
@@ -1,0 +1,62 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsElasticacheRedisCluster() *schema.Resource {
+
+	resourceSchema := resourceAwsElasticacheReplicationGroupCommon()
+
+	resourceSchema["automatic_failover_enabled"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  true,
+	}
+
+	resourceSchema["number_cache_clusters"] = &schema.Schema{
+		Type:     schema.TypeInt,
+		Computed: true,
+		ForceNew: true,
+	}
+
+	resourceSchema["num_node_groups"] = &schema.Schema{
+		Type:     schema.TypeInt,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by jlory** [comment]
Any news on that pull request? :)

**Comment by alexef** [comment]
Any news in here?

**Comment by stumyp** [comment]
so, I guess I'll need to work around this and use cloudformation for clusters, until this or any other solution is merged.

**Comment by nehalsyed** [comment]
Is there any progress on adding replicas_per_node_group parameter?

**Comment by randahlem** [comment]
What is blocking this from getting merged?

**Comment by catsby** [comment]
Hey @dpetzold I'm sincerely sorry for not getting to this earlier! This is a nice addition we'd love to support.

I'm OK with the PR _as is_, however:

     added this as a new resource however I would prefer to have it provision a cluster based on a flag:

I'm kind of in the same boat here. I _think_ we can make this just be a part of `aws_elasticache_replication_group`. What do you think about an `Optional` attribute called `cluster_mode` that's a `TypeSet` (with `MaxItems: 1`), used like this:

```hcl
resource "aws_elasticache_replication_group" "bar" {
    replication_group_id = "test"
    replication_group_description = "test"
    [...]

    # Becomes Optional
    # number_cache_clusters = 2 
    cluster_mode {
      replicas_per_node_group = 3
      num_node_groups = 3
    }
}
```

I feel like `number_cache_clusters` could then be made `Optional`. We could include logic in the `Create` method to catch if the right combination is not specified, albeit it would be caught at `apply`  time and not `plan` time. That's a downside, but I think there is considerable upside to not mudding the waters with a new resource, when this is really an enhancement to an existing one, made a bit tricky by APIs 😄 

What do you (or anyone else here) think?

Ninir:
Hi @dpetzold 

Would you have time to work on this? I can probably come up with something today, and can augment the work you made here with @catsby 's suggestions.

What do you think?

catsby:
Hey @Ninir the odds of hearing back are probably low, because I let this PR languish here for so long 😦 If you do proceed with that plan, please base your work off of this PR if possible, to preserve the original authors commit even if we undo a lot of it. 

If that's too much I'll understand, but I like to keep attribution when possible!

Ninir:
@catsby Sure, will do that, don't want to get all of the credits on this :)

catsby:
@Ninir I'd say at this point you may proceed, if you're so generous! Please let me know if you don't think you can though, so I can plan accordingly

dpetzold:
Hi! Sorry I have been out of town. I would be happy to followup on this work and would most likely have it done by the end of the week. If you have already started just let me know. Thanks.

catsby:
@dpetzold no worries! You already did a lot, we can't expect any more really 😄 Let us know how it goes!

Ninir:
Hey @dpetzold

I was planning to do it this weekend or so. Feel free to do it :)
And if you can't, will do later.

Thanks for the feedback though!

dpetzold:
Closed in favor of https://github.com/hashicorp/terraform/pull/14262

dpetzold:
https://github.com/hashicorp/terraform/pull/14262 is ready for review. Thank you for the feedback!

PR #6843, opened 2016-05-23
provider/aws: Implementing `vpc_peering_connection_accept`

BSick7:
## Problem

Currently `aws_vpc_peering_connection` provides a way to `auto_accept` a peering connection.  This works great until you have a peering connection between 2 VPCs that are in different aws accounts.  If you specify `auto_accept = true` to automate the peering connection accept, aws will error during a terraform apply that you are not authorized to accept.
## Solution

The solution is to utilize provider aliases to execute the accept under different aws account credentials.  See example below.

```
resource "aws_vpc_peering_connection" "peer1" {
  // use peering connection like normal
  auto_accept = false
}

provider "aws" {
  alias = "other"
  // other account credentials
}

resource "aws_vpc_peering_connection_accept" "peer1" {
  peering_connection_id = "${aws_vpc_peering_connection.peer1.id}"

  provider = "aws.other"
}
```

PR #9901, opened 2016-11-05
aws_vpc_peering_connection ambiguous language

AndHei:
The documentation mentions ownership of both VPCs for aws_vpc_peering_connection auto_accept to work but if both VPC are in separate accounts it does not matter if both account are owned or not.

In #6843 its stated that aws_vpc_peering_connection only works if both VPC are in the same AWS account.

The documentation fails to mention that peeing of two VPCs in two different regions is not supported by AWS.


jen20:
Thanks @AndHei!

PR #9903, opened 2016-11-06
provider/random: Add b64_std to random_id

jen20:
This commit makes three related changes to the `random_id` resource:

1. Deprecate the `b64` attribute.
1. Introduce a new `b64_url` attribut which functions in the same manner as the original `b64` attribute.
3. Introduce a new `b64_std` attribute which uses standard base64 encoding for the value rather than URL encoding.

Resource identifiers continue to use URL encoded base 64.

The reason for adding standard encoding of the base 64 value is to allow the use of generated values as a Serf Encryption Key for separating Consul clusters - these rely on standard encoding and do not permit some characters which are allowed by URL encoding. `b64_url` is introduced in order that there is consistency in specifying the desired encoding during interpolation.

We also clean up the code for the random provider and move to using the more recent schema-provided Noop functions in preference to custom functions in the provider.

cc @sean-, @stack72

### DIFF
--- a/builtin/providers/random/resource_id.go
+++ b/builtin/providers/random/resource_id.go
@@ -4,42 +4,54 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"encoding/hex"
-	"fmt"
+	"errors"
 	"math/big"
 
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
 func resourceId() *schema.Resource {
 	return &schema.Resource{
 		Create: CreateID,
-		Read:   stubRead,
-		Delete: stubDelete,
+		Read:   schema.Noop,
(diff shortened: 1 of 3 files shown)


builtin/providers/random/resource_id.go
```diff
@@ -4,71 +4,88 @@ import (
...
 
-func CreateID(d *schema.ResourceData, meta interface{}) error {
+func CreateID(d *schema.ResourceData, _ interface{}) error {
 
 	byteLength := d.Get("byte_length").(int)
 	bytes := make([]byte, byteLength)
 
 	n, err := rand.Reader.Read(bytes)
 	if n != byteLength {
-		return fmt.Errorf("generated insufficient random bytes")
```
phinze: No need to cover this here, but noting that this catches the global rand source and should probably be reworked to use a local one at some point.

PR #9880, opened 2016-11-04
Fixed typo on interpolation docs

jwieringa:
Removes extra slash in name.

PR #9906, opened 2016-11-07
Update interpolation.html.md

neildothunter:
template interpolation values in example need escaping (example does not work from 0.7.8 onward) https://github.com/hashicorp/terraform/pull/9880


mitchellh:
Missed this :) but we got this done and deployed. Thanks!

PR #9921, opened 2016-11-07
Add base_archive argument to archive_file resource

ponyfleisch:
Using the new argument `base_archive` inside an `archive_file` resource, an existing archive file can be amended.

My primary use case is for AWS lambda functions with a large dependency tree that i would prefer not adding to the repository as individual files. The main lambda function file needs to be run through template_file.

I'm new to golang, let me know if there is anything i can do better.


ponyfleisch:
(sorry, updated request comming)

PR #9924, opened 2016-11-07
Add base_archive option to archive_file resource

ponyfleisch:
Using the new argument `base_archive` inside an `archive_file` resource, an existing archive file can be amended.

My primary use case is for AWS lambda functions with a large dependency tree that i would prefer not adding to the repository as individual files. The main lambda function file needs to be run through template_file.

I'm new to golang, let me know if there is anything i can do better.


ponyfleisch:
@stack72 is there any interest in this? I can resolve the conflicts if need be. It strikes me as a reasonable feature to have when dealing with lambda, even though most of the customizations can now be done using lambda environment variables.

teamterraform:
Hi folks! We are undertaking a general GitHub issue/PR cleanup task now that terraform 0.12 has released. The archive provider now has its own repository (https://github.com/terraform-providers/terraform-provider-archive) and any future issues and PRs should be opened there. Thanks!

PR #9936, opened 2016-11-07
Update documentation for aws_spot_datafeed_subscription.

bljordan:
I've updated the documentation to be more accurate. I've just tested the function of "prefix" with AWS and confirmed it is a path inside an S3 bucket rather than a prefix on filenames.

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-data-feeds.html


catsby:
Thanks!

PR #9942, opened 2016-11-08
provider/google: throw an error for invalid disks

### LINKED ISSUE #6495: Terraform crashes when `disk` is wrongly declared in a `google_compute_instance`
### Terraform Version

Terraform v0.6.15
### Affected Resource(s)
- google_compute_instance
- google_compute_disk

``` hcl
resource "google_compute_instance" "openlava_nfs" {
    name = "${var.name}-nfs"
    machine_type = "${var.nfs_machine_type}"
    zone = "${var.zone}"
    disk {
      image = "${var.image}"
      type = "${var.storage_type}"
    }
    disk {
      disk = "${google_compute_disk.openlava_homes.name}"
      type = "${var.storage_type}"
      device_name = "xvdb"
    }
    disk {
      disk = "${google_compute_disk.openlava_pipelines.name}"
      type = "${var.storage_type}"
      device_name = "xvdc"
    }
    metadata {
      block-project-ssh-keys = "TRUE"
      ssh-keys = "${var.ssh_auth_string}"
    }
    network_interface {
    subnetwork = "${google_compute_subnetwork.private_subnet.name}"
    }
}

resource "google_compute_disk" "openlava_homes" {
    name = "openlava-homes"
    zone = "${var.zone}"
    type = "${var.storage_type}"
    size = "${var.homes_size}"
}

resource "google_compute_disk" "openlava_pipelines" {
    name = "openlava-pipelines"
    zone = "${var.zone}"
    type = "${var.storage_type}"
    size = "${var.pipelines_size}"
}
```
### Panic Output

https://gist.github.com/dvianello/b9b5124398b9ae8064d12b9ec45077aa
### Expected Behavior

Terraform exits with an error stating that `type` cannot be used when declaring a `disk` into a `google_compute_instance` referring to an already created  `google_compute_disk`. I understand that there is no logic in doing that (now!) but Terraform should exit with an error, not crashing. Docs should be updated, too.
### Actual Behavior

Terraform crashes
### Steps to Reproduce
1. `terraform apply`

paddyforan:
When configuring an instance's attached disk, if the attached disk has
both the disk and type attributes set, it would previously cause
terraform to crash with a nil pointer exception. The root cause was that
we only instantiate the InitializeParams property of the disk if its
disk attribute isn't set, and we try to write to the InitializeParams
property when the type attribute is set. So setting both caused the
InitializeParams property to not be initialized, then written to.

Now we throw an error explaining that the configuration can't have both
the disk and the type set.

Fixes #6495.

### DIFF
--- a/builtin/providers/google/resource_compute_instance.go
+++ b/builtin/providers/google/resource_compute_instance.go
@@ -361,6 +361,13 @@ func resourceComputeInstanceCreate(d *schema.ResourceData, meta interface{}) err
 		disk.Boot = i == 0
 		disk.AutoDelete = d.Get(prefix + ".auto_delete").(bool)
 
+		if _, ok := d.GetOk(prefix + ".disk"); ok {
+			if _, ok := d.GetOk(prefix + ".type"); ok {
+				return fmt.Errorf(
+					"Error: cannot define both disk and type.")
+			}
+		}
+
 		// Load up the disk for this disk if specified
 		if v, ok := d.GetOk(prefix + ".disk"); ok {
 			diskName := v.(string)
--- a/builtin/providers/google/resource_compute_instance_test.go
+++ b/builtin/providers/google/resource_compute_instance_test.go
@@ -2,6 +2,7 @@ package google
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


stack72:
nice work! LGTM :)

PR #9943, opened 2016-11-08
provider/openstack: Add Swauth/Swift Authentication

### LINKED ISSUE #9616: provider/openstack: Support Swift Authentication
OpenStack Swift is able to run without Keystone by enabling its own internal authentication.

The OpenStack Provider should be able to support this.

This issue is a result of #8016 and to track the progress of adding this support to the OpenStack provider.

/cc @renier

jtopjian:
This commit adds the ability to authenticate with Swauth/Swift. This can
be used in Swift-only environments that do not have a Keystone service
for authentication.

Fixes #9616

### DIFF
--- a/builtin/providers/openstack/config.go
+++ b/builtin/providers/openstack/config.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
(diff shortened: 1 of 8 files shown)


jtopjian:
@renier I apologize for the delay with this. Please let me know if you have any thoughts about how this is implemented.

PR #9968, opened 2016-11-08
terraform: destroy edges should never point to self

### LINKED ISSUE #9920: Experimental Feature Failure: aws_spot_instance_request destroy
Terraform told me to report this, so here it is.

Terraform version: 0.7.9

### Affected Resource(s)
- aws_spot_instance_request

File used is at: https://github.com/Pinafore/qb/blob/master/aws.tf

### Debug Output

```
Experimental feature failure! Please report a bug.

This is not an error. Your Terraform operation completed successfully.
Your real infrastructure is unaffected by this message.

While running, Terraform sometimes tests experimental features in the
background. These features cannot affect real state and never touch
real infrastructure. If the features work properly, you see nothing.
If the features fail, this message appears.

The following failures happened while running experimental features.
Please report a Terraform bug so that future Terraform versions that
enable these features can be improved!

You can report an issue at: https://github.com/hashicorp/terraform/issues

1 error(s) occurred:

* apply operation: Error building new graph: 1 error(s) occurred:

* Self reference: aws_spot_instance_request.master (destroy)

This is not an error. Your terraform operation completed successfully
and your real infrastructure is unaffected by this message.
```

### Steps to Reproduce
1. `terraform apply`

mitchellh:
Fixes #9920

This was an issue caught with the shadow graph. Self references in
provisioners were causing a self-edge on destroy apply graphs.

We need to explicitly check that we're not creating an edge to ourself.
This is also how the reference transformer works.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -3123,6 +3123,76 @@ func TestContext2Apply_provisionerMultiSelfRefCount(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_provisionerExplicitSelfRef(t *testing.T) {
+	m := testModule(t, "apply-provisioner-explicit-self-ref")
+	p := testProvider("aws")
+	pr := testProvisioner()
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+	pr.ApplyFn = func(rs *InstanceState, c *ResourceConfig) error {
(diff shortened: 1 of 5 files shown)


mitchellh:
Merging so we can get this shadow fix into a 0.7.10. Please review as needed though.

PR #8092, opened 2016-08-10
providers/google: Support IAM permissions for GCP projects

evandbrown:
WIP: needs acceptance tests, more validation, and code review. 

This change adds a data source to allow declaring IAM policies, as well as a
 new resource to represent an existing GCP project. The project resource may reference an IAM policy, allowing a user to set project-wide permissions.

PR #9969, opened 2016-11-08
provider/google: Add google_iam_policy and google_project to sidebar

cblecker:
These resources were created in #8092, but not added to the sidebar. This fixes that so the docs are available on the website.
This may be a candidate for a push to stable-website outside of a release cycle, as these resources are already live.


evandbrown:
Thank you, @cblecker!

Issue #9905, opened 2016-11-06
shadow state and real state diverge when output is removed from module


tkellen:
If, in the course of managing your infrastructure, you remove an output entry from a module, terraform does not purge it from `.tfstate`. This seems to be causing an issue with the new graph builders. This is valid for Terraform v0.7.9.

Here is a reproduction configuration:

```hcl
# ./test/terraform.tf
provider "aws" {
  profile = "<profile>"
  region = "us-east-1"
}

module "vpc" {
  source = "./vpc"
  cidr = "10.0.0.0/16"
}
```

```hcl
# ./test/vpc/main.tf
variable "cidr" { }

output "id" { value = "${aws_vpc.main.id}" }
output "temp" { value = "test" }

resource "aws_vpc" "main" {
  cidr_block = "${var.cidr}"
}
```

Given this, run `terraform apply`, then, remove this line from`./test/vpc/main.tf`
```hcl
output "temp" { value = "test" }
```
...then run `terraform apply` again. Every subsequent apply will show this output:
```
* apply operation: Real and shadow states do not match! Real state:

<no state>
module.vpc:
  aws_vpc.main:
    ID = vpc-9bb155fd
    cidr_block = 10.0.0.0/16
    default_network_acl_id = acl-06ff2560
    default_route_table_id = rtb-eaa2ec8c
    default_security_group_id = sg-b28454cf
    dhcp_options_id = dopt-b69d86d4
    enable_classiclink = false
    enable_dns_hostnames = false
    enable_dns_support = true
    instance_tenancy = default
    main_route_table_id = rtb-eaa2ec8c
    tags.% = 0

  Outputs:

  id = vpc-9bb155fd
  temp = test

Shadow state:

<no state>
module.vpc:
  aws_vpc.main:
    ID = vpc-9bb155fd
    cidr_block = 10.0.0.0/16
    default_network_acl_id = acl-06ff2560
    default_route_table_id = rtb-eaa2ec8c
    default_security_group_id = sg-b28454cf
    dhcp_options_id = dopt-b69d86d4
    enable_classiclink = false
    enable_dns_hostnames = false
    enable_dns_support = true
    instance_tenancy = default
    main_route_table_id = rtb-eaa2ec8c
    tags.% = 0

  Outputs:

  id = vpc-9bb155fd



This is not an error. Your terraform operation completed successfully
and your real infrastructure is unaffected by this message.
```

[...]

PR #9971, opened 2016-11-08
terraform: prune orphan outputs in old graph

mitchellh:
This makes the old graph also prune orphan outputs in modules.
This will fix shadow graph errors such as #9905 since the old graph will
also behave correctly in these scenarios.

Luckily, because orphan outputs don't rely on anything, we were able to
simply use the same transformer!


tkellen:
Nice!

PR #9975, opened 2016-11-08
Fix typo in the Google sidebar.

paddyforan:
We have Google Compute Platform when we really mean Google Cloud
Platform. Whoops. I totally let this sneak in in #9969.


cblecker:
Oops! Thanks for fixing that @paddyforan! My bad!

PR #9990, opened 2016-11-09
provider/aws: Fix panic when passing statuses to aws_acm_certificate

### LINKED ISSUE #9989: panic: aws_acm_certificate
### Terraform Version
v0.7.9

### Affected Resource(s)
- aws_acm_certificate

```hcl
provider "aws" {
  region = "eu-central-1"
}

// the example used here: https://www.terraform.io/docs/providers/aws/d/acm_certificate.html
data "aws_acm_certificate" "example" {
  domain = "tf.example.com"
  statuses = ["ISSUED"]
}
```

### Panic Output
https://gist.github.com/ligustah/fd63a16e80052ec3fff1a43a9377f6f7

### Expected Behavior
It should not panic

### Actual Behavior
It panicked

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
It doesn't appear to require valid AWS credentials, as it crashes before contacting the API.

stack72:
Fixes #9989

When passing a list of statuses to the acm_certificate data source, we
were trying to cast a schema.TypeList directly to []string

We need to do it via an []interface{} and then cast to string when
ranging over the results. Without this, we get a panic

### DIFF
--- a/builtin/providers/aws/data_source_aws_acm_certificate.go
+++ b/builtin/providers/aws/data_source_aws_acm_certificate.go
@@ -40,10 +40,10 @@ func dataSourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) e
 
 	statuses, ok := d.GetOk("statuses")
 	if ok {
-		statusStrings := statuses.([]string)
+		statusStrings := statuses.([]interface{})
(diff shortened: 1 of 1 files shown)


stack72:
Just to reiterate what @catsby and I spoke about - a PR will follow to add a test for this

PR #9992, opened 2016-11-09
Argument for aws_key_pair resource is incorrectly documented as "required"

nicksantamaria:
## Reasoning for docs update 

The [documentation for aws_key_pair](https://www.terraform.io/docs/providers/aws/r/key_pair.html#key_name) resource indicates that the **key_name** argument is required. It is actually optional - omitting that argument will result in an auto-generated key name, prefixed with `terraform-`.

## Relevant Terraform version

Documentation is incorrect on current live site. Behaviour has been tested with Terraform `0.7.9` .


stack72:
Hi @nicksantamaria 

You are indeed correct :) Nice spot! Pushing this to the site now

Paul

PR #9994, opened 2016-11-09
docs: spelling mistake in providers-vault index

SoMuchToGrok:


stack72:
Thanks so much for this :)

PR #10011, opened 2016-11-10
Fix postgres_extension docs

elblivion:


stack72:
Thanks for this @elblivion :)

PR #10015, opened 2016-11-10
Documentation Update: launch configuration for aws provider

dnABic:
Documentation for aws resource aws_launch_configuration in the section Attributes Reference is missing information about attribute "name". But that attribute is used in example on the same document page. Search on the page for:

launch_configuration = "${aws_launch_configuration.as_conf.name}"


stack72:
Hi @dnABic 

Thanks for the PR here - the general structure is as follows:

* Exported attributes are attributes that the user themselves cannot specify

*Any* of the attributes that are accepted as parameters to the resource are actually able to be used :)

I am happy to merge this PR so that the confusion that the docs create goes away a little

Hope this helps? Thanks for the work here

Paul

PR #10026, opened 2016-11-10
Update Azure RM remote config documentation

samirageb:
Current documentation on azure backend examples leaves out a critical parameter required for remote config to work.

### DIFF
--- a/website/source/docs/state/remote/azure.html.md
+++ b/website/source/docs/state/remote/azure.html.md
@@ -21,7 +21,8 @@ terraform remote config \
   -backend=azure \
   -backend-config="storage_account_name=terraform123abc" \
   -backend-config="container_name=terraform-state" \
-  -backend-config="key=prod.terraform.tfstate"
+  -backend-config="key=prod.terraform.tfstate" \
+  -backend-config="access_key=<primary or secondary storage account access key>"
 ```
 
 ## Example Referencing
@@ -34,6 +35,7 @@ data "terraform_remote_state" "foo" {
     storage_account_name = "terraform123abc"
     container_name       = "terraform-state"
     key                  = "prod.terraform.tfstate"
+    access_key           = "<primary or secondary storage account access key>"
   }
 }
 ```
--- a/website/source/docs/state/remote/azure.html.md
+++ b/website/source/docs/state/remote/azure.html.md
@@ -16,16 +16,19 @@ Use of environment variables or config file is recommended.
 
 ## Example Usage
 
+This example follows the recommended approach of storing the access_key in the environment variable `ARM_ACCESS_KEY`.
+
 ```
 terraform remote config \
   -backend=azure \
   -backend-config="storage_account_name=terraform123abc" \
   -backend-config="container_name=terraform-state" \
-  -backend-config="key=prod.terraform.tfstate" \
-  -backend-config="access_key=<primary or secondary storage account access key>"
+  -backend-config="key=prod.terraform.tfstate"
 ```
 
 ## Example Referencing
+
+This example follows the recommended approach of storing the access_key in the environment variable `ARM_ACCESS_KEY`.
 
 ```hcl
 # setup remote state data source


pmcatominey:
Hi @samirageb,

The reason it's left out of the config is described at the top of the document, use of an environment variable is recommended.

samirageb:
@pmcatominey Great. What is the right name for the env VAR? access_key? TF_VAR_access_key?  I personally would prefer to use a config file, but in this case I have no idea how I pass that in. Suggestions?

pmcatominey:
@samirageb it's `ARM_ACCESS_KEY`

samirageb:
@pmcatominey Thank you.  And how could one use a config file?

samirageb:
@pmcatominey Per your clarification, I've amended my PR. The current documentation is definitely lacking clarity around this use case. In general Hashicorp documentation is generally lacking. I do believe this PR will fall inline with the minimalist style to the existing docs.

stack72:
Hi @samirageb 

Thanks for opening the PR here - I am sorry that you feel the documentation is lacking in this area. Documentation is something we continue to try and improve all the time and, thankfully, the community we have help us to improve it

That being said, I think this PR may need a little more work. I think a way to make it more clear would be to edit the NOTE in the file rather than what is proposed here. I suggest something like this:

```
**Note:** Passing credentials directly via config options will
make them included in cleartext inside the persisted state.
Use of environment variables or config file is recommended.
Access key should be passed using the environment variable
`ARM_ACCESS_KEY` to follow this convention.
```

I also suggest we remove this line https://github.com/hashicorp/terraform/pull/10026/files#diff-0ff99894bd73bee9496c27811633e901R41 - as we are suggesting the use of the ENV VAR, the remote state resource actually reads the env var without needing to set that parameter

How does this sound? 

Paul

samirageb:
@stack72 Appending modifying the NOTE field definitely helps. I agree that removing the parameter makes sense if the examples assume use of the ENV VAR. I actually thought I removed the parameter in my update to my PR but perhaps I missed it.

I would also question the suggestion of using a config file as an alternative to setting ENV VARs. Using a config file doesn't appear to be a natively supported by Terraform, and as a new user the 'by Reference' section of the examples is misleading as it looks & smells like a configuration file, but in reality isn't.

I might consider adding some context to the config file recommendation by adding that either Terraform has no native functions to read/parse a config file or offer an example scenario that demonstrates how a config file can be used.

I'm speaking from my own experience in getting up & running on TF and found myself spending far too much time trying to infer the meaning behind this portion of documentation. Does that make sense?

stack72:
ok, comments within:

```
I would also question the suggestion of using a config file as an alternative to setting ENV VARs. Using a config file doesn't appear to be a natively supported by Terraform, and as a new user the 'by Reference' section of the examples is misleading as it looks & smells like a configuration file, but in reality isn't.
```

I agree on the config file - this is already a known issue #1964

Not sure what you mean by the `by reference` section? I am assuming you mean this:

```
# setup remote state data source
data "terraform_remote_state" "foo" {
  backend = "azure"
  config {
    storage_account_name = "terraform123abc"
    container_name       = "terraform-state"
    key                  = "prod.terraform.tfstate"
  }
}
```

If so, this is actually using the remote state as a datasource for use elsewhere - I can suggest the blog post 

```
I'm speaking from my own experience in getting up & running on TF and found myself spending far too much time trying to infer the meaning behind this portion of documentation. Does that make sense?
```

I again, am not sure what you mean here? IS this referring to using remote state? Or documentation in general? 

I am going to take the work you have done here and make the modifications I suggested to come to a happy medium - I agree that we can be better with our documentation. I can also direct you to the [community](https://www.terraform.io/community.html) page of the site - this will give you some links where you can go and ask questions to other members of the community

FYI - change reference is here 24348bad89ba7ec20bc9ed48c82ced9ee4409a21

Thanks again

Paul

samirageb:
@stack72 Thank you for the clarification. You were correct in your assumption when I was referring to the remote state 'reference example' code block. In my case, the confusion came from not understanding that the remote state code block was NOT a valid example of a configuration file alluded to in the best practices annotation earlier on the page.

As you've probably heard many times, the fact that remote state variables have to be entered at the command line and referenced in the .tf file was a serious source of confusion for me.

At any rate, I appreciate the assistance and implore you provide feedback to the Hashicorp team to err on the side of verbosity, specifically adding more common examples.  Again, thank you for the assistance.

stack72:
No worries at all :) We are going to look at how to load remote state variables from config files! We just have to get it prioritised internally

Thanks for getting back to me on this

P.

PR #10030, opened 2016-11-10
debug: next steps

jbardin:
This is some more progress towards detailed graph debugging.

The dot generation code was split between the terraform package, and the dot package. Having another in-memory graph representation in terraform wasn't too useful, so dot formatting has been unified in the dag package. This allows the Graph to serialize itself to dot format, as it already can to a String format.  

A MarshalJSON method is also added to the Graph, so we can serialize out to a more flexible data format. The common intermediary structs used for serialization will allow translation from the JSON output to dot for convenient viewing.

Once the Graph can serialize itself, there's now a path to recording Graph transformations. The Graph.SetDebugWriter method provides an io.Writer for the Graph to record it's initial state, and any subsequent operations and transformations. These are encoded as newline delimited json objects, in the same manner used by json.Encoder. This will provide a way to "replay" changes on a graph, without diff'ing multiple full graphs. This also fully removes the need for the terraform DebugGraph as a separate graph data structure, and we can use the canonical graph for all debugging output.

Finally we add AnnotateVertex and AnnotateEdge methods to the Graph itself, which will add metadata structures to the encoded JSON stream, allowing us to correlate the information when decoding the debug information.


mitchellh:
Looks great! 

The only concern I have is the annotation API. We plan on actually having actual "annotations" that Terraform users can provide on resources and various things. Since the annotations you have in this PR are _only_ related and used for debugging, could the API better reflect that?

I'm not sure what that looks like exactly but maybe `DebugAnnotateVertex` or something like that.

jbardin:
Changed the annotation methods to `VertexDebugInfo` and `EdgeDebugInfo` to avoid overloading "annotations" further in the future, and rebased against master.

mitchellh:
:+1:

PR #10034, opened 2016-11-11
terraform: when returning a raw attribute value, use hil conversion

mitchellh:
Because we now rely on HIL to do the computed calculation, we must make
sure the type is correct (TypeUnknown). Before, we'd just check for the
UUID in the string.

This changes all variable returns in the interpolater to run it through
`hil.InterfaceToVariable` which handles this lookup for us.

### DIFF
--- a/helper/schema/field_reader_config_test.go
+++ b/helper/schema/field_reader_config_test.go
@@ -1,11 +1,14 @@
 package schema
 
 import (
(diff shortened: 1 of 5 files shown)


mitchellh:
Merging since these are causing acc tests failures and I want to check on them tonight.

PR #10036, opened 2016-11-11
terraform: destroy graph must connect edges for proper target ordering

mitchellh:
This connects the destroy edges so that when a `-target` is specified on
a destroy, the proper dependencies get destroyed as well.

The `command` package actually caught this and I added a terraform 
context test as well.


mitchellh:
Merging since tests just failed before this.

PR #10037, opened 2016-11-11
terraform: default new graphs on, old graphs behind -Xlegacy-graph

mitchellh:
This turns the new graphs on by default and puts the old graphs behind a
flag `-Xlegacy-graph`. This effectively inverts the current 0.7.x
behavior with the new graphs.

We've incubated most of these for a few weeks now. We've found issues
and we've fixed them and we've been using these graphs internally for
awhile without any major issue. Its time to default them on and get them
part of a beta.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ install:
 - bash scripts/gogetcookie.sh
 script:
 - make test vet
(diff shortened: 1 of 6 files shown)


mitchellh:
Merging so I can get this beta out.

PR #10055, opened 2016-11-11
state/remote/swift: Add support for versioning state file in swift, and expiring objects

fatmcgav:
state/remote/swift: Add support for versioning state file in swift, and expiring objects after period of time. 
Website docs updated with additional config values.

### DIFF
--- a/state/remote/swift.go
+++ b/state/remote/swift.go
@@ -11,6 +11,8 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
+	"time"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
@@ -22,21 +24,24 @@ const TFSTATE_NAME = "tfstate.tf"
 
 // SwiftClient implements the Client interface for an Openstack Swift server.
 type SwiftClient struct {
-	client     *gophercloud.ServiceClient
-	authurl    string
-	cacert     string
-	cert       string
-	domainid   string
-	domainname string
-	insecure   bool
-	key        string
-	password   string
-	path       string
-	region     string
-	tenantid   string
-	tenantname string
-	userid     string
-	username   string
+	client      *gophercloud.ServiceClient
+	authurl     string
+	cacert      string
+	cert        string
+	domainid    string
+	domainname  string
+	insecure    bool
+	key         string
+	password    string
+	path        string
+	region      string
+	tenantid    string
+	tenantname  string
+	userid      string
+	username    string
+	archive     bool
+	archivepath string
+	expireSecs  int
 }
 
 func swiftFactory(conf map[string]string) (Client, error) {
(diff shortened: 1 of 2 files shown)


fatmcgav:
@jtopjian Would be interested in your thoughts on this one... 
Feels like a really useful addition to me.

The code works in my limited testing, but sure there's some cases I haven't encountered yet...

fatmcgav:
@jtopjian Ping for a review on this one :)

jtopjian:
@fatmcgav this looks really cool - nice work :) It might take me a day or two to look at this one in more detail.

jtopjian:
@fatmcgav Just finished reviewing. Just to make sure I'm not missing anything, there's still a bit of work to do here?

If I understand the feature correctly, this is making an archive copy of the state for preservation (as well as enable optional expiration)? I think that's pretty cool.

I don't see where the state object is being created though -- is that still to be done?

With regard to the time parsing, I think it looks OK. I think you might be able to handle the m/h/d parsing with a `switch` statement, checking the value of the suffix and having a `case` for each possibility.

fatmcgav:
@jtopjian No, I believe this is feature complete now... 

Yes, this enables Swift's object versioning to retain 'previous' tfstate data for change comparison, roll-back etc... 

The state object is being created on [Line 300](https://github.com/fatmcgav/terraform/blob/ddb2a3f59e159a2fecae1b902274539714152d5a/state/remote/swift.go#L300) - Haven't changed that behaviour, other than adding a 'DeleteAfter' attribute if required. 

On the time parsing, I only put handling in for 'days', as `Time.ParseDuration` can handle hours and minutes automatically...

jtopjian:
> The state object is being created on Line 300 - Haven't changed that behaviour, other than adding a DeleteAfter attribute if required. 

Ah - Gotcha. I had to check the docs and remember how versioning works :)

> On the time parsing, I only put handling in for 'days', as Time.ParseDuration can handle hours and minutes automatically... 

:+1: 

All sounds good. I've got a few PRs ready to be merged and I'll add this to the queue. 

Thank you for this!

fatmcgav:
Heh, np...

Cheers for that...

state/remote/swift.go
```diff
@@ -131,6 +136,35 @@ func (c *SwiftClient) validateConfig(conf map[string]string) (err error) {
...
 	c.path = path
 
+	if archivepath, ok := conf["archive_path"]; ok {
+		log.Printf("[DEBUG] Archivepath set, enabling object versioning")
+		c.archive = true
+		c.archivepath = archivepath
+	}
+
+	if expire, ok := conf["expire_after"]; ok {
+		log.Printf("[DEBUG] Requested that rmemote state expires after %s", expire)
```
jtopjian: I was about to merge this but saw the typo and now can't unsee it. Can you do a quick update and I'll merge right after?

fatmcgav: @jtopjian Doh... Done :)

PR #6910, opened 2016-05-28
Nsone provider 

sarguru:
- Retake of #6744
- Unlike #6744 this squashes everything into 2 commits.
- There are some extra changes in the vendor/\* dependencies after restoring from master.

PR #10058, opened 2016-11-11
[WIP] NS1 provider

pashap:
This is a WIP in progress, opening for help/early review.

This is a collection of the work put in by @bobtfish @sarguru @josephholsten and many others, with some changes that (i hope) help model the ns1 api a little better(with regards to ns1 metadata), and use the latest ns1 go rest client.

References:

 - https://github.com/hashicorp/terraform/pull/6910

Quick checklist of work i think is left. May be more, but this seems like good start.

- [ ] Sort out if im on the right track with structuring ns1 metadata
- [ ] Acceptance Test(all three left involve issue with ns1 `permissions` schema)
    - [ ] `AccNS1Team`
    - [ ] `AccNS1User`
    - [ ] `AccNS1Apikey`
- [x] Documentation


builtin/providers/ns1/metadata.go
```diff
@@ -0,0 +1,535 @@
...
+	"fmt"
+	"log"
+	"math"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/data"
+)
+
+func metadataSchema() *schema.Schema {
```
pashap: The main issue Ive been stuggling with is modeling the ns1 metadata structure. NS1 metadata can be attached at three different levels of a dns record, and is dynamic/flexible in nature.

From the [ns1 api docs](https://ns1.com/api/#record-put):

> Metadata tables (meta) may be specified in answers, in regions or in the record as a whole. (Note that regions can be any kind of group, not just geo-related. For this reason, the NS1 web portal at my.nsone.net uses the word "group" instead of "region".) The metadata tables may contain key/value pairs where valid keys and values are as described in the output of /metatypes.

> Anywhere metadata tables can go, data feeds can go as well. Connect a compatible feed to a metadata key by assigning its value to be a JSON object pointing to the data feed that should be consulted for the final value.

Metadata is also modeled in the [ns1 go rest client](https://github.com/ns1/ns1-go/blob/v2/rest/model/data/meta.go)

 
Not sure if `TypeList` of `MaxItems=1` is the best way to construct ns1 metadata. The main issues ive run into are:

 - "static fields" and "dynamic fields" should be in conflict with one another. But because im trying to make `metadataSchema` a reusable/shared struct for `record`s, `record.answer`s, and `record.region`s, im not sure how to use `ConflictsWith` for each "static" metadata field and its "dynamic" counterpart. 
 
 - With issue https://github.com/hashicorp/terraform/issues/5694 in the works, im struggling to get default bool/int/float values working. A good example is the boolean `metadata.Up` field, which has a different meaning when absent as opposed to `false`.
 
So I guess for ns1 metadata, does it look like im on the right track here? It feels like im not using the `helper` package correctly on this structure...

builtin/providers/ns1/permissions.go
```diff
@@ -0,0 +1,309 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/account"
+)
+
+func permissionsSchema() *schema.Schema {
+	return &schema.Schema{
```
pashap: Im not sure if im using the `helper` package correctly with the `permissions` schema here. This may a regression i introduced from the previous implementations of the ns1 provider.  

Im thinking that @bobtfish @sarguru (original authors) have much more wisdom than me, because they had orignally implemented permissions as such:

```
permissions = {
  dns_view_zones = true
  data_manage_datafeeds = true
}
```

Which works without error. 

But then i tried to mimic the ns1 api in making the individual permissions structures nested, but in doing so, caused plans to never be empty:


```
permissions = {
  dns = {
    view_zones = true
  }
  data = {
  	manage_datafeeds = true
  }
}
```


```
=== RUN   TestAccNS1Team_Basic
--- FAIL: TestAccNS1Team_Basic (7.42s)
        testing.go:265: Step 0 error: Check failed: Check 5/5 error: ns1_team.foobar: Attribute 'permissions.dns.#' expected "1", got ""
=== RUN   TestAccNS1Team_Updated
--- FAIL: TestAccNS1Team_Updated (9.47s)

                DIFF:

                UPDATE: ns1_team.foobar
                  permissions.0.account.#:    "1" => "0"
                  permissions.0.data.#:       "1" => "0"
                  permissions.0.monitoring.#: "1" => "0"
```

I can easily move the `permissions` structure back, but just wanted to know if it was possible to do this the 'nested' way. Or if nested structures should be avoided due to dag construction/diffs/all of the complicated cool stuff that the terraform library does for us.

josephholsten:
@pashap I've got a large branch that's currently being used by my company. I'll get it up so we can merge efforts.

josephholsten:
ok, I've got that PR up for your review. I'll try to look over yours and the differences between then soon.

pashap:
Thank you @josephholsten ! I will be looking over yours as well

pashap:
Closing in favor of https://github.com/hashicorp/terraform/pull/10172

PR #10060, opened 2016-11-11
terraform: detect compute counts and show a nicer error

mitchellh:
This will detect computed counts (which we don't currently support) and
change the error to be more informative that we don't allow computed
counts. Prior to this, the error would instead be something like
`strconv.ParseInt: "${var.foo}" cannot be parsed as int`.

### DIFF
--- a/terraform/context_plan_test.go
+++ b/terraform/context_plan_test.go
@@ -1309,6 +1309,26 @@ func TestContext2Plan_countComputed(t *testing.T) {
 	}
 }
 
+func TestContext2Plan_countComputedModule(t *testing.T) {
+	m := testModule(t, "plan-count-computed-module")
+	p := testProvider("aws")
+	p.DiffFn = testDiffFn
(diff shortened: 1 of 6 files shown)


mitchellh:
Merging for 0.8 beta 1. Without this you see the unknown UUID on this error.

sl1pm4t:
@mitchellh - I've been using computed counts for a while with success:
e.g.
```hcl
resource "aws_route53_record" "internal_dns_record" {
  count   = "${length(var.nodes)}"
...
```
where `var.nodes` is a list. 

I'm curious when you see the error you mention above, as I don't get it for the above config.
I'm concerned that this is going to be a bigger breaking change than you realise? It certainly will require a significant rework of many of our modules.

mitchellh:
@sl1pm4t The way this was implemented we're only catching where previously there was an error. If the count can compute without error (to a number) then we don't show the error. Your code should continue to work.

PR #10067, opened 2016-11-11
state/remote/s3: Fix Bug with Assume Role for Federated IAM Account

### LINKED ISSUE #8739: Support for AWS assume roles when configuring S3 remote state storage
Hi,

Since Terraform 0.7.3 we have been able to successfully use assume roles when dealing with the AWS provider.

It would be good if this functionality was also added to the Terraform state storage in S3, so we can use a combination of AWS credentials and assume role to access the S3 bucket.

darend:
## what

Add missing `role_arn` parameter necessary for federated IAM accounts when using s3 for remote state. The attribute is named `role_arn` for consistency with AWS provider: https://www.terraform.io/docs/providers/aws/index.html#role_arn

## usage

```sh
terraform remote config -backend=s3 \
  -backend-config="bucket=<bucket>" \
  -backend-config="key=<key>"  \
  -backend-config="region=$AWS_REGION" \
  -backend-config="profile=$AWS_PROFILE" \
  -backend-config="role_arn=arn:aws:iam::<account_id>:role/<Role>"
```

## testing

I could not see a reasonable way to test this change.  The call to get the credentials returns an aws [Credential](https://github.com/aws/aws-sdk-go/blob/5b341290c488aa6bd76b335d819b4a68516ec3ab/aws/credentials/credentials.go#L155). This does not provide any means of inspecting its internals. Alternatively the `s3Factory` would need to support a way of mocking the [GetCredentials](https://github.com/hashicorp/terraform/blob/f4cf443368486efb37573c67412b7c7beeda9fab/state/remote/s3.go#L63) call

## why

Fixes #8739

### DIFF
--- a/state/remote/s3.go
+++ b/state/remote/s3.go
@@ -66,6 +66,7 @@ func s3Factory(conf map[string]string) (Client, error) {
 		Token:         conf["token"],
 		Profile:       conf["profile"],
 		CredsFilename: conf["shared_credentials_file"],
+		AssumeRoleARN: conf["role_arn"],
 	})
 	// Call Get to check for credential provider. If nothing found, we'll get an
 	// error, and we can present it nicely to the user
--- a/state/remote/s3.go
+++ b/state/remote/s3.go
@@ -68,6 +68,10 @@ func s3Factory(conf map[string]string) (Client, error) {
 		CredsFilename: conf["shared_credentials_file"],
 		AssumeRoleARN: conf["role_arn"],
 	})
+	if err != nil {
+		return nil, err
+	}
+
 	// Call Get to check for credential provider. If nothing found, we'll get an
 	// error, and we can present it nicely to the user
 	_, err = creds.Get()
--- a/website/source/docs/state/remote/s3.html.md
+++ b/website/source/docs/state/remote/s3.html.md
@@ -98,3 +98,4 @@ The following configuration options or environment variables are supported:
    `~/.aws/credentials` will be used.
  * `token` - (Optional) Use this to set an MFA token. It can also be
    sourced from the `AWS_SESSION_TOKEN` environment variable.
+ * `role_arn` - (Optional) The role to be assumed


darend:
It would be ideal if remote config AND the aws provider both supported an ENV flag for the role, akin to how AWS_PROFILE or AWS_SESSION_TOKEN can be used. This would eliminate the need to edit the provider or remote config when the role changes.  Unfortunately AWS does not support an env flag for assume role. 

I don't have a lot of familiarity with the code so I did not pursue it very far. But I could see how an env flag for the role could be enabled for remote config here: https://github.com/hashicorp/terraform/blob/f4cf443368486efb37573c67412b7c7beeda9fab/state/remote/s3.go#L36.

Similarly for the provider the [assume_role](https://github.com/hashicorp/terraform/blob/f4cf443368486efb37573c67412b7c7beeda9fab/builtin/providers/aws/provider.go#L500) config could be modified to support an env flag like this https://github.com/hashicorp/terraform/blob/f4cf443368486efb37573c67412b7c7beeda9fab/builtin/providers/aws/provider.go#L62

darend:
I identified a related issue where the s3 remote does not check for errors after obtaining the AWS credentials. This would lead to crashes
See https://github.com/hashicorp/terraform/pull/10067/files#diff-7bd4c7cfc297dcc74cb94a512b5cece5R71

darend:
@stack72  I do not see this as an enhancement, but a critical bug fix in order to support federated IAM accounts with remote state. It’s critical because without it, it’s not possible to use use multiple AWS accounts with federated logins, which is the best practice for securing AWS environments

benjdlambert:
👍

shinofara:
+1

michelvocks:
Really looking forward to this feature/bug fix :-)

thebenwaters:
+1

stack72:
@darend after speaking with @jen20 on this - we agree this is good to merge :) There is no straight forward way to test but we can verify this locally

jen20:
@stack72, one thing I think we should add is an `errwrap.Wrapf` around the newly returned error so the provenance is more obvious.

darend:
> @darend after speaking with @jen20 on this - we agree this is good to merge :) There is no straight forward way to test but we can verify this locally

Thanks!

jtopper:
I've been trying this out since it merged - it looks like this works for initial remote state configuration, but the `role_arn` key doesn't get written as config into the local state file - as such, subsequent runs of terraform commands fail to assume the role before accessing the remote state.

jtopper:
Update: actually the failure case I mentioned above only holds if there's already a remote state in place without the `role_arn` config key.

jen20:
@jtopper that's a bug - did you just run remote config again?

jtopper:
I had an existing set of remote state, and ran remote config specifying a `role_arn` this time.  That command completed as expected, but the local copy of the state file didn't have a role arn key in it.

Since I'm hacking on an empty account, I deleted the S3 copy and the local state file, and ran remote config again with a role_arn, after which everything behaves as expected.

bnf2si:
I tried to configure remote state from an ec2 which uses the instance profile mechanism.  Allways get access denied.  Anyone facing the same problem?

bacoboy:
Yes, I'm seeing the same thing with 0.8.7.  Using the non-default account keys (in ~/.aws/credentials) of `security` for my role `role/Admin` (which works via CLI) using these additional s3 backend config fields:
```
  -backend-config="profile=security" \
  -backend-config="role_arn=arn:aws:iam::XXXXXXXXX:role/Admin" \
```
I get this error:
```
Error while performing the initial pull. The error message is shown
below. Note that remote state was properly configured, so you don't
need to reconfigure. You can now use `push` and `pull` directly.

Error preparing remote state: Failed to upload state: AccessDenied: Access Denied
	status code: 403, request id: B5EAECCE34839FA9
```

The thing that took me a second to figure out is that profiles specified in `~/.aws/config` are ignored by terraform.  It only looks for profiles in `~/.aws/credentials` (unless you've overridden that with the `shared_credentials_file` parameter) and if you use those credentials to assume a role you must also specify the `role_arn` (just like you did in the `~/.aws/config` file).

So best I can tell this is still broken.  Happy to work with somebody more familiar with the codebase to get this resolved.

bacoboy:
I was having an issue, but realized I was still using 0.8.6.  After upgrading to 0.8.7, I can confirm it is working.

@bnf2si take a look at your trust profile.  By default an EC2 IAM role just has the ec2 service listed as "trusted" in the "trust relationship" tab.  To make this works you need to do 2 things:

1) grant your EC2 box the `AssumeRole` permission for the specific role as in this example:
```
{
    "Version": "2012-10-17",
    "Statement": {
        "Effect": "Allow",
        "Action": "sts:AssumeRole",
        "Resource": "arn:aws:iam::YOUR_ACCOUNT:role/Admin"
    }
}
```
But then you need to also add a trust policy to the target role (in this example `role/Admin`) to the source principal.  The easiest way to do this is with a cross account role (but using your own account) in the Trust Relationship:
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::YOUR_ACCOUNT:root"
      },
      "Action": "sts:AssumeRole"
    }
  ]
}
```
Hope this helps.

bnf2si:
I also found a solution now for my usecase.
I wanted run the use terraform files from a) federated account access and b) from an EC2 via profile

With the following assume role configuration in my terraform files  I had to attach a policy to "JenkinsAgentCreator" role to allow assuming itself , otherwise I cannot run from terraform from an EC2 via profile.
Proposal: Terraform should recognize that it runs from an EC2 via profile and in this case not assuming the role because it is already assumed somehow.

My example code
`
provider "aws" {
  region = "${var.aws_region}"
  insecure = "${var.insecure}"
  allowed_account_ids = ["${var.account_id}"] // allow only to do things in sample account, nowhere else

  assume_role {
    role_arn = "arn:aws:iam::${var.account_id}:role/JenkinsAgentCreator"
    session_name = "creating_jenkins_agents"
  }
}

data "terraform_remote_state" "agents" {
    backend = "s3"
    config {
        bucket = "terraform-state-eclipseprojects.etas-dev.com"
        key = "agents_golden_ami/terraform.tfstate"
        region = "eu-central-1"
        encrypt = "true"
        role_arn = "arn:aws:iam::${var.account_id}:role/JenkinsAgentCreator"
    }
}
`

ben-robinson:
Problem: Running Terraform from an EC2 instance with an instance profile that allows assuming roles, Terraform can assume another role in the AWS provider block, however, cannot assume another role for remote config using S3 backend with the "role_arn" configuration variable. 

Workaround: Script that performs an STS assume role (assumes that assumed role has the appropriate s3 privileges), script disables remote config, script reenables remote config with the STS temporary credentials.

Note: Terraform will complain when the credentials expire. 

"Error reloading remote state: ExpiredToken: The provided token has expired."

Rerun the script. Run Terraform.

PR #10068, opened 2016-11-11
terraform: destroy edge must include resources through outputs

mitchellh:
This fixes: `TestContext2Apply_moduleDestroyOrder`

The new destroy graph wasn't properly creating edges that happened
_through_ an output, it was only created the edges for _direct_
dependents.

To fix this, the DestroyEdgeTransformer now creates the full transitive
list of destroy edges by walking all ancestors. This will create more
edges than are necessary but also will no longer miss resources through
an output.

### DIFF
--- a/terraform/transform_destroy_edge.go
+++ b/terraform/transform_destroy_edge.go
@@ -116,8 +116,10 @@ func (t *DestroyEdgeTransformer) Transform(g *Graph) error {
 	// in the graph. BUT if resource A is just pure destroy, then only
 	// destroy A is in the graph, and create A is not.
 	steps := []GraphTransformer{
+		&OutputTransformer{Module: t.Module},
 		&AttachResourceConfigTransformer{Module: t.Module},
 		&AttachStateTransformer{State: t.State},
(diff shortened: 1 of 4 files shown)


terraform/transform_destroy_edge.go
```diff
@@ -150,14 +154,22 @@ func (t *DestroyEdgeTransformer) Transform(g *Graph) error {
...
-	refMap := NewReferenceMap(tempG.Vertices())
+	log.Printf("[TRACE] DestroyEdgeTransformer: reference graph: %s", tempG.String())
 
 	// Go through all the nodes in the graph and determine what they
 	// depend on.
-	for _, v := range tempG.Vertices() {
-		// Find all the references
-		refs, _ := refMap.References(v)
+	for _, v := range tempDestroyed {
+		// Find all descendents of this to determine the edges we'll depend on
```
jbardin: should be "find all _ancestors_"?

mitchellh: Yeah :) Got my wording mixed up.

PR #10080, opened 2016-11-13
terraform: support version requirement in configuration

### LINKED ISSUE #1400: Feature Request: Terraform version requirements in configuration files
I'm starting to work on updating the [terraform community modules](https://github.com/terraform-community-modules/) to take advantage of the new [split function](http://www.terraform.io/docs/configuration/interpolation.html#split_delim_string_), but this will leave users on older versions unable to use the latest versions of these modules.

Being able to put some kind of version constraint as a comment in the configuration file could be useful, as new functions are added in newer versions.

I'm going to do my best to call out the version requirement in the documentation but an additional safeguard would also help.

mitchellh:
Fixes #1400 

This adds a new configuration to constrain the Terraform version that can perform operations on a given configuration. 

### Configuration

This introduces a new configuration block `terraform` which will be used now and in the future for meta-configuration (configuration that configures Terraform itself). For now, we introduce only a single configuration `required_version`:

```hcl
terraform {
    required_version = "> 0.7.0, < 0.8.0"
}
```

The `required_version` value can be a comma-separated list of version constraints. This flexibility allows the user to be as strict as they'd like for version requirements. The constraints are all of those accepted by https://github.com/hashicorp/go-version but we'll document this in the website.

### Execution

The required version is validated immediately upon loading Terraform in order to prevent any potential side effects (state change for example) as well. 

**If a module has a version requirement,** then that will also be validated. The entire module tree must have satisfied Terraform version constraints.

There is no way to "force" execution even if a required version is not met. There is a flag internally that allows this, but we don't currently expose it to the end user. I'm not sure there is a reason for doing this yet...

Here is what the error looks like:

2016-11-12 at 5 02 pm

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/hil"
 	"github.com/hashicorp/hil/ast"
 	"github.com/hashicorp/terraform/helper/hilmapstructure"
@@ -27,6 +28,7 @@ type Config struct {
 	// any meaningful directory.
 	Dir string
 
+	Terraform       *Terraform
 	Atlas           *AtlasConfig
 	Modules         []*Module
 	ProviderConfigs []*ProviderConfig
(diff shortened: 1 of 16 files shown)


mitchellh:
(Test failures unrelated and fixed by #10068, but I'm kicking the builds to get them green from races)

fatmcgav:
@mitchellh Could this config syntax be extended to support storing 'remote state' config in TF files?

As that would massively simplify the workflow when working with multiple TF projects - Different env's for example :)

mitchellh:
@fatmcgav Its possible, yes, but out of scope of course for this PR.

fatmcgav:
@mitchellh Of course :) 

Was more of a 'would it be a useful addition' question, and sounds like the answer is "Yes" 👍

Ninir:
@mitchellh Do you think interpolation would be possible in this very specific case?

I was thinking about having a global var that I get and use in many different places. It would allow me to update in one place and change everywhere else, for instance when upgrading major versions.

Do you think this is an acceptable use-case?

PR #10084, opened 2016-11-13
provider/scaleway: improve volume attachment

nicolai86:
this PR tries to improve the the issues troubling the `scaleway_volume_attachment` tests.

Some issues we're seeing:

### waiting for server times out.  
  this is related to the fact that you have to stop servers to change volumes, and this transfers ALL the data from the server to some storage medium. We've been using 20 minutes so far, and the [Scaleway community](https://community.online.net/t/solving-the-long-shutdown-boot-when-only-needed-to-attach-detach-a-volume/326) lists >1h times for bigger volumes. 
  For now, I've just bumped the limit to 60 minutes.

related tests: 
- https://travis-ci.org/hashicorp/terraform/builds/175439974
- https://travis-ci.org/hashicorp/terraform/builds/175273317

### verify volumes are unattached before proceeding.
  just to make sure that the attachment doesn't fail because of an already attached volume. Automatic detaching and reattaching is out of scope for now.

related tests:
- https://travis-ci.org/hashicorp/terraform/builds/175002940

### general stability
  I suspect that the `scaleway_volume_attachment` tests error because of the cache used within the official go sdk (state mismatch remote vs local). The easiest way to verify this for now is to just clear the cache when attaching/ detaching volumes.

ping @stack72 PTAL :)

### DIFF
--- a/builtin/providers/scaleway/helpers.go
+++ b/builtin/providers/scaleway/helpers.go
@@ -84,7 +84,7 @@ func deleteStoppedServer(scaleway *api.ScalewayAPI, server *api.ScalewayServer) 
 // the helpers.go file pulls in quite a lot dependencies, and they're just convenience wrappers anyway
 
 func waitForServerState(scaleway *api.ScalewayAPI, serverID, targetState string) error {
-	return resource.Retry(20*time.Minute, func() *resource.RetryError {
+	return resource.Retry(60*time.Minute, func() *resource.RetryError {
 		s, err := scaleway.GetServer(serverID)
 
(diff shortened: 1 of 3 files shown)


nicolai86:
travis tests are broken b/c of some unrelated breakage in master I assume: 

```
--- FAIL: TestContext2Apply_moduleDestroyOrder (0.01s)
	context_apply_test.go:1636: bad: []string{"a", "b"}
```

stack72:
LGTM! Thanks :)

PR #10086, opened 2016-11-13
fix cloudstack bugs

nyankichi820:
fix cloudstack provider bugs

* add zone id to ip associate parameter 
* cloudstack API list Firewall throw exception ( go-cloudstack bugs). invalid json type
   * create PR original repogitory
   * https://github.com/xanzy/go-cloudstack/pull/70

### DIFF
--- a/builtin/providers/cloudstack/resource_cloudstack_ipaddress.go
+++ b/builtin/providers/cloudstack/resource_cloudstack_ipaddress.go
@@ -16,6 +16,12 @@ func resourceCloudStackIPAddress() *schema.Resource {
 		Delete: resourceCloudStackIPAddressDelete,
 
 		Schema: map[string]*schema.Schema{
+			"zone_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
 			"network_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
@@ -56,6 +62,11 @@ func resourceCloudStackIPAddressCreate(d *schema.ResourceData, meta interface{})
 	if networkid, ok := d.GetOk("network_id"); ok {
 		// Set the networkid
 		p.SetNetworkid(networkid.(string))
+	}
+
+	if zoneid, ok := d.GetOk("zone_id"); ok {
+		// Set the zoneid
+		p.SetZoneid(zoneid.(string))
 	}
 
 	if vpcid, ok := d.GetOk("vpc_id"); ok {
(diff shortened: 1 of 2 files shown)


nyankichi820:
I know that it is better not to change the vendor code, but I'm daring to fix the problem. Since I am doing a Pull request to the original here, I think that if you can fix that there is no problem put it here. Since addition of zone id is optional, regardless of the version of the server, I think that there is no problem because it can deal with it

svanharmelen:
@nyankichi820 I understand and appreciate you are trying to fix a problem, but you still should never manually update vendored code. In a case like this the best you can do is make the needed Terraform changes in this PR and then in the comments clearly mention that this PR is dependent on another PR.

As for my other questions, please give me an example config (and any other info that might be relevant) so I can reproduce the problems you are trying to fix here.

You are right that adding the `zone_id` should not be a problem because it has always been an optional field in the API, yet you are changing the resource so I would say there needs to be a reason to do so. Until now I did never see a use case that needed this field to make things work, so please help me understand what you are trying to fix.

And this goed even more for the firewall related change as this seems to be a backwards incompatible fix. So accepting this PR maybe fixes things for your use case, but at the same time it will break thing for a lot of other users.

So that's why I want to be able to see and understand what you are trying to fix so we can look at the best way to do so for all users and all use cases.

nyankichi820:
My problem is the problem with a specific cloud service, so we are currently checking the version of Cloudstack.
If it is a problem that occurs only with this service, please discard this fix.

svanharmelen:
@nyankichi820 it would really help if you gave some more details so we can also have a look and understand what we are talking about here.

At the minimum it would be very helpful to know which service provider you are using...

nyankichi820:
I am checking whether this change is due to version upgrade of cloudstack or by original implementation of service side. I think that this modification should not be done if it is original implementation.

nyankichi820:
私がサービス提供者に確認したところ、cloudstackそのままを利用しているのではなく、商用のcloudplatformというものを利用しているそうです。
https://accelerite.com/products/cloudplatform/
そのため、純粋なcloudstackとは違うため、APIに違いが発生していそうです。そのためこの修正は入れるべきではないと思いますので、Pull requestを取り下げます。

nyankichi820:
When I checked with the service provider, I heard that they are using commercial cloudplatform instead of using cloudstack as it is.
Https://accelerite.com/products/cloudplatform/
Therefore, it is different from pure cloudstack, so it seems that there is a difference in API. Therefore, I think that this correction should not be included, so I withdraw the pull request.

svanharmelen:
@nyankichi820 thanks for the feedback. In that case I will close this PR...

Seems pity to me that they diverge their API as now they will not be able to be used by a lot of mainstream open source tooling that is compatible with CloudStack. But there is nothing I can do about that.

Thanks again...

Issue #8668, opened 2016-09-06
CloudFront expects different S3 website endpoint


llarsson:
I am creating an S3 bucket to host a static web page, and want to put CloudFront in front of it. Standard stuff.

To make the `origin` block in my `aws_cloudfront_distribution` resource work, I have to put the following hackish code into it for `domain_name`:

`domain_name = "${replace(aws_s3_bucket.origin.website_endpoint, aws_s3_bucket.origin.website_domain, "s3.amazonaws.com")}"`

The problem is that my `website_endpoint` includes a region-specific (`eu-west-1`) domain name, but CloudFront does not accept those.

How do people generally do this very standard thing (static web page on S3, distribute it via CloudFront)?

mikereinhold:
According to http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html, CloudFront supports both formats depending on your bucket configuration. That said, my experience has been similar to yours - the _{bucket-name}.s3.amazonaws.com_ format has been most reliable for me.

I've approached it in Terraform using the following:

````hcl
domain_name = "${aws_s3_bucket.origin.bucket}.s3.amazonaws.com"
````

It's a little less verbose than your example, but it works.

It would be nice if Terraform exposed an S3 resource output that had this format natively as that would streamline configuration, but the above isn't too bad.

llarsson:
Thank you, @mikereinhold. That is significantly easier to read. 

The point still remains, though -- it would be nice to just have Terraform wire the values together and never have to potentially worry about the format changing. :)

PR #10087, opened 2016-11-13
provider/aws: Improved the Cloudfront documentation when using s3

Ninir:
This "fixes" https://github.com/hashicorp/terraform/issues/8668 by better exposing the relation between S3 and CloudFront origin domain_names.

### DIFF
--- a/website/source/docs/providers/aws/r/cloudfront_distribution.html.markdown
+++ b/website/source/docs/providers/aws/r/cloudfront_distribution.html.markdown
@@ -25,9 +25,18 @@ want to wait, you need to use the `retain_on_delete` flag.
 The following example below creates a CloudFront distribution with an S3 origin.
 
 ```
+resource "aws_s3_bucket" "b" {
+  bucket = "mybucket"
+  acl    = "private"
+
+  tags {
+    Name = "My bucket"
+  }
+}
+
 resource "aws_cloudfront_distribution" "s3_distribution" {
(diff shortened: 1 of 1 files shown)


apparentlymart:
Thanks for this, @Ninir! This makes things clearer.

One thing I'm not sure of: I kinda remember in the past having to use a different domain name here when the bucket is not in the `us-east-1` region. Am I remembering that right? Just wondering if we need to mention something about that here, and in the long run perhaps expose an attribute on `aws_s3_bucket` that just takes care of providing the right full value here.

Ninir:
Hi @apparentlymart,

thanks for the quick feedback on this! 

I think you are mentionning section ***The bucket is not configured as a website endpoint*** in http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html, which states:
 
>In general, use the following format:
>
>    bucket-name.s3.amazonaws.com
>
>If your bucket is in the US Standard region and you want Amazon S3 to route requests to a facility in Northern Virginia, use the following format:
> 
>    bucket-name.s3-external-1.amazonaws.com

Regarding the second point, I initiated something in my other [PR](https://github.com/hashicorp/terraform/pull/10088) (which can be enhanced).

stack72:
LGTM - thanks for the work here @Ninir and the review @apparentlymart :)

PR #10088, opened 2016-11-13
provider/aws: Added a s3_bucket domain name attribute

Ninir:
This adds a new attribute to the s3_bucket resource called `bucket_domain_name`. It acts as a "helper" in the way it is computed on the Terraform side.

It just makes it easier to work with CloudFront when using S3 Origin Identities, and avoid #8668 for instance.
However, It is different from #10087 in the way that this work can be discussed (thus not mergeable now for example), whereas my other PR adds more insights on the documentation quickly.

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket.go
@@ -34,6 +34,11 @@ func resourceAwsS3Bucket() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"bucket_domain_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"arn": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -527,6 +532,8 @@ func resourceAwsS3BucketRead(d *schema.ResourceData, meta interface{}) error {
 	if _, ok := d.GetOk("bucket"); !ok {
 		d.Set("bucket", d.Id())
 	}
+
+	d.Set("bucket_domain_name", bucketDomainName(d.Get("bucket").(string)))
 
 	// Read the policy
 	if _, ok := d.GetOk("policy"); ok {
@@ -1193,6 +1200,10 @@ func websiteEndpoint(s3conn *s3.S3, d *schema.ResourceData) (*S3Website, error) 
 	}
 
 	return WebsiteEndpoint(bucket, region), nil
+}
+
+func bucketDomainName(bucket string) string {
+	return fmt.Sprintf("%s.s3.amazonaws.com", bucket)
 }
 
 func WebsiteEndpoint(bucket string, region string) *S3Website {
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/resource_aws_s3_bucket.go
```diff
@@ -1195,6 +1202,10 @@ func websiteEndpoint(s3conn *s3.S3, d *schema.ResourceData) (*S3Website, error)
 	return WebsiteEndpoint(bucket, region), nil
 }
 
+func bucketDomainName(bucket string) string {
+	return fmt.Sprintf("%s.s3.amazonaws.com", bucket)
```
stack72: is this always the case? Don't different regions mean different bucket names?

Ninir: Hi @stack72, 
Well, I think so. While looking at [this documentation](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html), it is stated under `The bucket is not configured as a website endpoint` that:
> In general, use the following format:
>
>    bucket-name.s3.amazonaws.com
>
> If your bucket is in the US Standard region and you want Amazon S3 to route requests to a facility in Northern Virginia, use the following format:
>
>    bucket-name.s3-external-1.amazonaws.com
>
> When you specify the bucket name in this format, you can use the following CloudFront features:
> 
> Configure CloudFront to communicate with your Amazon S3 bucket using SSL. For more information, see Using HTTPS with CloudFront.
> Use an origin access identity to require that your users access your content using CloudFront URLs, not by using Amazon S3 URLs. For more information, see Using an Origin Access Identity to Restrict Access to Your Amazon S3 Content.
> Update the content of your bucket by submitting POST and PUT requests to CloudFront. For more information, see HTTP Methods in the topic How CloudFront Processes and Forwards Requests to Your Amazon S3 Origin Server.

Tried creating a S3 bucket in "US Standard" and a CloudFront distrib from the console, and it tries to get the bucket using: `myname.s3.amazonaws.com`.

stack72:
Thanks for this @Ninir :)

This LGTM!

P.

builtin/providers/aws/resource_aws_s3_bucket.go
stack72: Hi @Ninir 

Did you manage to create this test in 2 different regions?

P.

stack72:
ping @Ninir :)

Ninir:
Hi @stack72 

Sorry for the latency. Will re-run the configurations to be sure.
I want to handle static & non-static website hostings.

As far as I checked, CF distributions were working nicely with S3 origins from us-east-1 & us-west-2.
Checking at the moment with non-website CF distributions, for sake purposes :)

Ninir:
Hi @stack72 ,

It works pretty nicely for both static hosting & non-static hosting origins and both us-east-1 & eu-west-2.
Do you want a configuration to check it out too?

stack72:
Hi @Ninir 

Thanks for the wait on this - while testing, I found a panic in master on s3 bucket notification. So I am going to try and work that out then i will get this merged

Paul

stack72:
ok, issue narrowed down to a testing code issue - merging - this LGTM! Thanks @Ninir :)

PR #10093, opened 2016-11-14
Add `terraform console` for REPL

### LINKED ISSUE #9750: Function REPL-esque Evaluator?
It would be nice to have a sort of REPL-like 'playground' where you could simply type in string blocks with interpolation functions. For example when I'm doing more complicated stuff with a series of functions I'd like to have a way outside of using a dummy Terraform project with a single file & output to make sure the functions will do what I think they'll do for a certain input as a quick sanity check.

mitchellh:
Fixes #9750

This PR adds a new command `terraform console` that opens an interactive console for experimenting with interpolations. I forsee in the future the console being much more powerful and being to invoke Terraform, modify state, etc. But, for v1 of this feature, it is only used for interacting with interpolations.

The `console` command _will_ read your state so you can access resources from the console. If you configure a remote state, it can read that as well.

There are a few use cases for this:

  * **Learning:** You can use the console to learn and play with interpolations.

  * **Testing:** You can try interpolations before you use them, or verify that they're working correctly.

  * **Scripting:** You can use interpolations to read values and transform them from the state in a Unix-friendly way.  This is especially true for remote state.

## Demo Gif 

Interactive mode:

repl

Piped input:

```
$ echo "1+5" | terraform console
6
```

## PR Contents

  * `helper/wrappedreadline`: New helper library for gaining access to the original standard streams and not the wrappers given to us by panicwrap. This is important for `readline` to work properly.

  * `repl`: New core Terraform library for REPL-behavior. This extraction allows us to unit test it very well instead of it all being stuffed in core. Also, by being a separate library there exists a future where we may be able to drop a REPL in scenarios where Terraform fails at runtime (with a debug flag or something). Maybe.

  * `command/console`: The console command.

  * `terraform/interpolation.go` changes: Improving the error messages that are now possible due to the REPL. This file continues to be a primary target for refactor and change in the future.

  * vendor update `panicwrap`: panicwrap needed to be updated to pass through raw standard streams.

## Future

The following are some ideas I have for the future of this that won't be included in this PR or targeted even for 0.8:

  * Tab-completion: we get this from the readline library we use, but we need to complete the completer. 

  * Setting variables

  * Listing resources

  * Actual resource configuration
  
  * Operations such as plan/apply

### DIFF
--- a/command/console.go
+++ b/command/console.go
@@ -1,0 +1,183 @@
+package command
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/wrappedreadline"
+	"github.com/hashicorp/terraform/repl"
+
+	"github.com/chzyer/readline"
+	"github.com/mitchellh/cli"
+)
+
+// ConsoleCommand is a Command implementation that applies a Terraform
+// configuration and actually builds or changes infrastructure.
+type ConsoleCommand struct {
+	Meta
+
+	// When this channel is closed, the apply will be cancelled.
+	ShutdownCh <-chan struct{}
+}
+
+func (c *ConsoleCommand) Run(args []string) int {
+	args = c.Meta.process(args, true)
+	cmdFlags := c.Meta.flagSet("console")
+	cmdFlags.StringVar(&c.Meta.statePath, "state", DefaultStateFilename, "path")
+	cmdFlags.Usage = func() { c.Ui.Error(c.Help()) }
+	if err := cmdFlags.Parse(args); err != nil {
+		return 1
+	}
+
+	pwd, err := os.Getwd()
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Error getting pwd: %s", err))
+		return 1
+	}
+
+	var configPath string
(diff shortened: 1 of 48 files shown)


apparentlymart:
Heh... when I was writing the new HIL parser I was thinking about whether I should add in a mode where it parses as if it were already inside `${ ... }` for this purpose, but I suppose just wrapping the user input in that directly is a reasonable enough way to get there.

This would just be polish that could be handled separately, but might be worth considering using the new entry point I added to HIL that allows the caller to specify the start position, so you can set it to be something like "line 1, column -1" so that the user-entered expression ends up starting at line 1, column 1. Though probably easier to wait to do that until Terraform's interpolator itself is able to pass config positions into the HIL parser, since this layer can't "see" the HIL interface at all.

mitchellh:
@apparentlymart Yeah I thought about both of those options too, but there are a couple things holding me back: 

  1.) Parsing 'interpolation mode' first would require significant refactors of TF since `config.RawConfig` is currently the easiest way to make this work rather than parsing the HIL directly. So I'd rather rethink that first vs updating HIL

  2.) ast.Pos is blocked by above too. :)

PR #10095, opened 2016-11-14
[WIP] provider/aws: Add separate resource for aws_elb_listener

### LINKED ISSUE #9807: Add a separate ELB listener resource - ie. aws_elb_listener.
Currently (as of 0.7.8), ELB listeners are defined and managed as in-line _sub_-resources. This makes it very difficult to create modules for re-use as the number of listeners is static.

Terraform should support separate resources for listeners so that these can be made more modular and allow for meta-parameters on listeners - ie. `aws_elb_listener`. This is similar to the `aws_elb_attachment` and `aws_security_group_rule` resources that exist.

##### Current:
```
resource "aws_elb" "bar" {
   ...
  listener {
    instance_port = 8000
    instance_protocol = "http"
    lb_port = 80
    lb_protocol = "http"
  }
}
```

##### Proposed:
```
resource "aws_elb" "bar" {
}

resource "aws_elb_listener" "bar" {
  instance_port = 8000
  instance_protocol = "http"
  lb_port = 80
  lb_protocol = "http"
}
```

stack72:
Fixes #9807

Allows similar listeners to be specified once and then attached to
multiple loadbalancers

Unfortunalely, the AWS SDK doesn't allow an AWS ELB to be created
without a Listener so we still have to keep that AWS ELB Listener
parameter as `required` - therefore, this may be blocked until AWS get
back to us on whether that is actually the usecase

### DIFF
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -244,6 +244,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_elastictranscoder_preset":                 resourceAwsElasticTranscoderPreset(),
 			"aws_elb":                                      resourceAwsElb(),
 			"aws_elb_attachment":                           resourceAwsElbAttachment(),
+			"aws_elb_listener":                             resourceAwsElbListener(),
 			"aws_emr_cluster":                              resourceAwsEMRCluster(),
 			"aws_emr_instance_group":                       resourceAwsEMRInstanceGroup(),
 			"aws_flow_log":                                 resourceAwsFlowLog(),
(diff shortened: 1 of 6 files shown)


catsby:
Can we close this? I have not seen any indication in the docs or the SDK that suggests that ELB version 1 will ever loosen up the restriction of requiring at least 1 listener defined at creation. With this restriction in place I think we shouldn't do this as is. 

With a resource like `aws_security_group`, we stipulate that you cannot (safely) use `aws_security_group` in-line rules and also `aws_security_group_rule` resources; doing so prevents the security group from detecting drift and can result in constant plans.

Because an in-line rule is required for a `aws_elb`, we can't make that stipulation. I worry that we'll end up hurting users at worst, and overall just a weird UX at best :/

stack72:
Closing this one for now - thanks for the nudge @catsby 

This isn't going to work easily and needs some real thought given to it

PR #10104, opened 2016-11-14
DebugOperation

jbardin:
The DebugOperation method marks the start of a set of operations on the Graph, with
extra information optionally provided in the second parameter. This
returns a function which can be called via its End method to mark the end of the set
in the logs.

Start by using this to record each Transform step in the debug log.


mitchellh:
This looks great :+1: I like the cleanup.

PR #10105, opened 2016-11-14
provider/chef: Migrate Chef to use KEY_MATERIAL rather than using a Pem file

stack72:
This will allow us to standardise on how we set KEY_MATERIAL


jen20:
This LGTM - it finishes the work that @phinze and I did a while back. However, we should update the documentation to make it clear that you need the key material and not a path.

stack72:
@jen20 documents updated as requested. Will merge when it goes green and @mitchellh has finished 0.8-beta

PR #10115, opened 2016-11-14
state/azure: support passing of lease ID when writing storage blob

pmcatominey:
This change allows users of the Azure remote state backend to provide a Lease ID for use when the state is being written.

## Reasoning

We use Terraform in a distributed team which brings challenges around managing state and avoiding conflicts. To solve these issues we are using a fork of [Terragrunt](https://github.com/gruntwork-io/terragrunt) to lock the remote state blobs while Terraform is performing operations. This is implemented by acquiring a lease on the storage blob and releasing it once Terraform is finished. Without a method of supplying the Lease ID to the Azure remote state backend, it is not possible to write to a locked blob.

### DIFF
--- a/state/remote/azure.go
+++ b/state/remote/azure.go
@@ -46,11 +46,13 @@ func azureFactory(conf map[string]string) (Client, error) {
 	}
 
 	blobClient := storageClient.GetBlobService()
+	leaseID, _ := confOrEnv(conf, "lease_id", "ARM_LEASE_ID")
 
 	return &AzureClient{
 		blobClient:    &blobClient,
 		containerName: containerName,
 		keyName:       keyName,
+		leaseID:       leaseID,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


state/remote/azure.go
```diff
@@ -46,11 +46,13 @@ func azureFactory(conf map[string]string) (Client, error) {
 	}
 
 	blobClient := storageClient.GetBlobService()
+	leaseID, _ := confOrEnv(conf, "lease_id", "ARM_LEASE_ID")
 
 	return &AzureClient{
 		blobClient:    &blobClient,
 		containerName: containerName,
 		keyName:       keyName,
+		leaseID:       leaseID,
```
stack72: what happens if we pass an empty string here?

pmcatominey: Only non empty values should be used in the request, adding a test to confirm

PR #10116, opened 2016-11-14
increase aws_lambda_function timeout

jeremy-asher:
I have been experiencing occasional timeouts creating a Lambda function on a good connection and consistent timeouts on slower ones with a 23 meg zip payload.  Based on testing on slow wifi I see the request taking 90-150 seconds.  This increases the timeout to a more reasonable 10 minutes.


stack72:
Hi @jeremy-asher 

Paul

Issue #10114, opened 2016-11-14
Resource variables that include attribute indexes stopped working


trodemaster:
Terraform v0.8.0-dev (7fbd880390d2009cedba6a24d2ea0d6cfd3bbd9d)
openstack_compute_instance_v2
vsphere_virtual_machine

Recent changes around handling attribute names with index values and periods in the name seem to be causing issues. I suspect this https://github.com/hashicorp/terraform/commit/6f347ebb3ae3ed4c0ffe6f7f8f4d3f449002e54b change may be related. 

### Terraform Configuration Files
```hcl
# create test vm and run test script
resource "vsphere_virtual_machine" "vmname-Ubuntu1604" {
  name   = "vmname-Ubuntu1604"
  depends_on = [ "null_resource.sync-vmname-Ubuntu1604" ]
  vcpu   = 2
  memory = 4096
  resource_pool = "clustername/Resources/Image_Lab_Validation"
  linked_clone = "false"
  folder = "Image_Lab_Validation"
  datacenter = "LAB"
  cluster = "clustername"
  dns_servers = ["8.8.8.8","4.4.4.4"]
  dns_suffixes = [".local",".nachos"]
  network_interface {
      label = "vmnet"
  }
  disk {
    template = "/LAB/vm/Templates/Ubuntu1604"
    datastore = "vsan"
    type = "thin"
  }
  connection {
  type = "ssh",
  agent = "false"
  host = "${vsphere_virtual_machine.vmname-Ubuntu1604.network_interface.0.ipv4_address}",
  user = "root"
  password = "PASS"
  script_path = "/root/tf.sh"
  }
  provisioner "remote-exec" {
    inline = [ 
    "date"
    ]
  }
}

output "vmname-vsphere-Ubuntu1604" {
  value = "${vsphere_virtual_machine.vmname-Ubuntu1604.network_interface.0.ipv4_address}"
}
```

### Debug Output

Error loading files Error loading /files/Ubuntu1604-vsphere.tf: Error reading config for output vmname-Ubuntu1604: vsphere_virtual_machine.vmname: resource variables must be three parts: TYPE.NAME.ATTR in:

${vsphere_virtual_machine.vmname-Ubuntu1604.network.0.floating_ip}

### Expected Behavior
The ip address should be retrieved from this variable. 

### Actual Behavior
It looks like recent changes to the "dot" package are causing issues when the attribute includes an index value. 

### Steps to Reproduce
[...]

PR #10119, opened 2016-11-15
vendor: update hil

mitchellh:


mitchellh:
For #10114 (never released)

Issue #7311, opened 2016-06-24
Intermittent aws_db_instance & aws_db_option_group lifecycle issues 


jrnt30:
We are seeing fairly regular issues deleting `aws_db_option_group` when removing the `aws_db_instance` first in the same destroy path. 

Looking at the debug logs, the issue is that while the DB instance is non-existent per the `Request rds/DescribeDBInstances` returning 404 however upon initial `rds/DeleteOptionGroup` request there is a 400 error thrown for `InvalidOptionGroupStateFault` due to `The option group 'dev-data-primary-justin-option-group' cannot be deleted because it is in use.`

### Terraform Version
0.6.16+

### Affected Resource(s)
- aws_db_option_group

### Debug Output
```
* aws_db_option_group.xmldb: Error Deleting DB Option Group: InvalidOptionGroupStateFault: The option group 'dev-data-primary-option-group' cannot be deleted because it is in use.
 	status code: 400, request id: d732f0b9-37a5-11e6-a48d-af638ac63cab
```

### Expected Behavior
DB Option group deletes properly seeing that AWS DB Instance had been deleted

### Actual Behavior
AWS exception stating that DB Option Group is still in use and fails deletion (intermittently but primarily with Oracle RDS instances)

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `terraform destroy`

### Important Factoids
I have been unable to replicate this yet with smaller MySQL instances however can consistently reproduce with the Oracle 11-ee instance we have based off a larger snapshot.

beakerman29:
I know that this was fixed by increasing the timeout limit. I am wondering if as a best practice it would be beneficial to allow an acceptable timeout switch. Not just for this bust for all items.

timvaines:
I get this issue consistently with SQL Server. Eventually after a few days you can delete the option set.

PR #10131, opened 2016-11-15
GH-7311 - Increasing deletion timeout wait

jrnt30:
- Per our discussion around the PR to increase this initially, we
weren't sure if 1 minute would be sufficient.  Well, it turns out it
wasn't for me today (we don't delete these often so not sure how
often people run into this).

Picking another somewhat arbitrary value of 5 minutes in the hopes
that it will be sufficient (today it took a little over 2 minutes).


catsby:
Thanks @jrnt30 ! Some kind of pre-resource configurable timeout mechanism is on the drawing board, hopefully to deliver sooner rather than later 😉

jrnt30:
That would be pretty interesting if it's what I am thinking (retries more at the framework level instead of the individual resources implementing that itself).  One thing that I see being pretty tricky is the actual "failure cases" that should be retried vs. those which are truly fatal.

PR #10134, opened 2016-11-15
helper/shadow: Close for auto-closing all values

### LINKED ISSUE #10122: Using an attribute in template_file data source hangs when inside a module
### Terraform Version
0.7.10

### Affected Resource(s)
- data.template_file

###### main.tf
```hcl
module "testing" {
  source = "./my-module"
}
```

###### my-module/main.tf
```hcl
data "template_file" "main" {
  template = "$${val}"
  vars {
    val = "${null_resource.main.id}"
  }
}

resource "null_resource" "main" { }
```

### Debug Output
```
...
2016/11/15 00:54:03 [ERROR] No DOT origin nodes found.
...
2016/11/15 00:54:08 [DEBUG] vertex "meta.count-boundary (count boundary fixup)", waiting for: "module.testing.data.template_file.main"
2016/11/15 00:54:13 [DEBUG] vertex "meta.count-boundary (count boundary fixup)", waiting for: "module.testing.data.template_file.main"
```

### Expected Behavior
It should create the resource and complete the process.

### Actual Behavior
It creates the resource, but Terraform completely hangs with the message `module.testing.data.template_file.main: Refreshing state...`. Pressing ctrl+c shows `Interrupt received. Gracefully shutting down...`, but still never completes the process. In debug mode, it shows the above "DEBUG" message every 5 seconds. Everything works fine after a second apply.

### Steps to Reproduce
1. `terraform get`
2. `terraform apply`

Works as expected if modules are not used. Could be related to https://github.com/hashicorp/terraform/issues/10014.

mitchellh:
Fixes #10122

The simple fix was that we forgot to close `ReadDataApply` for the
provider. But I've always felt that this section of the code was brittle
and I wanted to put in a more robust solution. The `shadow.Close` method
uses reflection to automatically close all values.

### DIFF
--- a/helper/shadow/closer.go
+++ b/helper/shadow/closer.go
@@ -1,0 +1,80 @@
+package shadow
+
+import (
+	"fmt"
+	"io"
+	"reflect"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/mitchellh/reflectwalk"
+)
+
+// Close will close all shadow values within the given structure.
+//
+// This uses reflection to walk the structure, find all shadow elements,
+// and close them. Currently this will only find struct fields that are
+// shadow values, and not slice elements, etc.
+func Close(v interface{}) error {
+	// We require a pointer so we can address the internal fields
+	val := reflect.ValueOf(v)
+	if val.Kind() != reflect.Ptr {
+		return fmt.Errorf("value must be a pointer")
+	}
+
+	// Walk and close
+	var w closeWalker
+	if err := reflectwalk.Walk(v, &w); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


helper/shadow/closer.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil
+	}
+
+	// We're looking for an io.Closer
+	raw := v.Interface()
+	if raw == nil {
+		return nil
+	}
+
+	var closer io.Closer
```
jbardin: don't need the declaration here, with the `:=` below

mitchellh:
Good catch. Fixed, merging.

Issue #9138, opened 2016-09-29
Feature Request: VPC Endpoint route table association resource


coen-hyde:
# Feature

A vpc endpoint route table association resource
# Why

So all route tables do not have to exist at the time the endpoint is created.
# My Use Case:

I use Terraform's remote state to logically group infrastructure. Having a this resource would allow me to create the endpoint in one deployment but associate route tables in another.

kwilczynski:
@coen-hyde hi there! Thank you for this! It is not a bad idea at all, actually.

I will have a look into it.

supergibbs:
I'd love this too but I assumed it didn't exist since the AWS CLI doesn't seem to support it:

http://docs.aws.amazon.com/cli/latest/reference/ec2/create-route.html

supergibbs:
Actually could be done with http://docs.aws.amazon.com/cli/latest/reference/ec2/modify-vpc-endpoint.html

kwilczynski:
@coen-hyde @supergibbs hi there! Yes, modifying the VPC endpoint is the key here, especially since the `route_table_ids` are already optional. A clear low hanging fruit. 👍

kwilczynski:
@coen-hyde @supergibbs hi! Apologies for the delay. I have began work in #9244.

coen-hyde:
@kwilczynski awsome! Hardly a delay. I wasn't expecting this to be worked on so soon

stack72:
Closed via #10137

PR #10137, opened 2016-11-15
Add new aws_vpc_endpoint_route_table_association resource

ewbankkit:
This PR follows on from #9244 and addresses #9138.
Here's the log from @kwilczynski's initial commit:
```
This commit adds a new resource which allows to a list of route tables to be
either added and/or removed from an existing VPC Endpoint. This resource would
also be complimentary to the existing `aws_vpc_endpoint` resource where the
route tables might not be specified (not a requirement for a VPC Endpoint to
be created successfully) during creation, especially where the workflow is
such where the route tables are not immediately known.
```
As it stands now the new `aws_vpc_endpoint_route_table_association` resource associates/dissociates as single `route_table_id` with a VPC endpoint.
Somehow I feel it's "better" Terraform to use a `count` attribute on the resource to associate multiple route tables with a VPC endpoint rather than accept a list of route tables in a single association, e.g.
```
resource "aws_vpc" "foo" {
    cidr_block = "10.0.0.0/16"
}

resource "aws_vpc_endpoint" "s3" {
    vpc_id = "${aws_vpc.foo.id}"
    service_name = "com.amazonaws.us-west-2.s3"
}

variable "c" {
    default = 2
}

resource "aws_route_table" "rt" {
    count = "${var.c}"

    vpc_id = "${aws_vpc.foo.id}"

    tags {
        Name = "test"
    }
}

resource "aws_vpc_endpoint_route_table_association" "a" {
    count = "${var.c}"

	vpc_endpoint_id = "${aws_vpc_endpoint.s3.id}"
	route_table_id  = "${element(aws_route_table.rt.*.id, count.index)}"
}
```
Since this is a "synthetic" resource (no underlying AWS resource maps exactly to this) this doesn't cause a cognitive disconnect, plus it greatly simplifies the implementation.
I welcome feedback on this, especially from @coen-hyde and @supergibbs who commented on the associated issue.

### DIFF
--- a/builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
+++ b/builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
@@ -1,0 +1,189 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsVpcEndpointRouteTableAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsVPCEndpointRouteTableAssociationCreate,
+		Read:   resourceAwsVPCEndpointRouteTableAssociationRead,
+		// "resource aws_vpc_endpoint_route_table_association: All fields are ForceNew or Computed w/out Optional, Update is superfluous"
+		// Update: resourceAwsVPCEndpointRouteTableAssociationUpdate,
+		Delete: resourceAwsVPCEndpointRouteTableAssociationDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"vpc_endpoint_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"route_table_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceAwsVPCEndpointRouteTableAssociationCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+	endpointId := d.Get("vpc_endpoint_id").(string)
+	rtId := d.Get("route_table_id").(string)
+
+	_, err := findResourceVPCEndpoint(conn, endpointId)
+	if err != nil {
+		return err
+	}
+
+	log.Printf(
+		"[INFO] Creating VPC Endpoint/Route Table association: %s => %s",
+		endpointId, rtId)
+
+	input := &ec2.ModifyVpcEndpointInput{
+		VpcEndpointId:    aws.String(endpointId),
+		AddRouteTableIds: aws.StringSlice([]string{rtId}),
+	}
+
+	_, err = conn.ModifyVpcEndpoint(input)
+	if err != nil {
+		return fmt.Errorf("Error creating VPC Endpoint/Route Table association: %s", err.Error())
+	}
+
+	return resourceAwsVPCEndpointRouteTableAssociationRead(d, meta)
+}
+
+func resourceAwsVPCEndpointRouteTableAssociationRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+	endpointId := d.Get("vpc_endpoint_id").(string)
+	rtId := d.Get("route_table_id").(string)
+
+	err := findResourceVPCEndpointRouteTableAssociation(conn, endpointId, rtId)
+	if _, notFound := err.(vpcEndpointNotFound); notFound {
+		// The VPC endpoint containing this association no longer exists.
+		d.SetId("")
+		return nil
+	}
+	if _, notFound := err.(vpcEndpointRouteTableAssociationNotFound); notFound {
+		// The association no longer exists.
+		d.SetId("")
+		return nil
+	}
+
+	id := vpcEndpointIdRouteTableIdHash(endpointId, rtId)
+	log.Printf("[DEBUG] Computed VPC Endpoint/Route Table ID %s", id)
+	d.SetId(id)
+
+	return nil
+}
+
+func resourceAwsVPCEndpointRouteTableAssociationDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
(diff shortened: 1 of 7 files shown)


ewbankkit:
Acceptance tests:
```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSVpcEndpointRouteTableAssociation_'
```

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
```diff
@@ -0,0 +1,189 @@
...
+	input := &ec2.ModifyVpcEndpointInput{
+		VpcEndpointId:    aws.String(endpointId),
+		AddRouteTableIds: aws.StringSlice([]string{rtId}),
+	}
+
+	_, err = conn.ModifyVpcEndpoint(input)
+	if err != nil {
+		return fmt.Errorf("Error creating VPC Endpoint/Route Table association: %s", err.Error())
+	}
+
```
jrnt30: I believe generally you should be calling the `d.SetId` here to populate the state, the fact that you call `Read` immediately makes it a bit less necessary but is the pattern I've seen around the code base.

ewbankkit: OK, I'll fix that.

ewbankkit:
Removing the **[WIP]** marker.
I am considering this as _ready for consideration_.

kwilczynski:
@ewbankkit hi there! Thank you for completing this when I could not do it. You are a ⭐️ !

I do agree that relying on `count` seem much more natural with Terraform - my idea of having the ability to pass multiple IDs definitely is not as flexible as using the `count` 👍 .

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsVpcEndpointRouteTableAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsVPCEndpointRouteTableAssociationCreate,
+		Read:   resourceAwsVPCEndpointRouteTableAssociationRead,
+		// "resource aws_vpc_endpoint_route_table_association: All fields are ForceNew or Computed w/out Optional, Update is superfluous"
```
kwilczynski: I would get rid of the comment here. If we are never going to use `UpdateFunc` then I would not bother with clarification.

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	_, err := conn.ModifyVpcEndpoint(input)
+	if err != nil {
+		ec2err, ok := err.(awserr.Error)
+		if !ok {
+			return fmt.Errorf("Error deleting VPC Endpoint/Route Table association: %s", err.Error())
+		}
+
+		switch ec2err.Code() {
```
ewbankkit: OK

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	return vpcEndpointRouteTableAssociationNotFound{endpointId, rtId}
+}
+
+func vpcEndpointIdRouteTableIdHash(endpointId, rtId string) string {
+	return fmt.Sprintf("a-%s%d", endpointId, hashcode.String(rtId))
+}
+
+type vpcEndpointNotFound struct {
```
ewbankkit: I'll revisit. I based that on another resource and maybe it's overkill in this situation.

ewbankkit:
Pushed changes after @kwilczynski's code review:
* Removed error types and simplified the error handling in `resourceAwsVPCEndpointRouteTableAssociationRead`
* Simplified logging in `resourceAwsVPCEndpointRouteTableAssociationDelete`

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
```diff
@@ -0,0 +1,151 @@
...
+	found := false
+	for _, id := range vpce.RouteTableIds {
+		if id != nil && *id == rtId {
+			found = true
+			break
+		}
+	}
+	if !found {
+		// The association no longer exists.
+		d.SetId("")
```
stack72: A quick follow up point for the future - we *usually* log a message here just to let the user know we have removed from state

stack72:
Hi @ewbankkit 

this LGTM! I left a few small questions inline - if you are happy with those then we can push forward

Thanks to @kwilczynski for the review as well

P.

ewbankkit:
Sorry, at re:Invent all last week. I'll address these now...

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
ewbankkit: OK. I won't change this now. Thanks for the tip.

builtin/providers/aws/resource_aws_vpc_endpoint_route_table_association.go
```diff
@@ -0,0 +1,151 @@
...
+		VpcEndpointIds: aws.StringSlice([]string{id}),
+	}
+
+	log.Printf("[DEBUG] Reading VPC Endpoint: %q", id)
+	output, err := conn.DescribeVpcEndpoints(input)
+	if err != nil {
+		return nil, err
+	}
+
+	return output.VpcEndpoints[0], nil
```
ewbankkit: Yes. The desired VPC endpoint ID is passed as an input parameter and the [API call](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpoints.html) will return either that VPC endpoint or a [client error](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#api-error-codes-table-client) such as `InvalidVpcEndpointId.Malformed` or `InvalidVpcEndpointId.NotFound`.
We should be good here.

stack72:
Hi @ewbankkit 

Thanks for the work here (and the responses to my nit questions)

This LGTM!

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSVpcEndpointRouteTableAssociation_'                                                                                               ✹
2016/12/05 12:53:35 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSVpcEndpointRouteTableAssociation_ -timeout 120m
=== RUN   TestAccAWSVpcEndpointRouteTableAssociation_basic
--- PASS: TestAccAWSVpcEndpointRouteTableAssociation_basic (50.68s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	50.700s
```

PR #10140, opened 2016-11-15
Update index.html.markdown

jjasghar:
Added a note about self-signed certs and vsphere.


stack72:
Hi @jjasghar 

Paul

PR #10142, opened 2016-11-15
provider/github: add GitHub labels resource

MrSaints:
In terms of use-case, I'm not entirely sure how useful this'll be. But, I quite like the idea of being able to quickly manage all labels for current, and future repositories in a consistent manner. The default labels GitHub provides aren't ideal. Understandably, this doesn't wipe the default labels.

Thoughts, and suggestions welcomed (especially with regards to naming).

---

This resource allows easy management of issue labels for an organisation's repositories. A name, and a color can be set. These attributes can be updated without creating a new resource.


stack72:
Thanks for the work here @MrSaints :)

This LGTM!

MrSaints:
Cheers @stack72 😄

PR #10145, opened 2016-11-15
Doc Update for ELB defaults

jrnt30:
- Doc update to explicitly call out ELB defaults (so I don't have to look
at the source again :)


stack72:
Hi @jrnt30 

Paul

PR #10149, opened 2016-11-15
Add debug command with json2dot

jbardin:
Add `terraform debug json2dot` to convert debug log graphs to dot
format. This is not meant to be in place of more advanced debug
visualization, but may continue to be a useful way to work with the
debug output.


jbardin:
Added a command test for `debug json2dot`

PR #10151, opened 2016-11-15
provider/docker: authentication via values instead of files

mkuzmin:
This PR adds new parameters for configuring remote Docker host for encryption and client authentication via inline values instead of external files.
```
provider "docker" {
  host = "tcp://hostname:2376"
  ca = "${tls_self_signed_cert.docker_ca.cert_pem}"
  cert = "${tls_locally_signed_cert.docker_client.cert_pem}"
  key = "${tls_private_key.docker_client.private_key_pem}"
}
```

This allows referencing other resources, for example from TLS and Vault providers.

See a complete example at https://github.com/mkuzmin/terraform-demo/tree/a851f57feda0e694e6a908baff019d6ea79b0316

### DIFF
--- a/builtin/providers/docker/config.go
+++ b/builtin/providers/docker/config.go
@@ -21,7 +21,7 @@ type Config struct {
 func (c *Config) NewClient() (*dc.Client, error) {
 	if c.Ca != "" || c.Cert != "" || c.Key != "" {
 		if c.Ca == "" || c.Cert == "" || c.Key == "" {
-			return nil, fmt.Errorf("ca, cert, and key must be specified")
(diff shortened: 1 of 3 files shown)


stack72:
Hi @mkuzmin 

Thanks so much for the PR here - so, we have been moving towards doing something like `key_material` rather than paths to files. I just merged the work for moving the chef provider towards this for 0.8 (#10105)

If possible, could you follow the same pattern here?

Thanks

Paul

mkuzmin:
I've renamed all three parameters, and also add handling for environment variables.
Please take a look.

stack72:
Looks good to me now :) Thanks for following up so fast on this :)

P.

PR #3507, opened 2015-10-15
provider/google: one more fix to GCE metadata

phinze:
In #3501 @lwander got us almost all the way there, but we still had
tests failing. This seemed to be because GCE sets
`metadata.startup-script` to a blank string on instance creation, and if
a user specifies any `metadata` in their config this is seen as the
desired full contents of metadata, so we get a diff trying to remove
`startup-script`.

Here, to address this, we just proactively remove the "startup-script"
key from `Read`, and then we enforce that "metadata_startup_script"
is the only way to configure startup scripts on instances.

PR #10167, opened 2016-11-16
google_compute_instance shouldn't favor metadata_startup_script over startup-script key in metadata

### LINKED ISSUE #9999: google_compute_instance: google_compute_instance.junk: Use metadata_startup_script instead of a startup-script key in "metadata"
### Terraform Version
`Terraform v0.7.9`

### Affected Resource(s)
- google_compute_instance

```hcl
resource "google_compute_instance" "junk" {
  name           = "inf-junk"
  machine_type   = "n1-standard-4"
  can_ip_forward = false
  zone           = "us-central1-b"

  tags = ["server"]

  disk {
    image = "centos-7"
  }

  network_interface {
    subnetwork = "us-central1"
    access_config {
      // Ephemeral IP
    }
  }

  metadata {
    startup-script = "${data.template_file.startup-script.rendered}"
  }
}
```

### Expected Behavior
This should work per the docs so we don't have to recreate instances when we change the startup script:
metadata_startup_script - (Optional) An alternative to using the startup-script metadata key, except this one forces the instance to be recreated (thus re-running the script) if it is changed. This replaces the **startup-script metadata key** on the created instance and thus the two mechanisms are not allowed to be used simultaneously.

### Actual Behavior
Forces you to use `metadata_startup_script` which forces a rebuild of the instance.

### References
https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance.go#L1005

JDiPierro:
Fixes #9999

### DIFF
--- a/builtin/providers/google/resource_compute_instance.go
+++ b/builtin/providers/google/resource_compute_instance.go
@@ -111,10 +111,9 @@ func resourceComputeInstance() *schema.Resource {
 			},
 
 			"metadata": &schema.Schema{
-				Type:         schema.TypeMap,
-				Optional:     true,
-				Elem:         schema.TypeString,
-				ValidateFunc: validateInstanceMetadata,
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem:     schema.TypeString,
 			},
 
 			"metadata_startup_script": &schema.Schema{
@@ -996,13 +995,4 @@ func resourceInstanceTags(d *schema.ResourceData) *compute.Tags {
 	}
 
 	return tags
-}
-
-func validateInstanceMetadata(v interface{}, k string) (ws []string, es []error) {
-	mdMap := v.(map[string]interface{})
-	if _, ok := mdMap["startup-script"]; ok {
-		es = append(es, fmt.Errorf(
-			"Use metadata_startup_script instead of a startup-script key in %q.", k))
-	}
-	return
 }


paddyforan:
See also #8407 :)

cblecker:
@JDiPierro --
What happens if you use both? I'm assuming one overwrites the other which wouldn't be good.

paddyforan:
Oh, also, sounds like this breaks some stuff: #3507.

Sorry, I'm working right now to try to find a good answer to this problem that can get us on a forward path again.

cblecker:
@paddyforan --
I think @evandbrown's suggestion [here](https://github.com/hashicorp/terraform/pull/8407#issuecomment-257723248) is probably the best way forward. Remove the metadata startup script, and do a state migration to put that key in one place. Just have instance metadata be in the metadata block, and work the same if you're using an instance or instance_group.

paddyforan:
Hey @cblecker,

In investigating this issue, I actually looked into that option--looking back at the original pull request, it looks like the metadata_startup_script attribute was included in #2375, because there's no way to make `metadata.startup-script` force instance recreation, which is the AWS behaviour. So I'm hesitant to roll back to that point without some discussion of pros and cons. There are a few ways forward that I can see, I'm just seeking some wisdom from some more experienced folks on the Terraform team before I commit to any of them. Apologies for the lengthy process, but I promise I'm investigating and trying to get this moving again. We have (from what I can tell) about 3 open PRs and an open issue that are all centered around this, so it has drawn my attention. :)

cblecker:
Totally understand the need for discussion and review! Just wanted to bring Evan's piece into it. I'd also note that forcing the instance to recreate when metadata is updated isn't always preferable in the GCP world (where as it might be in AWS). You can update metadata manually on an instance in Google and it doesn't force instance recreation.

Look forward to hearing what the result of those internal talks is :)

paddyforan:
Appreciate the direction. And the note; I totally see the value in updating the startup script without recreating things. I know it certainly makes sharing configuration code between instances and instance templates trickier, as forcing recreation of an instance template can have some pretty wide consequences. It's a balancing act at this point of trying to manage complexity while supporting as many use cases as possible, and prioritising those use cases.

I'm trying not to keep this as an internal discussion, but I did want to get feedback from other Terraform team members before I put my foot in my mouth, given how complicated this trade-off is.

JDiPierro:
re: `What happens if you use both?`
The provider yells at you that only one can be specified: https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance.go#L793

Glad to see this is being worked on, hope to have a clean way to manage this in a future version :) Feel free to close this PR in favor of a future one that addresses the issue.

paddyforan:
This should be covered by #10537. :) Thanks for the PR!

Issue #10168, opened 2016-11-16
aws_vpc: enable_dns_support not set correctly at initial creation


mattmoyer:
### Terraform Version
v0.7.11

### Affected Resource(s)
- aws_vpc

### Terraform Configuration Files
```hcl
resource "aws_vpc" "test" {
  cidr_block           = "10.0.0.0/16"
  enable_dns_support   = false
}

output "vpc_id" {
    value = "${aws_vpc.test.id}"
}
```

### Debug Output
https://gist.github.com/mattmoyer/38cc4322ed6a4ee5dc128e34935d61a8

### Expected Behavior
The VPC should be created with `enableDnsSupport` set to `false` during initial creation.

### Actual Behavior
The VPC is initially created with`enableDnsSupport` set to `true`, which is fixed by re-running `terraform apply`.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` (initial apply using the `.tf` linked above)
2. `aws ec2 describe-vpc-attribute --vpc-id $(terraform output vpc_id) --attribute enableDnsSupport` (shows that DNS support was enabled).
3. `terraform apply` (second apply, notices the discrepancy and fixes it)
4. `aws ec2 describe-vpc-attribute --vpc-id $(terraform output vpc_id) --attribute enableDnsSupport` (shows that DNS support was correctly disabled).

I haven't really looked at the resource yet but I can probably take a shot at fixing this. Would you want it cherry picked into 0.7.x?

This could potentially be considered a security vulnerability since one of the reasons to disable DNS support is to lock down egress from a VPC.

catsby:
Fixed in #10171 , thanks!

mattmoyer:
Thanks @Ninir!

PR #10171, opened 2016-11-16
provider/aws: Fixed the aws_vpc enable_dns_support handling on creation

Ninir:
Fixes https://github.com/hashicorp/terraform/issues/10168

This is an attempt in fixing the related issue. Still experimenting Go & Terraform codebase, be kind 😅

### DIFF
--- a/builtin/providers/aws/resource_aws_vpc.go
+++ b/builtin/providers/aws/resource_aws_vpc.go
@@ -23,59 +23,59 @@ func resourceAwsVpc() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"cidr_block": &schema.Schema{
+			"cidr_block": {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
 				ValidateFunc: validateCIDRNetworkAddress,
 			},
 
-			"instance_tenancy": &schema.Schema{
+			"instance_tenancy": {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
 				Computed: true,
 			},
 
-			"enable_dns_hostnames": &schema.Schema{
+			"enable_dns_hostnames": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Computed: true,
 			},
 
-			"enable_dns_support": &schema.Schema{
+			"enable_dns_support": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Computed: true,
 			},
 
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_vpc.go
```diff
@@ -260,7 +260,9 @@ func resourceAwsVpcUpdate(d *schema.ResourceData, meta interface{}) error {
 		d.SetPartial("enable_dns_support")
 	}
 
-	if d.HasChange("enable_dns_support") {
+	_, hasEnableDnsSupportOption := d.GetOk("enable_dns_support")
+
+	if !hasEnableDnsSupportOption || d.HasChange("enable_dns_support") {
```
Ninir: In fact, this will be the same for every attributes in this function, right?

mattmoyer:
I confirmed that this does fix the buggy behavior, but I'm not sure I understand how.

It's worth noting that I _don't_ see the buggy behavior with `enable_dns_hostnames`, I think because it defaults to `false` on new VPCs whereas `enable_dns_support` defaults to `true`.

I don't 100% understand the bug but I suspect it's related to the initial default value not being properly reflected in the initial resource data structure when it's first passed to `resourceAwsVpcUpdate`.

Ninir:
@mattmoyer In fact it does the following thing from the code POV:

1. It [creates the VPC](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_vpc.go#L95) without the option (it is how the API is designed).
2. Since there are no modifications on the given attribute, it is not written to the state.
3. After creating the resource, it [updates the resource](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_vpc.go#L131) (in order to add attributes).
4. Since [it does not have the attribute in the state](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_vpc.go#L263), this line resolves to false, because the value has not changed.
5. When you apply again, [Terraform will read the value from AWS](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_vpc.go#L167), and will set the retrieved value (true)
6. Since the value is different from your config one, it will warn about the update

Hope it explains better what happens!

builtin/providers/aws/resource_aws_vpc.go
catsby: seems `enable_dns_support` is the only attribute that defaults `true` on AWS side here, which is why this is the only one that needs this check (if I understand correctly)

PR #10178, opened 2016-11-16
Add name_prefix to aws_iam_policy

### LINKED ISSUE #10176: Please implement name_prefix for aws_iam_policy
Terraform supports a [`name_prefix` for `aws_iam_role`](https://www.terraform.io/docs/providers/aws/r/iam_role.html#name_prefix) but it doesn't support one for `aws_iam_policy`. This forces us to create our own unique names so we can apply our same terraform configuration with multiple state files in the same account. It would be nice if we could use the uniqueness-providing behavior that `aws_iam_role` has because of `name_prefix`.

2rs2ts:
This is intended to behave the same way as name_prefix does in `aws_iam_role`.

Fixes #10176

Feedback on the interface change would be highly appreciated. I was able to get the acceptance tests to run successfully thanks to some help from a peer, so I feel pretty confident about this.

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_policy.go
+++ b/builtin/providers/aws/resource_aws_iam_policy.go
@@ -2,11 +2,13 @@ package aws
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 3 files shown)


stack72:
Hi @2rs2ts 

This LGTM! Thanks for the work here:

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAWSPolicy_namePrefix'
2016/11/17 14:00:36 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAWSPolicy_namePrefix -timeout 120m
=== RUN   TestAWSPolicy_namePrefix
--- PASS: TestAWSPolicy_namePrefix (15.32s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	15.341s
```

Issue #10174, opened 2016-11-16
google_compute_instance_group_manager update_strategy not properly read


JDiPierro:
### Terraform Version
`0.7.11`

### Affected Resource(s)
- google_compute_instance_group_manager

### Terraform Configuration Files
```hcl
resource "google_compute_instance_template" "test" {
  name_prefix = "test-"
  machine_type = "f1-micro"
  can_ip_forward = false
  region = "us-central1"

  disk {
    source_image = "centos-7"
  }

  network_interface {
    access_config {
      // Ephemeral IP
    }
  }

  lifecycle {
    create_before_destroy = true
  }

  service_account {
    email = "compute@project.iam.gserviceaccount.com"
    scopes = ["cloud-platform"]
  }
}

resource "google_compute_instance_group_manager" "test" {
  name = "test"
  instance_template = "${google_compute_instance_template.test.self_link}"
  update_strategy = "NONE"
  base_instance_name = "test"
  zone = "us-central1-b"
  target_size = "1"
}
```

### Expected Behavior
The instance group manager persists the `"NONE"` setting into the `tfstate`.

### Actual Behavior
"RESTART" is saved into the `tfstate`. All subsequent plans will show a pending change:

    ~ google_compute_instance_group_manager.test
          update_strategy: "RESTART" => "NONE"

### References

Looks like [this line](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance_group_manager.go#L245) is causing the problem. After update we go to read and read defaults `update_strategy` to `"REFRESH"`. I think this should be attempting to pull the `update_strategy` using `d.Get(...)` and set it to the default if `nil` is returned.

evandbrown:
LGTM, thanks @paddyforan!

paddyforan:
Het @evandbrown, is that for #11013? Just want to be sure before I make assumptions. :P

stack72:
Closed via #11013

PR #10198, opened 2016-11-17
Check if the user set 'update_strategy' before using the default.

JDiPierro:
For #10174 which I reported yesterday :) I'm new to Go so please let me know if this can be written any cleaner.


paddyforan:
Hi @JDiPierro, sorry for the delay on the review of this. It looks great, but I wanted to see tests for the failing case (to make sure we didn't regress in the future). I noticed it had been a while since we heard from you in #10169, so I figured you ran out of time to contribute (which is understandable and okay!) Because this was such a small change, I just pulled it into #11013 to try and get it fixed. Hopefully that's OK. :)

Thanks for opening the PR and the issue! Feel free to let me know if you have any questions or concerns.

PR #10206, opened 2016-11-17
Checkpoint signature fixes

jrnt30:
- Currently the disable_checkpoint setting from $HOME/.terraformrc is never
respsected due to:
-- The runCheckpoint go routine being fired off prior to loading configuration
-- The config.Merge method not actually merging in the c2s settings


mitchellh:
LGTM, needs some tests which I'll add. Thanks!

PR #10224, opened 2016-11-18
ECS service ignore taks definition

zimbatm:
On AWS ECS services, allow to ignore task definition changes. This is to
provide an interface where terraform manages everything except the task
definitions and deploys to the services. You would have to provision the
initial services with a blank task definition.

```tf
# A noop-like task definition to bootstrap the cluster
resource "aws_ecs_task_definition" "null" {
  family = "null"

  container_definitions = <<EOF
[
  {
    "name": "null",
    "image": "busybox",
    "memory": 4,
    "essential": true,
    "command": ["sleep", "100000"]
  }
]
EOF
}

resource "aws_ecs_service" "example" {
  name                       = "example"
  cluster                     = "${aws_ecs_cluster.default.id}"
  task_definition        = "${aws_ecs_task_definition.null.arn}"
  # Allow to update task definitions independently
  update_task_definition = false
  desired_count          = 1
}
```

### DIFF
--- a/builtin/providers/aws/resource_aws_ecs_service.go
+++ b/builtin/providers/aws/resource_aws_ecs_service.go
@@ -44,6 +44,12 @@ func resourceAwsEcsService() *schema.Resource {
 				Required: true,
 			},
 
+			"update_task_definition": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
 			"desired_count": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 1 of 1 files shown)


zimbatm:
Looking for 20% percent review

catsby:
Hello @zimbatm – I'm curious , would using `ignore_changes` work here, instead? 

- https://www.terraform.io/docs/configuration/resources.html#ignore-changes

Ex: 

```
resource "aws_ecs_task_definition" "null" {
  family = "null"

  container_definitions = <<EOF
[
  {
    "name": "null",
    "image": "busybox",
    "memory": 4,
    "essential": true,
    "command": ["sleep", "100000"]
  }
]
EOF

  lifecycle {
    ignore_changes = ["container_definitions"]
  }
}
```

zimbatm:
OMG! I think you're right

zimbatm:
Yep. Neverming then, sorry for the noise.

catsby:
@zimbatm not at all! Thanks for opening the PR, even if not accepted we appreciate the thought, time, and energy 😄

Issue #10187, opened 2016-11-17
Resource map-typed arguments various crashes


moio:
### Terraform Version
0.8.0 dev, current master (7293630f644dc8f2c1db41730f70a7e6762b3e60)

### Affected Resource(s)
Terraform core, any resource which has an argument of type map (eg. [heroku_app's config_vars](https://www.terraform.io/docs/providers/heroku/r/app.html#config_vars)).

### Terraform Configuration Files
```hcl
provider "heroku" {
    email = "dummy"
    api_key = "dummy"
}

variable "app_config" {
  type = "list"

  default = [
    {
      foo = "bar"
      baz = "bam"
    },
    {
      foo = "bar"
      baz = "bam"
    },
  ]
}
  
resource "heroku_app" "test" {
  name   = "test-app"
  region = "us"

  // Scenario 1
  // this works partially (number of resources is 1, should be 2, see #8867):
  #config_vars = ["${var.app_config}"]

  // this leads to terraform crash (expected type error):
  #config_vars = "${var.app_config}"

  // Scenario 2
  // this works:
  #config_vars = ["${map("foo", "bar")}"]

  // this leads to terraform crash (expected type error):
  #config_vars = "${list(map("foo", "bar"))}"
    
  // Scenario 3
  // this works (type error: should be a map):
  #config_vars = ["${map("foo", "bar")}"]
  #config_vars = ["lalala"]

  // this leads to misleading error (unknown type for string *ast.ListType):
  #config_vars = ["${map("foo", "bar")}"]
  #config_vars = "lalala"
}
```

### Expected Behavior
See comments in scenarios above.

### Actual Behavior
See comments in scenarios above.

### Steps to Reproduce
1. uncomment one or more lines according to comments above
1. `terraform plan`

### Important Factoids
 - I am not sure if this should be reported as 3 separate issues, or if cause is common. Please advise.
 - I have another similar bug I cannot presently reproduce with builtin providers, only with terraform-provider-libvirt. I am fairly sure it is a bug in core - should I open an issue for that too?

### References
- GH-9549: these issues emerged while testing resolution of this issue

moio:
@jbardin great, thanks so far!

What about scenario 1, where plan is inconsistent on the number of vars? Even in latest terraform I get (emphasis mine):

> + heroku_app.test
>     all_config_vars.%: "<computed>"
>     config_vars.#:     "**2**"
>     config_vars.**0**.%:   "2"
[...]

PR #10228, opened 2016-11-18
ResourceConfig.get should never return (nil, true)

jbardin:
Fixes a case where ResourceConfig.get inadvertently returns a nil value.

Add an integration test where assigning a map to a list via
interpolation would panic.

This fixes the crashes seen in #10187. 

The silent errors in the list-of-maps corner cases may need to wait until after 0.8, when we are planning on a larger-scale type refactor, removing the dotted string format.

### DIFF
--- a/terraform/resource.go
+++ b/terraform/resource.go
@@ -316,7 +316,8 @@ func (c *ResourceConfig) get(
 			// prefix so were split as path components above.
 			actualKey := strings.Join(parts[i-1:], ".")
 			if prevMap, ok := previous.(map[string]interface{}); ok {
-				return prevMap[actualKey], true
+				v, ok := prevMap[actualKey]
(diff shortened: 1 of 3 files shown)


terraform/resource.go
```diff
@@ -316,7 +316,8 @@ func (c *ResourceConfig) get(
 			// prefix so were split as path components above.
 			actualKey := strings.Join(parts[i-1:], ".")
 			if prevMap, ok := previous.(map[string]interface{}); ok {
-				return prevMap[actualKey], true
+				v, ok := prevMap[actualKey]
+				return v, ok
```
mitchellh: In addition to the above test we should be able to reproduce this with a `ResourceConfig.Get` test, I'd feel much safer if we did that in addition to the above.

PR #10233, opened 2016-11-18
An empty module in state can panic

### LINKED ISSUE #10229: panic: runtime error: slice bounds out of range
### Terraform Version
0.7.10,0.7.11

### Affected Resource(s)
The issue affects multiple resources -- all that I've tried to test. It also occurs with no resources in the project (but still with the state file).

Corporate policy greatly restricts my ability to share information about the project. I'm happy to answer questions.

### Panic Output
https://gist.github.com/th7/08cb04196b8d086d967f4b11f28e8678

### Expected Behavior
A plan should have been generated.

### Actual Behavior
Panic

### Steps to Reproduce
1. `terraform plan -state=mystatefile.tfstate`

### Important Factoids
The issue even happens when I remove everything but the provider "aws" from the project. Refreshing state seems to work without issue, the panic happens shortly thereafter.

I've tried refreshing my state with incrementing versions of terraform. refresh/plan works from 0.7.6-0.7.9. The state file gets changes for the newer versions, but still breaks going from 0.7.9 to 0.7.10/0.7.11.

The issue doesn't occur against my staging environment. Staging / prod are not fully identical, so if the stack trace and info I've provided so far doesn't lead to any clear problems, I can do some more investigation into those differences and try to isolate a cause.

jbardin:
An empty module, or a module with an empty path can panic during graph
traversal. Make sure we clear these out when reading and writing the
state.

Fixes #10229

### DIFF
--- a/state/cache_test.go
+++ b/state/cache_test.go
@@ -1,6 +1,7 @@
 package state
 
 import (
+	"fmt"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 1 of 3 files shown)


mitchellh:
I agree with the test, but I think the implementation isn't safe enough:

This only clears the orphans on `init()`, but we've had past implementations (current old graph!) that can leave around empty module states after that. We should clean these up on `prune()`. We can safely call `prune()` from `init()` too, probably.

`ModuleByPath` can be considered to maybe call prune... for safety?

jbardin:
OK, let me look at these closer. The `init()` method should cover what we want here, because it's become more of a `normalize()` than just init, and is called after Read, before Write, and so on. Looks like some of this functionality needs to be split or combined with prune and cleaned up.

jbardin:
Moved the module pruning to `prune()`, and call that now during `init()`. 

Did a little cleanup if the prune/init dichotomy, and only had to fixup one test where we expected an empty module to hang around in the state.

PR #10249, opened 2016-11-20
fix aws_availability_zone data source doc example

ryane:
existing example returns an error like the following should you try to run `terraform plan` against it:

```
Error reading config for aws_subnet[example]: data.aws_availability_zone.name_suffix: data variables must be four parts: data.TYPE.NAME.ATTR in:

${cidrsubnet(aws_vpc.example.cidr_block, 4, var.az_number[data.aws_availability_zone.name_suffix])}
```


stack72:
Hi @ryane 

Thanks for the work here -this LGTM!

Paul

PR #10256, opened 2016-11-20
Closes #1768 - Implements taint with wildcard

### LINKED ISSUE #1768: Taint with wildcard?
Hey guys,

I've got an AWS instance resource for my workers in a module (`mymodule`):

```bash
resource "aws_instance" "worker" {
  count = "12"
 ...
}
```
Is it possible to taint all 12 resources? I've tried the following:

```bash
❯ terraform taint -module=mymodule "aws_instance.worker.*"
The resource aws_instance.worker.* couldn't be found in the module root.mymodule.
```

Rather than having to do this:
```bash
❯ terraform taint -module=mymodule "aws_instance.worker.0"
The resource aws_instance.worker.0 in the module root.mymodule has been marked as tainted!

❯ terraform taint -module=mymodule "aws_instance.worker.1"
The resource aws_instance.worker.1 in the module root.mymodule has been marked as tainted!

...
```

Cheers,
Alex

artburkart:
This implementation of wildcard functionality enables the following:

Given resources:
- test_instance.foo.a
- test_instance.foo.b
- test_instance.bar.a

Matches:

`test_instance.*`
- test_instance.foo.a
- test_instance.foo.b
- test_instance.bar.a

`test_instance.*.a`
- test_instance.foo.a
- test_instance.bar.a

`test_instance.foo.*`
- test_instance.foo.a
- test_instance.foo.b

`test_instance.foo*`
- Doesn't match anything

I'm sure I've broken some conventions along the way. I very much appreciate any feedback that is given. 

Cheers

### DIFF
--- a/command/taint.go
+++ b/command/taint.go
@@ -104,6 +104,67 @@ func (c *TaintCommand) Run(args []string) int {
 		return 1
 	}
 
+	// Taint resources using wildcard
+	if strings.Contains(name, ".*") {
+		nameParts := strings.Split(name, ".")
+		tainted := []string{}
+
+		for k, _ := range mod.Resources {
+			resourceParts := strings.Split(k, ".")
+			// Verify it's possible for name to match resource
+			if len(resourceParts) < len(nameParts) {
+				continue
+			}
+			if len(resourceParts) > len(nameParts) && !strings.HasSuffix(name, ".*") {
+				continue
+			}
+
+			// Verify whether resource paths match
+			isMatch := true
+			for i, p := range nameParts {
+				if p != resourceParts[i] && p != "*" {
+					isMatch = false
+					break
+				}
+			}
+
+			// Mark the resource for tainting if it matches
+			if isMatch {
+				rs, _ := mod.Resources[k]
+				rs.Taint()
+				tainted = append(tainted, k)
+			}
+		}
+
+		// Check whether it's okay if no resources were tainted
+		if len(tainted) == 0 {
+			if allowMissing {
+				return c.allowMissingExit(name, module)
+			}
+
+			c.Ui.Error(fmt.Sprintf(
+				"No resources matching pattern %s could be found in the module %s.",
+				name,
+				module))
+			return 1
+		}
+
+		// Attempt to write new state with tainted resources
+		log.Printf("[INFO] Writing state output to: %s", c.Meta.StateOutPath())
+		if err := c.Meta.PersistState(s); err != nil {
+			c.Ui.Error(fmt.Sprintf("Error writing state file: %s", err))
+			return 1
+		}
+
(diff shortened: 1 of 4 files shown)


afeld:
Nice timing - was just looking to see if this was possible. Just a suggestion: Might want to add a wildcard example to [the web page](https://github.com/hashicorp/terraform/blob/master/website/source/docs/commands/taint.html.markdown). Thanks!

artburkart:
Yep! I forgot to update the docs to reflect the change. I'll do that soon.

kwilczynski:
@artburkart hi there! Thank you for that! This is amazing! <3

artburkart:
Sorry, I'll make the change soon! I'm visiting relatives right now and don't wanna be rude.

artburkart:
Finally got around to adding some docs. If anyone thinks they should be more thorough, please tell me. Cheers!

mitchellh:
Hey @artburkart. Thanks a lot!

Your tests here are 100% perfect. Great job. :) Unfortunately your implementation though, while it works, isn't how we want to approach this. Instead, we'd prefer that we centralize on addressing/targeting in the core itself. We're currently at a messy state internally where we don't have a single clean targeting parser/structure. 

This has been the main holdup for this PR since while it _does_ add the feature -- no doubt -- it also adds a bit of technical debt for us. 😕 And that's tricky.

What I'd like to do is keep your tests (perfect, as I said), but reimplement this in the core. This probably won't happen super fast though and this PR is already working, so it puts us in an awkward position... but historically we've rejected PRs on this grounds and asked that folks who really want it manually compile it in (benefits of OSS) while we architect out the better long term solution.

artburkart:
@mitchellh,

I figured as much; especially considering the fact that this code would also need to be copied and pasted into `untaint`. I won't be offended at all if you close this PR. 😄 

If I'm understanding you correctly, however, this means there is an interest in adding consistency to the way we target (and `-target`) resources, so it's no longer `aws_instance.foo.0` or `aws_instance.foo[0]` depending on the command you're using? I noticed there were comments in the code indicating an interest in changing how state is managed generally. Are there issues in GitHub that point to this effort?

I guess I'm curious because I was actually considering taking the time to gain a good, holistic understanding of the targeting/parser structure and figuring out what could maybe be done.

mitchellh:
Thanks for understanding @artburkart.

For clarity: when we do do the longer term work, I'll be sure to make sure to cherry pick your tests so you can get contributor recognition with the project and so you don't feel we're just freeloading your work. We really appreciate what you did here. 

And you are understanding correctly: we need to unify so that its the same syntax to reference something everywhere. 

I'll close this now, but this is something we do want and I'll come back to cherry pick your tests at some point. :)

PR #10252, opened 2016-11-20
AWS EMR Autotermination and Termination protection

xt84:
* Add documentation 
* Autoterminate
* Termination protection

Fixes #10263

PR #10257, opened 2016-11-20
[WIP] adding ebs volume support emr

sepulworld:
Looking to add EMR EBS volume support to aws_emr_cluster and aws_emr_instance_group

### DIFF
--- a/builtin/providers/aws/resource_aws_emr_cluster.go
+++ b/builtin/providers/aws/resource_aws_emr_cluster.go
@@ -144,25 +144,15 @@ func resourceAwsEMRCluster() *schema.Resource {
 						"size": &schema.Schema{
 							Type:     schema.TypeInt,
 							Required: true,
-							Default:  50,
 						},
 						"type": &schema.Schema{
 							Type:     schema.TypeString,
-							Required: true,
-							Default:  "standard",
-						},
-						"iops": &schema.Schema{
-							Type:     schema.TypeInt,
 							Optional: true,
+							Default:  "standard",
 						},
 						"volumes_per_instance": &schema.Schema{
 							Type:     schema.TypeInt,
-							Required: true,
-							Default:  1,
-						},
-						"optimized": &schema.Schema{
-							Type:     schema.TypeInt,
-							Required: true,
+							Optional: true,
 							Default:  1,
 						},
 					},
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_emr_cluster.go
```diff
@@ -135,6 +135,29 @@ func resourceAwsEMRCluster() *schema.Resource {
 					},
 				},
 			},
+			"ebs_volume": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"size": &schema.Schema{
```
kwilczynski: In Go 1.7+ we can skip the explicit type, thus `&schema.Schema` can be removed.

builtin/providers/aws/resource_aws_emr_cluster.go
```diff
@@ -225,6 +248,46 @@ func resourceAwsEMRClusterCreate(d *schema.ResourceData, meta interface{}) error
...
+				},
+			},
+		}
+
+		*instanceConfigEbs = &emr.JobFlowInstancesConfig{
+			MasterInstanceType:          aws.String(masterInstanceType),
+			SlaveInstanceType:           aws.String(coreInstanceType),
+			InstanceCount:               aws.Int64(int64(coreInstanceCount)),
+			KeepJobFlowAliveWhenNoSteps: aws.Bool(true),
+			TerminationProtected:        aws.Bool(false),
```
kwilczynski: Do we want to expose this? Would someone find it useful?

sepulworld: I think KeepJobFlowAliveWhenNoSteps wouldn't be used for Terraform managed clusters.

Giving an option to change TerminationProtected on/off would be great though.

builtin/providers/aws/resource_aws_emr_cluster.go
sepulworld: Ok, so it would look like this instead?

```
"size":  {
    Type:     schema.TypeInt,
    Required: true,
 },
```

kwilczynski: Yes :)

builtin/providers/aws/resource_aws_emr_cluster.go
kwilczynski: Indeed. It might be useful to others.

kwilczynski:
@sepulworld hi there! I am not sure if #10252 would not conflict here?

sepulworld:
Good looking out @kwilczynski. I think that PR is basically done, while mine still needs work. So, perhaps Ill rebase once #10252 merges into master. Plan?

serdardalgic:
This PR can be closed, right? https://github.com/hashicorp/terraform/pull/11411 Here the EBS support for EMR Instance groups is implemented and released with 0.8.6

PR #10258, opened 2016-11-20
Use RDS::CreateDBInstance docs for more info on parameter values

jagregory:
The docs for the CreateDBInstance API call include quite a bit more information about each individual option, for example `Engine` has each of the possible options listed, whilst the cli reference doesn't include this level of detail.


jen20:
Good call - thanks for the contribution @jagregory!

PR #10259, opened 2016-11-20
provider/openstack: openstack_blockstorage_volume_attach_v2 resource

jtopjian:
This commit adds the openstack_blockstorage_volume_attach_v2 resource. This
resource enables a volume to be attached to an instance by using the OpenStack
Block Storage (Cinder) v2 API.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_blockstorage_volume_attach_v2.go
+++ b/builtin/providers/openstack/resource_openstack_blockstorage_volume_attach_v2.go
@@ -1,0 +1,254 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/volumeactions"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceBlockStorageVolumeAttachV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceBlockStorageVolumeAttachV2Create,
+		Read:   resourceBlockStorageVolumeAttachV2Read,
+		Update: nil,
+		Delete: resourceBlockStorageVolumeAttachV2Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				DefaultFunc: schema.EnvDefaultFunc("OS_REGION_NAME", ""),
+			},
+
+			"volume_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"instance_id": &schema.Schema{
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"host_name"},
+			},
+
+			"host_name": &schema.Schema{
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"instance_id"},
+			},
+
+			"device": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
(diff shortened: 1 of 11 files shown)


builtin/providers/openstack/resource_openstack_blockstorage_volume_attach_v2.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceBlockStorageVolumeAttachV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceBlockStorageVolumeAttachV2Create,
+		Read:   resourceBlockStorageVolumeAttachV2Read,
+		Update: nil,
```
stack72: If everything is now ForceNew, we can drop this - the schema is more relaxed about this now :)

builtin/providers/openstack/resource_openstack_blockstorage_volume_attach_v2.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	return nil
+}
+
+func blockStorageVolumeAttachV2AttachMode(v string) (volumeactions.AttachMode, error) {
+	var attachMode volumeactions.AttachMode
+	var attachError error
+	switch v {
+	case "":
```
stack72: Gotta say I am not a *huge* fan of this - this is a weird way of adding the default. IMO I Think it should be in the schema - but that's just IMO

website/source/docs/providers/openstack/r/blockstorage_volume_attach_v2.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+
+* `instance_id` - (Required if `host_name` is not used) The ID of the Instance
+  to attach the Volume to.
+
+* `host_name` - (Required if `instance_id` is not used) The host to attach the
+  volume to.
+
+* `device` - (Optional) The device to attach the volume as.
+
+* `attach_mode` - (Optional) Specify whether to attach the volume as Read-Only
```
stack72: We don't tell people that we default to `ro`

builtin/providers/openstack/resource_openstack_blockstorage_volume_attach_v2.go
jtopjian: Just a habit from the days of old :)

builtin/providers/openstack/resource_openstack_blockstorage_volume_attach_v2.go
jtopjian: Well if we're going to be honest with each other, I gotta say _IMO_, I'm not a fan either :)

So this is a legitimate `case` statement that is kind of equivalent to:

```go
switch v {
case "":
  nil
case ...:
```

This is because there could be three valid options:

1. Don't specify and let OpenStack apply the default (which is RW, but that _might_ be different depending on the cloud).
2. RW
3. RO

I'll fix this up and make the first case more explicit rather than obscure. I'd rather not enforce a default at the schema level since different OpenStack clouds _could_ be doing something different than the standard default (which is something else I don't agree with, but I respect the right do to it :)

jtopjian:
@stack72 ping

stack72:
LGTM! Thanks for making those changes :)

tadam:
What's the difference between `openstack_blockstorage_volume_attach_v2` and `
openstack_compute_volume_attach_v2`?

Also, there is a typo in example usage:
`instance_id = "${openstack_blockstorage_instance_v2.instance_1.id}"`
Should be `openstack_compute_instance_v2.instance_1.id`, I guess.

jtopjian:
@tadam `openstack_blockstorage_volume_attach_v2` uses the Cinder API to make the attachment while `openstack_compute_volume_attach_v2` uses the Nova API. 

For most people, it makes zero difference which one to use. I recommend sticking with `openstack_compute_volume_attach_v2` or flipping a coin (seriously). :)

Thanks for pointing out the typo! I will get a patch in for this. (Or if you'd like to submit a patch, ping me and I'll quickly review it :smile: )

tadam:
@jtopjian thank you for the clarification. In my case Nova API is working, but not Cinder API. But it's probably due to old version of OpenStack release that I use, or its settings.

Patch it by yourself, please =)

jtopjian:
@tadam Ah, yeah, there's a good chance your cloud is only using Cinder v1 at this time. :)

Will do - thanks again for pointing out the typo!

PR #10276, opened 2016-11-21
command: more manageable shadow error output

mitchellh:
This changes the shadow error message to be written to a file. By writing it to a file, the output is far less intimidating (especially since shadow error output tends to be long). We risk users submitting worse error messages but in general the errors reported have been quite good so far.

Looks like this:

2016-11-21 at 3 05 pm

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -5,10 +5,12 @@ import (
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


mitchellh:
Merging since this doesn't affect behavior in any real way. :)

PR #10281, opened 2016-11-22
providers/aws: cloudfront distribution 404 should mark as gone

mitchellh:
@clint Let me know if there is prior examples of how to test this and I'll gladly do so.

Just ran into this while trying to reproduce another issue.

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudfront_distribution.go
+++ b/builtin/providers/aws/resource_aws_cloudfront_distribution.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


stack72:
Hi @mitchellh 

An example of this can be found here: https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_efs_mount_target_test.go#L66

stack72:
Hi @mitchellh 

Ok, I tried to add a test to this and it's not quite as straight forward as I had hoped - the Delete func for a distribution (via SDK) requires an etag - we don't actually have an etag in the manual delete 

I am going to merge this for now and can rethink this later

P.

PR #10291, opened 2016-11-22
provider/aws: Add new aws_db_snapshot data source

### LINKED ISSUE #10287: provider/aws: Add a new aws_rds_snapshot data source
#10017 recently added the aws_ebs_snapshot data source, fixing #8828.

It would be great if we could have something similar for AWS RDS snapshots, which could be used like so:

```tcl
data "aws_rds_snapshot" "example" {
  most_recent = true
  filter {
    name = "name"
    values = ["foo-prod"]
  }
  owners = ["self"]
}

resource "aws_db_instance" "foo-dev-rds" {
    identifier = "foo-dev"
    snapshot_identifier = "${data.aws_rds_snapshot.example.id}"
    # ...
}
```

This would allow for the workflow of:
* taint existing development/staging RDS instance
* re-run terraform to recreate dev/stage RDS instance from latest snapshot of prod RDS

...without having to hard-code (and continually update) RDS snapshot identifier names like `"rds:foo-prod-2016-11-30-07-05"` in the Terraform config.

stack72:
Fixes #10287

```
% make testacc TEST=./builtin/providers/aws
% TESTARGS='-run=TestAccAWSDbSnapshotDataSource_'     2 ↵ ✭
2016/11/22 18:55:08 Generated command/internal_plugin_list.go
-run=TestAccAWSDbSnapshotDataSource_ -timeout 120m
=== RUN   TestAccAWSDbSnapshotDataSource_basic
--- PASS: TestAccAWSDbSnapshotDataSource_basic (966.68s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws966.699s
```

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSDBSnapshot_
2016/11/22 18:24:05 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSDBSnapshot_ -timeout 120m
=== RUN   TestAccAWSDBSnapshot_basic
--- PASS: TestAccAWSDBSnapshot_basic (892.75s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    892.773s
```

------

**Note:** catsby marked this as `WIP` for now

### DIFF
--- a/builtin/providers/aws/data_source_aws_ami.go
+++ b/builtin/providers/aws/data_source_aws_ami.go
@@ -179,7 +179,7 @@ func dataSourceAwsAmiRead(d *schema.ResourceData, meta interface{}) error {
 	nameRegex, nameRegexOk := d.GetOk("name_regex")
 	owners, ownersOk := d.GetOk("owners")
 
-	if executableUsersOk == false && filtersOk == false && nameRegexOk == false && ownersOk == false {
+	if !executableUsersOk && !filtersOk && !nameRegexOk && !ownersOk {
 		return fmt.Errorf("One of executable_users, filters, name_regex, or owners must be assigned")
 	}
 
--- a/builtin/providers/aws/data_source_aws_db_snapshot.go
+++ b/builtin/providers/aws/data_source_aws_db_snapshot.go
@@ -1,0 +1,182 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/rds"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsDbSnapshot() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsDbSnapshotRead,
+
+		Schema: map[string]*schema.Schema{
+			//selection criteria
+			"filter": dataSourceFiltersSchema(),
+
+			"db_instance_identifier": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"db_snapshot_identifier": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"snapshot_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"include_shared": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"include_public": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			//Computed values returned
+			"allocated_storage": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"availability_zone": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"db_snapshot_arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"encrypted": {
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"engine": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"engine_version": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"iops": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"kms_key_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"license_model": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"option_group_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"port": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"source_db_snapshot_identifier": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"source_region": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"status": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"storage_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"vpc_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsDbSnapshotRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).rdsconn
+
+	instanceIdentifier, instanceIdentifierOk := d.GetOk("db_instance_identifier")
+	filters, filtersOk := d.GetOk("filter")
+	snapshotIdentifier, snapshotIdentifierOk := d.GetOk("db_snapshot_identifier")
+
+	if instanceIdentifierOk == false && filtersOk == false && snapshotIdentifierOk == false {
+		return fmt.Errorf("One of db_snapshot_indentifier, db_instance_identifier or filter must be assigned")
+	}
+
+	params := &rds.DescribeDBSnapshotsInput{}
+	if instanceIdentifierOk {
+		params.DBInstanceIdentifier = aws.String(instanceIdentifier.(string))
+	}
+	if snapshotIdentifierOk {
+		params.DBSnapshotIdentifier = aws.String(snapshotIdentifier.(string))
+	}
+	if filtersOk {
+		params.Filters = buildAwsDataSourceFilters(filters.(*schema.Set))
+	}
+
+	if v, ok := d.GetOk("include_shared"); ok {
+		params.IncludeShared = aws.Bool(v.(bool))
+	}
+
+	if v, ok := d.GetOk("include_public"); ok {
+		params.IncludePublic = aws.Bool(v.(bool))
+	}
+
+	resp, err := conn.DescribeDBSnapshots(params)
+	if err != nil {
+		return err
+	}
+
+	if len(resp.DBSnapshots) < 1 {
+		return fmt.Errorf("Your query returned no results. Please change your search criteria and try again.")
+	}
+
+	if len(resp.DBSnapshots) > 1 {
+		return fmt.Errorf("Your query returned more than one result. Please try a more specific search criteria.")
+	}
+
+	return dbSnapshotDescriptionAttributes(d, resp.DBSnapshots[0])
+}
+
+func dbSnapshotDescriptionAttributes(d *schema.ResourceData, snapshot *rds.DBSnapshot) error {
+	d.SetId(*snapshot.DBInstanceIdentifier)
+	d.Set("db_instance_identifier", snapshot.DBInstanceIdentifier)
+	d.Set("db_snapshot_identifier", snapshot.DBSnapshotIdentifier)
+	d.Set("snapshot_type", snapshot.SnapshotType)
+
+	d.Set("allocated_storage", snapshot.AllocatedStorage)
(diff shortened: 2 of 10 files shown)


builtin/providers/aws/data_source_aws_db_snapshot.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+}
+
+func dataSourceAwsDbSnapshotRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).rdsconn
+
+	instanceIdentifier, instanceIdentifierOk := d.GetOk("db_instance_identifier")
+	snapshotIdentifier, snapshotIdentifierOk := d.GetOk("db_snapshot_identifier")
+
+	if instanceIdentifierOk == false && snapshotIdentifierOk == false {
```
kwilczynski: The only this here, which is more of a idiom, would be to do `!instanceIdentifierOk` over explicit comparison with `false`.

stack72: Totally agree - change made :)

builtin/providers/aws/data_source_aws_ami.go
```diff
@@ -181,7 +181,7 @@ func dataSourceAwsAmiRead(d *schema.ResourceData, meta interface{}) error {
 	nameRegex, nameRegexOk := d.GetOk("name_regex")
 	owners, ownersOk := d.GetOk("owners")
 
-	if executableUsersOk == false && filtersOk == false && nameRegexOk == false && ownersOk == false {
+	if !executableUsersOk && !filtersOk && !nameRegexOk && !ownersOk {
```
catsby: This seems like an unrelated change. Is there a bug here? Or is it a logically equivalent change? My gut says "no" to the later, but I'm happy to be wrong

website/source/docs/providers/aws/d/db_snapshot.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+---
+
+# aws\_db\_snapshot
+
+Use this data source to get information about a DB Snapshot for use when provisioning DB instances
+
+## Example Usage
+
+```
+data "aws_db_snapshot" "ebs_volume" {
```
catsby: nit: `ebs_volume` sounds like a weird data source name when referring to a db snapshot 😄

website/source/docs/providers/aws/d/db_snapshot.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `most_recent` - (Optional) If more than one result is returned, use the most
+recent Snapshot.
+
+* `db_instance_identifier` - (Optional) Returns the list of snapshots created by the specific db_instance
```
catsby: I'm not sure how to use `db_instance_identifier` based on this description

website/source/docs/providers/aws/d/db_snapshot.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+
+* `db_snapshot_identifier` - (Optional) Returns information on a specific snapshot_id.
```
catsby: I'm not sure how to use `db_snapshot_identifier ` based on this description

website/source/docs/providers/aws/d/db_snapshot.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+
+* `snapshot_type` - (Optional) The type of snapshots to be returned. If you don't specify a SnapshotType
```
catsby: `SnapshotType` -> `snapshot_type`

Should also list the allowed values; I assume `manual` and `automated`? Or is it `automatic`?

website/source/docs/providers/aws/d/db_snapshot.html.markdown
```diff
@@ -0,0 +1,65 @@
+---
+layout: "aws"
+page_title: "AWS: aws_db_snapshot"
+sidebar_current: "docs-aws-datasource-db-snapshot"
+description: |-
+  Get information on a DB Snapshot.
+---
+
+# aws\_db\_snapshot
```
catsby: I'm confused about this data source; it sounds singular but some of the parameters seem like it could return multiple?

website/source/docs/providers/aws/r/db_snapshot.html.md
```diff
@@ -0,0 +1,62 @@
...
+* `engine_version` - Specifies the version of the database engine.
+* `iops` - Specifies the Provisioned IOPS (I/O operations per second) value of the DB instance at the time of the snapshot.
+* `kms_key_id` - The ARN for the KMS encryption key.
+* `license_model` - License model information for the restored DB instance.
+* `option_group_name` - Provides the option group name for the DB snapshot.
+* `source_db_snapshot_identifier` - The DB snapshot Arn that the DB snapshot was copied from. It only has value in case of cross customer or cross region copy.
+* `source_region` - The region that the DB snapshot was created in or copied from.
+* `status` - Specifies the status of this DB snapshot.
+* `storage_type` - Specifies the storage type associated with DB snapshot.
+* `vpc_id` - Specifies the storage type associated with DB snapshot.
```
catsby: Can we include `InstanceCreateTime` or `SnapshotCreateTime` as a `Computed` attribute here? Might be useful. The docs suggest they are the same value so  pick one I guess? http://docs.aws.amazon.com/fr_fr/AmazonRDS/latest/APIReference/API_DBSnapshot.html

builtin/providers/aws/data_source_aws_db_snapshot_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestAccAWSDbSnapshotDataSource_basic(t *testing.T) {
+	rInt := acctest.RandInt()
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckAwsDbSnapshotDataSourceConfig(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsDbSnapshotDataSourceID("data.aws_db_snapshot.snapshot"),
```
catsby: Can we add a second step that adds another snapshot, and ensure that our data source then points to that newest snapshot? The test as it stands doesn't test the `most_recent` logic, because there is only 1 snapshot. If possible we should compare the ARNs and/or check that the "most recent" data source ARN changes

builtin/providers/aws/resource_aws_db_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	pollsWhereNotFound := 0
+	for {
+		res, err := conn.DescribeDBSnapshots(req)
+		if err != nil {
+			if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == "DBSnapshotNotFound" {
+				pollsWhereNotFound++
+				// We arbitrarily stop polling after getting a "not found" error five times,
+				// assuming that the Snapshot has been deleted by something other than Terraform.
+				if pollsWhereNotFound > 100 {
```
catsby: Why did we choose to count poll attempts, instead of using something like `resource.Retry()` or `WaitForState()`, which use time as the controlling factor instead of a count?

catsby: Should also say that we'd like to introduce a `Timeout` block to helper/schema and this style would likely make using that very hard

builtin/providers/aws/resource_aws_db_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+
+		if len(res.DBSnapshots) != 1 {
+			return nil, fmt.Errorf("new Snapshot vanished while pending")
+		}
+
+		state := *res.DBSnapshots[0].Status
+
+		if state == "creating" {
+			// Give it a few seconds before we poll again.
+			time.Sleep(4 * time.Second)
```
catsby: So we don't actually time out on snapshot creation, we poll indefinitely until it completes?  That's probably the right thing, considering this is _data_, but I wonder if some users would like to toggle this behavior, e.g. "create a snapshot, but I trust that AWS will complete it and I don't want/need Terraform to sit and wait for it to finish". Not a blocker, but a thought

website/source/docs/providers/aws/d/db_snapshot.html.markdown
catsby: After reading the code (I read the docs first) this is pretty clearly a singular thing, but the docs for `db_instance_identifier` and `db_instance_identifier` confuse me

website/source/docs/providers/aws/r/db_snapshot.html.md
catsby: also for the data source I think this data would be useful (especially for the data source, actually)

builtin/providers/aws/data_source_aws_ami.go
stack72: there is already a bool in play so i thought i would try and use that rather than test equality

website/source/docs/providers/aws/d/db_snapshot.html.markdown
stack72: @catsby this follows the same pattern as the other data sources - it can indeed return multiple results but we try to enforce people to return just 1.... for now

website/source/docs/providers/aws/d/db_snapshot.html.markdown
stack72: Updated docs :)

website/source/docs/providers/aws/r/db_snapshot.html.md
stack72: done!

Seraf:
Hello,

What is the status of this PR please ? I really need this feature to destroy infrastructure and rebuild them on demand.

Thanks !

joshuaspence:
+1

builtin/providers/aws/resource_aws_db_snapshot.go
catsby: `Timeout` has been merged and is available in `helper/schema`, should we keep this polling on a count, or use a timeout? Also the docs say `"not found" error five times` but it looks like we're counting to 100 with `pollsWhereNotFound`?

builtin/providers/aws/resource_aws_db_snapshot.go
catsby: We should use the `Timeout` here and let people specify it, but give it a high default like `40` or `60`

stack72:
@catsby just pushed the code with the new Timeouts work added and all looks good :) It also gets rid of all that nasty polling you asked about!

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSDBSnapshot_' 
2017/05/09 18:16:45 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSDBSnapshot_ -timeout 120m
=== RUN   TestAccAWSDBSnapshot_basic
--- PASS: TestAccAWSDBSnapshot_basic (879.32s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	879.339s
```

byrneo:
Hi - was this change supposed to also support 'name_regex' for data "aws_db_snapshot" (as is available in e.g. data "aws_ami")

PR #10299, opened 2016-11-22
Update elb_service_account docs

miketheman:
The current example using the ELB's account ID will trigger an update for a resource that uses the `.id` instead if the `.arn` syntax.

Once updated to the `.arn`, no changes are detected.

[ci skip]


miketheman:
Historical context: https://github.com/hashicorp/terraform/pull/8700

stack72:
Thanks :) This LGTM!

PR #10300, opened 2016-11-22
Remove extra dot from state backup file

jbardin:
The state subcommand was adding an extra dot to the backup file


mitchellh:
LGTM

PR #10301, opened 2016-11-22
provider/aws: data source for AWS Route Table

mathieuherbert:
Hi,

I added an AWS provider, aws_route_table.
We can retrieve, for example a route table from a subnet id and add a route with this route table
```
variable "subnet_id" {}

data "aws_route_table" "selected" {
  subnet_id = "${var.subnet_id}"
}

resource "aws_route" "route" {
  route_table_id = "${data.aws_route_table.selected.id}"
  destination_cidr_block = "10.0.1.0/22"
  vpc_peering_connection_id = "pcx-45ff3dc1"
}
```

Tests:

```
/opt/gopath/src/github.com/hashicorp/terraform$ TF_ACC=true go test -trace /tmp/toto.txt -v github.com/hashicorp/terraform/builtin/providers/aws -run ^TestAccDataSourceAwsRouteTable$
=== RUN   TestAccDataSourceAwsRouteTable
--- PASS: TestAccDataSourceAwsRouteTable (46.92s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	46.966s
```

### DIFF
--- a/builtin/providers/aws/data_source_aws_route_table.go
+++ b/builtin/providers/aws/data_source_aws_route_table.go
@@ -1,0 +1,80 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsRouteTable() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsRouteTableRead,
+
+		Schema: map[string]*schema.Schema{
+			"subnet_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"vpc_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"filter": ec2CustomFiltersSchema(),
+
+			"id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"tags": tagsSchemaComputed(),
+		},
+	}
+}
+
+func dataSourceAwsRouteTableRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+	req := &ec2.DescribeRouteTablesInput{}
+
+	req.Filters = buildEC2AttributeFilterList(
+		map[string]string{
+			"route-table-id":        d.Get("id").(string),
+			"vpc-id":                d.Get("vpc_id").(string),
+			"association.subnet-id": d.Get("subnet_id").(string),
+		},
+	)
+	req.Filters = append(req.Filters, buildEC2TagFilterList(
+		tagsFromMap(d.Get("tags").(map[string]interface{})),
+	)...)
+	req.Filters = append(req.Filters, buildEC2CustomFilterList(
+		d.Get("filter").(*schema.Set),
+	)...)
+	if len(req.Filters) == 0 {
+		// Don't send an empty filters list; the EC2 API won't accept it.
+		req.Filters = nil
+	}
+
+	log.Printf("[DEBUG] Describe Route Tables %v\n", req)
+	resp, err := conn.DescribeRouteTables(req)
+	if err != nil {
+		return err
+	}
+	if resp == nil || len(resp.RouteTables) == 0 {
+		return fmt.Errorf("no matching Route Table found")
+	}
+	if len(resp.RouteTables) > 1 {
+		return fmt.Errorf("multiple Route Table matched; use additional constraints to reduce matches to a single Route Table")
+	}
+
+	rt := resp.RouteTables[0]
+
+	d.SetId(*rt.RouteTableId)
+	d.Set("id", *rt.RouteTableId)
+	d.Set("vpc_id", rt.VpcId)
+	d.Set("tags", tagsToMap(rt.Tags))
+
+	return nil
+}
--- a/builtin/providers/aws/data_source_aws_route_table.go
+++ b/builtin/providers/aws/data_source_aws_route_table.go
@@ -13,19 +13,19 @@ func dataSourceAwsRouteTable() *schema.Resource {
 		Read: dataSourceAwsRouteTableRead,
 
 		Schema: map[string]*schema.Schema{
-			"subnet_id": &schema.Schema{
+			"subnet_id": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
 			},
-			"vpc_id": &schema.Schema{
+			"vpc_id": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
 			},
 			"filter": ec2CustomFiltersSchema(),
 
-			"id": &schema.Schema{
+			"id": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
@@ -72,7 +72,6 @@ func dataSourceAwsRouteTableRead(d *schema.ResourceData, meta interface{}) error
 	rt := resp.RouteTables[0]
 
 	d.SetId(*rt.RouteTableId)
-	d.Set("id", *rt.RouteTableId)
 	d.Set("vpc_id", rt.VpcId)
 	d.Set("tags", tagsToMap(rt.Tags))
 
--- a/builtin/providers/aws/data_source_aws_route_table.go
+++ b/builtin/providers/aws/data_source_aws_route_table.go
@@ -24,13 +24,71 @@ func dataSourceAwsRouteTable() *schema.Resource {
 				Computed: true,
 			},
 			"filter": ec2CustomFiltersSchema(),
-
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsRouteTable() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsRouteTableRead,
+
+		Schema: map[string]*schema.Schema{
+			"subnet_id": &schema.Schema{
```
kwilczynski: As of Go 1.7+ the explicit type is no longer needed here, so `&schema.Schema` can be removed.

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,80 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
```
kwilczynski: I am not sure if this should be optional? It seem that rest of the attributes are also optional, is this desirable?

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,80 @@
...
+			"association.subnet-id": d.Get("subnet_id").(string),
+		},
+	)
+	req.Filters = append(req.Filters, buildEC2TagFilterList(
+		tagsFromMap(d.Get("tags").(map[string]interface{})),
+	)...)
+	req.Filters = append(req.Filters, buildEC2CustomFilterList(
+		d.Get("filter").(*schema.Set),
+	)...)
+	if len(req.Filters) == 0 {
```
kwilczynski: Is this ever going to be empty? Given the first `req.Filters` getting some minimal filters.

builtin/providers/aws/data_source_aws_route_table.go
mathieuherbert: I think we can get a Route Table with

- `subnet_id`
- `tags`
- `vpc_id` and `tags`
- filters
What do you think ?

mathieuherbert:
Hi @kwilczynski,
Thank you for your review :) 
I think we can get a Route Table with
- subnet_id
- tags
- vpc_id and tags
- filters
The subnet is just one of the possibilities. What do you think ?

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,79 @@
...
+				Computed: true,
+			},
+			"vpc_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"filter": ec2CustomFiltersSchema(),
+
+			"id": {
```
kwilczynski: This is not needed (as per the earlier comment, you do not to set `id` explicitly - this is done via the `d.SetId()`.

kwilczynski:
@mathieuherbert hi there! Thank you for your updates!

I see, you are allowing to set either and then narrow the result down with either tags or filters. Which is fine, but in such a case I would add a check whether at least one is set, and issue an appropriate message should that not be the case.

kwilczynski:
@mathieuherbert hi there again!

I was wondering, how would you feel about adding `associations` and `routes` too? It would be an additional computed attribute (a sub-schema, etc.). Would it be something you could consider? It would make this data source more feature complete.

mathieuherbert:
Hi @kwilczynski,

So now, I export every `routes`and `associations` from the route table. I hope it is the good way. I take some code from aws_route_table resource.

builtin/providers/aws/data_source_aws_route_table.go
mathieuherbert: I add an error if there is no filter

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,190 @@
...
+
+	rt := resp.RouteTables[0]
+
+	d.SetId(*rt.RouteTableId)
+	d.Set("vpc_id", rt.VpcId)
+	d.Set("tags", tagsToMap(rt.Tags))
+
+	routes := make([]map[string]interface{}, 0, len(rt.Routes))
+	// Loop through the routes and add them to the set
+	for _, r := range rt.Routes {
```
kwilczynski: Same as above, if we turn this into a function it would be great.

builtin/providers/aws/data_source_aws_route_table.go
kwilczynski: I see, so you want to use the sheer fact that `buildEC2AttributeFilterList` would return empty slice given that values in the map are empty. Which is fine, albeit it can be make a whole lot simpler should you break even before getting to this helper. You can do this as follows:

```go
vpcId, vpcIdOk := d.GetOk("vpc_id")
subnetId, subnetIdOk := d.GetOk("subnet_id")

if !vpcIdOk && !subnetIdOk {
	return fmt.Errorf("One of vpc_id or subnet_id must be assigned")
}

req.Filters = buildEC2AttributeFilterList(
 	map[string]string{
		"vpc-id":                vpcId.(string),
		"association.subnet-id": subnetId.(string),
 	},
 )
```

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,190 @@
...
+		if r.VpcPeeringConnectionId != nil {
+			m["vpc_peering_connection_id"] = *r.VpcPeeringConnectionId
+		}
+		if r.NetworkInterfaceId != nil {
+			m["network_interface_id"] = *r.NetworkInterfaceId
+		}
+
+		routes = append(routes, m)
+	}
+	d.Set("routes", routes)
```
kwilczynski: When storing complex values, if would recommend adding an error check, for example:

```go
if err := d.Set("routes", routes); err != nil {
    return err
}
```

mathieuherbert:
@kwilczynski , thank you very much for your help, it is very interesting !

kwilczynski:
@mathieuherbert thank you for working on it!

mathieuherbert:
@kwilczynski  , do you have some comments on my last commit please ? Thanks !

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	if len(resp.RouteTables) > 1 {
+		return fmt.Errorf("Multiple Route Table matched; use additional constraints to reduce matches to a single Route Table")
+	}
+
+	rt := resp.RouteTables[0]
+
+	d.SetId(*rt.RouteTableId)
+	d.Set("vpc_id", rt.VpcId)
+	d.Set("tags", tagsToMap(rt.Tags))
```
kwilczynski:
```go
if err := d.Set("tags", tagsToMap(rt.Tags)); err != nil {
    return err
}
```

builtin/providers/aws/data_source_aws_route_table.go
```diff
@@ -0,0 +1,205 @@
...
+		routes = append(routes, m)
+	}
+	return routes
+}
+
+func dataSourceAssociationsRead(ec2Assocations []*ec2.RouteTableAssociation) []map[string]interface{} {
+	associations := make([]map[string]interface{}, 0, len(ec2Assocations))
+	// Loop through the routes and add them to the set
+	for _, a := range ec2Assocations {
+
```
kwilczynski: No need for the extra line.

kwilczynski:
@mathieuherbert hi there! Looks good! Two minor comments, and then I think ready to go.

stack72:
```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccDataSourceAwsRouteTable'          ✹
2016/11/28 18:02:00 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccDataSourceAwsRouteTable -timeout 120m
=== RUN   TestAccDataSourceAwsRouteTable
--- PASS: TestAccDataSourceAwsRouteTable (21.16s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	21.199s
```

Massive thanks to @kwilczynski for all the work here!

stack72:
Merged manually :)

website/source/docs/providers/aws/d/route_table.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+## Argument Reference
+
+The arguments of this data source act as filters for querying the available
+Route Table in the current region. The given filters must match exactly one
+Route Table whose data will be exported as attributes.
+
+
+* `filter` - (Optional) Custom filter block as described below.
+
+* `id` - (Optional) The id of the specific Route Table to retrieve.
```
swestcott: Was this parameter implemented? I can't see any reference to it in `data_source_aws_route_table.go` and get this when I try to use it,
```
Errors:
  * data.aws_route_table.selected: : invalid or unknown key: id
```

PR #10304, opened 2016-11-22
Update opsworks_application.html.markdown

mcheshier-switch:
'Secret' is incorrect in the environment argument documentation. It is used correctly in the example at the top as 'secure'.


stack72:
Thanks so much - this LGTM!

PR #10307, opened 2016-11-23
Clarified variable order precedence doc line

niravradia:


stack72:
Thanks for this - this LGTM!

PR #10310, opened 2016-11-23
Implements import with specified provider

### LINKED ISSUE #10029: terraform import does not handle multiple AWS profiles
Terraform does not seem to respect AWS profile choice when importing AWS resources. To handle this I have found I need to replace my default profile with the configuration from my chosen AWS profile, but I still needed to keep my profile in place. This works as a workaround, but is inconvenient.

### Terraform Version
Terraform v0.7.10

### Affected Resource(s)
- dynamodb_table
- aws_sqs_queue
(probably affects all providers)

### Expected Behavior
Terraform should have imported my resources.

### Actual Behavior
Terraform fails with an Access Denied error.

### Steps to Reproduce
1. Have an aws configuration with multiple profiles, and a terraform state which doesn't use the default profile.
2. `terraform import dynamodb_table.my_cool_table my.cool.table`

spangenberg:
This allows the import with a specified provider.
So far it was not possible to get resources imported from a different provider
than the default.

Example invocation that allows to import the specified AWS SQS queue with an AWS provider called "second":
`terraform import -provider=aws.second aws_sqs_queue.fancy https://sqs.us-east-1.amazonaws.com/123456789101/fancy`

This will fix #10029.

### DIFF
--- a/command/import.go
+++ b/command/import.go
@@ -32,6 +32,7 @@ func (c *ImportCommand) Run(args []string) int {
 	cmdFlags.StringVar(&c.Meta.stateOutPath, "state-out", "", "path")
 	cmdFlags.StringVar(&c.Meta.backupPath, "backup", "", "path")
 	cmdFlags.StringVar(&configPath, "config", pwd, "path")
+	cmdFlags.StringVar(&c.Meta.provider, "provider", "", "provider")
 	cmdFlags.Usage = func() { c.Ui.Error(c.Help()) }
 	if err := cmdFlags.Parse(args); err != nil {
 		return 1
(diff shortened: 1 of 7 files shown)


mitchellh:
Really well done. I'm going to touch up some things a bit but this is agreat.

jleclanche:
@mitchellh that command line argument should probably be mentioned in https://www.terraform.io/docs/import/usage.html and/or the error message - it's very confusing to get an error on an instance id not existing simply because it's outside the default provider.

quaeritate:
I ran into this issue and found the solution by searching and ending up on this page. The official documentation has not been updated. 

In the example given there is a mistake. The provider needs to be in ""

``
terraform import -provider="aws.second" aws_sqs_queue.fancy https://sqs.us-east-1.amazonaws.com/123456789101/fancy

ironhalo:
+1 on adding this to the documentation, I was only able to successfully import state with multiple providers only after finding this issue.

PR #10325, opened 2016-11-23
Fix some cases for nested maps and lists

jbardin:
When referencing a list of maps variable from within a resource, only
the first list element is included the plan. This is because GetRaw
can't access the interpolated values. Add some tests to document this
behavior for both Get and GetRaw.

In helper/schema, this now checks for interpolated values when reading a map.
Accessing an interpolated value in a map through ConfigFieldReader can
fail, because GetRaw can't access interpolated values, so we check if the
value exists at all by looking in the config. If the config has a value,
assume our map's value is interpolated and proceed as such.

We also need to lookup the correct schema to properly read a field from
a nested structure.

- Maps previously always defaulted to TypeString. Now
check if Elem is a ValueType and use that if applicable
- Lists now return the schema for nested element types, defaulting to a
  TypeString like maps.

This only allows maps and lists to be nested one level deep in a schema, and the
inner map or list must only contain string values.

### DIFF
--- a/helper/schema/field_reader.go
+++ b/helper/schema/field_reader.go
@@ -75,6 +75,8 @@ func addrToSchema(addr []string, schemaMap map[string]*Schema) []*Schema {
 				return nil
 			}
 		case TypeList, TypeSet:
+			isIndex := len(addr) > 0 && addr[0] == "#"
+
 			switch v := current.Elem.(type) {
 			case *Resource:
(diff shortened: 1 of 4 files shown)


jbardin:
This should mostly take care of #10187. I'm not sure if we can do any better until we get rid of the dotted string types, and possibly some helper/schema changes.

PR #10344, opened 2016-11-24
Make template_file change more explicit

jamtur01:
It wasn't clear to me what this change meant.


stack72:
Nice one! Good way to make it more clear :)

PR #10352, opened 2016-11-24
README: remove wrong reference to Linux

shaftoe:


stack72:
Thanks so much for this - we love documentation fixes @shaftoe :)

shaftoe:
@stack72 welcome! I'm looking forward to join the development, but I'm new to Go so it might take some time. In the meanwhile I'll take care of docs while going through them

PR #10353, opened 2016-11-24
feat(consul): Adds a data source to retrieve kv pairs under a specified path

samber:
Hi guys,

Example: 

```
provider "consul" {
    address = "1.2.3.4:8500"
    datacenter = "eu_central_1"
}

data "consul_key_prefix" "test" {
     path_prefix = "services/api/"
}

output "variables" { value = "${data.consul_key_prefix.test.var}" }
```

Read have been deprecated in resource consul_keys after adding the consul_keys data source. I think we should do the same for the resource consul_key_prefix. This data source should be the correct way to get data.

### DIFF
--- a/builtin/providers/consul/data_source_consul_key_prefix.go
+++ b/builtin/providers/consul/data_source_consul_key_prefix.go
@@ -1,0 +1,66 @@
+package consul
+
+import (
+	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceConsulKeyPrefix() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceConsulKeyPrefixRead,
+
+		Schema: map[string]*schema.Schema{
+			"datacenter": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"token": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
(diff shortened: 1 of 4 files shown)


stack72:
Hi @samber 

Thanks for the work here - looks like this doesn't build - any chance you could have a look at it?

Thanks

Paul

samber:
@stack72 sorry for the late fix.

This is working well ;)

Should we deprecate reads from consul_key_prefix resource ?

partsalliance:
@stack72 any ideas when this PR can be merged?

samber:
up ?

nanoz:
Having this PR checked and merged would be very helpful to us @stack72 :)

builtin/providers/consul/data_source_consul_key_prefix.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if err := d.Set("var", vars); err != nil {
+		return err
+	}
+
+	// Store the datacenter on this resource, which can be helpful for reference
+	// in case it was read from the provider
+	d.Set("datacenter", dc)
+
+	d.SetId("-")
```
stack72: If there is no Id, then we should use something like:

```
d.SetId(time.Now().UTC().String())
```

apparentlymart:
Note that `consul_key_prefix` has never supported read-only usage, so there's no need to deprecate anything there as we did with `consul_keys`.

teamterraform:
Hi folks! Now that terraform 0.12 is released, we are undertaking a general GH issue/PR cleanup. The providers now live in their own repositories, so any new issues or PRs should be opened there. Thanks!

PR #10370, opened 2016-11-25
Fix typo in test method name

paultyng:
Minor test method name typo update, shouldn't affect any runtime functionality.


stack72:
Thanks for adding this :) Nice Catch!

PR #10383, opened 2016-11-27
Fixed private_key param

pauljrob:
The provisioned argument key_file has been deprecated.  Updated the example to use the new argument private_key.


stack72:
LGTM! thanks for the work here @pauljrob :)

PR #10387, opened 2016-11-27
provider/google: allow session affinity for compute_backend_service

bpineau:
Google's Backend Services gives users control over the session affinity modes.
Let's allow Terraform users to leverage this option.
We don't change the default value ("NONE"), as provided by Google.

This is commited against master, but merges and works fine in main-0.7 too;).


bpineau:
Hi @paddyforan , thanks for the feedback and the test idea !
I amended the commit with a new test step that changes the affinity value (and test the changed result).

paddyforan:
This looks great, thanks so much for adding that, @bpineau. Great addition to our GCP support!

PR #10389, opened 2016-11-28
provider/azurerm: support import of routes, fix route_table

pmcatominey:
Added import functionality to:

* `azurerm_route`
* `azurerm_route_table`

Fix `azurerm_route_table` failing to actually set any routes declared within.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_route_table.go
+++ b/builtin/providers/azurerm/resource_arm_route_table.go
@@ -18,6 +18,9 @@ func resourceArmRouteTable() *schema.Resource {
 		Read:   resourceArmRouteTableRead,
 		Update: resourceArmRouteTableCreate,
 		Delete: resourceArmRouteTableDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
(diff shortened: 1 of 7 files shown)


builtin/providers/azurerm/resource_arm_route_table.go
```diff
@@ -73,7 +76,6 @@ func resourceArmRouteTable() *schema.Resource {
 
 			"subnets": {
 				Type:     schema.TypeSet,
-				Optional: true,
```
stack72: Is this a breaking change?

stack72:
Just got 1 question - but apart from that, LGTM!

builtin/providers/azurerm/resource_arm_route_table.go
pmcatominey: It shouldn't be, that value can't actually be set, the code was never writing the value to the API either.

stack72: Rgr! Thanks :)

stack72:
Thanks for the response - this LGTM! Thanks :)

PR #10391, opened 2016-11-28
The impossible ns1

pashap:
Combines https://github.com/hashicorp/terraform/pull/10058.

 - Renames nsone -> ns1
 - Contains awesome ns1 metadata restructuring from @cakoose
 - Small field name/test changes 
 - Adds ns1 rest client to govender json deps

The only failing test now is `TestAccRecord`, this is due to ns1 metadata still. But i believe that @cakoose has got us one step closer to finishing this.


pashap:
@josephholsten With the forks, dunno how to get this into one clean PR.

builtin/providers/nsone/resource_record.go
```diff
@@ -824,12 +339,20 @@ func RecordCreate(d *schema.ResourceData, meta interface{}) error {
 
 // RecordRead reads the DNS record from ns1
 func RecordRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*nsone.Client)
+	//client := meta.(*nsone.Client)
```
cakoose: @pashap: Whoops, looks like I left some debug code in here.

(Just in case you hadn't already noticed this yourself :-P)

pashap: Thank you @cakoose, i missed that one at first.

Just posted a [PR](https://github.com/hashicorp/terraform/pull/11300) against the terraform ns1-provider branch. Probably a bit more work to do, but its almost there :)

PR #10394, opened 2016-11-28
Fix for aws_opsworks_permission issue 9104

### LINKED ISSUE #9104: [AWS] aws_opsworks_permission: Setting 'level' has no effect, allow_ssh keeps reapplying
Hi there,

thanks for introducting the `aws_opsworks_permission` resource, it's been a great help already.

We encountered two issues with it that I'd like to report. Please have a look at the following resource configuration:

```
resource aws_opsworks_permission deploy_user {
  allow_ssh = true
  allow_sudo = false
  level = "deploy"
  user_arn = "${var.user_arn}"
  stack_id = "${var.stack_id}"
}
```

The issues we've encountered:
1. Setting the level to anything has no effect in OpsWorks, i.e. after `terraform apply`, Terraform saves the new level in it's state file, but the **OpsWorks permission level stays `iam_only`**, always.
2. When setting `allow_ssh = true` and `allow_sudo = false`, **Terraform keeps re-planning and -applying** `allow_ssh = true`. Note: This does not happen if `allow_sudo` is set to true also.

We're using `Terraform v0.7.4`.

Thanks in advance and looking forward to hearing from you!

janschumann:
This will fix #9104 

- [x] allow_ssh failure
- [x] permission level

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_permission.go
+++ b/builtin/providers/aws/resource_aws_opsworks_permission.go
@@ -14,8 +14,8 @@ import (
 
 func resourceAwsOpsworksPermission() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceAwsOpsworksPermissionCreate,
-		Update: resourceAwsOpsworksPermissionCreate,
+		Create: resourceAwsOpsworksSetPermission,
+		Update: resourceAwsOpsworksSetPermission,
 		Delete: resourceAwsOpsworksPermissionDelete,
 		Read:   resourceAwsOpsworksPermissionRead,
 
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_opsworks_permission.go
```diff
@@ -105,7 +105,7 @@ func resourceAwsOpsworksPermissionRead(d *schema.ResourceData, meta interface{})
 			found = true
 			d.SetId(id)
 			d.Set("id", id)
-			d.Set("allow_ssh", permission.AllowSudo)
+			d.Set("allow_ssh", permission.AllowSsh)
 			d.Set("allow_sodo", permission.AllowSudo)
```
Ninir: @janschumann I think the `allow_sodo` should be changed to allow_sudo to fix the typo

stack72:
Thanks for the work here @janschumann :) This LGTM!

Paul

PR #10402, opened 2016-11-28
Correct parent zone reference in route53_zone docs

ketzacoatl:
When creating a subdomain as exemplified in the docs, the NS records need to be added to the parent zone, and the docs had this reference pointed at the subdomain (rather than the parent).


stack72:
Thanks for fixing the example here :) This LGTM!

PR #10418, opened 2016-11-29
Fix count interpolation with computed values

### LINKED ISSUE #3888: When using count parameter w/ a split in aws_security_group_rule, receive parsing error "strconv.ParseInt" .. " invalid syntax"
This is using terraform v0.6.5

The following module is used:

```
resource "aws_security_group_rule" "rule" {
    count = "${length(split(",",var.ids))}"
    type = "${var.type}"
    from_port = "${var.port}"
    to_port = "${var.port}"
    protocol = "tcp"
    security_group_id = "${var.sg_id}"
    source_security_group_id = "${element(split(",",var.ids), count.index)}"
}
```

and when I run in conjunction with the wider project:

```
Errors:

  * strconv.ParseInt: parsing "${length(split(\",\",var.ids))}": invalid syntax
```

If I rewrite

```
count = "${length(split(",",var.ids))}"
```

as

```
count = "${length(split(",","a,b,c"))}"*
```

it works...

greenboxal:
This commit enables the use computed values when interpolating
the count special variable.

This is achieved by defering the expansion of `count` during
walkInput and walkValidate as the real value isn't necessary
at this point.

The value will be resolved during walkRefresh and DynamicExpand
will work as expected.

Partially fixes #3888.

@edit
I'm not sure on how write tests for this. I tested with this snippet and it works: https://gist.github.com/greenboxal/113469077550fa1f6c65d4db7f22ba27

Can someone guide me on this?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -88,6 +88,8 @@ type Resource struct {
 	Provider     string
 	DependsOn    []string
 	Lifecycle    ResourceLifecycle
+
+	DeferCountComputation bool
 }
 
 // Copy returns a copy of this Resource. Helpful for avoiding shared
@@ -213,6 +215,10 @@ func (r *Module) Id() string {
 
 // Count returns the count of this resource.
 func (r *Resource) Count() (int, error) {
+	if r.DeferCountComputation {
+		return 1, nil
+	}
+
 	v, err := strconv.ParseInt(r.RawCount.Value().(string), 0, 0)
 	if err != nil {
 		return 0, err
--- a/terraform/eval_context.go
+++ b/terraform/eval_context.go
@@ -77,4 +77,11 @@ type EvalContext interface {
 	// State returns the global state as well as the lock that should
 	// be used to modify that state.
 	State() (*State, *sync.RWMutex)
+
+	// True if we can ignore missing computed values when interpolating
+	// the count variable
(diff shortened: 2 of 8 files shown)


mitchellh:
This is a good set of information, but I think for input and validate we'll have a good solution when we switch to a new graph that simply ignores the count variable full out. :) `refresh`, `input`, and `validate` are still using the old graphs!

If this works for you though, feel free to patch and use it.

Thanks for the input! :)

greenboxal:
@mitchellh Nice. Do you have any prevision on when this will be available?

For now, I'll keep using this solution.

mitchellh:
Was hoping we could get it into 0.8 but the freeze came up on us. :)

So, 0.9, probably released as an experimental feature behind a flag in 0.8.x sometime.

cemo:
@mitchellh will this issue be fixed with 0.9? You have recently implemented a dozen features and this is a really huge problem for terraformers.

mitchellh:
I'm hoping it improves with 0.9, but I'm not 100% sure. 0.9 betas will start in Feb and we need to get it in before that window. I'm trying though!

cemo:
We are grateful for the all thing you have done. Thank you again.

PR #10421, opened 2016-11-29
Explain list-type variables, and their use

tamsky:
also include a Note regarding restrictions on the types environment variables can set.
+line length cleanup


mitchellh:
LGTM. Thanks.

PR #3508, opened 2015-10-15
[WIP] Add cli command to export provider/provisioner/resource schemas

VladRassokhin:
This feature allows to run command like `terraform schemas aws` and get json with information about required/optional properties of providers/provisioners/resources and their types. (and much more info)

Originally it was required to add 'properties completion' and 'required properties inspection' features in [IntelliJ HCL/Terraform plugin](https://github.com/VladRassokhin/intellij-hcl) and I think it could be shared with community, so vim/emacs/etc users could use it.

Example of generated output for provisioner `local-exec` (`terraform schemas -indent -json local-exec`) could be found [here](https://github.com/VladRassokhin/intellij-hcl/blob/df48a537bbee63c7d8b219c982962aef52e805fe/res/terraform/model/local-exec.json). There are other files for other providers/provisioners in the same directory, look around.

Still some things should be done:
- [ ] Documentation
- [ ] Tests
- [ ] Output formats
  - [X] JSON
  - [ ] XML
  - [ ] plain text (could be useful for bash completion or something like that)
- [ ] Check backward compatibility of third party plugins (new command added for provisioners and providers, at least recompilation required)
  - [ ] Source compatibility
  - [ ] Binary compatibility

PR #10422, opened 2016-11-29
Add `helpers.shema.Provisoner` base type for provisioners like there's one for providers

VladRassokhin:
Migrate provisioners to providers-like hierarchy when all of then extends `schema.Provisioner`.
Built-in plugins were migrated.
Also added support for schema validation for built-in provisioners: file, local-exec, remote-exec.

AFAIU external plugins would still compile but functions `Apply` and `Validate` from `schema.Provisioner` would not be used, so change is quite safe. Tried that by [reverting cahnges in 'chef' provisioner in separate commit](https://github.com/VladRassokhin/terraform/commit/309e0710b7d8894be3ef73845e0d12cd82f6dbb8)

### DIFF
--- a/builtin/bins/provisioner-chef/main.go
+++ b/builtin/bins/provisioner-chef/main.go
@@ -3,13 +3,10 @@ package main
 import (
 	"github.com/hashicorp/terraform/builtin/provisioners/chef"
 	"github.com/hashicorp/terraform/plugin"
-	"github.com/hashicorp/terraform/terraform"
 )
 
 func main() {
 	plugin.Serve(&plugin.ServeOpts{
-		ProvisionerFunc: func() terraform.ResourceProvisioner {
-			return new(chef.ResourceProvisioner)
-		},
+		ProvisionerFunc: chef.ResourceProvisioner,
 	})
 }
--- a/builtin/bins/provisioner-file/main.go
+++ b/builtin/bins/provisioner-file/main.go
@@ -3,13 +3,10 @@ package main
 import (
 	"github.com/hashicorp/terraform/builtin/provisioners/file"
 	"github.com/hashicorp/terraform/plugin"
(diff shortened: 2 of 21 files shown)


mitchellh:
Thanks this is a great start, but I'm not ready to merge as-is. I believe we should have the provisioners use the `ResourceData` structure similar to Providers, for easy config access. 

I might be willing to make these changes since we need to support the Stop API in core for provisioners (just doesn't yet) to make them interruptable... and I wanted to make that easy by having a helper/schema-type thing.

VladRassokhin:
Hi @mitchellh 
Thanks for fast feedback.
I've updated PR with your suggestion.

As mentioned in commit message:
Known problems:
* had to introduce `ResourceData.GetRawConfig()` since `chef` provisioner parses configuration manually;
* In `remote-exec` provisioner `inline` can no longer accept 'string' value, only list of strings allowed. - due to migration to schema. There's a workaround like it;s done in `chef` provisioner, but I'd like to keep code as is.

VladRassokhin:
If you think it's worth to spend time, I'd try to convert chef provisioner to use `Schema` and get rid of `ResourceData.GetRawConfig()`

VladRassokhin:
Hi @mitchellh, could you please take a look and provide some feedback? Maybe that PR could be merged before 0.8?

mitchellh:
Hey @VladRassokhin, thanks for this! I'm sorry to pre-empt your work but in introducing a Stop API for provisioners I ended up doing this in #10934. Thanks for your work and I'm sorry I didn't get this merged prior. :(

VladRassokhin:
Ok, fine. At least provisioners now has `schema` so I could rebase #3508 to use it and it would simplify generating metadata for [IntelliJ plugin](https://github.com/VladRassokhin/intellij-hcl).

Issue #9304, opened 2016-10-10
Add "Use previous value" functionality to AWS_CLOUDFORMATION_STACK resource parameters


gugahoi:
This is more of a feature request as I could not get a reply on the IRC channel wether this functionality is supported.
### Terraform Version

`Terraform v0.7.5`
### Affected Resource(s)
- AWS_CLOUDFORMATION_STACK
### Terraform Configuration Files

```
resource "aws_cloudformation_stack" "agents" {
  name = "${var.Team}-agents"
  template_url = "https://s3.amazonaws.com/aws-stack/aws-stack.json"
  capabilities = ["CAPABILITY_IAM"]
  parameters = {
    AgentToken = "${var.AgentToken}"
    Queue  = "${var.Team}"
  }

  lifecycle {
    ignore_changes = ["parameters.AgentToken"]
  }
}
```
### Expected Behavior

Need a way to tell the resource to "Use previous value" for a parameter just like Cloudformation supports. In this example the `AgentToken` should not be updated every time `terraform apply` is run, similarly I should be able to update my stack without overwriting the existing token. 

I could use `lifecycle { ignore_changes = ["parameters.AgentToken"] }` however this will only ignore the changes to this parameter but won't keep the original value.
### Actual Behavior

At the moment there is no documentation on how to tell terraform or if terraform supports configuring a cloudformation resource with a parameter to "Use previous value". The aws-sdk-go [supports this functionality](https://github.com/aws/aws-sdk-go/blob/6061082a165c9c3e8fb9c681a287af83c2464fb3/service/cloudformation/examples_test.go#L95) so it would be great if Terraform could support it too. The documentation does not mention anything at this point in time, if it does support then an example on how to use it would be great.

Ninir:
@gugahoi Hi!

It is possible to support it. However, it will introduce a breaking change, this it will the way to declare parameters. For instance, the first thing that comes to my mind from a dev POV is this declaration:

```hcl
parameter {
   Name = "AgentToken"
   Value  = "${var.AgentToken}"
   UsePreviousValue = true # Or false
}

parameter {
   Name = "Queue"
   Value  = "${var.Team}"
   UsePreviousValue = false
}
```

Can't really tell of another way to set it, since it is defined at the parameter level.

PR #10424, opened 2016-11-29
[WIP] Added parameter field & make parameters deprecated

Ninir:
Fixes https://github.com/hashicorp/terraform/issues/9304

This adds a new cloudformation_stack attribute called `parameter`. It deprecates the previous one called `parameters`.

- [x] Add the new parameter
- [ ] Fix the default tests
- [ ] Update the documentation

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudformation_stack.go
+++ b/builtin/providers/aws/resource_aws_cloudformation_stack.go
@@ -6,10 +6,12 @@ import (
 	"regexp"
 	"time"
 
+	"bytes"
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


stack72:
Hi @Ninir 

What is the reasoning for this? This feels like we are just adding a catch all

P.

Ninir:
Hi Paul,

The reasonning for this is with the related issue (exposed in the description).

Speaking of it, I made the assumption that the var.AgentToken was always being updated (which may be the case). While reading your comment, I started thinking the same thing in fact.
As TF covers the case by not updating the stack if the parameter didn't change, let's simply close it if you agree, no worries on that by my side :).

(learned a lot on this, but was having issues with the code at the moment).

Gauthier

PR #10425, opened 2016-11-30
providers/google: Create and manage Google Cloud Platform Projects

evandbrown:
This proposal allows a GCP project to be created, modified, and deleted with Terraform. A project may be created, APIs enabled (e.g., Container Engine and GCS), and IAM policies applied.

### DIFF
--- a/builtin/providers/google/import_google_project_test.go
+++ b/builtin/providers/google/import_google_project_test.go
@@ -4,12 +4,14 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
 func TestAccGoogleProject_importBasic(t *testing.T) {
 	resourceName := "google_project.acceptance"
-	conf := fmt.Sprintf(testAccGoogleProject_basic, projectId)
+	projectId := "terraform-" + acctest.RandString(10)
+	conf := testAccGoogleProject_import(projectId, org, pname)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
@@ -26,4 +28,13 @@ func TestAccGoogleProject_importBasic(t *testing.T) {
 			},
 		},
 	})
+}
+
+func testAccGoogleProject_import(pid, orgId, projectName string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+    project_id = "%s"
+    org_id = "%s"
+    name = "%s"
+}`, pid, orgId, projectName)
 }
--- a/builtin/providers/google/resource_google_project.go
+++ b/builtin/providers/google/resource_google_project.go
@@ -13,9 +13,7 @@ import (
 )
 
 // resourceGoogleProject returns a *schema.Resource that allows a customer
-// to declare a Google Cloud Project resource. //
-// Only the 'policy' property of a project may be updated. All other properties
-// are computed.
+// to declare a Google Cloud Project resource.
 //
 // This example shows a project with a policy declared in config:
 //
@@ -25,27 +23,64 @@ import (
 // }
 func resourceGoogleProject() *schema.Resource {
 	return &schema.Resource{
+		SchemaVersion: 1,
+
 		Create: resourceGoogleProjectCreate,
 		Read:   resourceGoogleProjectRead,
 		Update: resourceGoogleProjectUpdate,
 		Delete: resourceGoogleProjectDelete,
+
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+		MigrateState: resourceGoogleProjectMigrateState,
 
 		Schema: map[string]*schema.Schema{
 			"id": &schema.Schema{
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:       schema.TypeString,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "The id field has unexpected behaviour and probably doesn't do what you expect. See https://www.terraform.io/docs/providers/google/r/google_project.html#id-field for more information. Please use project_id instead; future versions of Terraform will remove the id field.",
 			},
-			"policy_data": &schema.Schema{
+			"project_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				ForceNew: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// This suppresses the diff if project_id is not set
+					if new == "" {
+						return true
+					}
+					return false
+				},
+			},
+			"skip_delete": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
 			},
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
+				Optional: true,
 				Computed: true,
+			},
+			"org_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"policy_data": &schema.Schema{
+				Type:             schema.TypeString,
+				Optional:         true,
+				Computed:         true,
+				Deprecated:       "Use the 'google_project_iam_policy' resource to define policies for a Google Project",
+				DiffSuppressFunc: jsonPolicyDiffSuppress,
+			},
+			"policy_etag": &schema.Schema{
+				Type:       schema.TypeString,
+				Computed:   true,
+				Deprecated: "Use the the 'google_project_iam_policy' resource to define policies for a Google Project",
 			},
 			"number": &schema.Schema{
 				Type:     schema.TypeString,
@@ -55,20 +90,55 @@ func resourceGoogleProject() *schema.Resource {
 	}
 }
 
-// This resource supports creation, but not in the traditional sense.
-// A new Google Cloud Project can not be created. Instead, an existing Project
-// is initialized and made available as a Terraform resource.
 func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	project, err := getProject(d, config)
-	if err != nil {
-		return err
+	var pid string
+	var err error
+	pid = d.Get("project_id").(string)
+	if pid == "" {
+		pid, err = getProject(d, config)
+		if err != nil {
+			return fmt.Errorf("Error getting project ID: %v", err)
+		}
+		if pid == "" {
+			return fmt.Errorf("'project_id' must be set in the config")
+		}
 	}
 
-	d.SetId(project)
-	if err := resourceGoogleProjectRead(d, meta); err != nil {
-		return err
+	// we need to check if name and org_id are set, and throw an error if they aren't
+	// we can't just set these as required on the object, however, as that would break
+	// all configs that used previous iterations of the resource.
+	// TODO(paddy): remove this for 0.9 and set these attributes as required.
+	name, org_id := d.Get("name").(string), d.Get("org_id").(string)
+	if name == "" {
+		return fmt.Errorf("`name` must be set in the config if you're creating a project.")
+	}
+	if org_id == "" {
+		return fmt.Errorf("`org_id` must be set in the config if you're creating a project.")
+	}
+
+	log.Printf("[DEBUG]: Creating new project %q", pid)
+	project := &cloudresourcemanager.Project{
+		ProjectId: pid,
+		Name:      d.Get("name").(string),
+		Parent: &cloudresourcemanager.ResourceId{
+			Id:   d.Get("org_id").(string),
+			Type: "organization",
+		},
+	}
+
+	op, err := config.clientResourceManager.Projects.Create(project).Do()
+	if err != nil {
+		return fmt.Errorf("Error creating project %s (%s): %s.", project.ProjectId, project.Name, err)
+	}
+
+	d.SetId(pid)
+
+	// Wait for the operation to complete
+	waitErr := resourceManagerOperationWait(config, op, "project to create")
+	if waitErr != nil {
+		return waitErr
 	}
 
 	// Apply the IAM policy if it is set
@@ -76,15 +146,14 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 		// The policy string is just a marshaled cloudresourcemanager.Policy.
 		// Unmarshal it to a struct.
 		var policy cloudresourcemanager.Policy
-		if err = json.Unmarshal([]byte(pString.(string)), &policy); err != nil {
+		if err := json.Unmarshal([]byte(pString.(string)), &policy); err != nil {
 			return err
 		}
+		log.Printf("[DEBUG] Got policy from config: %#v", policy.Bindings)
 
 		// Retrieve existing IAM policy from project. This will be merged
 		// with the policy defined here.
-		// TODO(evanbrown): Add an 'authoritative' flag that allows policy
-		// in manifest to overwrite existing policy.
-		p, err := getProjectIamPolicy(project, config)
+		p, err := getProjectIamPolicy(pid, config)
 		if err != nil {
 			return err
 		}
@@ -95,47 +164,98 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 
 		// Apply the merged policy
 		log.Printf("[DEBUG] Setting new policy for project: %#v", p)
-		_, err = config.clientResourceManager.Projects.SetIamPolicy(project,
+		_, err = config.clientResourceManager.Projects.SetIamPolicy(pid,
 			&cloudresourcemanager.SetIamPolicyRequest{Policy: p}).Do()
 
 		if err != nil {
-			return fmt.Errorf("Error applying IAM policy for project %q: %s", project, err)
+			return fmt.Errorf("Error applying IAM policy for project %q: %s", pid, err)
 		}
 	}
-	return nil
+
+	return resourceGoogleProjectRead(d, meta)
 }
 
 func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
-	project, err := getProject(d, config)
+	pid := d.Id()
+
+	// Read the project
+	p, err := config.clientResourceManager.Projects.Get(pid).Do()
+	if err != nil {
+		if v, ok := err.(*googleapi.Error); ok && v.Code == http.StatusNotFound {
+			return fmt.Errorf("Project %q does not exist.", pid)
+		}
+		return fmt.Errorf("Error checking project %q: %s", pid, err)
+	}
+
+	d.Set("project_id", pid)
+	d.Set("number", strconv.FormatInt(int64(p.ProjectNumber), 10))
+	d.Set("name", p.Name)
+
+	if p.Parent != nil {
+		d.Set("org_id", p.Parent.Id)
+	}
+
+	// Read the IAM policy
+	pol, err := getProjectIamPolicy(pid, config)
 	if err != nil {
 		return err
 	}
-	d.SetId(project)
 
-	// Confirm the project exists.
-	// TODO(evanbrown): Support project creation
-	p, err := config.clientResourceManager.Projects.Get(project).Do()
+	polBytes, err := json.Marshal(pol)
 	if err != nil {
-		if v, ok := err.(*googleapi.Error); ok && v.Code == http.StatusNotFound {
-			return fmt.Errorf("Project %q does not exist. The Google provider does not currently support new project creation.", project)
-		}
-		return fmt.Errorf("Error checking project %q: %s", project, err)
+		return err
 	}
 
-	d.Set("number", strconv.FormatInt(int64(p.ProjectNumber), 10))
-	d.Set("name", p.Name)
+	d.Set("policy_etag", pol.Etag)
+	d.Set("policy_data", string(polBytes))
 
 	return nil
 }
 
 func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
-	project, err := getProject(d, config)
+	pid := d.Id()
+
+	// Read the project
+	// we need the project even though refresh has already been called
+	// because the API doesn't support patch, so we need the actual object
+	p, err := config.clientResourceManager.Projects.Get(pid).Do()
 	if err != nil {
-		return err
+		if v, ok := err.(*googleapi.Error); ok && v.Code == http.StatusNotFound {
+			return fmt.Errorf("Project %q does not exist.", pid)
+		}
+		return fmt.Errorf("Error checking project %q: %s", pid, err)
 	}
 
+	// Project name has changed
+	if ok := d.HasChange("name"); ok {
+		p.Name = d.Get("name").(string)
+		// Do update on project
+		p, err = config.clientResourceManager.Projects.Update(p.ProjectId, p).Do()
+		if err != nil {
+			return fmt.Errorf("Error updating project %q: %s", p.Name, err)
+		}
+	}
+
+	return updateProjectIamPolicy(d, config, pid)
+}
+
+func resourceGoogleProjectDelete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	// Only delete projects if skip_delete isn't set
+	if !d.Get("skip_delete").(bool) {
+		pid := d.Id()
+		_, err := config.clientResourceManager.Projects.Delete(pid).Do()
+		if err != nil {
+			return fmt.Errorf("Error deleting project %q: %s", pid, err)
+		}
+	}
+	d.SetId("")
+	return nil
+}
+
+func updateProjectIamPolicy(d *schema.ResourceData, config *Config, pid string) error {
 	// Policy has changed
 	if ok := d.HasChange("policy_data"); ok {
 		// The policy string is just a marshaled cloudresourcemanager.Policy.
@@ -152,15 +272,13 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
 			newPString = "{}"
 		}
 
-		oldPStringf, _ := json.MarshalIndent(oldPString, "", "   ")
-		newPStringf, _ := json.MarshalIndent(newPString, "", "   ")
-		log.Printf("[DEBUG]: Old policy: %v\nNew policy: %v", string(oldPStringf), string(newPStringf))
+		log.Printf("[DEBUG]: Old policy: %q\nNew policy: %q", oldPString, newPString)
 
 		var oldPolicy, newPolicy cloudresourcemanager.Policy
-		if err = json.Unmarshal([]byte(newPString), &newPolicy); err != nil {
+		if err := json.Unmarshal([]byte(newPString), &newPolicy); err != nil {
 			return err
 		}
-		if err = json.Unmarshal([]byte(oldPString), &oldPolicy); err != nil {
+		if err := json.Unmarshal([]byte(oldPString), &oldPolicy); err != nil {
 			return err
 		}
 
@@ -199,7 +317,7 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
 		// with the policy in the current state
 		// TODO(evanbrown): Add an 'authoritative' flag that allows policy
 		// in manifest to overwrite existing policy.
-		p, err := getProjectIamPolicy(project, config)
+		p, err := getProjectIamPolicy(pid, config)
 		if err != nil {
 			return err
 		}
@@ -218,86 +336,15 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 
 		p.Bindings = rolesToMembersBinding(mergedBindingsMap)
-		log.Printf("[DEBUG] Setting new policy for project: %#v", p)
-
 		dump, _ := json.MarshalIndent(p.Bindings, " ", "  ")
-		log.Printf(string(dump))
-		_, err = config.clientResourceManager.Projects.SetIamPolicy(project,
+		log.Printf("[DEBUG] Setting new policy for project: %#v:\n%s", p, string(dump))
+
+		_, err = config.clientResourceManager.Projects.SetIamPolicy(pid,
 			&cloudresourcemanager.SetIamPolicyRequest{Policy: p}).Do()
 
 		if err != nil {
-			return fmt.Errorf("Error applying IAM policy for project %q: %s", project, err)
+			return fmt.Errorf("Error applying IAM policy for project %q: %s", pid, err)
 		}
 	}
-
 	return nil
-}
-
-func resourceGoogleProjectDelete(d *schema.ResourceData, meta interface{}) error {
-	d.SetId("")
-	return nil
-}
-
-// Retrieve the existing IAM Policy for a Project
-func getProjectIamPolicy(project string, config *Config) (*cloudresourcemanager.Policy, error) {
-	p, err := config.clientResourceManager.Projects.GetIamPolicy(project,
-		&cloudresourcemanager.GetIamPolicyRequest{}).Do()
-
-	if err != nil {
-		return nil, fmt.Errorf("Error retrieving IAM policy for project %q: %s", project, err)
-	}
-	return p, nil
-}
-
-// Convert a map of roles->members to a list of Binding
-func rolesToMembersBinding(m map[string]map[string]bool) []*cloudresourcemanager.Binding {
-	bindings := make([]*cloudresourcemanager.Binding, 0)
-	for role, members := range m {
-		b := cloudresourcemanager.Binding{
-			Role:    role,
-			Members: make([]string, 0),
-		}
-		for m, _ := range members {
-			b.Members = append(b.Members, m)
-		}
-		bindings = append(bindings, &b)
-	}
-	return bindings
-}
-
-// Map a role to a map of members, allowing easy merging of multiple bindings.
-func rolesToMembersMap(bindings []*cloudresourcemanager.Binding) map[string]map[string]bool {
-	bm := make(map[string]map[string]bool)
-	// Get each binding
-	for _, b := range bindings {
-		// Initialize members map
-		if _, ok := bm[b.Role]; !ok {
-			bm[b.Role] = make(map[string]bool)
-		}
-		// Get each member (user/principal) for the binding
-		for _, m := range b.Members {
-			// Add the member
-			bm[b.Role][m] = true
-		}
-	}
-	return bm
-}
-
-// Merge multiple Bindings such that Bindings with the same Role result in
-// a single Binding with combined Members
-func mergeBindings(bindings []*cloudresourcemanager.Binding) []*cloudresourcemanager.Binding {
-	bm := rolesToMembersMap(bindings)
-	rb := make([]*cloudresourcemanager.Binding, 0)
-
-	for role, members := range bm {
-		var b cloudresourcemanager.Binding
-		b.Role = role
-		b.Members = make([]string, 0)
-		for m, _ := range members {
-			b.Members = append(b.Members, m)
-		}
-		rb = append(rb, &b)
-	}
-
-	return rb
 }
--- a/builtin/providers/google/resource_google_project_iam_policy.go
+++ b/builtin/providers/google/resource_google_project_iam_policy.go
@@ -1,0 +1,417 @@
+package google
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"sort"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+func resourceGoogleProjectIamPolicy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGoogleProjectIamPolicyCreate,
+		Read:   resourceGoogleProjectIamPolicyRead,
+		Update: resourceGoogleProjectIamPolicyUpdate,
+		Delete: resourceGoogleProjectIamPolicyDelete,
+
+		Schema: map[string]*schema.Schema{
+			"project": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"policy_data": &schema.Schema{
+				Type:             schema.TypeString,
+				Required:         true,
+				DiffSuppressFunc: jsonPolicyDiffSuppress,
+			},
+			"authoritative": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"etag": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"restore_policy": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"disable_project": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceGoogleProjectIamPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	pid := d.Get("project").(string)
+	// Get the policy in the template
+	p, err := getResourceIamPolicy(d)
+	if err != nil {
+		return fmt.Errorf("Could not get valid 'policy_data' from resource: %v", err)
+	}
+
+	// An authoritative policy is applied without regard for any existing IAM
+	// policy.
+	if v, ok := d.GetOk("authoritative"); ok && v.(bool) {
+		log.Printf("[DEBUG] Setting authoritative IAM policy for project %q", pid)
+		err := setProjectIamPolicy(p, config, pid)
+		if err != nil {
+			return err
+		}
+	} else {
+		log.Printf("[DEBUG] Setting non-authoritative IAM policy for project %q", pid)
+		// This is a non-authoritative policy, meaning it should be merged with
+		// any existing policy
+		ep, err := getProjectIamPolicy(pid, config)
+		if err != nil {
+			return err
+		}
+
+		// First, subtract the policy defined in the template from the
+		// current policy in the project, and save the result. This will
+		// allow us to restore the original policy at some point (which
+		// assumes that Terraform owns any common policy that exists in
+		// the template and project at create time.
+		rp := subtractIamPolicy(ep, p)
+		rps, err := json.Marshal(rp)
+		if err != nil {
+			return fmt.Errorf("Error marshaling restorable IAM policy: %v", err)
+		}
+		d.Set("restore_policy", string(rps))
+
+		// Merge the policies together
+		mb := mergeBindings(append(p.Bindings, rp.Bindings...))
+		ep.Bindings = mb
+		if err = setProjectIamPolicy(ep, config, pid); err != nil {
+			return fmt.Errorf("Error applying IAM policy to project: %v", err)
+		}
+	}
+	d.SetId(pid)
+	return resourceGoogleProjectIamPolicyRead(d, meta)
+}
+
+func resourceGoogleProjectIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Reading google_project_iam_policy")
+	config := meta.(*Config)
+	pid := d.Get("project").(string)
+
+	p, err := getProjectIamPolicy(pid, config)
+	if err != nil {
+		return err
+	}
+
+	var bindings []*cloudresourcemanager.Binding
+	if v, ok := d.GetOk("restore_policy"); ok {
+		var restored cloudresourcemanager.Policy
+		// if there's a restore policy, subtract it from the policy_data
+		err := json.Unmarshal([]byte(v.(string)), &restored)
+		if err != nil {
+			return fmt.Errorf("Error unmarshaling restorable IAM policy: %v", err)
+		}
+		subtracted := subtractIamPolicy(p, &restored)
+		bindings = subtracted.Bindings
+	} else {
+		bindings = p.Bindings
+	}
+	// we only marshal the bindings, because only the bindings get set in the config
+	pBytes, err := json.Marshal(&cloudresourcemanager.Policy{Bindings: bindings})
+	if err != nil {
+		return fmt.Errorf("Error marshaling IAM policy: %v", err)
+	}
+	log.Printf("[DEBUG]: Setting etag=%s", p.Etag)
+	d.Set("etag", p.Etag)
+	d.Set("policy_data", string(pBytes))
+	return nil
+}
+
+func resourceGoogleProjectIamPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Updating google_project_iam_policy")
+	config := meta.(*Config)
+	pid := d.Get("project").(string)
+
+	// Get the policy in the template
+	p, err := getResourceIamPolicy(d)
+	if err != nil {
+		return fmt.Errorf("Could not get valid 'policy_data' from resource: %v", err)
+	}
+	pBytes, _ := json.Marshal(p)
+	log.Printf("[DEBUG] Got policy from config: %s", string(pBytes))
+
+	// An authoritative policy is applied without regard for any existing IAM
+	// policy.
+	if v, ok := d.GetOk("authoritative"); ok && v.(bool) {
+		log.Printf("[DEBUG] Updating authoritative IAM policy for project %q", pid)
+		err := setProjectIamPolicy(p, config, pid)
+		if err != nil {
+			return fmt.Errorf("Error setting project IAM policy: %v", err)
+		}
+		d.Set("restore_policy", "")
+	} else {
+		log.Printf("[DEBUG] Updating non-authoritative IAM policy for project %q", pid)
+		// Get the previous policy from state
+		pp, err := getPrevResourceIamPolicy(d)
+		if err != nil {
+			return fmt.Errorf("Error retrieving previous version of changed project IAM policy: %v", err)
+		}
+		ppBytes, _ := json.Marshal(pp)
+		log.Printf("[DEBUG] Got previous version of changed project IAM policy: %s", string(ppBytes))
+
+		// Get the existing IAM policy from the API
+		ep, err := getProjectIamPolicy(pid, config)
+		if err != nil {
+			return fmt.Errorf("Error retrieving IAM policy from project API: %v", err)
+		}
+		epBytes, _ := json.Marshal(ep)
+		log.Printf("[DEBUG] Got existing version of changed IAM policy from project API: %s", string(epBytes))
+
+		// Subtract the previous and current policies from the policy retrieved from the API
+		rp := subtractIamPolicy(ep, pp)
+		rpBytes, _ := json.Marshal(rp)
+		log.Printf("[DEBUG] After subtracting the previous policy from the existing policy, remaining policies: %s", string(rpBytes))
+		rp = subtractIamPolicy(rp, p)
+		rpBytes, _ = json.Marshal(rp)
+		log.Printf("[DEBUG] After subtracting the remaining policies from the config policy, remaining policies: %s", string(rpBytes))
+		rps, err := json.Marshal(rp)
+		if err != nil {
+			return fmt.Errorf("Error marhsaling restorable IAM policy: %v", err)
+		}
+		d.Set("restore_policy", string(rps))
+
+		// Merge the policies together
+		mb := mergeBindings(append(p.Bindings, rp.Bindings...))
+		ep.Bindings = mb
+		if err = setProjectIamPolicy(ep, config, pid); err != nil {
+			return fmt.Errorf("Error applying IAM policy to project: %v", err)
+		}
+	}
+
+	return resourceGoogleProjectIamPolicyRead(d, meta)
+}
+
+func resourceGoogleProjectIamPolicyDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Deleting google_project_iam_policy")
+	config := meta.(*Config)
+	pid := d.Get("project").(string)
+
+	// Get the existing IAM policy from the API
+	ep, err := getProjectIamPolicy(pid, config)
+	if err != nil {
+		return fmt.Errorf("Error retrieving IAM policy from project API: %v", err)
+	}
+	// Deleting an authoritative policy will leave the project with no policy,
+	// and unaccessible by anyone without org-level privs. For this reason, the
+	// "disable_project" property must be set to true, forcing the user to ack
+	// this outcome
+	if v, ok := d.GetOk("authoritative"); ok && v.(bool) {
+		if v, ok := d.GetOk("disable_project"); !ok || !v.(bool) {
+			return fmt.Errorf("You must set 'disable_project' to true before deleting an authoritative IAM policy")
+		}
+		ep.Bindings = make([]*cloudresourcemanager.Binding, 0)
+
+	} else {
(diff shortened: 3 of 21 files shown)


evandbrown:
cc @danawillow

builtin/providers/google/resource_google_project.go
```diff
@@ -31,18 +32,29 @@ func resourceGoogleProject() *schema.Resource {
 		Delete: resourceGoogleProjectDelete,
 
 		Schema: map[string]*schema.Schema{
-			"id": &schema.Schema{
+			"project_id": &schema.Schema{
```
paddyforan: This is a backwards-incompatible change, isn't it? Do we need a state migration here?

evandbrown:
@paddyforan @stack72 This change introduces a few new requirements for the GCP project that runs integration tests:

1. The project must be a member of an [organization](https://cloud.google.com/resource-manager/docs/quickstart)
1. The service account that runs the tests must have `roles/resourcemanager.projectCreator` in the organization
1. `GOOGLE_ORG` must be set to that org's idea when the tests run

I'm available to discuss here or off-thread at your leisure. Thank you!

builtin/providers/google/resource_google_project.go
```diff
@@ -31,18 +32,29 @@ func resourceGoogleProject() *schema.Resource {
...
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-			"policy_data": &schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
```
paddyforan: This looks like it's also a breaking change. Didn't this used to be computed? Is it not anymore? Why not?

builtin/providers/google/resource_google_project.go
```diff
@@ -58,81 +70,141 @@ func resourceGoogleProject() *schema.Resource {
...
 	}
 
-	// Apply the IAM policy if it is set
+	// Apply IAM policy
 	if pString, ok := d.GetOk("policy_data"); ok {
 		// The policy string is just a marshaled cloudresourcemanager.Policy.
 		// Unmarshal it to a struct.
-		var policy cloudresourcemanager.Policy
+		var policy *cloudresourcemanager.Policy
 		if err = json.Unmarshal([]byte(pString.(string)), &policy); err != nil {
```
paddyforan: Now that you've changed `policy` to a pointer, aren't we taking the address of a pointer here?

builtin/providers/google/resource_google_project.go
```diff
@@ -58,81 +70,141 @@ func resourceGoogleProject() *schema.Resource {
...
-
-		// Merge the existing policy bindings with those defined in this manifest.
-		p.Bindings = mergeBindings(append(p.Bindings, policy.Bindings...))
-
-		// Apply the merged policy
-		log.Printf("[DEBUG] Setting new policy for project: %#v", p)
-		_, err = config.clientResourceManager.Projects.SetIamPolicy(project,
-			&cloudresourcemanager.SetIamPolicyRequest{Policy: p}).Do()
+		// Apply the policy
+		log.Printf("[DEBUG] Setting policy for project: %#v", policy)
```
paddyforan: This used to merge the old policy in before setting the new one. Can you help me understand this change in behaviour?

builtin/providers/google/resource_google_project.go
```diff
@@ -58,81 +70,141 @@ func resourceGoogleProject() *schema.Resource {
...
 		if v, ok := err.(*googleapi.Error); ok && v.Code == http.StatusNotFound {
-			return fmt.Errorf("Project %q does not exist. The Google provider does not currently support new project creation.", project)
+			return fmt.Errorf("Project %q does not exist.", pid)
 		}
-		return fmt.Errorf("Error checking project %q: %s", project, err)
+		return fmt.Errorf("Error checking project %q: %s", pid, err)
 	}
 
 	d.Set("number", strconv.FormatInt(int64(p.ProjectNumber), 10))
-	d.Set("name", p.Name)
```
paddyforan: Why is name no longer being read?

builtin/providers/google/resource_google_project.go
```diff
@@ -298,3 +376,65 @@ func mergeBindings(bindings []*cloudresourcemanager.Binding) []*cloudresourceman
...
+
+		for _, v := range oldSs.List() {
+			oldMap[v.(string)] = true
+		}
+		for _, v := range newSs.List() {
+			newMap[v.(string)] = true
+		}
+
+		added = make([]string, 0)
+		removed = make([]string, 0)
```
paddyforan: nit: `var added, removed []string` does effectively the same thing and is a bit clearer, in my opinion.

builtin/providers/google/resource_google_project.go
```diff
@@ -298,3 +376,65 @@ func mergeBindings(bindings []*cloudresourcemanager.Binding) []*cloudresourceman
...
+				continue
+			}
+			removed = append(removed, k)
+		}
+
+		// Find added services
+		for k, _ := range newMap {
+			if !oldMap[k] {
+				added = append(added, k)
+				continue
```
paddyforan: nit: no real point to this `continue`, though it doesn't do any harm, either.

builtin/providers/google/resource_storage_object_acl.go
```diff
@@ -150,18 +150,9 @@ func resourceStorageObjectAclRead(d *schema.ResourceData, meta interface{}) erro
...
-				} else if key == "entity" {
-					entity = val.(string)
-				}
-			}
-			if _, in := re_local_map[entity]; in {
-				role_entity = append(role_entity, fmt.Sprintf("%s:%s", role, entity))
-				log.Printf("[DEBUG]: saving re %s-%s", role, entity)
+			if _, in := re_local_map[v.Entity]; in {
+				role_entity = append(role_entity, fmt.Sprintf("%s:%s", v.Role, v.Entity))
+				log.Printf("[DEBUG]: saving re %s-%s", v.Role, v.Entity)
```
paddyforan: These changes seem unrelated to the PR. Am I missing something?

builtin/providers/google/resourcemanager_operation.go
```diff
@@ -0,0 +1,67 @@
...
+
+type ResourceManagerOperationWaiter struct {
+	Service *cloudresourcemanager.Service
+	Op      *cloudresourcemanager.Operation
+}
+
+func (w *ResourceManagerOperationWaiter) RefreshFunc() resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		var op *cloudresourcemanager.Operation
+		var err error
```
paddyforan: Why declare these at the top instead of just instantiating and assigning at once on line 22?

website/source/docs/providers/google/r/google_project.html.markdown
```diff
@@ -38,8 +43,16 @@ data "google_iam_policy" "admin" {
...
 
-* `id` - (Required) The project ID.
-    Changing this forces a new project to be referenced.
+* `project_id` - (Required) The project ID.
+    Changing this forces a new project to be created.
+
+* `org_id` - (Required) The numeric ID of the organization this project belongs to.
+    Changing this forces a new project to be created.
+
+* `name` - (Optional) The display name of the project.
```
paddyforan: This says optional, but the schema says required. Which is right?

builtin/providers/google/resource_google_project.go
paddyforan: Update: I lied. You don't even need these two lines at all, the named return variables already are the equivalent of `var added, removed []string`.

cblecker:
This is great, and I'd love to see this in 0.8 if timing works out :)

builtin/providers/google/resource_google_project.go
evandbrown: Great catch. I'll get this added.

builtin/providers/google/resource_google_project.go
evandbrown: The initial version of this resource didn't support create. It assumed the resource (project) already existed, and it read/computed the values from the existing resource.

I'm wondering if it's worth breaking this out into two resources (`google_project` and `google_imported_project`), or adding an `imported` attribute? The latter is simpler, but would have issues with computed vs non-computed values. Thoughts?

builtin/providers/google/resource_google_project.go
evandbrown: I...don't know why I did this.

builtin/providers/google/resource_google_project.go
evandbrown: I mentioned this in an earlier comment, but basically we're assuming this resource creates new projects, vs importing existing ones now. I'm becoming more convinced that we want a `google_project` and `google_imported_project` to differentiate the two.

builtin/providers/google/resource_google_project.go
evandbrown: It's a required property now, and not computed. Is it best-practice to read-and-set all values?

builtin/providers/google/resource_storage_object_acl.go
evandbrown: The main feature of this PR required an update to a vendored library which required this change :-\

builtin/providers/google/resourcemanager_operation.go
evandbrown: Good point!

evandbrown:
@paddyforan **thank you** for such a quick and thorough review! Really great feedback there.

I had a serendipitous meeting this morning with @wendorf. He asked about this exact feature, and had some great real-world feedback. In a nutshell, I'd like to break out the `data_policy` and `services` properties of `google_project` into separate resources (`google_project_iam_policy` and `google_project_services`) that each take a project ID and apply an IAM policy or service configuration. This simplifies how we think about a `google_project`, and would let us import an existing project or create a new one if it doesn't exist.

On that last point, how do you think about handling imports of something that already exists? Should we auto-detect existence on create? Add an `imported` property to the resource?

builtin/providers/google/resource_google_project.go
paddyforan: Fair enough, this makes a tonne more sense to me now. Thanks :D I'll address the import stuff all at once in a separate comment (so it's easier to find in the future) if that's ok. :)

builtin/providers/google/resource_google_project.go
paddyforan: Ah, this makes total sense to me now. I'm less convinced, in terms of what the end result should look like, but am still thinking through the road to get there. :)

builtin/providers/google/resource_google_project.go
paddyforan: Oh jeez, I don't know about "best practice". But I imagine for things like refreshes, it'd be helpful to have it? That way `terraform import` (I think) will work properly. The idea being that if we set _all_ the properties in the state on read (even the ones we know are set in the config) then when we don't instantiate a resource from a config (e.g., when using `terraform import`) those properties will still be properly set. Or at least that's my rationale at the moment, I could be misunderstanding how any number of things work.

builtin/providers/google/resource_storage_object_acl.go
paddyforan: OH! Fair enough. :)

paddyforan:
@evandbrown thanks for the wonderful PR! This is some exciting new functionality, and I'd really like to see it in Terraform.

I think breaking out data policies and services is a great idea, and simplifies thinking about projects a little bit.

As for importing, I've thought about it a bit, and I think it's important to stay with the way Terraform behaves with other resources. That is to say, assume Terraform owns the resource, and will blindly reshape it to match the config, overwriting any changes made outside of Terraform.

When we run into projects that already exist, things get trickier. The "pure" approach, from what I can tell, is to assume the user wants to create a Terraform-managed project any time the project resource is used. In practice, though, I think people will want to reference their existing projects using interpolation in their configs. `terraform import` could load those projects for people, and that's an option, but they'd have to write their own config for it... though I'm not sure "add three lines to your config" is too much to ask?

My gut says right now to have the config assume the user wants to set the project information (blowing away anything not in the config) and rely on terraform import for projects that exist already. Separating out the IAM policy and enabled services seems like it limits the impact of that--looking at the API, it seems like the most damage we could cause would be removing labels.

Of course, this would mean we need to make projects [importable](https://www.terraform.io/docs/plugins/provider.html#Importer) but that seems like a really minor amount of work. I can even supply a patch--I just tested it out locally to make sure I wasn't speaking nonsense. :)

What do you think?

evandbrown:
@paddyforan This all sounds good! Here's my takeaway:

* You'll send a patch for import (Thank you! Haven't done one of these and looking forward to seeing it.)
* I'll break out IAM and services into `google_project_iam_policy` and `google_project_services`
* I'll add a new `google_project_labels` resource to address the astute observation you made about overwriting labels
* The `google_project` resources is for creating new projects. A user can use `import` to bind an existing project. Terraform owns that project.

Shout-out to @mitchellh and @armon for hiring @paddyforan. You're awesome, Paddy!

paddyforan:
Hey @evandbrown, sorry for the late reply, especially after you said such nice things about me (I blushed 😊)

That takeaway sounds good to me. I'm hoping to open that PR this morning, I just need to add a test case--yesterday got away from me with non-code stuff :(

But that sounds like a great path forward. If you want me to tackle some of those items in the interest of getting support for this released soon, just shout, I'm happy to help--just don't want to step on toes or duplicate effort. :)

evandbrown:
No worries and no rush! Wait, actually, what is the deadline for the next
release?

paddyforan:
Hm. I don't actually know if a date has been set yet. Importing projects got merged in today, so that's now in 0.8. :) Shout if you want me to pitch in on some of these next steps.

mitchellh:
@evandbrown A deadline for 0.8 final does exist internally, its mid-December. We have another RC coming out tomorrow. We're trying to not introduce any more core backwards incompats, but resources are okay still.

evandbrown:
Whew. That took a bit longer than I thought, @paddyforan! These latest changes:

1. Rebased off master, including your import changes (thanks!)
1. `google_project` continues to function as before (a user can embed IAM policy and services in that resource), but also adds `google_project_iam_policy` and `google_project_services` resources to allow easily modifying an existing project.
1. I didn't get labels in. Will do that in a separate PR.

Critical feedback welcome!

evandbrown:
@paddyforan conflicts resolved. When you have a moment, would appreciate any feedback and see if we can get this merged. Thank you!

paddyforan:
Taking a look at it right now. :)

builtin/providers/google/resource_google_project.go
```diff
@@ -55,87 +65,195 @@ func resourceGoogleProject() *schema.Resource {
...
+		if v, ok := err.(*googleapi.Error); ok && v.Code == http.StatusNotFound {
+			return fmt.Errorf("Project %q does not exist.", pid)
 		}
-		log.Printf("[DEBUG] Got existing bindings from project: %#v", p.Bindings)
+		return fmt.Errorf("Error checking project %q: %s", pid, err)
+	}
 
-		// Merge the existing policy bindings with those defined in this manifest.
-		p.Bindings = mergeBindings(append(p.Bindings, policy.Bindings...))
+	d.Set("number", strconv.FormatInt(int64(p.ProjectNumber), 10))
```
paddyforan: I think we're missing a few attributes we should be setting here. I think we need to `Set` the rest of the properties, so we can properly diff.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,87 +65,195 @@ func resourceGoogleProject() *schema.Resource {
...
+	// Project name has changed
+	if ok := d.HasChange("name"); ok {
+		p.Name = d.Get("name").(string)
+		// Do update on project
+		p, err = config.clientResourceManager.Projects.Update(p.ProjectId, p).Do()
 		if err != nil {
-			return fmt.Errorf("Error applying IAM policy for project %q: %s", project, err)
+			return fmt.Errorf("Error updating project %q: %s", p.Name, err)
 		}
 	}
```
paddyforan: Should we check for policy data updating, too?

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+	d.SetId(pid)
+	return nil
+}
+
+func resourceGoogleProjectIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	// Read is a no-op as we don't care about the actual state of the project's IAM
+	// policy. We always overwrite it during create or update
```
paddyforan: I think we need to read it anyways, just so we can set it in state to diff against. Will run some tests to find out if I'm wrong. :)

builtin/providers/google/resource_google_project_services.go
```diff
@@ -0,0 +1,63 @@
...
+				Set:      schema.HashString,
+			},
+		},
+	}
+}
+
+func resourceGoogleProjectServicesCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	pid := d.Get("project_id").(string)
+	err := setProjectIamPolicy(d, config, pid)
```
paddyforan: Is that the right function to call here?

builtin/providers/google/resource_google_project_services.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return err
+	}
+	d.SetId(pid)
+	return nil
+}
+
+func resourceGoogleProjectServicesRead(d *schema.ResourceData, meta interface{}) error {
+	// Read is a no-op as we don't care about the actual state of the project's IAM
+	// policy. We always overwrite it during create or update
```
paddyforan: We seem to have gotten IAM and services mixed up here, unless I'm misunderstanding. I think we _always_ want a read though, for things like refreshing state and for importing.

builtin/providers/google/resource_google_project_services.go
```diff
@@ -0,0 +1,63 @@
...
+	return nil
+}
+
+func resourceGoogleProjectServicesUpdate(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Deleting google_project_iam_policy")
+	config := meta.(*Config)
+	pid := d.Get("project_id").(string)
+	return updateProjectIamPolicy(d, config, pid)
```
paddyforan: I'm confused about what's happening here. Why is updating deleting stuff? (Also, we have another rogue IAM reference here)

builtin/providers/google/resource_google_project_services.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func resourceGoogleProjectServicesDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Deleting google_project_iam_policy")
+	if err := resourceGoogleProjectServicesUpdate(d, meta); err != nil {
```
paddyforan: Can you walk me through why we're updating when deleting? There's a logic here I'm not quite following, and feeling pretty sure I missed something.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,87 +65,195 @@ func resourceGoogleProject() *schema.Resource {
...
+	// Wait for the operation to complete
+	waitErr := resourceManagerOperationWait(config, op, "project to create")
+	if waitErr != nil {
+		// The resource didn't actually create
+		d.SetId("")
+		return waitErr
+	}
+
+	// Set the resourc ID
+	d.SetId(project.ProjectId)
```
evandbrown: Got it.

builtin/providers/google/resource_google_project.go
evandbrown: Ahh, so this would read the actual state from the APIs (which accounts for the possibility that non-computed values were changed out-of-band?) Makes sense, will correct.

builtin/providers/google/resource_google_project_iam_policy.go
evandbrown: Ahh yes. I was assuming that since we're merging the policy expressed in Terraform with the actual project policy (unless it's overridden with `authoritative`), we wouldn't care about the actual state. But reading is required to know there's been a change, so update is triggered. Duh. Thanks!

builtin/providers/google/resource_google_project_services.go
evandbrown: Blah, forgot to apply a stash and left all of the services stuff in a bad place. Will be fixed in next push.

builtin/providers/google/resource_google_project_services.go
evandbrown: Deletion is really dissociating a service from a project, and accounts for services that were bound outside of Terraform. The update code supports adding or removing services from existing projects that may have out-of-band service bindings. We're treating delete like an "update project to remove <set> of service bindings".

evandbrown:
Thanks for catching those things, @paddyforan! I'll submit fixes in a new commit tomorrow.

evandbrown:
@paddyforan Took longer than I thought, but I think we're there. We now handle services and IAM policy as separate resources, which is simpler when binding those to existing projects.

When running all of the acceptance tests in the "GoogleProject*" namespace, I noticed the import test fails. It seems like it tries to create a project that already exists. I'm unfamiliar with the import process, and since you added that feature I thought you might have some insight.

Very happy that my last commit of 2016 is to Terraform. Wishing everyone at HashiCorp a happy 2017!

paddyforan:
Hey @evandbrown! Thanks for the update. I'm going to take a look at this today. And super honoured to be your last commit in 2016, of course! Thanks so much for all the work you've put in. A very happy 2017 to you, as well--I hope we hear a lot from each other this year. 😀 

> When running all of the acceptance tests in the "GoogleProject*" namespace, I noticed the import test fails. It seems like it tries to create a project that already exists. I'm unfamiliar with the import process, and since you added that feature I thought you might have some insight.

That is interesting indeed. I'm going to go ahead and dig into this and report back with my findings. In the meantime, we'll try to get this merged. I'm realising the original projects resource probably should have been a data source, which exacerbates this PR because of backwards compatibility concerns, but no sense in crying over spilt milk. :)

I'm going to have to look for guidance from the rest of the team as to whether this needs to be entirely backwards compatible or wait for 0.9, or if we can live with some backwards incompatibilities in the name of getting this in for 0.8. I'll report back with that, too.

evandbrown:
Thanks, @paddyforan! I just pushed a change the preserves backwards-compatibility. I'm working on a separate commit to do a small migration, as well as a change that should fix the importer. Will ping when that's in.

evandbrown:
This should now be entirely backwards compatible. The original version of `google_project` existed before there was an API for creating or deleting projects. But there was an API for applying an IAM policy to a project. The original resource would effectively import a project during the create lifecycle event, and associate an IAM policy. The update lifecycle event would update the IAM policy. The delete lifecycle event would dissociate the IAM policy, but would not delete the project.

IAM policy association became more nuanced when projects became fungible, so policies were moved into their own resources. We're now marking the `policy_data` attribute as deprecated, but it behaves as always.

A state migration preserves the "do not delete the project" behavior of the previous version by setting a `skip_delete` attribute. Only `google_project` resources created after these changes (and therefore not subject to the state migration) can be deleted.

Original acceptance tests verify the backwards-compatible behavior. Additionally, the read lifecycle event was fixed to address the problem identified by the importer test. Here's acceptance test output:

```
$ make testacc TEST=./builtin/providers/google ... TESTARGS='-run=TestAccGoogle
Project...'
2017/01/05 21:08:11 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccGoogleProject... -timeout 120m
=== RUN   TestAccGoogleProject_importBasic
--- PASS: TestAccGoogleProject_importBasic (2.62s)
=== RUN   TestAccGoogleProjectIamPolicy_basic
--- PASS: TestAccGoogleProjectIamPolicy_basic (27.34s)
=== RUN   TestAccGoogleProjectServices_basic
--- PASS: TestAccGoogleProjectServices_basic (72.73s)
=== RUN   TestAccGoogleProject_create
--- PASS: TestAccGoogleProject_create (1.83s)
=== RUN   TestAccGoogleProject_iamPolicy1
--- PASS: TestAccGoogleProject_iamPolicy1 (8.71s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/google 113.291s
```

builtin/providers/google/resource_google_service_account_test.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
+var (
+	projectId = multiEnvSearch([]string{
+		"GOOGLE_PROJECT",
+		"GCLOUD_PROJECT",
+		"CLOUDSDK_CORE_PROJECT",
+	})
+)
```
paddyforan: Hmm. What brought about this addition?

builtin/providers/google/resource_google_project.go
```diff
@@ -25,28 +23,54 @@ import (
...
 			State: schema.ImportStatePassthrough,
 		},
+		MigrateState: resourceGoogleProjectMigrateState,
 
 		Schema: map[string]*schema.Schema{
 			"id": &schema.Schema{
 				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"project_id": &schema.Schema{
```
evandbrown: Ahh yes. `d.Id()` and `d.Get("id")` don't seem to return anything during create, even if they're set in the manifest. We were [testing with an ID](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_google_project_test.go#L448), but that was being ignored as [project was being assigned](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_google_project.go#L64) from the provider. So as far as I can tell, `id` appears to be reserved for state only. Does that sound correct?

builtin/providers/google/resource_google_project.go
```diff
@@ -55,36 +79,57 @@ func resourceGoogleProject() *schema.Resource {
...
 func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	project, err := getProject(d, config)
-	if err != nil {
-		return err
-	}
+	pid := d.Get("project_id").(string)
+
+	// See if project already exists. If it doesn't, create it.
```
evandbrown: Simply UX (and we weren't aware of import at the time this was built). Seems we can drop this behavior on create, since creating new things doesn't have the backwards-compat requirements of the other lifecycle events.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,36 +79,57 @@ func resourceGoogleProject() *schema.Resource {
...
+	if _, err := config.clientResourceManager.Projects.Get(pid).Do(); err != nil {
```
evandbrown: Fixed (this code path is gone; no longer accounting for project import in the create lifecycle)

builtin/providers/google/resource_google_project.go
```diff
@@ -222,82 +287,124 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
...
 			&cloudresourcemanager.SetIamPolicyRequest{Policy: p}).Do()
 
 		if err != nil {
-			return fmt.Errorf("Error applying IAM policy for project %q: %s", project, err)
+			return fmt.Errorf("Error applying IAM policy for project %q: %s", pid, err)
 		}
 	}
 
-	return nil
+	return resourceGoogleProjectCreate(d, meta)
```
evandbrown: Fixed.

builtin/providers/google/resource_google_project.go
```diff
@@ -222,82 +287,124 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
...
+		// JSON Unmarshaling would fail
+		if oldPString == "" {
+			oldPString = "{}"
+		}
+		if newPString == "" {
+			newPString = "{}"
+		}
+
+		oldPStringf, _ := json.MarshalIndent(oldPString, "", "   ")
+		newPStringf, _ := json.MarshalIndent(newPString, "", "   ")
```
evandbrown: Fixed.

builtin/providers/google/resource_google_project.go
```diff
@@ -222,82 +287,124 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
...
-	// Get each binding
-	for _, b := range bindings {
-		// Initialize members map
-		if _, ok := bm[b.Role]; !ok {
-			bm[b.Role] = make(map[string]bool)
+		// Find any Roles and Members that were removed (i.e., those that are present
+		// in the old but absent in the new
+		oldMap := rolesToMembersMap(oldPolicy.Bindings)
+		newMap := rolesToMembersMap(newPolicy.Bindings)
+		deleted := make(map[string]map[string]bool)
```
evandbrown: That would work. `if deleted["role"]["member"]` is just simpler to check for the existence of a person in a role.

builtin/providers/google/resource_google_project.go
```diff
@@ -222,82 +287,124 @@ func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error
...
 		}
-		rb = append(rb, &b)
-	}
 
-	return rb
+		p.Bindings = rolesToMembersBinding(mergedBindingsMap)
+		log.Printf("[DEBUG] Setting new policy for project: %#v", p)
+
+		dump, _ := json.MarshalIndent(p.Bindings, " ", "  ")
+		log.Printf(string(dump))
```
evandbrown: Fixed.

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+	}
+	return policy, nil
+}
+
+// Get the restore_policy that can be used to restore a project's IAM policy to its
+// state before it was adopted into Terraform
+func getRestoreIamPolicy(d *schema.ResourceData) (*cloudresourcemanager.Policy, error) {
+	if v, ok := d.GetOk("restore_policy"); ok {
+		var policy *cloudresourcemanager.Policy = &cloudresourcemanager.Policy{}
```
evandbrown: Fixed.

builtin/providers/google/resource_google_service_account_test.go
evandbrown: This is a simpler version of a `schema.MultiEnvDefaultFunc`. Could abstrace that out in `google/provider.go` and re-use here. Will add a TODO for that.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,36 +81,53 @@ func resourceGoogleProject() *schema.Resource {
...
-	d.SetId(project)
-	if err := resourceGoogleProjectRead(d, meta); err != nil {
-		return err
+	// Wait for the operation to complete
+	waitErr := resourceManagerOperationWait(config, op, "project to create")
+	if waitErr != nil {
+		return waitErr
 	}
 
+	d.SetId(pid)
```
paddyforan: If you put the call to `SetId` before the wait, a timeout will be recoverable (`terraform plan` or `terraform apply` will notice that the project exists and not try to recreate it).

builtin/providers/google/resource_google_project.go
```diff
@@ -95,45 +138,69 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+	if ok := d.HasChange("name"); ok {
+		p.Name = d.Get("name").(string)
+		// Do update on project
+		p, err = config.clientResourceManager.Projects.Update(p.ProjectId, p).Do()
+		if err != nil {
+			return fmt.Errorf("Error updating project %q: %s", p.Name, err)
+		}
 	}
 
 	// Policy has changed
```
paddyforan: Is this not the same as `updateProjectIamPolicy`? It looks like it does the same thing. Maybe use that new abstraction here?

builtin/providers/google/resource_google_project_services.go
```diff
@@ -0,0 +1,200 @@
...
+		return err
+	}
+	d.SetId(pid)
+	return resourceGoogleProjectServicesRead(d, meta)
+}
+
+func resourceGoogleProjectServicesRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	// Put the services defined in the manifest in a map
```
paddyforan: This is actually "services defined in the config/manifest file _or_ that exists in the state file, IIUC.

builtin/providers/google/resource_google_project_services.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+
+	services, err := config.clientServiceMan.Services.List().ConsumerId("project:" + d.Id()).Do()
+	if err != nil {
+		return err
+	}
+
+	// Inventory and record actual services state, but only care about
+	// services that are in the manifest
```
paddyforan: I'm nervous about the drift this could open us up to--services changing outside the manifest means the manifest no longer documents which services are enabled. This also means our diffs get weird and we can no longer accurately remove services from a project. Kinda concerned about the implications of this.

builtin/providers/google/resource_google_project.go
evandbrown: Fixed.

builtin/providers/google/resource_google_project_services.go
evandbrown: Fixed.

builtin/providers/google/resource_google_project_services.go
evandbrown: Fixed. On create, existing services that aren't in the config are removed. Same on update. I like the authoritative-only approach.

evandbrown:
@paddyforan Thanks for the latest review. Addressed all of your points, chief of which is making service management authoritative.

paddyforan:
Hey @evandbrown! Thanks for the updates. And man, I kinda feel like the worst right now.

I'm using this config:

```tf
resource "google_project" "acctest" {
  id = "hashicorp-terraform-testing"
}
```

* removing `getProject` broke the existing behaviour--if I run `terraform apply` on that config under 0.8.4, then run it against your branch, I get the following error:
    ```
    * google_project.acctest: One of 'project_id' or 'id' must be set in the config
    ```
    
    The upshot of this is that merging this as-is would mean that anyone using the `google_project` resource would have to update their config, which I'd like to avoid. I'm pretty sure if, instead of trying to read the `id` key, you use `getProject` like the old code did, that will work as intended (using `project_id` when set, but falling back on old behaviour otherwise).
* I was talking about it last night a bit after hours with some of the other core team, and my understanding is we _really_ want to keep Terraform's behaviour consistent across providers, which means the conf file should be considered authoritative pretty much everywhere. This means we should drop the merging of IAM policies. If you want to talk about it some more, that's cool, and I'm happy to have the conversation, but from my understanding the core team's stance on the issue is any time the provider can drift from the config file without Terraform noticing or trying to bring it back inline, it's a bug. If there's a special case for IAM in this, I'd love to hear more about it.

Really sorry to move the goalposts on that one on you. :(

For the sake of getting this merged (I still want to hit that merge button this week), I'm going to take a crack at updating this with the changes above to do some more backwards compatibility testing. I'm happy to contribute those changes back to the PR (if you [enable edits from maintainers on the PR](https://github.com/blog/2247-improving-collaboration-with-forks) or want me to just PR them into your branch) but I also don't want to step on your toes--you've worked on this for a long time and been very patient, so I'm not bothered by doing some exploratory coding to uncover all the potential problems, then just deleting it and letting you push this across the finish line. Just let me know how you want to proceed. :)

evandbrown:
Hey @paddyforan! The `getProject` piece sounds good. I'll get that added back today.

I do think IAM policy merging is a behavior we want for project IAM. If you get an IAM policy wrong (and it happens a lot), there's a good chance you lose access to your project and have to find an admin to re-enable it. This is particularly true with the `google_project_iam_policy` resource that allows attachment to an existing project. The workflow we're supporting here is to set `authoritative=false` while you're building the config. Your project would have some existing policy that Terraform won't damage. When you're sure the Terraform config is right, flip `authoritative=true`. Sound good?

evandbrown:
@paddyforan Focusing on project ID here:

 Using `getProject` restores the behavior where the project ID is retrieved from the env (`GOOGLE_PROJECT`) or the provider configuration config (if it exists). This is that bug where I thought the `id` attribute of the config was working, but it was actually that project was specified in the provider config and was being used instead of `d.Get("id")`

So we can restore that behavior, but the create lifecycle still won't work with the template you showed. The reason for that is that the old version was using the create event to effectively import an existing project (this was before importing resources existed, and before the project API supported creation.) Now we're using Terraform's import functionality, and the create function actually really creates a project now.

So from a backwards compatibility perspective, this change should work fine with existing deployments/state. But you can't use that old template to create a new deployment since the meaning of create has changed (and been replaced with import).

Thoughts?

paddyforan:
Re: authoritative: I can see both sides of this, and it feels like a gray area to me. I'm trying to get more guidance on it from people who have more experience than I do, trying to avoid further confusion and goalpost moving. :) I'll update you here when I get more info, hopefully today.

Re: project ID:

> So from a backwards compatibility perspective, this change should work fine with existing deployments/state. But you can't use that old template to create a new deployment since the meaning of create has changed (and been replaced with import).

I think this matches what I'm hoping for. My goal is to let people that have configs that work for them right now (even if they only work because of a bug) continue to use those configs and not need to update everything as soon as this is merged. But also, we should probably tell those people that the configs don't actually work the way they think they do. Which is what the deprecation message does.

So reading environment variables on _create_ doesn't feel super important, because any config that already exists won't have the create run for it--it already has the project in state. But on _read_, it's important, because those configs would think they can't find the ID anymore, which breaks things. On update and destroy, I'd argue it's not important, because that should only happen when the user is changing things, and if they're changing things, they can change to use the project_id attribute anyways. I just want to avoid having `terraform apply` break immediately after you upgrade when your config hasn't changed at all, if that makes sense.

The way I'm testing this is by using that config file, using the Terraform 0.8.4 binary, and running `terraform apply`. Then I run `terraform apply` using a binary built from your branch. Ideally, what I'd see is that no change is necessary, because the config didn't change, but also a warning that `id` doesn't work. I get the warning, but at the moment, it fails with the error I pasted above.

Hopefully that clears things up! Sorry for the confusion. Let me know if you have questions, or if a quick Hangout would help.

evandbrown:
Cool! So looking at the `terraform apply` approach you're using, it seems like the state migration func for `resource_google_project` isn't executing. The second apply sees that `project_id` is missing, and because of `force_new` tries to delete then re-create the resource.

The migration func should have taken care of that. Any thoughts on why that's not running?

paddyforan:
I think, if I understand what I'm seeing correctly, it's saying that because `project_id` is not in the config file, it needs to be removed from the state, which prompts the force_new. We want it to be OK that project_id isn't set in the config, but we also want switching from id to project_id to not force a delete/create cycle.

The answer to the delete/create cycle is the state migration. The answer to the diff... I'm not sure. My gut says we can do a DiffSuppressFunc that ignores diffs where the project_id is set in state but not in the config, but I need to think through the ramifications of that. Is probably worth trying out, though?

evandbrown:
I believe the `DiffSupressFunc` in my latest push gets us there:

`terraform apply` this config with 0.8.4:

```
resource "google_project" "acctest" {
  id = "evandbrown17"
}
```

Result:

```
provider.google.region
  Enter a value: 

google_project.acctest: Creating...
  name:   "" => "<computed>"
  number: "" => "<computed>"
google_project.acctest: Creation complete

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

The state of your infrastructure has been saved to the path
below. This state is required to modify and destroy your
infrastructure, so keep it safe. To inspect the complete state
use the `terraform show` command.

State path: terraform.tfstate
```

---

`terraform apply` the same config with dev build.

Result:

```
terraform apply
[WARN] /home/evanbrown/dev/go-workspaces/terraform/bin/terraform-provider-google overrides an internal plugin for google-provider.
provider.google.region
  Enter a value: 

fix these before continuing.

Warnings:

  * google_project.acctest: "id": [DEPRECATED] Please replace this attribute with `project_id`. It will be removed from the schema in Terraform 0.9 and will be computed automatically. See https://www.terraform.io/docs/providers/google/r/google_project.html for a complete explanation.

No errors found. Continuing with 1 warning(s).

google_project.acctest: Refreshing state... (ID: evandbrown17)

```

Migration function is applied and state file contains new attributes, including `project_id`.

---

Add `project_id` to config:

```
resource "google_project" "acctest" {
  id = "evandbrown17"
  project_id = "evandbrown17"
}
```

Result: no diff.

---

Remove `id` from config:

```
resource "google_project" "acctest" {
  id = "evandbrown17"
  project_id = "evandbrown17"
}
```

Result: no diff

builtin/providers/google/resource_google_project.go
```diff
@@ -25,28 +23,63 @@ import (
...
 			State: schema.ImportStatePassthrough,
 		},
+		MigrateState: resourceGoogleProjectMigrateState,
 
 		Schema: map[string]*schema.Schema{
 			"id": &schema.Schema{
+				Type:       schema.TypeString,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "Please replace this attribute with `project_id`. It will be removed from the schema in Terraform 0.9 and will be computed automatically. See https://www.terraform.io/docs/providers/google/r/google_project.html for a complete explanation.",
```
paddyforan: This feels a bit ambiguous to me. Not to bikeshed, but I think something like

> The id field has unexpected behaviour and probably doesn't do what you expect. See https://www.terraform.io/docs/providers/google/r/google_project.html#id-field for more information. Please use project_id instead; future versions of Terraform will remove the id field.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,36 +88,57 @@ func resourceGoogleProject() *schema.Resource {
...
+	var err error
+	pid = d.Get("project_id").(string)
+	if pid == "" {
+		pid, err = getProject(d, config)
+		if err != nil {
+			return fmt.Errorf("Error getting project ID: %v", err)
+		}
+	}
+	if pid == "" {
+		return fmt.Errorf("One of 'project_id' or 'id' must be set in the config")
```
paddyforan: I'd remove `or 'id'` from this--we should try to discourage any new use of `id`.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,36 +88,57 @@ func resourceGoogleProject() *schema.Resource {
...
 
 	// Apply the IAM policy if it is set
 	if pString, ok := d.GetOk("policy_data"); ok {
 		// The policy string is just a marshaled cloudresourcemanager.Policy.
 		// Unmarshal it to a struct.
 		var policy cloudresourcemanager.Policy
-		if err = json.Unmarshal([]byte(pString.(string)), &policy); err != nil {
+		if err := json.Unmarshal([]byte(pString.(string)), &policy); err != nil {
 			return err
 		}
```
paddyforan: For debug purposes, it'd be great to log policy.Bindings here, just so we can track any potential weirdness in people's merging.

builtin/providers/google/resource_google_project.go
```diff
@@ -95,47 +149,89 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+	if p.Parent != nil {
+		d.Set("org_id", p.Parent.Id)
+	}
+
+	// Read the IAM policy
+	pol, err := getProjectIamPolicy(pid, config)
+	if err != nil {
+		return err
+	}
+	d.Set("policy_etag", pol.Etag)
```
paddyforan: Should we be setting `policy_data` here, as well?

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceGoogleProjectIamPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	pid, err := getProject(d, config)
```
paddyforan: This is new behaviour, right? There are no backwards compatibility concerns here? If so, I think we can just use `d.Get("project")`, right? Then we keep configs authoritative, and since the config requires a project to be set, there's no reason we'd need to fall back on anything, right?

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+		d.Set("restore_policy", "")
+	} else {
+		log.Printf("[DEBUG] Updating non-authoritative IAM policy for project %q", pid)
+		// Get the previous policy from state
+		pp, err := getPrevResourceIamPolicy(d)
+		if err != nil {
+			return fmt.Errorf("Error retrieving previous version of changed project IAM policy: %v", err)
+		}
+
```
paddyforan: Let's log the previous policy from state here for debugging.

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+		// Get the existing IAM policy from the API
+		ep, err := getProjectIamPolicy(pid, config)
+		if err != nil {
+			return fmt.Errorf("Error retrieving IAM policy from project API: %v", err)
+		}
+
```
paddyforan: Let's log the existing policy from the API here for debugging.

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+		// Subtract the previous and current policies from the policy retrieved from the API
+		rp := subtractIamPolicy(ep, pp)
+		rp = subtractIamPolicy(rp, p)
```
paddyforan: Let's log the results of each of these subtractions.

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+	}
+	d.SetId(pid)
+	return resourceGoogleProjectIamPolicyRead(d, meta)
+}
+
+func resourceGoogleProjectIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Reading google_project_iam_policy")
+	config := meta.(*Config)
+	pid, err := getProject(d, config)
```
paddyforan: Same here, `project` is required, let's use it.

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	log.Printf("[DEBUG]: Setting etag=%s", p.Etag)
+	d.Set("etag", p.Etag)
+	return nil
+}
+
+func resourceGoogleProjectIamPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG]: Updating google_project_iam_policy")
+	config := meta.(*Config)
+	pid, err := getProject(d, config)
```
paddyforan: Same here, `project` is required, let's use it.

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,343 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Get the policy in the template
+	p, err := getResourceIamPolicy(d)
+	if err != nil {
+		return fmt.Errorf("Could not get valid 'policy_data' from resource: %v", err)
+	}
```
paddyforan: Probably excessive, but let's log the policy just for debuggability.

website/source/docs/providers/google/r/google_project.html.markdown
```diff
@@ -8,54 +8,69 @@ description: |-
...
-data "google_iam_policy" "admin" {
-  binding {
-    role = "roles/storage.objectViewer"
-    members = [
-      "user:evandbrown@gmail.com",
-    ]
-  }
+    project_id = "your-project-id"
+    org_id = "1234567"
+    services = ["compute_component", "storage-component-json.googleapis.com", "iam.googleapis.com"
```
paddyforan: Missing an `]` at the end here, I think.

website/source/docs/providers/google/r/google_project.html.markdown
```diff
@@ -8,54 +8,69 @@ description: |-
...
+
+* `services` - (Optional) The services/APIs that are enabled for this project.
+    For a list of available services, run `gcloud beta service-management list`
+
+* `skip_delete` - (Optional) If true, the Terraform resource can be deleted
+    without deleting the Project via the Google API.
+
+* `policy_data` - (Deprecated) This argument is no longer supported, and
+    an error will be thrown if it is present. It must be replaced with a 
+    `google_project_iam_policy` resource.
```
paddyforan: I think we should probably soften the language around this, because it makes it sound like it no longer works even though it used to, which isn't (AFAIK) actually the case. It still works, but we suggest moving off it, and people will get a warning to move off it.

website/source/docs/providers/google/r/google_project.html.markdown
```diff
@@ -8,54 +8,69 @@ description: |-
...
 ## Attributes Reference
 
 In addition to the arguments listed above, the following computed attributes are
 exported:
 
-* `name` - The name of the project.
-
 * `number` - The numeric identifier of the project.
+* `policy_etag` - (Deprecated) The etag of the project's IAM policy, used to
+    determine if the IAM policy has changed.
```
paddyforan: We should mention that people should use the `google_project_iam_policy` resource instead.

website/source/docs/providers/google/r/google_project_iam_policy.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project` - (Required) The project ID.
+    Changing this forces a new project to be created.
+
+* `policy_data` - (Optional) The `google_iam_policy` data source that represents
```
paddyforan: This is actually required, according to the Schema.

website/source/docs/providers/google/r/google_project_iam_policy.html.markdown
```diff
@@ -0,0 +1,66 @@
...
+    merged with any existing policy applied to the project.
+
+    Changing this updates the policy.
+
+    Deleting this removes the policy, but leaves the original project policy
+    intact. If there are overlapping `binding` entries between the original
+    project policy and the data source policy, they will be removed.
+
+* `authoritative` - (Optional) A boolean value indicating if this policy
+    should overwrite any existing IAM policy on the project. If this argument
```
paddyforan: Let's mention here that this includes removing any IAM policies not specified in your config, and that that could mean you get locked out of your project.

website/source/docs/providers/google/r/google_project_services.html.markdown
```diff
@@ -0,0 +1,32 @@
...
+Allows management of enabled API services for an existing Google Cloud
+Platform project. Services in an existing project that are not defined
+in the config will be removed.
+
+## Example Usage
+
+```js
+resource "google_project_services" "project" {
+    project_id = "your-project-id"
+	services = ["iam.googleapis.com", "cloudresourcemanager.googleapis.com"]
```
paddyforan: Weird indentation.

evandbrown:
Haha I can't approve the changes either as I'm the PR author :laughing: 

But here's my old-school approval: LGTM

danawillow:
I can look at it, I'll do that now.

paddyforan:
I'd be satisfied with @evandbrown checking my changes and verifying they make sense, and I've checked his changes and verified they make sense, so at least two pairs of eyes hit everything.

That said, having @danawillow check things is excellent / 💯 / even better, so if you have the bandwidth and don't mind doing it, that'd be fantastic, and we could then be super confident merging.

Merging still needs to block until we get the quota thing worked out, but it sounds like we're super close on that.

builtin/providers/google/resource_google_project.go
```diff
@@ -55,36 +89,66 @@ func resourceGoogleProject() *schema.Resource {
...
+	project := &cloudresourcemanager.Project{
+		ProjectId: pid,
+		Name:      d.Get("name").(string),
+		Parent: &cloudresourcemanager.ResourceId{
+			Id:   d.Get("org_id").(string),
+			Type: "organization",
+		},
+	}
+
+	op, err := config.clientResourceManager.Projects.Create(project).Do()
```
danawillow: does this actually build? (since err was already declared)

builtin/providers/google/resource_google_project.go
```diff
@@ -95,45 +140,69 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
+
 	return nil
 }
 
 func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
-	project, err := getProject(d, config)
+	pid := d.Id()
+
+	// Read the project
```
danawillow: Can you make this a comment in the code in case others have the same question?

builtin/providers/google/resource_google_project_iam_policy.go
```diff
@@ -0,0 +1,347 @@
...
+
+		// First, subtract the policy defined in the template from the
+		// current policy in the project, and save the result. This will
+		// allow us to restore the original policy at some point (which
+		// assumes that Terraform owns any common policy that exists in
+		// the template and project at create time.
+		rp := subtractIamPolicy(ep, p)
+		rps, err := json.Marshal(rp)
+		if err != nil {
+			return fmt.Errorf("Error marhsaling restorable IAM policy: %v", err)
```
danawillow: typo: marshaling

builtin/providers/google/resource_google_project_services_test.go
```diff
@@ -0,0 +1,178 @@
...
+				),
+			},
+		},
+	})
+}
+
+func testAccGoogleProjectAssociateServicesBasic(services []string, pid, name, org string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+    project_id = "%s"
```
danawillow: fyi the reason the indentation looks messed up here is because it's a mix of tabs and spaces. up to you if you want to fix it.

builtin/providers/google/resource_storage_object_acl.go
danawillow: I'm not sure changing this file is necessary anymore- we've definitely revendored since November

builtin/providers/google/import_google_project_test.go
```diff
@@ -23,7 +25,19 @@ func TestAccGoogleProject_importBasic(t *testing.T) {
 				ResourceName:      resourceName,
 				ImportState:       true,
 				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"skip_delete",
```
paddyforan: I can't actually explain this one. @evandbrown said it was necessary to make tests pass, though I just ran the test now, and it passed without the ImportStateVerifyIgnore. I've removed it, and we can fix the test later if the bug pops up again.

builtin/providers/google/resource_google_project.go
paddyforan: Yup! You only get errors about variables already being declared if they're the only variable being declared: https://play.golang.org/p/b0qIqFpVrv vs https://play.golang.org/p/L_G5UBLL4d

paddyforan:
128 comments, 21 emails, and a Hangout later, I think this one is _finally_ ready. :D My tests all pass, and manual testing has shown that UX is as expected, as far as I can tell. If either of y'all can verify that your LGTMs stand in light of the recent changes, we can merge this when ready. :D

builtin/providers/google/resource_google_project.go
```diff
@@ -71,10 +71,11 @@ func resourceGoogleProject() *schema.Resource {
...
 			},
 			"policy_data": &schema.Schema{
-				Type:       schema.TypeString,
-				Optional:   true,
-				Computed:   true,
-				Deprecated: "Use the the 'google_project_iam_policy' resource to define policies for a Google Project",
+				Type:             schema.TypeString,
+				Optional:         true,
+				Computed:         true,
+				Deprecated:       "Use the the 'google_project_iam_policy' resource to define policies for a Google Project",
```
danawillow: typo: the the

danawillow:
New changes LGTM

evandbrown:
LGTM OMG WOOOO!

PR #10438, opened 2016-11-30
Add `aws_s3_bucket` to importable resources

mwarkentin:
According to the [`aws_s3_bucket` docs](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html) it is importable.

### DIFF
--- a/website/source/docs/import/importability.html.md
+++ b/website/source/docs/import/importability.html.md
@@ -82,6 +82,7 @@ To make a resource importable, please see the
 * aws_route53_health_check
 * aws_route53_zone
 * aws_route_table
+* aws_s3_bucket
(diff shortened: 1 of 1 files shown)


website/source/docs/import/importability.html.md
```diff
@@ -82,6 +82,7 @@ To make a resource importable, please see the
 * aws_route53_health_check
 * aws_route53_zone
 * aws_route_table
+* aws_s3_bucket
```
mwarkentin: Not sure on if this should be at the start of the `aws_s*` list or at the end.

stack72:
Thanks for the addition here @mwarkentin :)

This LGTM!

P.

PR #10444, opened 2016-11-30
Increase `aws_emr_cluster` timeout

fosskers:
### TODO
- [x] Boost timeout to 75 minutes

### Motivation
As has been done to other resources when the need arose, I've boosted the EMR resource timeout. I found that my EMR bootstrap scripts run longer than the default 40 minutes, due to custom compilation of some large libraries. 

Being able to control this timeout in external config would be nice, but for now these values seem hard-coded, hence this PR.


stack72:
Hi @fosskers 

Paul

Issue #10002, opened 2016-11-09
Make `aws_iam_access_key` verify existence only


ajlanghorn:
The `aws_iam_access_key` resource is used to generate an IAM access key for a given user, which I suppose makes some sense given it allows for the creation of keys in AWS in the same way as any other resource.

When a `terraform apply` is run, the generated IAM access key and secret key are written to the Terraform state file, which - if you don't read the small warning in the documentation properly - might come as a surprise.

AWS only let you view the IAM secret key once - when you first create the access and secret key pair. But by inserting it in to the state file, it's then visible multiple times and can be checked in to Git (this problem also occurs for secrets kept for other resources, such as RDS instances).

I wonder whether or not the `aws_iam_access_key` resource should be re-purposed to simply check for the existence of an access/secret key pair, rather than try to track the state of the pair and therefore its values?

Thoughts welcome, since this goes a little against the grain of AWS only giving us secret keys once and does have the possibility of some fairly serious security implications in certain scenarios.

jen20:
My opinion here is that we should follow the model of `aws_iam_login_profile`, and allow specification of a PGP key or key base login with which to encrypt the access keys, so they may be securely communicated without being exposed to the operator. This would be a breaking change, but we could consider it for Terraform 0.8.

Any thoughts @phinze, @mitchellh?

ajlanghorn:
@jen20 I think that makes sense; if there's a standard to be followed from another resource, I'd support its use. That also keeps similar functionality as at the moment, too. Nice idea! 👍

phinze:
Agreed - let's follow the PGP/keybase pattern and queue this up for 0.8

catsby:
Hey there – we implemented optional encrypting of the `secret` in https://github.com/hashicorp/terraform/pull/10615 . Thanks!

ajlanghorn:
@catsby <3

PR #10450, opened 2016-11-30
[WIP] provider/aws: Encrypt aws_iam_access_key.secret with pgp

catsby:
#### This is meant for v0.8.x and not to be released in v0.7.x series

------

Instead of storing the one-time revealed secret in plain text, encrypt it with a PGP key.

I've marked this as `[WIP]`, for discussion.

I haven't found a smooth upgrade path beyond completely breaking this resource. I considered making `pgp_key` optional, and letting people limp along with the `secret` exposed in the state file until they decided to add a `pgp_key` to encrypt it. Normally we'd mark `secret` as `Deprecated` to get a nice warning message for users to migrate off of it, however `secret` is a computed field and `Deprecated` does not work there. 

Another consideration was to make `pgp_key` optional, but fail in the `CREATE` step if it wasn't found, that way **new** users or **new** `aws_iam_access_key` resources did require the pgp encryption route, however that results in a **runtime** error / warning instead of at **plan** time, and I felt that a lesser user experience. 

#### Upgrade UX:

- user on pre-v0.8 is using resource as normal, with `secret` stored in state ( :scream: )
- upgrade to v0.8+, and get `plan` failure/warning message about missing required `pgp_key`
- `pgp_key` is a `ForceNew` attribute, so adding one will show a diff to re-create the resource ( 🐼 )
- `apply` will recreate the resource. `secret` is no longer set, so no longer in state. Users can decrypt the password as show in the documentation. 

#### Changes:

- `secret` is dropped from the resource
- `pgp_key` added, `Required`
- `key_fingerprint` added, `Computed`
- `encrypted_secret` added, `Computed`
 
The encryption behavior is lifted from `aws_iam_login_profile`. 

Refs #10002

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_access_key.go
+++ b/builtin/providers/aws/resource_aws_iam_access_key.go
@@ -26,16 +26,28 @@ func resourceAwsIamAccessKey() *schema.Resource {
 				ForceNew: true,
 			},
 			"status": &schema.Schema{
-				Type: schema.TypeString,
-				// this could be settable, but goamz does not support the
-				// UpdateAccessKey API yet.
+                                Type:     schema.TypeString,
 				Computed: true,
 			},
 			"secret": &schema.Schema{
+                                Type:       schema.TypeString,
+                                Computed:   true,
+                                Deprecated: "Secrets are no longer stored. Please use a PGP key to encrypt",
+                        },
+                        "ses_smtp_password": &schema.Schema{
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"ses_smtp_password": &schema.Schema{
+                        "pgp_key": {
+                                Type:     schema.TypeString,
+                                ForceNew: true,
+                                Required: true,
+                        },
+                        "key_fingerprint": {
+                                Type:     schema.TypeString,
+                                Computed: true,
+                        },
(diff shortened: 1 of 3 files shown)


mitchellh:
I'm okay with this BC UX, I think it may be worth adding to the upgrade guide.

One question: for the upgrade case, what happens to the destroyed access key? Is it revoked (as in no longer works?) If we can make it so that pre-0.8 users don't have to worry about this then its pretty clean but unsure if that exists.

catsby:
> what happens to the destroyed access key? Is it revoked (as in no longer works?)

Yes, it's revoked and destroyed

> If we can make it so that pre-0.8 users don't have to worry about this then its pretty clean but unsure if that exists.

Not sure I follow. I intend to merge this into v0.8 branch and **not** v0.7.x series, so the BC is only when a user upgrades to v0.8. In a multi-user scenario and one user upgrades to v0.8 then the key will get rotated, but I think that's a bad scenario anyway

mitchellh:
Its not a requirement, but it'd be best case if a user who upgrades just doesn't have to use the new behavior. But its all good! 

I think this is a fair tradeoff.

catsby:
closing this, will re-open with refactored work to reflect the change we decided on

PR #10470, opened 2016-12-01
provider/aws: Support eu-west-2

jen20:
This is the new London region - we don't have access yet but several enquiries have come from customers who do.

cc @garylawuk


radeksimko:
Just FYI - those folks will be unable to use S3 as website and 2 other data sources, unless we add more details about this new region, see https://github.com/hashicorp/terraform/commit/4b36bc3210433618be76df2202f22080a25f41a0

mitchellh:
Since we don't have access to it we can't add that yet, thanks @radeksimko.

I think its important enough to get this in and then add those as we learn more.

stack72:
@radeksimko / @mitchellh I have already spoken to @jen20 about this :) he's trying to get the info now

afoldesi-gds:
Hi all, do you have any news on this one? Would it be possible to decouple this particular change from whatever it is waiting on & get it merged?

stack72:
The only metadata missing from this list if the AWS Redshift Service Account ID - this doesn't seem to exist yet http://docs.aws.amazon.com/redshift/latest/mgmt/db-auditing.html#db-auditing-enable-logging

Will add it when it's available

afoldesi-gds:
Brilliant, thanks!

PR #10472, opened 2016-12-01
Have `terraform graph` use the plan builder

jbardin:
The graph displayed based solely on config may be different than the
graph built using the new PlanGraphBuilder. Have `terraform graph` use
the same builder that will be used by `terraform plan`.


jbardin:
oops, lost the verbose option, just going to push a fix for that

jbardin:
This is going to miss a lot of nodes from the new graph. REDO!

PR #10473, opened 2016-12-01
provider/aws: Add Lightsail Instance

catsby:
Support for [AWS Lightsail](https://lightsail.aws.amazon.com/ls/docs/all)

Includes:

- [x] Docs for `aws_lightsail_instance`
- [x] Tests for `aws_lightsail_instance`
- [x] `aws_lightsail_instance` resource


catsby:
We're limited on `key_pair` right now, we'll need a separate resource for that

builtin/providers/aws/resource_aws_lightsail_instance.go
```diff
@@ -0,0 +1,264 @@
...
+		BlueprintId:      aws.String(d.Get("blueprint_id").(string)),
+		BundleId:         aws.String(d.Get("bundle_id").(string)),
+		InstanceNames:    aws.StringSlice([]string{iName}),
+	}
+
+	if v, ok := d.GetOk("key_pair_name"); ok {
+		req.KeyPairName = aws.String(v.(string))
+	}
+	// // encoding required?
+	// if v, ok := d.GetOk("user_data"); ok {
```
stack72: This intentionally missing? :)

builtin/providers/aws/resource_aws_lightsail_instance.go
```diff
@@ -0,0 +1,264 @@
...
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		// We don't return an error here because the Create call succeded
+		log.Printf("[ERR] Error waiting for instance (%s) to become ready: %s", d.Id(), err)
+	}
+
+	// return resourceAwsLightsailInstanceUpdate(d, meta)
```
stack72: worth removing this old comment?

catsby: whoops :)

catsby:
@stack72 thanks for the review, I've updated it!

> 1 other thing, is it worth adding a few more tests? This would allow us to check for the _disappears

I added `TestAccAWSLightsailInstance_disapear` for this 

> import

The `_basic` test has the `IDRefreshName` thing, any more than that?

> maybe an update

Sadly everything is `ForceNew` at the moment, the API doesn't offer much for updates at all. The resource lacks an `Update` method

stack72:
LGTM! thanks for the changes :)

PR #10476, opened 2016-12-01
provider/github: Implements github_branch_protection

alindeman:
This PR introduces a new resource, `github_branch_protection`, that configures branch protection for a given repository and branch.

The API required in go-github landed in google/go-github#476 and are included in a single commit a2da126 in this branch.

The Protected Branches API is still in a preview period, meaning it may change (including breaking changes). I'm open to keeping this PR open until it stabilizes, but I could also see it being useful to folks now (with a warning that it is not fully stable). I'm open to hearing feedback on that point.

/cc: @sr

### DIFF
--- a/builtin/providers/github/resource_github_branch_protection.go
+++ b/builtin/providers/github/resource_github_branch_protection.go
@@ -1,0 +1,278 @@
+package github
+
+import (
+	"context"
+	"errors"
+
+	"github.com/google/go-github/github"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceGithubBranchProtection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubBranchProtectionCreate,
+		Read:   resourceGithubBranchProtectionRead,
+		Update: resourceGithubBranchProtectionUpdate,
+		Delete: resourceGithubBranchProtectionDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


jason-riddle:
Any update to this?

alindeman:
@stack72 @jason-riddle I've rebased this on 0.9.1 and added support for `required_pull_request_reviews`. This is ready for some feedback/review if anyone has a chance :heart:

alindeman:
I resolved new conflicts and rebased on 0.9.3. Any feedback @stack72 @mbfrahry? :)

stack72:
Hi @alindeman 

Thanks for the work here - can you tell me if this is still in preview?

P.

builtin/providers/github/resource_github_branch_protection.go
```diff
@@ -0,0 +1,278 @@
...
+	return resourceGithubBranchProtectionRead(d, meta)
+}
+
+func resourceGithubBranchProtectionRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Organization).client
+	r, b := parseTwoPartID(d.Id())
+
+	githubProtection, _, err := client.Repositories.GetBranchProtection(context.TODO(), meta.(*Organization).name, r, b)
+	if err != nil {
+		d.SetId("")
```
stack72: Do we want to refresh from state on *every* error ?

alindeman: @stack72 Thanks for pointing this out. I think it would be frustrating to have the item cleared from state on, say, a network issue or timeout. I've proposed a more narrow version over in #13887

PR #10502, opened 2016-12-02
Make sure that a Context.diff is never nil

jbardin:
The context and diff are passed along during a walk, and the diff is assumed
to be valid.

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -162,13 +162,18 @@ func NewContext(opts *ContextOpts) (*Context, error) {
 		}
 	}
 
+	diff := opts.Diff
+	if diff == nil {
+		diff = &Diff{}
+	}
+
 	return &Context{
 		components: &basicComponentFactory{
 			providers:    opts.Providers,
 			provisioners: opts.Provisioners,
 		},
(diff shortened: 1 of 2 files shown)


mitchellh:
Is this potentially another scenario where we should assume that the diff MIGHT be nil? Where was the crash originally out of curiosity?

jbardin:
The crash happens in EvalWriteDiff, where it gets a diff and mutex from the context. The context here returned a nil diff with a valid mutex. I think we need to ensure there's a valid diff, since it's trying to modify it at this point.

mitchellh:
Yes, but we do ensure that plan graphs have a diff in `Plan()`. I'm just curious how this was happening otherwise?

jbardin:
Ah, true, I missed that. This was purely in my own tests that happened upon this. 

Maybe this isn't necessary at all then, though maybe it could be useful to ensure that diff is never nil?

mitchellh:
Yeah, it could be totally fine, my fear is just that it might mask other issues but I have no idea. Its safe enough to merge this probably. LGTM!

PR #10508, opened 2016-12-02
[docs] Correct typo in azure storage account docs

sheerun:


stack72:
Hi @sheerun 

Thanks for the work here :) Nice spot!

paul

PR #10509, opened 2016-12-02
provider/openstack: Detect Region for Importing Resources

### LINKED ISSUE #9054: How to import resource on a openstack provider with multiple region ?
### Terraform Version

0.7.4
### Affected Resource(s)
- any importable openstack resources

No configuration file
### Debug Output

https://gist.github.com/xtero/d048e6b0235a13ea8622eb197781883e
### Panic Output

N.A
### Expected Behavior

In my environment, I have set the varibale OS_REGION_NAME to "fr1".
So during the import, I expect terraform to filter endpoint based on this value.
### Actual Behavior

Nothing is filtered. And I end up with 2 endpoint instead of one
### Steps to Reproduce
1. We need to have at least 2 region with endpoint for each region for the same component.
2. We have to defined tenant variables :

```
OS_REGION_NAME=fr1
OS_PASSWORD=123
OS_AUTH_URL=https://identity.fr1.cloudwatt.com/v2.0/
OS_USERNAME=toto
OS_TENANT_NAME=My_tenant
```
1. The problem occure when we run something like :

```
$ terraform import openstack_networking_network_v2.any_name any-id-goes-here
```
1. We got the following error message : "Error creating OpenStack networking client: Discovered 2 matching endpoints"
### Important Factoids

We are using a public cloud based on Openstack that provide 2 regions "fr1" and "fr2".

jtopjian:
This commit changes the way the OpenStack region is detected and set.
Any time a region is required, the region attribute will first be
checked. Next, the OS_REGION_NAME environment variable will be checked.
While schema.EnvDefaultFunc handles this same situation, it is not
applicable when importing resources.

Fixes #9054

### DIFF
--- a/builtin/providers/openstack/import_openstack_blockstorage_volume_attach_v2_test.go
+++ b/builtin/providers/openstack/import_openstack_blockstorage_volume_attach_v2_test.go
@@ -19,10 +19,9 @@ func TestAccBlockStorageVolumeAttachV2_importBasic(t *testing.T) {
 			},
 
 			resource.TestStep{
-				ResourceName:            resourceName,
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"region"},
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
 			},
 		},
 	})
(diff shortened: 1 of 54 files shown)


jtopjian:
Thank you for the review and merge!

PR #10517, opened 2016-12-03
provider/openstack added back personality which was missing from the createmap request body

brucemartins:
when terraform migrated over from the rackspace/gophercloud project to the gophercloud/gophercloud project the personality capability was lost as it seems the block of code in the request createmap was missing.. This is required for booting with config_drive

### DIFF
--- a/vendor/github.com/gophercloud/gophercloud/openstack/compute/v2/servers/requests.go
+++ b/vendor/github.com/gophercloud/gophercloud/openstack/compute/v2/servers/requests.go
@@ -260,6 +260,10 @@ func (opts CreateOpts) ToServerCreateMap() (map[string]interface{}, error) {
 		b["flavorRef"] = flavorID
 	}
 
+	if len(opts.Personality) > 0 {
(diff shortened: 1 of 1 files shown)


jtopjian:
@brucemartins Nice catch!

Unfortunately this is a vendor file, so the patch will have to be submitted to github.com/gophercloud/gophercloud. Once the PR is merged, I can pull the updated file into Terraform.

The good news is that I also hang out over at Gophercloud, so although I can't merge the request for you, I can vouch for its validity.

Also, the better way to solve this is to edit the JSON Go tag:

```go
- Personality Personality `json:"-"`
+ Personality Personality `json:"personality,omitempty"`
```

`ToServerCreateMap` is only for attributes that have custom logic applied to them. The attributes that can be sent to OpenStack as-is are handled by the JSON tags.

I'm going to close this PR, though please let me know if you have any questions. Thank you again for catching this.

brucemartins:
thank you for pointing that out, I will make the change, test and submit a PR to the other repo

Issue #10459, opened 2016-12-01
terraform should gracefully quit on SIGTERM


mildred:
Terraform Version: 0.7.5-132-g7e04082

Looking at code, it seems that terraform gracefully shuts down on `SIGINT` (`os.Terminate` in go) but does not respond to `SIGTERM` (`syscall.SIGTERM`). The first signal is used when interrupting from command line (Ctrl-C0 while the latter is used when using the `kill` command.

I am using terraform unattended in a system service, and systemd generally uses `SIGTERM` to notify a service to stop. It can be reconfigured to use `SIGINT` but it would be nice if terraform could make use of `SIGTERM`.

mitchellh:
I agree.

For impl notes (maybe me): we'll need to use conditional build files here since syscall.SIGTERM isn't available on Windows, I believe.

mildred:
Probably, panicwrap will need to be changed as well

https://github.com/mitchellh/panicwrap/blob/master/panicwrap.go#L170

mildred:
I have the master branch here modified to make it work. Still WIP https://github.com/squarescale/terraform

mitchellh:
Done!

mildred:
@mitchellh you merged part of the solution only. there is still a problem with panicwrap. it ignores the SIGTERM signal instead of forwarding it to the main process. Contrary to SIGINT, SIGTERM is generally sent only to the parent process, so ignoring it in panicwrap is not a guarantee that it will reach the main process.

On some circumstances (using killall or systemd where it kill the complete cgroup, this is configurable) this will work. on some other setups (running inside docker for instance) this will fail.

That's why I think that panicwrap should be modified to allow forwarding some signals.

Also, I was just thinking that forwarding the SIGTERM signal could cause it to be received twice on the main process, forcing the shutdown instead of keeping graceful. In that circumstance, I think SIGTERM should not trigger a force shutdown when received more than once. here are the two use cases:

1. where the signal is sent to the parent process only (docker, ...):

* terraform starts
* SIGTERM is sent to the parent process (panicwrap)
* panicwrap forwards the SIGTERM signal to the main process
* the mainprocess received SIGTERM only once, it gracefully shuts down
* to force shutdown, SIGINT could be used on the main process

2. where the signals is sent to all processes of the control group, or through killall:

* terraform starts
* SIGTERM is sent to all processes, including the panicwrap process and the main process
* panicwrap forwards the SIGTERM signal to the main process
[...]

PR #10534, opened 2016-12-05
Gracefully stops on SIGTERM

mildred:
Add SIGTERM handling on UNIX platforms.

closes issue #10459

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -223,7 +223,7 @@ func makeShutdownCh() <-chan struct{} {
 	resultCh := make(chan struct{})
 
 	signalCh := make(chan os.Signal, 4)
-	signal.Notify(signalCh, os.Interrupt)
+	signal.Notify(signalCh, interruptSignals...)
 	go func() {
 		for {
 			<-signalCh
--- a/signal_unix.go
+++ b/signal_unix.go
@@ -1,0 +1,10 @@
+// +build !windows
+
+package main
+
+import (
+	"os"
+	"syscall"
+)
+
+var interruptSignals []os.Signal = []os.Signal{os.Interrupt, syscall.SIGTERM}
--- a/signal_windows.go
+++ b/signal_windows.go
@@ -1,0 +1,9 @@
+// +build windows
+
+package main
+
+import (
+	"os"
+)
+
+var interruptSignals []os.Signal = []os.Signal{os.Interrupt}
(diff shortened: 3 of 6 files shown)


mildred:
I had another patch, but I tested it and it seems it works without. How do signals interact with plugins ?

Here it is:

```
diff --git a/vendor/github.com/hashicorp/go-plugin/server.go b/vendor/github.com/hashicorp/go-plugin/server.go
index 4e3a552..02519a2 100644
--- a/vendor/github.com/hashicorp/go-plugin/server.go
+++ b/vendor/github.com/hashicorp/go-plugin/server.go
@@ -11,6 +11,7 @@ import (
        "runtime"
        "strconv"
        "sync/atomic"
+       "syscall"
 )
 
 // CoreProtocolVersion is the ProtocolVersion of the plugin system itself.
@@ -117,7 +118,7 @@ func Serve(opts *ServeConfig) {
 
        // Eat the interrupts
        ch := make(chan os.Signal, 1)
-       signal.Notify(ch, os.Interrupt)
+       signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
        go func() {
                var count int32 = 0
                for {
```

mildred:
This PR doesn't work yet, unless `TF_FORK=0`. I will provide a fix soon.

mildred:
Modifying panicwrap was necessary to allow signal forwarding. See PR mitchellh/panicwrap#15. This PR also contains a commit to update the panicwrap dependency.

mitchellh:
Merged to master. I made the updates necessary for the panicwrap passthrough config to be setup. Thanks!

mildred:
Hello,

I'm afraid this won't work. Panicwrap is blocking SIGTERM and does not forward the signal to the main terraform process. See mitchellh/panicwrap#15:

> SIGINT is generally sent to all processes connected to the virtual terminal when typing Ctrl-C. SIGTERM is generally sent to a single process. In that case, it is wise not to ignore that signal but to forward it to the child process instead to give it the opportunity to handle it.
>
> This is required for correct handling of SIGTERM inside terraform. When SIGTERM is sent to the parent process, ignoring the signal won't help, it must be forwarded.

mitchellh:
Ah I see, chatting over at panicwrap I didn't realize that SIGTERM wasn't sent to all processes like SIGINT. I'll check that out.

Issue #10227, opened 2016-11-18
provider/google: Resolve the conversation around instance templates and startup scripts


paddyforan:
Pinging @JDiPierro, @cblecker, @mikemcrill, and @evandbrown, as they've all shown interest in this issue.

Essentially, the shortcoming is that instances support a `metadata_startup_script` property, but instance templates do not. This makes it less intuitive than it should be to turn an instance into an instance template.

We've got three PRs and an open issue on this already, so rather than having this conversation in all of them, I wanted to centralise stuff in a single issue.

After researching the history of this, it turns out that:

* `metadata_startup_script` was born because we wanted instances to be recreated when the `metadata.startup-script` property changed, but because it's embedded in `metadata`, and we don't want instances to recreate when `metadata` changes, `metadata_startup_script` is necessary to force that.
* `metadata.startup-script` was deprecated because having two places to set the variable was causing issues with our reads, so now an error is thrown if it is used.
* Now that we have instance templates, we want to be able to use startup scripts on instance templates.
* Instance templates are immutable, so updating their startup script involves creating a new instance template and destroying the old one.
* Instance templates cannot be destroyed while they're in use by instance group managers.
* Destroying an instance group manager destroys all the instances it's managing.
* It's desirable that an instance should be recreated when its startup script changes, but it's not desirable that an instance group manager should be destroyed when its startup script changes.

So, after a lot of discussion yesterday, here's the proposed path forward:

* I'll open a PR that adds the `metadata_startup_script` to instance templates.
* Our documentation [already shows](https://www.terraform.io/docs/providers/google/r/compute_instance_template.html#using-with-instance-group-manager) that instance templates need `lifecycle { create_before_destroy = true }` to be used with instance group managers, which resolves the problem of destroying templates. The workflow is now: the new template is created, the instance group manager is updated to use that template, the old template is destroyed. This is all accomplished with `create_before_destroy`.
* We'll deprecate the `name` attribute on instance templates, preferring the `name_prefix`, as `create_before_destroy` will have conflict issues if trying to use the same name (we can't have two templates with the same name at the same time). Because instance templates are used primarily for managed instance groups (according to the docs), the desire to use a static name for instance templates seems like an edge case.

The upshot of all this:

* Creating managed instance groups will continue to work as documented, now with the added benefit of steering users away from giving their templates static names.
* Updating a template attached to a managed instance group will work as expected: new instances will be created with that template, instances that already exist will not be recreated.
[...]

PR #10537, opened 2016-12-05
provider/google: Fix instance/template metadata support

paddyforan:
Update our instance template to include metadata_startup_script, to
match our instance resource. Also, we've resolved the diff errors around
metadata.startup-script, and people want to use that to create startup
scripts that don't force a restart when they're changed, so let's stop
disallowing it.

Also, we had a bunch of calls to `schema.ResourceData.Set` that ignored
the errors, so I added error handling for those calls. It's mostly
bundled with this code because I couldn't be sure whether it was the
root of bugs or not, so I took care of it while addressing the startup
script issue.

**Notes:**
* When _importing_ instance templates, we assume any startup scripts are for `metadata.startup-script`. This means we can't do an import test for `metadata_startup_script` because it will check for `metadata_startup_script` in the state, and it will get `metadata.startup-script` instead. We have to make this assumption because until we have state to work off, we can't detect that `metadata_startup_script` was used. Imports (to my knowledge?) don't have any state, so we can't figure out what the user specified. This does mean when a template is imported, then its conf file is created, if the conf file uses `metadata_startup_script`, the next `terraform plan` will show changes to use `metadata_startup_script` in the state, instead.
* Changing from `metadata.startup-script` to `metadata_startup_script` requires a destroy/create. I don't know of a good way around this, or if this should even be the desired behaviour.
* In the event that both `metadata.startup-script` _and_ `metadata_startup_script` are specified, `metadata_startup_script` wins.
* `metadata_startup_script` is for instances that should be destroyed/recreated when the startup script changes. `metadata.startup-script` is for instances that want to modify it without a destroy/recreate. Instance templates destroy and recreate on metadata changes, anyways, so there's no real reason to use `metadata_startup_script` in an instance template, but cutting down on the inconsistencies between resources seems like a nice thing to do.

This undoes the work @phinze did in #3507, so pinging him. See also #10227, where @JDiPierro, @cblecker, and @sparkprime weighed in on the importance of having both supported for instances.

### DIFF
--- a/website/source/docs/providers/google/r/compute_instance_template.html.markdown
+++ b/website/source/docs/providers/google/r/compute_instance_template.html.markdown
@@ -136,6 +136,11 @@ The following arguments are supported:
 * `metadata` - (Optional) Metadata key/value pairs to make available from
     within instances created from this template.
 
+* `metadata_startup_script` - (Optional) An alternative to using the
+    startup-script metadata key, mostly to match the compute_instance resource.
+    This replaces the startup-script metadata key on the created instance and 
+    thus the two mechanisms are not allowed to be used simultaneously.
+
 * `network_interface` - (Required) Networks to attach to instances created from
     this template. This can be specified multiple times for multiple networks.
     Structure is documented below.
--- a/builtin/providers/google/resource_compute_instance.go
+++ b/builtin/providers/google/resource_compute_instance.go
@@ -111,10 +111,9 @@ func resourceComputeInstance() *schema.Resource {
 			},
 
 			"metadata": &schema.Schema{
-				Type:         schema.TypeMap,
-				Optional:     true,
-				Elem:         schema.TypeString,
-				ValidateFunc: validateInstanceMetadata,
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem:     schema.TypeString,
 			},
 
 			"metadata_startup_script": &schema.Schema{
@@ -634,10 +633,10 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	md := instance.Metadata
 
 	_md := MetadataFormatSchema(d.Get("metadata").(map[string]interface{}), md)
-	delete(_md, "startup-script")
 
 	if script, scriptExists := d.GetOk("metadata_startup_script"); scriptExists {
 		d.Set("metadata_startup_script", script)
+		delete(_md, "startup-script")
 	}
 
 	if err = d.Set("metadata", _md); err != nil {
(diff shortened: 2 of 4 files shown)


website/source/docs/providers/google/r/compute_instance_template.html.markdown
```diff
@@ -136,6 +136,11 @@ The following arguments are supported:
 * `metadata` - (Optional) Metadata key/value pairs to make available from
     within instances created from this template.
 
+* `metadata_startup_script` - (Optional) An alternative to using the
+    startup-script metadata key, mostly to match the compute_instance resource.
+    This replaces the startup-script metadata key on the created instance and 
+    thus the two mechanisms are not allowed to be used simultaneously.
```
phinze: We have a `ConflictsWith` key in `helper/schema` to enforce this as a validation - I think we should do that on both resources to prevent folks from shooting themselves in the foot. 👣 🔫

phinze: > It would make `metadata_startup_script` incompatible with setting any metadata at all

Ah I thought that was true. Given:

```
  metadata {
    foo = "bar"
  }
  metadata_startup_script = "somescript"
```

Won't you get a perpetual diff in the metadata map as the two fields fight over whether or not `startup-script` should be in there?

paddyforan: Using this instance template conf:

```tf
resource "google_compute_instance_template" "template" {
  name        = "terraform-test"
  description = "test instance template"

  tags = ["testing"]

  machine_type = "f1-micro"

  disk {
    source_image = "debian-cloud/debian-8"
    auto_delete  = true
    boot         = true
  }

  network_interface {
    network = "default"
  }

  metadata {
    foo = "bar"
  }

  metadata_startup_script = "somescript"
}
```

and this instance conf:

```tf
resource "google_compute_instance" "instance" {
  name        = "terraform-test"
  description = "test instance"
  zone        = "us-central1-a"

  tags = ["testing"]

  machine_type = "f1-micro"

  disk {
    image = "debian-cloud/debian-8"
  }

  network_interface {
    network = "default"
  }

  metadata {
    foo = "bar"
  }

  metadata_startup_script = "somescript"
}
```

I'm able to create, then plan, and get no diff. My understanding is that's what [this part](https://github.com/hashicorp/terraform/pull/10537/files#diff-999eb677a0030ed7bdcceb48295d6784R719) is for.

phinze: Ah yep you are right. So the read method filters out `startup-script` if and only if `metadata_startup_script` is set on both resources.

If the user clears out `metadata_startup_script` in the config there might be an odd edge case there, but I think that's far enough off the beaten path to accept some slightly confusing behavior.

Thanks for helping clarify! 👍

PR #10539, opened 2016-12-05
provider/azurerm support `license_type` virtual_machine property for Windows machines.

TStraub-rms:
The property `license_type` already exists in the resource `azurerm_virtual_machine`, however it was not being used [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_virtual_machine.go#L77).

The azure golang SDK supports this as one of its properties [here](https://github.com/Azure/azure-sdk-for-go/blob/master/arm/compute/models.go#L834).

This PR is aimed at closing this gap, mainly to support [Microsoft Azure Hybrid Use Benefit](https://azure.microsoft.com/en-us/pricing/hybrid-use-benefit/).

 - [x] Proper formatting
 - [x] Updated Documentation
 - [x] Wrote additional acceptance test
 - [x] Passing Travis CI Build

### DIFF
--- a/builtin/providers/azurerm/resource_arm_virtual_machine.go
+++ b/builtin/providers/azurerm/resource_arm_virtual_machine.go
@@ -75,9 +75,10 @@ func resourceArmVirtualMachine() *schema.Resource {
 			},
 
 			"license_type": {
-				Type:     schema.TypeString,
-				Optional: true,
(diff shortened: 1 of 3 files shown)


stack72:
This LGTM! Thanks for the work here @TStraub-rms :)

Issue #10544, opened 2016-12-05
azurerm_availability_set missing a force recreate


cchildress:
### Terraform Version
0.7.13

### Affected Resource(s)
- azurerm_availability_set

### Terraform Configuration Files
Before:

```hcl
resource "azurerm_availability_set" "foobar" {
  name                         = "foobar"
  location                     = "westeurope"
  resource_group_name          = "${azurerm_resource_group.rg.name}"
  platform_fault_domain_count  = "2"
  platform_update_domain_count = "3"
}
```

After:
```hcl
resource "azurerm_availability_set" "foobar" {
  name                         = "foobar"
  location                     = "westeurope"
  resource_group_name          = "${azurerm_resource_group.rg.name}"
  platform_fault_domain_count  = "3"
  platform_update_domain_count = "5"
}
```

### Debug Output
```
15:09:14 * azurerm_availability_set.foobar: compute.AvailabilitySetsClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="PropertyChangeNotAllowed" Message="Changing property 'platformFaultDomainCount' is not allowed."
15:09:14 * azurerm_availability_set.foobar: compute.AvailabilitySetsClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="PropertyChangeNotAllowed" Message="Changing property 'platformFaultDomainCount' is not allowed."
```

### Expected Behavior
Terraform should force a resource recreate for the availability set if the fault domain count is altered.

### Actual Behavior
Terraform attempts to change the fault domain count in Azure, but Azure returns a failure code.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create an availability set with n fault domains.
2. Alter the availability set to add to the number of fault domains.
3. Run `terraform apply` and it should fail to alter the AS.

stack72:
Closed via #10545

PR #10545, opened 2016-12-05
Force a recreate if the numbers for the AS are changed.

cchildress:
This PR is a proposed fix to handle [Issue 10544](https://github.com/hashicorp/terraform/issues/10544).

An update to either the fault domain count or the update domain count should force a recreate of the resource.


stack72:
LGTM! Thanks for the quick turnaround here

PR #10548, opened 2016-12-06
[WIP] support floating_ip when create instance with port

takaishi:
Hello, I fix the following behavior.

1. I run `terraform apply` with following tf file.

```
resource "openstack_compute_instance_v2" "myserver" {
  name = "myserver"
  image_name = "${var.image}"
  flavor_name = "${var.flavor}"
  security_groups = "${var.security_groups}"

  network {
    port = "${openstack_networking_port_v2.myserver-port.id}"
    floating_ip = "${openstack_compute_floatingip_v2.myserver-floating-ip.address}"
    access_network = true
  }
}
```

Success to create instance with floating_ip.

2. run `terraform plan` with no changes

```
~ openstack_compute_instance_v2.myserver
    network.0.floating_ip: "" => "192.0.2.10"
```

myserver has floating_ip exists, but terraform to attach floating_ip.
I expect terraform plan say 'no changes', so I create this PR.

Thanks.

### DIFF
--- a/builtin/providers/openstack/devstack/deploy.sh
+++ b/builtin/providers/openstack/devstack/deploy.sh
@@ -147,11 +147,14 @@ wget http://download.cirros-cloud.net/0.3.4/cirros-0.3.4-x86_64-disk.img
 glance image-create --name CirrOS --disk-format qcow2 --container-format bare < cirros-0.3.4-x86_64-disk.img
 nova flavor-create m1.tform 99 512 5 1 --ephemeral 10
 _NETWORK_ID=$(nova net-list | grep private | awk -F\| '{print $2}' | tr -d ' ')
+neutron port-create --name test-port $_NETWORK_ID
+_PORT_ID=$(neutron port-list | grep test-port | awk -F\| '{print $2}' | tr -d ' ')
 _EXTGW_ID=$(nova net-list | grep public | awk -F\| '{print $2}' | tr -d ' ')
 _IMAGE_ID=$(nova image-list | grep CirrOS | awk -F\| '{print $2}' | tr -d ' ' | head -1)
 echo export OS_IMAGE_NAME="cirros-0.3.4-x86_64-uec" >> openrc
 echo export OS_IMAGE_ID="$_IMAGE_ID" >> openrc
 echo export OS_NETWORK_ID=$_NETWORK_ID >> openrc
+echo export OS_PORT_ID=$_PORT_ID >> openrc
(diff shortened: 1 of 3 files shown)


jtopjian:
@takaishi Thanks for looking into this - this has been an issue that I've been meaning to research for a while and I'm not sure what the best solution is at the moment or even if there _is_ a good solution.

Right now, the best way to achieve this is to:

* Create a `openstack_networking_port_v2` resource.
* Create a `openstack_networking_floatingip_v2` resource and specify the port above
* Create a `openstack_compute_instance_v2` resource and specify the port in the `network` block.

The downside is that the instance resource won't know what the floating IP is since there is a degree of separation with the resources. _But_ the floating IP _will_ work and communicate with the instance.

_But_, that's how the Neutron API works. Floating IPs are attached to Ports and the Nova API has little insight into port details. What you've provided is definitely one solution -- coupling the Instance resource to the Neutron API, but I don't think that's the best way to handle this. Creating this tight coupling will cause some inflexibility down the road -- it has already happened with the Instance's support of the `volume` block which I'm currently working on decoupling.

One solution I'm thinking of is creating "floating ip attach" resources just like the new "volume attach" resources. 

The downside with both the "volume attach" and "floating ip attach" resources is that the Instance resource can't natively tell what floating IPs and volumes it has attached to it. I think this is OK. There are other ways in Terraform to aggregate this data into templates and meta-resources. Originally, the Instance resource was thought of like the "hub" of all OpenStack resources. Decoupling attributes like this will break this hub and be better in the long-term.

Also, the other downside (although this is only short-term) to adding the Neutron API to the instance resource is that older OpenStack deployments that don't use Neutron will break.

Given all of the above, I'd like to hear your thoughts and opinion on this :)

takaishi:
@jtopjian Thank you for your comment.

> Right now, the best way to achieve this is to:
> 
> - Create a openstack_networking_port_v2 resource.
> - Create a openstack_networking_floatingip_v2 resource and specify the port above
> - Create a openstack_compute_instance_v2 resource and specify the port in the network block.

In this method, I could to make instance well.

I agree your thought, because I did not fully understand openstack component.
I think that when only port and floating_ip are specified in network block, `terraform plan` and `terraform apply`  need to output an error.

jtopjian:
> I think that when only port and floating_ip are specified in network block, terraform plan and terraform apply need to output an error.

This is a good idea! I will take a look at implementing this. I'll leave this PR open until it's in place.

PR #10549, opened 2016-12-06
communicator/ssh: don't share rand object to guarantee unique values

### LINKED ISSUE #10463: Null resource count.index not unique in remote-exec
Hi there,

### Terraform Version
Terraform v0.8.0-dev (b15b8bd99aae33c5b68cbecf5aef375c0798147c+CHANGES)

### Affected Resource(s)
- null_resource
      - remote-exec
          - inline

```hcl
resource "null_resource" "deploy_app" {
  count = "${var.instance_count}"
  triggers {
    current_ec2_instance_id = "${element(aws_instance.mod.*.id, count.index)}"
  }
  provisioner "remote-exec" {
    connection {
      type = "ssh"
      user = "deploy"
      host = "bastion.${var.vpc}.${var.environment}.foo.com"
      timeout = "10s"
      agent = true
    }
    inline = [
      "echo INSTANCE_NUMBER=${count.index + 1}"
    ]
  }
  lifecycle {
    create_before_destroy = true
  }
}
```

### Expected Behavior
module.marketing_site.null_resource.deploy_app.0 (remote-exec): Connected!
module.marketing_site.null_resource.deploy_app.1 (remote-exec): Connected!
module.marketing_site.null_resource.deploy_app.1 (remote-exec): INSTANCE_NUMBER=1
module.marketing_site.null_resource.deploy_app.0 (remote-exec): INSTANCE_NUMBER=0

### Actual Behavior
module.marketing_site.null_resource.deploy_app.0 (remote-exec): Connected!
module.marketing_site.null_resource.deploy_app.1 (remote-exec): Connected!
module.marketing_site.null_resource.deploy_app.1 (remote-exec): INSTANCE_NUMBER=1
module.marketing_site.null_resource.deploy_app.0 (remote-exec): INSTANCE_NUMBER=1

OR

module.marketing_site.null_resource.deploy_app.0 (remote-exec): Connected!
module.marketing_site.null_resource.deploy_app.1 (remote-exec): Connected!
module.marketing_site.null_resource.deploy_app.0 (remote-exec): INSTANCE_NUMBER=2
module.marketing_site.null_resource.deploy_app.1 (remote-exec): INSTANCE_NUMBER=2
module.marketing_site.null_resource.deploy_app.0: Creation complete
module.marketing_site.null_resource.deploy_app.1: Creation complete

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
This is a dev build from master to pre-empt an rc2 release. To test #10338 and #10337.

### References
This seems to work correctly on 0.8.0 beta2, could this have been broken by #10338 ?

mitchellh:
Fixes #10463

I'm really surprised this flew under the radar for years...

By having unique PRNGs, the SSH communicator could and would
generate identical ScriptPaths and two provisioners running in parallel
could overwrite each other and execute the same script. This would
happen because they're both seeded by the current time which could
potentially be identical if done in parallel...

Instead, we share the rand now so that the sequence is guaranteed
unique. As an extra measure of robustness, we also multiple by the PID
so that we're also protected against two processes at the same time.

Don't think there is a realistic way to test this.

### DIFF
--- a/builtin/provisioners/remote-exec/resource_provisioner.go
+++ b/builtin/provisioners/remote-exec/resource_provisioner.go
@@ -178,7 +178,6 @@ func (p *ResourceProvisioner) runScripts(
 
 		remotePath := comm.ScriptPath()
 		err = retryFunc(comm.Timeout(), func() error {
-
 			if err := comm.UploadScript(remotePath, script); err != nil {
 				return fmt.Errorf("Failed to upload script: %v", err)
 			}
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/hashicorp/terraform/communicator/remote"
@@ -27,6 +28,13 @@ const (
 	DefaultShebang = "#!/bin/sh\n"
 )
 
+// randShared is a global random generator object that is shared.
+// This must be shared since it is seeded by the current time and
+// creating multiple can result in the same values. By using a shared
+// RNG we assure different numbers per call.
+var randLock sync.Mutex
+var randShared *rand.Rand
+
 // Communicator represents the SSH communicator
 type Communicator struct {
 	connInfo *connectionInfo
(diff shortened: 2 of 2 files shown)


chrisrlong:
If it is any help, this pr seems to have fixed my issues in #10463

Thanks

PR #10553, opened 2016-12-06
Explicitly state the count parameter cannot be used by modules

walterdolce:
As pointed out [in the Google Group today](https://groups.google.com/forum/#!msg/terraform-tool/T3eB4pZ6waU/5hqs2SyGCAAJ), the `count` parameter is ineffective when used in modules.

In order to avoid confusion and doubts, this PR makes it explicit in the documentation so that future comers will have a definitive answer should they have the same question in mind.


stack72:
Hi @walterdolce 

Thanks for this - we believe this is a useful addition :)

Paul

jamtur01:
Fixed formatting in #10584. It's always a good idea to use middleman to generate the docs to check it's all working. :)

PR #10560, opened 2016-12-06
provider/aws: Extract helper/encryption library

jen20:
This commit extracts the GPG code used for aws_iam_user_login_profile into a library that can be reused for other resources, and updates the call sites appropriately.


stack72:
LGTM!

mitchellh:
This is going to need unit tests, despite the code we're using being tested heavily. I'll add them later.

Issue #10551, opened 2016-12-06
Get aws_emr_cluster endpoint


vkulov:
Hi,

I have successfully installed provisioned an EMR cluster using the aws_emr_cluster tag.

I'm having troubles accessing the EMR cluster endpoint or public DNS so I can use my cluster. I don't see an exported variable that points to any private or public address of my EMR cluster.

Please help. How do I use cluster if I don't have its IP or endpoint.

Ninir:
Hi @vkulov 
Can you try using the `master_public_dns` attribute and tell us how it goes?

I think this is the thing you are looking for :)

Ninir:
It is live in https://github.com/hashicorp/terraform/pull/10563 !

vkulov:
Thank you. This is what I was looking for. :)

dhaneshcode:
any chance can we get the hostname of the master.

PR #10563, opened 2016-12-06
provider/aws: Improved the documentation for EMR Cluster

Ninir:
Fixes https://github.com/hashicorp/terraform/issues/10551

This adds a key information on the EMR cluster which is `master_public_dns`, allowing users to know the endpoint to call.
It also adds the attribute descriptions for missing ones.


stack72:
LGTM! Thanks for this @Ninir :)

PR #10564, opened 2016-12-06
provider/azurerm: Do not pass an empty string of license_type to AMR VMs

stack72:


jen20:
LGTM, should fix several currently broken acceptance tests.

PR #10583, opened 2016-12-07
provider/aws: Add Lightsail Key Pair resource

catsby:
This adds an `aws_lightsail_keypair` resource. KeyPairs in lightsail are their own thing, and not (yet?) compatible with the `aws_key_pair`. Users can create new key pairs and receive the public and private keys in this resource. They can optionally (recommended) encrypt the private key with a pgp key, for secure storing in state.

Users can also import an existing pub key but suppling `public_key`.


builtin/providers/aws/resource_aws_lightsail_key_pair.go
```diff
@@ -0,0 +1,214 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name_prefix": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
stack72: Should this conflict_with name?

builtin/providers/aws/resource_aws_lightsail_key_pair.go
```diff
@@ -0,0 +1,214 @@
...
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsLightsailKeyPairCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).lightsailconn
+
+	kName := d.Get("name").(string)
```
stack72: You are not checking that name_prefix has been used?

builtin/providers/aws/resource_aws_lightsail_key_pair.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+		if resp.KeyPair == nil {
+			return fmt.Errorf("[ERR] No KeyPair information found for CreateKeyPair response")
+		}
+		d.SetId(kName)
+
+		// private_key and public_key are only available in the response from
+		// CreateKey pair. Here we set the public_key, and encrypt the private_key
+		// if a pgp_key is given, else we store the private_key in state
+		d.Set("public_key", resp.PublicKeyBase64)
```
stack72: Will this be problematic for import scenarios?

builtin/providers/aws/resource_aws_lightsail_key_pair.go
```diff
@@ -0,0 +1,214 @@
...
+
+		// encrypt private key if pgp_key is given
+		pgpKey, err := encryption.RetrieveGPGKey(d.Get("pgp_key").(string))
```
stack72: is it worth calling the encryption helper here? Wouldn't it be better to do that only if d.Get("pgp_key").(string) has a value other than "" ?

builtin/providers/aws/resource_aws_lightsail_key_pair.go
catsby: What do you mean by "import" here? If you mean `terraform import`, that is not supported. AWS does not provide any way to get the `private_key` except on create.

builtin/providers/aws/resource_aws_lightsail_key_pair.go
catsby: If `strings.HasPrefix` [here](https://github.com/hashicorp/terraform/blob/f-lightsail-key-pair/helper/encryption/encryption.go#L19) is `false`, then `RetrieveGPGKey` will return whatever it got for `pgpKey` as input; in the scenario you describe, thats `""`. So, instead of doing `GetOk` and only going down this chain `if ok`, I just did this. I can change it if you feel strongly enough about it

builtin/providers/aws/resource_aws_lightsail_key_pair.go
catsby: done in 97ac4bde

catsby:
Updated @stack72 , take a look when you can

stack72:
LGTM! Thanks for making the changes so fast :)

PR #10589, opened 2016-12-07
Keep a consistent provider order.

kitsunde:
In the rest of the document the order is organization, access_key. Except in this one spot.


jen20:
Thanks @kitsunde!

PR #10608, opened 2016-12-08
Validate `effect` in aws_iam_policy_document data source

dougneal:
@stack72 : Replaces https://github.com/hashicorp/terraform/pull/10021

AWS allows only the case-sensitive strings `Allow` and `Deny` to appear in the `Effect` fields of IAM policy documents. Catch deviations from this, including mis-casing, before hitting the API and generating an error (the error is a generic 400 and doesn't indicate what part of the policy doc is invalid).


stack72:
You rock!

PR #10610, opened 2016-12-08
provider/azurerm: Fix issue that Backend Address Pools could not be used in Scale Sets

vitorarins:
Added support for Load Balancer Backend Address Pools in Virtual Machine Scale Set

### DIFF
--- a/builtin/providers/azurerm/resource_arm_virtual_machine_scale_set.go
+++ b/builtin/providers/azurerm/resource_arm_virtual_machine_scale_set.go
@@ -813,10 +813,19 @@ func expandAzureRmVirtualMachineScaleSetNetworkProfile(d *schema.ResourceData) *
 					},
 				},
 			}
-			//TODO: Add the support for the load balancers when it drops
-			//if v := ipconfig["load_balancer_backend_address_pool_ids"]; v != nil {
-			//
-			//}
+
+			if v := ipconfig["load_balancer_backend_address_pool_ids"]; v != nil {
+				bePoolIds := v.(*schema.Set).List()
+				subResources := make([]compute.SubResource, 0, len(bePoolIds))
+				for _, bePoolId := range bePoolIds {
+					id := bePoolId.(string)
+					subResource := compute.SubResource{
+						ID: &id,
+					}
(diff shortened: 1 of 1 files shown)


builtin/providers/azurerm/resource_arm_virtual_machine_scale_set.go
```diff
@@ -813,10 +813,19 @@ func expandAzureRmVirtualMachineScaleSetNetworkProfile(d *schema.ResourceData) *
...
-			//TODO: Add the support for the load balancers when it drops
-			//if v := ipconfig["load_balancer_backend_address_pool_ids"]; v != nil {
-			//
-			//}
+
+			if v := ipconfig["load_balancer_backend_address_pool_ids"]; v != nil {
+				bePoolIds := v.(*schema.Set).List()
+				subResources := make([]compute.SubResource, 0, len(bePoolIds))
+				for _, bePoolId := range bePoolIds {
+					id := bePoolId.(string)
```
katcipis: This can panic, is that ok ? Or perhaps it should be better to handle the possible failure and return an error

vitorarins: The rest of the code isn't handling the possible failure in this scenario. Maybe I should refactor the rest of the code to handle this.

pmcatominey: It's safe to assume that the entries in the slice will be of type `string`, no harm in sanity checking though 😀

stack72:
@pmcatominey what do we think is needed here? I'd like to try and get this merged for 0.8.3 as it's a simple addition

pmcatominey:
Theres no acceptance test for attaching to a load balancer, also the `ip_configuration` set doesn't have a hash function so I'm not sure if changes will be detected properly.

stack72:
Hi @vitorarins 

Thanks for opening the PR here - I am going to close this out for now. We haven't heard back from you in a while and we still believe that this is a WIP.

Please do let us know if you want to update the PR

Thanks again

Paul

PR #10613, opened 2016-12-08
Test for minimum replacement set when using splat operator

jbardin:
See if we can reduce the replaced resources when a singe tainted node is referenced via the splat operator.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -6902,3 +6902,115 @@ module.middle.bottom:
 		t.Fatalf("expected: \n%s\n\nbad: \n%s", expected, actual)
 	}
 }
+
+// Test that replacing resources using using the splat operator aren't replaced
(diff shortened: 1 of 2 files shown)


jbardin:
@mitchellh: this is just a test that's rigged to pass for now. You think this is valid enough to warrant merging?

mitchellh:
@jbardin I'd rather attach the test (maybe as a diff, maybe just as code) to issue #3449 so we can just make it fail right away when we start fixing that. I plan on keeping that issue open until its fixed and its on the roadmap for 0.9 so we'll run into it.

jbardin:
SGTM, just didn't want to throw this test away if it's useful. I'll reference it over there and close this.

mitchellh:
Sounds good. Yeah, as long as we have a link to it or the source I think we won’t lose it since #3449 won’t close.

PR #10615, opened 2016-12-08
provider/aws: Encrypt aws_iam_access_key.secret with pgp

catsby:
Optionally encrypt the iam access key secret with a pgp key.

This is an alternative to #10450 where we forced users to supply a PGP Key, and was a breaking change. 

Here we maintain the current behavior, but offer a `pgp_key` parameter to optionally encrypt the returned secret. 

The upgrade UX becomes "nothing"; there is no diff or plan or change. Adding a `pgp_key` will result in a ForceNew plan.


mcanevet:
@catsby @mitchellh `secret` is not set anymore with this patch. Would it be possible to have a deprecation period of one release before removing this feature so that code written for 0.7 still works in 0.8?

PR #10616, opened 2016-12-08
Forward SIGTERM and handle that as an interrupt

mitchellh:
SIGTERM, unlike SIGINT, isn't sent to the process group. It is only sent to the parent process. So we need to first manually forward this from panicwrap and then second handle it as part of our interrupt handling.


mitchellh:
Also /cc @mildred I think this fixes everything :D

Issue #10617, opened 2016-12-08
AWS: Support Canada Central 1 Region


tecnobrat:
Canada Central region has just been released! 🎉 

https://aws.amazon.com/about-aws/whats-new/2016/12/announcing-the-aws-canada-central-region/

The region code is `ca-central-1`

Issue #9249, opened 2016-10-06
AWS API GW should support AWS_PROXY integration type.


BerndWessels:
Hi there,

we desperately need `AWS_PROXY` as type for our `API Gateway integration` to the `Lambda function`.

AWS actually supports that, but terraform seems not to be updated with the new types.

### Terraform Version
v0.7.4

### Affected Resource(s)
- aws_api_gateway_integration

### Terraform Configuration Files
```hcl
resource "aws_api_gateway_integration" "SeoWebsite" {
  rest_api_id = "${aws_api_gateway_rest_api.SeoWebsite.id}"
  resource_id = "${aws_api_gateway_resource.SeoWebsite.id}"
  http_method = "${aws_api_gateway_method.SeoWebsite.http_method}"
  type = "AWS_PROXY"
  uri = "arn:aws:apigateway:${var.aws_region}:lambda:path/2015-03-31/functions/${aws_lambda_function.SeoWebsite.arn}/invocations"
  integration_http_method = "${aws_api_gateway_method.SeoWebsite.http_method}"
}
```

### Expected Behavior
Allow `AWS_PROXY` value for `type`.

### Actual Behavior
`Errors: aws_api_gateway_integration.SeoWebsite: "type" must be one of 'AWS', 'MOCK', 'HTTP'`

### References
- https://aws.amazon.com/blogs/aws/api-gateway-update-new-features-simplify-api-development/

kwilczynski:
@BerndWessels hi there! I have a very good news for you!

The work on adding this has already been done, and is due to release together with next point-release, as per: https://github.com/hashicorp/terraform/pull/9213

Keep an eye on an upcoming release.

stack72:
Closed via #9213 :)

PR #10618, opened 2016-12-08
Add support for AWS CA Central 1 Region

eddywashere:
New region `ca-central-1`, and [blog post](https://aws.amazon.com/about-aws/whats-new/2016/12/announcing-the-aws-canada-central-region/).

Closes https://github.com/hashicorp/terraform/issues/10617

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -309,6 +309,7 @@ func (c *Config) ValidateRegion() error {
 		"ap-south-1",
 		"ap-southeast-1",
 		"ap-southeast-2",
+		"ca-central-1",
 		"cn-north-1",
 		"eu-central-1",
 		"eu-west-1",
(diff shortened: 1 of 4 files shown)


catsby:
Hey @eddywashere thanks! Is there any more info on this new region that we can use in our other files that relate to regions? See https://github.com/hashicorp/terraform/pull/9429 as an example

apparentlymart:
Thanks for this, @eddywashere.

There are a few other things we need to set to get full support for a new region (lookup tables for service ids that differ in each region, as discussed in #10470) but this is a good start.

It looks like there is some debate in #10470 about whether to enable the region before all of the relevant metadata has been updated. Presumably whatever decision is made there will apply here too, so I'm mainly just posting this comment to create the connection between the two issues.

eddywashere:
Added additional metadata. This uses the new region syntax for s3 website endpoints, so no change needed there. Is there a full list of potential data to add or is the info in #9249 sufficient? If it is, then I'm just waiting on elb info to be published.

radeksimko:
Here's the missing ALB/ELB account ID: `985666609251`.

Even though it's not published anywhere yet, you can go ahead and enable logging for ALB/ELB in the AWS Console and let the console create the S3 bucket - which includes creation of the policy. The policy then contains the relevant account ID. 😉

eddywashere:
Thanks @radeksimko. Just verified and added it.

stack72:
Thanks for the fast turnaround here @eddywashere :)

Paul

darrin-wortlehock:
Will this make it into 0.8 release?

stack72:
it certainly will @darrin-wortlehock  - it's already in master :)

PR #10635, opened 2016-12-09
provider/aws: aws_vpc_peering_connection: import fix

dougneal:
When importing an `aws_vpc_peering_connection`, the code assumes that the account under Terraform control is the initiator (requester) of the VPC peering request. This holds true when the peering connection is between two VPCs in the same account, or when the peering connection has
been initiated from the controlled account to another.

However, when the peering connection has been initiated from a foreign account towards the account under management, importing the peering connection into the statefile results in values of `peer_vpc_id` and `vpc_id` being the opposite way round to what they should be, and in the `peer_owner_id` being set to the managed account's ID rather than the foreign account's ID.

This patch checks the Accepter and Requester Owner IDs against the AWS connection's reported owner ID, and reverses the mapping if it is determined that the VPC peering connection is owned by the foreign account.


catsby:
Looks good, thanks!

PR #10639, opened 2016-12-09
command/plan: user friendly error if plan file given to plan command

### LINKED ISSUE #10627: Terraform plan acts strangely when given a plan file as input
Hi,

Apologies if this is a duplicate, I couldn't find an existing issue.

The usage information for `terraform plan` says the following:

```
Usage: terraform plan [options] [dir]

  Generates an execution plan for Terraform.

  This execution plan can be reviewed prior to running apply to get a
  sense for what Terraform will do. Optionally, the plan can be saved to
  a Terraform plan file, and apply can take this plan file to execute
  this plan exactly.
```

But it appears that providing an existing plan file as the `dir` option causes the plan to be a noop, even though it appears to work as normal. This isn't critical or anything, but it's something that can easily catch one out if one accidentally omits the `-out`.

### Terraform Version

v0.7.13

### Affected Resource(s)

All resource types.

Any config should work, for the purposes of reproduction let's assume just use the config from the [aws_instance example](https://www.terraform.io/docs/providers/aws/r/instance.html).

```hcl
provider "aws" {
    region = "us-west-2"
}

data "aws_ami" "ubuntu" {
  most_recent = true
  filter {
    name = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
  }
  filter {
    name = "virtualization-type"
    values = ["hvm"]
  }
  owners = ["099720109477"] # Canonical
}

    ami = "${data.aws_ami.ubuntu.id}"
    instance_type = "t2.micro"
    tags {
        Name = "HelloWorld"
    }
}
```

### Expected Behavior

One of the following seems best:

 1. `terraform plan` should throw an error since the plan file isn't a directory containing terraform config.
 2. `terraform plan` should act as `terraform show` when given a plan file (i.e. not refresh and immediately output the human-readable plan) and the behaviour should be documented.

### Actual Behavior

Terraform performs a refresh, but outputs the previous plan file without any modifications.

### Steps to Reproduce

 1. Save the example config above as `example.tf`.

 2. `terraform plan -out plan-bug.tfplan`

    Terraform correctly determines that it has to create a new instance.

 3. `rm example.tf`.

 4. `terraform plan plan-bug.tfplan` (note the lack of `-out`).

    Terraform appears to act normally: it performs a refresh then outputs a human-readable plan. However, the displayed plan and the plan file will still contain the resources that were previously defined in `example.tf`.

mitchellh:
Fixes #10627 

This detects if a plan file was given to the `plan` command and gives an error. This didn't have any detrimental effect in practice but was just a weird case where we should be more explicit about what is happening. In practice it is a noop so we just show an error saying it isn't allowed since there is no practical reason to do it.

### DIFF
--- a/command/plan.go
+++ b/command/plan.go
@@ -61,7 +61,7 @@ func (c *PlanCommand) Run(args []string) int {
 	// This is going to keep track of shadow errors
 	var shadowErr error
 
-	ctx, _, err := c.Context(contextOpts{
+	ctx, planned, err := c.Context(contextOpts{
 		Destroy:     destroy,
 		Path:        path,
 		StatePath:   c.Meta.statePath,
@@ -69,6 +69,13 @@ func (c *PlanCommand) Run(args []string) int {
 	})
 	if err != nil {
 		c.Ui.Error(err.Error())
+		return 1
+	}
+	if planned {
+		c.Ui.Error(
+			"The plan command cannot be called with a saved plan file.\n\n" +
+				"The plan command expects a configuration directory as an argument.",
+		)
 		return 1
 	}
 
--- a/command/plan.go
+++ b/command/plan.go
@@ -72,11 +72,15 @@ func (c *PlanCommand) Run(args []string) int {
 		return 1
 	}
 	if planned {
-		c.Ui.Error(
-			"The plan command cannot be called with a saved plan file.\n\n" +
-				"The plan command expects a configuration directory as an argument.",
-		)
-		return 1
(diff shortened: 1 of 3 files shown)


mitchellh:
I'm actually going to change this. I talked to a few folks and they liked this UX for checking what a plan does (tell me the plan of this plan), but we should more explicitly warn that that is what is happening.

PR #10642, opened 2016-12-09
config: allow names to start with numbers

mitchellh:
This rolls back the change introduced in 0.8 RC3 and allows names to start with numbers.

In HIL, identifiers are still restricted to NOT start with numbers, but since identifiers in Terraform are always qualified with `var.` or `resource.` and so on, this becomes a non-issue for us.

This should significantly improve the upgrade process to 0.8. :)


stack72:
LGTM

PR #10643, opened 2016-12-09
Disaply interpolation strings for computed fields

jbardin:
Displaying interpolations strings when possible for computed fields in
the plan output makes for a nicer UX


jbardin:
Fixes #10643

PR #10648, opened 2016-12-09
Update Vagrantfile to Ubuntu 16.04

cblecker:
As the issues with Parallels and the Ubuntu 16.04 bento image have been fixed, this bumps the Vagrantfile to 16.04, and variable-izes the version numbers so it's clear and easy to do so in the future (or manually if you need to test a certain version). Also uses variables for RAM and CPUCOUNT, so that these can be altered manually depending on your system.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,8 +4,16 @@
 # Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
 VAGRANTFILE_API_VERSION = "2"
 
+# Software version variables
+GOVERSION = "1.7.4"
+UBUNTUVERSION = "16.04"
+
+# CPU and RAM can be adjusted depending on your system
+CPUCOUNT = "2"
+RAM = "4096"
+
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -4,8 +4,16 @@
 # Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
 VAGRANTFILE_API_VERSION = "2"
 
+# Software version variables
+GOVERSION = "1.7.4"
```
stack72: Why is this defined here and then again in line 16?

cblecker: This is the ruby variable, and then it's defined again lower in the shell script that vagrant runs. on the box. It's slightly redundant, but it allows the variables to be grouped together. I could however pull out the shell script variable, and just replace all of those with the ruby variable if you think that would be cleaner.

stack72:
Let's merge as is :) Thanks for the work here

PR #10651, opened 2016-12-10
documentation/aws: Fix example syntax

kjmkznr:


stack72:
Thanks for spotting this :)

PR #10654, opened 2016-12-10
Maps with -var-file or -var always merge

### LINKED ISSUE #8219: Multiple var files with maps behaviour with 0.7.0
In updating some existing 0.6.x files to 0.7.x I'm running into a problem with multiple var files and maps. Previously I had one public vars file and one secret vars file, the result would be merged together for a full set, like so:

`a.tfvars`:

``` hcl
mymap.foo = "foo"
```

`b.tfvars`:

``` hcl
mymap.bar = "bar"
```

But when I've updated this to the newer syntax, it seems the merging behaviour has been lost. So this is either a bug report or a usage question, I'm not sure which. I looked but did not find documentation on this change coming into 0.7.x.

I believe I can work around this for now by using top level vars for secrets instead of trying to partially apply / marge a map like this, but I'd like to know the proper approach for this if this is not a bug.
### Terraform Version

Terraform v0.7.0
### Affected Resource(s)

Terraform core

`test.tf`:

``` hcl
variable "mymap" {
  default = {
    foo = "DEFAULT"
    bar = "DEFAULT"
  }
}

output "foo" {
  value = "${var.mymap["foo"]}"
}

output "bar" {
  value = "${var.mymap["bar"]}"
}
```

`a.tfvars`:

``` hcl
mymap = {
  foo = "foo"
}
```

`b.tfvars`:

``` hcl
mymap = {
  bar = "bar"
}
```
### Debug Output

https://gist.github.com/gwilym/63bcab48720682e868043c5cc177f1cc
### Expected Behavior

```
$ terraform apply -var-file=a.tfvars -var-file=b.tfvars

Outputs:

bar = bar
foo = foo
```
### Actual Behavior

```
$ terraform apply -var-file=a.tfvars -var-file=b.tfvars

Outputs:

bar = bar
foo = DEFAULT
```
### Steps to Reproduce
1. `terraform apply -var-file=a.tfvars -var-file=b.tfvars` with the above files

If further information is required I'm happy to provide.

mitchellh:
Fixes #8219

Conflicting map values are always merged when setting it via the CLI or files.

This unifies the behavior across Terraform to be consistent: map values are _always merged_. There is another issue open concerning unsetting values, which we may have to consider at some point. However, until then the logic of "does this merge or override?" for maps was inconsistent depending _where you used the map_. This makes it uniform throughout.

**This shouldn't be backwards incompatible in any way.** In past versions where overriding happened, you may have lost keys. Now those keys likely stick around. But since it was an error before to reference non-existent keys and there is no way to set a default value, the end behavior should not change in any config.

This also introduces a new extracted helper library `helper/variables` which aims to unify all variable parsing, merging, etc. logic. This helps keep it uniform as well as is laying the foundation for the future for us to support nested maps/lists.

I didn't refactor the `terraform` package to use the new helper just because we're so close to Terraform 0.8. I think that is a good idea for 0.9, since I'd want it to go through beta sweeps.


helper/variables/parse.go
```diff
@@ -0,0 +1,118 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/hcl"
+)
+
+// ParseInput parses a manually inputed variable to a richer value.
+//
+// This will turn raw input into rich types such as `[]` to a real list or
+// `{}` to a real map. This function should be used to parse any manual untyped
```
jbardin: is this supposed to be "manually typed"?

mitchellh: @jbardin "untyped" in the sense it isn't an ast.Variable, its just an `interface{}` map. I'll make it clearer though.

Issue #5340, opened 2016-02-26
Postgres provider: failure on destroy


saliceti:
It looks like the dependency is not respected when destroying postgres roles, databases and RDS instance. However, the graph shows the dependency clearly.
![image](https://cloud.githubusercontent.com/assets/8416694/13349463/e8ab5798-dc71-11e5-95ee-427d5ae272ca.png)
This is with terraform 0.6.12.

Code:

```
resource "aws_db_instance" "cf" {
  identifier = "cf"
  allocated_storage = 10
  engine = "postgres"
  engine_version = "9.4.5"
  instance_class = "db.t2.micro"
  username = "dbadmin"
  password = "abcd"
  db_subnet_group_name = "${aws_db_subnet_group.cf_rds.name}"
  parameter_group_name = "${aws_db_parameter_group.default.id}"

  vpc_security_group_ids = ["${aws_security_group.cf_rds.id}"]
}

provider "postgresql" {
  host = "${aws_db_instance.cf.address}"
  username = "abcd"
  password = "abcd"
}

resource "postgresql_role" "api" {
  name = "api"
  login = true
  password = "abcd"
}

resource "postgresql_database" "apidb" {
   name = "apidb"
   owner = "${postgresql_role.api.name}"
}
```

Error:

```
...
aws_db_instance.cf: Destroying...
...
aws_db_instance.cf: Destruction complete
...
postgresql_database.uaadb: Destroying...
postgresql_database.apidb: Destroying...
Error applying plan:

2 error(s) occurred:

* postgresql_database.uaadb: Error granting membership: dial tcp 10.0.1.185:5432: getsockopt: connection timed out
* postgresql_database.apidb: Error granting membership: dial tcp 10.0.1.185:5432: getsockopt: connection timed out
```

The only way to solve this issue is to edit the state file manually...

The state file shows only the remaining postgres resources:

```
{
    "version": 1,
    "serial": 24,
    "modules": [
        {
            "path": [
                "root"
            ],
            "resources": {
                "postgresql_database.apidb": {
                    "type": "postgresql_database",
                    "depends_on": [
                        "postgresql_role.api"
                    ],
                    "primary": {
                        "id": "apidb",
                        "attributes": {
                            "id": "apidb",
                            "name": "apidb",
                            "owner": "api"
                        }
                    }
                },
                "postgresql_role.api": {
                    "type": "postgresql_role",
                    "primary": {
                        "id": "api",
                        "attributes": {
[...]

PR #10659, opened 2016-12-11
terraform: destroy resources in dependent providers first

### LINKED ISSUE #4645: Destruction order for dependent providers is incorrect
Situation:
- I have RDS database created with aws_rds_instance resource
- I have MySQL/Postgre/MSSQL providers with <type>_database resource

For MySQL/Postgre/MSSQL provider endpoint is created RDS.

But I found that if I'm doing `terraform destroy` RDS instance is destroyed before database on it. I checked dependency graph and found that <type>_database resource is not linked to RDS resource. The only option here is to add `depends_on` flag

Should we think how to fix dependencies or at least we should document this somewhere, may be on MySQL/Postgre/MSSQL providers pages

My graph for 2 cases in attached files:
[with_depends_on.txt](https://github.com/hashicorp/terraform/files/87085/with_depends_on.txt)
[without_depends_on.txt](https://github.com/hashicorp/terraform/files/87086/without_depends_on.txt)

mitchellh:
Fixes #4645

This is something that never worked (even in legacy graphs), but as we
push forward towards encouraging multi-provider usage especially with
things like the Vault data source, I want to make sure we have this
right for 0.8.

When you have a config like this:

```
resource "foo_type" "name" {}
provider "bar" { attr = "${foo_type.name.value}" }
resource "bar_type" "name" {}
```

Then the destruction ordering MUST be:

  1. `bar_type`
  2. `foo_type`

Since configuring the client for `bar_type` requires accessing data from
`foo_type`. Prior to this PR, these two would be done in parallel. This
properly pushes forward the dependency.

**THIS PR:**

  - ✅  Resources of dependent providers are destroyed first
  - ✅  Resources of dependent providers _in child modules_ are destroyed first

There are likely more edge cases but I think I got the most common ones.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -2747,6 +2747,102 @@ func TestContext2Apply_multiProvider(t *testing.T) {
 	}
 }
 
+func TestContext2Apply_multiProviderDestroy(t *testing.T) {
+	m := testModule(t, "apply-multi-provider-destroy")
+	p := testProvider("aws")
+	p.ApplyFn = testApplyFn
+	p.DiffFn = testDiffFn
+
+	p2 := testProvider("do")
+	p2.ApplyFn = testApplyFn
+	p2.DiffFn = testDiffFn
+
+	var state *State
+
+	// First, create the instances
+	{
+		ctx := testContext2(t, &ContextOpts{
+			Module: m,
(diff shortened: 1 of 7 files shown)


mtekel:
Looks like this could also fix #5340, but I'm not 100% sure...

mitchellh:
You're right that does look the same!

PR #10662, opened 2016-12-11
provider/openstack: More Import and Region Fixes

jtopjian:
This has a few more import and region fixes from #10509

The most notable fix is changing the subnet allocation pool to computed so that automatically created allocation pools are correctly stored in state. The other fixes are just cleaning up acceptance tests (which I plan to do in greater detail at some point).

I came across these while running the full set of acceptance tests in preparation for the upcoming 0.8 release. All acceptance tests pass with this PR in place.


stack72:
LGTM! Nice work :)

jtopjian:
Thank you, sir!

Issue #9988, opened 2016-11-09
AWS RDS - Restoring from snapshot doesn't set DBname


kxseven:
### In a Nutshell
 - Creating an RDS instance by restoring from an existing Snapshot where the DBname will change results in a new RDS instance getting created with the DBname from the Snapshot.
 - Subsequent 'terraform apply' actions detect the incorrect DBname and force a new (RDS) resource, Re-creating the problem all over again

### Example
 - An RDS snapshot is created `myapp-sample-db` from an existing RDS instance
 - A new RDS instance is requested using the `myapp-sample-db` Snapshot identifier but with a different DBname `MYAPPDEV`

### Terraform Version
Terraform v0.7.7, Terraform v0.7.9

### Affected Resource(s)
- aws_db_instance

### Terraform Configuration Files
```hcl
resource "aws_db_instance" "development" {

    identifier                 = "myppdevdb"
    allocated_storage          = 500
    storage_type               = "gp2"
    engine                     = "oracle-ee"
    engine_version             = "12.1.0.2.v5"
    instance_class             = "db.r3.2xlarge"

    apply_immediately          = "true"

    password                   = "<redacted>"
    username                   = "<redacted>"
    name                       = "MYAPPDEV"

    publicly_accessible        = "false"
    db_subnet_group_name       = "<redacted>"
    parameter_group_name       = "<redacted>"
    option_group_name          = "<redacted>"
    copy_tags_to_snapshot      = "true"

    multi_az                   = "true"
    storage_encrypted          = "true"
    kms_key_id                 = "${var.kms_key_id}"
    vpc_security_group_ids     = [ "<redacted>" ]

    backup_retention_period    = 7
    backup_window              = "23:00-23:45"
    auto_minor_version_upgrade = "true"

    snapshot_identifier        = "myapp-sample-db"

    skip_final_snapshot        = "true"
    final_snapshot_identifier  = "devdb-final-snapshot"

}
```

### Expected Behavior
 - A new RDS instance should be created as a [RestoreDBInstanceFromDBSnapshot](https://docs.aws.amazon.com/AmazonRDS/latest/CommandLineReference/CLIReference-cmd-RestoreDBInstanceFromDBSnapshot.html) applying the DBname change if specified in the HCL

### Actual Behavior
 - The RDS instance gets created from the Snapshot but with the DBname stored in the Snapshot
[...]

PR #10664, opened 2016-12-11
provider/aws: Set the DBName when restoring from a snapshot

Ninir:
Closes https://github.com/hashicorp/terraform/issues/9988

Not sure if we want to add a test for that (since it can be verrrrrrry long).


stack72:
Hi @Ninir 

This LGTM! Thanks for the work here :)

P.

Issue #9556, opened 2016-10-24
Proposal: State Encryption


apparentlymart:
Currently we have several resources that retrieve or generate secrets, and for any where these secrets are used to populate other resources or configure other providers these secrets must necessarily be stored in the state.

Such resources include:
- `aws_db_cluster` (password attribute)
- `azurerm_virtual_machine` (machine login passwords)
- `tls_private_key`
- `vault_generic_secret` (both managed resource and data source) (#9158)
- ...and many other resources that use or produce passwords and keys

This causes some conflict, because Terraform's design originally assumed that the state was just a local cache of some remote data, and was fine to e.g. check into a git repo alongside the configuration, or to publish somewhere for consumption in other downstream Terraform configurations. It can be surprising and troublesome for secret values to show up in Terraform states that are being used in these ways.
# Proposal: Encrypt the state at rest

To address this issue in a way that does not significantly increase Terraform's core complexity, I propose that we address this by allowing Terraform state to _optionally_ be encrypted, as a whole, at rest. That is to say that the state file stored on local disk and on the remote storage target would be some sort of ciphertext of the state, and for each operation Terraform would retrieve this and decrypt it _in memory only_ for use during that operation, re-encrypting it before writing any changes.

Encrypting the entire state is a rather blunt instrument, but it has the advantage of allowing the encryption to be orthogonal to other concerns in Terraform, and thus makes it easy to reason about its behavior and understand what is and is not encrypted: all, or nothing.
## State Encryption Backends

Terraform already has the concept of a remote state _storage_ backend. This proposal introduces a similar but orthogonal concept alongside that: a state _encryption_ backend.

An encryption backend is responsible for translating from a cleartext state file to an encrypted one and vice-versa. The backend defines exactly what format the encrypted state file is stored in, and its contents are opaque to the rest of Terraform.

This would be enabled along with remote state storage in `terraform remote config`:

```
terraform remote config \
    -backend="consul" \
    -backend-config="address=consul.example.com:8500" \
    -backend-config="path=terraform/foo.tfstate" \
    -encrypt="vault" \
    -encrypt-config="address="https://vault.example.com/" \
    -encrypt-config="mount_point=foobaz" \ # "transit" by default
    -encrypt-config="key=terraform"
```

[...]

PR #4610, opened 2016-01-10
[WIP] JSON plan format

apparentlymart:
This change switches to a JSON serialization format for plans, rather than the `gob`-based binary serialization.

The motivation for this is to enable application-specific extra tools to do various sorts of checks or visualizations on the generated plan, to guide operators. For example:

* Automatically apply most changes to a QA/dev environment, but flag changes to certain resources for manual human review.
* Completely block certain changes that the application-specific logic knows must never happen, such as destroying a database.
* Perhaps just render the plan in a different medium than plain text, such as a web UI for Terraform that presents an HTML-formatted version of the plan.

Implementing this entails adding annotations to all of the structures included in a plan so that they can be JSON-serialized in a reasonable way. For just this change the diff feels rather disruptive, but it feels more reasonable if it's considered also a building block to enable machine readable progress output of various sorts as described in #2460: having a defined JSON serialization for Terraform's key structures makes Mitchell's suggestion there (having a new UI hook that produces JSON output) easier to achieve. 

My intent is that at first this wouldn't really be documented as a stable format; similar to the JSON state files, it's an implementation detail that happens to be human- and machine-readable to enable expert users to do certain things that wouldn't otherwise be possible, as long as they're willing to maintain those things as the format evolves.

Whereas JSON state files are long-lived and so deserve their own version numbering, I figured that plan files are very short-lived and so it's reasonable to refuse to apply plan files generated by different versions of Terraform. It seems pretty unlikely that someone would upgrade Terraform between making a plan and applying that plan, and even if they do it's a trivial matter to produce an updated plan with `terraform plan -refresh=false` to update the format.

This seems to basically work but I've not tested it extensively with various kinds of input and different sorts of plan. I intend to do some more thorough testing, but I figured I'd share this early to get feedback on the idea and in case others might want to try creating JSON plans against their own configurations and let me know if they run into any issues.

* [x] JSON tagging on relevant structs
* [x] JSON-serialized plan files
* [ ] Test against various configuration constructs

PR #3170, opened 2015-09-03
Initial stab at JSON plan output

glenjamin:
Guidelines said to submit even if it's early work, so here goes.

I'm pretty new to go, but have found it pretty easy to onboard into a build environment and navigating the codebase, good work everyone!

This relates to #2460 - specifically supporting JSON as plan output.

The particular scenario that led me to implement this went like this:
1. Run terraform apply to create a VM (via cloudstack)
2. VM takes ages to provision, terraform times out
3. VM was actually created, but terraform doesn't know about it

At this point, one option is to manually fill in the state file with a name and id for the resource, and terraform is then able to refresh and figure out where it got to.

My thinking was that it'd be nice to be able to machine read the plan (or the whole config perhaps?), and then consume this with a different tool that knows which fields should be unique, and can go fill out my state file based on what the plan might have acheived.

What do you think?

PR #10665, opened 2016-12-12
command: terraform show -format=json

apparentlymart:
This set of changes adds a `-format=json` option to the `terraform show` command to allow users to obtain machine-readable plan and state output.

There are (and have been) several other variants of adding some JSON output support to Terraform:

* In #4610 I tried to make JSON the primary format for plans, for consistency with states. However, this would've likely tempted people to try to hand-edit plan files, which is likely to cause Terraform to misbehave or crash.
* #3170 proposes to add JSON output to the `terraform plan` command itself. That would work, but is a little problematic because the `plan` command produces more output than just the plan, including the log of refreshing any existing resources and warnings about various things, and so it feels better (to me) to leave `terraform plan` as human-readable and provide a separate command for JSONifying plans.
* #2460 is a bit of a "meta-issue" aiming to get JSON support in a bunch of different places. JSON-readable logs of Terraform's ongoing behavior would be interesting, but is a lot of work to bite off.

So this change focuses on one primary use-case: I already made a plan using `terraform plan -out=...` and now I want to use some separate tool to analyze the plan.

```
$ terraform plan -out=tfplan


Your plan was also saved to the path below. Call the "apply" subcommand
with this plan file and Terraform will exactly execute this execution
plan.

Path: tfplan

(... etc, etc ...)

$ terraform show -format=json tfplan
{
   ...a big blob of JSON...
}
```

`terraform show` is used because it is focused on obtaining information without any other side-effects. This keeps the multi-format complexity from spreading all over the `command` module *and* should make a reasonable interface for producing a plan once and then safely rendering it in a number of different formats. (A maximum of two to start, but who knows...)

Since it is the `terraform show` command that is being extended here, it also supports JSON output of states. This is pretty redundant, since the state is already JSON-readable in exactly the same format in the local state file. It is thus included just for consistency and completeness for now, but could become more useful in future if, for example, we were to implement encryption of the local state file at rest (#9556), or if the local cache were to go away altogether as part of some forthcoming redesign of the remote state management workflow (#1964 or something like it).

This includes some of the work I did for #4610, though it required some tweaking to catch up with changes to Terraform's data structures in the mean time.

----

A big reason to be cautious about supporting something like this is that it creates a pretty big API surface area to maintain compatibility with. Terraform is still evolving very fast, so it's basically impossible to retain 100% compatibility with any machine-readable format. As a pragmatic compromise this change adds a JSON format for adventurous people to use, but specifically disclaims any defined behavior of it across Terraform versions and declines to document it in any way.

My thought was that for now this would remain a "buyer beware" sort of feature for adventurous early adopters, and then we'll get some experience with what people tend to do with it and then, at some later point, possibly commit to some level of compatibility for this format and produce some documentation on it.

### DIFF
--- a/command/format_plan.go
+++ b/command/format_plan.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strings"
@@ -228,4 +229,22 @@ func formatPlanModuleSingle(
 		"    %d resource(s)",
 		len(m.Resources)))
 	buf.WriteString(opts.Color.Color("[reset]\n"))
+}
+
+func FormatPlanJSON(plan *terraform.Plan) string {
+	err := plan.State.PrepareForWrite()
+	if err != nil {
+		// should never happen
+		panic(err)
+	}
+
+	dst := bytes.NewBuffer(make([]byte, 0, 64))
+	enc := json.NewEncoder(dst)
+	enc.SetIndent("", "  ")
+	err = enc.Encode(plan)
+	if err != nil {
+		// should never happen
+		panic(err)
+	}
+	return dst.String()
 }
--- a/command/format_state.go
+++ b/command/format_state.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strings"
@@ -149,4 +150,16 @@ func formatStateModuleSingle(
 
 	// Now just write how many resources are in here.
 	buf.WriteString(fmt.Sprintf("  %d resource(s)\n", len(m.Resources)))
+}
+
+func FormatStateJSON(state *terraform.State) string {
+	dst := bytes.NewBuffer(make([]byte, 0, 64))
+	enc := json.NewEncoder(dst)
+	enc.SetIndent("", "  ")
+	err := enc.Encode(state)
+	if err != nil {
+		// should never happen
+		panic(err)
+	}
+	return dst.String()
 }
--- a/command/show.go
+++ b/command/show.go
@@ -20,8 +20,11 @@ func (c *ShowCommand) Run(args []string) int {
 
 	args = c.Meta.process(args, false)
 
+	var format string
+
 	cmdFlags := flag.NewFlagSet("show", flag.ContinueOnError)
 	c.addModuleDepthFlag(cmdFlags, &moduleDepth)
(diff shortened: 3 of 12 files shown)


mitchellh:
There is probably meaningful work in here, but we're going to be completely revamping `terraform show` since it is not usable very quickly as state size increases. JSON output may help with that but there are other things I don't like about this:

The JSON format now creates some sort of API promise implicitly (though we definitely explicitly don't) of the structure, but the diff/state structure is going to continue to change heavily in upcoming versions, so anyone relying on this format will have to update their tools for each new version.

This has been a primary de-motivator for me for JSON output for state/diffs and why I push folks to just use UNIX tools with `terraform state` commands if possible for example. I understand there is a large class of problems that can't be solved that way, though.

And the people super motivated can easily (Go knowledge required, but past that it is easy) use the Terraform Go API (not promised stable but you can vendor) to parse states/plans and do clever things. That is a VERY high touch solution but at least its definitely "buyer beware" as you might say. 

So given that... I'm gonna say "no" to this and close it but I don't think your work will be a waste: we probably can and should adopt various parts of your diff in the future and knowing this exists will be useful.

apparentlymart:
Cool... that all makes sense. Thanks for the explanation, @mitchellh. A chunk of the code here was already from another rejected PR so maybe it will get a third chance later. :grinning:

As an aside: I previously tried to make use of Terraform's "API" from another Go program in the past; I guess it was probably my prototype of wrapping Terraform in a Packer-like UI. Ever since Terraform started using govendor this sort of usage has been generally broken, because any time a vendored library appears in Terraform's public interface (which happens with e.g. HIL's AST and HCL's concepts) you end up with types whose names cannot be written down by any code outside of the Terraform package.

Of course I don't expect anything to be done to fix that, since Terraform is primarily an application rather than a library, but just something to keep in mind when thinking about ways people might want to integrate with Terraform.

jalil:
Hey Michell,
Can  you give use the link to the Terraform Go API with some samples usages.

Thank you

hraban:
Hi @jalil I think he's talking about this very package we're commenting on right now. See https://github.com/hashicorp/terraform/issues/415#issuecomment-59308567

gazoakley:
It's a shame we can't have this built into Terraform - it'd be particularly helpful to include an option to strip away any attributes marked as sensitive in the schema. You could then build tools which take in plans/states for analysis. Maybe the exported JSON could include a schema semver?

PR #10453, opened 2016-12-01
providers/google: Add support for Internal Load Balancing

danawillow:
This change enables Internal Load Balancing with the introduction of two new resources: `google_compute_region_backend_service` and `google_compute_health_check`.

For all intents and purposes, this is ready for review. I still need to add documentation.

@evandbrown

PR #10676, opened 2016-12-12
providers/google: Add missing doc links

danawillow:
This accidentally got left out of #10453.
@evandbrown @paddyforan


paddyforan:
Oops, that's embarrassing, I should have caught that. Thanks for doubling back and fixing it :)

PR #10681, opened 2016-12-12
[WIP] providers/azurerm_vm - Only pass LicenseType if set in the terraform configuration.

TStraub-rms:
Only set the LicenseType property if the "license_type" string is not empty.


pmcatominey:
Hi @TStraub-rms 

It appears this change has already been made and is on [master](https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_virtual_machine.go#L499), which is why there is a conflict.

TStraub-rms:
@pmcatominey you are too fast. Also this issue was already fixed in the mainline :)

PR #3810, opened 2015-11-08
provider/aws: Support scratch volumes in ecs_task_definition

radeksimko:
Fixes https://github.com/hashicorp/terraform/issues/2370

Originally terraform was constructing the struct with sourcePath being empty string (when not defined in the DSL) which was causing API error.

### Test plan

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=EcsTaskDefinition' 2>~/tf.log
```
```
go generate ./...
TF_ACC=1 go test ./builtin/providers/aws -v -run=EcsTaskDefinition -timeout 90m
=== RUN   TestAccAWSEcsTaskDefinition_basic
--- PASS: TestAccAWSEcsTaskDefinition_basic (22.38s)
=== RUN   TestAccAWSEcsTaskDefinition_withScratchVolume
--- PASS: TestAccAWSEcsTaskDefinition_withScratchVolume (12.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	34.399s
```

PR #10683, opened 2016-12-12
Document ECS scratch volumes

xoob:
Update docs to say that volumes can be created without host_path, which will create an nonpersistent volume, as described in [Using Data Volumes in Tasks](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_data_volumes.html).

This feature was implemented in #3810 but did not appear in the docs.

@radeksimko


stack72:
LGTM! Thanks @xoob :)

PR #10694, opened 2016-12-13
Don't remove secret, just deprecate it

mcanevet:
Fix regression in #10615


catsby:
Thanks @mcanevet , great catch. I think though we should only save the `secret` if no PGP key is given, so I opened https://github.com/hashicorp/terraform/pull/10702 which builds off of this. 

Let me know if you agree 😄

catsby:
Closed in favor of #10702, thanks again for catching this!

Issue #10699, opened 2016-12-13
`vendor.json` references `github.com/lrsmith/terraform`


mdb:
It appears that `vendor/vendor.json` contains the following, as per [this commit](https://github.com/hashicorp/terraform/commit/015e96d0ddeab096b1b8e1c76190fe697504db27):

```
"rootPath": "github.com/lrsmith/terraform"
```

### Expected Behavior

I would expect it to contain:

```
"rootPath": "github.com/hashicorp/terraform"
```

stack72:
Closed via #10700

PR #10700, opened 2016-12-13
correct `rootPath`

mdb:
This previously referenced `github.com/lrsmith/terraform`

Fixes issue #10699


stack72:
Hi @mdb 

Thanks for this - good spot!

Paul

PR #10702, opened 2016-12-13
provider/aws: Save `iam_access_key` secret to state if no PGP key given (supersedes #10694)

catsby:
This pr builds of off #10694 and only saves the secret if no PGP is given.


catsby:
Pulling in

PR #10707, opened 2016-12-13
Dept of second thoughts: remove authorization support ASAP.

sean-:
When `postgresql_schema_policy` lands this attribute should be removed in
order to provide a single way of accomplishing setting permissions on
schema objects.  Removing ASAP to minimize any usage.


sean-:
Insta-merging before docs go up, or in case of a re-tag.

PR #10717, opened 2016-12-14
helper/variables: trim whitespace around the key in -var

### LINKED ISSUE #10716: The -var option seems totally broken in Terraform 0.8
### Terraform Version

Terraform v0.8.0

### Affected Resource(s)

All

Very simple repro example:

```hcl
variable "foo" {}
```

### Expected Behavior

Here's what happens with Terraform 0.7.x:

```
> terraform apply -var 'foo = "bar"' -input=false

```

### Actual Behavior

Here's what happens with the exact same template in Terraform 0.8.0:

```
> terraform apply -var 'foo = "bar"' -input=false
fix these before continuing.

Errors:

  * 1 error(s) occurred:

* Required variable not set: foo
```

### Steps to Reproduce

Run `terraform apply` and pass variables using the `-var` option.

mitchellh:
Fixes #10716

This trims whitespace around the key in the `-var` flag.

This is a regression from 0.7.x.

The value is whitespace sensitive unless double-quoted. This is the same
behavior as 0.7.x. I considered rejecting whitespace around the '='
completely but I don't want to introduce BC and the behavior actually
seems quite obvious to me.

Non-regression from 0.7.x but added in here trivially: verify the key is non-empty.

### DIFF
--- a/helper/variables/flag.go
+++ b/helper/variables/flag.go
@@ -21,6 +21,9 @@ func (v *Flag) Set(raw string) error {
 	}
 
 	key, input := raw[0:idx], raw[idx+1:]
+
+	// Trim the whitespace on the key
+	key = strings.TrimSpace(key)
 	if key == "" {
 		return fmt.Errorf("No key to left '=' in arg: %s", raw)
 	}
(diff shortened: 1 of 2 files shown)


helper/variables/flag.go
```diff
@@ -21,11 +21,18 @@ func (v *Flag) Set(raw string) error {
...
+	if key == "" {
+		return fmt.Errorf("No key to left '=' in arg: %s", raw)
+	}
+
 	value, err := ParseInput(input)
 	if err != nil {
 		return err
 	}
 
+	// Trim the whitespace on the key
```
jbardin: shoulnd't we trim the whitespace before the `""` comparison above

PR #10734, opened 2016-12-14
provider/aws: Added aws_iam_login_profile custom password setting

Ninir:
#### Reasoning for this change
This is the continuation of https://github.com/hashicorp/terraform/pull/9605.

The previous PR introduced PGP-based keys, encrypting and decrypting using keybase.io.
As the login_profile password was generated, it was missing the ability to generate a custom password. 
This allows to set it, making this even more flexible.

#### Real Configuration
```hcl
 resource "aws_iam_user" "user" {
     name = "demo"
     path = "/"
     force_destroy = true
 }
 
 data "aws_iam_policy_document" "user" {
     statement {
         effect = "Allow"
         actions = ["iam:GetAccountPasswordPolicy"]
         resources = ["*"]
     }
     statement {
         effect = "Allow"
         actions = ["iam:ChangePassword"]
         resources = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:user/&{aws:username}"]
     }
 }
 
 resource "aws_iam_user_policy" "user" {
     name = "AllowChangeOwnPassword"
     user = "${aws_iam_user.user.name}"
     policy = "${data.aws_iam_policy_document.user.json}"
 }
 
 resource "aws_iam_access_key" "user" {
     user = "${aws_iam_user.user.name}"
 }
 
 resource "aws_iam_user_login_profile" "user" {
     user = "${aws_iam_user.user.name}"
     password = "test"
 }
```

#### Use cases
Demos, trainings or handsons.

#### Acceptance tests
```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSUserLoginProfile_'
2016/12/14 21:58:25 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSUserLoginProfile_ -timeout 120m
=== RUN   TestAccAWSUserLoginProfile_basic
--- PASS: TestAccAWSUserLoginProfile_basic (22.11s)
=== RUN   TestAccAWSUserLoginProfile_keybase
--- PASS: TestAccAWSUserLoginProfile_keybase (20.71s)
=== RUN   TestAccAWSUserLoginProfile_keybaseDoesntExist
--- PASS: TestAccAWSUserLoginProfile_keybaseDoesntExist (12.11s)
=== RUN   TestAccAWSUserLoginProfile_notAKey
--- PASS: TestAccAWSUserLoginProfile_notAKey (14.28s)
=== RUN   TestAccAWSUserLoginProfile_customPassword
--- PASS: TestAccAWSUserLoginProfile_customPassword (19.61s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	88.850s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_user_login_profile_test.go
+++ b/builtin/providers/aws/resource_aws_iam_user_login_profile_test.go
@@ -169,7 +169,7 @@ func testDecryptPasswordAndTest(nProfile, nAccessKey, key string) resource.TestC
 				NewPassword: aws.String(generatePassword(20)),
 			})
 			if err != nil {
-				if awserr, ok := err.(awserr.Error); ok && awserr.Code() == "InvalidClientTokenId" {
+				if awserr, ok := err.(awserr.Error); ok && (awserr.Code() == "InvalidClientTokenId" || awserr.Code() == "EntityTemporarilyUnmodifiable") {
 					return resource.RetryableError(err)
 				}
 
--- a/builtin/providers/aws/resource_aws_iam_user_login_profile_test.go
+++ b/builtin/providers/aws/resource_aws_iam_user_login_profile_test.go
@@ -90,7 +90,27 @@ func TestAccAWSUserLoginProfile_notAKey(t *testing.T) {
 			{
 				// We own this account but it doesn't have any key associated with it
 				Config:      testAccAWSUserLoginProfileConfig(username, "/", "lolimnotakey"),
-				ExpectError: regexp.MustCompile(`Error encrypting password`),
+				ExpectError: regexp.MustCompile(`Error encrypting Password`),
+			},
+		},
+	})
+}
+
+func TestAccAWSUserLoginProfile_customPassword(t *testing.T) {
+	var conf iam.GetLoginProfileOutput
+
+	username := fmt.Sprintf("test-user-%d", acctest.RandInt())
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSUserLoginProfileDestroy,
+		Steps: []resource.TestStep{
+			{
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_iam_user_login_profile_test.go
```diff
@@ -169,7 +189,7 @@ func testDecryptPasswordAndTest(nProfile, nAccessKey, key string) resource.TestC
 				NewPassword: aws.String(generatePassword(20)),
 			})
 			if err != nil {
-				if awserr, ok := err.(awserr.Error); ok && awserr.Code() == "InvalidClientTokenId" {
+				if awserr, ok := err.(awserr.Error); ok && (awserr.Code() == "InvalidClientTokenId" || awserr.Code() == "EntityTemporarilyUnmodifiable") {
```
Ninir: This is required since I encountered this error:

> Check failed: Check 2/2 error: Error changing decrypted password: EntityTemporarilyUnmodifiable: Login Profile for User test-user-5423997845776594466 cannot be modified while login profile is being created.

jen20:
Hi @Ninir! The issue with setting a custom password is that it then appears in configuration unencrypted - is this the desired outcome?

Ninir:
Hi @jen20 !
Well, we can still hash the password in the state. In my case, it is sometimes useful to provide sequence-based passwords, like "user1", "user2" etc (training purposes).

Also, the variable can be passed using `-var file`, or using a Vault Provider in cases you really need to hide it. What do you think?

Ninir:
Also, I am able to do the update stuff (already prepared locally in fact), but it will require to impersonate the user, and update the profile.
Indeed, if you try to change the password without impersonating the user, the following error will be thrown:

    InvalidUserType: Only IAM Users can change their own password.

Thus, it will require to do the same as in tests, i.e. getting credentials for the user and updating the password AS the user.

Do you think we should go for the update in this way?

PR #10745, opened 2016-12-15
terraform: destroy edges should take into account module variables

### LINKED ISSUE #10729: destroying multiple modules uses incorrect order and causes race conditions
Destroying resources in multiple modules does not appear to order dependencies correctly.

### Terraform Version

master - (a9637f6)

### Affected Resource(s)

Tested with the following resources but I believe this is a core issue.

- azurerm_public_ip
- azurerm_resource_group
- azurerm_lb
- azurerm_lb_backend_address_pool

https://github.com/BedeGaming/tf-module-bug

### Debug Output

See `debug.log` in linked repo, sensitive values scrubbed.

### Expected Behavior

The resources should have been deleted in the correct order, that is the reverse of the order applied.

### Actual Behavior

The resources defined in each module were deleted at the time causing an error as the Public IP resource cannot be deleted while it is attached to the Load Balancer.

### Steps to Reproduce
1. clone linked [repo](https://github.com/BedeGaming/tf-module-bug)
1. configure AzureRM provider credentials
1. `terraform get`
1. `terraform apply`
1. `terraform destroy`

mitchellh:
Fixes #10729

Destruction ordering wasn't taking into account ordering implied through
variables across module boundaries.

This is because to build the destruction ordering we create a
non-destruction graph to determine the _creation_ ordering (to properly
flip edges). This creation graph we create wasn't including module
variables. This PR adds that transform to the graph.

### DIFF
--- a/terraform/graph_builder_apply_test.go
+++ b/terraform/graph_builder_apply_test.go
@@ -192,6 +192,46 @@ func TestApplyGraphBuilder_destroyCount(t *testing.T) {
 	}
 }
 
+func TestApplyGraphBuilder_moduleDestroy(t *testing.T) {
+	diff := &Diff{
+		Modules: []*ModuleDiff{
+			&ModuleDiff{
+				Path: []string{"root", "A"},
+				Resources: map[string]*InstanceDiff{
+					"null_resource.foo": &InstanceDiff{
+						Destroy: true,
+					},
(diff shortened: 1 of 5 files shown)


mitchellh:
Merging w/o review to expedite 0.8.1.

PR #10749, opened 2016-12-15
Make it clearer that ipv4_range is deprecated

danielbryantuk:
Removed google_compute_network example code usage that demonstrated deprecated ipv4_range argument, and made clearer that this argument is deprecated further down in the docs.


stack72:
Thanks @danielbryantuk 

this LGTM

P.

Issue #5664, opened 2016-03-16
Creating dynamically named aws_lambda_alias results in badness


dansteen:
Hey!

I created an aws_lambda_alias with a dynamic name:

```
resource "aws_lambda_alias" "assessment-trait-scores-scaler" {
        name = "${consul_keys.assessment-trait-scores-scaler.var.version}"
        description = "Git Hash ${consul_keys.assessment-trait-scores-scaler.var.version}"
        function_name = "${aws_lambda_function.assessment-trait-scores-scaler.arn}"
        function_version = "$LATEST"
}
```

The intention was to have an alias named after the git hash of the jar being generated.  This would create a new alias each time the jar is updated (I realize there are some issues with this - this was an early exploration).  Unfortunately, I get the following error after the first run:

```
* aws_lambda_alias.assessment-trait-scores-scaler: ResourceNotFoundException: Cannot find alias arn: arn:aws:lambda:us-east-1:xxxxxxx
:function:assessment-trait-scores-scaler-prod:ae91e35ffdc42c614fad63983d39ff5f21ff2fb1
```

Which is true, but not at all what I expected.  I expected it to remove the old alias, and create a new one, rather than attempting to "update" the new alias (which doesn't exist yet).   To resole this, I had to manually delete the entries out of the state file.

Thanks!
Dan

dthagard:
This PR doesn't address the underlying issue you are experiencing (namely that Terraform isn't properly removing the alias resource when changing the lambda), but it would fix the symptom you are having:  https://github.com/hashicorp/terraform/pull/10759

catsby:
I merged #10759 to address a small part of this.

> This would create a new alias each time the jar is updated

Unfortunately, this is not quite how Lambda Alias's work(at least, not how `aws_lambda_alias` works). The Alias is a pointer to a function version. On modification, `aws_lambda_alias` does not create a new alias, it simply updates it. 

If by chance you didn't strictly mean "create a new alias" and maybe just meant "update this alias" then there's nothing to see here 😄

PR #10759, opened 2016-12-15
Drop alias from state file if missing from lambda.

dthagard:
This commit fixes an issue where if you remove a AWS Lambda, the corresponding alias for that Lambda is also deleted.


dthagard:
This would also address the issue reported in https://github.com/hashicorp/terraform/issues/5664

catsby:
Thanks @dthagard !

Issue #10747, opened 2016-12-15
A list variable have inconsistant ordering


yamamoto-febc:
### Terraform Version
v0.8.0

### Affected Resource(s)
Core

### Terraform Configuration Files
```hcl
resource "docker_image" "image" {
  name = "alpine"
}

resource docker_container "arg10" {
  name = "arg10"
  image = "${docker_image.image.latest}" 
  entrypoint = ["/usr/bin/yes"]
  command = ["0","1","2","3","4","5","6","7","8","9"]
}

resource docker_container "arg11" {
  name = "arg11"
  image = "${docker_image.image.latest}" 
  entrypoint = ["/usr/bin/yes"]
  command = ["0","1","2","3","4","5","6","7","8","9","10"]
}

output arg10{
  value = ["${docker_container.arg10.command}"]
}
output arg11{
  value = ["${docker_container.arg11.command}"]
}
```

### Debug Output
```sh
arg10 = [
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
]
arg11 = [
    0,
    1,
    10,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
]
```

### Expected Behavior
List order should be consistent and independent on number of items in list

### Actual Behavior
When 11 or more items are included in the list, order was disturbed

### Steps to Reproduce
1. `terraform apply`

martin-flaregames:
Looks like as if the output got sorted using an alphanumeric collecting order

jbardin:
The outputs here are sorted in lexicographical order, based on the address key of the value internally. 

The order is only shown incorrectly in the output, but I'll see if we can clean that up for display.

jbardin:
Actually I was mistaken, it's not only in this order for display. 

The interpolated list is stored in the incorrect order, which means that the indexing order will be incorrect for interpolated lists, both within a module and through outputs.

jbardin:
closed via #10765

PR #10765, opened 2016-12-15
Sort interpolated lists by index number

jbardin:
Interpolated lists need to maintain the original list order

### DIFF
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -620,6 +620,33 @@ func (i *Interpolater) computeResourceMultiVariable(
 	return &variable, err
 }
 
+type indexKeys []string
+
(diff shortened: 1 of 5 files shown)


jbardin:
Fixes #10747

terraform/interpolate.go
```diff
@@ -648,7 +675,7 @@ func (i *Interpolater) interpolateComplexTypeAttribute(
 				keys = append(keys, id)
 			}
 		}
-		sort.Strings(keys)
+		sort.Sort(indexKeys(keys))
```
mitchellh: May be helpful to have a comment here on why we're sorting in what order.

jbardin: good point, added another comment

PR #10780, opened 2016-12-16
provider/aws: Fix broken documentat

f440:
- Fix broken document ... [aws/opsworks_rds_db_instance](https://www.terraform.io/docs/providers/aws/r/opsworks_rds_db_instance.html)


stack72:
Hi @f440 

Thanks so much for the fix here :) This LGTM!

Paul

PR #10803, opened 2016-12-16
Correct 'aws_hosted_zone' -> 'aws_route53_zone' in data source documentation

ewbankkit:
Some places in the documentation still referred to the data source as `aws_hosted_zone` after it had been renamed to `aws_route53_zone`.
I also renamed the HTML file and the once place that linked to it.


stack72:
thanks for this @ewbankkit :)

Good spot!

PR #10807, opened 2016-12-17
config: smarter provider alias usage validation

### LINKED ISSUE #4789: Provider Aliases Not Working w/ Modules?
I'm having an issue regarding a particular multi-region cluster configuration. Note that I'm currently only using modules to create separation within my configuration rather than focusing on creating reusable modules, so I'll have a root module with variables & providers and then, say, an AWS module that configures things like security groups or S3 buckets.

So in my top level (root) module I have a `providers.tf` file like:

```
provider "aws" {
  access_key = "${var.aws_access_key}"
  secret_key = "${var.aws_secret_key}"
  region = "${var.aws_region}"
}

provider "aws" {
  alias = "eu"
  region = "eu-central-1"
}
```

And then in a module I use to define bare AWS infrastructure resources I have something like this:

```
resource "aws_instance" "some_us_cluster_node" {
  # config ...
}

resource "aws_instance" "some_eu_cluster_node" {
  provider = "aws.eu"
  # config ...
}

...
```

However when I do this I get an error:
`aws_instance.some_eu_cluster_node: resource depends on non-configured provider 'aws.eu'`

So I tried instead defining the provider aliases in the `module` block rather than the root level module (while keeping the default `aws` provider at the root level):

```
module "aws" {
  source = "./aws"

  provider "aws" {
    alias = "eu"
    region = "eu-central-1"
  }
}
```

But then, interestingly enough, I get no errors but I get a prompt asking me what region to use for the `provider.aws.eu` resource as if I had left it out or as if I'm using a variable without a default value:

``` log
provider.aws.eu.region
  [description...]

  Default: us-east-1
  Enter a value:
```

So am I using the provider incorrectly? Do I need to propagate the provider alias down to the `aws` module like I do with variables? Or can I simply not use top level provider aliases inside of submodules? (which would be surprising since I can use a default `aws` provider defined at the top level inside of my submodules with no problem)

Finally I am on terraform version `0.6.9`

mitchellh:
Fixes #4789

This improves the validation that valid provider aliases are used.

Previously, we required that provider aliases be defined in every module
they're used. This isn't correct because the alias may be used in a
parent module and inherited.

This removes that validation and creates the validation that a provider
alias must be defined in the used module or _any parent_. This allows
inheritance to work properly.

We've always had this type of validation for aliases because we believe
its a good UX tradeoff: typo-ing an alias is really painful, so we
require declaration of alias usage. It may add a small burden to
declare, but since relatively few aliases are used, it improves the
scenario where a user fat-fingers an alias name.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -553,15 +553,6 @@ func (c *Config) Validate() error {
 		// Validate DependsOn
 		errs = append(errs, c.validateDependsOn(n, r.DependsOn, resources, modules)...)
 
-		// Verify provider points to a provider that is configured
-		if r.Provider != "" {
-			if _, ok := providerSet[r.Provider]; !ok {
-				errs = append(errs, fmt.Errorf(
-					"%s: resource depends on non-configured provider '%s'",
-					n, r.Provider))
-			}
-		}
-
 		// Verify provisioners don't contain any splats
 		for _, p := range r.Provisioners {
 			// This validation checks that there are now splat variables
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -456,13 +456,6 @@ func TestConfigValidate_providerMultiRefGood(t *testing.T) {
 	}
 }
 
-func TestConfigValidate_providerMultiRefBad(t *testing.T) {
-	c := testConfig(t, "validate-provider-multi-ref-bad")
-	if err := c.Validate(); err == nil {
-		t.Fatal("should not be valid")
-	}
-}
-
 func TestConfigValidate_provConnSplatOther(t *testing.T) {
 	c := testConfig(t, "validate-prov-conn-splat-other")
 	if err := c.Validate(); err != nil {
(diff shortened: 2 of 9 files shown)


apparentlymart:
It always felt to me like inheriting unaliased providers was a nice pragmatic decision to get the right behavior in the common case of "I have a bunch of modules with AWS stuff in them and I only want to configure this once".

Inheriting *aliased* providers feels a little more dubious to me because it creates a hidden dependency from a child module to once of its ancestors which can't be seen in the variables and outputs, and there's no way to loosen that coupling because the `provider` pseudo-attribute can't be interpolated from variables.

I realize that this PR is about validating for the existing behavior rather than changing the behavior, so this comment is not really on topic... I guess it just never really occurred to me that alias providers were inherited too, since that felt kinda counter-intuitive. This comment is more "thinking aloud" than a specific objection to the change at hand... :grinning:

mitchellh:
@apparentlymart I don't disagree, using a provider alias will limit the reusability of a module. Personally, I find it to be a relatively advanced feature and making inheritance available isn't such a bad thing.

PR #10821, opened 2016-12-18
provider/openstack: Handle Volume Creation Errors

### LINKED ISSUE #10376: OpenStack openstack_blockstorage_volume_v2 fails to become ready
Hi there,

### Terraform Version
Terraform v0.8.0-dev 
(after doing `brew upgrade terraform`, which claimed to install v0.7.13)

### Affected Resource(s)

- openstack_blockstorage_volume_v2
- openstack_blockstorage_volume_v1

This issue appears to affect multiple resources, it may be an issue with Terraform's core.

```hcl
resource "openstack_blockstorage_volume_v2" "glusterfs_volume_2" {
  name = "Fake volume"
  count = "1"
  description = "My fake volume"
  size = "10"
}
```

### Debug Output

See debug.log file in the gist:
https://gist.github.com/pcm32/dcd8f6eff19d54b75f65329790a03b21

### Panic Output

No panic error.

### Expected Behavior
Volume should be created.

### Actual Behavior
It fails saying:
```
Error applying plan:

1 error(s) occurred:

* openstack_blockstorage_volume_v2.glusterfs_volume_2: Error waiting for volume (483a4cc6-39fd-4281-a1f7-8f4e977fe0c7) to become ready: unexpected state 'error', wanted target 'available'. last error: %!s(<nil>)
```

if you apply again, it works. But it always fails the first time, on 2 out of 3 openstack instances we have access. This was part of a larger deployment, but I have reduced it to what is necessary to show the failure only.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Three different openstack installations, fails in two of them (RDO Liberty one of them, will find out other versions).

### References
Couldn't find anything related.

jtopjian:
This commit makes the openstack_blockstorage_volume resources better able
to handle volume creation errors upon resource creation. The cause of this
change is because there could be some storage backend error that happens
during storage provisioning that won't manifest in an "err" but will set
the volume's status to "error". We now check for a status of "error" and
propagate the error up the stack.

Fixes #10376

### DIFF
--- a/builtin/providers/openstack/resource_openstack_blockstorage_volume_v1.go
+++ b/builtin/providers/openstack/resource_openstack_blockstorage_volume_v1.go
@@ -130,9 +130,6 @@ func resourceBlockStorageVolumeV1Create(d *schema.ResourceData, meta interface{}
 	}
 	log.Printf("[INFO] Volume ID: %s", v.ID)
 
-	// Store the ID now
-	d.SetId(v.ID)
-
 	// Wait for the volume to become available.
 	log.Printf(
 		"[DEBUG] Waiting for volume (%s) to become available",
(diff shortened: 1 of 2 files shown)


stack72:
LGTM! Thanks for this :)

jtopjian:
Thanks, Paul!

PR #10822, opened 2016-12-18
provider/aws: Add support for iam_roles to aws_rds_cluster

### LINKED ISSUE #10166: Feature Request: Specify IAM role(s) for RDS Cluster
AWS Aurora [recently](https://aws.amazon.com/about-aws/whats-new/2016/10/amazon-aurora-new-features-aws-lambda-integration-and-data-load-from-amazon-s3-to-aurora-tables/) added supports adding roles to the RDS cluster. This can be used to integrate Aurora with other AWS services, such as S3 or Lambda.

For details on this please see: http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Aurora.Authorizing.AWSServices.html

It would be nice to have the ability to add IAM roles to an an `aws_rds_cluster` in much the same way as the `aws_redshift_cluster` resource's `iam_roles` attribute.

stack72:
Fixes #10166


willejs:
@stack72 looks like this works, any reason it cant get merged? Is it because its missing tests?

stack72:
Hi @willejs 

I am actually working on this code *right* now to get it finished :) Stick with me for a little while - it *will* be in the next version of TF

P.

builtin/providers/aws/resource_aws_rds_cluster.go
```diff
@@ -499,6 +511,16 @@ func resourceAwsRDSClusterRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("reader_endpoint", dbc.ReaderEndpoint)
 	d.Set("replication_source_identifier", dbc.ReplicationSourceIdentifier)
 
+	var roles []string
+	for _, r := range dbc.AssociatedRoles {
+		if *r.Status == "ACTIVE" {
+			roles = append(roles, *r.RoleArn)
```
catsby: The role status is one of `ACTIVE`, `PENDING`, or `INVALID`; do you not think it's worth storing the role even if it's not in the right state? I worry something will be `INVALID` and the user won't see it, and then Terraform will think it's just not there and try to reapply it.

builtin/providers/aws/resource_aws_rds_cluster.go
```diff
@@ -579,6 +601,27 @@ func resourceAwsRDSClusterUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	if d.HasChange("iam_roles") {
+		oraw, nraw := d.GetChange("iam_roles")
+		o := oraw.([]interface{})
+		n := nraw.([]interface{})
+		create, remove := rdsClusterDiffIamRoles(o, n)
+
+		for _, role := range create {
```
catsby: Are you aware of any reason where 2 roles could conflict, such that adding the new ones before removing the old ones, could cause a conflict or failure here?

builtin/providers/aws/resource_aws_rds_cluster.go
stack72: So the issue I faced here is that the UI only shows ACTIVE roles but the API returns the full history of the role associations - so i decided that we would only grab the attached ones

Maybe this isn't correct?

P.

builtin/providers/aws/resource_aws_rds_cluster.go
stack72: No, they have to be uniquely named - they cannot attach the same role twice

builtin/providers/aws/resource_aws_rds_cluster.go
grubernaut: I would have `roles` be a map or a set, so the user could query what the `STATUS` is of a role. It looks to me, that if a role is in the `INVALID` state, it's still attached to the db cluster, preventing that role from being attached again.

builtin/providers/aws/resource_aws_rds_cluster_test.go
```diff
@@ -557,6 +557,71 @@ resource "aws_rds_cluster" "default" {
     Environment = "production"
   }
 }`, n, n, n)
+
```
mikecook: missing ```}``` here

mikecook: Here's the diff...

```diff --git a/builtin/providers/aws/resource_aws_rds_cluster_test.go b/builtin/providers/aws/resource_aws_rds_cluster_test.go
index 798db9c4c..d73a0697c 100644
--- a/builtin/providers/aws/resource_aws_rds_cluster_test.go
+++ b/builtin/providers/aws/resource_aws_rds_cluster_test.go
@@ -557,6 +557,7 @@ resource "aws_rds_cluster" "default" {
     Environment = "production"
   }
 }`, n, n, n)
+}
 
 func testAccAWSClusterConfig_namePrefix(n int) string {
        return fmt.Sprintf(`
```

stack72:
Closed as migrating to new repository

PR #10830, opened 2016-12-19
provider/openstack: Block Device Tests and Docs

jtopjian:
This commit adds some basic tests for block device functionality. It also
expands the existing block device documentation as well references the
new "volume attach" resources for further block storage functionality.


stack72:
LGTM! Nice work :)

PR #10839, opened 2016-12-19
provider/openstack: Acceptance Test Cleanup: Provider & Block Storage

jtopjian:
This is the first of a series of commits to clean up the OpenStack acceptance tests. Updates are an attempt to better align with the styles of other Terraform acceptance tests.

All PRs in this series have been confirmed to pass testing, so the only review required would be for style formatting.

### DIFF
--- a/builtin/providers/openstack/provider_test.go
+++ b/builtin/providers/openstack/provider_test.go
@@ -9,8 +9,14 @@ import (
 )
 
 var (
-	OS_REGION_NAME = ""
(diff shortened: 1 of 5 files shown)


stack72:
LGTM! Thanks :)

jtopjian:
@stack72 Thanks, Paul! I have the rest of these ready to go, all in the same format. I can just push them through if that's OK?

stack72:
@jtopjian keep em coming sir :)

PR #10844, opened 2016-12-19
google: Add example for Internal Load Balancing

danawillow:
@evandbrown

### DIFF
--- a/examples/google-internal-load-balancing/README.md
+++ b/examples/google-internal-load-balancing/README.md
@@ -2,9 +2,20 @@
 
 This provides a template for setting up internal load balancing in Google Cloud. It directly mirrors the tutorial in the [GCP Internal Load Balancing Documentation](https://cloud.google.com/compute/docs/load-balancing/internal/).
 
-To run the example, you'll need to [download your credentials from Google Cloud Console](https://www.terraform.io/docs/providers/google/#credentials). A suggested path for the downloaded file is `~/.gcloud/Terraform.json`.
+To run the example,
 
-Optionally update `variables.tf` to specify a default value for the `project_name` variable, and check other variables.
+* Log in to gcloud with an account that has permission to create the necessary resources using `gcloud init`.
+* Optionally update `variables.tf` to specify a default value for the `project_name` variable, and check other variables.
+* Run with a command like this:
+
+```
+terraform apply \
+	-var="region=us-central1" \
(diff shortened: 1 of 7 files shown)


examples/google-internal-load-balancing/README.md
```diff
@@ -0,0 +1,38 @@
+# Internal Load Balancing in Google Cloud
+
+This provides a template for setting up internal load balancing in Google Cloud. It directly mirrors the tutorial in the [GCP Internal Load Balancing Documentation](https://cloud.google.com/compute/docs/load-balancing/internal/).
+
+To run the example, you'll need to [download your credentials from Google Cloud Console](https://www.terraform.io/docs/providers/google/#credentials). A suggested path for the downloaded file is `~/.gcloud/Terraform.json`.
```
evandbrown: @danawillow, what do you think about having the example use Cloud Shell, and not requiring credential download?

danawillow: Went with @cblecker's solution. Cloud Shell is a bit more complicated since then you'd have to download terraform onto the cloud shell vm.

danawillow:
Also, question for all. Right now it's not a perfect example of internal load balancing because the resulting compute instances have external IPs. In the tutorial, it explains that this is in order to download/install apache, and they have a step later where they delete the IPs. Any ideas of how to get a similar result without lingering external IPs at the end? I can just say "now delete them using gcloud" but that feels wrong.

cblecker:
It's starts to complicate things, but perhaps implementing a NAT gateway? Or maybe the external IPs aren't that big a deal, because it's the ILB IP itself that matters?

evandbrown:
@danawillow You should be able to do a `gcloud compute instances delete-access-config ...` as the last line in the startup script, so the VM is removing its own address. 

Fair point about Cloud Shell not having the `terraform` binary installed. But! Docker is installed, and there's an official Terraform image, so you can do something like:

```
docker run -i -t \
  hashicorp/terraform:light apply \
    -var projectid=${project_id} \
    -var region=${region} \
    -var zone=${zone}
```

But I think that's an exercise for another day. Agree that keeping it simple and consistent is the way to go here.

Nice work!

danawillow:
Ok, got delete-access-config working and moved the startup script into its own file.

evandbrown:
LGTM. Thanks, @danawillow!

Issue #10843, opened 2016-12-19
provider/docker: Add support for multiple docker_image pull_triggers


hmcgonig:
### Terraform Version
0.7.13

### Affected Resource(s)
Please list the resources as a list, for example:
- docker_image

### Terraform Configuration Files
```hcl
data "docker_registry_image" "elasticsearch" {
  name = "elasticsearch:latest"
}

resource "docker_image" "elasticsearch" {
  keep_locally = true
  name = "${data.docker_registry_image.elasticsearch.name}"
  pull_trigger = "${data.docker_registry_image.elasticsearch.sha256_digest}"
}
```

Currently, the above will pull the elasticsearch image into every host in my docker swarm cluster. In my environment I have found that when scaling up my docker swarm, I have no way to trigger another docker_image.elasticsearch pull so that my new swarm node gets an ES image. Currently, the workaround is that I would have to taint this resource to get it on my new node. Alternatively I can interpolate the cluster size and sha256 together.

I'm proposing support for something like this:
```hcl 
data "docker_registry_image" "elasticsearch" {
  name = "elasticsearch:latest"
}

resource "docker_image" "elasticsearch" {
  keep_locally = true
  name = "${data.docker_registry_image.elasticsearch.name}"
  pull_triggers = ["${data.docker_registry_image.elasticsearch.sha256_digest}", "${var.swarm_size}"]
}
```

So pull_trigger becomes a list called pull_triggers so that image pulls can be invoked by state changes other than just sha changes.

The other change required is that currently when pulling images, the docker provider first checks to see which images are already pulled, then if the image you want is not pulled, it pulls it. When using swarm, while setting up a new swarm host, terraform will not pull existing images onto new hosts, because those same images already exist on some nodes of the cluster already. The simplest fix I can see is to remove this check that sees if an image is already pulled. Pulling an image that you already have just results in a no-op with the side-effect that new swarm hosts will now also download the image as needed.

Later today I'll be throwing together a PR for this enhancement since it should be a very easy change, but I just wanted to get an issue out there for tracking. Thanks!

stack72:
Closed via #10845

PR #10845, opened 2016-12-19
provider/docker: Add support for a list of pull_triggers within the docker_image resource.

hmcgonig:
This PR is in response to an issue I opened earlier today: #10843 

Changes included are:
* Modify docker_image.pull_trigger (string) to now be a list of strings called docker_image.pull_triggers.
* Always pull images regardless of if they are found by searchLocalImages()
* Fixed acceptance test which used pull_trigger to now use the pull_triggers list.
* Updated relevant documentation

Why the PR?: When using the terraform provider to orchestrate a docker swarm cluster there are various times when you need to invoke a docker image pull outside of just the registry image sha256 changing. Take for example, adding a swarm node to the cluster. Once the new swarm node is up and accessible, an image pull must be triggered so that images are pulled onto the brand new node. To accommodate this I have changed pull_trigger into a list so that image pulls can be invoked by any number of things.

When testing this new functionality I found that images still were not being pulled onto the new host in the swarm. The reasoning behind this was because the provider was checking if an image was already found in the docker images list. When using docker swarm, 'docker images' will return an image if it is found in *any* node in the swarm, so in the case of a brand new swarm node, it will have no images and never be able to receive any images since the provider thinks the swarm already has the image. The best fix was to remove the check which sees if the image is already pulled and always attempt a pull operation. During this pull operation, nodes that already have the required docker image result in a no-op while nodes that need the image pull the image.

Example usage of pull_triggers attribute:
```hcl 
data "docker_registry_image" "elasticsearch" {
  name = "elasticsearch:latest"
}

resource "docker_image" "elasticsearch" {
  name = "${data.docker_registry_image.elasticsearch.name}"
  pull_triggers = ["${data.docker_registry_image.elasticsearch.sha256_digest}", "${var.swarm_cluster_size}"]
}
```

Thanks for taking a look, hopefully this is useful to others!

### DIFF
--- a/builtin/providers/docker/resource_docker_image_funcs.go
+++ b/builtin/providers/docker/resource_docker_image_funcs.go
@@ -188,15 +188,11 @@ func findImage(d *schema.ResourceData, client *dc.Client) (*dc.APIImages, error)
 		return nil, fmt.Errorf("Empty image name is not allowed")
 	}
 
-	foundImage := searchLocalImages(data, imageName)
-
-	if foundImage == nil {
-		if err := pullImage(&data, client, imageName); err != nil {
-			return nil, fmt.Errorf("Unable to pull image %s: %s", imageName, err)
-		}
+	if err := pullImage(&data, client, imageName); err != nil {
+		return nil, fmt.Errorf("Unable to pull image %s: %s", imageName, err)
 	}
 
-	foundImage = searchLocalImages(data, imageName)
+	foundImage := searchLocalImages(data, imageName)
 	if foundImage != nil {
 		return foundImage, nil
 	}
--- a/builtin/providers/docker/resource_docker_image_test.go
+++ b/builtin/providers/docker/resource_docker_image_test.go
@@ -89,6 +89,22 @@ func TestAccDockerImage_data(t *testing.T) {
 	})
 }
 
+func TestAccDockerImage_data_pull_trigger(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:                  func() { testAccPreCheck(t) },
+		Providers:                 testAccProviders,
+		PreventPostDestroyRefresh: true,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccDockerImageFromDataConfigWithPullTrigger,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestMatchResourceAttr("docker_image.foobarbazoo", "latest", contentDigestRegexp),
+				),
+			},
+		},
+	})
+}
+
 func testAccDockerImageDestroy(s *terraform.State) error {
 	for _, rs := range s.RootModule().Resources {
 		if rs.Type != "docker_image" {
@@ -131,6 +147,16 @@ data "docker_registry_image" "foobarbaz" {
 }
 resource "docker_image" "foobarbaz" {
 	name = "${data.docker_registry_image.foobarbaz.name}"
-	pull_trigger = "${data.docker_registry_image.foobarbaz.sha256_digest}"
+	pull_triggers = ["${data.docker_registry_image.foobarbaz.sha256_digest}"]
+}
+`
+
+const testAccDockerImageFromDataConfigWithPullTrigger = `
+data "docker_registry_image" "foobarbazoo" {
+	name = "alpine:3.1"
+}
+resource "docker_image" "foobarbazoo" {
+	name = "${data.docker_registry_image.foobarbazoo.name}"
+	pull_trigger = "${data.docker_registry_image.foobarbazoo.sha256_digest}"
 }
 `
--- a/builtin/providers/docker/resource_docker_image.go
+++ b/builtin/providers/docker/resource_docker_image.go
@@ -28,9 +28,19 @@ func resourceDockerImage() *schema.Resource {
 			},
 
 			"pull_trigger": &schema.Schema{
-				Type:     schema.TypeString,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"pull_triggers"},
+				Deprecated:    "Use field pull_triggers instead",
+			},
+
+			"pull_triggers": &schema.Schema{
+				Type:     schema.TypeSet,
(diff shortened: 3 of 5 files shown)


stack72:
Hi @hmcgonig 

Thanks for the work here - unfortunately, this is a backwards incompatibility that will affect our users. The correct way to handle this would be to mark the old parameter as `deprecated` and then have it conflicting with the new way of working

```
"pull_trigger": &schema.Schema{
				Type:          schema.TypeString,
				Optional:      true,
				ConflictsWith: []string{"pull_triggers"},
				Deprecated:    "Use field pull_triggers instead",
			},
```

Then in the code itself, we would need to cater for both pull_trigger and pull_triggers as we need to be able to support both sets of parameters

Sound ok?

Thanks

Paul

hmcgonig:
@stack72 Sure, that sounds great, I can make these changes sometime today. Thanks for the review.

stack72:
👍

hmcgonig:
Hey @stack72, I just reverted the removal of pull_trigger, marked it as a deprecated field in the schema+docs, then squashed the commit. Is there anything else I should do?

builtin/providers/docker/resource_docker_image_funcs.go
```diff
@@ -188,15 +188,11 @@ func findImage(d *schema.ResourceData, client *dc.Client) (*dc.APIImages, error)
 		return nil, fmt.Errorf("Empty image name is not allowed")
 	}
 
-	foundImage := searchLocalImages(data, imageName)
```
stack72: Why are we deleting the searchLocalImages check?

builtin/providers/docker/resource_docker_image_funcs.go
```diff
@@ -188,15 +188,11 @@ func findImage(d *schema.ResourceData, client *dc.Client) (*dc.APIImages, error)
...
-
-	if foundImage == nil {
-		if err := pullImage(&data, client, imageName); err != nil {
-			return nil, fmt.Errorf("Unable to pull image %s: %s", imageName, err)
-		}
+	if err := pullImage(&data, client, imageName); err != nil {
```
stack72: does this `pullImage` work for both pull_trigger and pull_triggers?

builtin/providers/docker/resource_docker_image_test.go
```diff
@@ -131,6 +131,6 @@ data "docker_registry_image" "foobarbaz" {
 }
 resource "docker_image" "foobarbaz" {
 	name = "${data.docker_registry_image.foobarbaz.name}"
```
stack72: I think it would be useful to have 2 tests - 1 for pull_trigger and 1 for pull_triggers

builtin/providers/docker/resource_docker_image_funcs.go
hmcgonig: Yes, this functionality is correct in either case.

builtin/providers/docker/resource_docker_image_test.go
hmcgonig: I can do this, sure.

builtin/providers/docker/resource_docker_image_funcs.go
hmcgonig: This was removed for two reasons, first because we do not need to care about if the image is found locally or not. A docker pull on an image that you already have just results in a no-op. Second, this is actually incorrect when using a docker-swarm. If you search local images in the swarm, it will return any image on any host in the swarm. Because of this behavior, it was previously possible for terraform to try and start a container on a swarm host that did not have the image pulled, which is incorrect.

hmcgonig:
@stack72 I left the existing "TestAccDockerImage_data" using the pull_triggers list since that is the new functionality going forward, but added in a test which uses pull_trigger which will need to be deleted when we remove that functionality.

stack72:
Hi @hmcgonig 

This looks much better now :) A nice upgrade path for users - we can remove the deprecation for 0.9 :)

```
% make testacc TEST=./builtin/providers/docker                                                                                ✭
2017/01/03 16:06:41 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/docker -v  -timeout 120m
=== RUN   TestAccDockerRegistryImage_basic
--- PASS: TestAccDockerRegistryImage_basic (6.08s)
=== RUN   TestAccDockerRegistryImage_private
--- PASS: TestAccDockerRegistryImage_private (2.02s)
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccDockerContainer_basic
--- PASS: TestAccDockerContainer_basic (20.17s)
=== RUN   TestAccDockerContainer_volume
--- PASS: TestAccDockerContainer_volume (21.19s)
=== RUN   TestAccDockerContainer_customized
--- PASS: TestAccDockerContainer_customized (31.77s)
=== RUN   TestAccDockerContainer_upload
--- PASS: TestAccDockerContainer_upload (18.97s)
=== RUN   TestAccDockerImage_basic
--- PASS: TestAccDockerImage_basic (4.05s)
=== RUN   TestAccDockerImage_private
--- PASS: TestAccDockerImage_private (2.16s)
=== RUN   TestAccDockerImage_destroy
--- PASS: TestAccDockerImage_destroy (2.41s)
=== RUN   TestAccDockerImage_data
--- PASS: TestAccDockerImage_data (8.50s)
=== RUN   TestAccDockerImage_data_pull_trigger
--- PASS: TestAccDockerImage_data_pull_trigger (8.13s)
=== RUN   TestAccDockerNetwork_basic
--- PASS: TestAccDockerNetwork_basic (0.43s)
=== RUN   TestAccDockerNetwork_internal
--- PASS: TestAccDockerNetwork_internal (1.03s)
=== RUN   TestAccDockerVolume_basic
--- PASS: TestAccDockerVolume_basic (0.23s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/docker	127.160s
```

Thanks for the work here

Paul

hmcgonig:
Woo! Thanks for guiding me through my first official terraform PR! Looking forward to more in the future. 👍

stack72:
Please continue to submit them :)

Issue #7690, opened 2016-07-18
DNS record change is not reflected in Cloudflare


makered:
Terraform: v0.6.16
Resource: cloudflare_record

I was able to successfully create a CNAME record on Cloudflare. Later I needed to change the DNS name so I modified the `name` attribute and applied the change. Result was:

```
module.elb.cloudflare_record.cf_dns_record_elb: Modifying...
  name: "oldname" => "newname"
module.elb.cloudflare_record.cf_dns_record_elb: Modifications complete
```

However, the Cloudflare record was unaltered and the output showed:

```
Outputs:
  cloudflare_elb_fqdn       = oldname.mydomain.com
```

HCL:

```
variable "dns_hostname" {
  description = "ELB DNS hostname"
  default = "newname"
}
```

```
resource "cloudflare_record" "cf_dns_record_elb" {
  domain = "${var.cloudflare_domain}"
  name = "${var.dns_hostname}"
  value = "${aws_elb.elb.dns_name}"
  type = "CNAME"
}
output "cloudflare_elb_fqdn" {
  value = "${cloudflare_record.cf_dns_record_elb.hostname}"
}
```

catufunwa:
I'm still running into this bug with the cloudflare provider on v0.7.3. Is it me or has the cloudflare provider been really flaky the last few version?

jlgeering:
issue still present in Terraform v0.7.13

joshuajabbour:
I can confirm this is present in `0.7.13`. The only workaround I have found is to comment out the individual `cloudflare_record` block so it's destroyed and then uncomment it so it's created anew with the proper name. Unfortunately, there's no `lifecycle` option to force recreation in this manner.

mwarkentin:
Also ran into this on `0.7.13` - wonder if `taint` would work to force a destroy + recreate?

mwarkentin:
Opened https://github.com/cloudflare/cloudflare-go/issues/97 to try and figure out if this is an issue with the upstream library.

mwarkentin:
It looks like this may just be an issue with the upstream Cloudflare library: https://github.com/cloudflare/cloudflare-go/issues/97#issuecomment-268362646

Trying to work with their engineer on a fix here: https://github.com/cloudflare/cloudflare-go/issues/98

mwarkentin:
[...]

PR #10846, opened 2016-12-19
Update test for updating cloudflare record

mwarkentin:
Start of work on #7690 - may tackle all or some of the following (depending on my ability):

* Update acceptance tests to demonstrate the bug
* Fix the resource to allow for record name updates

### DIFF
--- a/builtin/providers/cloudflare/resource_cloudflare_record_test.go
+++ b/builtin/providers/cloudflare/resource_cloudflare_record_test.go
@@ -119,11 +119,13 @@ func TestAccCloudFlareRecord_Updated(t *testing.T) {
 					testAccCheckCloudFlareRecordExists("cloudflare_record.foobar", &record),
 					testAccCheckCloudFlareRecordAttributesUpdated(&record),
 					resource.TestCheckResourceAttr(
-						"cloudflare_record.foobar", "name", "terraform"),
+						"cloudflare_record.foobar", "name", "terraform2"),
 					resource.TestCheckResourceAttr(
 						"cloudflare_record.foobar", "domain", domain),
 					resource.TestCheckResourceAttr(
 						"cloudflare_record.foobar", "value", "192.168.0.11"),
+					resource.TestCheckResourceAttr(
+						"cloudflare_record.foobar", "ttl", "3601"),
 				),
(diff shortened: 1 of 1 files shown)


builtin/providers/cloudflare/resource_cloudflare_record_test.go
```diff
@@ -119,11 +119,13 @@ func TestAccCloudFlareRecord_Updated(t *testing.T) {
 					testAccCheckCloudFlareRecordExists("cloudflare_record.foobar", &record),
 					testAccCheckCloudFlareRecordAttributesUpdated(&record),
 					resource.TestCheckResourceAttr(
-						"cloudflare_record.foobar", "name", "terraform"),
+						"cloudflare_record.foobar", "name", "terraform2"),
```
mwarkentin: Doesn't look like this is doing what I hoped - it seems to be checking the terraform state, instead of hitting the cloudflare API and checking the record there.

mwarkentin:
This is a terrible commit, but I was able to reproduce what we've seen
with cloudflare records in practice:

```
make testacc TEST=./builtin/providers/cloudflare TESTARGS='-run=TestAccCloudFlareRecord_UpdatedApi'
gofmt -w $(find . -name '*.go' | grep -v vendor)
2016/12/19 20:50:25 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/cloudflare -v -run=TestAccCloudFlareRecord_UpdatedApi -timeout 120m
=== RUN   TestAccCloudFlareRecord_UpdatedApi
--- FAIL: TestAccCloudFlareRecord_UpdatedApi (6.45s)
	testing.go:265: Step 1 error: Check failed: Check 7/7 error: Cloudflare record doesn't match state file:%!(EXTRA string=terraform.wave-juno.com, string=terraform2.wave-juno.com)
FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/cloudflare	6.472s
make: *** [testacc] Error 1
```

mwarkentin:
This turned out to be an upstream library issue - fixed in https://github.com/cloudflare/cloudflare-go/pull/98, closing.

PR #10848, opened 2016-12-19
Add CLI output example to Usage of Modules Docs

kford55:
Adding a CLI Example to grab output variables from a module.

**Reasoning for docs update**: It can be confusing for how to grab the output variables for a module from the CLI tool, so I added an example to show using the -module flag. It was confusing and a bit of a time sink for me so I figured it would be useful to show.

**Relevant Terraform version**: This could go into the docs now, we are using 0.7 and used this syntax.


mitchellh:
Thanks!

PR #10849, opened 2016-12-19
Add MD5 to archive_file's output

mtougeron:
This is a small change that adds the MD5 to the `archive_file` output. This is useful for cases like the way `aws_s3_bucket_object` requires an `etag`.


mtougeron:
Arg, commit ended up with the wrong user. Closing to fix...

PR #10851, opened 2016-12-20
Add the archive file's MD5 to the output

mtougeron:
This is a small change that adds the MD5 to the `archive_file` output. This is useful for cases like the way `aws_s3_bucket_object` requires an `etag`


apparentlymart:
Thanks, @mtougeron!

While merging I took the opportunity to add some tests for the various hash attributes in 6ad1e596dd3e2885741307a706ae8448c1a9919d.

mtougeron:
Awesome, thanks!

Issue #10823, opened 2016-12-18
google_compute_instance ForceNew argument create_timeout added without state migration


apparentlymart:
In 49c7d272a35b8cfbbb2c6231e4d1bd496e59d54b a new argument `create_timeout` was added that has a default value and sets `ForceNew` true. This forces a replace diff for anyone upgrading from an older Terraform version where this attribute wasn't present.

This commit seems to have shown up for the first time in 0.8, so this can be an upgrade blocker for people coming from 0.7.

The general solution to this sort of problem is to provide a state migration function so that a default can be inserted into the state. In this case however it seems that this attribute is used only on create anyway, so perhaps a simpler fix is to just drop the `ForceNew` altogether... it feels weird to assume that if someone changes the creation timeout that they necessarily want to recreate the instance in order to experience the new timeout. :grinning:

---

For anyone that hits this bug and wants a workaround to make progress:

Manually editing the state is probably the least-effort workaround here, though care must always be taken when editing the state since the state is Terraform's "memory" and careless editing can cause Terraform to misbehave.

To adjust this, locate the relevant `google_compute_instance` resource in the state file and add the new attribute with its default value of `4`:

```js
    "create_timeout": "4"
```

Be sure to preserve correct JSON syntax by adding a comma to the end to delimit from a following attribute value. The value must be quoted because Terraform treats all state attribute values as strings.

If you have remote state enabled, the file to edit is `.terraform/terraform.tfstate` and you must also increment the `serial` value near the top of the file so Terraform can understand that this was an intentional change rather than state corruption.

If you do not have remote state enabled, the file to edit is `terraform.tfstate`.

cblecker:
Completely agree with the dropping of ForceNew. If resource creation fails because of timeout, the resource would be marked as tainted already, wouldn't it?

paddyforan:
I'm on board with both dropping the ForceNew and adding a migration to set a default. Definitely don't want to block people on updating or have un-labeled backwards incompatibilities. Thanks for bringing it up. :) Just want to mention @aditya87 (the original author) and @evandbrown (reviewer) here to make sure we're not missing nuance or context, but I'll go ahead and make a PR for this today.

cblecker:
If we drop `ForceNew`, would we end up needing a state migration? The next apply should set the default, wouldn't it?

paddyforan:
[...]

PR #10858, opened 2016-12-20
Remove create_timeout backwards incompatibilities.

paddyforan:
A new create_timeout attribute was added that had some backwards
incompatibilities, and as per discussion in #10823, it was determined we
could make upgrading to 0.8.x easier by fixing them, without really
losing any functionality.

Because create_timeout is not something stored or transmitted to the
API, it's not something we need a ForceNew on. Also, because an update
wouldn't result in an API call, we can add a state migration to avoid a
false positive diff that requires people to plan and apply but doesn't
actually make an API call.


stack72:
LGTM! Thanks for removing this!

PR #10871, opened 2016-12-20
provider/aws: fixes double base64 encode

### LINKED ISSUE #10786: Provider AWS: Autoscaling Launch Configuration base64 encodes user_data regardless of input
### Affected Resource(s)
- aws_launch_configuration

### Supporting Links
How EC2 Instance handles User Data: https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_instance.go#L1062
How Launch Config handles User Data:
https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_launch_configuration.go#L313

### Expected Behavior
Based on functionality in aws_instance, launch configuration should base64 encode only if content is not already base64 encoded.

### Actual Behavior
aws_launch_configuration base64 encodes user_data regardless of the input.

curtisallen:
Add a utility that ensures a byte array is not doubly base64 encoded

Reference https://play.golang.org/p/RnEBFCJ9h0
Fixes #10786

### DIFF
--- a/builtin/providers/aws/utils.go
+++ b/builtin/providers/aws/utils.go
@@ -1,0 +1,19 @@
+package aws
+
+import "encoding/base64"
+
+// Base64Encode encodes data if the input isn't already encoded using base64.StdEncoding.EncodeToString.
+// If the input is already base64 encoded, return the original input unchanged.
+func base64Encode(data []byte) string {
+	// Check whether the data is already Base64 encoded; don't double-encode
+	if isBase64Encoded(data) {
+		return string(data)
+	}
+	// data has not been encoded encode and return
+	return base64.StdEncoding.EncodeToString(data)
+}
+
+func isBase64Encoded(data []byte) bool {
+	_, err := base64.StdEncoding.DecodeString(string(data))
+	return err == nil
+}
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -1058,14 +1058,7 @@ func buildAwsInstanceOpts(
 
 	user_data := d.Get("user_data").(string)
 
-	// Check whether the user_data is already Base64 encoded; don't double-encode
-	_, base64DecodeError := base64.StdEncoding.DecodeString(user_data)
-
-	if base64DecodeError == nil {
-		opts.UserData64 = aws.String(user_data)
-	} else {
-		opts.UserData64 = aws.String(base64.StdEncoding.EncodeToString([]byte(user_data)))
-	}
+	opts.UserData64 = aws.String(base64Encode([]byte(user_data)))
 
 	// check for non-default Subnet, and cast it to a String
 	subnet, hasSubnet := d.GetOk("subnet_id")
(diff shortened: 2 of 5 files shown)


builtin/providers/aws/utils.go
```diff
@@ -0,0 +1,13 @@
+package aws
```
nathany: Rather than a generic "utils" file, maybe this file should be named base64.go and likewise for the tests?

builtin/providers/aws/utils.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "encoding/base64"
+
+// Base64Encode encodes data if the input isn't already encoded useing base64.StdEncoding.EncodeToString.
+// If the input is already base64 encoded, return the original input unchanged.
+func base64Encode(data []byte) string {
+	// Check whether the user_data is already Base64 encoded; don't double-encode
+	if _, base64DecodeError := base64.StdEncoding.DecodeString(string(data)); base64DecodeError != nil {
+		return base64.StdEncoding.EncodeToString(data)
```
nathany: Since an error represents the expected path of not being encoded and therefore needing encoded, this may be clearer with a new more comments. `// not encoded, so encode it` `// already encoded`.

Or maybe even splitting out an `isBase64Encoded() bool` function?

builtin/providers/aws/utils.go
curtisallen: Yeah, I was looking for examples in other packages on what to name this file. In the `aws` package, there's an `auth_helper.go`. In other packages there are `utils.go`, just not sure which is the best fit for this use case.

builtin/providers/aws/utils.go
```diff
@@ -0,0 +1,19 @@
+package aws
+
+import "encoding/base64"
+
+// Base64Encode encodes data if the input isn't already encoded useing base64.StdEncoding.EncodeToString.
+// If the input is already base64 encoded, return the original input unchanged.
+func base64Encode(data []byte) string {
+	// Check whether the user_data is already Base64 encoded; don't double-encode
```
nathany: in this context it's just `data`, not `user_data`

builtin/providers/aws/utils.go
```diff
@@ -0,0 +1,19 @@
+package aws
+
+import "encoding/base64"
+
+// Base64Encode encodes data if the input isn't already encoded useing base64.StdEncoding.EncodeToString.
```
nathany: s/useing/using/

curtisallen: whoops good catch

builtin/providers/aws/utils.go
curtisallen: thanks will fix

jen20:
This LGTM, I'm going to rename the file after merge however. Thanks @curtisallen!

nathany:
Thanks James. Did you mean to rename utils.go?

[hub](https://github.com/github/hub) is pretty useful for amending pull requests (see [article](https://blog.spreedly.com/2014/06/24/merge-pull-request-considered-harmful/))

PR #10894, opened 2016-12-21
docs/aws: Add documentation for iam_role/policy import

stack72:


Hashfyre:
This is not yet documented at: https://www.terraform.io/docs/import/importability.html

PR #10903, opened 2016-12-22
removes region param from google_compute_backend_service

zachgersh:
- this param was not being used in this service
- you need a regional_backend_service if you want to pass this
- docs also updated

### DIFF
--- a/builtin/providers/google/resource_compute_backend_service.go
+++ b/builtin/providers/google/resource_compute_backend_service.go
@@ -117,12 +117,6 @@ func resourceComputeBackendService() *schema.Resource {
 				Computed: true,
 			},
 
-			"region": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-			},
-
 			"self_link": &schema.Schema{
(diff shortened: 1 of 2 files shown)


zachgersh:
@stack72 - can I get a review here?This should be a super easy merge.

stack72:
Hi @zachgersh 

Thanks for the work here - sorry it took so long to get back to you on it. So, the usual way we do this is to deprecate the parameter. IF we don't deprecate it, anyone who has set the param in their terraform config will get an error message. I suggest the following change:

```
"region": {
	Type:       schema.TypeString,
	Optional:   true,
  ForceNew:   true,
	Deprecated: "This parameter has been removed as it was never used",
},
```

Then for Terraform 0.9, we can completely remove the parameter

Does this make sense?

Thanks

Paul

zachgersh:
You got it, easy change and I will fire that in shortly. No apologies needed - you all are busy, totally understandable.

zachgersh:
@stack72 this should be good to go now - let me know if there is anything else here.

stack72:
LGTM! Thanks @zachgersh :)

PR #10787, opened 2016-12-16
Inteprolate list of maps

jbardin:
Fix `flatmap.Expand` to work correctly with nested maps. This can then be used to unpack all nested values for the Interpolator, rather than parsing the keys and building the data structures in the Interpolator itself. 

Fixes #9693

PR #10920, opened 2016-12-23
Fix string representation of sets during interpolation

### LINKED ISSUE #10908: aws_network_interface.private_ips empty string in 0.8.2
Hi there,

I upgraded to 0.8.2 and am having issues getting ip addresses off of `aws_network_interface`s.

### Terraform Version
0.8.2 (does not happen in 0.8.1)

### Affected Resource(s)
- aws_network_interface

The full config is here: https://gist.github.com/eric-johnson/e73255b833844fbb1e010daf1b5c107b
```hcl
...

resource "aws_network_interface" "test" {
  subnet_id = "${element(split(",", module.vpc.external_subnets), 0)}"
  subnet_id = "${aws_subnet.external.id}"
  security_groups = ["${aws_security_group.sg.id}"]
  tags { Name = "test-8.2" }
}

resource "null_resource" "test" {
  triggers {
    id = "${aws_network_interface.test.subnet_id}"
  }

  provisioner "local-exec" {
    command = "echo ${join("-", aws_network_interface.test.private_ips)} ${jsonencode(aws_network_interface.test.private_ips)}"
  }
}
```

### Output
https://gist.github.com/eric-johnson/e73255b833844fbb1e010daf1b5c107b
I have a debug log, but I'd rather not have to strip out sensitive info from it.

### Expected Behavior
The local-exec provisioner should be printing out the private_ips of the network_interface.
[This line](https://gist.github.com/eric-johnson/e73255b833844fbb1e010daf1b5c107b#file-output-txt-L45) should have the private ip address in it.

### Actual Behavior
The network interface's private_ips was a list with an empty string.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
This does not occur in terraform 0.8.1

2rs2ts:
The change in #10787 used `flatmap.Expand` to fix interpolation of nested maps, but it broke interpolation of sets such that their elements were not represented. For example, the expected string representation of a splatted `aws_network_interface.whatever.*.private_ips` should be:

```
[{Variable (TypeList): [{Variable (TypeString): 10.41.17.25}]} {Variable (TypeList): [{Variable (TypeString): 10.41.22.236}]}]
```

But instead it became:

```
[{Variable (TypeList): [{Variable (TypeString): }]} {Variable (TypeList): [{Variable (TypeString): }]}]
```

This is because the `expandArray` function of expand.go treated arrays to exclusively be lists, e.g. not sets. The old code used to match for numeric keys, so it would work for sets, whereas `expandArray` just assumed keys started at 0 and ascended incrementally. Remember that sets' keys are numeric, but since they are hashes, they can be any integer. The result of assuming that the keys start at 0 led to the recursive call to `flatmap.Expand` not matching any keys of the set, and returning `nil`, which is why the above example has nothing where the IP addresses used to be.

So we bring back that matching behavior, but we move it to `expandArray` instead. We've modified it to not reconstruct the data structures like it used to when it was in the `Interpolator`, and to use the standard int sorter rather than implementing a custom sorter since a custom one is no longer necessary thanks to the use of `flatmap.Expand`.

Fixes #10908, and restores the viability of the workaround I posted in #8696.

Big thanks to @jszwedko for helping me with this fix. I was able to diagnose the problem alone, but couldn't fix it without his help.

### DIFF
--- a/flatmap/expand.go
+++ b/flatmap/expand.go
@@ -2,6 +2,8 @@ package flatmap
 
 import (
 	"fmt"
+	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 )
@@ -42,9 +44,27 @@ func expandArray(m map[string]string, prefix string) []interface{} {
 		panic(err)
 	}
 
+	keySet := make(map[int]struct{})
+	listElementKey := regexp.MustCompile("^" + prefix + "\\.([0-9]+)(?:\\..*)?$")
+	for key := range m {
+		if matches := listElementKey.FindStringSubmatch(key); matches != nil {
+			k, err := strconv.ParseInt(matches[1], 0, 0)
+			if err != nil {
+				panic(err)
+			}
+			keySet[int(k)] = struct{}{}
+		}
+	}
+
+	keysList := make([]int, 0, num)
+	for key := range keySet {
+		keysList = append(keysList, key)
+	}
+	sort.Ints(keysList)
+
 	result := make([]interface{}, num)
-	for i := 0; i < int(num); i++ {
-		result[i] = Expand(m, fmt.Sprintf("%s.%d", prefix, i))
+	for i, key := range keysList {
+		result[i] = Expand(m, fmt.Sprintf("%s.%d", prefix, key))
 	}
 
 	return result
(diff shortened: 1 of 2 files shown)


jbardin:
Hi @2rs2ts,

Thanks for the fix on this. I made a few changes and added another test, so I moved your commit over to #11042. 

I'll close this and we can merge from there.

2rs2ts:
Cool, I dig it.

PR #10922, opened 2016-12-25
provider/postgresql: Quote connection string parameters

sean-:
Build upon what @idubinskiy correctly identified and properly escape DSN parameters when connecting to PostgreSQL.

CC @stack72 

Fixes: #10915


idubinskiy:
Thanks!

PR #10934, opened 2016-12-28
core: stoppable provisioners, helper/schema for provisioners

mitchellh:
⚠️  For **Terraform 0.9** (Do not merge until master is 0.9) ⚠️ 

**Fast provisioner cancellation!** 💃 

asdf

This PR includes two things:

  * Create a new Stop API for provisioners that is called when terraform's Stop API (triggered by things like interrupts) is called.

  * Introduce a `helper/schema` framework for provisioners. This makes it easy to consume the Stop API also introduced in this PR. 

This PR uses `context` heavily to make it easy to consume the cancellation, and helper/schema handles managing this context. Terraform's core stop mechanism was also changed to use `context`.

### DIFF
--- a/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/builtin/provisioners/local-exec/resource_provisioner.go
@@ -68,8 +68,24 @@ func applyFn(ctx context.Context) error {
 		"Executing: %s %s \"%s\"",
 		shell, flag, command))
 
-	// Run the command to completion
-	err := cmd.Run()
+	// Start the command
+	err := cmd.Start()
+	if err == nil {
+		// Wait for the command to complete in a goroutine
+		doneCh := make(chan struct{})
+		go func() {
+			defer close(doneCh)
+			err = cmd.Wait()
+		}()
+
+		// Wait for the command to finish or for us to be interrupted
+		select {
+		case <-doneCh:
+		case <-ctx.Done():
+			cmd.Process.Kill()
+			err = cmd.Wait()
+		}
+	}
 
 	// Close the write-end of the pipe so that the goroutine mirroring output
 	// ends properly.
--- a/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/builtin/provisioners/local-exec/resource_provisioner.go
@@ -72,15 +72,14 @@ func applyFn(ctx context.Context) error {
 	err := cmd.Start()
 	if err == nil {
 		// Wait for the command to complete in a goroutine
-		doneCh := make(chan struct{})
+		doneCh := make(chan error, 1)
 		go func() {
-			defer close(doneCh)
-			err = cmd.Wait()
+			doneCh <- cmd.Wait()
 		}()
 
 		// Wait for the command to finish or for us to be interrupted
 		select {
-		case <-doneCh:
+		case err = <-doneCh:
 		case <-ctx.Done():
 			cmd.Process.Kill()
 			err = cmd.Wait()
--- a/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/builtin/provisioners/local-exec/resource_provisioner.go
@@ -1,13 +1,14 @@
 package localexec
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os/exec"
 	"runtime"
 
 	"github.com/armon/circbuf"
-	"github.com/hashicorp/terraform/helper/config"
+	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/mitchellh/go-linereader"
 )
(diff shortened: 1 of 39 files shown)


helper/schema/schema.go
```diff
@@ -118,9 +118,16 @@ type Schema struct {
...
 	// cause instability.
-	Elem     interface{}
-	MaxItems int
-	MinItems int
+	//
+	// PromoteSingle, if true, will allow single elements to be standalone
+	// and promote them to a list. For example "foo" would be promoted to
+	// ["foo"] automatically. This is primarily for legacy reasons and the
+	// ambiguity is not recommended for new usage. Promotion is only allowed
+	// for primitive element types.
```
jbardin: Interesting idea, but I don't understand why we run into this. Did we allow defining single elements for a list in the config previously?

jbardin:
sorry, didn't mean to approve yet. still reviewing some, so please don't merge ...

terraform/context.go
```diff
@@ -634,30 +633,34 @@ func (c *Context) Refresh() (*State, error) {
 //
 // Stop will block until the task completes.
 func (c *Context) Stop() {
+	log.Printf("[WARN] terraform: Stop called, initiating interrupt sequence")
+
 	c.l.Lock()
```
jbardin: there's no special-case lock handling here, so Unlock should be defered.

terraform/context.go
```diff
@@ -634,30 +633,34 @@ func (c *Context) Refresh() (*State, error) {
...
+		log.Printf("[WARN] terraform: run context exists, stopping")
+
+		// Tell the hook we want to stop
+		c.sh.Stop()
 
-	// Tell the hook we want to stop
-	c.sh.Stop()
+		// Stop the context
+		c.runContextCancel()
+		c.runContextCancel = nil
```
jbardin: Do we need to set this to nil? This one is OK right now I think, but sets the code of for races in the future,

terraform/context.go
```diff
@@ -892,7 +906,15 @@ func (c *Context) walk(
 	return walker, realErr
 }
 
-func (c *Context) watchStop(walker *ContextGraphWalker, stopCh, doneCh <-chan struct{}) {
+func (c *Context) watchStop(walker *ContextGraphWalker, doneCh <-chan struct{}) {
+	// Get the stop channel. runContext might be nil only during tests.
+	// If this is called during a proper run operation, this will never
+	// be nil.
+	var stopCh <-chan struct{}
+	if ctx := c.runContext; ctx != nil {
```
jbardin: This definitely races with `acquireRun`, which sets the context fields under a mutex. What we can do here is have `walk` get the `stopCh` and pass it in to avoid reading that field without the mutex:

    func (c *Context) watchStop(walker *ContextGraphWalker, stopCh, doneCh <-chan struct{})

jbardin: I'm not a fan of special casing this during tests either; can we ensure that it's set for tests too and get rid of these nil checks?

helper/schema/schema.go
VladRassokhin: That's the case of `inline` parameter of 'remote-exec' provisioner, see what is removed in `builtin/provisioners/remote-exec/resource_provisioner.go`

mitchellh: Yeah, backwards compatibility here is the reason this was necessary.

terraform/context.go
mitchellh: We have to to avoid double-cancels. I'm not sure if that actually has an effect on context though. See Stop()

terraform/context.go
mitchellh: Its safe to access runContext because an entire Run is wrapped in a lock (acquireRun acquires a lock via cond vars). Fixing the special case though!

terraform/context.go
mitchellh: Done!

mitchellh:
Addressed feedback. Can you check the latest commits? 

All the `run*` unexported vars are safe to access without a lock since `acquireRun`/`releaseRun` protect access. Even if `Stop` is called, another operation can't start until `releaseRun` completes which requires that watchStop/walk and friends aren't running anymore.

terraform/context.go
jbardin: yes, but `watchStop` doesn't take a lock, and can race with any method writing to the context fields. Just verified this shows up readily with the race detector.

This should fix it by removing the read of that field from watchStop entirely:

```
@@ -811,7 +811,9 @@ func (c *Context) walk(

        // Watch for a stop so we can call the provider Stop() API.
        doneCh := make(chan struct{})
-       go c.watchStop(walker, doneCh)
+       stopCh := c.runContext.Done()
+
+       go c.watchStop(walker, stopCh, doneCh)

        // Walk the real graph, this will block until it completes
        realErr := graph.Walk(walker)
@@ -906,12 +908,7 @@ func (c *Context) walk(
        return walker, realErr
 }

-func (c *Context) watchStop(walker *ContextGraphWalker, doneCh <-chan struct{}) {
-       // Get the stop channel. runContext will never be nil since this should
-       // only be called within the context of an operation started with
-       // acquireRun
-       stopCh := c.runContext.Done()
-
+func (c *Context) watchStop(walker *ContextGraphWalker, stopCh, doneCh <-chan struct{}) {
        // Wait for a stop or completion
        select {
        case <-stopCh:
```

builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -62,8 +68,24 @@ func (p *ResourceProvisioner) Apply(
...
+			defer close(doneCh)
+			err = cmd.Wait()
+		}()
+
+		// Wait for the command to finish or for us to be interrupted
+		select {
+		case <-doneCh:
+		case <-ctx.Done():
+			cmd.Process.Kill()
+			err = cmd.Wait()
```
jbardin: this `err` assignment races with the one in the goroutine above

mitchellh:
All current feedback addressed. Please review again. :)

PR #10959, opened 2016-12-29
provider/openstack: Updating FWaaS and LBaaS v2 Acceptance Tests

jtopjian:
Part 3 of 5. I'd submit all, but I'm rebasing and retesting them as I go just to make sure everything is good.


stack72:
LGTM! Thanks @jtopjian :)

PR #10960, opened 2016-12-29
provider/aws: handle aws_lambda_function missing s3 key error

kung-foo:
This PR traps `S3 Error Code: NoSuchKey` as a non-retryable error when creating a new lambda function. Otherwise if the S3 key doesn't exist, then the function just loops for 10 minutes. I don't think this is the intended behavior.

The original code seemed to treat all `InvalidParameterValueException` as retryable, but the comment above it indicates that that was only for an IAM role propagation issue. So I've also added an explicit check for the IAM error.

I could also just remove the S3 check now that I have the explicit IAM check.

Side note, this came up because I was using `aws_s3_bucket_object` and _that_ check was showing the object existed. `aws_s3_bucket_object` seems to allow a leading `/` whereas `aws_lambda_function.s3_key` does not.

-- jonathan

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -274,12 +274,8 @@ func resourceAwsLambdaFunctionCreate(d *schema.ResourceData, meta interface{}) e
 		if err != nil {
 			log.Printf("[DEBUG] Error creating Lambda Function: %s", err)
 
-			if isAWSErr(err, "InvalidParameterValueException", "S3 Error Code: NoSuchKey") {
-				return resource.NonRetryableError(err)
-			}
-
 			if isAWSErr(err, "InvalidParameterValueException", "The role defined for the function cannot be assumed by Lambda") {
-				log.Printf("[ERROR] Received %q, retrying CreateFunction", err)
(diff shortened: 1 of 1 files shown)


bclodius:
@kung-foo @apparentlymart any updates on this?  I just faced this same issue scratching my head why my lambda functions were taking a while to create.  In my case it was for ```InvalidParameterValueException: The role defined for the function cannot be assumed by Lambda.``` which is included in this PR.

radeksimko:
Hi @kung-foo 
sorry to keep you waiting here and thank you for the valuable contribution.

Since we're retrying on the error we tend to avoid `ERROR` log severity to avoid the feeling that something is wrong, when it's something that happens nearly every time during the creation of Lambda function.

Secondly I believe there's no need to treat `S3 Error Code: NoSuchKey` as a special case - if the error doesn't match, it will fall back to `NonRetryableError` anyway.

Finally there's another error which we need to retry on - `InvalidParameterValueException: The provided execution role does not have permissions to call DescribeNetworkInterfaces on EC2` - that occurs if you try to create Lambda function in a VPC.

I made the mentioned tweaks in b8a7774 to speed things up. I hope you don't mind.

Your contribution is still included (and still credited to you), with the appropriate modifications. Feel free to ask about any of the changes.

Issue #10977, opened 2016-12-31
Empty interpolation of schema.Set values


vancluever:
I didn't see anything in the TF changelog that looked inparticularly damning on this one, but this definitely seems like a regression between v0.8.1. and v0.8.2:

## Code

```
data "aws_alb" "autoscaling_alb" {
  arn = "arn.goes.here"
}

output "foo" {
  value = "${data.aws_alb.autoscaling_alb.security_groups}"
}

```

## v0.8.1

```
Outputs:

foo = [
    sg-abcdef12
]
```

## v0.8.2

```
Outputs:

foo = [
    
]
```

## Other Curiosities

The lookup result is not a bad index, but a literally blank string. If you do this on sets that contain multiple items (the ALB I was looking at only had 1 security group, you actually get something like:

```
foo = [
    ,
    
]
```

What does work is looking up the set item directly via its hash. So if you were to do something like `value = "${data.aws_alb.autoscaling_alb.security_groups.123456789}"`, that would work.

Going to keep digging and see what I can find.

vancluever:
Think I found it:

https://github.com/hashicorp/terraform/blob/master/flatmap/expand.go#L39

`expandArray` (which complex value interpolation was switched to a few weeks ago on part of that functionality moving to `flatmap.Expand`) does not take into account list structures that are actually of type `schema.Set`, because it bases its index lookups off a sequence of `0` - `len-1` of the attribute. Since the "keys" for the set are actually hashed, you can be pretty much always sure that you will not get a proper lookup here.

vancluever:
I think I got a fix in for this! Commit references back to this issue

mitchellh:
Will be fixed by #11042

PR #10978, opened 2016-12-31
core: Fix interpolation on schema.Sets

vancluever:
A change introduced in v0.8.2 simplified how interpolations on maps and
lists worked, moving this functionality to `flatmap.Expand`. However, it
seems that the new functionality did not take into account that lists
come from sets as well, where the "indexes", which are more like keys,
and being hashed, do not have a deterministic range bound to the length
of the list. This caused issues when interpolating sets where the
resulting list was almost always likely returned as a set of blank
strings.

This commit takes some of the old, pre-v0.8.2 functionality, and rolls
it into `flatmap.expandArray`. Rather than use a custom sort
implementation, we just derive all numeric indexes from the current
resource and build a sorted, de-duplicated (in the case of subkeys) list
that is then used with the actual length of the list to build the proper
interpolated set.

Fixes hashicorp/terraform#10977.

### DIFF
--- a/flatmap/expand.go
+++ b/flatmap/expand.go
@@ -2,6 +2,8 @@ package flatmap
 
 import (
 	"fmt"
+	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


mitchellh:
This is all linked properly to the open PR so I'm going to close it. We may use details from it if we find your work was an improvement. Sorry we missed this completely before starting work on our fix! Thanks for your help.

vancluever:
Fair enough @mitchellh, thanks for the response!

PR #10979, opened 2016-12-31
provider/openstack: Updating LBaaSV1 acceptance tests

jtopjian:
Part 5/5!

### DIFF
--- a/builtin/providers/openstack/resource_openstack_lb_member_v1_test.go
+++ b/builtin/providers/openstack/resource_openstack_lb_member_v1_test.go
@@ -20,7 +20,7 @@ func TestAccLBV1Member_basic(t *testing.T) {
 			resource.TestStep{
 				Config: testAccLBV1Member_basic,
 				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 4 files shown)


stack72:
Thanks for this again @jtopjian :)

I'd like to start a conversation on how we can start running the openstack acceptance tests as part of our nightly builds. We should chat on slack about this

LGTM

Paul

jtopjian:
@stack72 Thanks! I've been tied up with work, but I'll try to ping you on Slack soon.

I've been thinking about this as well. I used to use devstack to run the acceptance tests, but I've recently switched to an all-in-one [Packstack](https://www.rdoproject.org/install/quickstart/) deployment. devstack became too buggy in some areas and I found myself spending too much time fiddling with just getting it up and running. The other downside to devstack is that it does not provide any type of service init scripts, so the environment could never be rebooted -- this nixed any idea of using it as a base image for a quick, automated acceptance test environment.

Packstack deploys a functional all-in-one OpenStack environment _with_ init scripts. I need to dig into it a bit more and get it working without hard-coded interface IPs, but if that's possible, it could be rolled into an image. :)

PR #10987, opened 2017-01-01
aws_key_pair: Ensure key_name attribute is set

### LINKED ISSUE #10983: AWS: key_pair name sometimes not resolvable before reference
~~Found another issue caused by AWS' eventual consistency, this time around ssh key pair creation.~~

If I declare a resource such as the following:

```
resource "aws_key_pair" "hsimpson" {
    key_name_prefix = "hsimpson"
    public_key = "${file("/Users/hsimpson/.ssh/id_rsa.pub")}"
}
```

Then subsequently refer to it:

```
resource "aws_instance" "test_instance" {
   key_name = "${aws_key_pair.hsimpson.key_name}"
   # etc
}
```

It's quite possible the key name will not yet be known by the time the instance provisioner runs:

```
Error applying plan:

1 error(s) occurred:

* Resource 'aws_key_pair.hsimpson' does not have attribute 'key_name' for variable 'aws_key_pair.hsimpson.key_name'
```

~~As with other objects, we'll need to poll for a reasonable amount of time until the name can be determined.~~

mfischer-zd:
Ensure that the `key_name` attribute is available to `aws_key_pair`
resource dependents, even when the attribute is not specifically
set (i.e., when `key_name_prefix` or automatic naming is performed).

Fixes #10983.

### DIFF
--- a/builtin/providers/aws/resource_aws_key_pair.go
+++ b/builtin/providers/aws/resource_aws_key_pair.go
@@ -75,8 +75,10 @@ func resourceAwsKeyPairCreate(d *schema.ResourceData, meta interface{}) error {
 		keyName = v.(string)
 	} else if v, ok := d.GetOk("key_name_prefix"); ok {
 		keyName = resource.PrefixedUniqueId(v.(string))
+		d.Set("key_name", keyName)
 	} else {
 		keyName = resource.UniqueId()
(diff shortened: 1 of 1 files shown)


catsby:
Wonderful, thanks!

PR #10991, opened 2017-01-02
provider/azurerm: avoid registering providers already registered, add option to disable

### LINKED ISSUE #8824: Update azurerm provider  to not require provider resource registration privileges
### Actual Behavior

Using the **azurerm** provider **registerAzureResourceProvidersWithSubscription** is called. This requires the account in use to have access to register resources w/ a subscription. 
### Expected Behavior

The call being used could be replaced w/ https://msdn.microsoft.com/en-us/library/azure/dn790534.aspx that displays resource registration state without requiring the user to have access to change it. 
### Terraform Version

0.7.3  
### Related Commit

https://github.com/hashicorp/terraform/commit/c90718d1acbdc9ba66dfe0c88158245e3f44cff9

### LINKED ISSUE #10931: AzureRM registers for Resource Providers unnecessarily
When replying to this issue (https://github.com/Azure/azure-rest-api-specs/issues/836), I remembered an issue with how AzureRM operates.

**Scenario:**

1. I (an Admin on the Subscription) pre-create resource group `blah`.

2. I create an SP and grant it Contributor level access to `blah`.

3. I run Terraform with those credentials, only asking it to create resources in the resource group that the SP has access to.

This fails today, because every call tries to register the Resource Provider, even if it's already registered and this registration action is at the Subscription scope.

**Suggested Fix:** 

If possible, interogate ARM to see if registration is really needed up front.

Also, the error message in this case is unhelpful. It just looks like the generic "auth" error from ARM, which is unfortunate because people do struggle with getting the SP working properly and they might not realize the problem. It actually took a couple weeks of sitting in the back of my brain before I realized why this hadn't been working.

cc: @tombuildsstuff

pmcatominey:
Fixes #10931.
Fixes #8824.

The provider now consults the Providers API to detect which providers are already
registered and uses this call to test the credentials

A new option `skip_provider_registration` / `ARM_SKIP_PROVIDER_REGISTRATION` is
now available to opt out of provider registration entirely.

The error message from the call used to test credentials has also been changed, this will give us better errors once Azure/go-autorest#101 makes it's way into the SDK. Example of current errors:

```bash
# Bad Client ID
* Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: autorest#WithErrorUnlessStatusCode: POST https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/token?api-version=1.0 failed with 400 Bad Request: StatusCode=400

# Bad Client Secret
* Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: autorest#WithErrorUnlessStatusCode: POST https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/token?api-version=1.0 failed with 401 Unauthorized: StatusCode=401

# Bad Subscription ID
* Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: resources.ProvidersClient#List: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="SubscriptionNotFound" Message="The subscription '00000000-0000-0000-0000-000000000000' could not be found."

# Bad Tenant ID
* Unable to list provider registration status, it is possible that this is due to invalid credentials or the service principal does not have permission to use the Resource Manager API, Azure error: autorest#WithErrorUnlessStatusCode: POST https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/token?api-version=1.0 failed with 400 Bad Request: StatusCode=400
```

### DIFF
--- a/builtin/providers/azurerm/config.go
+++ b/builtin/providers/azurerm/config.go
@@ -144,15 +144,6 @@ func (c *Config) getArmClient() (*ArmClient, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Error creating Riviera client: %s", err)
 	}
-
-	// validate that the credentials are correct using Riviera. Note that this must be
-	// done _before_ using the Microsoft SDK, because Riviera handles errors. Using a
-	// namespace registration instead of a simple OAuth token refresh guarantees that
-	// service delegation is correct. This has the effect of registering Microsoft.Compute
-	// which is neccessary anyway.
-	if err := registerProviderWithSubscription("Microsoft.Compute", rivieraClient); err != nil {
-		return nil, err
-	}
 	client.rivieraClient = rivieraClient
 
 	oauthConfig, err := azure.PublicCloud.OAuthConfigForTenant(c.TenantID)
--- a/builtin/providers/azurerm/provider.go
+++ b/builtin/providers/azurerm/provider.go
@@ -7,6 +7,9 @@ import (
 
 	"sync"
 
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/arm/resources/resources"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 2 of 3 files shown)


jen20:
This looks like a good idea to me. :+1: for moving to the official SDK where practical - the current solution was always a hack.

jen20:
Thanks @pmcatominey! I've resolved the conflicts and merged.

PR #10993, opened 2017-01-02
provider/aws: Bump SDK version to 1.6.9

stack72:
There is 1 breaking change. The OpsWorks Update API no longer supports
`tenancy` as an update. So that has to be marked as ForceNew: true

```
% make testacc TEST=./builtin/providers/aws
% TESTARGS='-run=TestAccAWSOpsworksInstance'
% 2 ↵ ✹ ✭
2017/01/02 17:11:29 Generated command/internal_plugin_list.go
-run=TestAccAWSOpsworksInstance -timeout 120m
=== RUN   TestAccAWSOpsworksInstance
--- PASS: TestAccAWSOpsworksInstance (99.89s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws99.911s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_instance.go
+++ b/builtin/providers/aws/resource_aws_opsworks_instance.go
@@ -280,6 +280,7 @@ func resourceAwsOpsworksInstance() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Computed:     true,
+				ForceNew:     true,
 				ValidateFunc: validateTenancy,
 			},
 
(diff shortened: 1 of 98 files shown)


jen20:
LGTM if this compiles, but the change isn't mentioned in the library release notes, so this might be worth opening an issue on the SDK repo.

stack72:
AFAICT, there is no update supported by the CLI either. I think the struct just allowed the value to be passed and was an unsupported operation

There were no tests in place for tenancy updates  and it was only added in this same release so this is a safe change :)

PR #10985, opened 2017-01-01
provider/aws: implement CloudFront Lambda Function Associations

MrGossett:
Implements [Lambda Function Associations](http://docs.aws.amazon.com/AmazonCloudFront/latest/APIReference/API_LambdaFunctionAssociations.html) for CloudFront cache behaviors.

Within a `cache_behavior` or `default_cache_behavior` block:

```
lambda_function_association {
  event_type = "viewer-request"
  lambda_arn = "arn:aws:lambda:us-east-1:${data.aws_caller_identity.current.account_id}:some-function:$LATEST"
}
```

should fix #10938 

My first contribution to Terraform. How'd I do?

PR #10997, opened 2017-01-03
provider/aws: fix expectation messages that did not align with code

MrGossett:
While working on #10985, I noticed that there were a few test failure messages that didn't match up with the assertions the code was actually performing. After some `sed` and `grep` tomfoolery, I'm pretty sure these three locations were the only places this kind of mismatch existed.


jen20:
Thanks @MrGossett!

PR #11013, opened 2017-01-03
Read update_strategy before overwriting it.

paddyforan:
As brought up in #10174, our update_strategy property for instance group
managers in GCP would always be set to "RESTART" on read, even if the
user asked for them to be "NONE" in the config.

This adds a test to ensure that the user wishes were respected, which
fails until we check for update_strategy in the ResourceData before we
update it within the Read function. Because the update_strategy property
doesn't map to anything in the API, we never need to read it from
anywhere but the config, which means the ResourceData should be
considered authoritative by the time we get to the Read function.

The fix for this was provided by @JDiPierro in #10198 originally, but
was missing tests, so it got squashed into this.


stack72:
LGTM! Thanks @paddyforan :)

PR #11015, opened 2017-01-03
No 'bucket' attribute exported for aws_s3_bucket resource

huyanhvn:


stack72:
Hi @huyanhvn 

Thanks for the PR here - in this specific case, `bucket` is indeed an exported attribute. Anything that can be passed to the resource is also available as exported

This isn't currently something that the documentation explains very well - we are going to be looking at our documentation shortly and will make sure that this is actually the case

Thanks - for taking the time to open this PR - in this case, we can merge this. ID still works as expected :)

Paul

PR #11018, opened 2017-01-04
[GKE] Allow additional zones to be configured

### LINKED ISSUE #6397: Feature request: additional-zones on google_container_cluster resource
https://cloud.google.com/sdk/gcloud/reference/container/clusters/create

> --additional-zones ZONE,[ZONE,…]
> The set of additional zones in which the specified node footprint should be replicated. All zones must be in the same region as the cluster's primary zone. If additional-zones is not specified, all nodes will be in the cluster's primary zone.
> Multiple locations can be specified, separated by commas. For example:
> 
>  $ {command} example-cluster --zone us-central1-a \ 
>      --additional-zones us-central1-b,us-central1-c

unguiculus:
Fixes #6397.


builtin/providers/google/resource_container_cluster.go
```diff
@@ -282,6 +288,24 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.InitialClusterVersion = v.(string)
 	}
 
+	if v, ok := d.GetOk("additional_zones"); ok {
```
danawillow: Can you add support to the Update method as well?

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -26,6 +26,23 @@ func TestAccContainerCluster_basic(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckContainerClusterDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccContainerCluster_withAdditionalZones,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckContainerClusterExists(
+						"google_container_cluster.with_additional_zones"),
```
danawillow: I'd love to see a check that confirms that the cluster not only exists but has the additional zones.

builtin/providers/google/resource_container_cluster.go
unguiculus: Is there a way to update this without recreating the whole cluster? I did not find any. I guess I should set `ForeNew: True` then. Currently, the only argument that can be updated is `node_version`.

builtin/providers/google/resource_container_cluster_test.go
unguiculus: OK, I'll have a look at this. I basically just looked at what the other tests do and followed that principle.

builtin/providers/google/resource_container_cluster.go
danawillow: Ah, you're right. The API docs claim you can update locations but the client library hasn't been updated with that change yet (hopefully that'll get fixed soon, I've been doing a fair amount of poking but it's apparently complicated). `ForceNew: True` sounds good then.

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -143,6 +163,29 @@ func testAccCheckContainerClusterExists(n string) resource.TestCheckFunc {
...
+
+		var (
+			additionalZonesSize int
+			err                 error
+		)
+
+		if additionalZonesSize, err = strconv.Atoi(rs.Primary.Attributes["additional_zones.#"]); err != nil {
+			return err
+		}
+		if additionalZonesSize < 2 {
```
danawillow: != 2?

builtin/providers/google/resource_container_cluster_test.go
danawillow: Thanks! Yeah, I'm trying to make sure that new tests that get added are more comprehensive than what we currently have, and that they're testing the right things.

builtin/providers/google/resource_container_cluster_test.go
unguiculus: Fixed.

builtin/providers/google/resource_container_cluster.go
unguiculus: Added `ForceNew: True`.

danawillow:
Looks good, thanks @unguiculus!

PR #11019, opened 2017-01-04
Rancher stack upgrade fails

raphink:
Terraform 0.8.2 fails to upgrade Rancher stacks when it meets the `finishing-upgrade` state:

```
* rancher_stack.conplicity: Error waiting for stack (1e18) to be upgraded: unexpected state 'finishing-upgrade', wanted target 'active'. last error: %!s(<nil>)
```


stack72:
Thanks for this @raphink :)

P.

PR #11020, opened 2017-01-04
provider/google: remote_traffic_selector for google_compute_vpn_tunnel

drebes:
The google_compute_vpn_tunnel resource was missing support for the RemoteTrafficSelector parameter. As such, establishment of route based VPN tunnels fails due to the inability to negotiate a child SA. This should fix that.

### DIFF
--- a/builtin/providers/google/resource_compute_vpn_tunnel.go
+++ b/builtin/providers/google/resource_compute_vpn_tunnel.go
@@ -199,6 +199,18 @@ func resourceComputeVpnTunnelRead(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Error Reading VPN Tunnel %s: %s", name, err)
 	}
 
+	localTrafficSelectors := []string{}
+	for _, lts := range vpnTunnel.LocalTrafficSelector {
+		localTrafficSelectors = append(localTrafficSelectors, lts)
+	}
+	d.Set("local_traffic_selector", localTrafficSelectors)
+
+	remoteTrafficSelectors := []string{}
+	for _, rts := range vpnTunnel.RemoteTrafficSelector {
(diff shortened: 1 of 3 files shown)


drebes:
Any chance to get this reviewed? I'm working on something else that would have this a requirement.

builtin/providers/google/resource_compute_vpn_tunnel.go
```diff
@@ -72,6 +72,14 @@ func resourceComputeVpnTunnel() *schema.Resource {
 				Set:      schema.HashString,
 			},
 
+			"remote_traffic_selector": &schema.Schema{
+				Type:     schema.TypeSet,
```
danawillow: Why a Set instead of a List?

drebes: Three reasons:

-  local_traffic_selector also uses a set (this by itself is not an excuse).
- Cloud VPN doesn't care about the order of the traffic selectors, so even if a remote side has the local traffic selector configured in the specific order as "A, B", and on Cloud VPN side the remote traffic selector has been configured as "B, A", Cloud VPN will still successfully negotiate the child SA.
- In general, the main reason to use a set instead of a list in terraform is that if the user changes the definition from ["A", "B"] to ["B", "A"], terraform does not consider this as a delta and will not recreate the resource (or update, for resources that support that). 

I'll update the PR with some tests.

drebes:
Tests added.

builtin/providers/google/resource_compute_vpn_tunnel.go
drebes: I've updated the commit with (local and remote) selector tests.

danawillow:
LGTM, thanks!

PR #11021, opened 2017-01-04
provider/aws: aws_kinesis_firehose_delivery_stream code cleanup

stack72:
We `removed` these items in 0.7 and they can be deleted from 0.8 IMO

//cc @catsby


catsby:
Get out of here, dead code 
hulk_bear 2

PR #11023, opened 2017-01-04
provider/aws: Updating the aws_efs_mount_target dns_name

stack72:
Fixes:#10902

AWS introduced a change to the Mount Target DNS Name to remove the
availability_zone from it -
https://aws.amazon.com/about-aws/whats-new/2016/12/simplified-mounting-of-amazon-efs-file-systems/

This was because there used to be a limit of 1 mount target per AZ -
this has been raised.

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSEFSMountTarget_'
2017/01/04 10:45:35 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEFSMountTarget_ -timeout 120m
=== RUN   TestAccAWSEFSMountTarget_importBasic
--- PASS: TestAccAWSEFSMountTarget_importBasic (236.19s)
=== RUN   TestAccAWSEFSMountTarget_basic
--- PASS: TestAccAWSEFSMountTarget_basic (445.52s)
=== RUN   TestAccAWSEFSMountTarget_disappears
--- PASS: TestAccAWSEFSMountTarget_disappears (228.31s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	910.044s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_efs_mount_target.go
+++ b/builtin/providers/aws/resource_aws_efs_mount_target.go
@@ -200,13 +200,13 @@ func resourceAwsEfsMountTargetRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// DNS name per http://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-dns-name.html
-	az, err := getAzFromSubnetId(*mt.SubnetId, meta.(*AWSClient).ec2conn)
+	_, err = getAzFromSubnetId(*mt.SubnetId, meta.(*AWSClient).ec2conn)
 	if err != nil {
 		return fmt.Errorf("Failed getting Availability Zone from subnet ID (%s): %s", *mt.SubnetId, err)
 	}
 
 	region := meta.(*AWSClient).region
-	err = d.Set("dns_name", resourceAwsEfsMountTargetDnsName(az, *mt.FileSystemId, region))
+	err = d.Set("dns_name", resourceAwsEfsMountTargetDnsName(*mt.FileSystemId, region))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


catsby:
I'm ( 🤔 ), should this region-only DNS Name be a separate computed attribute? Is the AZ specific one still valid, or is it deprecated? I'm thinking that anyone referencing `efs_mount_target.mount.dns_name` would see a downstream change because of this

catsby:
so, I re-read #10902 and it seems the AZ is just dropped on AWS side? Then this LGTM  👍

stack72:
Thanks @catsby :) yes, they dropped it on their side

builtin/providers/aws/resource_aws_efs_mount_target.go
```diff
@@ -200,13 +200,13 @@ func resourceAwsEfsMountTargetRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// DNS name per http://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-dns-name.html
-	az, err := getAzFromSubnetId(*mt.SubnetId, meta.(*AWSClient).ec2conn)
+	_, err = getAzFromSubnetId(*mt.SubnetId, meta.(*AWSClient).ec2conn)
```
tjwallace: Is there any point of calling `getAzFromSubnetId` if the return value isn't used?

tjwallace:
Could the `dns_name` attribute also be added to the `aws_efs_file_system` resource now, since the value of `efs_mount_target.dns_name` will be the same for the same `file_system_id`?

PR #11050, opened 2017-01-05
provider/aws: aws_db_instance restored from snapshot had problem with subnet_group

### LINKED ISSUE #11024: Creating RDS instance from snapshot errors - InvalidVPCNetworkStateFault: You cannot move DB instance _rds_instance_name_ to subnet group
Hi,

I'm creating an issue off the back of a thread on the following PR - https://github.com/hashicorp/terraform/pull/10818

In there I've found that I can successfully create a RDS from snapshot on using 0.8.0 but cannot on 0.8.3 (yet to personally test .1 and .2)

The error message I get looks like the AWS api returning an error on a modify operation :

`Error applying plan:`
`1 error(s) occurred:`

`* aws_db_instance._resource_name_: Error modifying DB Instance _rds_instance_name_: InvalidVPCNetworkStateFault: You cannot move DB instance _rds_instance_name_ to subnet group _subnet_group_name_. The specified DB subnet group and DB instance are in the same VPC. Choose a DB subnet group in different VPC than the specified DB instance and try again. status code: 400, request id: 7f594101-cced-11e6-b996-2141c7b3473f`

`Terraform does not automatically rollback in the face of errors. Instead, your Terraform state file has been partially updated with any resources that successfully completed. Please address the error above and apply again to incrementally change your infrastructure.`

Terraform gets passed in the correct values for security group id and subnet group etc, creates it but bombs out when the rds instance goes into a modifying state.

I believe @stack72 is already looking into this, but I'm raising the issue here so others can see a fix is being looked into.

stack72:
Fixes #11024

A change was introduced in 0.8.2 that allows db_instances to change
their db_subnet_group. Unfortunately, this caused an issue for
db_instances that were being restored from snapshot. The restore from
snapshot part of create calls the Update func whereas a normal create
calls the Read func

When calling the Update func, the db_instance was trying to go through a
db_subnet_group_name change and was failing for the following reason:

```
InvalidVPCNetworkStateFault: You cannot move DB instance _rds_instance_name_ to subnet group _subnet_group_name_. The specified DB subnet group and DB instance are in the same VPC. Choose a DB subnet group in different VPC than the specified DB instance and try again.
```

### DIFF
--- a/builtin/providers/aws/resource_aws_db_instance.go
+++ b/builtin/providers/aws/resource_aws_db_instance.go
@@ -941,7 +941,7 @@ func resourceAwsDbInstanceUpdate(d *schema.ResourceData, meta interface{}) error
 		req.DBPortNumber = aws.Int64(int64(d.Get("port").(int)))
 		requestUpdate = true
 	}
-	if d.HasChange("db_subnet_group_name") {
+	if d.HasChange("db_subnet_group_name") && !d.IsNewResource() {
 		d.SetPartial("db_subnet_group_name")
 		req.DBSubnetGroupName = aws.String(d.Get("db_subnet_group_name").(string))
 		requestUpdate = true


edmundcraske:
What needs to happen to get this merged? :)

paddyforan:
(Update: we have none, because they would take hours to run. Fair enough.)

paddyforan:
LGTM.

PR #11075, opened 2017-01-06
Updated the instructions to install Terraform

aifred:
The string value within the PATH environment variable should be the directory that contains the binary file and not include the filename.


stack72:
Hi @aifred 

Thanks for this - this is actually just a hangup from when we had multiple binaries. Thanks for fixing it

Paul

PR #11076, opened 2017-01-06
Update cloudwatch_metric_alarm.html.markdown

csukcc:
Added the missing `=` in the "Example in Conjunction with Scaling Policies" script


stack72:
Hi @csukcc 

Thanks for the PR here - this isn't actually a requirement so I don't think we should merge this

Please let me know if you feel this is an issue

Thanks

Paul

PR #11079, opened 2017-01-06
Remove key.name from example, deprecated since 0.7

gerr1t:


stack72:
Hi @gerr1t 

Thanks for noticing this here :) You are correct and this LGTM

Paul

Issue #10505, opened 2016-12-02
Crash when using plan with vault provider


fcantournet:
### Terraform Version
Terraform v0.8.0-rc1 (aac47ecf8421da8573155951e36ce5c1aa3fd3a2)

### Affected Resource(s)
- vault_generic_secret

### Terraform Configuration Files
```hcl
provider "vault" {
	address = "https://127.0.0.1:8200"
}
data "vault_generic_secret" "example" {
  path = "secret/foo"
  data_json = <<EOT
{
 "foo":   "bar",
 "pizza": "cheese"
}
EOT
}
```

### Panic Output
https://gist.github.com/fcantournet/e63ee73bc926ee4a0ad147dbfd8de9f5

### Expected Behavior
terraform should have made a plan and reported that it was going to create missing secret
OR
yell at me because I screwed up

### Actual Behavior
Crash

### Steps to Reproduce
1. have a vault instance running on localhost with ssl
2. `terraform plan`

### Important Factoids
terraform prompted me for : 
- provider.vault.ca_cert_dir
- provider.vault.ca_cert_file

I answer with `enter` and I supposed it would default to vault behavior which is to use the system's CAs

apparentlymart:
```
2016/12/02 17:04:56 [DEBUG] plugin: terraform: panic: Unknown type for input: 5
2016/12/02 17:04:56 [DEBUG] plugin: terraform: 
2016/12/02 17:04:56 [DEBUG] plugin: terraform: goroutine 50 [running]:
2016/12/02 17:04:56 [DEBUG] plugin: terraform: panic(0x2cbf280, 0xc4201ef4f0)
2016/12/02 17:04:56 [DEBUG] plugin: terraform: 	/opt/go/src/runtime/panic.go:500 +0x1a1
2016/12/02 17:04:56 [DEBUG] plugin: terraform: github.com/hashicorp/terraform/helper/schema.schemaMap.Input(0xc4201a6d20, 0x50d1160, 0xc420394298, 0xc420384ab0, 0x301f900, 0xc420391a01, 0xc420394298)
2016/12/02 17:04:56 [DEBUG] plugin: terraform: 	/opt/gopath/src/github.com/hashicorp/terraform/helper/schema/schema.go:503 +0x420
[...]

PR #11082, opened 2017-01-06
provider/vault: Remove user input for optional vault provider fields

Mongey:
Hello again @apparentlymart 👋 

When `ca_cert_dir` and `ca_cert_file` are not set in the `vault` provider, Terraform will prompt the user for input. 

Looking at [`helper/schema/schema.go`](https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L501) it seems that a nil `DefaultValue` for a `TypeString` will ask the user for input. Defaulting to `""` will allow us to continue without user input.

It also seems like Terraform isn't handling `TypeList`, causing the crash described in #10505.


Mongey:
Adding a `WIP` tag, `ConflictsWith` always gets triggered with either `ca_cert_file` or `ca_cert_dir` set

Mongey:
It seems that the Vault Client allows `ca_cert_file` and `ca_cert_dir` to be set, so I've removed the `ConflictsWith` on these fields.

gerr1t:
+1 , thanks @Mongey

stack72:
LGTM! thanks for the work here :)

PR #11093, opened 2017-01-07
provider/aws require db option group name to be lower case

### LINKED ISSUE #11040: aws_db_option_group name is always lowercase in AWS
### Terraform Version
All

### Affected Resource(s)
- aws_db_option_group

```hcl
resource "aws_db_option_group" "aws_my_option_group" {
    name = "ABCDEF"
    option_group_description  = "ABCDEF"
    engine_name = "oracle-ee"
    major_engine_version = "11.2"
}
```

### Expected Behavior
The resource is successfully created and is successfully refreshed in subsequent runs

### Actual Behavior
The resource is successfull created but in subsequent runs it cannot be refreshed as the state
has recorded the uppercase containing name as the resource id but AWS has implicitly lowercased it and thus the code errors [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_option_group.go#L158)

### Steps to Reproduce
1. `terraform apply`
2. `terraform apply` again

### Important Factoids
We only encountered the issue with oracle option group but I do not think this is specific to oracle.

### Potential fixes
Validate that the group name is always lowercase [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_option_group.go#L347) is probably the cleanest because the non lowercase will be lost by AWS anyway ?

Alternatively the id could be [set to a lowercase version of the name](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_option_group.go#L124) but this might lead to unexpected behaviors in other situations ... ?

travertischio:
AWS is automatically changing these names to lower case as described in #11040 
This is causing errors when terraform apply is re-run.

I don't believe this needs any documentation updates.

Closes #11040

### DIFF
--- a/builtin/providers/aws/resource_aws_db_option_group.go
+++ b/builtin/providers/aws/resource_aws_db_option_group.go
@@ -354,6 +354,10 @@ func validateDbOptionGroupName(v interface{}, k string) (ws []string, errors []e
 		errors = append(errors, fmt.Errorf(
 			"only alphanumeric characters and hyphens allowed in %q", k))
 	}
+	if regexp.MustCompile(`[A-Z]`).MatchString(value) {
+		errors = append(errors, fmt.Errorf(
+			"only lowercase letters allowed in %q", k))
+	}
 	if regexp.MustCompile(`--`).MatchString(value) {
 		errors = append(errors, fmt.Errorf(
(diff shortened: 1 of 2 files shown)


travertischio:
This was addressed in #14192

PR #11097, opened 2017-01-08
Added example of variable reference inside file

arehmandev:
This tripped me up earlier as I was referencing inside the file as $${variable} when it should be ${variable}. Think this change to the website brings clarity, thanks!


stack72:
Thanks @arehmandev :) This LGTM! Sorry this tripped you up

Paul

Issue #11114, opened 2017-01-09
aws_route53_record: plan should catch wrong type value


jonatanblue:
### Terraform Version
`v0.8.2`

### Affected Resource
- aws_route53_record

### Terraform Configuration Files
```hcl
# modules/dns_alias/main.tf:
provider "aws" {
  access_key = "${var.DNS_AWS_ACCESS_KEY_ID}"
  secret_key = "${var.DNS_AWS_SECRET_ACCESS_KEY}"
  region = "eu-west-1"
}
resource "aws_route53_record" "alias" {
    count = "${var.tf_count}" # If zero won't create resource
    name = "${var.sub_domain}.example.com"
    zone_id = "ABCDEFGHIJKLMN"
    type = "alias"

    alias {
      name = "${var.alias}"
      zone_id = "${var.zone_id}"
      evaluate_target_health = true
    }
}

# main.tf:
...
module "dns" {
  source = "modules/dns_alias"
  sub_domain = "example1"
  alias = "${aws_alb.my_alb.dns_name}"
  zone_id = "${aws_alb.my_alb.zone_id}"
  DNS_AWS_ACCESS_KEY_ID = "${var.DNS_AWS_ACCESS_KEY_ID}"
  DNS_AWS_SECRET_ACCESS_KEY = "${var.DNS_AWS_SECRET_ACCESS_KEY}"
}
...
```

### Expected Behavior
The plan should error out. `type = "alias"` is a typo, and `alias` is not a valid value. The correct value should be `type = "A"`.

### Actual Behavior
Plan succeeds without issues. Then the apply step fails with:

```
* aws_route53_record.alias: [ERR]: Error building changeset: InvalidInput: Invalid XML ;
   cvc-enumeration-valid: Value 'alias' is not facet-valid with respect to enumeration
   '[SOA, A, TXT, NS, CNAME, MX, NAPTR, PTR, SRV, SPF, AAAA]'. It must be a value
   from the enumeration.
	status code: 400, request id: 3c9fc39d-ab23-11e6-b940-83d82bd81212
```

When changing the value to `A` and re-running both the plan and apply steps work.

### Steps to Reproduce
1.  `terraform plan -out planfile`
2. `terraform apply planfile`

### Important Factoids
I'm declaring the provider inside the module, because the DNS provider is different (a separate AWS account) from that of other resources.

PR #11119, opened 2017-01-09
provider/aws: Route53 Record: Add Type validation

grubernaut:
Adds validation for the `type` parameter of an `aws_route53_record` resource.

This will allow Terraform to catch any user errors of a `type` parameter during a `terraform plan` instead of during a `terraform apply`.

Fixes: #11114


stack72:
LGTM! Nice work @grubernaut :)

PR #10862, opened 2016-12-20
New provider arukas

yamamoto-febc:
This PR adds a provider for the [Arukas](https://arukas.io/en/).

Arukas is a Docker hosting service and is explained on the official website as follows.

> Arukas is the simplest-to-use Container-as-a-service that makes it easy to deploy and manage apps at scale.

One can manage the container on arukas like this:
```hcl
provider "arukas" {
    token = "[put_your_api_token]"
    secret = "[put_your_api_secret]"
}

resource "arukas_container" "foobar" {
    name = "terraform_for_arukas_test_foobar"
    image = "nginx:latest"
    instances = 1
    memory = 256
    ports = {
        protocol = "tcp"
        number = "80"
    }
    environments {
        key = "key1"
        value = "value1"
    }
}
```

The PR includes tests & documentation.

Test output :
```sh
TF_ACC=1 go test ./builtin/providers/arukas -v  -timeout 120m
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProvider_impl
--- PASS: TestProvider_impl (0.00s)
=== RUN   TestAccArukasContainer_Basic
--- PASS: TestAccArukasContainer_Basic (80.82s)
=== RUN   TestAccArukasContainer_Update
--- PASS: TestAccArukasContainer_Update (164.25s)
=== RUN   TestAccArukasContainer_Minimum
--- PASS: TestAccArukasContainer_Minimum (79.97s)
=== RUN   TestAccArukasContainer_Import
--- PASS: TestAccArukasContainer_Import (69.71s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/arukas	394.961s
```

Looking forward to your feedback!

PR #11140, opened 2017-01-11
provider/arukas: Fix broken build on windows

yamamoto-febc:
This fixes the problem that the build broke on Windows by merging #10862.

### DIFF
--- a/vendor/gopkg.in/alecthomas/kingpin.v2/guesswidth.go
+++ b/vendor/gopkg.in/alecthomas/kingpin.v2/guesswidth.go
@@ -1,0 +1,9 @@
+// +build appengine !linux,!freebsd,!darwin,!dragonfly,!netbsd,!openbsd
+
+package kingpin
+
+import "io"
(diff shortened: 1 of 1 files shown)


jbardin:
Hi @yamamoto-febc,

Was this fixed upstream? 

You can't patch a vendored package directly, since the next person who updates the vendor directory will revert the change.  If it was fixed upstream, we ned to update the manifest so that the hash matches the correct revision.

yamamoto-febc:
Hi @jbardin,

Sorry , I don't seem to have understood the behavior of `govendor`.
Surely this method is not working well...

Since the original PR #10862 has been reverted, this PR's will also be included when re-issuing PullRequest, so please close this PR.

Regarding the case where the build on Windows has been broken, I will create another PullRequest which adds the Arukas provider, so let me discuss it there.

Thanks.

yamamoto-febc:
It closes because it corresponds with #11171

Issue #11134, opened 2017-01-10
aws_route_table crash on terraform plan


mkolbert:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

0.8.3 as well as 0.8.2 prior to upgrading today.

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_instance
- dnsimple_record

aws_route_table

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```
variable "AWS_REGION" { default = "us-east-1" }

variable "PEERED_ENV" { default = "XXXXXXXX" }

provider "aws" {
  alias = "this"
  region = "${var.AWS_REGION}"
  profile = "${var.PEERED_ENV}"
}

data "aws_route_table" "RTB" {
  provider = "aws.this"
  tags = {
    Name = "Private Route Table"
    Region = "us-east-1a"
  }
}
```
### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

<script src="https://gist.github.com/mkolbert/465e0e5ec4da96b48ff255ba29a1839b.js"></script>

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

<script src="https://gist.github.com/mkolbert/dc6ad90a6bb0647bd286d83597dce00b.js"></script>

### Expected Behavior
What should have happened?

Using same syntax with aws_subnet works fine.

### Actual Behavior
What actually happened?

Crash

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

terraform plan

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

VPC

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-1234

grubernaut:
Hi @mkolbert, thanks for the issue!

I'm having trouble reproducing the issue you're facing. Is there any other configuration that would allow for a successful reproduction? 

[...]

PR #11152, opened 2017-01-11
provider/aws: Fix panic caused by main route table lookup

grubernaut:
A VPC's main route table has an implicit subnet association, not an explicit subnet association. This caused a Terraform panic when using the `data_source_aws_route_table` resource to query the main route table for a VPC.

This fixes the Terraform panic, and allows the data lookup to complete successfully. Also added an acceptance test to verify the bugfix.

Fixes: #11134


paddyforan:
LGTM :+1:

PR #11155, opened 2017-01-11
provisioner/remote-exec: fail on first inline script with bad exit code

### LINKED ISSUE #11145: remote-exec: inline scripts only fail if the last command does
A remote-exec provisioner configured with multiple inline commands will only fail if the last command returns a non zero exit code. This is due to the implementation of the provisioner collecting each inline command into a single script which is uploaded and executed. Is this the intended behaviour, if so then it would be helpful for the documentation to mention this.

### Terraform Version

Terraform v0.8.4-dev (6bafd4c8961f13edf0effda9ebfeb208c5ec5ffe)

### Affected Resource(s)
- remote-exec

```hcl
resource "null_resource" "test" {
	provisioner "remote-exec" {
		connection {
		    host = "..."
			type = "winrm"
			user = "peter"
			password = "..."
		}

        inline = [
			"cmd /c exit 1",
			"cmd /c exit 0"
        ]
    }
}
```

### Expected Behavior

Terraform should detect the non zero exit code of the first command.

### Actual Behavior

Terraform reports that the apply was successful.

### Steps to Reproduce

1. `terraform apply`

pmcatominey:
Fixes #11145.

The provisioner collected all inline commands into a single script which meant
only the exit code of the last command was actually checked for an error.

### DIFF
--- a/builtin/provisioners/remote-exec/resource_provisioner.go
+++ b/builtin/provisioners/remote-exec/resource_provisioner.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform/communicator"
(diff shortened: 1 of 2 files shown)


stack72:

dcousens:
Is there any way to set this?  The change appears to have been reverted?

PR #11157, opened 2017-01-11
provider/aws: Add missing id argument for Route Table data source

grubernaut:
Documentation for the `aws_route_table` data source mentions that it supports a route table `id` as an argument, however it was missing from the actual provider code.

Adds in the missing provider code, adds a test, and updates the documentation to use `rtb_id` as the argument, instead of the more ambiguous `id`.


paddyforan:
LGTM :+1:

Issue #11131, opened 2017-01-10
For Azure China Error obtaining authorization token: Error refreshing token: invalid character '<' looking for beginning of value


nicocesar:
### Terraform Version
terraform version: 0.8.3  b845cb7093c24c49ece854bb38b55aff3783f195

### Affected Resource(s)
azure_resource_group

(but I think this affects the refreshing of the authentication token

### Terraform Configuration Files
https://www.dropbox.com/s/ij7u77i37mvwbb1/invalid_character_error.zip?dl=0

### Debug Output
https://www.dropbox.com/s/su4rqy8bm90xon4/invalid_character_error.txt.asc?dl=0
encrypted for: 
```
pub  2048R/32EA77F3 2014-02-26 HashiCorp Security <security@hashicorp.com>
 Primary key fingerprint: 91A6 E7F8 5D05 C656 30BE  F189 5185 2D87 348F FC4C
      Subkey fingerprint: E5E7 92FA E106 A1B3 B272  B65D D5F4 3B64 32EA 77F3

```

### Expected Behavior
Detect that the resource is already created ,  or create new ones 

also #6769 should provide better HTTP debug. 

### Actual Behavior
```
Error applying plan:

1 error(s) occurred:

* azurerm_resource_group.rg: Error creating resource group: Error obtaining authorization token: Error refreshing token: invalid character '<' looking for beginning of value

```

### Steps to Reproduce
terraform apply -state mystate.tfstate -var-file invalid_character_error.tfvars -target=azurerm_virtual_machine.switchyard > invalid_character_error.txt 2>&1

### Important Factoids
This is in azure China.  There could be a possibility that the queried endpoints are wrong leading to a html/xml reply at some point of the authorization token refresh. If we had the full string instead of only the offending character maybe we can debug it better  .

nicocesar:
I ran  tcpdump in the background and I see that is connecting to **login.chinacloudapi.cn**

```
2017/01/10 15:40:58 [DEBUG] root: eval: *terraform.EvalApply
azurerm_resource_group.rg: Creating...
  location:  "" => "chinaeast"
  name:      "" => "d8kp5"
  tags.%:    "" => "1"
  tags.Name: "" => "d8kp5"
2017/01/10 15:40:58 [DEBUG] apply: azurerm_resource_group.rg: executing Apply
[...]

PR #11158, opened 2017-01-11
Fixing terraform's issue #11131

nicocesar:
is "?version-api" a leftover from some older server version deployed ?  it doesn't work for me in China - see https://github.com/hashicorp/terraform/issues/11131


pmcatominey:
Hi @nicocesar,

Since the SDK is a vendored package, we'll need the fix to be made upstream and then we can update out vendored version. I see that you've opened a PR to fix this in the go-autorest library, once that is merged and is released in a new SDK version we'll be able to pull the change in.

Thanks

nicocesar:
@pmcatominey sorry about that .. did  https://github.com/Azure/go-autorest/pull/105

PR #11160, opened 2017-01-11
provider/aws: Find matching public zones even when passing in a vpc_id to aws_route53_zone data source 

tomelliff:
I'm hoping to use the `aws_route53_zone` data source more generically so that when I create a route53 record, whether in a public or a private zone, and just have it return the appropriate zone ID for me (currently hardcoding a lookup list of IDs from the pre-data source days).

A simple example would look something like this:

### modules/elb/main.tf:
```hcl
variable "environment" {}
variable "internal" {}
variable "subnets" {}
variable "vpc_id" {}
variable "dns_record" { default = "www" }
variable "dns_domain" {}

resource "aws_elb" "balancer" {
  name = "elb-${var.environment}"
  internal = "${var.internal}"
  subnets = [ "${split(",", var.subnets)}" ]

  listener {
    lb_port = "443"
    lb_protocol = "https"
    instance_port = "443"
    instance_protocol = "https"
  }
}

data "aws_route53_zone" "zone" {
  name = "${var.dns_domain}"
  vpc_id = "${var.vpc_id}"
  private_zone = "${var.internal}"
}

resource "aws_route53_record" "frontend" {
  name = "${var.dns_record}.${var.dns_domain}"
  zone_id = "${data.aws_route53_zone.zone.id}"
  type = "A"

  alias {
    name = "${aws_elb.balancer.dns_name}"
    zone_id = "${aws_elb.balancer.zone_id }"
    evaluate_target_health = "false"
  }
}
```

### dev/main.tf:
```hcl
module "elb" {
  source = "../modules/elb"
  environment = "dev"
  internal = "true"
  subnets = "subnet-123abc,subnet-124abc,subnet-125abc"
  vpc_id = "vpc-123abc"
  dns_domain = dev.example.com
}
```

### prod/main.tf:
```hcl
module "elb" {
  source = "../modules/elb"
  environment = "prod"
  internal = "false"
  subnets = "subnet-123abd,subnet-124abd,subnet-125abd"
  vpc_id = "vpc-123abd"
  dns_domain = example.com
}
```

In the above example the generic ELB module comes with a Route53 record that may or may not be in a public zone (in prod it is public, in dev it's not). With the current code on master running prod would still pass in a `vpc_id` parameter to the data source and so force it to look for private zones only and fail to match a private zone called `example.com` with there only being a `dev.example.com` private zone.

This pull request aims to fix that so if the `private_zone` parameter is set to false then it will search for public zones even if `vpc_id` is provided.

### DIFF
--- a/builtin/providers/aws/data_source_aws_route53_zone.go
+++ b/builtin/providers/aws/data_source_aws_route53_zone.go
@@ -104,6 +104,10 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 							break
 						}
 					}
+					// if we specified a vpc_id but private_zone is false and the name matches
+					if *hostedZone.Config.PrivateZone == false {
+						matchingVPC = true
+					}
 				} else {
 					matchingVPC = true
 				}
--- a/builtin/providers/aws/data_source_aws_route53_zone_test.go
+++ b/builtin/providers/aws/data_source_aws_route53_zone_test.go
@@ -28,9 +28,13 @@ func TestAccDataSourceAwsRoute53Zone(t *testing.T) {
 					testAccDataSourceAwsRoute53ZoneCheck(
 						publicResourceName, "data.aws_route53_zone.by_name", publicDomain),
 					testAccDataSourceAwsRoute53ZoneCheck(
+						publicResourceName, "data.aws_route53_zone.public_with_vpc", publicDomain),
+					testAccDataSourceAwsRoute53ZoneCheck(
 						privateResourceName, "data.aws_route53_zone.by_vpc", privateDomain),
 					testAccDataSourceAwsRoute53ZoneCheck(
 						privateResourceName, "data.aws_route53_zone.by_tag", privateDomain),
+					testAccDataSourceAwsRoute53ZoneCheck(
+						privateResourceName, "data.aws_route53_zone.by_name_explicit_private", privateDomain),
 				),
 			},
 		},
@@ -79,9 +83,21 @@ func testAccDataSourceAwsRoute53ZoneConfig(rInt int) string {
 		cidr_block = "172.16.0.0/16"
 	}
 
+	resource "aws_vpc" "test_multiple" {
+		cidr_block = "172.17.0.0/16"
+	}
+
 	resource "aws_route53_zone" "test_private" {
 		name = "test.acc-%d."
 		vpc_id = "${aws_vpc.test.id}"
+		tags {
+			Environment = "dev-%d"
+		}
+	}
+
+	resource "aws_route53_zone" "test_multiple_private" {
+		name = "test.acc-%d."
+		vpc_id = "${aws_vpc.test_multiple.id}"
 		tags {
 			Environment = "dev-%d"
 		}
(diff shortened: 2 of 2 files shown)


tomelliff:
I wanted to raise this as is while I have a think about the approach but would appreciate any feedback on what's there now or ideas for how best to make this work.

tomelliff:
That looks to have sorted it now. Appreciate any feedback on potential missing test cases or approach though.

tomelliff:
Does this need anything else from me or is it just waiting to be picked up into a release?

tomelliff:
@radeksimko Sorry for bumping again but does this need anything more from me?

tomelliff:
@stack72 Not sure if I need to be doing something else get any feedback/eyes on this PR?

tomelliff:
@radeksimko Thanks again for agreeing to take a look at this tonight and sorry for grabbing you unaware at the Hashicorp meetup ;)

I took a look at trying to fold in the changes made by @grubernaut in https://github.com/hashicorp/terraform/commit/3d090b203e5309ede975b6bc110b8baafa8a890a but I'm wary that my change was also adding in a missing test for multiple Route53 zones named the same but attached to different VPCs and I can't see how I can test that now that the zone has gone from `test.acc.` to `test.acc-%d.` (where `%d` is a RandInt) in the normal test_private resource. Should I leave that as is and create 2 new VPCs and zones instead of just the extra one with the same name and not randomise the name?

I'm a little wary that those changes were put in to help around parallelisation of test runs so the above change might then go back to breaking that.

tomelliff:
I made a complete mess of trying to rebase the merge conflicts out of this so I've force pushed the work back over a clean pull from upstream taking into account the work for the parallel tests that @grubernaut did. And I was misreading the code after a long day the other night because obviously it's a single random integer used for all the `%d`s so the randomisation wasn't an issue.

Should be good for review now @radeksimko if that's okay?

radeksimko:
Hi @tomelliff 
it was nice meeting you at the meetup and sorry for the long feedback loop here. 😅 

I was thinking about the root cause of this issue for a while and I'm wondering if a simple conditional would help here - see below:

```hcl
variable "environment" {}
variable "internal" {}
variable "subnets" {}
variable "vpc_id" {}
variable "dns_record" { default = "www" }
variable "dns_domain" {}

resource "aws_elb" "balancer" {
  name = "elb-${var.environment}"
  internal = "${var.internal}"
  subnets = [ "${split(",", var.subnets)}" ]

  listener {
    lb_port = "443"
    lb_protocol = "https"
    instance_port = "443"
    instance_protocol = "https"
  }
}

data "aws_route53_zone" "zone" {
  name = "${var.dns_domain}"
  vpc_id = "${var.internal == "true" ? var.vpc_id : ""}"
  private_zone = "${var.internal}"
}

resource "aws_route53_record" "frontend" {
  name = "${var.dns_record}.${var.dns_domain}"
  zone_id = "${data.aws_route53_zone.zone.id}"
  type = "A"

  alias {
    name = "${aws_elb.balancer.dns_name}"
    zone_id = "${aws_elb.balancer.zone_id}"
    evaluate_target_health = "false"
  }
}
```

I was able to `plan` that successfully as long as there was 1 and only 1 public zone with that name.

----

My worry about the proposed solution is that we'd be loosening the filter in a way that passed parameters are no longer true (because public zones simply don't have any `vpc_id` assigned to them).

I understand how that may be expected and desired behaviour for you, but some other folks may rely on more restrictive search.

I think that finding wrong zone and potentially modifying wrong zone because of such loose match is a risk I'd prefer not to expose our users to.

I hope it's understandable - let me know what you think. 😉

tomelliff:
Yep, I think you're right there. I thought I'd tried that but couldn't get the syntax right (was just when 0.8 was released) but trying that now looks like it works without any changes to Terraform.

Thanks for spotting the mistake in my configuration though :)

PR #11161, opened 2017-01-11
provider/postgres set default postgres connection limit

pbthorste:
Hi, there seems to have been a regression for the postgres provider. When creating a new database or role, the connection limit is set to 0, making them unusable. The docs say that the default should be "-1".

This small change sets the default to -1, making the posgres provider act as it did previously.

### DIFF
--- a/builtin/providers/postgresql/resource_postgresql_database.go
+++ b/builtin/providers/postgresql/resource_postgresql_database.go
@@ -86,7 +86,6 @@ func resourcePostgreSQLDatabase() *schema.Resource {
 			dbConnLimitAttr: {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Computed:     true,
 				Default:      -1,
 				Description:  "How many concurrent connections can be made to this database",
 				ValidateFunc: validateConnLimit,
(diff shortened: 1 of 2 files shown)


sean-:
Thank you, good catch.  I'm adding a follow up commit to add a unit test in just a minute.

sean-:
Anecdotally, I remember why I set this to computed: the connection limit goes from `-1` to `0` when you change a `ROLE` from `LOGIN` to `NOLOGIN` the `connection_limit` changes from `-1` to `0`.  Ah well, can't win them all.  Explicitly setting this is the way to go.

pbthorste:
Does that mean that if you create a role with login: false, you also must set the connection limit to 0 because of this change?

sean-:
No, that's what happens inside of PostgreSQL.  If you set a `postgresql_role` to `login=false`, inside of PostgreSQL it changes the `canlogin` attribute to `f` *and* sets their `connlimit` attribute to `0`.  If you re-enable login in `postgresql_role`, it will re-enable the `connection_limit` of `-1` (in the next release, I found a small syntax error and fixed it yesterday).

Issue #10941, opened 2016-12-28
Support importing an aws_route53_record


kitsunde:
Currently if you want to have the `aws_route53_zone` as a data source, but absorb a record into terraform as managed resource you have to either delete the records before running terraform or create duplicate records and rename them in the tfstate file.

This was the first thing I did when transitioning an existing project.

### Terraform Version

0.8.2

### Affected Resource(s)

- aws_route53_record

### Terraform Configuration Files
```hcl
data "aws_route53_zone" "sg" {
  name = "mediapop.sg."
}

resource "aws_route53_record" "mediapop-sg" {
  zone_id = "${data.aws_route53_zone.sg.zone_id}"
  name = "${data.aws_route53_zone.sg.name}"
  type = "A"
  ttl = "300"
  records = [
    "54.169.106.147"
  ]
}
```

### Expected Behavior

I should be able to import a record from a zone.

### Actual Behavior

> * import aws_route53_record.mediapop-co (id: Z38P2EBBHKFAZI_mediapop.sg_A): resource aws_route53_record doesn't support import

### Steps to Reproduce

1. `terraform import aws_route53_record.mediapop-co Z38P2EBBHKFAZI_mediapop.sg_A`

PR #11162, opened 2017-01-11
provider/aws: Add import to route53 record

grubernaut:
Allows the user to run:

```
terraform import aws_route53_record.record_name <record_id>
```

Fixes: #10941

### DIFF
--- a/builtin/providers/aws/import_aws_route53_record_test.go
+++ b/builtin/providers/aws/import_aws_route53_record_test.go
@@ -1,0 +1,28 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAwsRoute53Record_importBasic(t *testing.T) {
+	resourceName := "aws_route53_record.default"
+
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/import_aws_route53_record_test.go
```diff
@@ -0,0 +1,28 @@
...
+		CheckDestroy: testAccCheckRoute53RecordDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccRoute53RecordConfig,
+			},
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"weight"},
```
stack72: Why are you ignoring weight here?

grubernaut: Weight is now removed in favor of using `weighted_routing_policy`, if a non-weighted record is imported the `weight` attribute is set to a default of `-1`. We should probably clean this up in a future major release.

PR #11165, opened 2017-01-11
Add a test to make sure the implied `connection_limit` default of `-1` is always set.

sean-:
Covers: #11161


sean-:
Cleaned up the excessive newlines from the original style of this resource.

PR #11171, opened 2017-01-12
New provider arukas

yamamoto-febc:
==UPDATED==

By updating upstream, dependency library `gopkg.in/alecthomas/kingpin.v2` is no longer used, so the following problem is solved.
And updating the ignore section of `vendor/vendor.json` is no longer necessary.

---

This PR was recreated to solve the problem occurring at #10862.

## Problems

This provider includes a dependency that at time of writing requires a *nix system, and will not build on Windows.

## Cause

The cause of this case is probably in `govendor` or `gopkg.in/alecthomas/kingpin.v2` package.

First of all, the direct cause of this case was that `vendor/gopkg.in/alecthomas/kingpin.v2/guesswidth.go` file necessary for building on Windows was not included in the vendor directory.

Next, it is about why this file was not included in the vendor directory, but this seems to be caused by the existing "appengine" in the ignore section of vendor.json as follows.

https://github.com/hashicorp/terraform/blob/v0.8.4/vendor/vendor.json#L3

```json
    "ignore": "appengine test",
```

The build tag of the target file and the fallback file is described as follows.

https://github.com/alecthomas/kingpin/blob/v2.2.3/guesswidth.go#L1

```go
// +build appengine !linux,!freebsd,!darwin,!dragonfly,!netbsd,!openbsd
```

https://github.com/alecthomas/kingpin/blob/v2.2.3/guesswidth_unix.go#L1

```go
// +build !appengine,linux freebsd darwin dragonfly netbsd openbsd
```

On the other hand, as discussed in [this Issue](https://github.com/kardianos/govendor/issues/262), in govendor it seems that if the build tag matches even one of the ignore in vendor.json, it does not include the target file under the vendor directory.

In this case, since "appengine" is included in the ignore of vendor.json, `guesswidth.go` will not be copied under the vendor directory even if you execute `govendor add or fetch`.

## Solution

- Remove "appengine" from the ignore section of vendor.json

In this case, the size of the vendor directory will become large.
And there may be other problems as well.   

What do you think about this method?

Looking forward to your feedback!


yamamoto-febc:
rebased.

stack72:
Hi @yamamoto-febc 

Thanks for this - @jbardin has checked out the vendor file and is happy with it

Pulling it locally to run the tests again

Thanks

Paul

stack72:
Test results post merge conflicts:

```
% make testacc TEST=./builtin/providers/arukas
2017/02/13 19:02:31 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/arukas -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccArukasContainer_Basic
--- PASS: TestAccArukasContainer_Basic (22.90s)
=== RUN   TestAccArukasContainer_Update
--- PASS: TestAccArukasContainer_Update (101.32s)
=== RUN   TestAccArukasContainer_Minimum
--- PASS: TestAccArukasContainer_Minimum (80.00s)
=== RUN   TestAccArukasContainer_Import
--- PASS: TestAccArukasContainer_Import (22.16s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/arukas	226.409s
```

PR #11179, opened 2017-01-12
Add support for StatusCake confirmation servers

samcrang:
StatusCake supports confirmation servers which makes StatusCake tests check with additional servers before detecting, and possibly alerting on, downtime.

I've updated the documentation to describe the new agrument; this doesn't need to go out now but the docs site would need to be updated when/if this is merged and released.

### DIFF
--- a/builtin/providers/statuscake/resource_statuscaketest.go
+++ b/builtin/providers/statuscake/resource_statuscaketest.go
@@ -57,6 +57,10 @@ func resourceStatusCakeTest() *schema.Resource {
 				Type:     schema.TypeInt,
 				Optional: true,
 			},
+			"confirmations": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


stack72:
Hi @samcrang 

Thanks for the work here - according to the tests, I can see that you are settting 1 confirmation in the test fixture but when it get's set to state we check for a 0

Does this mean that it doesn't work as expected?

Thanks

Paul

samcrang:
Hi @stack72,

It should work correctly, however, I think it is confusingly tested. I was basically following how `contact_id` is already tested when adding `confirmations`. You’ve made me look at it a bit closer and we should be able to make it clearer.

[These assertions](https://github.com/samcrang/terraform/blob/fdc64ec08750bd13248ffb872a3beb112bb28c63/builtin/providers/statuscake/resource_statuscaketest_test.go#L52-L55) (which I think is what makes the diff confusing) appear redundant given the call to `testAccTestCheckAttributes` that immediately precedes it.

I’m happy to give it a refactor to make it clearer. If it were me I’d probably just inline all the assertions and drop `testAccTestCheckAttributes` entirely. Does that seem reasonable?

stack72:
Hi @samcrang 

I'd love a follow up PR that allows us to have better tests for statuscake - if that is something you could do that would be amazing!

I will merge this as is as the tests pass:

```
% make testacc TEST=./builtin/providers/statuscake                                                                          ✭
2017/01/12 20:50:35 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/statuscake -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccStatusCake_basic
--- PASS: TestAccStatusCake_basic (2.53s)
=== RUN   TestAccStatusCake_withUpdate
--- PASS: TestAccStatusCake_withUpdate (2.95s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/statuscake	5.494s
```

Paul

PR #11181, opened 2017-01-12
Example aws-elb fix

dannyleesmith:
While testing the aws-elb example ran into the following error when performing a terraform apply:
```

Error applying plan:

2 error(s) occurred:

* aws_security_group.default: Error authorizing security group egress rules: InvalidParameterValue: Only Amazon VPC security groups may be used with this operation.
	status code: 400, request id: 061c2299-8be5-4620-9c24-226a0f1c32ba
* aws_security_group.elb: Error authorizing security group egress rules: InvalidParameterValue: Only Amazon VPC security groups may be used with this operation.
	status code: 400, request id: 4a7225f1-95f9-4696-b55e-54e9487a50d6

```

This initial error led to a few other errors which should be resolved in this version.

Most of the changes involved leveraging arguments that are supported by EC2 Classic because the current master references VPC options. Because of the changes to make this work completely outside of a VPC there was a need to change the instance type to a t1.micro as the t2.micro was VPC only. To facilitate this change the variables.tf file was updated to use the appropriate AMI type.

Tested against 0.8.4 to validate that this had not been resolved in a newer release.

### DIFF
--- a/examples/aws-elb/main.tf
+++ b/examples/aws-elb/main.tf
@@ -25,13 +25,6 @@ resource "aws_security_group" "default" {
     cidr_blocks = ["0.0.0.0/0"]
   }
 
-  # outbound internet access
-  egress {
-    from_port   = 0
-    to_port     = 0
-    protocol    = "-1"
-    cidr_blocks = ["0.0.0.0/0"]
-  }
 }
 
 # Our elb security group to access
@@ -48,13 +41,6 @@ resource "aws_security_group" "elb" {
     cidr_blocks = ["0.0.0.0/0"]
   }
 
-  # outbound internet access
-  egress {
-    from_port   = 0
-    to_port     = 0
-    protocol    = "-1"
-    cidr_blocks = ["0.0.0.0/0"]
-  }
 }
 
 resource "aws_elb" "web" {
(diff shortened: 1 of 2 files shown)


catsby:
Hey @dannyleesmith – thanks for updating the examples. Do they now work in both VPC and Classic environments? While we want to still support Classic, unless otherwise specified we assume VPC enabled accounts.

dannyleesmith:
Hi @catsby,
The example had VPC arguments in use but was not fully formed to create all the necessary components to function in a VPC environment. I think this should therefore stand as a standalone EC2 Classic example, and perhaps it would be good to have an aws-elb-vpc example? It may also be worth updating the README for this example to make it clearer that it will use EC2 Classic.

catsby:
> The example had VPC arguments in use but was not fully formed to create all the necessary components to function in a VPC environment.

Thanks for pointing this out! I should have investigated further and found this myself. 

I would prefer we address this issue by adding in the necessary VPC items to this example, rather than modifying to fit Classic. I don't have the numbers available but I imagine a vast majority of our users are VPC only. 

Are you up for contributing this change? I'm happy to pick it up myself if not, we do appreciate the work you've done so far 😄

catsby:
I've opened https://github.com/hashicorp/terraform/pull/11200 as an alternative

dannyleesmith:
Hi @catsby, I'd certainly like to put that forward. I'll let you know when I've got something ready and I'll work out raising it in #11200.

dannyleesmith:
Closing this PR as I can see the VPC changes have now been made in #11200

PR #11183, opened 2017-01-12
provider/scaleway improve bootscript data source

nicolai86:
this PR fixes the broken scaleway `bootscript` data source tests:

- it corrects a long standing error in the datasource: when `name` and `architecture` where present, `architecture` was not used for filtering the results. This behaviour however exists when a `name_filter` is present, so now the data source behaves the same in both cases.
- it improves the reliability of the tests by not hardcoding bootscript names.  these tend to change rather quickly and we don't want to update the test cases all the time because of some external change.

tests are green:

```
make testacc TEST=./builtin/providers/scaleway TESTARGS='-run=TestAccScalewayDataSourceBootscript'
2017/01/12 22:55:22 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/scaleway -v -run=TestAccScalewayDataSourceBootscript -timeout 120m
=== RUN   TestAccScalewayDataSourceBootscript_Basic
--- PASS: TestAccScalewayDataSourceBootscript_Basic (2.16s)
=== RUN   TestAccScalewayDataSourceBootscript_Filtered
--- PASS: TestAccScalewayDataSourceBootscript_Filtered (1.58s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/scaleway	3.763s
```

@stack72 PTAL

### DIFF
--- a/builtin/providers/scaleway/data_source_scaleway_bootscript_test.go
+++ b/builtin/providers/scaleway/data_source_scaleway_bootscript_test.go
@@ -20,7 +20,7 @@ func TestAccScalewayDataSourceBootscript_Basic(t *testing.T) {
 					testAccCheckBootscriptID("data.scaleway_bootscript.debug"),
 					resource.TestCheckResourceAttr("data.scaleway_bootscript.debug", "architecture", "x86_64"),
 					resource.TestCheckResourceAttr("data.scaleway_bootscript.debug", "public", "true"),
-					resource.TestMatchResourceAttr("data.scaleway_bootscript.debug", "kernel", regexp.MustCompile("4.8.3")),
+					resource.TestMatchResourceAttr("data.scaleway_bootscript.debug", "kernel", regexp.MustCompile("4.8.14")),
 				),
 			},
 		},
@@ -67,7 +67,7 @@ func testAccCheckBootscriptID(n string) resource.TestCheckFunc {
 
 const testAccCheckScalewayBootscriptConfig = `
 data "scaleway_bootscript" "debug" {
-  name = "x86_64 4.8.3 debug #1"
+  name = "x86_64 4.8.14 debug #2"
 }
 `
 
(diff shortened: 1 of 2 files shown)


builtin/providers/scaleway/data_source_scaleway_bootscript_test.go
```diff
@@ -2,25 +2,44 @@ package scaleway
...
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
+	"github.com/scaleway/scaleway-cli/pkg/api"
+	"github.com/scaleway/scaleway-cli/pkg/scwversion"
 )
 
 func TestAccScalewayDataSourceBootscript_Basic(t *testing.T) {
+	testAccPreCheck(t)
+	client, err := api.NewScalewayAPI(
```
stack72: this is causing the build to fail:

```
ok  	github.com/hashicorp/terraform/builtin/providers/rundeck	0.192s
--- FAIL: TestAccScalewayDataSourceBootscript_Basic (0.00s)
	provider_test.go:33: SCALEWAY_ORGANIZATION must be set for acceptance tests
FAIL
```

stack72: We should assume these are in place and passed as variables IMO

nicolai86: @stack72 I tried `testAccProvider.Meta().(*Client).scaleway` but Meta returned nil. it was only present if a) I have a test with multiple steps and read it after the first `TestStep` or b) use it inside the verify function.

Sadly I did not find the place where `SetMeta` is being called, so I'm not sure why it's not available there.

nicolai86:
@stack72 I've removed the dynamic bootscript lookup for now. will try in a separate pr.

this change will make our ci green again for scaleway 👍

stack72:
LGTM :) Thanks for following up with this to make it green 

Paul

PR #11187, opened 2017-01-12
State Locking initial implementations

jbardin:
Basic state locking.

This all starts with the `state.Locker` interface:

    type StateLocker interface {
        Lock(reason string) error
        Unlock() error
    }

This is implemented by the internal state wrappers (CacheState, BackupState), as well as LocalState. Any commands that operating on the state can use these functions to prevent concurrent modification. 

Remote state clients can also implement this same interface, which will be called via the State structures that wraps them. 

This PR contains the initial LocalState and AWS locking implementations. No state locking actually occurs yet, and adding the Lock calls and associated commands will come in another PR.

### DIFF
--- a/state/local_lock_windows.go
+++ b/state/local_lock_windows.go
@@ -1,0 +1,140 @@
+// +build windows
+
+package state
+
+import (
+	"math"
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+type stateLock struct {
+	handle syscall.Handle
+}
+
+var (
+	modkernel32      = syscall.NewLazyDLL("kernel32.dll")
+	procLockFileEx   = modkernel32.NewProc("LockFileEx")
+	procCreateEventW = modkernel32.NewProc("CreateEventW")
+
+	lockedFiles = map[*os.File]syscall.Handle{}
+)
+
+const (
+	_LOCKFILE_FAIL_IMMEDIATELY = 1
+	_LOCKFILE_EXCLUSIVE_LOCK   = 2
+)
+
+func (s *LocalState) lock() error {
+	name, err := syscall.UTF16PtrFromString(s.PathOut)
+	if err != nil {
+		return err
+	}
+
+	handle, err := syscall.CreateFile(
+		name,
+		syscall.GENERIC_READ|syscall.GENERIC_WRITE,
+		// since this file is already open in out process, we need shared
+		// access here for this call.
+		syscall.FILE_SHARE_READ|syscall.FILE_SHARE_WRITE,
+		nil,
+		syscall.OPEN_EXISTING,
+		syscall.FILE_ATTRIBUTE_NORMAL,
+		0,
+	)
+	if err != nil {
+		return err
+	}
+
+	lockedFiles[s.stateFileOut] = handle
+
+	// even though we're failing immediately, an overlapped event structure is
+	// required
+	ol, err := newOverlapped()
+	if err != nil {
+		return err
+	}
+	defer syscall.CloseHandle(ol.HEvent)
+
+	return lockFileEx(
+		handle,
+		_LOCKFILE_EXCLUSIVE_LOCK|_LOCKFILE_FAIL_IMMEDIATELY,
+		0,              // reserved
+		0,              // bytes low
+		math.MaxUint32, // bytes high
+		ol,
+	)
+}
+
+func (s *LocalState) unlock() error {
+	handle, ok := lockedFiles[s.stateFileOut]
+	if !ok {
+		// we allow multiple Unlock calls
+		return nil
+	}
+	delete(lockedFiles, s.stateFileOut)
+	return syscall.Close(handle)
+}
+
+func lockFileEx(h syscall.Handle, flags, reserved, locklow, lockhigh uint32, ol *syscall.Overlapped) (err error) {
+	r1, _, e1 := syscall.Syscall6(
+		procLockFileEx.Addr(),
+		6,
+		uintptr(h),
+		uintptr(flags),
+		uintptr(reserved),
+		uintptr(locklow),
+		uintptr(lockhigh),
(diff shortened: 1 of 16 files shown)


apparentlymart:
This seems related to #5036. :grinning:

radeksimko:
Is there any specific reason you decided to use DynamoDB to hold the lock for S3 remote state backend instead of saving a "lockfile" in S3? Is it just for the [potential strong consistency](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html)?

jbardin:
@radeksimko, yes it's specifically for the strong consistency offered by dynamodb. S3 is only eventually consistent in almost all cases. 

IIRC S3 might be consistent after a PUT of a new object and a GET of that object specifically, but even in that case there's no way to make a PUT conditional, so if multiple locks were in-flight at the same time the last timestamp wins. In this case 2 clients could see no lock file, PUT a new lock file, and even verify that only their unique lock file exists while continuing to run concurrently.

state/remote/s3.go
```diff
@@ -188,3 +195,73 @@ func (c *S3Client) Delete() error {
...
+		return nil
+	}
+
+	stateName := fmt.Sprintf("%s/%s", c.bucketName, c.keyName)
+
+	putParams := &dynamodb.PutItemInput{
+		Item: map[string]*dynamodb.AttributeValue{
+			"LockID":  {S: aws.String(stateName)},
+			"Created": {S: aws.String(time.Now().UTC().Format(time.RFC3339))},
+			"Expires": {S: aws.String(time.Now().Add(time.Hour).UTC().Format(time.RFC3339))},
```
radeksimko: Could the expiration period be configurable somewhere? Maybe not necessarily from user's perspective (CLI), but I'm thinking it could be passed in as `time.Duration` argument to `Lock()`? That might also make developer implementing new remote backends realise that the expiration should be respected/implemented properly where possible.

radeksimko: Actually I discussed this with @mitchellh and we came to a conclusion that expiration generally may introduce unnecessary edge cases - e.g. long TF runs or leap seconds/years so it's probably best to just not implement expiration at all.

jbardin: Yes, I was going to leave expiration handling up to the individual state client, and have it be strictly informational to start. The user can see the expiration value in the error message and choose to override manually.

apparentlymart:
Probably-too-late design question: would it be useful to make locking method orthogonal to state storage? For example: maybe I want to make a stronger locking guarantee by sharing the same lock across multiple states, ensuring that only one can change at a time.

Probably less-compelling example: I store my state in S3 but I want to lock in Consul because I don't want to complicate my world with DynamoDB, or because (continuing the last use-case) I want other non-terraform processes to be able to hold the lock for operations that affect Terraform. (e.g. an admin cron that deletes old AMIs that don't appear to be used anymore, but which Terraform is simultaneously trying to start using.)

Having a default seems important for UX, but having the ability to override in more complex cases would be useful, I think.

Probably none of this for this first pass though. :)

jbardin:
@mitchellh,

 - symlinks are atomic on POSIX systems, which is why they are sort of the standard to avoid the issues surrounding inconsistent support for locking in unix systems. I'm happy to change to `fcntl()` and `LockFileEx` as long as we can accept the issue that some shared filesystems won't support it, and the user should beware. 

- I'm OK with removing expiration. It was only in there as an informational hint, but since it requires user intervention anyway they may as well use their best judgement based on the initial lock time.

- I agree that we shouldn't block on the lock and immediately return an error. It makes the implementation semantics easier, and we can always retry. 


@apparentlymart,

I also like that idea, and was toying around with that while implementing the locks (there was a lot more code that's no longer here, as it ended up conflicting with the new Backends). I think that this should will eventually be able to extend to that use case, if only by separating the lock implementations and making them configurable.

jbardin:
Re-wrote the local locks to use POSIX and Windows file locks. This turned out to be quite a bit more involved than I expected, because the state files were often deleted and/or recreated, and locking requires that we keep track of the locked file handles. 

I'd like to review and merge the PR at this stage, since it's getting rather large and unwieldy. 

The commands don't hook into the state.Lock methods yet at all, so the goal here is to have this PR maintain existing behavior. A notable change in behavior is that the state output file is almost always created, since that is what will be locked when using local state. The backend tests had to be modified to check for non-empty state files, rather than simply checking for their existence. 

I'll start the work of enabling the locks and the associated commands in a new PR.

state/local_lock_windows.go
```diff
@@ -0,0 +1,140 @@
...
+		nil,
+		syscall.OPEN_EXISTING,
+		syscall.FILE_ATTRIBUTE_NORMAL,
+		0,
+	)
+	if err != nil {
+		return err
+	}
+
+	lockedFiles[s.stateFileOut] = handle
```
mitchellh: We probably need to wrap this in a lock. It'd be surprising as a consumer to learn you can't lock two seemingly distinct local states in parallel.

state/local_lock_windows.go
```diff
@@ -0,0 +1,140 @@
...
+	modkernel32      = syscall.NewLazyDLL("kernel32.dll")
+	procLockFileEx   = modkernel32.NewProc("LockFileEx")
+	procCreateEventW = modkernel32.NewProc("CreateEventW")
+
+	lockedFiles = map[*os.File]syscall.Handle{}
+)
+
+const (
+	_LOCKFILE_FAIL_IMMEDIATELY = 1
+	_LOCKFILE_EXCLUSIVE_LOCK   = 2
```
mitchellh: I generally find it useful when interacting with DLLs and constants like this to put the MSDN link above to the docs so that we can always easily look it back up.

jbardin:
added mutex and MSDN link

Issue #11173, opened 2017-01-12
No validation performed on domain names for cloudflare_record


jleclanche:
cloudflare_record currently does not perform any validation on its value.

Take the example where you mistakenly add an A record instead of a CNAME record:

```hcl
resource "cloudflare_record" "invalid_a_record" {
	domain = "example.org"
	name = "invalid"
	type = "A"
	value = "test.example.com"
}
```

PR #11197, opened 2017-01-13
provider/cloudflare: Add validation for record types and record content

grubernaut:
Adds a validation function for cloudflare record types. Also adds an apply-time validation for the record's content based on record type. Currently only validating `A` and `AAAA` records, can be expanded to verify record content for every possible record type in the future.

```
$ make test TEST=./builtin/providers/cloudflare
==> Installing errcheck...
2017/01/13 14:41:37 Generated command/internal_plugin_list.go
TF_ACC= go test ./builtin/providers/cloudflare  -timeout=30s -parallel=4
ok      github.com/hashicorp/terraform/builtin/providers/cloudflare     0.018s
```

Fixes: #11173


stack72:
LGTM! Nice work :)

PR #11214, opened 2017-01-15
'user' is not a valid NewRelic alert channel type

pickgr:
"user" does not appear to be a valid channel type:

```
Error applying plan:

1 error(s) occurred:

* newrelic_alert_channel.foo-user: Invalid channel type
```


pickgr:
FYI @paultyng

stack72:
LGTM! Thanks for this change @pickgr :)

PR #11217, opened 2017-01-15
provider/aws: Add Inspector Support

DaveBlooman:
Ready for reviewing now, added tests and docs for 3 new resources.  

The API is a little odd in that it doesn't allow for deletions on resource groups, so i just set ID and the ARN to an empty string.  If this is not the right approach for resources without delete options, let me know.


builtin/providers/aws/resource_aws_inspector_assessment_target.go
```diff
@@ -0,0 +1,117 @@
...
+		Update: resourceAwsInspectorAssessmentTargetUpdate,
+		Delete: resourceAwsInspectorAssessmentTargetDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"arn": &schema.Schema{
```
stack72: Can ARN be anything other than Computed?

builtin/providers/aws/resource_aws_inspector_assessment_template.go
```diff
@@ -0,0 +1,135 @@
...
+				ForceNew: true,
+			},
+			"target_arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
stack72: Can ARN be anything other than Computed?

builtin/providers/aws/resource_aws_inspector_assessment_template.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return resourceAwsInspectorAssessmentTemplateRead(d, meta)
+}
+
+func resourceAwsInspectorAssessmentTemplateRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).inspectorconn
+
+	resp, err := conn.DescribeAssessmentTemplates(&inspector.DescribeAssessmentTemplatesInput{
+		AssessmentTemplateArns: []*string{
+			aws.String(d.Get("arn").(string)),
```
stack72: I would tend to use d.Id() here

builtin/providers/aws/resource_aws_inspector_assessment_template.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		if inspectorerr, ok := err.(awserr.Error); ok && inspectorerr.Code() == "InvalidInputException" {
+			return nil
+		} else {
+			log.Printf("[ERROR] Error finding Inspector Assessment Template: %s", err)
+			return err
+		}
+	}
+
+	d.Set("name", *resp.AssessmentTemplates[0].Name)
```
stack72: I think we need to check if AssessmentTemplates is nil || empty before we do this assignment or we could get a panic

also note that d.Set handles dereferencing for you

builtin/providers/aws/resource_aws_inspector_assessment_template.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return nil
+}
+
+func resourceAwsInspectorAssessmentTemplateUpdate(d *schema.ResourceData, meta interface{}) error {
+	readTemplateErr := resourceAwsInspectorAssessmentTemplateRead(d, meta)
```
stack72: I am not sure what the point of the Update func here is - should the schema be ForceNew and Terraform will handle this for you?

builtin/providers/aws/resource_aws_inspector_assessment_template.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return resourceAwsInspectorAssessmentTemplateCreate(d, meta)
+}
+
+func resourceAwsInspectorAssessmentTemplateDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).inspectorconn
+
+	_, err := conn.DeleteAssessmentTemplate(&inspector.DeleteAssessmentTemplateInput{
+		AssessmentTemplateArn: aws.String(d.Get("arn").(string)),
```
stack72: I would tend to use d.Id()

builtin/providers/aws/resource_aws_inspector_assessment_template.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+	if err != nil {
+		if inspectorerr, ok := err.(awserr.Error); ok && inspectorerr.Code() == "AssessmentRunInProgressException" {
```
stack72: Do we want to retry the delete when the AssessmentRun isn't in progress?

builtin/providers/aws/resource_aws_inspector_resource_group.go
```diff
@@ -0,0 +1,87 @@
...
+
+func resourceAWSInspectorResourceGroup() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsInspectorResourceGroupCreate,
+		Read:   resourceAwsInspectorResourceGroupRead,
+		Update: resourceAwsInspectorResourceGroupUpdate,
+		Delete: resourceAwsInspectorResourceGroupDelete,
+
+		Schema: map[string]*schema.Schema{
+			"tags": &schema.Schema{
```
stack72: I think we should use:

```
"tags": tagsSchema(),
```

builtin/providers/aws/resource_aws_inspector_resource_group.go
```diff
@@ -0,0 +1,87 @@
...
+				ForceNew: true,
+				Type:     schema.TypeMap,
+				Required: true,
+			},
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
stack72: Can this even be optional?

builtin/providers/aws/resource_aws_inspector_resource_group.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return resourceAwsInspectorResourceGroupRead(d, meta)
+}
+
+func resourceAwsInspectorResourceGroupRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).inspectorconn
+
+	_, err := conn.DescribeResourceGroups(&inspector.DescribeResourceGroupsInput{
+		ResourceGroupArns: []*string{
+			aws.String(d.Get("arn").(string)),
```
stack72: I tend to use d.Id() here

builtin/providers/aws/resource_aws_inspector_resource_group.go
```diff
@@ -0,0 +1,87 @@
...
+			log.Printf("[ERROR] Error finding Inspector resource group: %s", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func resourceAwsInspectorResourceGroupUpdate(d *schema.ResourceData, meta interface{}) error {
+	createErr := resourceAwsInspectorResourceGroupCreate(d, meta)
```
stack72: I am not sure why we need an Update here?

builtin/providers/aws/resource_aws_inspector_resource_group.go
```diff
@@ -0,0 +1,87 @@
...
+	if createErr != nil {
+		return createErr
+	}
+
+	return nil
+}
+
+func resourceAwsInspectorResourceGroupDelete(d *schema.ResourceData, meta interface{}) error {
```
stack72: Does this leave resources in AWS? If so, do these resources cost money?

builtin/providers/aws/resource_aws_inspector_assessment_target.go
DaveBlooman: No, its the ARN generated by AWS.

builtin/providers/aws/resource_aws_inspector_assessment_target.go
```diff
@@ -0,0 +1,117 @@
...
+			},
+		},
+	}
+}
+
+func resourceAwsInspectorAssessmentTargetCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).inspectorconn
+
+	targetName := d.Get("name").(string)
+	resourceGroupArn := d.Get("resource_group_arn").(string)
```
DaveBlooman: I have changed the field to required, so now it should fail validation on plan if resource group isn't set.

builtin/providers/aws/resource_aws_inspector_assessment_target.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		if inspectorerr, ok := err.(awserr.Error); ok && inspectorerr.Code() == "InvalidInputException" {
+			return nil
+		} else {
+			log.Printf("[ERROR] Error finding Inspector Assessment Target: %s", err)
+			return err
+		}
+	}
+
+	d.Set("name", *resp.AssessmentTargets[0].Name)
```
DaveBlooman: If the input ARN isn't valid, the error will be caught, if any other error occurs, that should be caught.  I could set to this to the name input, rather than the name response from the API, which is preferred in this instance?

builtin/providers/aws/resource_aws_inspector_assessment_target.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return nil
+}
+
+func resourceAwsInspectorAssessmentTargetUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).inspectorconn
```
DaveBlooman: OK, I have made changes for review

builtin/providers/aws/resource_aws_inspector_assessment_target.go
```diff
@@ -0,0 +1,117 @@
...
+
+	_, err := conn.UpdateAssessmentTarget(&inspector.UpdateAssessmentTargetInput{
+		AssessmentTargetArn:  aws.String(d.Get("arn").(string)),
```
DaveBlooman: Changed

builtin/providers/aws/resource_aws_inspector_assessment_template.go
DaveBlooman: As above, need on advice, can either add a check or set to input name

builtin/providers/aws/resource_aws_inspector_assessment_template.go
DaveBlooman: I wasn't sure how to do this correctly. Do I still implement an update func, or just return a read to validate that a new template has been created?

builtin/providers/aws/resource_aws_inspector_assessment_template.go
DaveBlooman: Assessment runs can lasts for hours, so I would say no.  AWS recommend the minimum also be an hour, so in this instance, the error message should be enough to inform the user they need to stop the run, or wait for completion before this change.

builtin/providers/aws/resource_aws_inspector_resource_group.go
DaveBlooman: When `"tags": tagsSchema()` is used without an Update resource, tests fail due to force new not being set on tags.  What is the correct approach here?

builtin/providers/aws/resource_aws_inspector_resource_group.go
DaveBlooman: You are only charged for assessment runs, not resources created. https://aws.amazon.com/inspector/pricing/

DaveBlooman:
I have left some questions, so have made most of the changes requested. Will complete once I have guidance on the others.

Thanks for the review 👍

builtin/providers/aws/resource_aws_inspector_assessment_target.go
stack72: Ok, we should remove the Optional: true :)

builtin/providers/aws/resource_aws_inspector_assessment_target.go
stack72: 👍

builtin/providers/aws/resource_aws_inspector_assessment_target.go
stack72: Usually, we wrap this as follows:

if resp.AssessmentTargets != nil && len(resp.AssessmentTargets) > 0 {
    d.Set("name", resp.AssessmentTargets[0].Name)
}

builtin/providers/aws/resource_aws_inspector_assessment_template.go
stack72: So I think the issue we will have here is that during our acceptance tests, we will spin up an inspector and then try to destroy it. This will fail and cause our tests to fail

builtin/providers/aws/resource_aws_inspector_resource_group.go
stack72: If tags cannot be updated for this, then we should ignore my comments

stack72:
Responses left to most of your questions :)

builtin/providers/aws/resource_aws_inspector_resource_group.go
DaveBlooman: Sorry, i meant within the code.  So 

```hcl
func resourceAWSInspectorResourceGroup() *schema.Resource {
	return &schema.Resource{
		Create: resourceAwsInspectorResourceGroupCreate,
		Read:   resourceAwsInspectorResourceGroupRead,
		Delete: resourceAwsInspectorResourceGroupDelete,

			"tags": tagsSchema(),
			"arn": &schema.Schema{
				Type:     schema.TypeString,
				Computed: true,
			},
		},
	}
}
```
will fail with `: No Update defined, must set ForceNew on: []string{"tags"}`

DaveBlooman:
Everything except the update Funcs on resourceGroup and assessmentTemplate have been changed.  Does terraform require all resources to have CRUD?

stack72:
@DaveBlooman excellent - if there is no update, then we can simply omit Update BUT all values would have to be ForceNew: true at that point

P.

builtin/providers/aws/resource_aws_inspector_assessment_template.go
stack72: I still believe ARN should be Computed here NOT Optional

DaveBlooman:
OK, i've pushed up, could take a quick look at the retry logic in `resource_aws_inspector_assessment_target.go`

Also, I have set `resource_aws_inspector_resource_group.go` to use my original method for tags as tagsSchema() does not use ForceNew.

stack72:
Hi @DaveBlooman 

This will only retry for a max of 1 minute, that correct?

P.

stack72:
This LGTM now @DaveBlooman :) Thanks for all the work getting through the feedback

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSInspector'                                                           ✭
2017/01/17 17:27:30 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSInspector -timeout 120m
=== RUN   TestAccAWSInspectorTarget_basic
--- PASS: TestAccAWSInspectorTarget_basic (37.87s)
=== RUN   TestAccAWSInspectorTemplate_basic
--- PASS: TestAccAWSInspectorTemplate_basic (43.06s)
=== RUN   TestAccAWSInspectorResourceGroup_basic
--- PASS: TestAccAWSInspectorResourceGroup_basic (29.91s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	110.890s
```

joe-zhou-ripple:
Anyway to add schedules, right now the support just create the static target and template WITHOUT schedule, so basically the assessment never runs...

PR #11223, opened 2017-01-16
Update interpolation.html.md

telepath:
correct typo


stack72:
Thanks for this @telepath :)

P.

PR #11229, opened 2017-01-16
build: Update travis for go1.8rc1

cblecker:
Update travis to use go1.8rc1 for builds instead of go1.8beta2.


stack72:
LGTM! Thanks :)

PR #10874, opened 2016-12-21
Fix contradiction in terms with provider documentation (#10815)

OJFord:
This PR fixes a contradiction in the API access terminology used by Scaleway and Terraform; each previously using `access_key` for different parameters. Discussion in #10815.

I've also included a removal of 'ARM' from the documentation (e.g. 'ARM server') to reflect the fact that Scaleway now also provides x86 servers.

PR #11234, opened 2017-01-17
provider/scaleway: fix provider configuration

nicolai86:
properly fetch configuration from env - leftover from #10874


stack72:
Thanks for fixing this up so fast :)

PR #11250, opened 2017-01-18
provider/openstack: Increase deletion timeout for router interfaces

### LINKED ISSUE #11192: OpenStack Neutron Router Interface deletion: timeout too short
Hi,

I have a working deployment using Terraform v0.8.2 against an OpenStack installation.
When deleting, I always get 

```
* openstack_networking_router_interface_v2.main: Error deleting OpenStack Neutron Router Interface: timeout while waiting for state to become 'DELETED' (last state: 'ACTIVE', timeout: 2m0s
```

and on re-running, the deletion of the OpenStack Neutron Router Interface succeeds,
so total deletion time is probably <4min. 

I don't have a PR for @jtopjian , (who asked for it in https://github.com/hashicorp/terraform/pull/4223#issuecomment-163113311) but I think the change is required here:
https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_networking_router_interface_v2.go#L73

Yours,
Steffen

jtopjian:
Fixes #11192


stack72:
LGTM!

PR #11256, opened 2017-01-18
clarify text: existing -> user-supplied

tamsky:
I think this clarifies the text.
The source of the public key material is not existing (which could be interpreted through the lens that existing resources are eligible for `terraform import`).


stack72:
Hi @tamsky

Paul

Issue #8423, opened 2016-08-23
Adding Route53 records with nested double quotes escaped, causes Terraform plan to always show the resource as modified.


c4milo:
### Background

Route53 does not accept record values longer than 255 characters. This makes adding [DKIM](https://en.wikipedia.org/wiki/DomainKeys_Identified_Mail) TXT records a little bit tricky. However, it seems there is a way to do it by splitting the value using `[double quoute][space][double quote]` as explained in https://forums.aws.amazon.com/thread.jspa?threadID=93940

I was able to create the DKIM record by escaping the nested double qoutes, but Terraform plan keeps showing the resource as modified every time. 
### Terraform Version

v0.7.2-dev
### Affected Resource(s)
- aws_route53_record
### Terraform Configuration Files

``` hcl
resource "aws_route53_record" "google_dkim" {
    zone_id = "${aws_route53_zone.primary.zone_id}"
    name = "google._domainkey"
    type = "TXT"
    ttl = "300"
    records = [
        "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAiajKNMp\" \"/A12roF4p3MBm9QxQu6GDsBlWUWFx8EaS8TCo3Qe8Cj0kTag1JMjzCC1s6oM0a43JhO6mp6z/\" \"MzVf01GWGxxvrtv0lJLHBjWSQqDTQwAZAsa/eKUrsDTbTHYDgXwKeYB2W192oNOTf1oy/\" \"CHvvDsPqnavOYBg8/y18SKjfrc768n4t41ZZcs6GRX4erMlspsUkS6cmSpsyu/\" \"OpMzM1qq05PduKs26VfD2r+mQYdB6D1QOeYodhhNNFxRxMZeepbh/\" \"SRiV3AbLlawTYJbpKWuFfBGT0j1ieNKHVyNdssYf4DWveP2pjNs0MePx/hVASkCM3JkXH8MfrtF/nwIDAQAB"
    ]
}
```
### Expected Behavior

Terraform plan should not show the resource as modified as the record is already created in Route53.
### Actual Behavior

Terraform plan shows the resource as modified despite the record value being the same.
### Steps to Reproduce
1. Just try creating a Route53 TXT record with a value longer than 255 characters, split as explained above.

cakoose:
[...]

PR #11257, opened 2017-01-18
aws_route53_record: Make unquoting slightly less hacky.

cakoose:
Before, the code would remove up to the first two quotes.  Now it will
remove the first and last quote if they are both present.

There are still a bunch of problems with the way quoting is handled, but
I believe this change makes the situation slightly better.  In
particular, it helps with hashicorp/terraform#8423.

### DIFF
--- a/builtin/providers/aws/structure.go
+++ b/builtin/providers/aws/structure.go
@@ -814,11 +814,14 @@ func flattenStepAdjustments(adjustments []*autoscaling.StepAdjustment) []map[str
 	return result
 }
 
-func flattenResourceRecords(recs []*route53.ResourceRecord) []string {
+func flattenResourceRecords(recs []*route53.ResourceRecord, typeStr string) []string {
 	strs := make([]string, 0, len(recs))
 	for _, r := range recs {
 		if r.Value != nil {
-			s := strings.Replace(*r.Value, "\"", "", 2)
+			s := *r.Value
+			if typeStr == "TXT" || typeStr == "SPF" {
+				s = strings.Replace(s, "\"", "", 2)
+			}
 			strs = append(strs, s)
 		}
 	}
@@ -829,13 +832,11 @@ func expandResourceRecords(recs []interface{}, typeStr string) []*route53.Resour
 	records := make([]*route53.ResourceRecord, 0, len(recs))
 	for _, r := range recs {
 		s := r.(string)
-		switch typeStr {
-		case "TXT", "SPF":
-			str := fmt.Sprintf("\"%s\"", s)
-			records = append(records, &route53.ResourceRecord{Value: aws.String(str)})
-		default:
-			records = append(records, &route53.ResourceRecord{Value: aws.String(s)})
+		if typeStr == "TXT" || typeStr == "SPF" {
+			// `flattenResourceRecords` removes quotes.  Add them back.
+			s = fmt.Sprintf("\"%s\"", s)
 		}
+		records = append(records, &route53.ResourceRecord{Value: aws.String(s)})
 	}
 	return records
 }
--- a/builtin/providers/aws/structure_test.go
+++ b/builtin/providers/aws/structure_test.go
@@ -819,23 +819,56 @@ func TestFlattenStepAdjustments(t *testing.T) {
 }
 
 func TestFlattenResourceRecords(t *testing.T) {
-	expanded := []*route53.ResourceRecord{
-		&route53.ResourceRecord{
-			Value: aws.String("127.0.0.1"),
-		},
-		&route53.ResourceRecord{
-			Value: aws.String("127.0.0.3"),
-		},
+	original := []string{
+		`127.0.0.1`,
+		`"abc def"`,
 	}
 
-	result := flattenResourceRecords(expanded)
+	dequoted := []string{
+		`127.0.0.1`,
+		`abc def`,
+	}
+
+	var wrapped []*route53.ResourceRecord = nil
+	for _, original := range original {
+		wrapped = append(wrapped, &route53.ResourceRecord{Value: aws.String(original)})
+	}
+
+	sub := func(recordType string, expected []string) {
+		t.Run(recordType, func(t *testing.T) {
+			checkFlattenResourceRecords(t, recordType, wrapped, expected)
+		})
+	}
+
+	// These record types should be dequoted.
+	sub("TXT", dequoted)
+	sub("SPF", dequoted)
+
+	// These record types should not be touched.
+	sub("CNAME", original)
+	sub("MX", original)
+}
+
+func checkFlattenResourceRecords(
+	t *testing.T,
+	recordType string,
+	expanded []*route53.ResourceRecord,
+	expected []string) {
+
+	result := flattenResourceRecords(expanded, recordType)
 
 	if result == nil {
 		t.Fatal("expected result to have value, but got nil")
 	}
 
-	if len(result) != 2 {
-		t.Fatal("expected result to have value, but got nil")
+	if len(result) != len(expected) {
+		t.Fatalf("expected %v, got %v", expected, result)
+	}
+
+	for i, e := range expected {
+		if result[i] != e {
+			t.Fatalf("expected %v, got %v", expected, result)
+		}
 	}
 }
 
--- a/builtin/providers/aws/resource_aws_route53_record.go
+++ b/builtin/providers/aws/resource_aws_route53_record.go
@@ -484,7 +484,7 @@ func resourceAwsRoute53RecordRead(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	err = d.Set("records", flattenResourceRecords(record.ResourceRecords))
+	err = d.Set("records", flattenResourceRecords(record.ResourceRecords, *record.Type))
 	if err != nil {
 		return fmt.Errorf("[DEBUG] Error setting records for: %s, error: %#v", d.Id(), err)
 	}


cakoose:
Hey, is there anything in particular that's holding up the acceptance of this pull request?  If so, please let me know and I'll try to help out.

cakoose:
Is there any chance this will get merged?  If not, please let me know and I'll try to solve my problem another way.

builtin/providers/aws/structure.go
```diff
@@ -805,11 +805,17 @@ func flattenStepAdjustments(adjustments []*autoscaling.StepAdjustment) []map[str
...
 	for _, r := range recs {
 		if r.Value != nil {
-			s := strings.Replace(*r.Value, "\"", "", 2)
+			s := *r.Value
+			if typeStr == "TXT" || typeStr == "SPF" {
+				l := len(s)
+				if l > 1 && s[0] == '"' && s[l-1] == '"' {
+					s = s[1:(l - 1)]
+				}
+			}
```
zimbatm: It would make sense to add a comment to explain why this is required

cakoose: Yeah, good idea.  Done.

pkoch:
Any idea when this can get to master/be released?

zimbatm:
Unfortunately I don't have merging rights.

/cc @jbardin, @mitchellh, ??

builtin/providers/aws/structure.go
```diff
@@ -813,11 +813,20 @@ func flattenStepAdjustments(adjustments []*autoscaling.StepAdjustment) []map[str
...
-			s := strings.Replace(*r.Value, "\"", "", 2)
+			s := *r.Value
+			if typeStr == "TXT" || typeStr == "SPF" {
+				// On Route53, these records are usually quoted.  This plugin defines the
+				// canonical format to be unquoted.  So when converting from the Route53 value
+				// to the Terraform config value, remove quotes.
+				l := len(s)
+				if l > 1 && s[0] == '"' && s[l-1] == '"' {
+					s = s[1:(l - 1)]
+				}
```
radeksimko: Apologies if this will sound like nitpicking... wouldn't [`strings.Trim`](https://golang.org/pkg/strings/#Trim) make this look less complex and cleaner?

builtin/providers/aws/structure_test.go
```diff
@@ -824,18 +824,29 @@ func TestFlattenResourceRecords(t *testing.T) {
...
 		},
 	}
 
-	result := flattenResourceRecords(expanded)
+	expected := []string{
+		"127.0.0.1",
+		"abc def\" \"xxx yyy",
+	}
+
+	result := flattenResourceRecords(expanded, "TXT")
```
radeksimko: I think we should also test at least 1 other record type which will remain as is (i.e. verify that we're not removing quotes in such case and won't in the future).

Would you mind adding such test?

builtin/providers/aws/structure.go
cakoose: When inputs are well-formed, I think `strings.Trim` would be equivalent.  Sadly, the plugin currently doesn't strictly validate inputs, so I'm worried about the behavior on strange inputs (e.g. `abc "def"` or `"abc" ""`).

My thinking was that the current code is slightly more predictable: if there are quotes at both ends, remove them, otherwise do nothing.

That said, maybe we should modify this plugin to be more strict about inputs?
- When pulling from Route53, remove all quotes and create a single string.
- When pushing to Route53, split on the maximum length and re-quote every segment.

Two downsides:
- I *think* this might have to be a backwards-incompatible change.  But maybe it's worth it?
- Values won't round-trip verbatim, but they'll be semantically equivalent.

builtin/providers/aws/structure_test.go
cakoose: Will add.

zimbatm:
Are quotes are a valid value in TXT records?

After thinking about this for a while, I would rather have the `aws_route53_record` fail when quotes are given as input and take care of the chunking for me. As a user I don't want to know about these AWS particularities. When comparing the state it should get the quotes version from the AWS API, remove all the quotes and concat the string, compare that against the input.

builtin/providers/aws/structure_test.go
```diff
@@ -824,18 +824,46 @@ func TestFlattenResourceRecords(t *testing.T) {
...
+	expected_verbatim := []string{
+		"127.0.0.1",
+		"abc def\" \"xxx yyy",
+	}
+
+	checkFlattenResourceRecords(t, "TXT", expanded, expected_unquoted)
+	checkFlattenResourceRecords(t, "SPF", expanded, expected_unquoted)
+
+	checkFlattenResourceRecords(t, "CNAME", expanded, expected_verbatim)
+	checkFlattenResourceRecords(t, "MX", expanded, expected_verbatim)
```
radeksimko: This new test is now failing:

```
--- FAIL: TestFlattenResourceRecords (0.00s)
	structure_test.go:865: expected [127.0.0.1 abc def" "xxx yyy], got [127.0.0.1 "abc def" "xxx yyy"]
```

A side note - it would be easier to identify the failing test here if we used named [subtests](https://blog.golang.org/subtests).

cakoose: Ah, a stupid mistake in the test.  Will fix.

And thanks for the tip about subtests.  Will switch to using them.  (I didn't realize that tests failures don't show the whole stack.)

cakoose:
> It looks like this is already a breaking change then.https://play.golang.org/p/jM_nofpCuV
>
> While I believe that this (un)quoting logic may be worth changing I feel like it's a topic for a separate PR where we can also discuss the expected/right TXT format and impact on the user.
>
> Do you agree?

I'm probably less familiar with Terraform's policy on this stuff, but I think it's possible to view this change as *not* being a breaking change.

For example: https://github.com/hashicorp/terraform/issues/8423.  With the old code, people just wouldn't be able to sync those records.  So yes, the behavior changes for the input string `"abc" "def"`, but the old behavior would just cause Terraform to not work.

I could be mistaken, but I believe that the new code only changes the behavior in cases where the old behavior resulted in Terraform not working.  So I don't think it will break any working cases.

cakoose:
Uh oh, I just realized a problem with this PR.  I think merging it will make it more annoying to perform a "proper" fix.  Here's why:

For the Route 53 value `"abc" "def"`:
- Current behavior: unquote to `abc "def"`, requote to `"abc "def""`.  This is broken.
- Behavior in this PR: unquote to `abc" "def`, requote to `"abc" "def"`.  This is hacky, but works.
- "Ideal" behavior: unquote to `abcdef`, requote to `"abcdef"` (or any valid split, e.g. `"abcd" "ef"`)

By my logic of what constitutes a "breaking change", it's ok to change the behavior for `"abc" "def"`, because the current behavior causes Terraform do something that isn't useful, so nobody would rely on it (I believe).

If we merge this PR, people can now sync the value `"abc" "def"`.  This would appear as `abc" "def` in the config and state.  But now that Terraform is doing something useful for this input, people will start relying on it.

Later on, if we want to implement the "ideal" solution, we can no longer change the behavior for `"abc" "def"`.  If we jump straight to the ideal solution now, I think it will be easier because we can change the behavior of `"abc" "def"`.

Does this make sense, @radeksimko?

radeksimko:
@cakoose Right, so I'd like to avoid any changes in the quoting behaviour _in this PR_, which means I'd like to keep `strings.Replace(*r.Value, "\"", "", 2)` there. I think the PR is still valuable without that change as it cleans the code. 🚿 👍 

I think the go playground clearly demonstrates the differences
https://play.golang.org/p/jM_nofpCuV

Does that make sense? 😃

cakoose:
Totally agree that the new behavior is different.  My goal for this PR _was_ to change the behavior.  The `strings.Replace` call is what I considered to be "hacky" since it seemed like a lazy way of unquoting that didn't work in all cases.

I only now realize that you see the record type check as the important part of this PR.  I am in favor of this check because the old code did the wrong thing and people probably shouldn't have quotes in other record types.  But technically isn't this also a change in behavior?

radeksimko:
> But technically isn't this also a change in behavior?

Practically no, assuming cases where users were managing the R53 record in Terraform and didn't modify it outside. The flattener is only responsible for flattening API struct into state to compare it with config. Expander is responsible for the real stuff (creating API structs from config) - and that one functionally remains as is (it already had `switch type { case "TXT", "SPF`).

radeksimko:
> I would rather have the aws_route53_record fail when quotes are given as input and take care of the chunking for me. As a user I don't want to know about these AWS particularities.

I'm not convinced there's any way for us to reliably decide about what/whether to enclose in quotes, see the below example:

```hcl
resource "aws_route53_record" "www" {
  zone_id = "${aws_route53_zone.primary.zone_id}"
  name    = "notexample.com"
  type    = "TXT"
  ttl     = "300"
  records = ["one\" \"two"]
}

resource "aws_route53_zone" "primary" {
  name = "notexample.com"
}
```

Setting the complexity of HCL escaping aside (imagine the user input was `one" "two`), the question is - did the user want `"one \" \"two"` or `"one" "two"`? Probably the latter, but how do we know for sure?

Terraform's behaviour here is admittedly slightly confusing because it tries to do some work for the user already and encloses the whole value in outer quotes. So `one" "two` becomes `"one" "two"` when it's sent to the R53 API and that is in turn accepted as valid record.

One thing we could do is to allow outer quotes (to make the above example cleaner), but avoid double quoting. i.e. `s = fmt.Sprintf("\"%s\"", strings.Trim(s, `"`))`. That _should_ deal with the example from https://github.com/hashicorp/terraform/issues/8423 , hopefully?

cakoose:
Currently, values with nested quotes don't work correctly.  Terraform will keep showing the config as modified (#8423).  So maybe it's safe to assume that nobody is successfully using a value like `one" "two`, which means we can now decide how it should be interpreted?

Options:
1. Interpret as `one" "two`: Let's the user control where the quotes are placed, but is sort of ugly.  (This is what my original PR did.)
2. Interpret as `one\" \"two`: The user doesn't have to worry about quoting, but also *can't* control where the quotes are placed.  Terraform would split long strings into multiple quoted strings automatically.

If I put `"one" "two"` into the Route53 config UI, doing a query with `dig` gives me `"one" "two"`.  So it looks like clients can tell where you put the quotes.  This makes it seem like option 1 is the safe route.  I'd hate for some poor Terraform user to run into a situation where they need to control the quote placement but are now unable to.

On the other hand, the SPF spec ([https://tools.ietf.org/html/rfc4408#section-3.1.3](RFC 4408, §3.1.3)) says that a client should merge the quoted strings before interpreting the record.  I can't track down what the DNS spec itself says about this situation.  Maybe if we can get high confidence that quote placement never matters, we can make Terraform handle it automatically?

radeksimko:
@cakoose I'm in favour of (1) too, generally but I think we can support both use cases.

----

I was thinking about the differences between `strings.Trim` and your original implementation that literally cuts first and last `"`.

The edge case where it probably matters the most is `"abc" ""` - `strings.Trim` would cut off the two consecutive quotes from the right + the first one. Then by adding quotes back while sending it to the API we would get `"abc" "` which is invalid.

Your function would trim it to `abc" "` which admittedly isn't a valid record either, but by adding quotes back we'd just get the original, valid record, which is desired (I think).

I originally thought `strings.Trim` would prevent excessive use of quotes (e.g. treat `"""myexample""` as valid record), but it probably brings more confusion than your solution and makes thing way too magic, so I'm in favour of your custom trimming function.

-------

> On the other hand, the SPF spec says that a client should merge the quoted strings before interpreting the record (RFC 4408, §3.1.3).

That is a very good point, admittedly Route53 API would reject anything over 255 bytes as they seem to do the merge on their side and expect the client to figure it out. It would be nice if Terraform was able to do this behind the scenes for the user, so they can keep long strings (e.g. DKIM) as consecutive strings in the HCL config - which is kind of what @zimbatm suggested and it doesn't go against quotes _inside_ the record (unlike what I thought originally) - these would just need to be escaped properly - due to HCL twice, not amazing but hopefully there isn't too many such use cases anyway.

I'm not against returning errors if the user has entered quote, but only if it's literal quote (`"`), not an escaped one (`\"`) - here we're talking post-HCL-processed value, so this should be accepted as **valid** thing:

```hcl
resource "aws_route53_record" "www" {
  zone_id = "${aws_route53_zone.primary.zone_id}"
  name    = "notexample.com"
  type    = "TXT"
  ttl     = "300"
  records = ["one\\\" \\\"two"]
}
```
and this can be raised as **invalid**:
```hcl
resource "aws_route53_record" "www" {
  zone_id = "${aws_route53_zone.primary.zone_id}"
  name    = "notexample.com"
  type    = "TXT"
  ttl     = "300"
  records = ["one\" \"two"]
}
```
and the error message should be clearly suggesting the user to either remove or escape quotes.

I'm happy for the implementation to split values by `" "`, but _not_ `\" "` (maybe it's obvious, maybe not but I thought it should be mentioned). The RFC mentions "spaces", so I'd assume any number of consecutive spaces. Also we should probably use `StateFunc` for this and always store the value as consecutive characters and only split it in `Create()` and `Update()` prior to sending it to the API. That way we can avoid any spurious diffs and expectations of what the API returns.

-----

I'm happy to review a separate PR for this - if we roughly agree here. I'll make sure it's labelled as breaking change which was also the reason I wanted it to be separate - sorry if my request originally sounded too harsh 🙈

cakoose:
I don't totally follow your scheme to allow both use cases.  With this example:

```
records = ["one\" \"two"]
```

To enable option 1 (give users control over quotes), we should send the value to Route 53 as `"one" "two"`.  To enable option 2 (handle quotes automatically), it seems like we should reject that value.  Am I missing something?

Or maybe you saying that users who want option 1 should do:

```
records = ["one\\\" \\\"two"]
```

radeksimko:
> Or maybe you saying that users who want option 1 should do:

Yes, that's exactly what I meant. 👍

cakoose:
[FYI, for the examples in this comment, I'm omitting the extra level of HCL quoting.]

I'm worried that `abc\" \"def` is too confusing.  For one, the DNS spec already uses `\"` to insert a literal double-quote ([RFC 1464, §2](https://tools.ietf.org/html/rfc1464#section-2) see example `string="Cat"`).

**Alternative idea**

- Parse the value and identify all the quoted strings.
- For quoted strings that are less than 255 characters, leave them as-is.
- For quoted strings that are more than 255 characters, split them so they fit the limit.

So, for example, in the common case, lets say we had a TXT record that was 256 "x" characters:

```
xxx...xxx
```

When pushing to Route 53, we would break at 255 and produce:

```
"xxx...xx" "x"
```

This would work even if the user manually quotes some things:

```
abc" "xxx...xxx
```

Becomes:

```
"abc" "xxx...xx" "x"
```

Most of the time, the auto-splitting would be good enough for people.  But if someone ever needs control over the quoting, they still have a (slightly ugly) option.

radeksimko:
@cakoose that sounds like a good alternative idea. As long as we have expected edge cases covered by tests I'm fine with that. 😉  👍 

Do you plan to update the linked PR to reflect this?

cakoose:
I would prefer land the current PR as-is.  For one, it would unblock people (#8423).

Further, we can safely add splitting later.  For cases where the string is longer than 255 characters, Terraform won't sync it properly.  So if we add splitting later, it will change the semantics of those cases, but I think that's fine because they never worked in the first place.

cakoose:
BTW, I really like the idea of auto-splitting.  I actually started working on it, but ran into some edge cases around the [quoting rules](https://tools.ietf.org/html/rfc1464#section-2) that I didn't have time to investigate thoroughly.  If I screw something up, fixing it later might require a breaking change.

PR #11261, opened 2017-01-18
S3 Bucket Object Sever Side Encryption

mubeta06:
This PR sees support for server side encryption added to s3 bucket object resources. This functionality is equivalent to

```
aws s3 cp --sse aws:kms source destination
```

which facilitates using S3 default master key for server side encryption.

 Associated acceptance testing and documentation has been included.

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket_object.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket_object.go
@@ -89,6 +89,13 @@ func resourceAwsS3BucketObject() *schema.Resource {
 				ValidateFunc: validateS3BucketObjectStorageClassType,
 			},
 
+			"server_side_encryption": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validateS3BucketObjectServerSideEncryption,
+				ConflictsWith: []string{"kms_key_id"},
+			},
+
 			"kms_key_id": {
 				Type:         schema.TypeString,
 				Optional:     true,
@@ -171,9 +178,13 @@ func resourceAwsS3BucketObjectPut(d *schema.ResourceData, meta interface{}) erro
 		putInput.ContentDisposition = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("server_side_encryption"); ok {
+		putInput.ServerSideEncryption = aws.String(v.(string))
+	}
+
 	if v, ok := d.GetOk("kms_key_id"); ok {
 		putInput.SSEKMSKeyId = aws.String(v.(string))
-		putInput.ServerSideEncryption = aws.String("aws:kms")
+		putInput.ServerSideEncryption = aws.String(s3.ServerSideEncryptionAwsKms)
 	}
 
 	resp, err := s3conn.PutObject(putInput)
@@ -218,6 +229,7 @@ func resourceAwsS3BucketObjectRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("content_language", resp.ContentLanguage)
 	d.Set("content_type", resp.ContentType)
 	d.Set("version_id", resp.VersionId)
+	d.Set("server_side_encryption", resp.ServerSideEncryption)
 	d.Set("kms_key_id", resp.SSEKMSKeyId)
 	d.Set("etag", strings.Trim(*resp.ETag, `"`))
 
@@ -325,6 +337,22 @@ func validateS3BucketObjectStorageClassType(v interface{}, k string) (ws []strin
 			"%q contains an invalid Storage Class type %q. Valid types are either %q, %q, or %q",
 			k, value, s3.StorageClassStandard, s3.StorageClassReducedRedundancy,
 			s3.StorageClassStandardIa))
+	}
+	return
+}
+
+func validateS3BucketObjectServerSideEncryption(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+
+	serverSideEncryption := map[string]bool{
+		s3.ServerSideEncryptionAes256: true,
+		s3.ServerSideEncryptionAwsKms: true,
+	}
+
+	if _, ok := serverSideEncryption[value]; !ok {
+		errors = append(errors, fmt.Errorf(
+			"%q contains an invalid Server Side Encryption value %q. Valid values are %q and %q",
+			k, value, s3.ServerSideEncryptionAes256, s3.ServerSideEncryptionAwsKms))
 	}
 	return
 }
--- a/builtin/providers/aws/resource_aws_s3_bucket_object.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket_object.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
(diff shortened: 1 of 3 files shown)


stack72:
Hi @mubeta06

Thanks for the PR here - unfortunately this PR doesn't pass the acceptance tests:

```
=== RUN   TestAccAWSS3BucketObject_kms
--- FAIL: TestAccAWSS3BucketObject_kms (66.03s)

		DIFF:

		UPDATE: aws_s3_bucket_object.object
		  server_side_encryption: "aws:kms" => ""

		STATE:

		aws_kms_key.kms_key_1:
		  ID = 335963c0-210f-447f-a940-10982a37d2b8
		  arn = arn:aws:kms:us-west-2:187416307283:key/335963c0-210f-447f-a940-10982a37d2b8
		  description =
		  enable_key_rotation = false
		  is_enabled = true
		  key_id = 335963c0-210f-447f-a940-10982a37d2b8
		  key_usage = ENCRYPT_DECRYPT
		  policy = {"Id":"key-default-1","Statement":[{"Action":"kms:*","Effect":"Allow","Principal":{"AWS":"arn:aws:iam::187416307283:root"},"Resource":"*","Sid":"Enable IAM User Permissions"}],"Version":"2012-10-17"}
		aws_s3_bucket.object_bucket_2:
		  ID = tf-object-test-bucket-7612941759793338343
		  acceleration_status =
		  acl = private
		  arn = arn:aws:s3:::tf-object-test-bucket-7612941759793338343
		  bucket = tf-object-test-bucket-7612941759793338343
		  force_destroy = false
		  hosted_zone_id = Z3BJ6K6RIION7M
		  region = us-west-2
		  request_payer = BucketOwner
		  tags.% = 0
		  versioning.# = 1
		  versioning.0.enabled = false
		  versioning.0.mfa_delete = false
		  website.# = 0
		aws_s3_bucket_object.object:
		  ID = test-key
		  acl = private
		  bucket = tf-object-test-bucket-7612941759793338343
		  cache_control =
		  content = stuff
		  content_disposition =
		  content_encoding =
		  content_language =
		  content_type = binary/octet-stream
		  etag = 6e0a90a7169634120ebc69a8037322c2
		  key = test-key
		  kms_key_id = arn:aws:kms:us-west-2:187416307283:key/335963c0-210f-447f-a940-10982a37d2b8
		  server_side_encryption = aws:kms
		  storage_class = STANDARD
		  version_id =

		  Dependencies:
		    aws_s3_bucket.object_bucket_2
		    aws_kms_key.kms_key_1
=== RUN   TestAccAWSS3BucketObject_sse
--- FAIL: TestAccAWSS3BucketObject_sse (50.04s)

		DIFF:

		UPDATE: aws_s3_bucket_object.object
		  kms_key_id: "arn:aws:kms:us-west-2:187416307283:key/12610ebb-e6f8-4afc-981c-f96e00bc123c" => ""

		STATE:

		aws_s3_bucket.object_bucket_2:
		  ID = tf-object-test-bucket-7117879692753440864
		  acceleration_status =
		  acl = private
		  arn = arn:aws:s3:::tf-object-test-bucket-7117879692753440864
		  bucket = tf-object-test-bucket-7117879692753440864
		  force_destroy = false
		  hosted_zone_id = Z3BJ6K6RIION7M
		  region = us-west-2
		  request_payer = BucketOwner
		  tags.% = 0
		  versioning.# = 1
		  versioning.0.enabled = false
		  versioning.0.mfa_delete = false
		  website.# = 0
		aws_s3_bucket_object.object:
		  ID = test-key
		  acl = private
		  bucket = tf-object-test-bucket-7117879692753440864
		  cache_control =
		  content = stuff
		  content_disposition =
		  content_encoding =
		  content_language =
		  content_type = binary/octet-stream
		  etag = 5bce9549b498bfeede2cd1a219b59dcd
		  key = test-key
		  kms_key_id = arn:aws:kms:us-west-2:187416307283:key/12610ebb-e6f8-4afc-981c-f96e00bc123c
		  server_side_encryption = aws:kms
		  storage_class = STANDARD
		  version_id =

		  Dependencies:
		    aws_s3_bucket.object_bucket_2
```

I ran these using the command:

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSS3BucketObject_'
```

Please can you have a look at these?

thanks

Paul

mubeta06:
@stack72 i've managed to get the acceptance tests passing. I am reasonably unfamiliar with the codebase and, hence, am not 100% sure of the fix. It all seems to work however.

mubeta06:
@stack72 it'd be great to get some feedback on this if at all possible. this is a feature we are relying on.

builtin/providers/aws/resource_aws_s3_bucket_object.go
```diff
@@ -218,7 +229,11 @@ func resourceAwsS3BucketObjectRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("content_language", resp.ContentLanguage)
 	d.Set("content_type", resp.ContentType)
 	d.Set("version_id", resp.VersionId)
-	d.Set("kms_key_id", resp.SSEKMSKeyId)
+	if _, ok := d.GetOk("server_side_encryption"); ok {
+		d.Set("server_side_encryption", resp.ServerSideEncryption)
+	} else if _, ok := d.GetOk("kms_key_id"); ok {
+		d.Set("kms_key_id", resp.SSEKMSKeyId)
```
radeksimko: While this seems to be fixing the acceptance test which was previously failing 👍 , there's still one more edge case most resources always try to cover and we should cover here too.

If things change outside of Terraform (either by hand or via a different tool/script), Terraform should be able to detect such changes and pull down what's the real key ID in order to present it to the user in the diff. At this point your implementation would cause TF to ignore these changes.

Here comes the tricky part - if you don't specify `SSEKMSKeyId` attribute in the `Put` API request, AWS will find the default master S3 KMS key and then return that key ID back - i.e. empty field in `Put` != empty field in `Head` 😿 . To avoid `Computed` `kms_key_id` (because that would suppress the diff in certain cases) we could work around that by checking what the default Key ID for a given region is and then compare it here.

e.g.
```go
// Find out whether the default S3 KMS key is in use
out, err := kmsconn.DescribeKey(&kms.DescribeKeyInput{
	KeyId: aws.String("alias/aws/s3"),
})
if err != nil {
	return fmt.Errorf("Failed to describe default S3 KMS key (alias/aws/s3): %s", err)
}

// Only set non-default KMS key ID (one that doesn't match default)
if resp.SSEKMSKeyId != nil && *resp.SSEKMSKeyId != *out.KeyMetadata.Arn {
	log.Printf("[DEBUG] S3 object is encrypted using a non-default KMS Key ID: %s", *resp.SSEKMSKeyId)
	d.Set("kms_key_id", resp.SSEKMSKeyId)
}
```

The alias unfortunately isn't very well documented, but from what I can see it's always available - even if it's new  brand new AWS account - It's created for the first time you try interacting with it.

mubeta06: @radeksimko thanks for the feedback. That makes sense. I figured there was a better way of handling this. My grasp on the colloquial TF knowledge was too thin. I'll give those changes a go.

mubeta06:
@radeksimko requested changes have been made. thanks for the feedback. it anything else remains to get this over the line just let me know!

builtin/providers/aws/resource_aws_s3_bucket_object.go
```diff
@@ -218,7 +230,22 @@ func resourceAwsS3BucketObjectRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("content_language", resp.ContentLanguage)
 	d.Set("content_type", resp.ContentType)
 	d.Set("version_id", resp.VersionId)
-	d.Set("kms_key_id", resp.SSEKMSKeyId)
+	d.Set("server_side_encryption", resp.ServerSideEncryption)
+
+	// retrieve S3 KMS Default Master Key
+	kmsconn := meta.(*AWSClient).kmsconn
+	kmsresp, err := kmsconn.DescribeKey(&kms.DescribeKeyInput{
```
radeksimko: I think there's no reason to call any KMS API at all if the S3 bucket/object isn't encrypted, i.e. if `SSEKMSKeyId` is `nil`.

It may look like an optimisation nitpick, but it's also about what IAM permissions we expect/require the user to have - in this case _everyone_ using `aws_s3_bucket_object` would _always_ need `kms:DescribeKey` even though they may not be using the encryption at all.

mubeta06:
@radeksimko No problem makes perfect sense... anything to speed up the running of the tests 👍

radeksimko:
@mubeta06 This looks good to me.

Thanks a lot for all the changes and for making them in timely manner 👍

Issue #10713, opened 2016-12-14
AWS Provider Enhancement -- Attach AWS WAF policies to ALBs


chroto:
### Terraform Version
0.7.13

### Affected Resource(s)
- aws_alb

AWS announced that they now support WAF policies attached to ALBs ( https://aws.amazon.com/about-aws/whats-new/2016/12/AWS-WAF-now-available-on-Application-Load-Balancer/ )

This would be an amazing feature to have on the 0.7 branch!

4dz:
+1 definitely need this feature :)

goduncan:
+1 from me too. Any thoughts on what release it might make it in to? Cheers!

charles-d-burton:
+1 I also need this feature.

henryng24:
+1 need

zoharsanchez:
+1 for me as well! 🙏

sbeasey:
we need this feature

hylaride:
+1 (million) ;-)

arevell89:
+1

PR #11263, opened 2017-01-18
[WIP] provider/aws Add support for WAF Regional

neildothunter:
This is a straight copy of the current Terraform WAF resources (tests and docs) amended to use the WAF Regional API (the resulting WAF resources can then be associated with an Application Load Balancer in that region). As such it contains the same bugs as the current WAF resources.

WAF Regional endpoints (http://docs.aws.amazon.com/general/latest/gr/rande.html#waf_region ) are only available in us-east-1, us-west-2, eu-west-1 and ap-northeast-1. There is nothing here to provide early warning that the region may not support wafregional. 

Use aws_wafregional_* resources in addition to aws_waf_* follows the pattern set by the API, but doesn't feel "completely right".


neildothunter:
looks like my branch is out of step. Is this something i need to fix?

neildothunter:
aws_wafregional_web_acl_association facilitates link between WAF Regional WebACL and ALB. Fixes https://github.com/hashicorp/terraform/issues/10713.

Abrishges:
How can I associate my alb with WAF ACL? 

`resource "aws_wafregional_web_acl_association" "waf_alb_regional" {
  ............
}`

neildothunter:
That resource takes a webacl and alb id. The resource documentation is with the PR (https://github.com/neildothunter/terraform/blob/6e81561bde53e8e78baa767ff61b706bde138ab0/website/source/docs/providers/aws/r/wafregional_web_acl_association.html.markdown ). However this PR is getting a little old.

Abrishges:
@neildothunter : thank you! I think the provider is not supported by terraform yet. I getting the following error.

Errors:
  * aws_wafregional_web_acl_association.wafassociation: Provider doesn't support resource: aws_wafregional_web_acl_association

neildothunter:
This is a Pull Request (a request for this code to be merged into the main Terraform branch at which point it can be cut for a release). It will be closed once/if merged.

Until that point you could for example use the local-exec provisioner to call AWS CLI command http://docs.aws.amazon.com/cli/latest/reference/waf-regional/associate-web-acl.html or a Cloudformation stack or which ever method you use to create resources not supported by Terraform.

gjobin:
I'm not sure, but I think you must resolve the conflicts before it can be merged.
Nice addition by the way.

gjobin:
I think you must fix the conflicts before it can be merged back.
Nice addition by the way !

radeksimko:
@neildothunter Do you mind if we close this PR in favour of these?

 - https://github.com/hashicorp/terraform/pull/13705
 - https://github.com/hashicorp/terraform/pull/13707
 - https://github.com/hashicorp/terraform/pull/13708
 - https://github.com/hashicorp/terraform/pull/13709
 - https://github.com/hashicorp/terraform/pull/13710
 - https://github.com/hashicorp/terraform/pull/13711
 - https://github.com/hashicorp/terraform/pull/13712

Issue #11268, opened 2017-01-18
Support tilde '~' in shared credentials path for providers


joshlove:
### Terraform Version
v.0.8.2

### Affected Resource(s)
- aws provider

### Terraform Configuration Files
```hcl
provider "aws" {
  region                   = "us-west-2"
  shared_credentials_file  = "~/.aws/creds"
  profile                  = "customprofile"
}
```

### Expected Behavior
Evaluate to my $HOME, and then use the aws creds file from there. This would be nice in a team environment so we don't have usernames hardcoded.

Workaround is a wrapper script for this, but this would be nice as an enhancement.

grubernaut:
Hi @joshlove, thanks for the issue!

I've submitted #11277, which should allow you to use the `pathexpand()` interpolation function inside your provider config, and elsewhere.

```hcl
provider "aws" {
  region                   = "us-west-2"
  shared_credentials_file  = "${pathexpand(~/.aws/creds)}"
  profile                  = "customprofile"
}
```

joshlove:
Appreciate your work on this, thanks!

PR #11277, opened 2017-01-18
core: Add pathexpand interpolation function

grubernaut:
Adds the `pathexpand` interpolation function to allow users to expand `~` to the home directory in filepath strings.

Fixes: #11268


apparentlymart:
Should there be a warning in the docs that using this will likely make your config detect spurious diffs when run by multiple users or on multiple hosts? Fine to use this with `file(...)` but possibly less fine to use it directly in any resource argument.

(People already get confused by the similar problem caused by `path.module` and ask about it somewhat frequently.)

grubernaut:
@apparentlymart good point, yeah I can add some documentation to state that.

PR #11285, opened 2017-01-19
provider/openstack: Volume Attachment Updates

jtopjian:
This commit adds a StateRefresh func for volume attachments. Mostly
this is to add a buffer of time between the request and the return
of the attachment to give time for the volume to become attached,
however, in some cases the refresh function could work as specified.

Docs have also been updated to reflect that a device could be specified,
but to use with caution.


stack72:
LGTM! Thanks @jtopjian :)

PR #11288, opened 2017-01-19
Additionnal information for machine type

tpoindessous:
Additionnal information for machine type, which is the new name of "tiers".


stack72:
LGTM! Thanks @tpoindessous :)

Issue #11231, opened 2017-01-17
Idea: poll and verify success of *-exec commands


OJFord:
I'd like to be able to do something like:

```hcl
provisioner "local-exec" {
  command = "curl --fail -X POST /initialisation"
  verify = "curl --fail -X GET /running"
}
```

which is a bit silly, but just an example. Imagine for less-silly use the API is not under my own control (so remote-exec wouldn't be better) and provides a second 'operation success' endpoint like our `/running`.

Of course, this is currently possible by collapsing into the `command`, but it gets messy - especially if we want to poll the verify command until it works (and eventually timeout, failing the `apply`).

pkolyvas:
I'm doing a clean-up of older issues that haven't gained traction but also don't coincide with some of our current focus. While provisioners, as a whole, aren't a top priority for the team, when some time is available to focus on them we're going to use that attention for some long-standing defects. 

We appreciate every idea and contribution to Terraform, even if we can't get to them all. My door is always open to talk about Terraform, the use cases, and needs of our practitioners.

OJFord:
I can't really remember what I was trying to achieve - I think it was to do with checking kubernetes apiserver was correctly started and alive, but can't be sure.

Fair enough on closing for limited interest - note though that someone else (not I) did go to the effort to open a PR, twice, rebased when requested, and only then gave up on it after receiving no further feedback.

It is currently a forgotten use case to me though :slightly_smiling_face:.

PR #11293, opened 2017-01-19
provisioner/exec: Add verify attribute

grubernaut:
Adds the `verify` attribute for `local-exec` and `remote-exec`. Allows a user to specify a specific verification script, in order to verify that the provisioning script actually performed the actions that it needed to.

Fixes: #11231


cblecker:
@grubernaut Would you mind rebasing? :)

PR #11300, opened 2017-01-20
Ns1 provider

pashap:
Ref https://github.com/hashicorp/terraform/pull/10391

/cc @cakoose @josephholsten


pashap:
Just for reference, some thoughts/description around the ns1 metadata field:

 - https://github.com/hashicorp/terraform/pull/10058#discussion_r87617034
 - https://github.com/hashicorp/terraform/pull/10058#discussion_r88322244
 - https://github.com/hashicorp/terraform/pull/11300/commits/ab76b55bbca53f5fd076fbcaf8cc78fa8a06bb90

The metadata field is commented out in this PR(both in code and acctests), with the idea that we can get the provider in the next release and iterate on how to implement the metadata field after.

stack72:
Going to merge this back to the other PR - that way we can run the acctests and see what happens :)

builtin/providers/ns1/resource_apikey.go
```diff
@@ -3,7 +3,7 @@ package ns1
 import (
 	"github.com/hashicorp/terraform/helper/schema"
 
-	nsone "gopkg.in/ns1/ns1-go.v2/rest"
+	ns1 "gopkg.in/ns1/ns1-go.v2/rest"
```
josephholsten: I don't love importing into `ns1` while in a packaged named `ns1`. But there are worse things in like, and at least this reads cleanly when you grok it. Hopefully new readers of this code will realize what's going on.

pashap: Ah, thats a really good point, didnt consider it at all. im down to change

cakoose:
@pashap: What are the remaining issues blocking metadata?  The ones I can remember:
1. Using a nested single-item list seems clunky.  Maybe there's a cleaner way?
2. The issue where even optional omitted fields are written to the state, causing conflicts (since "up" and "up_feed" are not both allowed to be present).

Are there others?  Is there anything I can do to help?

pashap:
@cakoose: The second issue is still whats holding us back. We cant get metadata working without the `ConflictsWith` prohibiting using "up" and "up_feed". Sorry for the delayed response, Ive been focused on trying to get the provider green without metadata, so we can then solely focus on that.

cakoose:
Got it.  I definitely don't want to interfere with the initial merge.  I was just wondering if there was was good place to discuss the remaining issue (I'm thinking of ideas but don't want to do too much work before hearing your perspective as well).

PR #11301, opened 2017-01-20
Basic authentication for HTTP remote state backend

syskill:
A simple fix to make the HTTP remote state backend pass muster for production use.

### DIFF
--- a/state/remote/http.go
+++ b/state/remote/http.go
@@ -44,20 +44,40 @@ func httpFactory(conf map[string]string) (Client, error) {
 		}
 	}
 
-	return &HTTPClient{
+	ret := &HTTPClient{
 		URL:    url,
 		Client: client,
-	}, nil
+	}
+	if username, ok := conf["username"]; ok && username != "" {
+		ret.Username = username
+	}
+	if password, ok := conf["password"]; ok && password != "" {
+		ret.Password = password
+	}
+	return ret, nil
 }
 
 // HTTPClient is a remote client that stores data in Consul or HTTP REST.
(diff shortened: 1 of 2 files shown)


state/remote/http.go
```diff
@@ -44,20 +44,40 @@ func httpFactory(conf map[string]string) (Client, error) {
...
 
 func (c *HTTPClient) Get() (*Payload, error) {
-	resp, err := c.Client.Get(c.URL.String())
+	req, err := http.NewRequest("GET", c.URL.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Prepare the request
+	if c.Username != "" {
```
grubernaut: What if `username` is set and `password` is not set? Possible to add in extra verification for this case?

state/remote/http.go
```diff
@@ -141,6 +161,9 @@ func (c *HTTPClient) Put(data []byte) error {
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Content-MD5", b64)
 	req.ContentLength = int64(len(data))
+	if c.Username != "" {
```
grubernaut: Same here

state/remote/http.go
syskill: Unnecesary. The empty string is the zero value, so in this case the HTTP client will just use an empty password, which is totally legit for HTTP basic auth.

state/remote/http.go
syskill: Same here

PR #11303, opened 2017-01-20
provider/aws: Add DataSource to get a list of Autoscaling groups in a region

stack72:
```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAutoscalingGroups_'
2017/01/20 11:39:37 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAutoscalingGroups_basic -timeout 120m
=== RUN   TestAccAWSAutoscalingGroups_basic
--- PASS: TestAccAWSAutoscalingGroups_basic (121.29s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	121.335s
```

### DIFF
--- a/builtin/providers/aws/provider.go
+++ b/builtin/providers/aws/provider.go
@@ -147,6 +147,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_alb":                      dataSourceAwsAlb(),
 			"aws_alb_listener":             dataSourceAwsAlbListener(),
 			"aws_ami":                      dataSourceAwsAmi(),
+			"aws_autoscaling_groups":       dataSourceAwsAutoscalingGroups(),
 			"aws_availability_zone":        dataSourceAwsAvailabilityZone(),
 			"aws_availability_zones":       dataSourceAwsAvailabilityZones(),
 			"aws_billing_service_account":  dataSourceAwsBillingServiceAccount(),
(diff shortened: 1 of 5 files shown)


website/source/docs/providers/aws/d/autoscaling_groups.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+    "autoscaling:EC2_INSTANCE_LAUNCH_ERROR",
+    "autoscaling:EC2_INSTANCE_TERMINATE_ERROR"
+  ]
+  topic_arn = "TOPIC ARN"
+}
+```
+
+## Argument Reference
+
+The datasource currently takes no arguments as it uses the current region that the provider works in.
```
jen20: s/datasource/data source/

builtin/providers/aws/provider.go
```diff
@@ -147,6 +147,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_alb":                      dataSourceAwsAlb(),
 			"aws_alb_listener":             dataSourceAwsAlbListener(),
 			"aws_ami":                      dataSourceAwsAmi(),
+			"aws_autoscaling_groups":       dataSourceAwsAutoscalingGroups(),
```
jen20: Should we name this `aws_autoscaling_group_names` since it doesn't return all the properties?

stack72:
Discussed the potential name change with @jen20 - this follows the same lines as aws_availaiblity_zones

we expose a names parameter. IF we want to expose any other info then we can do it on a single resource, again like we do with aws_availability_zone :)

P.

PR #11308, opened 2017-01-20
Fix vendored packages

jbardin:
The code in the vendor folder was out of sync with the vendor manifest.

This first re-syncs the vendor folder with the spec'ed packages. Any packages that were modified because of this were in an unknown state. Rather than try to match the file state in each of those packages to a known commit, I simply updated from their respective master branches, and verified that all tests pass.

The only exception here is nomad, which was reverted to the state in master, removed from vendor.json and added to the ignore list. We can work on finding an acceptable way to vendor nomad separately. 

Finally, this adds a vendor-status make target which is also checked during the CI build process so that a broken vendor directory can no longer be merged into master.


catsby:
Failing due to dependency things 🤔 
```
2017/01/20 14:47:26 Generated command/internal_plugin_list.go
TF_ACC= go test $(go list ./... | grep -v '/terraform/vendor/' | grep -v '/builtin/bins/')  -timeout=30s -parallel=4
vendor/golang.org/x/oauth2/google/default.go:16:2: cannot find package "cloud.google.com/go/compute/metadata" in any of:
	/home/travis/gopath/src/github.com/hashicorp/terraform/vendor/cloud.google.com/go/compute/metadata (vendor tree)
```

Is that expected?

jbardin:
nope, this govendor task failed miserably ;(

jbardin:
Not as bad as I thought. I forgot to try building under a clean GOPATH, and some packages we're properly vendored at all.

jbardin:
Now those packages are in the vendor folder, but it still says they're not found 🤔

jbardin:
All fixed, forgot to `git add` 🤦‍♂️

PR #11309, opened 2017-01-20
documentation: Add warning for pathexpand interpolation function

grubernaut:
Note that the `pathexpand` interpolation function could cause an unwanted diff on two different hosts.

cc: @apparentlymart


apparentlymart:
:+1: looks good to me!

Does the second line of that list item need to be indented to align with the first, as is done for the other bullets? (I remember this behaving differently between Markdown and reStructuredText but I can't remember which one behaves in which way...)

PR #11319, opened 2017-01-20
 Clarify location of the remote state file in Manta backend

travispaul:
The user's private object storage directory (`/$MANTA_USER/stor/`) will automatically be prepended to the supplied path. It's helpful to know that you can't use an absolute Manta path such as `/$MANTA_USER/stor/random/path` or `~~/stor/random/path`.


jen20:
Thanks @travispaul!

Issue #11325, opened 2017-01-21
ignition_systemd_unit should support enable-only definitions


rtlong:
### Terraform Version
`0.8.4` and `master`

### Affected Resource(s)
- `ignition_systemd_unit`

### Terraform Configuration Files
```hcl
resource "ignition_systemd_unit" "iptables_restore" {
  name = "iptables-restore.service"
  enable = true
}
```
The intent here is to ask Ignition merely to enable an already defined unit, without redefining the contents of the unit file or adding any drop-ins.

### Expected Behavior
This resource should result in this snippet of Ignition JSON config:
```json
{
  "systemd": {
    "units": [
      {
        "name": "iptables-restore.service",
        "enable": true
      }
    ]
  }
}
```

Which, as far as I can tell, is valid, according to https://coreos.com/validate/

### Actual Behavior
```
Error applying plan:

1 error(s) occurred:

* ignition_systemd_unit.iptables_restore: invalid or empty unit content
```

### Steps to Reproduce
1. define an `ignition_systemd_unit` with `enable = true` and neither `content` nor any `dropin` defined
1. `terraform apply`

PR #11327, opened 2017-01-21
provider/ignition: Fix systemd unit errors

grubernaut:
According to the coreos [documentation](https://coreos.com/ignition/docs/latest/configuration.html), systemd units only require the name attribute per each unit. This can also be validated with the CoreOS config validator. This change allows the `ignition_systemd_unit` resource to no longer fail if given an empty `content` and `dropin`.

Also adds a test to cover this use case.

Fixes: #11325


stack72:
LGTM :)

rtlong:
Thanks for this! :+1:

gappan:
The docs are not updated with this information. Just wanted to point this out. 
https://www.terraform.io/docs/providers/ignition/d/systemd_unit.html

grubernaut:
Hey @gappan, thanks for noticing! 

https://github.com/terraform-providers/terraform-provider-ignition/pull/6, should include the fix needed.

PR #11329, opened 2017-01-21
Destroy Provisioners

### LINKED ISSUE #386: Destroy 'provisioner' for instance resources
I would be great to have sort of a 'provisioner' for destroying an instance resource.

Example: 
When creating a instance, I bootstrap it with chef and the node is registered with the chef server. Now I need a way of automatically deleting the node from the chef server after terraform destroys the instance.

mitchellh:
Fixes #386 

I'm going to go light on the details in this PR description since there is a really good discussion and RFC in #386. You can also view the website changes in this PR to see how it works!

Highlights:

  * Provisioners that run during destruction
  * Option `on_failure` to determine behavior when a provisioner fails (works for both create and destroy provisioners)

### DIFF
--- a/website/source/docs/provisioners/index.html.markdown
+++ b/website/source/docs/provisioners/index.html.markdown
@@ -3,15 +3,104 @@ layout: "docs"
 page_title: "Provisioners"
 sidebar_current: "docs-provisioners"
 description: |-
-  When a resource is initially created, provisioners can be executed to initialize that resource. This can be used to add resources to an inventory management system, run a configuration management tool, bootstrap the resource into a cluster, etc.
+  Provisioners are used to execute scripts on a local or remote machine as part of resource creation or destruction.
 ---
 
 # Provisioners
 
-When a resource is initially created, provisioners can be executed to
-initialize that resource. This can be used to add resources to an inventory
-management system, run a configuration management tool, bootstrap the
-resource into a cluster, etc.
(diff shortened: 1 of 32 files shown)


cemo:
wuhuu :) Don't forget to add tags.

website/source/docs/provisioners/index.html.markdown
```diff
@@ -3,15 +3,104 @@ layout: "docs"
...
+how to communicate with the resource.
+
+## Creation-Time Provisioners
+
+Provisioners by default run when the resource they are defined within is
+created. Creation-time provisioners are only run during _creation_, not
+during updating or any other lifecycle. They are meant as a means to perform
+bootstrapping of a system.
+
+If a creation-time provisioner fails, the resource is marked as **tainted**.
```
jbardin: I think this is a good idea, but we should probably document it as a backwards compatibility change since buggy scripts that didn't exit cleanly would have been OK previously.

mitchellh: How so? This behavior is the same as always.

jbardin: Oh, maybe I'm mis-remembering. I was thinking a failed provisioner _didn't_ mark the resource as tainted, and left it as-is in the state. No problem then, if nothing changed.

PR #11341, opened 2017-01-22
core: New Refresh Graph

mitchellh:
This PR modifies the refresh graph to use a new custom graph builder. This work is just continuing the work I started and shipped for 0.8 but on the less complex and less dangerous graphs. See #9973 or #9388 as prior examples. 

The goal at this point now that plan/apply have shipped on new graphs is to complete the less dangerous graphs so we can ultimately fully remove the legacy graph building code. We still have to do refresh, validate, and input.

The changes here are minimal since refresh is simple. **What this PR changes/adds:**

  * New refresh graph, based on state as primary source of truth
  * `StateFilter` made more robust to handle cases where state may not have resource type. This was necessary for tests to pass without changes.
  * `ConfigTransformer` can filter the types of things added. This was necessary to add only data sources for refresh.

**This PR does not enable/use the shadow graph.** Refresh is not a scary operation (compared to plan/apply) and therefore we won't be using the shadow graph for this. We'll be shipping in default-on in 0.9 betas and let any bugs come through there.

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -233,6 +233,15 @@ func (c *Context) Graph(typ GraphType, opts *ContextGraphOpts) (*Graph, error) {
 			Validate: opts.Validate,
 		}).Build(RootModulePath)
 
+	case GraphTypeRefresh:
+		return (&RefreshGraphBuilder{
+			Module:    c.module,
+			State:     c.state,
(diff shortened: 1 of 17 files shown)


mitchellh:
Closing for #11426 which builds on top of this.

PR #11347, opened 2017-01-23
Ns1 provider

pashap:
- Includes merge from master that fixed failing acctest.
- Adds ns1-go rest client via govendor
- Adds docs around some more resources
- `go vet` is green on provider


stack72:
LGTM! Thanks :)

PR #11348, opened 2017-01-23
fix #10788 skip validation of chef attributes_json compute values

### LINKED ISSUE #10788: Chef provisioner attributes_json cannot parse rendered template_file
### Terraform Version
0.8.1

### Affected Resource(s)
- `null_resouce`
- `template_file` (data)

```hcl
data "template_file" "json" {
  template = <<EOF
    {
      "abc": "def"
    }
EOF
}

resource "null_resource" "chef" {
  triggers { json = "${data.template_file.json.rendered}" }

  provisioner "chef" {
    node_name = "chef"
    server_url = "http://chef.chef/chef"
    user_name = "chef"
    user_key = "chef"
    run_list = []
    attributes_json = "${data.template_file.json.rendered}"
  }
}
```

### Debug Output
https://gist.github.com/b-dean/5e1b28f9f03a48823c2aa8bb515d5f36

### Expected Behavior
plan has the following output (this output is from running `terraform plan` with 0.7.13):

```
+ null_resource.chef
    triggers.%:    "1"
    triggers.json: "    {\n      \"abc\": \"def\"\n    }\n"
```

### Actual Behavior
plan fails:

```
fix these before continuing.

Errors:

  * null_resource.chef: Error parsing attributes_json: invalid character 'D' after top-level value
```

### Steps to Reproduce
1. `terraform plan`

sjwl:
The problem starts when `attributes_json` has an interpolation of a computed value.  And so during `terraform plan`, the value turns into `74D93920-ED26-11E3-AC10-0800200C9A66` (the value of `const UnknownVariableValue`) which subsequently fails because the validation attempts to decode that value as json format, which it obviously isn't.

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/hashicorp/terraform/communicator"
 	"github.com/hashicorp/terraform/communicator/remote"
+	"github.com/hashicorp/terraform/config"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/mitchellh/go-homedir"
 	"github.com/mitchellh/go-linereader"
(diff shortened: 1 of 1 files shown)


brainrape:
Is there anything neede to merge this? The bug is preventing proper use of the chef provisioner, the last usable version being v0.7.13

mitchellh:
Thanks for this! I'm working on this now. While this works, it isn't strictly correct since long-term we don't actually want anybody using UnknownVariableValue directly. The approach I'd prefer is to use `IsComputed()` API to check this. I'm going to do this now and merge. 

Your PR is basically correct except swapping one line with the IsComputed check instead. 

Thanks! Sorry for not jumping on this earlier. :(

Issue #7183, opened 2016-06-15
Don't validate AWS Regions in provider config bootstrap


dastbe:
Hey,

Wanted to open this issue before I supplied a code change. I noticed that terraform is using a static array of available regions to whitelist terraforming. Of course, new regions are coming all the time and some people are able to make use of them before they're publicly available. For unlaunched services one could create a shim provider until its early access/full release, but here the only solution is to branch and build :(.

Is there a compelling reason to have any validation here, or at most warn that this is not a known public region?

Here's the [method](https://github.com/hashicorp/terraform/blob/cd160d3f3f6c80bbf12fae11b0d9b6d6c56bbeff/builtin/providers/aws/config.go#L326) in question.

Thanks!

cjeanneret:
It would be better to ask AWS directly what regions are available I guess? For example, the following awscli command returns the required things:
```Bash
aws ec2 describe-regions
```
Pretty sure the aws-sdk for go supports such a thing — might even be cached for some time if needed.

stack72:
Hi 

To quote @radeksimko from another issue (https://github.com/hashicorp/terraform/pull/7383#issuecomment-229169574)

```
I think this is purely for static/offline validation of the aws provider configuration - so that the user gets a human-friendly error for cases like this:

provider "aws" {
  region = "not-exists"
}
We cannot verify this online, even though it's available from the API, because we'd need to choose a valid region (and hard-coding a single region is not a good idea since that region becomes a single point of failure) + have valid credentials so this is kind of a 🐔 & 🍳 problem when validating provider which is responsible for setting up the connection.

Also the way most SDKs work is that they simply take the given region as a string and make a hostname of it, so the error message user would see if Terraform wouldn't perform this offline validation is something like this:

Could not connect to the endpoint URL: "https://sts.not-exists.amazonaws.com/"
What I think would make sense to do though is to keep such list of valid regions in the SDK and let the SDK perform the validation, but SDK doesn't currently offer such functionality neither a list of regions.
```

Hope this helps

Paul

radeksimko:
> For unlaunched services one could create a shim provider until its early access/full release, but here the only solution is to branch and build

All AWS tooling would typically also need to be rebuilt using the JSON blueprints to get a new API functionality - i.e. it is non-trivial to use off-the-shelf tools to control private, unreleased APIs anyway and I don't expect this to be easy in Terraform either.

That said this may not be the case for regions that offer the exact same APIs.

[...]

PR #11358, opened 2017-01-23
provider/aws: Allow bypassing region validation

radeksimko:
This was requested in a few places, e.g. https://github.com/hashicorp/terraform/issues/7183 and could take the pressure off the users (rushing to upgrade) & maintainers (rushing to cut release) when AWS announces new region.

btw. [Packer already has a similar option](https://www.packer.io/docs/builders/amazon-ebs.html#skip_region_validation).

It is important to say that there will be still certain data sources & resources which simply won't work as Terraform doesn't have the metadata for unknown regions, e.g.

 - `aws_s3_bucket` due to [zone ID](https://github.com/hashicorp/terraform/blob/2efe489d59ed2391aa128541eae5ddcb73976bd3/builtin/providers/aws/resource_aws_s3_bucket.go#L879-L883)
 - `aws_redshift_service_account`
 - `aws_elb_service_account`
 - `aws_elb_hosted_zone_id`

### Before

```hcl
provider "aws" {
  region = "eu-west-x"
}

resource "aws_vpc" "tada" {
  cidr_block = "10.0.0.0/16"
}
```
```

* Not a valid region: eu-west-x
```

### After

```hcl
provider "aws" {
  region = "eu-west-x"
  skip_region_validation = true
}

resource "aws_vpc" "tada" {
  cidr_block = "10.0.0.0/16"
}
```
Assuming the region doesn't _actually_ exist, the SDK will just keep retrying for ~5mins and eventually return error due to failed DNS lookup:
```
Error applying plan:

1 error(s) occurred:

* aws_vpc.tada: Error creating VPC: RequestError: send request failed
caused by: Post https://ec2.eu-west-x.amazonaws.com/: dial tcp: lookup ec2.eu-west-x.amazonaws.com on 10.0.0.1:53: no such host
```
^ that is also why most users using valid and publicly available regions want to avoid setting this new parameter.

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -92,6 +92,7 @@ type Config struct {
 	Insecure         bool
 
 	SkipCredsValidation     bool
+	SkipRegionValidation    bool
 	SkipRequestingAccountId bool
 	SkipMetadataApiCheck    bool
 	S3ForcePathStyle        bool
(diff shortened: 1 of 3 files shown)


stack72:
LOVE THIS! LGTM :)

Issue #10061, opened 2016-11-11
Terraform wrongly assumes arn to be always be arn:aws (govcloud uses arn:aws-us-gov)


jc-m:
The following code:
https://github.com/hashicorp/terraform/blob/351c6bed79abbb40e461d3f7d49fe4cf20bced41/builtin/providers/aws/resource_aws_s3_bucket.go#L787

d.Set("arn", fmt.Sprint("arn:aws:s3:::", d.Id()))

wrongly assumes that the arn is always aws. In govcloud region, it is aws-us-gov

### Terraform Version
Terraform v0.7.8

### Affected Resource(s)
-aws_s3_bucket
-maybe others

### Terraform Configuration Files
```hcl
resource "aws_s3_bucket" "test" {
  provider = "aws"
  bucket = "bucket-test"
  region = "us-gov-west-1"
  acl = "private"
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
state show aws_s3_bucket.test with the correct arn:
arn                 = arn:aws-us-gov:s3:::bucket-test

### Actual Behavior
arn                 = arn:aws:s3:::bucket-test

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

jc-m:
Just as an update - this also breaks the capability to use kms_key_id in aws_ebs_volume.

I wonder why this issue is ignored - see http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#genref-arns

the assumptions are wrong - this breaks in govcloud and china region.

grubernaut:
Hi @jc-m, thanks for the issue! 

I'll make sure this is prioritized early next week and we'll internally begin to put together a solution for you. Sorry for the delay!

jc-m:
Hi @grubernaut - sorry for the push - i patched terraform to use a map for the namespaces - but i think it might be useful to have a data source which provides this information in addition to internally changing  few regexp and sprintf. There are few other peculiarities in each zones that might be helpful to give visibility into (endpoints is one of them)

radeksimko:
Related:
 - https://github.com/hashicorp/terraform/pull/9273
[...]

PR #9273, opened 2016-10-07
Parse AWS partition from ARN.

jmcarp:
Helper functions used to construct arn strings in several aws resources assume the "aws" partition: "arn:aws:%s...". This results in incorrect arn strings when using a different aws partition, like aws-us-gov or aws-cn. This patch parses the current partition while getting the current account id and passes the partition to the various arn helper functions.

[Resolves #5307]

PR #11359, opened 2017-01-23
provider/aws: Allow ARN identifier to be set

grubernaut:
Allows users from govcloud and other regions (aws-cn) to now use the following resources correctly:

```
- data "aws_billing_service_account"
- data "aws_elb_service_account"
- resource "aws_cloudfront_origin_access_identity"
- resource "aws_ecs_service"
- resource "aws_iam_saml_provider"
- resource "aws_lambda_permission"
- resource "aws_sns_topic_policy"
```

Fixes: #7309 

Related:
 - #9273 
 - #10061

### DIFF
--- a/builtin/providers/aws/resource_aws_ecs_service.go
+++ b/builtin/providers/aws/resource_aws_ecs_service.go
@@ -282,7 +282,7 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("name", service.ServiceName)
 
 	// Save task definition in the same format
-	if regexp.MustCompile(`^arn:[\w-]+:ecs:`).MatchString(d.Get("task_definition").(string)) {
+	if strings.HasPrefix(d.Get("task_definition").(string), "arn:"+meta.(*AWSClient).partition+":ecs:") {
 		d.Set("task_definition", service.TaskDefinition)
 	} else {
 		taskDefinition := buildFamilyAndRevisionFromARN(*service.TaskDefinition)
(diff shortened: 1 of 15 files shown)


builtin/providers/aws/resource_aws_ecs_service.go
```diff
@@ -292,7 +292,7 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("desired_count", service.DesiredCount)
 
 	// Save cluster in the same format
-	if strings.HasPrefix(d.Get("cluster").(string), "arn:aws:ecs:") {
+	if regexp.MustCompile(`^arn:[\w-]+:ecs:`).MatchString(d.Get("cluster").(string)) {
```
radeksimko: I know I originally suggested using regexp matching here, but now that I'm looking at the whole diff I'm thinking we could just use the same pattern as elsewhere and avoid the loose matching - i.e.

```go
if strings.HasPrefix(d.Get("cluster").(string), "arn:"+meta.(*AWSClient).partition+":ecs:") {
```
or something similar w/ `fmt.Sprintf` - depending on what looks better.

What do you think?

builtin/providers/aws/resource_aws_lambda_permission_test.go
```diff
@@ -63,6 +63,18 @@ func TestLambdaPermissionGetQualifierFromLambdaAliasOrVersionArn_alias(t *testin
...
+	arnWithAlias := "arn:aws-us-gov:lambda:us-west-2:187636751137:function:lambda_function_name:testalias"
+	expectedQualifier := "testalias"
+	qualifier, err := getQualifierFromLambdaAliasOrVersionArn(arnWithAlias)
+	if err != nil {
+		t.Fatalf("Expected no error when getting qualifier: %s", err)
+	}
+	if qualifier != expectedQualifier {
+		t.Fatalf("Expected qualifier to match (%q != %q)", qualifier, expectedQualifier)
+	}
+}
```
radeksimko: 👍 for the test

builtin/providers/aws/resource_aws_ecs_service.go
grubernaut: That looks/sounds good to me, I'll make that change and push

PR #11368, opened 2017-01-23
provider/ns1/record: Fix "use_client_subnet".

cakoose:
The support for "use_client_subnet" was half finished.
- Field was defined in schema.
- ResourceData-to-struct code was present but incorrect.
- struct-to-ResourceData code was missing.

Made the change and verified with manual testing:
1. In NS1 UI, switched "Use Client Subnet" between checked and
   unchecked.
2. In Terraform config file, switched "use_client_subnet" field between
   "true", "false", and omitted.
3. The output of "terraform plan" was as expected in all six cases.

@pashap, @stack72 

BTW, if there are minor tweaks that need to be made before committing, it might be quicker to just make those changes yourself and then commit.  I don't care about retaining authorship here.


pashap:
LGTM, thanks for fixing/testing!

Issue #11355, opened 2017-01-23
Better support for partial version in aws_db_instances


alext:
The AWS API allows specifying a partial version (ie "9.5" instead of "9.5.4") when creating an RDS instance. This when combined with `auto_minor_version_upgrade=true` can be used to specify that you want an instance using the latest point release of a given major version.

At present, this mostly works with terraform, but results in terraform trying to change the version on every run, which takes 30-40 seconds, even though the actual change is a no-op.

### Terraform Version

```
$ terraform -v
Terraform v0.8.4
```

### Affected Resource(s)

- aws_db_instance

### Terraform Configuration Files
```hcl
resource "aws_db_instance" "test" {
  identifier              = "tf-test"
  allocated_storage       = 5
  storage_type            = "gp2"
  instance_class          = "db.t2.micro"
  backup_retention_period = 0
  username                = "dbadmin"
  password                = "super-secret"

  engine                     = "postgres"
  engine_version             = "9.5"
  auto_minor_version_upgrade = true
}
```

### Debug Output

```
$ TF_LOG=DEBUG TF_LOG_PATH=rds_partial_version_debug.log terraform apply
aws_db_instance.test: Refreshing state... (ID: tf-test)
aws_db_instance.test: Modifying...
  engine_version: "9.5.4" => "9.5"
aws_db_instance.test: Still modifying... (10s elapsed)
aws_db_instance.test: Still modifying... (20s elapsed)
aws_db_instance.test: Still modifying... (30s elapsed)
aws_db_instance.test: Modifications complete

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.

The state of your infrastructure has been saved to the path
below. This state is required to modify and destroy your
infrastructure, so keep it safe. To inspect the complete state
use the `terraform show` command.

State path: terraform.tfstate
```

Debug log: https://gist.github.com/alext/89824c542bf4d598e03af4cb90dd0d2e

### Expected Behavior

Terraform should not have attempted to update the resource.

### Actual Behavior

Terraform attempted to change the version from "9.5.4" to "9.5"

### Steps to Reproduce

1. `terraform apply` to create the initial resource
1. `terraform apply` a second time to see it being updated.

stack72:
Hi @alext 

Thanks for getting in touch here. From looking at your terraform config, you are missing `apply_immediately = true` 

[...]

PR #11369, opened 2017-01-23
provider/aws: Add diff suppress function for aws_db_instance

grubernaut:
Adds a diff suppress function for the `engine_version` attribute of the `db_instance` AWS resource.

The function only supresses the state diff, if the attribute key `auto_minor_version_upgrade` is set, and if the returned `engine_version` from the running RDS instance shares the same prefix as the configured `engine_version`.

 ```
 $ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSDBInstance_MinorVersion'
 2017/01/23 17:59:14 Generated command/internal_plugin_list.go
 TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSDBInstance_MinorVersion -timeout 120m
 === RUN   TestAccAWSDBInstance_MinorVersion
 --- PASS: TestAccAWSDBInstance_MinorVersion (503.48s)
 PASS
 ok      github.com/hashicorp/terraform/builtin/providers/aws    503.518s
 ```

Fixes: #11355

### DIFF
--- a/builtin/providers/aws/diff_suppress_funcs.go
+++ b/builtin/providers/aws/diff_suppress_funcs.go
@@ -1,6 +1,9 @@
 package aws
 
 import (
+	"log"
+	"strings"
+
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/jen20/awspolicyequivalence"
 )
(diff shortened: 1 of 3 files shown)


dcarley:
Awesome, thanks 👍

PR #11373, opened 2017-01-24
core: input graph uses new graph builders

mitchellh:
This is similar to #11341 but for inputs.

There is no tangible benefit to this except to continue marching towards our goal of removing the legacy graph completely. Almost there! No tests added, removed, or changed. All existing tests pass.

This is the simplest so far in that it actually shares the graph with plan. The goal of the new graph builders was to create unique builders for each operation to simplify the graph. For input in particular, the graph is identical to the plan graph since in order to determine what we need input for we need to know what we're going to plan.

Therefore, input just shares the plan graph. Creating an entire new graph builder would be silly since it'd always have to be identical. While Refresh, Plan, Apply have unique ways to build their graphs, I believe Input and Validate likely won't since it is very similar to planning... but we'll see about Validate. 

The way the graph is built (using `GraphTypeInput`) gives us some play in the future if we choose to create a custom graph builder.

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -219,15 +219,32 @@ func (c *Context) Graph(typ GraphType, opts *ContextGraphOpts) (*Graph, error) {
 	case GraphTypeInput:
 		// The input graph is just a slightly modified plan graph
 		fallthrough
+	case GraphTypeValidate:
+		// The validate graph is just a slightly modified plan graph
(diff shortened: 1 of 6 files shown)


mitchellh:
Closing, going to reopen with validate, refresh, input all in one since they're related.

Issue #10447, opened 2016-11-30
Amazon Lightsail


jgartrel:
This is a feature request to support Amazon Lightsail, per their public announcement there are separate APIs.

> As is almost always the case with AWS, there’s a full set of APIs behind all of the console functionality that we just reviewed. Here are just a few of the more interesting functions:
> 
> - GetBundles – Get a list of the bundles (machine configurations).
> - CreateInstances – Create one or more Lightsail instances.
> - GetInstances – Get a list of all Lightsail instances.
> - GetInstance – Get information about a specific instance.
> - CreateInstanceSnapshot – Create a snapshot of an instance.
> - CreateInstanceFromSnapshot – Create an instance from a snapshot.

https://docs.aws.amazon.com/lightsail/2016-11-28/api-reference/Welcome.html?fid=71746835D2E3C6B0-212737D1B3CC5C3F

Thank you,

James

nodesocket:
👍  💯

jessefulton:
Looks like this was partially implemented by #10473 - it would be amazing to add support for static ips as well!

nodesocket:
👍  for static ips support.

jgartrel:
@catsby It would be great to allow CreateInstanceFromSnapshot in the existing implementation of lightsail instance.   Are you interested in working on that?

catsby:
@jgartrel I'm strapped ATM but otherwise would love to. If you're up for taking a stab at it, please ping me on the PR 😄

radeksimko:
Hi folks,
I think we now have the initial support of Lightsail in place, including static IPs which I've done in #13175 & #13207

There's clearly a potential for 1 or 2 more resources (`aws_lightsail_domain_entry` & `aws_lightsail_public_port`), but as a "house-keeper" 🚿 I do not want to see this issue expanding into "anything Lightsail we might ever support" and going stale, hence I'm closing it. With that said we certainly welcome PRs with for the mentioned resources and will gladly review them. It is possible that the HC team may implement these resources at some point, but unlikely in the near-term.

Thanks!

k99bw01:
Load balancer support and Snapshot support would be great.

PR #11376, opened 2017-01-24
instantiate lightsail with an instance_snapshot_name

ahammond:
Per https://github.com/hashicorp/terraform/issues/10447 it would be nice to be able to instantiate a lightsail instance based on a snapshot.

### DIFF
--- a/builtin/providers/aws/resource_aws_lightsail_instance.go
+++ b/builtin/providers/aws/resource_aws_lightsail_instance.go
@@ -35,7 +35,12 @@ func resourceAwsLightsailInstance() *schema.Resource {
 			},
 			"blueprint_id": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				ForceNew: true,
+			},
+			"instance_snapshot_name": {
+				Type:     schema.TypeString,
+				Optional: true,
 				ForceNew: true,
 			},
 			"bundle_id": {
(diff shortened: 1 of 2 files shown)


ahammond:
I couldn't find the relevant UTs to update. I've managed a single run to confirm using the following:

```
resource "aws_lightsail_instance" "from_snapshot_test" {
  name              = "snapshot_instance"
  availability_zone = "us-east-1b"
  instance_snapshot_name  = "snapshot_base-system-1485251845526"
  bundle_id         = "nano_1_0"
}
```

builtin/providers/aws/resource_aws_lightsail_instance.go
```diff
@@ -35,7 +35,12 @@ func resourceAwsLightsailInstance() *schema.Resource {
 			},
 			"blueprint_id": {
 				Type:     schema.TypeString,
-				Required: true,
```
ahammond: This shouldn't really be Optional, so much as requiring exactly one of blueprint_id or instance_snapshot_name. I don't know how to express that. Either that or I should replace this with "image_name" and then detect if it's one of the blueprints or if it's a snapshot. Which might be the more elegant solution.

white-hat:
I'd love to have this soon in TF

onorua:
Any plans to include tests?

valch85:
I would like to have it in terraform.

PR #11379, opened 2017-01-24
Add support for importing openstack compute instance

yagnik:
Hey @jtopjian wanted to get your opinion on this before I send it upstream. 
Would you mind giving me some context as to why we didn't implement import for compute instances before ?
Thanks!

### DIFF
--- a/builtin/providers/openstack/import_openstack_compute_instance_v2_test.go
+++ b/builtin/providers/openstack/import_openstack_compute_instance_v2_test.go
@@ -1,0 +1,28 @@
+package openstack
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccComputeV2Instnace_importBasic(t *testing.T) {
+	resourceName := "openstack_compute_instance_v2.sg_1"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeV2InstanceDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeV2Instance_basic,
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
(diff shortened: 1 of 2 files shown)


yagnik:
Whoops, I made the pr to upstream instead of my own repo :/. 
Oh well, now that I'm here already.

yagnik:
cc @vicky999

jtopjian:
@yagnik Thanks! I've tried getting this to work in the past, but have run into some issues around networks and volumes. A lot of the improvements I've been making to the OpenStack provider have been to get instance importing working, so it's very possible that the provider is now in good shape!

I'll take a look either today or tomorrow and let you know. 😄

builtin/providers/openstack/import_openstack_compute_instance_v2_test.go
```diff
@@ -0,0 +1,28 @@
+package openstack
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccComputeV2Instnace_importBasic(t *testing.T) {
+	resourceName := "openstack_compute_instance_v2.sg_1"
```
jtopjian: This should be `openstack_compute_instance_v2.instance_1`

jtopjian:
@yagnik I did a quick test of this and with the fix mentioned above, the test is still failing:

```
--- FAIL: TestAccComputeV2Instnace_importBasic (79.65s)
        testing.go:265: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

                (map[string]string) (len=1) {
                 (string) (len=14) "network.0.uuid": (string) ""
                }


                (map[string]string) (len=2) {
                 (string) (len=14) "network.0.uuid": (string) (len=36) "93e3ae34-0b8b-4ec8-b06a-f9d7d1b9f033",
                 (string) (len=19) "stop_before_destroy": (string) (len=5) "false"
                }
```

The `stop_before_destroy` attribute was added after I tested this a few months ago, but that shouldn't be hard to fix at all. The issue I need to dig into is the network part. The other prior issue was with volume attachments, but now that there are dedicated resources for volume attachments, I'm not concerned about those anymore.

I have a few other Terraform/OpenStack bugs to look into and I'll add this to my list. I'm going to leave this PR open for:

* Reminder to me to not drop the ball on this one again
* If you wanted to look into getting the `network` attribute read in correctly, go for it. I'd be happy to answer any questions :)

yagnik:
I don't think I'm going to get to this anytime soon :(. Going to close this mr for now.

PR #11381, opened 2017-01-24
Add pass provider

mcanevet:
Add a new provider for `pass` (http://www.passwordstore.org/).

Works like the `Vault` provider but uses pass.


mcanevet:
@stack72 is there something preventing this PR to be merged?

builtin/providers/pass/data_source_password_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	r "github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestDataSourcePassword(t *testing.T) {
+	r.Test(t, r.TestCase{
+		Providers: testProviders,
+		//PreCheck:  func() { testAccPreCheck(t) },
```
kwilczynski: Please more code that is not needed.

builtin/providers/pass/resource_pass_password.go
```diff
@@ -0,0 +1,78 @@
...
+			"path": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Full path where the pass password will be written.",
+			},
+
+			// Data is passed as JSON so that an arbitrary structure is
+			// possible, rather than forcing e.g. all values to be strings.
+			"data": &schema.Schema{
```
kwilczynski: Would this benefit from JSON validation helper to be added here?

kwilczynski:
@mcanevet would this require someone to install to get the `pass` binary and for the binary to be available in the `PATH`?

mcanevet:
@kwilczynski yes, this provider will not work without the `pass` program available in the `PATH`.

builtin/providers/pass/resource_pass_password.go
mcanevet: I'm not sur this is required, as if the password can't be unmarshaled, we fallback to plain text.

mcanevet:
@stack72 is there something preventing this PR to be merged?

raphink:
@stack72 is this missing something to be merged?

mcanevet:
@kwilczynski is there something preventing this PR to be merged?

mcanevet:
Could anyone take a look at this?
I can perfectly understand that you don't want to merge this because you don't want to maintain this provider on the long term, but if so, please tell me so that I close this PR and maintain it as a plugin instead.
/cc @stack72 @kwilczynski

beddari:
We're another org interested in the outcome of this. Not sure about what the current policy is, however I'm pretty certain the Terraform using pass community is large enough to maintain this 👍

mcanevet:
https://github.com/camptocamp/terraform-provider-pass

PR #11383, opened 2017-01-24
Fix: Incorrect AWS Lambda Qualifier Regexp

brad-larson:
Type of change:
===============
- Bug fix

What changed? ... and Why:
==========================
The regexp is currently set to:
`pattern := ^[a-zA-Z0-9$_]+$`

The AWS docs state that qualifer names must conform to the following
regexp:
`Pattern: (|[a-zA-Z0-9$_-]+)`

As you can see, the current regexp in Terraform is missing the `-` at
the end.

This addresses that.

How has it been tested?
=======================
Added a few test cases to the existing spec for `AwsLambdaQualifier`
validation.

### DIFF
--- a/builtin/providers/aws/validators.go
+++ b/builtin/providers/aws/validators.go
@@ -251,7 +251,7 @@ func validateLambdaQualifier(v interface{}, k string) (ws []string, errors []err
 			"%q cannot be longer than 128 characters: %q", k, value))
 	}
 	// http://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html
-	pattern := `^[a-zA-Z0-9$_]+$`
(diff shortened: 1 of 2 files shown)


catsby:
Hey @brad-larson – can you clarify for me what resources this impacts, and the documentation you're referencing? I'm looking at the API docs here:

- http://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html

I'm not seeing the trading `-` as you mention. Can you show me what I'm missing?

brad-larson:
This impacts the `AwsLambdaPermission` resource,  and the docs link is: http://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html

catsby:
AH ok, thanks!

Issue #3551, opened 2015-10-19
aws_iam_instance_profile accepts roles[] but should only accept role


gtmtech:
aws_iam_instance_profiles accepts a roles[] array, indicating attaching 2 roles, but:

http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html

indicates that you can only have 1 role attached to an instance_profile.

johnrengelman:
That's only a default maximum. I believe AWS can increase that limit at your request.

simonvanderveldt:
This is a duplicate of #3851

stack72:
Closing as a duplicate of #3851 - there is a lot of discussion in that issue that can be kept track of

PR #11385, opened 2017-01-24
Update documentation for roles attribute for iam_instance_profile

misham:
- When creating an `iam_instance_profile` you will receive an error if you have multiple roles defined but have not increased your AWS limit for the number of roles you can assign to an `iam_instance_profile`.
- See more on defaults: http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html


misham:
Hit similar issues as detailed in #3851 and found related comments in #3551 that pointed to the problem I am seeing

catsby:
Thank!

PR #11388, opened 2017-01-24
dependency: update go-getter

catsby:
Update `go-getter` to include the latest updates from everyone's favorite @ryanuber


ryanuber:
This changed ever so slightly - if you could pull again before merge, that would be fantastic!

catsby:
@ryanuber bumped to cc80f38c726badeae53775d179755e1c4953d6cf, thanks!

PR #11389, opened 2017-01-24
ip_address is an array and requires the instance number

cblecker:
Fixes #9088.
Original PR is stale since Sept.


stack72:
LGTM! Thanks @cblecker :)

PR #11431, opened 2017-01-26
core: MergeDiff to delete nested entries in fields that became empty

radeksimko:
Related:
https://github.com/hashicorp/terraform/pull/11245

After merging #11245 a few acceptance tests started to panic.
I've verified that those panicking tests are testing genuine behaviour - something user would normally do too. The problem occurs when you have a resource w/ `TypeList` that has an existing state and you decide to remove that whole `TypeList` block in the config.

I don't know if the motivation behind the panic was the expectation that we'd always produce `InstanceDiff` where `TypeList` w/ `NewRemoved: true` also has all nested fields `NewRemoved: true` - that is currently _not_ the case. If that's expected, then this should be fixed elsewhere.

If it's supposed to be the responsibility of `MergeDiff` to figure this out then I think this is ok as is. 

I ran the full aws acceptance test suite and the panics all disappeared & new failures were just coming from flappy tests.

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -1678,15 +1678,14 @@ func (s *InstanceState) MergeDiff(d *InstanceDiff) *InstanceState {
 	}
 
 	for _, k := range deleted {
-		// Sanity check for invalid structures.
-		// If we removed the primary count key, there should have been no
+		// If we removed the primary count key, we also need to remove
 		// other keys left with this prefix.
 
 		// this must have a "#" or "%" which we need to remove
 		base := k[:len(k)-1]
 		for k, _ := range result.Attributes {
 			if strings.HasPrefix(k, base) {
-				panic(fmt.Sprintf("empty structure %q has entry %q", base, k))
+				delete(result.Attributes, k)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


jbardin:
I wonder if we _should_ fix this elsewhere as you mention here. 

The goal of the panic here was to enforce that we only merge precise diffs, and in this case the diff is missing `NewRemoved: true` for `"alarm_configuration.0.ignore_poll_alarm_failure"` (probably because the default value doesn't change).

While this will certainly patch up the diff at this stage, would fixing the diff generation benefit us in any other way? I was kind of hoping that flushing out inconsistencies would catch other diff related bugs, but if this puts too much onus on provider itself we might as well patch it here.

radeksimko:
Yeah, we should fix it elsewhere. Thanks for confirming my thoughts. 👍 

I'll dive deeper. 🌊

danawillow:
I ran into the issue that inspired this PR today when trying to figure out why my tests in the google provider were failing- I was deleting a `TypeList` block that had not every field filled in (and hence some were left in the `InstanceState` with default values). Mind keeping me in the loop when a fix has been decided on? Thanks!

Issue #10968, opened 2016-12-30
[aws]ECS Task placement


spy-tech:
Please add support for the following feature to aws_ecs_service:

![image](https://cloud.githubusercontent.com/assets/23215674/21561807/24e94fa6-ce72-11e6-8250-0ef8d7a07a92.png)

This is *not* part of the task definition attribute as it exists in Terraform.

catsby:
Should be covered in https://github.com/hashicorp/terraform/pull/11030

mykyta-protsenko:
#11030 covers placement_constraints for ECS *tasks definitions*

ECS now supports different placementStrategies for ECS *services*, which is a different thing and should be defined on aws_ecs_service level:

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-strategies.html

It is a part of AWS' CreateService call:

http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html
...
`   "placementConstraints": [ 
      { 
         "expression": "string",
         "type": "string"
      }
   ],
   "placementStrategy": [ 
      { 
         "field": "string",
         "type": "string"
      }
   ],`

spy-tech:
@catsby It isn't. https://github.com/hashicorp/terraform/issues/11077.

catsby:
Ok I'll take a look thanks

radeksimko:
The following attributes were introduced:
 - `placement_strategy` for `aws_ecs_service` - https://github.com/hashicorp/terraform/pull/11201
 - `placement_constraint` for `aws_ecs_service` - https://github.com/hashicorp/terraform/pull/11242
 - `placement_constraints` for `aws_ecs_task_definition` - https://github.com/hashicorp/terraform/pull/11030 - already released; part of `v0.8.4` and `v0.8.3`

The `aws_ecs_service` updates will be part of a next release.

Related documentation for ECS service is also being fixed in https://github.com/hashicorp/terraform/pull/11338

I think it's time to close this issue. Thanks to all involved. 🙂  🎉

spy-tech:
@radeksimko No, it's not time to close this, because it does not work. 

Using Terraform v0.8.5-dev (cb4ffd466222a2bc6839ce5989a73a63ed911202+CHANGES)

I get: 
```
InvalidParameterException: distinctInstance expression should not be specified.
        status code: 400, request id: `<snip>
```
for the following placement_constraints definition as part of a `aws_ecs_service` definition. 

```
[...]

PR #11446, opened 2017-01-26
AWS ECS placement constraints fix

bgetsug:
I've been working with the `master` branch because I want to take advantage of AWS ECS Task Placement Constraints. When trying to create the following ECS Service resource, I received a 400 error from the AWS API:

```hcl
resource "aws_ecs_service" "my_service" {
  name                               = "my-service"
  cluster                            = "${aws_ecs_cluster.aws_ecs_cluster.id}"
  task_definition                    = "${aws_ecs_task_definition.my_api.arn}"
  desired_count                      = 6
  deployment_minimum_healthy_percent = 33
  deployment_maximum_percent         = 200
  iam_role                           = "${var.ecs_service_role_arn}"

  load_balancer {
    target_group_arn = "${module.alb.alb_target_group_arn}"
    container_name = "my-api"
    container_port = "${var.api_port}"
  }

  placement_constraints {
    type = "distinctInstance"
  }
}
```

The error:
```
* aws_ecs_service.my_service: InvalidParameterException: distinctInstance expression should not be specified.
        status code: 400, request id: 44618ece-e411-11e6-8e4c-0b431ab05b6b

```

 This seems to be caused by the constraint `expression` being sent as an empty string ("") with the `type` of `distinctInstance`. I guess they don't want "expression" in the request body at all with that type.

Additionally the ECS Task Definition resource documentation is incorrect. `distinctInstance` can be used on ECS Services, but not Task Definitions. See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#constraints.

The changes outlined here seem to fix the issue, though I'm not sure if it's the best approach, since this is my first time digging into Terraform code. If I have more time, I'll try to work through the acceptance test docs. **I have confirmed that the issue still exists in 0.8.5**


spy-tech:
Yes, I reported the same bug here: https://github.com/hashicorp/terraform/issues/10968 (see my last comment).

grubernaut:
@bgetsug thanks for this, it's much appreciated! Closing in favor of #11475

PR #11452, opened 2017-01-27
provider/postgres grant role when creating database

pbthorste:
The code here is reintroduced from older commits. See:

https://github.com/hashicorp/terraform/commit/2e529146a5e92251081b3c75fba6ecabbac37c3d

and

https://github.com/hashicorp/terraform/commit/f3add9e7efbb3066a743d88a86cf8c8f61cc4c0a

(the part related to grantRoleMembership).

Without this code - creating a new database with a terraform generated role as the owner will not work. You will get the error:

```bash
Error applying plan:

1 error(s) occurred:

* postgresql_database.my-db: Error creating database database1: pq: must be member of role "my-role"
```

### DIFF
--- a/builtin/providers/postgresql/resource_postgresql_database.go
+++ b/builtin/providers/postgresql/resource_postgresql_database.go
@@ -125,7 +125,7 @@ func resourcePostgreSQLDatabaseCreate(d *schema.ResourceData, meta interface{}) 
 	//needed in order to set the owner of the db if the connection user is not a superuser
 	err = grantRoleMembership(conn, d.Get(dbOwnerAttr).(string), c.username)
 	if err != nil {
-		return err
+		return errwrap.Wrapf(fmt.Sprintf("Error granting role membership on  database %s: {{err}}", dbName), err)
 	}
 
 	// Handle each option individually and stream results into the query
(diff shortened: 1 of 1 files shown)


pbthorste:
Another issue that this commit fixes: Roles cannot be destroyed after they are created. Error is:

```bash
Error applying plan:

1 error(s) occurred:

* postgresql_role.my-role: Error deleting role: pq: permission denied to reassign objects
```

martynsmith:
This is definitely an issue with AWS RDS, any idea if/when it'll be merged?

pbthorste:
hopefully soon. There is a workaround - but then you have to manually log into the postgres instance and grant the role.

mindw:
@pbthorste can it be done from provisioner? do you have a code example handy?

Thanks!

pbthorste:
@mindw No, it is a manual operation - I'll try to explain the steps :)

1. Create db / use a provider
If you create a postgres database with "aws_db_instance", you have a username
which is usually the same one as the one used in the postgresql provider. Lets
call that user: 'provider_user' (this user is not a superuser)

2. Create a role
Then you create a new role with 'postgresql_role' - lets call that 'new_role'.

3. Create a database
This will fail: you can't then create a database with postgresql_database where the owner is the 'new_role'.

That is until you log into your postgres instance and run:

```sql
GRANT new_role to provider_user;
```
after this granting has taken place, you can run terraform and the db will be created.

FlorinAndrei:
Without this PR the Postgresql provider in Terraform is broken if you use RDS. Please merge.

mindw:
Any news on this one?
Thanks!

PR #11467, opened 2017-01-27
Mention how to use git modules over SSH

eraserhd:
How this came about:

The only mention of using private modules is from a CI server.  This suggests using a username and password.  This doesn't work for me, since I (and everyone in our org) have 2FA enabled, so instead of a username and password, we would need a username and a personal auth token.

However, GitHub will not let us generate a personal auth token that can access private content.  So that is a dead end.  And this is not for a CI type environment.

Digging into the code of `go-getter`, I found out that strings prefixed with "git@github.com:" will be translated into SSH URLs.  This works for us.

### DIFF
--- a/website/source/docs/modules/sources.html.markdown
+++ b/website/source/docs/modules/sources.html.markdown
@@ -57,6 +57,14 @@ module "consul" {
 }
 ```
 
+These will fetch the modules using HTTPS.  If you want to use SSH instead:
+
+```
+module "consul" {
+	source = "git@github.com:hashicorp/example.git//subdir"
+}
+```
+
 **Note:** The double-slash, `//`, is important. It is what tells Terraform that that is the separator for a subdirectory, and not part of the repository itself.
 
 GitHub source URLs require that Git is installed on your system and that you have access to the repository.


apparentlymart:
There is also a more general syntax that works for non-Github git repositories:

```
git::ssh://username@hostname/path/to/repo.git//path-within-repo
```

For github (should you not want to use the shorthand you mentioned for some reason) this would look like the following:

```
git::ssh://git@github.com/username/reponame.git//path-within-repo
```

stack72:
@apparentlymart anything you want to see added here or is this GTG?

kojiromike:
Some folks looking at this PR may find this useful. We use local-path modules in our day-to-day, but sometimes when we update a local-path module we have to maintain resources with the previous version of that module. Since we are using a complete, nonshallow clone of the repo for the module, I've found it's fast and convenient to refer to the gitdir itself. So whereas we normally have
```
source  = "../../modules/cluster/iam"
```
sometimes we have
```
= "git::../../../.git?ref=tf-pub-dns//terraform/modules/cluster/iam"
```

This seems to work even if the .gitdir is detached.

apparentlymart:
Sorry I think when I read this the first time in understood it as an issue rather than a PR, so I was "me-tooing" it. Didn't mean to block it being merged.

stack72:
after speaking to @apparentlymart - we believe this is good to mege :)

PR #11472, opened 2017-01-27
WIP: circonus provider

sean-:
There are a few rough edges that I need to work through but I want to get a pair of eyes on this sooner rather than later.  I plan on merging this with a squash commit.


sean-:
Either after this PR or before this is merged I'm going to move the tests over to a new regression testing account (credentials and account have just been provisioned).

catsby:
Hey Sean –

Thanks for the PR here! I have some concerns about the code however, so I’m afraid we’ll be unable to merge it at this time. 

The main objections may seem to be simple style conflicts, but none the less I feel they are important. This Provider looks and may perhaps behave very different compared to our other providers. For example, most things here are prefaced with `_`, which doesn’t match up with the style of any of our other Providers. Can you elaborate why functions, methods, and types all use this prefix? Resource attribute names are constants instead of the typical inline strings, and many common things like saving an attribute to state are wrapped in their own methods. 

A specific example is `_StateSet`, a wrapper around `d.Set`. This method panics at runtime if it fails. In the event an attribute fails to save to state, Terraform should continue running (ideally logging the error) and complete the plan to the best of its ability. Future `plan` runs should reveal any drift or otherwise missing state information, and the logs will provide information as to why the data/state is missing. For new provider/resources, we do routinely check the return of `d.Set` for errors and log accordingly, but a panic in the case of error here can lead to data loss (specifically for that resource at least), and a significant confidence lost in terms of an operator using Terraform to manage their infrastructure. 

The pattern of prefixing most things with the underscore does not match any of the established code style of the project. Further, the underscored constants, functions, etc. also seem to be exported. This seems to suggest that in theory someone could consume those APIs across another package. But the underscore seems to suggest they're private in some way. This is confusing and not idiomatic to Go nor the Terraform project.

The custom methods and functions (ex. attr_reader_config.go) that seem to simply wrap Terraform’s DSL methods (d.Set, d.Get, d.GetOk), appear unnecessary and just add more to the codebase for future maintainers to navigate and reason about, for little gain. The minor improvement in lines of code from these does not benefit a maintainer long term since they'll appear foreign and have to be re-learned for any maintenance. 

There is also a lack of commenting in many places where it is unclear what the purpose of a piece of code is. For example, `_AttrReader`. @mitchellh in many cases encourages erring on the side of over-commenting. It often adds redundancy for what may be construed as "obvious" in the code, but it helps for inexperienced folks approaching the codebase (i.e. long term maintainers). As a rough check: an interface such as `_AttrReader` should probably have a minimum one paragraph comment for code docs, but perhaps more. This ignores the fact that stylistically it's odd to begin with (prior paragraph). 

I feel that the maintenance burden brought on by merging this Provider as is would be unreasonably high for anyone but yourself. Terraform has a wonderful community that engages in all areas of the project, and I’m concerned that the stark differences here create a very high barrier to entry for the community to contribute to this Provider. 

I welcome discussion via Zoom or chat if you’d like, but I can’t merge the pull request in it’s current form. I've double checked with additional team members on this, since you obviously put in a great deal of work here and it never feels good as a maintainer to have to reject that, but we've treated this PR just as we would any other.

Please let me know if you have any questions or need me to clarify anything here. I hope I’ve given you guidance on a path forward here, but I’m happy to clarify if needed.

Thanks!
Clint

mitchellh:
I want to add that @catsby requested I take a look as well and I came to the same general conclusions as above. I believe @catsby hit the stylistic issues correctly, which is 100% spot on, so I'll just talk about implementation itself.

Ultimately, the issue I had with this provider was that it does too much: its both trying to implement a provider as well as introduce a _new way_ of doing so. Both are valid discussions, but separate.

Again, sorry. I know there is a lot of work already done here and its super painful as a maintainer to have to say no to that. But there is a lot of improvement to be done here still so I don't think its right to hold open a "WIP" PR until it is more ready. In the future please feel free to ping us with work-in-progress samples (for example after doing 1 resource + basic config) and we can provide guidance to avoid overcommitting in the future. This PR was titled "WIP" but it was very large for an initial request for feedback.

Issue #11464, opened 2017-01-27
terraform import aws_alb_listener fails on listener rules with priority value "default"


AnthonyNeace:
`terraform import aws_alb_listener` fails during refresh for listener rules with priority value "default". This appears to be because [priority is expected to be an int](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_alb_listener_rule.go#L35), however AWS supplies a valid priority value "default" for default rules, and this value doesn't appear to be correctly supported (or otherwise mapped to an int?) by Terraform.

### Terraform Version

0.8.5

### Affected Resource
- aws_alb_listener

### Expected Behavior

Successful refresh, with aws_alb_listener imported to tfstate.

### Actual Behavior

Failure on refresh, with this error:

```
Error importing: 1 error(s) occurred:

* aws_alb_listener_rule.test: Cannot convert rule priority %q to int: strconv.ParseInt: parsing "default": invalid syntax
```

### Steps To Reproduce

Manually create an ALB with a single listener (of arbitrary configuration, HTTP 80 forward would be fine) and some arbitrary target group.  It doesn't need any targets registered/instances attached, or anything like that... can have a very simple configuration aside from the single listener. After creation, note that your rule priority's value is "default" in the web console, for this created default rule.

This can also be verified via CLI:

```
aws elbv2 describe-rules --rule-arns <myRuleArn>
{
    "Rules": [
        {
            "Priority": "default",
            "Conditions": [],
            "RuleArn": "<myRuleArn>",
            "IsDefault": true,
            "Actions": [
                {
                    "TargetGroupArn": "<myTargetArn>",
                    "Type": "forward"
                }
            ]
        }
    ]
}
```

Now, attempt to import, and observe that refresh fails with the following error:

```
terraform import aws_alb_listener_rule.test <myArn>

aws_alb_listener_rule.test: Importing from ID "<myArn>"...
aws_alb_listener_rule.test: Import complete!
  Imported aws_alb_listener_rule (ID: <myArn>)
aws_alb_listener_rule.test: Refreshing state... (ID: <myArn>)
Error importing: 1 error(s) occurred:

* aws_alb_listener_rule.test: Cannot convert rule priority %q to int: strconv.ParseInt: parsing "default": invalid syntax
```

PR #11474, opened 2017-01-27
provider/aws: Import aws_alb_listener_rule fix

grubernaut:
Allows the user to import a default aws_alb_listener_rule. When creating the resource with TF, the AWS API requires that `priority` be an integer.
However, the `DescribeRules` API call returns a string for `priority`. This would work in every case except for the `default` listener rule, which sets the `priority` value to the string "default".

This fixes the previous error:
```
Error importing: 1 error(s) occurred:

* aws_alb_listener_rule.test: Cannot convert rule priority %q to int: strconv.ParseInt: parsing "default": invalid syntax
```

Fixes: #11464

### DIFF
--- a/builtin/providers/aws/resource_aws_alb_listener_rule.go
+++ b/builtin/providers/aws/resource_aws_alb_listener_rule.go
@@ -144,10 +144,15 @@ func resourceAwsAlbListenerRuleRead(d *schema.ResourceData, meta interface{}) er
 	rule := resp.Rules[0]
 
 	d.Set("arn", rule.RuleArn)
-	if priority, err := strconv.Atoi(*rule.Priority); err != nil {
-		return errwrap.Wrapf("Cannot convert rule priority %q to int: {{err}}", err)
+	// Rules are evaluated in priority order, from the lowest value to the highest value. The default rule has the lowest priority.
+	if *rule.Priority == "default" {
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_alb_listener_rule.go
```diff
@@ -144,10 +144,15 @@ func resourceAwsAlbListenerRuleRead(d *schema.ResourceData, meta interface{}) er
 	rule := resp.Rules[0]
 
 	d.Set("arn", rule.RuleArn)
-	if priority, err := strconv.Atoi(*rule.Priority); err != nil {
-		return errwrap.Wrapf("Cannot convert rule priority %q to int: {{err}}", err)
+	// Rules are evaluated in priority order, from the lowest value to the highest value. The default rule has the lowest priority.
+	if *rule.Priority == "default" {
+		d.Set("priority", 99999)
```
stack72: why 99999?

grubernaut: See the comment above, the `default` rule priority is the lowest value, which happens to be `99999` :)

stack72: SHIT IT :)

PR #11477, opened 2017-01-27
provider/google: Fix master_instance_name to prevent slave rebuilds

gh-mlfowler:
The v1beta4 API returns the `master_instance_name` for a Second Generation Replica prefixed with the `project` which is not what Terraform is expecting, nor is it required. With the `project` prefixed Terraform plans to recreate the slave on every subsequent apply. The read method has been adjusted to strip the `project` prefix so slaves can persist.

Two minor fixes are included that were uncovered in the acceptance test. `crash_safe_replication` is only relevant for First Generation instances so I have not specified it. With the slave the API returns `true` even if we haven't set the value. Terraform will attempt to set it to `false` on the next apply which appears to be a no-op in Google as the next plan will show `true`. I have changed this to a `Computed` value as it is a boolean.

Not specifying the optional `start_time` for `backup_configuration` causes subsequent plans to attempt to adjust the Google default to an empty value. As this is a string I have changed the read to return no value if we haven't set the value.

### DIFF
--- a/builtin/providers/google/resource_sql_database_instance.go
+++ b/builtin/providers/google/resource_sql_database_instance.go
@@ -3,6 +3,7 @@ package google
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -70,6 +71,7 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 						"crash_safe_replication": &schema.Schema{
 							Type:     schema.TypeBool,
 							Optional: true,
+							Computed: true,
 						},
 						"database_flags": &schema.Schema{
 							Type:     schema.TypeList,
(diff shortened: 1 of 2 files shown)


danawillow:
```
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccGoogleSqlDatabaseInstance -timeout 120m
--- PASS: TestAccGoogleSqlDatabaseInstance_basic (40.18s)
--- PASS: TestAccGoogleSqlDatabaseInstance_basic2 (50.78s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_basic (31.48s)
--- PASS: TestAccGoogleSqlDatabaseInstance_slave (643.12s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_upgrade (54.92s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_downgrade (35.78s)
PASS
```

LGTM. Thanks for the contribution!

Issue #1497, opened 2015-04-12
count parameter needs to be able to interpolate variables from modules.


bobtfish:
I've been trying to extend my Terraform examples to support multiple AZs by default.

As such, I have an _az_count_ variable output by my module to detect the AZs you have available:

https://github.com/terraform-community-modules/tf_aws_availability_zones#outputs

And I then want to reuse it for instances, for example here:

https://github.com/bobtfish/terraform-vpc/blob/master/main.tf#L37

My example (https://github.com/bobtfish/terraform-example-vpc/tree/master/eucentral1-demo) crashes out with the error:
- aws_instance.nat: resource count can't reference module variable: module.vpc.az_count

and if I remove this error from the source (hope springs eternal!), I run into:
- strconv.ParseInt: parsing "${module.azs.az_count}": invalid syntax

This inability to interpolate count variables is a blocker for me being able to write region independent modules - as (for example) I want to be able to allocate one subnet per AZ, writing code like:

```
module "azs" {
    source = "github.com/terraform-community-modules/tf_aws_availability_zones"
    account = "${var.account}"
    region = "${var.region}"
}

resource "aws_subnet" "front" {
     count = "${module.vpcs.az_count}"
     ...
 }     
```

Even better, I'd like to be able to interpolate variables out of one module, and into the user variables of another, for example:

```
module "azs" {
    source = "github.com/terraform-community-modules/tf_aws_availability_zones"
    account = "${var.account}"
    region = "${var.region}"
}

variable "az_count" {
    default = "${module.azs.az_count}"
}

resource "aws_subnet" "front" {
     count = "${var.az_count}"
     ...
}
```

jtopper:
:+1: 

Surprised to see I can't do `count = ${var.az_count}` here.

progrium:
+1

phinze:
We should definitely do this, the tricky part comes from the fact that count expansion is currently done statically, before the primary graph walk, which means we can't support "computed" counts right now. (A "computed" value in TF is one that's flagged as not known until all its dependencies are calculated.)

Related: https://github.com/hashicorp/terraform/pull/354 

So we'd need to rework resource count expansion to occur "just in time" to support this. Probably the right thing to do in the long run, but that's what makes this nontrivial to fix.

jtopper:
Sounds like I might need to preprocess these .tf files for a little while then. I don't suppose you have a sense of where in the roadmap this rework might happen?

blalor:
[...]

PR #11482, opened 2017-01-28
core: allow non-computed data source values in "count"

mitchellh:
Based on #10418, but simpler due to new graphs (more LOC change due to tests).

This disables the computed value check for `count` during the validation
pass. This enables partial support for #3888 or #1497: as long as the
value is non-computed during the plan, complex values will work in
counts.

**Notably, this allows data source values to be present in counts!**

The "count" value can be disabled during validation safely because we
can treat it as if any field that uses `count.index` is computed for
validation. We then validate a single instance (as if `count = 1`) just
to make sure all required fields are set.

This still won't allow "computed" data source values: data sources that
themselves depend on the value of some computed resource (non-data source).

As promised, this loosens the strict requirement of computed count values.
I'm still trying to make this even better for 0.9...

### DIFF
--- a/builtin/providers/test/data_source_test.go
+++ b/builtin/providers/test/data_source_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 	"testing"
 
@@ -48,6 +49,49 @@ resource "test_resource" "foo" {
 					}
 					if res.Primary.Attributes["list.2"] != "count-2" {
 						return errors.New("Wrong list.0, expected count-2")
+					}
+					return nil
+				},
+			},
+		},
+	})
+}
+
+// Test that the output of a data source can be used as the value for
+// a "count" in a real resource. This would fail with "count cannot be computed"
+// at some point.
+func TestDataSource_valueAsResourceCount(t *testing.T) {
+	resource.UnitTest(t, resource.TestCase{
+		Providers: testAccProviders,
+		CheckDestroy: func(s *terraform.State) error {
+			return nil
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: strings.TrimSpace(`
+data "test_data_source" "test" {
+  input = "4"
+}
+
+resource "test_resource" "foo" {
+  count = "${data.test_data_source.test.output}"
+
+  required     = "yep"
+  required_map = {
+    key = "value"
+  }
+}
+				`),
+				Check: func(s *terraform.State) error {
+					count := 0
+					for k, _ := range s.RootModule().Resources {
+						if strings.HasPrefix(k, "test_resource.foo.") {
+							count++
+						}
+					}
+
+					if count != 4 {
+						return fmt.Errorf("bad count: %d", count)
 					}
 					return nil
 				},
--- a/command/plan_test.go
+++ b/command/plan_test.go
@@ -575,7 +575,7 @@ func TestPlan_validate(t *testing.T) {
 	}
 
 	actual := ui.ErrorWriter.String()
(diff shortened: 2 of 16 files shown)


myoung34:
@mitchellh will this have any effect on #7034 ? namely inline settings blocks such as

```hcl
   setting {
     count = "${var.spin_down_at_night == "true" ? 1 : 0}"
     namespace = "aws:autoscaling:scheduledaction"
     resource = "ScheduledPeriodicScaleup"
     name = "Recurrence"
     value = "0 10 * * *" #10am UTC = 4am CST
   }
```

mitchellh:
@myoung34 Zero effect for or against, we'd like to support that though.

myoung34:
Thanks for the quick reply. Too bad, but at least it's not a "no"

odupuy:
Can you give an example of what becomes doable with this fix?

mitchellh:
@odupuy here is an abstract example: 

```
data "foo" "bar" {}
resource "baz" "qux" { count = "${data.foo.bar.value}" }
```

More concrete examples just in human words since I don't know their exact syntax/attributes: reading a Consul key to get the count, reading data about an AWS instance and using that for count, etc.

RichardBronosky:
I just discovered that deterministic math is acceptable even though count must be "non-computed".

https://gist.github.com/RichardBronosky/9ba397e85e4af86e517b33af32ad8706

mitchellh:
@RichardBronosky that is non-computed! I meant in the Terraform definition of the word "computed", which is: value cannot be determined until apply-time (outputs of to-be-changed resources basically).

thatderek:
Another example for this is setting up a subnet in every availability-zone, regardless of which region you're executing a plan in. A la: 

```
data "aws_availability_zones" "available" {}

resource "aws_subnet" "public" {
  count = "${length(data.aws_availability_zones.available.id)}"

  availability_zone = "${data.aws_availability_zones.available.id[count.index]}"
}
```

whereas before my subnetting module had a integer variable with a default value (of three), and I had to have extensive explanations in my README about what defaults to change for regions with only 2 AZs.

kojiromike:
For a slightly more concrete example, I'm (expecting) to be able to do this:

```
data "template_file" "cname" {
  template = "/dev/null"

  vars {
    count = "${1 + length(var.additional_cnames)}"
    primary = "${join("-", compact(list(var.customer, var.environment == "prod" ? "" : var.environment)))}"
  }
}

resource "null_resource" "cname" {
  count = "${data.template_file.cname.vars.count}"
  triggers {
    value = "${count.index == 0 ? data.template_file.cname.vars.primary : element(var.additional_cnames, 1 + count.index)}"
  }
}

resource "aws_route53_record" "cname" {
  count = "${data.template_file.cname.vars.count}"
  name    = "${element(null_resource.cname.*.triggers.value, count.index)}"
  records = ["${aws_instance.ngnx.public_dns}"]
  ttl     = "60"
  type    = "CNAME"
  zone_id = "${data.terraform_remote_state.base.my_zone_id}"
}
```

mitchellh:
@kojiromike looking at that, I believe that'll work.

greenboxal:
Awesome! What are your plans on computed counts?

I'm using the changes that I made on #10418 and I have one issue when creating new resources with computed resource counts.

Basically I need to run apply two times: On the first it creates one resource (assumes count = 1, as in validation). In the second, it creates the resources correctly as the dependencies are resolved in the state.

apparentlymart:
@greenboxal I think you're describing something like what was discussed in #4149. That's a more complete solution to the "computed count" problem based on multiple runs of `terraform apply`, as you're suggesting.

Issue #9861, opened 2016-11-04
provider/pagerduty: allow "null" as an option for auto_resolve_timeout in pagerduty_service


EvanKrall:
### Terraform Version
`Terraform v0.7.8`

### Affected Resource(s)
- pagerduty_service

### Terraform Configuration Files
```hcl
resource "pagerduty_service" "something" {
    name = "something"
    escalation_policy = "Some escalation policy id"
    acknowledgement_timeout = "${24 * 60 * 60}"
    auto_resolve_timeout = null
}
```

### Expected Behavior
This should set auto_resolve_timeout to null, so as to disable the feature, as described on https://www.terraform.io/docs/providers/pagerduty/r/service.html.

### Actual Behavior
```
Errors:

  * pagerduty_service.something: cannot parse '' as int: strconv.ParseInt: parsing "null": invalid syntax
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

heimweh:
Hi @EvanKrall. Thanks for reporting this.

It seems like the API allows the value **null** (tested it via https://v2.developer.pagerduty.com/v2/page/api-reference#!/Services/put_services_id) to be passed which would disable the `auto_resolve_timeout` (also affects `acknowledgement_timeout`) but it doesn't seem like the client library support it yet. 

Really sorry about this, will take a look and see if I can get this fixed asap.

A possible work-around for now is to not set the `auto_resolve_timeout` and then manually disable it in the UI (tested this quick and this shouldn't cause a diff in Terraform).

paddyforan:
I've labeled this as an enhancement to track the progress of bringing the feature to Terraform, but in the meantime, is it worth updating the docs to correct the available options?

heimweh:
@paddyforan thanks, absolutely that sounds good :)
Gimme a few minutes and I'll push a doc update.

heimweh:
@paddyforan I pushed a fix for this here (#9881), hopefully that should be enough until we get a proper fix out. Once again, really sorry about this.

paddyforan:
Hey @heimweh, thanks for addressing that so quickly! I've gone ahead and updated the Terraform website with the fix, so it's live now. No worries, things happen. Thanks so much for all your work on the pagerduty provider. I'm gonna actually reopen this issue just so we can track the missing feature, but it's great we've got the accurate docs live now.

heimweh:
Hi everyone,

[...]

PR #11483, opened 2017-01-28
provider/pagerduty: Allow timeouts to be disabled (pagerduty_service)

heimweh:
This PR aims to solve the issue described in #9861 where 
auto_resolve_timeout & acknowledgement_timeout can't be disabled via Terraform.

Old behaviour:
If auto_resolve_timeout & acknowledgement_timeout wasn't set, 
API defaults would be used which is:
```
acknowledgement_timeout: 1800
auto_resolve_timeout:    14400
```

New behaviour:
The default for both auto_resolve_timeout & acknowledgement_timeout is now:
`0` which disables the timeouts. 

Please note though that this is a breaking change because the user would now have to explicitly pass the values for these two timeouts in order for them to be enabled.

### DIFF
--- a/builtin/providers/pagerduty/resource_pagerduty_service.go
+++ b/builtin/providers/pagerduty/resource_pagerduty_service.go
@@ -117,6 +117,10 @@ func resourcePagerDutyServiceRead(d *schema.ResourceData, meta interface{}) erro
 	service, err := client.GetService(d.Id(), o)
 
 	if err != nil {
+		if isNotFound(err) {
+			d.SetId("")
+			return nil
+		}
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


heimweh:
Tests are passing:
```
$ make testacc TEST=./builtin/providers/pagerduty TESTARGS='-run=TestAccPagerDutyService_Basic'                                                                                                     10:30:25
2017/01/28 10:30:47 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/pagerduty -v -run=TestAccPagerDutyService_Basic -timeout 120m
=== RUN   TestAccPagerDutyService_Basic
--- PASS: TestAccPagerDutyService_Basic (11.65s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/pagerduty	11.668s
```

stack72:
Hi @heimweh 

This LGTM!

Thanks

P.

PR #11490, opened 2017-01-28
build: Update travis to go1.7.5 and go1.8rc3 and Vagrantfile to go1.7.5

cblecker:


jen20:
Prior to this being merged, we need to ensure we won't get a repeat of the issues that plagued Go 1.8.4 with respect to root certificates on OS X when cross compiling - other than that, :+1:

cblecker:
That's actually what 1.7.5 fixes!
golang/go#18688
https://github.com/golang/go/commit/5c2b5ee3c8271c3d19d8f0c38804983950c752bf
https://github.com/golang/go/milestone/52?closed=1

stack72:
@cblecker in that case, LGTM! Thanks :)

PR #11500, opened 2017-01-29
fix wrong usage of hashicorp/errwrap

julius-bonial:
According to https://github.com/hashicorp/errwrap
'{{err}}' has to be used instead of '%s'

Without this patch, error output from terraform is missing important information:
* aws_cloudwatch_log_group.logs: Error Getting CloudWatch Logs Tag List: %s

With this patch, I get the important information. E.g.:
* aws_cloudwatch_log_group.logs: Error Getting CloudWatch Logs Tag List: AccessDeniedException: User: arn:aws:sts::XYZ:assumed-role/AAA-BBB-CCC/terraform-assuming-role-assume-role-ReadOnly is not authorized to perform: logs:ListTagsLogGroup on resource: arn:aws:logs:us-east-1:XYZ:log-group:logs:log-stream:


stack72:
Thanks for the work here @julius-bonial :)

this is a great catch!

Paul

Issue #11461, opened 2017-01-27
AWS assume_role doesn't support policy as a parameter as SDK supports


kumar999222:
Dear All,

As I can see AWS SDK support 5 parameter for Assume_Role.
            RoleArn                              // supported in terraform as well
            ExternalId                           // supported in terraform as well
            RoleSessionName              // supported in terraform as well
            Policy                                  // not supported
            DurationSeconds                // not supported

Requesting terraform team if "policy" support can be provided.

Thanks in Advance.

PR #11501, opened 2017-01-29
provider/aws: Add support for policy to AWS provider assume_role

stack72:
Fixes: #11461

This will allow the user to pass a policy to further restrict the use
of AssumeRole. It is important to note that it will NOT allow an
expansion of access rights

### DIFF
--- a/builtin/providers/aws/auth_helpers.go
+++ b/builtin/providers/aws/auth_helpers.go
@@ -146,8 +146,8 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 
 	// Otherwise we need to construct and STS client with the main credentials, and verify
 	// that we can assume the defined role.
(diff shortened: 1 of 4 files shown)


builtin/providers/aws/auth_helpers.go
```diff
@@ -146,8 +146,8 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 
 	// Otherwise we need to construct and STS client with the main credentials, and verify
 	// that we can assume the defined role.
-	log.Printf("[INFO] Attempting to AssumeRole %s (SessionName: %q, ExternalId: %q)",
-		c.AssumeRoleARN, c.AssumeRoleSessionName, c.AssumeRoleExternalID)
+	log.Printf("[INFO] Attempting to AssumeRole %s (SessionName: %q, ExternalId: %q, Policy: %q)",
```
jen20: We might want to consider logging the policy on a separate line here as it will be quoted JSON?

PR #11503, opened 2017-01-29
Closes #11053. Use environment specific client for accessing resources.

### LINKED ISSUE #11053: Rancher provider crash when Rancher API key is not valid
### Terraform Version
v0.8.2

### Affected Resource(s)
- rancher_environment
- rancher_registration_token
- rancher_registry_credential
- rancher_registry
- rancher_stack

```hcl
resource "rancher_stack" "foo" {
  name            = "foo"
  environment_id  = "${var.environment_id}"
  start_on_create = true
}
```

### Expected Behavior
Rancher stack creation

### Actual Behavior
Terraform crashes with
* rancher_stack.foo: unexpected EOF

johnrengelman:
When using access control, the Rancher global API may return 404 for
resources that exist and are accessible via the environment API.

### DIFF
--- a/builtin/providers/rancher/resource_rancher_registration_token.go
+++ b/builtin/providers/rancher/resource_rancher_registration_token.go
@@ -97,7 +97,11 @@ func resourceRancherRegistrationTokenCreate(d *schema.ResourceData, meta interfa
 
 func resourceRancherRegistrationTokenRead(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[INFO] Refreshing RegistrationToken: %s", d.Id())
-	client := meta.(*Config)
+	client, err := meta.(*Config).EnvironmentClient(d.Get("environment_id").(string))
+	if err != nil {
+		return err
+	}
+	// client := meta.(*Config)
 
 	regT, err := client.RegistrationToken.ById(d.Id())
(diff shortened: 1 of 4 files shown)


stack72:
Hey @johnrengelman 

How can we test this is actually fixing the bug?

Thanks

Paul

johnrengelman:
create a user account in rancher that is of type `user` and not owner. 
When that user logs in the first time, Rancher will create a new environment for it (this is how you can verify).
Then create an API key for that user account http://docs.rancher.com/rancher/v1.3/en/api/v2-beta/#api-keys-for-an-account and use those keys to communicate with Rancher.

stack72:
so is this backwards compatible for what is currently out there? or will it potentially cause breakages?

johnrengelman:
It's backward compatible.

stack72:
in that case :) I am happy!

stack72:
Which bug does this fix? It mentions itself

johnrengelman:
Weird, Github is doing something fancy with pull request - issue numbers now: https://github.com/hashicorp/terraform/issues/11053

johnrengelman:
oh no it's not...I flipped a number in the commit message!

PR #11520, opened 2017-01-30
Terraform provider ProfitBricks - Data Sources

jasminSPC:
While this PR is fresh I have two issues with unit tests:

1. resource_profitbricks_loadbalancer_test - The issue is within ProfitBricks API. When a load balancer is provisioned it creates a private LAN and attaches it to existing NIC. This is leaving .tfstate with "dangling resources". This is a known issue with ProfitBricks

2. resource_profitbricks_volume_test - The script used for unit tests is working when ran separately but once unit tests run I get this error: 
```
        testing.go:265: Step 1 error: Check failed: Check 1/2 error: Bad name: webserver
```
I'm not sure how to fix this.
@stack72 do you have any ideas?


jasminSPC:
I'm using this PR to update to ProfitBicks REST V3. Adding following parameters:

- nat parameter for Nics
- availabilityZone for Volumes

stack72:
👀 are on this now @jasminSPC

builtin/providers/profitbricks/data_source_datacenter.go
```diff
@@ -0,0 +1,76 @@
...
+	config := meta.(*Config)
+	profitbricks.SetAuth(config.Username, config.Password)
+
+	datacenters := profitbricks.ListDatacenters()
+
+	if datacenters.StatusCode > 299 {
+		return fmt.Errorf("An error occured while fetching datacenters %s", datacenters.Response)
+	}
+
+	name, _ := d.GetOk("name")
```
stack72: as it's a required field we can guarantee it's there so we can write

`name := d.Get("name").(string)`

builtin/providers/profitbricks/data_source_datacenter_test.go
```diff
@@ -0,0 +1,31 @@
+package profitbricks
+
+import (
+	"github.com/hashicorp/terraform/helper/resource"
+	"regexp"
+	"testing"
+)
+
+func TestAccDataSourceDatacenter_basic(t *testing.T) {
```
stack72: I'd like to see a test that matches a datacenter name here :)

builtin/providers/profitbricks/data_source_image.go
```diff
@@ -0,0 +1,107 @@
...
+	profitbricks.SetAuth(config.Username, config.Password)
+	profitbricks.SetDepth("5")
+
+	images := profitbricks.ListImages()
+
+	if images.StatusCode > 299 {
+		return fmt.Errorf("An error occured while fetching ProfitBricks locations %s", images.Response)
+	}
+
+	name, _ := d.GetOk("name")
```
stack72: Name isn't required, so we should check it's ok before using it

builtin/providers/profitbricks/data_source_datacenter.go
```diff
@@ -0,0 +1,76 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"location": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"id": {
```
stack72: We actually don't need to expose Id as a param - it's there by default

builtin/providers/profitbricks/data_source_image.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+			"location": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"version": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"id": {
```
stack72: We actually don't need to expose Id as a param - it's there by default

builtin/providers/profitbricks/data_source_image.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if len(results) > 1 {
+		return fmt.Errorf("There is more than one image that match the search criteria")
+	}
+
+	if len(results) == 0 {
+		return fmt.Errorf("There are no images that match the search criteria")
+	}
+
+	d.Set("name", results[0].Properties.Name)
```
stack72: should we check that Properties != nil ? Otherwise we could panic here

builtin/providers/profitbricks/data_source_location.go
```diff
@@ -0,0 +1,81 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"feature": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"id": {
```
stack72: We actually don't need to expose Id as a param - it's there by default

builtin/providers/profitbricks/data_source_datacenter_test.go
jasminSPC: This one is tricky in order to get a match we need to have a data center. I'm not sure how to accomplish that/

stack72: In the test, have 2 steps, the first as a setup to create the datacenter, then the second will include a datacenter and a lookup

have a look at this test as an example https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/data_source_aws_autoscaling_groups_test.go#L15

builtin/providers/profitbricks/data_source_image.go
jasminSPC: ```Properties ImageProperties            `json:"properties,omitempty"```
It can't be nil it is not a pointer.

stack72: 👍 sorry - still in AWS land where *everything* is a pointer :)

builtin/providers/profitbricks/data_source_datacenter_test.go
jasminSPC: I tried something similar: 

In first step I do this:

```
resource "profitbricks_datacenter" "foobar" {
	name       =  "datacenter-test"
	location = "us/las"
}
```
and then this.
```
resource "profitbricks_datacenter" "foobar" {
	name       =  "datacenter-test-updated"
	location = "us/las"
}

data "profitbricks_datacenter" "data_source"{
	name = "datacenter-test-updated"
	location= "us/las"
}
```

Each time I do this it simply doesn't get work. 
How do you want me to proceed?

jasminSPC:
The loadbalancer thing is a known issue with ProfitBricks and we can't do much about that. 
I can add cleanup in check destroy to wipe all orphaned resource.

stack72:
Hey

so i just went through this and it looks good - I ran the tests and got 1 error

```
% make testacc TEST=./builtin/providers/profitbricks TESTARGS='-run=TestAccDataSource'                                                   1 ↵ ✭
2017/02/02 10:39:44 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/profitbricks -v -run=TestAccDataSource -timeout 120m
=== RUN   TestAccDataSourceDatacenter_basic
--- PASS: TestAccDataSourceDatacenter_basic (0.64s)
=== RUN   TestAccDataSourceImage_basic
--- FAIL: TestAccDataSourceImage_basic (0.55s)
	testing.go:265: Step 0 error: Check failed: Check 2/3 error: data.profitbricks_image.img: Attribute 'name' expected "Ubuntu-16.04-LTS-server-2017-01-01", got "Ubuntu-16.04-LTS-server-2017-02-01"
=== RUN   TestAccDataSourceLocation_basic
--- PASS: TestAccDataSourceLocation_basic (0.61s)
FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/profitbricks	1.824s
make: *** [testacc] Error 1
```

I think we are going to have an issue with that test if it revs a version each month though so all in all this is looking good

P.

jasminSPC:
The problem with the test is that ProfitBricks is changing image names on monthly basis. 
I guess I should add custom check for this. 
@stack72 What do you think?

stack72:
This LGTM! :)

```
% make testacc TEST=./builtin/providers/profitbricks TESTARGS='-run=TestAccDataSource'                                                       ✭
2017/02/02 13:25:28 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/profitbricks -v -run=TestAccDataSource -timeout 120m
=== RUN   TestAccDataSourceDatacenter_matching
--- PASS: TestAccDataSourceDatacenter_matching (12.29s)
=== RUN   TestAccDataSourceImage_basic
--- PASS: TestAccDataSourceImage_basic (0.67s)
=== RUN   TestAccDataSourceLocation_basic
--- PASS: TestAccDataSourceLocation_basic (0.46s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/profitbricks	13.433s
```

PR #11529, opened 2017-01-30
Read vault token from ~/.vault-token if none supplied

### LINKED ISSUE #11365: vault provider should use `~/.vault-token` if `VAULT_TOKEN` isn't set
This allows simple operation of terraform if the user is already authenticated with vault.

eraserhd:
This closes #11365 (a feature request).

### DIFF
--- a/builtin/providers/vault/provider.go
+++ b/builtin/providers/vault/provider.go
@@ -129,7 +129,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		// Use the vault CLI's token, if present.
 		homePath, err := homedir.Dir()
 		if err != nil {
-			return nil, fmt.Errorf("No vault token found: %s", err)
+			return nil, fmt.Errorf("Can't find home directory when looking for ~/.vault-token: %s", err)
 		}
 		tokenBytes, err := ioutil.ReadFile(homePath + "/.vault-token")
 		if err != nil {
--- a/builtin/providers/vault/provider.go
+++ b/builtin/providers/vault/provider.go
@@ -2,7 +2,9 @@ package vault
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
+	"os"
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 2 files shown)


builtin/providers/vault/provider.go
```diff
@@ -122,6 +124,18 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, fmt.Errorf("failed to configure Vault API: %s", err)
 	}
 
+	token := d.Get("token").(string)
+	if token == "" {
+		// Use the vault CLI's token, if present.
+		tokenFile := fmt.Sprintf("%s/.vault-token", os.Getenv("HOME"))
```
apparentlymart: It looks like [vault's own implementation of this](https://github.com/hashicorp/vault/blob/f8c657a80a6b8e6dc5baa06febe0b67f00ced96e/command/token/helper_internal.go#L21) uses the `go-homedir` library, which tries various different means to locate the user's home directory, including some specialized support for Windows systems.

Do you think using that here would make sense, so we can ensure that the behavior will be consistent with Vault itself?

eraserhd: Haha, I noticed and pushed a patch apparently just as you were commenting.  This is done.

eraserhd:
I think the CI issue is not related.  This should be good to go.

builtin/providers/vault/provider.go
```diff
@@ -122,6 +124,21 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		return nil, fmt.Errorf("failed to configure Vault API: %s", err)
 	}
 
+	token := d.Get("token").(string)
+	if token == "" {
+		// Use the vault CLI's token, if present.
+		homePath, err := homedir.Dir()
+		if err != nil {
+			return nil, fmt.Errorf("No vault token found: %s", err)
```
cblecker: Wouldn't this error be called if the homedir was unable to be found? The error text might need to be clarified.

eraserhd: Clarified.

apparentlymart:
Thanks @eraserhd! This is merged in f1803ba4b6c5a5d6388d9977429f1637f6a251f5.

PR #11539, opened 2017-01-31
Recreate Rancher stack if it was removed

raphink:
Currently, a stack that is manually removed on Rancher is never seen as removed by Terraform, and thus not recreated. This patch forces Terraform to recreate a removed stack.


stack72:
LGTM! Thanks for the work - ignore the build failure here! We have an issue with one of the unit tests for remote state that isn't affected by this change

Paul

mcanevet:
@stack72 will/could this be backported to 0.8-stable? Or do we have to wait for 0.9?

stack72:
@mcanevet this will be in 0.8.6

mcanevet:
great! thanks

PR #11541, opened 2017-01-31
Fix StatusCake test timeout schema

ruippeixotog:
Although `timeout` in StatusCake tests is [defined correctly in the documentation](https://www.terraform.io/docs/providers/statuscake/r/test.html#timeout) as an optional value, in the resource it is defined as a computed value - most probably a copy-paste error.

The default value of 40 is the one shown in the StatusCake Web UI when a test is created manually.


stack72:
Thanks for this :) This LGTM!

PR #11543, opened 2017-01-31
Rancher: reset Id for unknown resources

raphink:
When switching from one Rancher server to another, we want Terraform to recreate Rancher resources. This currently leads to ugly `EOF` errors.    

This patch resets resource Ids when they can't be found in the Rancher API.


stack72:
LGTM! Thanks :)

PR #11546, opened 2017-01-31
provider/cloudstack: make the hash consistent again

svanharmelen:
The existing hash function for set items cannot generate consistent hashes when using both `Optional` and `Computed` on a schema field.

I tried to add this use case to the existing code base, but came to the conclusion this would be quite an endeavor.

That together with the fact this is the only field in all sets used in all builtin providers/resources that would be using both options at the same time, made me decide to change this single resource instead.


svanharmelen:
All is green and in addition the acceptance tests are also successfully executed locally. So moving ahead with this one...

Issue #11055, opened 2017-01-05
terraform import of rancher_stack does not import current docker_compose and rancher_compose


mcanevet:
### Terraform Version
v0.8.2

### Affected Resource(s)
- rancher_stack

### Terraform Configuration Files
```hcl
resource "rancher_stack" "foo" {
  name = "foo"
  environment_id = "${var.environment_id}"

  docker_compose =<<EOF
foo:
  image: foo/bar
EOF

  rancher_compose =<<EOF
EOF
foo:
  scale: 1
}
```

### Expected Behavior
Terraform import should import current docker_compose and rancher_compose.

### Actual Behavior
Terraform import does not import current docker_compose and rancher_compose.

johnrengelman:
So you created a stack in Rancher and you're trying to import it into TF?
How was the stack created? Manually or using a catalog entry?

mcanevet:
@johnrengelman exactly. I created the stack using a catalog entry.

johnrengelman:
The `docker_compose` and `rancher_compose` values will not be imported in this case. However, there will be values in `rendered_docker_compose` and `rendered_rancher_compose` that correspond to the config that you can see in the Rancher API.
The `docker_compose` and `rancher_compose` values are incompatible with using the `catalog` entry for a stack.
This is due to how the Rancher API stores this data.

mcanevet:
@johnrengelman actually I have the same issue without using a catalog entry

johnrengelman:
Please post the config and what terraform plan says.

stack72:
Closed via #11500

johnrengelman:
@stack72 - how does this refreneced PR affect this issue? Doesnt seem related.

stack72:
I'm going by the the OP put on the PR...

johnrengelman:
The "closed via" message has the wrong number (11500 vs 11550)

stack72:
d'oh! Sorry! Will fix now

PR #11550, opened 2017-01-31
rancher_stack: Retrieve docker_compose and rancher_compose in Read

raphink:
Without this, when importing an existing stack into Terraform,
the first Terraform run is non convergent as it needs to retrieve
the existing values.


mcanevet:
Fixes #11055

stack72:
LGTM! Thanks :)

johnrengelman:
@raphink did you test this with importing a stack that was created with a catalog entry? 
I ran into issues with that, so I think maybe setting the `docker_compose` and `rancher_compose` attributes only when the catalog entry is not set?

PR #11554, opened 2017-01-31
Remove race around local-exec Wait

jbardin:
There was still a race around the local-exec Command, where we were
calling Wait in 2 places which you can't do. There's no need for the
extra goroutine however, since we now have a context that we can pass to
the command itself.

### DIFF
--- a/builtin/provisioners/local-exec/resource_provisioner.go
+++ b/builtin/provisioners/local-exec/resource_provisioner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"runtime"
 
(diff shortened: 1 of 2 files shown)


jbardin:
hmm, my tests passed locally -- investigating....

jbardin:
Tests were passing locally on macOS. The problem looks like a Go issue on Linux: https://github.com/golang/go/issues/18874

jbardin:
It appears this is completely broken, and we can't Wait on canceled process on Linux right now. This was only working before by returning the wrong error and leaving the process running.

builtin/provisioners/local-exec/resource_provisioner.go
```diff
@@ -58,32 +59,30 @@ func applyFn(ctx context.Context) error {
...
 	// Setup the command
-	cmd := exec.Command(shell, flag, command)
+	cmd := exec.CommandContext(ctx, shell, flag, command)
 	output, _ := circbuf.NewBuffer(maxBufSize)
-	cmd.Stderr = io.MultiWriter(output, pw)
-	cmd.Stdout = io.MultiWriter(output, pw)
+
+	multiWriter := io.MultiWriter(output, pw)
+
+	// now we need an os.Pipe to pump the outout on Linux
```
mitchellh: I would add the link to the Go issue here so we know when we can remove it.

jbardin:
Turns out it's not a Linux problem, it just showed up on Linux as `/bin/sh` there is often symlinked to `/bin/dash`, and dash always starts a subprocess for commands whereas bash will exec single commands in the same process. I changed the test to make this fail the same way on all systems.

Having the process output to an os.Pipe prevents the exec package from needing to start goroutines to copy the data manually, and shifts the blocked Read into our own code. This lets us conditionally select on the copy goroutine finishing or the context expiring. 

@mitchellh: pushed a cleaner much fix to look at

Issue #8943, opened 2016-09-20
Support Google CloudSQL Second Generation instance type


evandbrown:
Second Generation instance types became generally available in August 16 (https://cloudplatform.googleblog.com/2016/08/Cloud-SQL-Second-Generation-performance-and-feature-deep-dive.html). Modify the Cloud SQL provider to support this instance type.

csainty:
I'd love to see this included.
In particular I want to be able to set...

```
dataDiskSizeGb: 100
dataDiskType: PD_SSD
storageAutoResize: true
failoverReplica:
    name: xxx
maintenanceWindow:
    day: 2
    hour: 10
```

Do you need external contributions around it or is there work underway to make these updates?

marcjoha:
Basic support for 2nd gen instances is already available. The key thing to notice is that 'tier' should specify the instance type if 2nd gen instances are created, and the pricing tier if 1st gen instances are created. This is mentioned here: https://cloud.google.com/sql/docs/admin-api/v1beta4/tiers.

Here's an example creating a 2nd gen instance. Also notice that the region parameter is different depending which generation is used.

resource "google_sql_database_instance" "master" {
  name = "master-instance"
  region = "us-central1"
  database_version = "MYSQL_5_7"
  settings = {
    tier = "db-n1-standard-1"
  }
}

stack72:
Closed via #11571

PR #11571, opened 2017-01-31
provider/google-cloud: Add second generation disk specification options

gh-mlfowler:
This is an enhancement to the `google_sql_database_instance` resource to add three optional settings for use with Second Generation instances:

- `disk_autoresize`: Toggles automatic resizing of the underlying disk
- `disk_size`: Allows specification of the initial disk size and can be used to increase disk size
- `disk_type`: Allows you to choose between the two disk types currently offered.

With the documentation I have indicated that these are Second Generation options by specifying this along with the Optional flag and default value.

I found that someone has raised an issue requesting these: https://github.com/hashicorp/terraform/issues/8943

### DIFF
--- a/builtin/providers/google/resource_sql_database_instance.go
+++ b/builtin/providers/google/resource_sql_database_instance.go
@@ -89,6 +89,18 @@ func resourceSqlDatabaseInstance() *schema.Resource {
 								},
 							},
 						},
+						"disk_autoresize": &schema.Schema{
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+						"disk_size": &schema.Schema{
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"disk_type": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+						},
 						"ip_configuration": &schema.Schema{
 							Type:     schema.TypeList,
 							Optional: true,
@@ -323,6 +335,18 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
 
 	if v, ok := _settings["crash_safe_replication"]; ok {
 		settings.CrashSafeReplicationEnabled = v.(bool)
+	}
+
+	if v, ok := _settings["disk_autoresize"]; ok && v.(bool) {
+		settings.StorageAutoResize = v.(bool)
+	}
+
+	if v, ok := _settings["disk_size"]; ok && v.(int) > 0 {
+		settings.DataDiskSizeGb = int64(v.(int))
+	}
+
+	if v, ok := _settings["disk_type"]; ok && len(v.(string)) > 0 {
+		settings.DataDiskType = v.(string)
 	}
 
 	if v, ok := _settings["database_flags"]; ok {
@@ -577,6 +601,24 @@ func resourceSqlDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
 
 	if v, ok := _settings["crash_safe_replication"]; ok && v != nil {
 		_settings["crash_safe_replication"] = settings.CrashSafeReplicationEnabled
+	}
+
+	if v, ok := _settings["disk_autoresize"]; ok && v != nil {
+		if v.(bool) {
+			_settings["disk_autoresize"] = settings.StorageAutoResize
+		}
+	}
+
+	if v, ok := _settings["disk_size"]; ok && v != nil {
+		if v.(int) > 0 && settings.DataDiskSizeGb < int64(v.(int)) {
+			_settings["disk_size"] = settings.DataDiskSizeGb
+		}
+	}
+
+	if v, ok := _settings["disk_type"]; ok && v != nil {
+		if len(v.(string)) > 0 {
+			_settings["disk_type"] = settings.DataDiskType
+		}
 	}
 
 	if v, ok := _settings["database_flags"]; ok && len(v.([]interface{})) > 0 {
@@ -840,6 +882,20 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 
 		if v, ok := _settings["crash_safe_replication"]; ok {
 			settings.CrashSafeReplicationEnabled = v.(bool)
+		}
+
+		if v, ok := _settings["disk_autoresize"]; ok && v.(bool) {
+			settings.StorageAutoResize = v.(bool)
+		}
+
+		if v, ok := _settings["disk_size"]; ok {
+			if v.(int) > 0 && int64(v.(int)) > instance.Settings.DataDiskSizeGb {
+				settings.DataDiskSizeGb = int64(v.(int))
+			}
+		}
+
+		if v, ok := _settings["disk_type"]; ok && len(v.(string)) > 0 {
+			settings.DataDiskType = v.(string)
 		}
 
 		_oldDatabaseFlags := make([]interface{}, 0)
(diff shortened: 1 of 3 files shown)


stack72:
Hi @gh-mlfowler 

I just looked at this - the code looks good but for some reason 1 of the tests fail

```
% make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccGoogleSqlDatabaseInstance_'                                             ✭
2017/01/31 21:53:06 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccGoogleSqlDatabaseInstance_ -timeout 120m
--- PASS: TestAccGoogleSqlDatabaseInstance_basic (52.55s)
--- PASS: TestAccGoogleSqlDatabaseInstance_basic2 (50.55s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_basic (30.45s)
--- FAIL: TestAccGoogleSqlDatabaseInstance_slave (839.66s)
	testing.go:265: Step 0 error: Check failed: Check 2/4 error: Error settings.disk_size mismatch, (10, 0)
=== RUN   TestAccGoogleSqlDatabaseInstance_diskspecs
--- PASS: TestAccGoogleSqlDatabaseInstance_diskspecs (411.84s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_upgrade (55.86s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_downgrade (45.16s)
=== RUN   TestAccGoogleSqlDatabaseInstance_authNets
--- PASS: TestAccGoogleSqlDatabaseInstance_authNets (61.70s)
FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/google	1547.777s
make: *** [testacc] Error 1
```

Can you take a look?

Thanks

Paul

gh-mlfowler:
@stack72 Apologies, I must not have run the full test suite. The issue is the test comparison itself. The scenario of the local specification being unspecified with the server having a value was catered for in the resource code but I missed it in the test code. Odd that it only appears for the slave, I guess the Google SDK treats them differently.

Regards,

Mike

stack72:
Thanks so much for the fast turnaround on the fix here - this LGTM! Tests are all green :)

```
% make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccGoogleSqlDatabaseInstance_'                                             ✭
2017/02/01 15:51:35 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccGoogleSqlDatabaseInstance_ -timeout 120m
--- PASS: TestAccGoogleSqlDatabaseInstance_basic (52.68s)
--- PASS: TestAccGoogleSqlDatabaseInstance_basic2 (51.42s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_basic (40.95s)
--- PASS: TestAccGoogleSqlDatabaseInstance_slave (893.66s)
=== RUN   TestAccGoogleSqlDatabaseInstance_diskspecs
--- PASS: TestAccGoogleSqlDatabaseInstance_diskspecs (400.75s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_upgrade (67.69s)
--- PASS: TestAccGoogleSqlDatabaseInstance_settings_downgrade (37.35s)
=== RUN   TestAccGoogleSqlDatabaseInstance_authNets
--- PASS: TestAccGoogleSqlDatabaseInstance_authNets (70.03s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	1614.546s
```

Paul

builtin/providers/google/resource_sql_database_instance.go
```diff
@@ -579,6 +603,24 @@ func resourceSqlDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
...
 	}
 
+	if v, ok := _settings["disk_autoresize"]; ok && v != nil {
+		if v.(bool) {
+			_settings["disk_autoresize"] = settings.StorageAutoResize
+		}
+	}
+
+	if v, ok := _settings["disk_size"]; ok && v != nil {
+		if v.(int) > 0 && settings.DataDiskSizeGb < int64(v.(int)) {
```
danawillow: I know the review for this has already been done, but can you explain this line to me? Why check that settings.DataDiskSizeGb < v?

gh-mlfowler: I came across the scenario where a database with auto-resize had increased it's size to a value higher than specified in the Terraform file. When you run the plan terraform wishes to reduce the size of the disk to match the specification which we don't want to do and the API will reject anyway. This check allows you to increase the size and ignore a server increased size.

danawillow: Wait, I'm confused. Can you tell me what's wrong with the way I'm reading this?

Say we start with a size of 20GB.

If auto resize increases it to 30, then we do an update for some other reason:
* the read would notice that the size increased, and update the state file
* the next update after that wouldn't do anything because the sizes would match

If auto resize decreases it to 10, then we do an update:
* the read wouldn't update the state file because what was read from the api was less than what's currently there
* the next update after that would try to reduce the size because the smaller one is what's in the state file

gh-mlfowler: My original version of this (which lives in https://github.com/BashtonLtd/terraform-gcebashtonsql) I didn't have these checks. After some of our instances resized I found the read was making a plan that wished to shrink the database. On a test database I found that if you applied this change the API would reject the request with a 400. This seems to match my understanding of the documentation that the auto resize feature will never shrink the database.

At the time I wrote this I didn't see any other way to prevent the apply. Is there a way to update the state file without necessitating a failing apply?

danawillow: > I found the read was making a plan that wished to shrink the database

Why was that happening? Does auto-resize not update the dataDiskSize field? If it does, wouldn't a read just update the state file to match what the server has?

gh-mlfowler: Yes, but the problem is the configuration file is demanding the lower value. Taking the following configuration:

```
provider "google" {
  region      = "us-central"
  project     = "redacted"
  credentials = "${file("redacted.json")}"
}

resource "google_sql_database_instance" "instance_master" {
        region = "us-central1"

        settings {
                tier = "db-f1-micro"
                disk_size = 10

                backup_configuration {
                        enabled = true
                        binary_log_enabled = true
                }
        }
}
```

After `apply` the state file will show `"settings.0.disk_size": "10",`. By using the GCP console I manually increase the size to 11 to simulate the auto-resize, without the check in place the next plan will produce the below:
```
~ google_sql_database_instance.instance_master
    settings.0.disk_size: "11" => "10"
```
So the read is getting the correct current value but the configuration is wishing to impose the originally specified 10.

Going on to the `apply` I get the 400 error I was expecting:
```
google_sql_database_instance.instance_master: Refreshing state... (ID: terraform-hex)
google_sql_database_instance.instance_master: Modifying...
  settings.0.disk_size: "11" => "10"
Error applying plan:

1 error(s) occurred:

* google_sql_database_instance.instance_master: Error, failed to update instance terraform-hex: googleapi: Error 400: Invalid instance property., invalidInstanceProperty
```

Side note, the error from the Google API is useless. We first hit this problem when adding a new user and a new schema, I burned a lot of time trying to find the problem with what we were adding before I noticed the shrinking disk.

So the check is in the read so that you don't generate a plan that includes a shrink and it's in the update so that a shrink isn't even attempted.

danawillow: Ah that makes more sense, thanks for the detailed explanation!

PR #11573, opened 2017-01-31
Fix quoting in README.md

spy-tech:
Fix one correctness issue, the rest is stylistic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ For local dev first make sure Go is properly installed, including setting up a [
 Next, using [Git](https://git-scm.com/), clone this repository into `$GOPATH/src/github.com/hashicorp/terraform`. All the necessary dependencies are either vendored or automatically installed, so you just need to type `make`. This will compile the code and then run the tests. If this exits with exit status 0, then everything is working!
 
 ```sh
-$ cd $GOPATH/src/github.com/hashicorp/terraform
+$ cd "$GOPATH"/src/github.com/hashicorp/terraform
 $ make
 ```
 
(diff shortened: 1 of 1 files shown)


mitchellh:
Hey there! Thanks. I tried our current quoting (or lack thereof) in sh, bash, and zsh and it worked fine in all three. Given those are three of the most major shells out there, I think our quoting is likely fine unless there is something I'm missing!

spy-tech:
@mitchellh Try adding a space to the GOPATH. 

```
GOPATH='/home/foo bar' cd $GOPATH
bash: cd: /home/foo: No such file or directory
```

mitchellh:
Ah, in that case we should probably just double quote the entire value. Will edit live.

spy-tech:
No

mitchellh:
Can you please explain further? "No" is not particularly helpful and I am interested in getting in any changes I am misunderstanding. Thank you!

spy-tech:
Quoting just the variable tells the human that a small part is interpreted using variable substitution rules. The way I wrote it is how it is supposed to be done.

mitchellh:
I see. Unless there are practical dangers of quoting the entire string here, I think that is more aesthetically appealing to me, correctness aside. Please correct me if I'm misunderstanding something, though. Even our "incorrect" path has worked without a single issue being raised for 3 years so I imagine we're having a mostly theoretical discussion here in any case.

spy-tech:
I think it's sloppy. Other people will also think that, but might not say it. To me it communicates that any shell script being written by the Terraform people is suspect. It undermines confidence on what is supposed to be a marketing page.

Feel free to disagree, but I won't change my opinion.

mitchellh:
Fair enough. Thanks for your input! I have no intention of changing your opinion, just wanted to hear it. 😄  In any case, the quoting now covers cases where there may be a space in your GOPATH. If there are edge cases still that aren't handled, I'd be happy to make any changes.

PR #11594, opened 2017-02-01
Rancher imports broken

raphink:
fa656429373d4bdfe3894d1f5fd85ff2bc225e4d breaks all imports, because  `environment_id` is not specified in imported resources.

I see two ways to fix that:

* either use a global client when `environment_id` is not set
* or import the `environment_id` from `account_id` on import, so it is set. This is probably the cleanest option.

/cc @johnrengelman

### DIFF
--- a/builtin/providers/rancher/resource_rancher_registration_token.go
+++ b/builtin/providers/rancher/resource_rancher_registration_token.go
@@ -16,7 +16,7 @@ func resourceRancherRegistrationToken() *schema.Resource {
 		Read:   resourceRancherRegistrationTokenRead,
 		Delete: resourceRancherRegistrationTokenDelete,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: resourceRancherRegistrationTokenImport,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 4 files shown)


raphink:
I just attached some code to fix this with option 2 in rancher_stack resources. If you agree with that approach, I'll apply the same patch to other resource types.

raphink:
OK then, I'll implement it for all resource types.

raphink:
There, I updated the commit with all resource types. Does that look good to you @johnrengelman ?

mcanevet:
LGTM

stack72:
are we all happy for this to be merged? :)

raphink:
@johnrengelman I don't see a way to import without a global key, since you can't specify the environment in the import command, so I don't think this can be fixed in the provider.

mcanevet:
At least this patch fixes the regression introduced by https://github.com/hashicorp/terraform/commit/fa656429373d4bdfe3894d1f5fd85ff2bc225e4d even if https://github.com/hashicorp/terraform/issues/11053 is not fully resolved.
I'd say +1 to merge this.

johnrengelman:
@stack72 is there a way to provide multiple arguments to an import command?

stack72:
@johnrengelman all you can do is to provide a custom argument and inside the import func, you could parse it into pieces

We do this in aws_route_53_record as follows:

```
parts := strings.Split(d.Id(), "_")

		d.Set("zone_id", parts[0])
		d.Set("name", parts[1])
		d.Set("type", parts[2])
		if len(parts) > 3 {
			d.Set("set_identifier", parts[3])
		}
```

johnrengelman:
@raphink what do you thinking of that? We could keep the existing method but also allow some to provide something like `<stack_id>:<environment_id>` if they need to?

raphink:
I'd be fine with that, although I'd prefer `<environment_id>/<stack_id>` as the format.

So if `environment_id` is provided, we keep it, otherwise, we get the `account_id` as the default value in import, right?

johnrengelman:
Yeah, I like that format to..consistent with the Rancher api.

Yeah, if `environment_id` is provided, then use it to create the client. Otherwise, default to the global client and try to import.

mcanevet:
Could this PR still be merged, as it fixes a regression even if it doesn't solve completely the issue?
It would be great to have 0.8.6 released with this fix.

raphink:
@johnrengelman @stack72 I suggest we merge this fix and implement @johnrengelman 's proposed syntax later.

johnrengelman:
Go for it. I'll open a PR today for the new syntax.

stack72:
Thanks for all this work so far folks!

PR #11604, opened 2017-02-01
vSphere virtual machine don't ignore VM power on errors

darrenswhite:
This prevents Terraform from hanging indefinitely while waiting for the VM to power on.


stack72:
Thanks so much for this - LGTM!

PR #11607, opened 2017-02-01
Closes #11567. Handled Rancher deleted resources.

### LINKED ISSUE #11567: Rancher provider does not removed resources that are deleted out of band.
The Rancher API never actually deletes resources. It marks their state as 'removed' and then 'purged', but querying for them by ID will still return the payload.

This causes resources create by Terraform but then removed via the Rancher API to not be re-created.

All the `read` functions from Rancher resources should not consider resources in a `removed` or `purged` state to be valid resources.

/cc @raphink

johnrengelman:
/cc @raphink 

I wasn't sure how to add a test for this, since I need TF to create a resource, then call the Rancher API directly to delete it, then run TF again and see that the ID changes.

@stack72 if you can point me to a spec that does something like this, I'd be happy to add it.

### DIFF
--- a/builtin/providers/rancher/resource_rancher_environment_test.go
+++ b/builtin/providers/rancher/resource_rancher_environment_test.go
@@ -3,6 +3,7 @@ package rancher
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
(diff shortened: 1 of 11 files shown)


stack72:
@johnrengelman :) https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_efs_mount_target_test.go#L68

johnrengelman:
Thank you sir. Will update tests and update this PR shortly.

johnrengelman:
Updated.

builtin/providers/rancher/resource_rancher_environment_test.go
```diff
@@ -39,6 +40,50 @@ func TestAccRancherEnvironment(t *testing.T) {
 	})
 }
 
+func TestAccRancherEnvironment_diappears(t *testing.T) {
```
stack72: _disappears not _diappers ;)

johnrengelman: fixed and pushed.

stack72:
This LGTM!

PR #11608, opened 2017-02-01
Update index.html.markdown

luisamador:
replaced 'supercedes' for 'supersedes'.

https://en.oxforddictionaries.com/definition/supersede


stack72:
Thanks so much for this :) LGTM!

PR #11620, opened 2017-02-01
provider/pagerduty: pagerduty_addon resource

heimweh:
This PR adds support for `add-ons` in PagerDuty. 

With add-ons, third-party developers can write their own add-ons to PagerDuty's UI. Given a configuration containing a src parameter, that URL will be embedded in an iframe on a page that's available to users from a drop-down menu.

Example usage:

```bash
resource "pagerduty_addon" "example" {
  name = "Internal Status Page"
  src  = "https://intranet.example.com/status"
}
```


stack72:
LGTM!

Issue #9474, opened 2016-10-20
IAM instance profile attachment race condition


tomelliff:
It looks like there's yet another eventual consistency issue with IAM roles being created and the API returning before the role is fully created leading to errors being thrown by the AWS API when an instance attempts to use it.

https://github.com/hashicorp/terraform/pull/7324 and https://github.com/hashicorp/terraform/pull/7938 each fixed one set of issues we were seeing when creating the IAM role and profile in the same folder as the instance that used it.

We're now seeing the following error:

```
04:12:49 * aws_instance.scope.2: Error launching source instance: InvalidParameterValue: Value (test01-tp-pipe-selenium-node) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
04:12:49    status code: 400, request id: f4b4e246-f1c2-49ee-bf03-782a85a28260
04:12:49 * aws_instance.scope.6: Error launching source instance: InvalidParameterValue: Value (test01-tp-pipe-selenium-node) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
04:12:49    status code: 400, request id: 73afb476-3ebd-4208-a9a3-def9f3152cd9
04:12:49 * aws_instance.scope.1: Error launching source instance: InvalidParameterValue: Value (test01-tp-pipe-selenium-node) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
04:12:49    status code: 400, request id: 1f512f73-6493-44d4-bad0-202402bb4130
04:12:49 * aws_instance.scope.4: Error launching source instance: InvalidParameterValue: Value (test01-tp-pipe-selenium-node) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
04:12:49    status code: 400, request id: 7774f7b9-7818-454b-b7e4-5e0d521b6a58
```

which follows these creating/creation complete logs:

```
04:12:28 [0m[1mmodule.selenium_node_instance.aws_instance.scope.2: Still creating... (10s elapsed)[21m[0m
04:12:28 [0m[1mmodule.selenium_node_instance.aws_instance.scope.6: Still creating... (10s elapsed)[21m[0m
04:12:28 [0m[1mmodule.selenium_node_instance.aws_instance.scope.1: Still creating... (10s elapsed)[21m[0m
04:12:28 [0m[1mmodule.selenium_node_instance.aws_instance.scope.3: Still creating... (10s elapsed)[21m[0m
[...]

PR #11634, opened 2017-02-02
[WIP] provider/aws: 60s instance profile creation retry

snakeb1t:
fixes https://github.com/hashicorp/terraform/issues/9474

### DIFF
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -368,7 +368,7 @@ func resourceAwsInstanceCreate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Run configuration: %s", runOpts)
 
 	var runResp *ec2.Reservation
-	err = resource.Retry(15*time.Second, func() *resource.RetryError {
+	err = resource.Retry(60*time.Second, func() *resource.RetryError {
 		var err error
 		runResp, err = conn.RunInstances(runOpts)
 		// IAM instance profiles can take ~10 seconds to propagate in AWS:


catsby:
I think that if we need to ensure an instance profile is created, we should do so in the instance profile creation

-  https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_iam_instance_profile.go#L98

I think line #123 would be a perfect place to use the AWS SDK's IAM waiter:

- https://github.com/aws/aws-sdk-go/blob/master/service/iam/waiters.go#L13

What do you think?

snakeb1t:
@catsby : whoa, that's really neat! I'll give it a try. Though I wonder if there's a race condition here: suppose we keep trying the GetInstanceProfile until it succeeds, but then terraform moves on to creating a resource that relied on the instance profile resource (say, an EC2 instance). But when terraform tries to make that resource, the EC2 service's equivalent of GetInstanceProfile still returns nothing when it tries to make the resource, so the resource fails.

Even though there might not be any guarantees, using the Waiter would still be way better (and a general fix) than what we have now. KMS keys can't be created if any principal listed in its policy doesn't exist (from the perspective of the KMS service), so my original PR wouldn't do anything about that.

Thanks so much; I'll push something up soon.

catsby:
> but then terraform moves on to creating a resource that relied on the instance profile resource (say, an EC2 instance). 

If we're waiting in the `instance_profile` then anything that depends on it should be themselves waiting. Assuming they are related correctly with interpolation in the config files, Terraform ensures that A comes before B. 

Unfortunately IAM does sometime, seemingly, tell us directly that resource A is ready, but when trying to create resource B, resource B's service ask's IAM for A and IAM says it 'doesn't know what we're talking about 😄

catsby:
I'm going to close this for now as we discussed I'm not sure this is the right fix. Please reopen (ping me if you can't) or open a new PR if/when the time comes. Thanks!

PR #11641, opened 2017-02-02
provider/aws: Better handle parallel tests spot-fleet-req

grubernaut:
Better handle parallel testing of the `aws_spot_fleet_request` resource. Previously we were catching a lot of errors due to naming the policy_attachment resource the same between tests.


stack72:
LGTM!

PR #11642, opened 2017-02-02
provider/aws: Improve IAM Group AccTests

grubernaut:
Improve the iam_group acceptance tests. Previously, the iam_group would overlap when running our tests in parallel. This fixes that.


stack72:
LGTM

PR #11631, opened 2017-02-02
provider/google: make local_traffic_selector computed now that we rea…

drebes:
As part of the acceptance tests for #11020 the traffic selectors are now read back from the server. If local_traffic_selector is not provided during tunnel creation, Cloud VPN still returns it as "0.0.0.0/0". If the Computed attribute is not set, the tunnel will get recreated if the user does not set local_traffic_selector explicitly, until its value is set to that value.

Setting it as computed skips the recreation.

PR #11650, opened 2017-02-03
provider/google: set additional_zones to computed and disallow the original zone from appearing in the list

danawillow:
This fixes a bug introduced by #11018.

`cluster.Locations` will never be empty on read (because it will contain at least `cluster.Zone`), so even if `additional_zones` wasn't set in the config, it was getting set on read, causing a subsequent `terraform plan` to be non-empty.

### DIFF
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -444,7 +444,10 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 
 	d.Set("name", cluster.Name)
 	d.Set("zone", cluster.Zone)
-	d.Set("additional_zones", cluster.Locations)
+
+	if _, ok := d.GetOk("additional_zones"); ok {
+		d.Set("additional_zones", cluster.Locations)
+	}
 	d.Set("endpoint", cluster.Endpoint)
 
 	masterAuth := []map[string]interface{}{
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -442,15 +442,15 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("name", cluster.Name)
 	d.Set("zone", cluster.Zone)
 
+	locations := []string{}
 	if len(cluster.Locations) > 1 {
-		locations := []string{}
 		for _, location := range cluster.Locations {
 			if location != cluster.Zone {
 				locations = append(locations, location)
 			}
 		}
-		d.Set("additional_zones", locations)
 	}
+	d.Set("additional_zones", locations)
 
 	d.Set("endpoint", cluster.Endpoint)
 
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -95,6 +95,7 @@ func resourceContainerCluster() *schema.Resource {
 			"additional_zones": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
 				ForceNew: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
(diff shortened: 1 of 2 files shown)


builtin/providers/google/resource_container_cluster.go
```diff
@@ -444,7 +444,10 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 
 	d.Set("name", cluster.Name)
 	d.Set("zone", cluster.Zone)
-	d.Set("additional_zones", cluster.Locations)
+
+	if _, ok := d.GetOk("additional_zones"); ok {
+		d.Set("additional_zones", cluster.Locations)
```
drebes: Alternatively, you can set "additional_zones" to be Computed. See #11631 for the same problem there.

danawillow: Done

danawillow:
I realized that there's no good way of being able to allow the original zone in the `additional_locations` list, so I went ahead and disallowed it, since it's called additional_locations anyway.

Why this was necessary: if I don't include the original zone, then when I get back `locations` from the API it will include it and update the state file. The next update then will show the addition of the original zone as a change. If I proactively remove the original zone from the `locations` list on read, then the opposite issue occurs if I did have the original zone in the list.

paddyforan:
As a general note, I'm in support of getting away from the practice of only setting state if it's set in config. That's causing headaches and is kind of a special-case thing, in my opinion. So glad to see the change to computed. I have some code-level questions, but I'm going to read the code a bit more closely before I pester you with silly questions. Queuing this up for post-lunch. :)

paddyforan:
OK, after reading through all this, all my questions were, in fact, silly. So the last big question I have is it seems the only test changes were to parameterize an existing test without changing the logic of it. Did the test already fail for this? Do we have a test case that shows this bug as failing yet? If not, is it possible to add one?

danawillow:
Yup, the test is already failing. From TeamCity:
```
TestAccContainerCluster_withAdditionalZones 	
 	Current failure:	refs/heads/master	#244992c 	Changes (33) 	07 Feb 17 06:13
	First failure:		#998155f 	Changes (41) 	19 Jan 17 05:09
 
------- Stdout: -------
=== RUN   TestAccContainerCluster_withAdditionalZones
--- FAIL: TestAccContainerCluster_withAdditionalZones (311.81s)
    testing.go:265: Step 0 error: Check failed: Check 2/2 error: number of additional zones did not match 2
FAIL
```

All the other ContainerCluster tests are failing too (which prompted this fix):
```
TestAccContainerCluster_basic 	
 	Current failure:	refs/heads/master	#244992c 	Changes (33) 	07 Feb 17 06:13
	First failure:		#cb8d7bf 	Changes (27) 	12 Jan 17 05:09
 
------- Stdout: -------
=== RUN   TestAccContainerCluster_basic
--- FAIL: TestAccContainerCluster_basic (301.73s)
        
        DIFF:
        
        DESTROY/CREATE: google_container_cluster.primary
          additional_zones.#:                   "1" => "0" (forces new resource)
          additional_zones.0:                   "*******-a" => "" (forces new resource)
[etc.]
```

The test change was because I thought I was going to have to make a change but then didn't. Happy to take it out of the PR if you don't think it belongs here.

paddyforan:
I think the test change is totally great :) I just wanted to be sure we had a failing test, to test regressions and such. I don't know if you can tell, but I'm trying to make sure every code change has a test associated with it. :)

builtin/providers/google/resource_container_cluster.go
```diff
@@ -444,7 +441,17 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 
 	d.Set("name", cluster.Name)
 	d.Set("zone", cluster.Zone)
-	d.Set("additional_zones", cluster.Locations)
+
+	if len(cluster.Locations) > 1 {
```
paddyforan: It probably doesn't matter, but I'd love to set this in state, even if it's getting set to an empty value. I think, as a general rule of thumb, things should get set in state unless there's a core bug that makes things not work, we can articulate why diffing on it is undesirable, or we can articulate a strong security concern for storing it in state.

danawillow: Done.

paddyforan:
Looks great, merge at will! Thanks for tackling those test failures, @danawillow. You're awesome.

PR #11653, opened 2017-02-03
providers/google: google_project supports billing account

evandbrown:
This change allows a Terraform user to set and update the billing account associated with their project.

### DIFF
--- a/builtin/providers/google/resource_google_project.go
+++ b/builtin/providers/google/resource_google_project.go
@@ -183,13 +183,13 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 		ba := cloudbilling.ProjectBillingInfo{
 			BillingAccountName: "billingAccounts/" + v.(string),
 		}
-		_, err = config.clientBilling.Projects.UpdateBillingInfo("projects/"+pid, &ba).Do()
+		_, err = config.clientBilling.Projects.UpdateBillingInfo(prefixedProject(pid), &ba).Do()
 		if err != nil {
 			d.Set("billing_account", "")
 			if _err, ok := err.(*googleapi.Error); ok {
-				return fmt.Errorf("Error setting billing account %q for project %q: %v", ba.Name, "projects/"+pid, _err)
+				return fmt.Errorf("Error setting billing account %q for project %q: %v", ba.BillingAccountName, prefixedProject(pid), _err)
 			}
-			return fmt.Errorf("Error setting billing account %q for project %q: %v", ba.Name, "projects/"+pid, err)
+			return fmt.Errorf("Error setting billing account %q for project %q: %v", ba.BillingAccountName, prefixedProject(pid), err)
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


evandbrown:
This one is really tricky to write an integration test for. We'd need to create two dummy Billing Accounts in the testing Organization, and give the service account that runs the integration test the proper permission in the Organization.  Then introduce the two Billing Account IDs as env vars to the tests. Not impossible, but really raises the bar for someone running the integration tests locally.

For now, I'm testing manually to verify create and update of a project with a billing account. Thoughts, @danawillow and @paddyforan?

Thanks!

builtin/providers/google/resource_google_project.go
```diff
@@ -172,6 +178,21 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	// Set the billing account
+	if v, ok := d.GetOk("billing_account"); ok {
+		ba := cloudbilling.ProjectBillingInfo{
+			BillingAccountName: "billingAccounts/" + v.(string),
+		}
+		_, err = config.clientBilling.Projects.UpdateBillingInfo("projects/"+pid, &ba).Do()
```
danawillow: nit: even though it's a teeny tiny thing, since you do "projects/" + pid a few times I wouldn't mind seeing that in a function

builtin/providers/google/resource_google_project.go
```diff
@@ -210,6 +231,18 @@ func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("policy_etag", pol.Etag)
 	d.Set("policy_data", string(polBytes))
 
+	// Read the billing account
+	ba, err := config.clientBilling.Projects.GetBillingInfo("projects/" + pid).Do()
+	if err != nil {
+		return fmt.Errorf("Error reading billing account for project %q: %v", "projects/"+pid, err)
+	}
+	if ba.BillingAccountName != "" {
+		_ba := strings.Split(ba.BillingAccountName, "billingAccounts/")
```
danawillow: TrimPrefix is probably cleaner than split

paddyforan:
Have a few thoughts on this, let me circle around tomorrow. :)

builtin/providers/google/resource_google_project.go
```diff
@@ -172,6 +178,21 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
...
 
+	// Set the billing account
+	if v, ok := d.GetOk("billing_account"); ok {
+		ba := cloudbilling.ProjectBillingInfo{
+			BillingAccountName: "billingAccounts/" + v.(string),
+		}
+		_, err = config.clientBilling.Projects.UpdateBillingInfo("projects/"+pid, &ba).Do()
+		if err != nil {
+			d.Set("billing_account", "")
+			if _err, ok := err.(*googleapi.Error); ok {
```
paddyforan: Is there a difference between how `*googleapi.Error` and `error` get printed when using `%v`?

builtin/providers/google/resource_google_project.go
paddyforan: The way this is written, I don't actually know what the string is expected to be... it could be any one of these:

* projects/{PROJECT}/billingAccounts/{account name}
* billingAccounts/{account name}
* https://www.googlespis.com/projects/{PROJECT}/billingAccounts/{account name}

Honestly, would love some clarification (esp as a code comment) around what `ba.BillingAccountName` is expected to be at this point. And agree `strings.Split` may not be the most semantic function to use.

builtin/providers/google/resource_google_project.go
```diff
@@ -210,6 +231,18 @@ func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("policy_etag", pol.Etag)
 	d.Set("policy_data", string(polBytes))
 
+	// Read the billing account
+	ba, err := config.clientBilling.Projects.GetBillingInfo("projects/" + pid).Do()
```
sethvargo: It looks like it's possible to create projects without a billing account (now), and this line of code is causing any projects which do not have a billing account to return a 403 on read.

builtin/providers/google/resource_google_project.go
evandbrown: There is. The former will unpack and print any nested errors.

builtin/providers/google/resource_google_project.go
evandbrown: Not a nit at all! Done.

builtin/providers/google/resource_google_project.go
evandbrown: TIL `TrimPrefix`. Done.

builtin/providers/google/resource_google_project.go
evandbrown: Hmm, I'm not seeing this. I've created a few projects (and just added tests that do the same) that don't have billing accounts but make it through read. GetBillingInfo should return an empty billing account if none is set. Possible to see your config?

evandbrown:
@paddyforan totally reasonable; tests are added.

builtin/providers/google/resource_google_project.go
sethvargo: @evandbrown it's sadder than that 😦 . Steps to repro:

1. Compile Terraform without this PR
2. Create project
3. Compile Terraform with this PR
4. Change a project attr
5. Get error

builtin/providers/google/resource_google_project.go
paddyforan: Fair enough!

builtin/providers/google/resource_google_project_iam_policy_test.go
```diff
@@ -624,3 +624,13 @@ resource "google_project" "acceptance" {
 	org_id = "%s"
 }`, pid, name, org)
 }
+
+func testAccGoogleProject_createBilling(pid, name, org, billing string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+    project_id = "%s"
```
sethvargo: Looks like tabs vs spaces happened here

paddyforan:
Thanks for adding those tests, @evandbrown! This looks great. Looks like @sethvargo had a nit about tabs/spaces, don't know if you want to update that or not. Either is fine by me. But this has a 👍 from me.

PR #11658, opened 2017-02-03
rancher: parse Rancher client cli.json config file

raphink:
This doesn't really work yet, because the URL is required in the provider parameters. I see several options, not sure which one is the best:

* make the URL optional, fail in providerConfigure if it's not set
* use our own method instead of `schema.EnvDefaultFunc` to return default values, but then that means we can't provide `config` as a parameter as it needs to be known beforehand.

@stack72 @johnrengelman any idea?

### DIFF
--- a/builtin/providers/rancher/provider.go
+++ b/builtin/providers/rancher/provider.go
@@ -1,9 +1,21 @@
 package rancher
 
 import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
+
+type CLIConfig struct {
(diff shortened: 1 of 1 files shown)


mcanevet:
I think making the URL optional is a good idea

mcanevet:
The idea is to pass a configuration file created by "rancher config" that we could pass at the instanciation of the provider like that:

```hcl
provider "rancher" {
  config = "${pathexpand(~/.rancher/config.json)}"
}
```

raphink:
@stack72 @mcanevet fixed my code, so it works now, using option 1

mcanevet:
@stack72 could you please review this?

stack72:
LGTM! Thanks for this @raphink :)

PR #11660, opened 2017-02-03
terraform: remove ConnectDependents and related interfaces

mitchellh:
This is just more code removal that I missed when removing the old legacy graph.

We no longer use the dependency map and `ConnectDependents` function on the Graph itself. This allows us to remove a lot of internal accounting as well as some interfaces. Pretty great!


mitchellh:
Merging this one since the code removed was never called by any non-test code.

Issue #10877, opened 2016-12-21
Provisioners fail when using IPv6 address with WinRM connections


mgoldsmi:
When using the file provisioner with a WinRM connection where the host attribute is set to an IPv6 address, the provisioner fails as follows:

Error applying plan:
1 error(s) occurred:
* Upload failed: Error uploading file to $env:TEMP\winrmcp-585a3f0c-cc63-9788-f17b-59b253c19e8b.tmp: Couldn't create shell: unknown error Post http://2402:9900:111:1485:7918:dde1:7866:3770:5985/wsman: dial tcp: too many colons in address 2402:9900:111:1485:7918:dde1:7866:3770:5985

And if you change to a remote-exec provisioner, you get a slightly different output (however the same root cause is evident in the log file):

[deploy@cxs-de-mgmt01 winrm]$ terraform apply
null_resource.windows-ipv6-host: Creating...
null_resource.windows-ipv6-host: Provisioning with 'remote-exec'...
null_resource.windows-ipv6-host (remote-exec): Connecting to remote host via WinRM...
null_resource.windows-ipv6-host (remote-exec):   Host: [2402:9900:111:1485:7918:dde1:7866:3770]
null_resource.windows-ipv6-host (remote-exec):   Port: 5985
null_resource.windows-ipv6-host (remote-exec):   User: Administrator
null_resource.windows-ipv6-host (remote-exec):   Password: true
null_resource.windows-ipv6-host (remote-exec):   HTTPS: false
null_resource.windows-ipv6-host (remote-exec):   Insecure: false
null_resource.windows-ipv6-host (remote-exec):   CACert: false
null_resource.windows-ipv6-host (remote-exec): Connected!
null_resource.windows-ipv6-host: Still creating... (10s elapsed)
^CInterrupt received. Gracefully shutting down...
^CTwo interrupts received. Exiting immediately. Note that data
loss may have occurred.

### Terraform Version
Terraform v0.8.2-dev

### Affected Resource(s)
Please list the resources as a list, for example:
- file provisioner
- remote-execute provisioner

### Terraform Configuration Files
Sample file for file provisioner:

resource "null_resource" "windows-ipv6-host" {

  connection {
    host = "2402:9900:111:1485:7918:dde1:7866:3770"
    user = "Administrator"
    password = "*******"
    type = "winrm"
  }

  provisioner "file" {
      content = "Simple string to write to host"
      destination = "AppData\\Local\\Terraform\\Setup-Windows.ps1"
[...]

PR #11661, opened 2017-02-03
Proposed fix for issue #10877 - Provisioners fail when using IPv6 address with WinRM connections

mgoldsmi:
Packages masterzen/winrm and packer-community/winrmcp have been updated with govendor fetch to bring IPv6 fix from winrmcp into build. Changes to communicator.go required to resolve minor changes in msaterzen/winrm.

Some notes:
1. There was a directory change in the masterzen/winrm package (files were moved one directory down) which is the reason for the removal of the masterzen/winrm/winrm sub-directory.
2. masterzen/winrm now includes a new transport for NTLM authentication support (ntlm.go) which is based on the github.com/Azure/go-ntlmssp and github.com/masterzen/azure-sdk-for-go packages. These were fetch by the govendor fetch used to update winrmcp and I have left these in this pull request ... but it makes this pull request a far bigger change than just a fix for IPv6 support in winrmcp.

Finally - I'd appreciate a review of my treatment of the endpoint.CACert slice in communicator/winrm/communicator.go. The winrm package changed this from a pointer []byte, to a plain []byte. My code change simply sets the slice reference but does not copy slice contents. Hope this is correct


teamterraform:
Hi @mgoldsmi ! 
We are sorry that we let this PR sit for so long. Since the underlying code base has changed so much, and there are many conflicts, we are going to close this PR. 

If you are still interested in working on this, please do feel free to open a new PR with conflicts resolved.

Issue #11448, opened 2017-01-26
aws_db_instance: skip_final_snapshot should be false by default


aerickson:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.8.4

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_db_instance

### Terraform Configuration Files
```hcl
resource "aws_db_instance" "db" {
  // skip_final_snapshot = true by default
}
```

### Expected Behavior
`skip_final_snapshot = false` by default

The AWS Console's default is to not skip the final snapshot. I think TF should follow AWS defaults (I guess I agree with security deviations from AWS like with SG egress rules).

### Actual Behavior
`skip_final_snapshot = true`

### Steps to Reproduce
It's part of the documented and tested behavior.

### Notes

This caused me data loss. I assume that it will for many others also.

stack72:
Hi @aerickson 

Thanks for coming back to this - I agree with you on this. I think this is something we should do. I don't think we should rush this into a point release as this is backwards incompatible and may actually mean our users incur an extra cost going forward for their resources

I hope you can wait a little while longer for this - please note though, it *will* definitely be part of 0.9 and i will be working on it this week

Thanks

Paul

stack72:
Closed via #11668

PR #11668, opened 2017-02-03
provider/aws: Set aws_db_cluster to snapshot by default

stack72:
As requested in #11448

Terraform didn't snapshot AWS DB Instances by default. We are going to
change that behaviour going forward.

### DIFF
--- a/builtin/providers/aws/resource_aws_db_instance_test.go
+++ b/builtin/providers/aws/resource_aws_db_instance_test.go
@@ -150,15 +150,13 @@ func TestAccAWSDBInstanceReplica(t *testing.T) {
 	})
 }
 
-func TestAccAWSDBInstanceSnapshot(t *testing.T) {
+func TestAccAWSDBInstanceNoSnapshot(t *testing.T) {
 	var snap rds.DBInstance
 
 	resource.Test(t, resource.TestCase{
-		PreCheck:  func() { testAccPreCheck(t) },
-		Providers: testAccProviders,
-		// testAccCheckAWSDBInstanceSnapshot verifies a database snapshot is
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_db_instance_test.go
```diff
@@ -711,7 +717,7 @@ provider "aws" {
   region = "us-east-1"
 }
 resource "aws_db_instance" "snapshot" {
-	identifier = "tf-snapshot-%d"
+	identifier = "tf-test-%s
```
grubernaut: `acctest.RandInt()` should return an `int`, suggest changing to `%d` here

tomelliff:
@stack72 Looking at this change am I right in thinking that by default it still won't snapshot unless `final_snapshot_identifier` is specified? If so is that intentional?

Looking at the original issue I think the person who raised it was expecting a snapshot to be created without specifying any extra config.

Is it possible to define something as mandatory if something else is set? Or should it be defaulting a snapshot name to something sensible?

stack72:
hi @tomelliff 

We follow the AWS standard here - they require a final-snapshot-name and the cli / api requests fail without them - IMO, we would have to discuss the addition of default snapshot names - what would the format be? would we timestamp? version? etc

I understand the need for extra config - but with the addition of a single param, it will work as expected

Hope this helps?

Paul

tomelliff:
Fair enough and thanks for the reply. I guess it's still an improvement because then you only need to set the `final_snapshot_identifier` to something suitable to enable snapshots but it would still allow you to cleanly disable snapshots by setting `skip_final_snapshot` to `true`.

PR #11669, opened 2017-02-03
Adding in root level .gitattributes file.

liamjbennett:
This is to pin the .go files to LF line endings.

Developing on windows and using the vagrant environment to build,
running `make fmt` will cause all the *.go files to be touched,
alterting their line endings and causing git to see them as changed.


mitchellh:
This seems reasonable to me, especially because `go fmt` forces it anyways.

Issue #6016, opened 2016-04-05
Github Provider: Support API Rate Limiting


johnrengelman:
The Github API has a rate limit than can be easily overcome when managing an organization.
This ticket is simply a spot to start a conversation about how to properly handle the rate limit.

Details: https://developer.github.com/v3/#rate-limiting

JacobASeverson:
If I'm reading this right we are authenticating with OAuth so the limit is 5,000/hour? Or is that not the case?

apparentlymart:
In other providers the approach has been for Terraform to politely retry things for a while and then fail if they can't be dealt with in a reasonable amount of time (a few minutes, I think.)

This window-based rate limiting strategy doesn't really play well with our usual approach: if the limit is hit at 15 minutes after the hour, it's not reasonable for Terraform to pause for 45 minutes before completing the request. Instead, it'd be better for it to fail with an error message telling the user when to retry.

If the limit is indeed 5,000/hour then it's probably reasonable as a starting point to just handle this error response gracefully by giving a user-friendly error message. We could potentially have a special case for waiting if the reset time is less than 2min in the future, but that seems like it'd be rare enough to not be worth the extra complexity.

[Elsewhere in the docs](https://developer.github.com/v3/#conditional-requests) it's suggested that clients can use conditional GET requests and they won't count against the rate limit if `304 Not Modified` is returned; we could potentially store the resource etag as a computed resource attribute and use `If-None-Match` in our refresh requests, thus reducing Terraform's impact on hitting the rate limit. Unfortunately it looks like [the Github client library Terraform is using delegates conditional request handling to the `http.Transport` layer](https://godoc.org/github.com/google/go-github/github#hdr-Conditional_Requests), which means it'd probably take some significant hoop-jumping to use the Terraform state as the "cache" for it to read.

JacobASeverson:
Thanks for the elaboration @apparentlymart. Recognizing a rate limit error and providing a more friendly message does indeed sound like a good starting point. I'll start looking at that as well as options for conditional GETs.

PR #11670, opened 2017-02-03
Add some support for Github conditional requests

nicolasvan:
This is an attempt at fixing #6016

This is done by storing the `ETag` or `Last-Modified` headers from github's API responses into the terraform state.
Next time the resource is refreshed, add a `If-None-Match` or `If-Modified-Since` header to the request.
Github then returns a HTTP 304 if the resource has not changed since last terraform refresh. (which does not count against our rate limits, as per https://developer.github.com/v3/#conditional-requests)

I had to change the provider config so that a new `github.Client` is created for each resource, so that we can pass in a separate `conditionalTransport` object for each resource (`conditionalTransport` is used to add the headers to the request)

I have absolutely no experience with Go and there might be better ways to do it?

### DIFF
--- a/builtin/providers/github/config.go
+++ b/builtin/providers/github/config.go
@@ -1,6 +1,7 @@
 package github
 
 import (
+	"net/http"
 	"net/url"
 
 	"github.com/google/go-github/github"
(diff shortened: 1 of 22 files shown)


stack72:
Hi @nicolasvan 

Please can you rebase this to fix the merge conflicts?

Thanks

Paul

nicolasvan:
Not relevant to this repo anymore.

PR #11672, opened 2017-02-03
Update storage_bucket_acl.html.markdown

mbfrahry:
Removing `default_acl` since Terraform doesn't manage it anymore


danawillow:
Hey @mbfrahry, I know it was a while ago that you submitted this so you might not remember, but @rileykarson just pointed out to me that Terraform is still managing `default_acl`: https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_storage_bucket_acl.go#L27. Were you intending to deprecate it or was this PR meant to remove something else?

Issue #11663, opened 2017-02-03
provider/aws: Set undefined value to aws_elasticache_replication_group: snapshot_window


xsellier:
Hi there,

### Terraform Version
Terraform v0.8.5

### Affected Resource(s)
- aws_elasticache_replication_group

### Terraform Configuration Files
```hcl
resource "aws_elasticache_replication_group" "redis" {
  replication_group_id          = "tf-rep-group-1"
  replication_group_description = "test description"
  node_type                     = "cache.m1.small"
  number_cache_clusters         = 2
  port                          = 6379
  parameter_group_name          = "default.redis3.2"
  availability_zones            = ["us-west-2a", "us-west-2b"]
  automatic_failover_enabled    = true
  snapshot_window               = ""
}
```

### Panic Output

```
Errors:

  * aws_elasticache_replication_group.redis: "snapshot_window" must satisfy the format of "hh24:mi-hh24:mi".
```
### Expected Behavior
- As in terraform 0.8.2, it does not set the field `snapshot_window`

### Actual Behavior
- As in terraform 0.8.5, it tries to validate "zero" value as a date.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Nope

### References
- https://github.com/hashicorp/terraform/issues/5471

xsellier:
Bug introduced with this commit: https://github.com/hashicorp/terraform/commit/c560c72e58f96603873f54570511c4f58251d1fb

grubernaut:
Hi @xsellier, thanks for the issue!

It doesn't look to me like terraform would return an error if the `snapshot_window` attribute was not specified. Is there a need for specifying the following?
```
snapshot_window = ""
```

xsellier:
In fact, we have a variable list that describe where each item describe an environment type:

```hcl
variable "cache_snapshot_window" {
  description = "Snapshot window"
  default = {
    develop = ""
    qa = ""
    uat = ""
    stage = ""
    loadtest = "04:00-05:00"
    production = "04:00-05:00"
  }
}
```

Meaning, for develop, uat and stage environments don't need snapshot at all.

PR #11677, opened 2017-02-03
provider/aws: Fix OnceADayWindowFormat Validator

grubernaut:
If an attribute that would normally be validated by `validateOnceADayWindowFormat()` is an empty string, in every case AWS supplies a default time window to use.

Thus, an empty string for these attributes is a valid value.

Fixes: #11663


xsellier:
Interesting I have created another kind of merge request:
https://github.com/hashicorp/terraform/pull/11679

Since every value can be empty, we should allow them all by default.

grubernaut:
@xsellier we can't allow them through validation by default, as some attributes, if specified, cannot be an empty string. This works for all resource attributes that use these validators because the API will allow an empty string as a default value. 
Closing in favor of #11679

PR #11678, opened 2017-02-03
provider/aws: fix instance profile creation false negative

snakeb1t:
fixes https://github.com/hashicorp/terraform/issues/9474

discussion of approach in https://github.com/hashicorp/terraform/pull/11634

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_instance_profile.go
+++ b/builtin/providers/aws/resource_aws_iam_instance_profile.go
@@ -121,6 +121,17 @@ func resourceAwsIamInstanceProfileCreate(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("Error creating IAM instance profile %s: %s", name, err)
 	}
 
+	waiterRequest := &iam.GetInstanceProfileInput{
+		InstanceProfileName: aws.String(name),
+	}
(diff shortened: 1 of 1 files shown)


snakeb1t:
Hey @catsby , here's the PR you suggested!

catsby:
Merged and cherry-picked into `0-8-stable`

tomelliff:
@radeksimko Am I right in thinking that this means that the naive 15s wait on instance creation https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_instance.go#L371 can now be removed?

catsby:
> Am I right in thinking that this means that the naive 15s wait on instance creation https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_instance.go#L371 can now be removed?

I don't believe so – IAM has been a source of inconsistency in the API for some time, so while the IAM API may report the profile to exist, an immediate follow up call from EC2 may not get the same result, or perhaps nay not report that the profile has any Roles yet. 

Ideally everything will be fine, and leaving the retry will be harmless because we'll never need to retry. However, APIs for distributed clouds are not always ideal so we should leave the retry for now 😄

PR #11680, opened 2017-02-03
terraform: data source on refresh should just delete from state

mitchellh:
This was caught by an acceptance test. We've now added a unit test. When
refreshing, an orphan (no config) data source should just be deleted.

This only affects master/0.9 since these graph changes aren't in 0.8.x


mitchellh:
Going to merge, doesn't affect running users since its master only and its fairly straightforward, want to get tests green!

PR #11686, opened 2017-02-03
Enable state locking for plan/apply/destroy/refresh/taint/untaint

jbardin:
This enables the locking of state through the command UI. 

This does change the behavior of 2 tests. Previously when running a plan with no existing state, the plan would be written out and then backed up on the next WriteState by another BackupState instance. Since we now maintain a single State instance throughout an operation, the backup happens before any state exists so no backup file is created. This shouldn't be a problem, as there really was nothing that required backing up. Now those tests will create the state file before running. 

The lock/unlock terraform commands will be added in another PR. Only local state is supported so far, so the commands aren't yet required.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -22,7 +22,8 @@ type Backend interface {
 
 	// State returns the current state for this environment. This state may
 	// not be loaded locally: the proper APIs should be called on state.State
-	// to load the state.
+	// to load the state. If the state.State is a state.Locker, it's up to the
+	// caller to call Lock and Unlock as needed.
 	State() (state.State, error)
 }
 
@@ -38,6 +39,9 @@ type Enhanced interface {
 	// It is up to the implementation to determine what "performing" means.
 	// This DOES NOT BLOCK. The context returned as part of RunningOperation
 	// should be used to block for completion.
+	// If the state used in the operation can be locked, it is the
+	// responsibility of the Backend to lock the state for the duration of the
+	// running operation.
 	Operation(context.Context, *Operation) (*RunningOperation, error)
 }
 
@@ -99,6 +103,10 @@ type Operation struct {
 	// Input/output/control options.
 	UIIn  terraform.UIInput
 	UIOut terraform.UIOutput
+
+	// If LockState is true, the Operation must Lock any
+	// state.Lockers for its duration, and Unlock when complete.
+	LockState bool
 }
 
 // RunningOperation is the result of starting an operation.
--- a/backend/local/backend_apply.go
+++ b/backend/local/backend_apply.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform/backend"
+	"github.com/hashicorp/terraform/state"
 	"github.com/hashicorp/terraform/terraform"
 )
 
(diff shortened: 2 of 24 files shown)


pchaganti:
👍

backend/backend.go
```diff
@@ -22,7 +22,8 @@ type Backend interface {
 
 	// State returns the current state for this environment. This state may
 	// not be loaded locally: the proper APIs should be called on state.State
-	// to load the state.
+	// to load the state. If the state.State is a state.Locker, it's up to the
+	// caller to call Lock and Unlock as needed.
```
mitchellh: Thanks for updating the comment, this is the correct behavior I wanted!

backend/local/backend_apply.go
```diff
@@ -28,12 +29,22 @@ func (b *Local) opApply(
...
 		runningOp.Err = err
 		return
 	}
 
+	// context acquired the state, and therefor the lock.
+	// Unlock it when the operation is complete
+	defer func() {
+		if s, ok := opState.(state.Locker); op.LockState && ok {
+			if err := s.Unlock(); err != nil {
+				log.Printf("[ERROR]: %s", err)
```
mitchellh: We should multierror append the error to `runningOp.Err` so that the error shows up to the end user. I would make a long const (like other error messages in the package) for it so that the user knows what to do: verify everything is okay, manually call `terraform unlock`

command/meta.go
```diff
@@ -83,12 +83,15 @@ type Meta struct {
...
 	// provider is to specify specific resource providers
+	//
+	// lockState is set to false to disable state locking
 	statePath    string
 	stateOutPath string
 	backupPath   string
 	parallelism  int
 	shadow       bool
 	provider     string
+	lockState    bool
```
mitchellh: Nitpick: let's name this `stateLock` just to match the other state-related fields above (statePath, stateOutPath)

command/apply.go
```diff
@@ -272,6 +274,8 @@ Options:
                          modifying. Defaults to the "-state-out" path with
                          ".backup" extension. Set to "-" to disable backup.
 
+  -lock-state=true       Lock the state file when locking is supported.
```
mitchellh: Bike shed: Let's just use `-lock`. Maybe we'll lock more in the future maybe we won't but I think its clear regardless and I'd prefer the aesthetic of it.

backend/backend.go
```diff
@@ -99,6 +103,10 @@ type Operation struct {
 	// Input/output/control options.
 	UIIn  terraform.UIInput
 	UIOut terraform.UIOutput
+
+	// If LockState is true, the Operation must Lock any
+	// state.Lockers for its duration, and Unlock when complete.
```
mitchellh: Add to the comment: if using `backend.Local`, it is up to the caller to unlock the state.

backend/local/backend_apply.go
jbardin: Ah yes. I was thinking this would only apply to LocalState, where Unlock shouldn't error, and the lock is gone on exit anyway.

backend/backend.go
jbardin: I don't think `backend.Local` is an exception here. The state is acquired and used solely within `Enhanced.Operation`, which is done for `backend.Local` too. I did note that `Backend.State` expects the caller to lock the state as needed, and that's called from within an Operation.

PR #11688, opened 2017-02-03
[Rancher] - allow for importing resources using environment ID to target

johnrengelman:
/cc @raphink @mcanevet 

This is the follow-up to https://github.com/hashicorp/terraform/pull/11594 to allow for specify additional metadata about Rancher resources on import.


stack72:
LGTM!

PR #11699, opened 2017-02-04
validate aws_alb_target_group name is less than 32 characters

netjunki:
aws_alb_target_group.test: Error creating ALB Target Group: ValidationError: Target group name 'VeryLongGroupNameWhichCausesAnError' cannot be longer than '32' characters
    status code: 400, request id: cb020159-ea78-11e6-8546-15674f4fb6d2

I kept encountering this error during terraform apply and thought it would be better to validate that the field conforms to validation rules during plan. This PR adds a validation to the name field to ensure it is less than 32 characters.


stack72:
LGTM! Thanks for this :)

PR #9729, opened 2016-10-30
Add 'slice' interpolation function

ewbankkit:
`slice(list, from, to)` - Returns the portion of `list` between `from` (inclusive) and `to` (exclusive)
e.g. `slice(var.list_of_strings, 0, length(var.list_of_strings) - 1)`.

Fixes https://github.com/hashicorp/terraform/issues/8846.

PR #11704, opened 2017-02-04
Update for new HIL type system

apparentlymart:
This is the Terraform counterpart to hashicorp/hil#42.

After that change is merged, HIL's type system is extended to support parameterized types. The initial use of this is to represent explicitly the element types of lists and map values, where previously HIL merely represented the list-ness and map-ness of these values without speaking about the element type.

A lot of this change is simply adjusting the function definitions to use the new type definition syntax, where `TypeList` and `TypeMap` have become struct types rather than enumeration values as before.

However, this also presented an opportunity to remove some user-frustrating limitations in many of the functions. In Terraform 0.7 we introduced the idea of maps and lists but that initial implementation was slightly hobbled by this HIL type system limitation, leading to the introduction of the concepts of "flat lists" and "flat maps", where functions would take only lists or maps of string values due to being unable to operate generically over all types.

As a consequence of this, several list and map functions are now able to operate generically over arbitrary element types, including arbitrarily-nested structures as long as the element types remain homogeneous. For example, the `concat` function can now concatenate two lists of lists of strings into a single list of lists of strings, and `lookup` can allow defaulted lookups into maps of lists. Full details on these changes can be found in the diff for the interpolation function documentation, and when merging this we may wish to enumerate these separately in CHANGELOG to present this change in a user-oriented manner, rather than a type-system-nerd-oriented manner.

This also inherits the updated HIL functionality that allows the index syntax to be a true operator in its own right, allowing chaining of indexing and indexing function return values:

* `something[i][j]`
* `somefunc()[i]`

**Note that the vendor update for HIL is included in this PR**. This will need to be rebased if there are any updates to the hashicorp/hil#42 branch during its review.


apparentlymart:
(There are still some bugs in the HIL change that need to be resolved before the tests will pass here.)

apparentlymart:
As well as the conflict that Github already detected, this will likely also need an adjustment to the implementation of the new `slice` function that was added in #9729, which will need to be altered to use `ReturnTypeFunc`.

apparentlymart:
I have fixed the merge config (which was in docs) and updated the `slice` function implementation to be compatible with the new approach.

It looks like this is now getting tripped up on a test added for #12210 in the mean time:

```
--- FAIL: TestResourceGH12183_basic (0.03s)
		
		* test_resource_gh12183.b: inconsistent map element types; previously string but found list of string at key rules
FAIL
FAIL	github.com/hashicorp/terraform/builtin/providers/test	0.740s
```

I didn't yet have a chance to dig into what's going on here. Probably it's a bug that was always there and this new test is just exposing it for the first time. @mitchellh if you happen to have a good guess as to what's going on here I'd appreciate it, but if it's not immediately obvious I'll try to dig in next weekend and debug in more detail.

apparentlymart:
This is unlikely to land in its current state and I don't have the energy to keep it mergeable for a long period so I am going to close this and plan to revisit at some later date when it's more of a core priority.

Issue #4335, opened 2015-12-16
aws_cloudformation_stack indefinitely reapplies changes in NoEcho parameters


mtb-xt:
If a cloudformation template contains a parameter with NoEcho set to true (like a password), the parameter value is shown as asterisks  "****".

Terraform 0.6.8 detects this as a parameter change and tries to reapply it on every 'plan' or 'apply':

```
aws_cloudformation_stack.redshift-prod: Modifying...
  parameters.MasterUserPassword: "****" => "lolmegapassword"

```

radeksimko:
Thanks for this report too. :smiley: 

A solution could be ignoring changes in `NoEcho` parameters coming from AWS, the only difficult part is in decoding the whole template as this attribute isn't available from [the API directly](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Parameter.html).

radeksimko:
This project may be helpful here:
https://github.com/awslabs/aws-cfn-go-template

bobzoller:
Perhaps this is silly but it might also save someone from a headache:  _Don't_ be like me and try to work around this by adding:

```
lifecycle {
  ignore_changes = ["parameters.Password"]
}
```

All it means is that the next `apply` will write the `****` value back to AWS (instead of your secret).

dylanvaughn:
Maybe if the parameter name ends in ```NoEcho``` then that tells Terraform to ignore it?  i.e. ```MasterDbPasswordNoEcho```

I could implement this in a PR if it seems like an acceptable solution.

dylanvaughn:
I ended up doing a slightly different approach of just checking if the old value of the parameter coming from AWS is ```****``` - see https://github.com/hashicorp/terraform/pull/11707

jsierles:
This issue still exists today, and the fix in the PR was never addressed. Does anyone have a workaround?

PR #11707, opened 2017-02-05
provider/aws: ignore NoEcho parameters in diff

dylanvaughn:
Hi there - this ignores NoEcho params when determining if a CFN stack should be updated - fixes:

https://github.com/hashicorp/terraform/issues/4335

I thought about parsing the ```template_body``` to really see if a parameter has been set to ```NoEcho``` but since ```template_url``` could also be specified it seems pretty heavy-handed to potentially pull down the template and then parse it for each compared parameter.

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudformation_stack.go
+++ b/builtin/providers/aws/resource_aws_cloudformation_stack.go
@@ -67,6 +67,10 @@ func resourceAwsCloudFormationStack() *schema.Resource {
 				Type:     schema.TypeMap,
 				Optional: true,
 				Computed: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// ignore NoEcho parameters
+					return old == "****"
+				},
 			},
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @dylanvaughn 
thanks for the PR.

I'm afraid we're going to have to parse the template. This solution would in practice be ignoring _four stars_ - which may (or may not) be genuine parameter value (may or may not have `NoEcho` flag).

Why do you think that parsing the template would be heavy-handed? https://github.com/awslabs/aws-cfn-go-template/blob/master/template.go could help with parsing itself and even if the user has specified `template_url` we pull down the body as part of the `Read()` function already anyway:

https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_cloudformation_stack.go#L289-L293

dylanvaughn:
@radeksimko ok sounds good - I didn't realize we already had the template_body from s3 - I will give parsing the template a shot and reopen

pauldraper:
I assume this effort is dead?

PR #11724, opened 2017-02-06
add force-unlock command

jbardin:
Implement the basics for the lock and unlock commands


command/lock.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *LockCommand) Help() string {
+	helpText := `
+Usage: terraform lock [DIR]
+
+  Manually lock the state for the defined configuration.
+
+  This will not modify your infrastructure. This command obtains a lock on the
+  state for the current configuration. The behavior of this lock is dependent
+  on the backend being used. A lock on a local state file only lasts for the
+  duration of the calling process.
```
apparentlymart: If I'm understanding this message, it seems to be effectively saying "this command doesn't work for local state files". Perhaps it would be better to spell that out specifically, rather than merely alluding to it.

(I'm actually kinda skeptical about manual locking and unlocking general to be honest, but I trust y'all already convinced yourselves that the benefits are worth the workflow complexity cost. :grinning:)

jbardin: The command does "work" for local state files, but yes, it's not very useful with the current local lock implementation 😉 

The `unlock` command is mostly meant to recover from a failed unlock situation (TF crash, network failure, etc). The `lock` command probably isn't required, but does provide parity plus the possibility of using it for testing. Neither of them are intended to be used as part of a general workflow, so shouldn't be considered part of one any more than the `terraform state` commands.

apparentlymart: The symmetry here may work against us by making this feel similar to Subversion locks, explicit database transactions, and other such things that Terraform users are likely to have experience with. This creates a "false cognate" that is likely to lead users astray.

I would define "work" here -- from a user's perspective -- as "create a lock that persists until the user runs `unlock`"... I worry that `terraform lock` command that *can't* guarantee to behave that way would lead people into believing that Terraform has an explicit locking workflow and only later finding that it's been ineffective due to the specific behavior of the backend, possibly after it's too late and data has already been lost.

I think I'd lean towards renaming `unlock` to `force-unlock`, to make it clearer that this is something exceptional that you shouldn't normally need to do, and then eliminating the `lock` command entirely.

If the testing use-case is important then perhaps we can have a `test-lock` command that acquires a lock *and then immediately releases it*, allowing users to exercise the locking behavior (e.g. to validate configuration/credentials) without creating the impression that long-lived locks are an expected workflow.

jbardin: @apparentlymart, thanks, it's probably a good idea to consider user's preconceived notions about what the locking interface may actually do. I left this slightly ambiguous so backend implementations could choose the semantics, and in theory a backend could make a workflow around manual lock/unlock calls if desired. I'm not saying this is a god idea, and this might be the place to thwart that ahead of time.

Maybe `unlock` on its own is OK without the associated `lock` function, and in this case `lock` may not be needed at all. 

@mitchellh: I'm not tied to lock/unlock at all; do you have any preference on the UI here?

command/lock.go
```diff
@@ -0,0 +1,78 @@
...
+
+	st, err := b.State()
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Failed to load state: %s", err))
+		return 1
+	}
+
+	s, ok := st.(state.Locker)
+	if !ok {
+		c.Ui.Error("Current state does not support locking")
```
mitchellh: I think this needs to be a lot more detailed to explain what locking is, why it may not be supported, etc.

command/lock.go
```diff
@@ -0,0 +1,78 @@
...
+		return 1
+	}
+
+	if err := s.Lock("lock"); err != nil {
+		c.Ui.Error(fmt.Sprintf("Failed to lock state: %s", err))
+		return 1
+	}
+
+	return 0
```
mitchellh: Above this we probably need a green success message for locking and explaining the semantics of that, the importance of calling `unlock`, etc.

command/lock.go
mitchellh: Good reasons here. Let's start with force-unlock only (remove the lock command). If the use case arises for a manual lock command we can add it later.

jbardin:
Removed `lock` command, changed `unlock` to `force-unlock`, added success output message (in red because continuing could still be dangerous), and added confirmation prompt for unlocking.

command/unlock.go
```diff
@@ -0,0 +1,131 @@
...
+	force := false
+	cmdFlags := c.Meta.flagSet("force-unlock")
+	cmdFlags.BoolVar(&force, "force", false, "force")
+	cmdFlags.Usage = func() { c.Ui.Error(c.Help()) }
+	if err := cmdFlags.Parse(args); err != nil {
+		return 1
+	}
+
+	// assume everything is initialized. The user can manually init if this is
+	// required.
```
mitchellh: If things aren't initialized the Backend call below will error anyways. :)

PR #11729, opened 2017-02-06
Consul Data Sources for nodes and services

sean-:
More to come, including docs, but putting this up for review quickly before I pull down the rest of the end points and merge it into this scheme.


sean-:
Offline LGTM from @catsby .  Merging.

partsalliance:
@sean- following the docs in the PR with terraform v0.8.7 I am getting: 
 Provider doesn't support data source: consul_catalog_services

Anyone else had any luck using the new data sources?

sean-:
This data source isn't in 0.8.7, but will be available starting with the 0.8.8 release.  https://github.com/hashicorp/terraform/commit/b7c9dedf4ee4949af7ff09a3e6169bee3c5c748e was committed after https://github.com/hashicorp/terraform/commit/c1df88ac4633b61b83802bbb10c343c20d3035b5 by about 12hrs.  If you checkout the `0-8-stable` build and run a `make dev` you could get early access to this data source.

PR #11743, opened 2017-02-07
Docs: Remove duplicate NoMad provider, fix sorting

vladgolubev:
Ignition wasn't in A-Z order, fixed identation, removed duplicated NoMad provider:

image


apparentlymart:
Looks good to me! Thanks for doing this, @vladgolubev!

PR #11745, opened 2017-02-07
Fix aws_lambda_function attributes when more than 50 versions

### LINKED ISSUE #11720: aws_lambda_function attributes for version and qualified_arn not set
When there are more than 50 versions of a lambda function, the `version` and `qualified_arn` attrinutes are always returned as empty strings.

There is a logic inversion issue between the rturn value of the callback passed to `listVersionsByFunctionPages` in `resourceAwsLambdaFunctionRead` and it's usage as `should_continue` in `listVersionsByFunctionPages`.  If this were fixed the function would probably never exit, as `page.NextMarker` is not propagated from the output of one request to the inout of the next.

Note that using the aws command line client, the `--max-items` argument is ignored if it is greater than 50.  I suspect that the value specified for MaxItems in the code (10000) is also ignored.

Checked on 2e0b30e.

hugoduncan:
Fixes the version and qualified_arn attributes, when the lambda
function has more than 50 versions.

Closes #11720


stack72:
LGTM! Thanks for this :)

PR #11762, opened 2017-02-08
website: describe API Gateway ARN's

kevinburke:
I lost a few hours figuring out the right way to describe an ARN for an API
Gateway resource. Specifically I translated the example poorly since I didn't
realize I had to append the path onto the end of the ARN.

Adds two links to an Amazon documentation page describing the format for API
Gateway ARN's. Adds an additional path component to the ARN example so you can
see you need to specify paths.


stack72:
Hi @kevinburke 

Thanks for this - this is a great enhancement to our documentation!

Paul

PR #11764, opened 2017-02-08
Minor fix to the triton_machine provider. The docs referred to a `net…

zero-below:
Terraform v0.8.6 triton provider seems to be out of sync with the docs. The docs describe a `networks` list of strings, while the code seems to be expecting a `nics` list of maps, with the `network` listed in an appropriate `nic` entry. 

There seems to be some additional functionality in the code for configuring a `nic`, but I haven't tried it, so didn't add the docs.


stack72:
Hi @zero-below 

thanks for this - I am going to pull this locally and make 1 change before pushing it - should be a "list of NIC blocks" not list of maps

Thanks for the work here

Paul

Issue #11591, opened 2017-02-01
Terraform does not set id attribute not default to name attribute in consul_catalog_entry's service block


manuwelakanade:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.8.5

### Affected Resource(s)
consul_catalog_entry

### Terraform Configuration Files
resource "consul_catalog_entry" "service_consul_registration" {
  depends_on = ["null_resource.some_resource"]
  count = "${var.service_count}"
  address = "127.0.0.1"
  node = "node_name"
  service = {
      name = "service_name"
      address = "${var.service_ip}"
      port = "${var.service_port}"
  }
}

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: 

### Expected Behavior
What should have happened?
I should have seen the ID generated for consul_catalog_entry as ID: node-name-127.0.0.1[service_name]

### Actual Behavior
What actually happened?
The ID generated for consul_catalog_entry is node-name-127.0.0.1[ ]

### Steps to Reproduce
`terraform apply`

manuwelakanade:
Hi, I would like to work on this issue.

stack72:
@manuwelakanade please feel free to open a PR for this :) We would certainly appreciate it!

manuwelakanade:
PR is merged hence closing the issue

PR #11772, opened 2017-02-08
provider/consul: catalog entry service id should default to service name

manuwelakanade:
provider/consul: catalog entry service id should default to service name


manuwelakanade:
This PR fixes: https://github.com/hashicorp/terraform/issues/11591

sean-:
I have my fingers in this code, but this LGTM.  Merging.

PR #11787, opened 2017-02-08
Add consul state locking

jbardin:
Lock state stored in the consul backend. 

Use consul locks to implement state locking. The lock path is state path+"/.lock" which matches the consul cli default for locks. Lockinfo is stored at path+"/.lockinfo"

Refactor to a single `TestRemoteLocks` function that all backends can use.

### DIFF
--- a/backend/remote-state/consul/client.go
+++ b/backend/remote-state/consul/client.go
@@ -2,15 +2,35 @@ package consul
 
 import (
 	"crypto/md5"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
 
 	consulapi "github.com/hashicorp/consul/api"
+	"github.com/hashicorp/errwrap"
+	multierror "github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform/state/remote"
 )
+
+const (
+	lockSuffix     = "/.lock"
+	lockInfoSuffix = "/.lockinfo"
+)
+
+// TODO: use single LockInfo struct
+type lockInfo struct {
+	Created time.Time
+	Info    string
+}
 
 // RemoteClient is a remote client that stores data in Consul.
 type RemoteClient struct {
 	Client *consulapi.Client
 	Path   string
+
+	consulLock *consulapi.Lock
+	lockCh     <-chan struct{}
 }
 
 func (c *RemoteClient) Get() (*remote.Payload, error) {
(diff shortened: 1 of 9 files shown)


jbardin:
Since consul tests have to run as acceptance tests right now:

```
$ CONSUL_HTTP_ADDR=127.0.0.1:8500 go test -v github.com/hashicorp/terraform/backend/remote-state/consul
=== RUN   TestRemoteClient
2017/02/08 11:40:41 [DEBUG] New state was assigned lineage "8942c680-0fc0-4416-a381-70a88978fc97"
2017/02/08 11:40:41 [TRACE] Preserving existing state lineage "8942c680-0fc0-4416-a381-70a88978fc97"
--- PASS: TestRemoteClient (0.00s)
=== RUN   TestConsul_stateLock
--- PASS: TestConsul_stateLock (1.04s)
	testing.go:68: lock info error: state locked. path:"tf-unit/2017-02-08 11:40:41.358860741 -0500 EST", created:2017-02-08 16:40:41.360109197 +0000 UTC, info:"test client A"
PASS
ok  	github.com/hashicorp/terraform/backend/remote-state/consul	1.069s
```

backend/remote-state/consul/client.go
```diff
@@ -2,15 +2,29 @@ package consul
...
+	lockInfoSuffix = "/.lockinfo"
+)
+
 // RemoteClient is a remote client that stores data in Consul.
 type RemoteClient struct {
 	Client *consulapi.Client
 	Path   string
+
+	consulLock *consulapi.Lock
+	lockCh     <-chan struct{}
```
mitchellh: I think both of the above need a mutex slapped around it, probably around Lock/Unlock here since we read/write them in there and it is possible that we call those concurrently (though we probably shouldn't within-process).

jbardin: Concurrent calls of Lock would definitely be a bug. Lock implementations aren't guaranteed to be reentrant, so we can't ever count on multiple Lock calls working.

I think it's better to ensure that we internally maintain unambiguous locations for taking and releasing the lock, rather than sprinkling Unlock around to hope it gets called.

mitchellh: I agree, but even in that case it may be worth it to protect r/w access to these values with a sync.Mutex. As it stands currently if you called both concurently you'd 1.) probably cause problems in general but 2.) violate the Go memory model regardless.

jbardin: Good point, can't hurt to drop it in ;)

jbardin: Actually, now that I'm writing the comment for this commit, it seems more superfluous. It's basically stating that Lock and Unlock are not safe to call concurrently, but in this one instance we have a lock just in case we forget and break our own API. 🤔

mitchellh:
That's a good point. Merge on!

backend/remote-state/consul/client.go
mitchellh: Commented via email didn't work as well as I'd hope. This is a good point. Merge on :)

Issue #8605, opened 2016-09-01
[GKE] Support autoscaling


thecodeassassin:
hey all,

Google cloud supports autoscaling for nodes which is very useful in high traffic kubernetes deployments. Could you add this feature to terraform? Currently i use local exec like this:

``` bash
  # enable autoscaling
  provisioner "local-exec" {
      command = "gcloud container clusters update ${var.google_container_cluster.name} --enable-autoscaling --min-nodes=${var.min_nodes} --max-nodes=${var.max_nodes} --project='${var.project_name}'"
  }
```

Documentation is here
https://cloud.google.com/container-engine/docs/clusters/operations

cblecker:
Hey @stack72 --
I think this should be provider/google instead of provider/aws :)

cblecker:
@evandbrown @roberthbailey --
It appears that `google.golang.org/api/container/v1` has fallen behind the public documentation (similar to how it did in https://github.com/google/google-api-go-client/issues/127). The API documentation (https://cloud.google.com/container-engine/reference/rest/v1/projects.zones.clusters.nodePools) indicates there is an `autoscaling` field, however https://www.googleapis.com/discovery/v1/apis/container/v1/rest doesn't specify that this field exists.

I'm guessing that if https://www.googleapis.com/discovery/v1/apis/container/v1/rest is updated, that the auto-generated `google.golang.org/api/container/v1` will update, and then can be re-vendored into Terraform.

Is there a place to open an issue to get the API discovery point updated?

roberthbailey:
https://container.googleapis.com/$discovery/rest?version=v1 shows the `autoscaling` field. And that link is referenced from https://cloud.google.com/container-engine/reference/rest/ as the discovery URL for the API. What led you to the googleapis.com URL?

evandbrown:
@cblecker I'm mobile at the moment but agree this is likely due to an old version of the vendored package. I'll get that updated tomorrow and add the necessary field to the cluster resource.

cblecker:
@roberthbailey: That Github issue I linked to from earlier this year pointed me at that URL: https://github.com/google/google-api-go-client/issues/127 .

It also appears that https://github.com/google/google-api-go-client/blob/master/api-list.json#L1334 in the `google-api-go-client` package points to that discovery URL. Perhaps that means it could be as simple as updating the discovery URL is updated in the `google-api-go-client`?

[...]

PR #11802, opened 2017-02-08
provider/google: initial commit for node pool resource

danawillow:
Purposely kept to just required fields for now to keep the PR small.

Will enable further development in this resource, such as is necessary for #8605.

### DIFF
--- a/builtin/providers/google/container_operation.go
+++ b/builtin/providers/google/container_operation.go
@@ -1,0 +1,59 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"google.golang.org/api/container/v1"
+)
+
+type ContainerOperationWaiter struct {
+	Service *container.Service
+	Op      *container.Operation
+	Project string
+	Zone    string
+}
+
+func (w *ContainerOperationWaiter) Conf() *resource.StateChangeConf {
+	return &resource.StateChangeConf{
+		Pending: []string{"PENDING", "RUNNING"},
+		Target:  []string{"DONE"},
+		Refresh: w.RefreshFunc(),
+	}
+}
+
+func (w *ContainerOperationWaiter) RefreshFunc() resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		resp, err := w.Service.Projects.Zones.Operations.Get(
+			w.Project, w.Zone, w.Op.Name).Do()
+
+		if err != nil {
+			return nil, "", err
+		}
+
+		log.Printf("[DEBUG] Progress of operation %q: %q", w.Op.Name, resp.Status)
+
+		return resp, resp.Status, err
+	}
+}
+
+func containerOperationWait(config *Config, op *container.Operation, project, zone, activity string, timeoutMinutes, minTimeoutSeconds int) error {
+	w := &ContainerOperationWaiter{
+		Service: config.clientContainer,
+		Op:      op,
+		Project: project,
+		Zone:    zone,
+	}
+
+	state := w.Conf()
+	state.Timeout = time.Duration(timeoutMinutes) * time.Minute
+	state.MinTimeout = time.Duration(minTimeoutSeconds) * time.Second
+	_, err := state.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error waiting for %s: %s", activity, err)
+	}
+
+	return nil
+}
--- a/builtin/providers/google/resource_container_node_pool.go
+++ b/builtin/providers/google/resource_container_node_pool.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/container/v1"
 	"google.golang.org/api/googleapi"
@@ -17,6 +18,17 @@ func resourceContainerNodePool() *schema.Resource {
 		Exists: resourceContainerNodePoolExists,
 
 		Schema: map[string]*schema.Schema{
+			"project": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{
+					"GOOGLE_PROJECT",
+					"GCLOUD_PROJECT",
+					"CLOUDSDK_CORE_PROJECT",
+				}, nil),
+			},
+
 			"name": &schema.Schema{
 				Type:          schema.TypeString,
 				Optional:      true,
@@ -125,7 +137,7 @@ func resourceContainerNodePoolRead(d *schema.ResourceData, meta interface{}) err
 			return nil
 		}
 
-		return err
+		return fmt.Errorf("Error reading NodePool: %s", err)
 	}
 
 	d.Set("name", nodePool.Name)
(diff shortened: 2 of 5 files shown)


builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+func resourceContainerNodePoolCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
```
paddyforan: I think we want to use a project_id inline here, so we can take advantage of creating projects in Terraform and using them as the host for node pools.

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,148 @@
...
+
+	req := &container.CreateNodePoolRequest{
+		NodePool: nodePool,
+	}
+
+	op, err := config.clientContainer.Projects.Zones.Clusters.NodePools.Create(project, zone, cluster, req).Do()
+
+	if err != nil {
+		return err
```
paddyforan: If we could add some context to this error, that would be awesome. Something like "Error creating node pool:" or something.

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,148 @@
...
+
+	d.SetId(name)
+
+	return resourceContainerNodePoolRead(d, meta)
+}
+
+func resourceContainerNodePoolRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
```
paddyforan: Same here, I think we want project_id in the config, instead.

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,148 @@
...
+	d.Set("name", nodePool.Name)
+	d.Set("initial_node_count", nodePool.InitialNodeCount)
+
+	return nil
+}
+
+func resourceContainerNodePoolDelete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
```
paddyforan: This should probably be read from the Config.

builtin/providers/google/container_operation.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func (w *ContainerOperationWaiter) RefreshFunc() resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		resp, err := w.Service.Projects.Zones.Operations.Get(
+			w.Project, w.Zone, w.Op.Name).Do()
+		log.Printf("[DEBUG] Progress of operation %q: %q", w.Op.Name, resp.Status)
+
+		return resp, resp.Status, err
```
danawillow: Fixed

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+	"google.golang.org/api/container/v1"
+	"google.golang.org/api/googleapi"
+)
+
+func resourceContainerNodePool() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceContainerNodePoolCreate,
+		Read:   resourceContainerNodePoolRead,
+		Delete: resourceContainerNodePoolDelete,
```
danawillow: Done. Can you help me now figure out whether there's anything that can be taken out of the other functions now that the Exists one is there?

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,148 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
danawillow: Done.

builtin/providers/google/resource_container_node_pool.go
danawillow: `project_id` or `project`? I'm seeing both: https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_google_project_iam_policy.go#L21 vs https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_google_project.go#L47.

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	zone := d.Get("zone").(string)
+	name := d.Get("name").(string)
+	cluster := d.Get("cluster").(string)
+
+	nodePool, err := config.clientContainer.Projects.Zones.Clusters.NodePools.Get(
+		project, zone, cluster, name).Do()
+	if err != nil {
+		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
```
paddyforan: I _think_ this should no longer be necessary, as Exists should never let this code execute if the resource doesn't exist. But @radeksimko said something that made me doubt that, so I'm going to follow up with him before I start handing out wrong info.

builtin/providers/google/resource_container_node_pool.go
paddyforan: Still wrong, just need a project property on the resource.

builtin/providers/google/resource_container_node_pool.go
danawillow: Seems reasonable. I added a project field.

builtin/providers/google/resource_container_node_pool.go
danawillow: Done

builtin/providers/google/resource_container_node_pool.go
danawillow: If that is the case, we (I think) still need somewhere to mark the id as empty. I could do it in Exists but the comments around the declaration of it say `The *ResourceData passed to Exists should _not_ be modified.` which certainly is a deterrent for doing it there.

radeksimko: Unless I'm terribly mistaken I don't think you need to run `d.SetId("")` in `Exists` in order to mark it as gone.
That's what happens automatically if `Exists` returns false, see https://github.com/hashicorp/terraform/blob/master/helper/schema/resource.go#L237-L243

radeksimko: That said the `Exists` function will need some extra care because right now it's returning `false` if it encounters _any_ error (could be 503), so the conditional w/ 404 here should be moved down there and `Exists` should always properly return genuine error as `true, error`.

danawillow: Makes sense, thanks @radeksimko!

danawillow:
All right @paddyforan, I think all comments have been resolved at this point. Over to you.

builtin/providers/google/resource_container_node_pool.go
```diff
@@ -0,0 +1,198 @@
...
+		Schema: map[string]*schema.Schema{
+			"project": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{
+					"GOOGLE_PROJECT",
+					"GCLOUD_PROJECT",
+					"CLOUDSDK_CORE_PROJECT",
+				}, nil),
```
danawillow: Huh, for some reason tests were failing when I had it this way before but I just tried again and they passed, so \shrug. Done.

ragdaj:
Is it possible to set machine type for node pool ?

PR #11812, opened 2017-02-09
provider/aws: output the log group name when create fails

netjunki:
I found this error output to not be particularly helpful in tracking down the actual failure:

    * aws_cloudwatch_log_group.app: Creating CloudWatch Log Group failed: ResourceAlreadyExistsException: The specified log group already exists
	status code: 400, request id: f3db0aab-ee7d-11e6-af7b-b1b77eba3b01

So I've added the log group name to the output.

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudwatch_log_group.go
+++ b/builtin/providers/aws/resource_aws_cloudwatch_log_group.go
@@ -54,7 +54,7 @@ func resourceAwsCloudWatchLogGroupCreate(d *schema.ResourceData, meta interface{
 		LogGroupName: aws.String(d.Get("name").(string)),
 	})
 	if err != nil {
-		return fmt.Errorf("Creating CloudWatch Log Group (%s) failed: %s", d.Get("name").(string), err)
+		return fmt.Errorf("Creating CloudWatch Log Group failed: %s:  The CloudWatch Log Group '%s' already exists.", err, d.Get("name").(string))
 	}
 
 	d.SetId(d.Get("name").(string))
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_cloudwatch_log_group.go
```diff
@@ -54,7 +54,7 @@ func resourceAwsCloudWatchLogGroupCreate(d *schema.ResourceData, meta interface{
 		LogGroupName: aws.String(d.Get("name").(string)),
 	})
 	if err != nil {
-		return fmt.Errorf("Creating CloudWatch Log Group failed: %s", err)
+		return fmt.Errorf("Creating CloudWatch Log Group failed: %s:  The CloudWatch Log Group '%s' already exists.", err, d.Get("name").(string))
```
radeksimko: I'm afraid we cannot assume here that `ResourceAlreadyExistsException` is the only kind of error we can ever receive during creation, so we'll need to check via `awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceAlreadyExistsException"` and only print out the helpful message in such cases.

netjunki:
@radeksimko how is this looking now?

No worries about the missing data I know this stuff is tough. Gotta scratch your own itches at the end of the day. :-) And better to contribute the fix back so others can benefit.

Was just thinking though... wouldn't it be useful to know the name of the resource generating the error no matter what type of error is being thrown?

radeksimko:
Looking good, thanks 👍

PR #11829, opened 2017-02-09
provider/aws: Set the qualifier to an empty string if the parsing fai…

catsby:
…ls, to attempt to detect drift

Should fix `TestAccAWSLambdaPermission_basic`


builtin/providers/aws/resource_aws_lambda_permission.go
```diff
@@ -212,8 +212,9 @@ func resourceAwsLambdaPermissionRead(d *schema.ResourceData, meta interface{}) e
 
 	qualifier, err := getQualifierFromLambdaAliasOrVersionArn(statement.Resource)
 	if err == nil {
```
grubernaut: `!=`?

catsby: derp

PR #11836, opened 2017-02-09
Add locking during backend configuration

jbardin:
During backend initialization, especially during a migration, there is a
chance that an existing state could be overwritten.

Attempt to get a locks when writing the new state. It would be nice to
always have a lock when reading the states, but the recursive structure
of the Meta.Backend config functions makes that quite complex.

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -1808,7 +1808,9 @@ var ErrNoState = errors.New("no state")
 // was written by WriteState.
 func ReadState(src io.Reader) (*State, error) {
 	buf := bufio.NewReader(src)
-	if _, err := buf.Peek(1); err == io.EOF {
+	if _, err := buf.Peek(1); err != nil {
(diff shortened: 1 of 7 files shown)


terraform/state.go
```diff
@@ -1808,7 +1808,9 @@ var ErrNoState = errors.New("no state")
 // was written by WriteState.
 func ReadState(src io.Reader) (*State, error) {
 	buf := bufio.NewReader(src)
-	if _, err := buf.Peek(1); err == io.EOF {
+	if _, err := buf.Peek(1); err != nil {
+		// the error is either io.EOF or "invalid argument", and both are from
+		// an empty state.
```
mitchellh: Glad you got this, a unit test for this would be good.

jbardin: Noticed bc some command tests started failing.
Added a test specifically for this too.

PR #11840, opened 2017-02-10
provider/google: Update node version in container cluster test

danawillow:
The previous version is no longer valid, so the test was consistently failing.


paddyforan:
You're on a roll! :+1: Thank you!

PR #11848, opened 2017-02-10
Added scope alias for cloud-dns management

mblankleder:


danawillow:
Can you point me to where in the documentation you found the shorthand cloud-dns-ro and cloud-dns-rw?

Issue #10272, opened 2016-11-21
provider/aws: Cleanup the Lambda ENI deletion fails on destroy (after update Terraform to v0.7.11)


mgagliardo:
Hi there,

I was running v0.7.8 and recently upgraded to v0.7.11 (without jumping to 0.7.9 or 0.7.10 first). It seems that [GH-5767](https://github.com/hashicorp/terraform/issues/5767) came back at some point between those versions.

Basically we have an SNS topic that triggers a lambda function within a VPC, if at least once this lambda function is used, the ENI attached to it won't get removed and the destroy command will fail and we'll need to remove the rest of the VPC components by hand.

Also need to mention that between upgrade 0.7.8 -> 0.7.11 I did not even touch the code, neither from the lambda function nor the terraform config files.

Should you need any additional info please let me know.

Thanks a lot in advance! (and sorry if I forgot to add some info, first issue on TF here!)

---

### Terraform Version
```
$ terraform --version
Terraform v0.7.11
```

### Affected Resource(s)
- aws_security_group.lambda_sg
- aws_subnet

### Terraform Configuration Files
I will just add some example code as there is some sensitive information on mine, but this should be enough to check out.

```hcl
resource "aws_security_group" "lambda_sg" {
  name        = "lambda-cert-removal-sg-service"
  vpc_id      = "${var.vpc_id}"
}

resource "aws_lambda_function" "main" {
    function_name    = "${var.function_name}"
    description      = "Lambda function"
    filename         = "./my_file_path.zip"
    role             = "${aws_iam_role.iam_role.arn}"
    runtime          = "python2.7"
    timeout          = "60"
    handler          = "certs_removal_script.lambda_handler"

    vpc_config       = {
      subnet_ids         = [ "subnet-4ab34e12,subnet-9j5gs95s" ]
      security_group_ids = [ "${aws_security_group.lambda_sg.id}" ]
    }
}
```

### Panic Output
```
[...]
module.base_vpc.aws_subnet.private.1: Still destroying... (4m0s elapsed)
module.base_vpc.aws_subnet.private.1: Still destroying... (4m10s elapsed)
module.base_vpc.aws_subnet.private.1: Still destroying... (4m20s elapsed)
module.base_vpc.aws_subnet.private.1: Still destroying... (4m30s elapsed)
module.base_vpc.aws_subnet.private.1: Still destroying... (4m40s elapsed)
[...]

PR #11849, opened 2017-02-10
ISSUE-10272 Lets not look at the requestor id.

grayaii:
This is to fix: https://github.com/hashicorp/terraform/issues/10272

Copied from that link:
I can't find (nor can I remember) why RequesterId was even added here (most likely as an extra safety net to make sure we don't delete the wrong ENI), but deleteLingeringLambdaENIs only gets called when resourceAwsSecurityGroupDelete is called, and the filters are "security group-id" and "description", so it should be "good enough". I still don't know who added "awslambda_" to the requestor-id (terraform? or AWS?), but I removed that filter and things seem to be working fine.


grayaii:
@gposton , @radeksimko , can you guys comment on this?

gozer:
For reference, here is what such an ENI looks like right now:

```json
{
            "Status": "in-use", 
            "MacAddress": "00:XX:00:XX:00:XX", 
            "SourceDestCheck": true, 
            "VpcId": "vpc-xxxx", 
            "Description": "AWS Lambda VPC ENI: 4d30b479-eadf-4293-ad47-034ecbd3c3e5", 
            "NetworkInterfaceId": "eni-XXXXXX", 
            "PrivateIpAddresses": [
                {
                    "PrivateDnsName": "ip-x-x-x-x.us-west-2.compute.internal", 
                    "Primary": true, 
                    "PrivateIpAddress": "10.x.x.x"
                }
            ], 
            "RequesterManaged": false, 
            "Groups": [
                {
                    "GroupName": "xxx-admin-earzgwd5cfcepa6m6qeutegqgm", 
                    "GroupId": "sg-xxxx"
                }, 
            ], 
            "PrivateDnsName": "ip-xx-xx-xx-xx.us-west-2.compute.internal", 
            "AvailabilityZone": "us-west-2a", 
            "Attachment": {
                "Status": "attached", 
                "DeviceIndex": 1, 
                "AttachTime": "2017-02-10T14:36:03.000Z", 
                "DeleteOnTermination": false, 
                "AttachmentId": "eni-attach-xxxx", 
                "InstanceOwnerId": "aws-lambda"
            }, 
            "RequesterId": "AROAITGDG6GCK7DAPOMEW:<lambda-function-name>", 
            "Ipv6Addresses": [], 
            "SubnetId": "subnet-xx", 
            "OwnerId": "11111111", 
            "TagSet": [], 
            "PrivateIpAddress": "10.xx.xx.xx"
}
```

And as far as I can tell, *AROAITGDG6GCK7DAPOMEW* in this case, is the unique_id attribute of the aws_iam_role I have assigned to the lambda function.

Could it be possible that the requester-id is different when the Lambda function doesn't have an attached role ?

And for the UUID in the description, *4d30b479-eadf-4293-ad47-034ecbd3c3e5*, I can't find that anywhere.

grayaii:
@gozer This PR would have caught that ENI for deletion because it matches the security group id AND description :)  I'm not sure if attached roles play a role in this.  Good point though. The only other "safety net" I see is trapping ""InstanceOwnerId": "aws-lambda"", but that may be overkill, just like I think requester-id is overkill.  I built terraform with this PR and I've had no problems so far building/destroying resources :)

gposton:
@radeksimko, I authored the original PR here: https://github.com/hashicorp/terraform/pull/8033  The code being removed in the PR was an 'extra' check that was just a CMA safety measure, but is ultimately unnecessary.  I support this PR and would like to see it merged ASAP :)

grayaii:
@radeksimko Any chance we can merge this in?  We've been having to compile patched versions of terraform with this fix for quite a while now.

crossing:
I think the filter on description should be removed too. It should be enough to filter on the security group itself. Also `Description` is inherit unreliable field. Its content can change anytime.

willejs:
@radeksimko can you review this please?

willejs:
@stack72 could you take a peek please?

kristinn:
Would be nice to get this fix merged.

brikis98:
+1 on getting this merged. Can't really use Lambda in a VPC without it!

radeksimko:
Hi folks,
sorry to keep you waiting here.

I took some time to read the thread and related data, thanks @gozer for data about ENI - that's very useful in this context. As for the reasons why this started failing - it is possible that Amazon has just changed the format of the requestorId - I have observed tiny changes like this in the past as we run all acceptance tests nightly and sometimes tests start to fail without us changing any code and the root cause is change in the API.

@grayaii I added an acceptance test to cover this behaviour so that we have a chance to notice earlier in the future. I hope you don't mind me pushing that into your branch.

This PR looks good to me now. 👍

PR #11856, opened 2017-02-10
Add support for missing attributes for PagerDuty service resource 

jtsaito:
Currently, the `PagerDuty` `service` resource does not support the `incident_urgency_rule`, `scheduled_actions` and `support_hours` attribute. This PR adds support for the three attributes.

We also add acceptance tests which can be used as examples. If the tests are not essential enough, they can of course be removed.

Please note that update operations (`PUT /services/:id`) on the three attributes don't seem to be working with the SDK or v2 PagerDuty API in spite of being [documented](https://v2.developer.pagerduty.com/v2/page/api-reference#!/Services/put_services_id). We therefore made the attributes `ForceNew: true`.

~~Furthermore, there is an issue with a PagerDuty acceptance test in the current master on my build system. (I did not find an open issue of this; the problem may or may not have to do with my PagerDuty account.) The current implementation does not fix this error but it should be independent of the changes. The other unit and acceptance tests are fine. The error is as follows (both on master and this PR):~~ -> 

**UPDATE**: sorry, my account on PagerDuty contained something braking the tests; all tests are passing fine on master and this branch.

### DIFF
--- a/builtin/providers/pagerduty/import_pagerduty_service_test.go
+++ b/builtin/providers/pagerduty/import_pagerduty_service_test.go
@@ -26,3 +26,24 @@ func TestAccPagerDutyService_import(t *testing.T) {
 		},
 	})
 }
+
+func TestAccPagerDutyServiceWithIncidentUrgency_import(t *testing.T) {
+	resourceName := "pagerduty_service.foo"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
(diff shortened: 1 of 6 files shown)


jtsaito:
@stack72 Thanks for your quick response! I added a commit with documentation now.

stack72:
Thanks @jtsaito :) This now LGTM!!

Paul

jtsaito:
woo-hoo! thanks!

PR #11857, opened 2017-02-10
provider/aws: (optional) append timestamp to final snapshot identifier

netjunki:
tl;dr: add a `timestamp_final_snapshot` parameter to `aws_db_instance`. When deleting the instance this generates a snapshot named `<final_snapshot_identifier>-<timestamp>`

Example: generates a snapshot called mydb-20170210-083019

    resource "aws_db_instance" "default" {
      allocated_storage    = 10
      engine               = "mysql"
      instance_class       = "db.t1.micro"
      name                 = "mydb"
      username             = "foofoofoo"
      password             = "barbarbar"
      db_subnet_group_name = "mydb_subnet_group"
      final_snapshot_identifier = "mydb"
      timestamp_final_snapshot = "20060102-150405"
    }

My team have been encountering the following error intermittently the past couple of weeks:
    4 error(s) occurred:
    
    * aws_db_instance.default: DBSnapshotAlreadyExists: Cannot create the snapshot because a snapshot with the identifier finalsnapshotcifull-service-mysql already exists.
        status code: 400, request id: 456a579a-ee6a-11e6-a5b8-cdbcbd0418d6

To help mitigate the issue we had been appending a `md5(timestamp())` to the final_snapshot_identifier parameter of aws_db_instance, but this was causing some less than optimal behavior when running terraform apply against the same environment repeatedly, not to mention the fact that the snapshot name wasn't particularly readable.

To mitigate the issue I decided it would be useful to have a real timestamp appended to the snapshot identifier. It uses a [golang time Format](https://golang.org/pkg/time/#Time.Format) string to determine the formatting.


gdelgado:
+1...

PR #11864, opened 2017-02-10
config: Resource.Count should type check

### LINKED ISSUE #11800: TF 0.8.6 crash -- panic: when `count` doesn't have a value of integer vs list...
Reporting TF crash after making route53 alterations. If you need HCLs, please let me know. The crash eludes to something about interfaces, but I made no changes to any interfaces in my HCLs as far as I'm aware.

### Terraform Version
0.8.6

### Affected Resource(s)
It's not very clear what the exact resource is, but the provider is aws.

If you can determine which HCLs you need from the crash, please let me know. I'd rather not send my entire code base as it would take me a while to make sure they're sufficiently scrubbed, and I don't have a lot of time to do so unless it's necessary.

### Debug Output
No debug output is currently available. See crash.log.

### Panic Output
https://gist.github.com/notjames/0c165f9874d2e8bd91a04cc6b661c445

### Expected Behavior
No crash should have occurred... :)

### Actual Behavior
A crash occurred.

### Steps to Reproduce
1. `terraform plan -var azones=["us-west-2a","us-west-2b","us-west-2c"] -var region=us-west-2 -var environment=test -var private-key=$HOME/.ssh/redaptcust-forge.pem -var privkey-name=keyname.pem -var privkey-path=$HOME/.ssh -var ssh_key_name=key_name -var aws-id=ID_WAS_HERE -var aws-secret=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx -var aws-creds-profile=default`
2. same for apply and refresh

### Important Factoids
No, other than I run the above TF command from a wrapper script to develop the dynamic bits required for my HCLs to run properly. The above reflects the baked command that the wrapper executes.

### References
No.

mitchellh:
Fixes #11800

Type check the value of count so we don't panic on the conversion.

I wondered "why didn't we do this before?" There is no excuse for NOT
doing it at all but the reasoning was beacuse prior to the list/map work
in 0.7, the value couldn't be anything other than a string since any
primitive can turn into a string.

Regardless, we should've always done this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -212,7 +212,14 @@ func (r *Module) Id() string {
 
 // Count returns the count of this resource.
 func (r *Resource) Count() (int, error) {
-	v, err := strconv.ParseInt(r.RawCount.Value().(string), 0, 0)
+	raw := r.RawCount.Value()
+	count, ok := r.RawCount.Value().(string)
+	if !ok {
+		return 0, fmt.Errorf(
+			"expected count to be a string or int, got %T", raw)
+	}
+
+	v, err := strconv.ParseInt(count, 0, 0)
 	if err != nil {
 		return 0, err
 	}
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/hashicorp/hil/ast"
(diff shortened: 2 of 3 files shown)


jbardin:
Thanks, I've been meaning to do this for a while ;)

PR #11871, opened 2017-02-10
Clarify assn of App Registration with subscription

elijah:
In my first attempt at using this provider, I did not understand that this was a step that applied regardless of whether one used the ARM portal or the classic portal.  I fixed the grammar and added words to make this clearer.  Maybe it should be a section instead of a thing that could be mistaken for the tail of the classic portal setup instructions?


tombuildsstuff:
Hi @elijah

Thanks for your contribution! Apologies for the delay in reviewing this.

This Guide was first written whilst Azure Resource Manager/the New Portal was under development - and so references functionality in both the New and the Old Portal. We're now at a point where the New Portal is now feature-complete - allowing us to remove references to the Old Portal altogether.

I've been working through separately to this PR to review the Guide, with the intention of simplifying it - and hope to have something to show in the near-future.

So that we can consolidate our efforts down here, I'm going to close this PR for the moment - I'd like to thank you for taking the time to submit this PR.

Thanks!

Issue #6463, opened 2016-05-03
Cannot create api-gateway integration with stage variable for lambda function name


JefStat:
### Terraform Version

Terraform v0.6.15
### Affected Resource(s)
- aws_api_gateway_integration
### Terraform Configuration Files

https://gist.github.com/JefStat/9cb343c0ee66366aa060e7db55fbcef6
### Debug Output

https://gist.github.com/JefStat/9ffa536e973066c67e627d7e77aa0ac9
### Expected Behavior

Should update lambda function
### Actual Behavior

Error invalid lambda function
### Steps to Reproduce
1. `terraform apply`

radeksimko:
I haven't personally used stage variables (yet), I'm wondering if API Gateway deployment is eventually consistent ...

I plan to work on `aws_api_gateway_stage` resource so I might have a look at this bug too.

dkade:
Any new on this? I think is mandatory when using Lambda, or you would have to create different API for different stages.

Bschuster3434:
Is there a time frame on integrating this feature into the solution?

neil-power-hs:
I am also running into this issue. I would like to create one deployment per lambda alias but I am unable to reference the stage variables in the uri in the aws_api_gateway_integration

gnp:
I'm using 0.9.2 and its working for me like this:

```
resource "aws_api_gateway_integration" "my-integration" {
...
  uri = "arn:aws:apigateway:${var.region}:lambda:path/2015-03-31/functions/${aws_lambda_function.my-lambda.arn}:$${stageVariables.lambdaAlias}/invocations"
...
}
```

neil-power-hs:
@gnp how do you avoid the circular dependency? My deployment depends on the integration
```
resource "aws_api_gateway_deployment" "query_builder_deployment" {
  depends_on = [
    "aws_api_gateway_method.query_builder_post",
    "aws_api_gateway_integration.query_builder_integration",
    "aws_api_gateway_integration_response.query_builder_integration_response",
    "aws_api_gateway_method_response.query_builder_post_200"
  ]
...
```

neil-power-hs:
oh I see, you're not referencing the variable directly so that avoids it, would you mind posting your aws_api_gateway_deployment block?

gnp:
```
resource "aws_api_gateway_deployment" "test" {
  depends_on = ["aws_api_gateway_integration.my-integration"]

[...]

PR #11877, opened 2017-02-11
Address #6463 to allow :${stageVariables.___} suffix in function_name

gnp:
I set up a dev env per instrux but when I added the tests in this PR and tried to see if they would fail, other unrelated stuff was failing. I would have expected the qualifier tests to fail because I didn't change any regex related to the separate qualifier property. Seems like both qualifier and function_name suffix should allow ${stageVariables.___} format.

I was hoping to build locally and use this fix on my real use case but since the tests don't work reliably here I cannot trust the local build.

Anyone willing to collaborate on making this a real fix and getting it into master?

### DIFF
--- a/builtin/providers/google/resource_container_cluster_test.go
+++ b/builtin/providers/google/resource_container_cluster_test.go
@@ -215,7 +215,7 @@ var testAccContainerCluster_withVersion = fmt.Sprintf(`
 resource "google_container_cluster" "with_version" {
 	name = "cluster-test-%s"
 	zone = "us-central1-a"
-	node_version = "1.4.7"
+	node_version = "1.5.2"
 	initial_node_count = 1
 
 	master_auth {


gnp:
I'm trying to work around this limitation of Terraform, and in doing so, I'm not certain that making this change to the Permission side of things is actually right. I started off trying to fix the function_name parsing in aws_api_gateway_integration to allow me to put the ":${stageVariables.lambdaAlias}" I need and can do from the AWS console. Then I found this permission stuff and thought that might be it. I just made my change in the console and discovered the permissions resources do need to be set for each possible value of the stage variable (TEST, PROD in my case) NOT to allow the variable itself.

I think the regex change I propose needs to be applied to aws_api_gateway_integration instead. I'm happy to edit this pull request to back out the permissions changes in favor of the integration change if I can get someone to partner on testing...

Issue #11838, opened 2017-02-09
Unable to use a tag that starts with aws: in a datasource filter


Dchamard:
### Terraform Version
0.8.4

### Affected Resource(s)

- aws_security_group (datasource)

It is propably affecting most of the aws datasources as well.

### Terraform Configuration Files
```hcl
data "aws_security_group" "beanstalk_sg" {
  count = "${var.beanstalk_sg_manage_enabled}"
  tags = {
     "aws:cloudformation:logical-id"     =  "AWSEBLoadBalancerSecurityGroup"
     "Name"                               = "${var.stack_name}"
     "service"                           =  "${var.stack_name}-beanstalk"
  }
}
```

### Debug Output
2017/02/09 17:14:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/02/09 17:14:25 [DEBUG] Matching ^aws:* with Name
2017/02/09 17:14:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/02/09 17:14:25 [DEBUG] Matching ^aws:* with service
2017/02/09 17:14:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/02/09 17:14:25 [DEBUG] Matching ^aws:* with aws:cloudformation:logical-id
**2017/02/09 17:14:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/02/09 17:14:25 [DEBUG] Found AWS specific tag aws:cloudformation:logical-id (val: AWSEBLoadBalancerSecurityGroup), ignoring.**
2017/02/09 17:14:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/02/09 17:14:25 [DEBUG] Describe Security Groups {

### Expected Behavior
I should have been abled to use this tag to find the proper resource for my data source

### Actual Behavior
It is ignoring the tag because it starts with aws:
This was introduced in the following issue.
https://github.com/hashicorp/terraform/issues/4513

I understand that those tags cannot be modifed because its used internally by aws but we should be able to use them as data source to find resources.

The data source is using the same  'tagsFromMap' generic method that calls the 'tagIgnored' method that will filter tags that start with aws:

### References
https://github.com/hashicorp/terraform/issues/4513

ericwestfall:
I have submitted a PR that provides the ability for data sources to opt-out of the internal AWS tag filtering. I've confirmed the bugfix resolves your issue.

**Sample Terraform**
```
data "aws_security_group" "selected" {
[...]

PR #11885, opened 2017-02-11
provider/aws: Provide ability to opt-out of internal AWS tag filtering

ericwestfall:
See the following issue: #11838

In PR #7454, functionality was introduced to filter and ignore any AWS internally reserved tags (e.g. tags that begin with `aws:`). This functionality is integrated directly into various functions, including the shared `tagsToMap()` and `tagsFromMap()` functions which are called by quite a few resources and data sources. 

While we cannot modify these AWS reserved tags, we absolutely should be able to reference them for read-only operations. 

The proposed solution here is to add a boolean flag, `allowInternalTags`, to the common `tagsFromMap` and `tagsToMap` functions. This flag will allow callers to decide how they want to process AWS reserved tags. 

All standard resources that call the common `tagsFromMap` and `tagsToMap` functions will pass `false` and retain the normal filtering behavior where any tag that begins with `aws:` will be ignored. 

Data sources can pass `true` to opt-out of the filtering and thereby gain the ability to reference AWS reserved tags. 

It is worth noting, that quite a few resources contain independent implementations of the `tagsFromMap`, `tagsToMap` and `tagIgnored` functionality - [here is an example](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/tagsEC.go#L105) of the ElastiCache implementation. This change does not presently modify the behavior of resources that have their own internal implementation, only the common `tagsFromMap` and `tagsToMap` functions in [tags.go](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/tags.go).

### DIFF
--- a/builtin/providers/aws/data_source_aws_instance.go
+++ b/builtin/providers/aws/data_source_aws_instance.go
@@ -225,7 +225,7 @@ func dataSourceAwsInstanceRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	if tagsOk {
 		params.Filters = append(params.Filters, buildEC2TagFilterList(
-			tagsFromMap(tags.(map[string]interface{})),
+			tagsFromMap(tags.(map[string]interface{}), true),
 		)...)
 	}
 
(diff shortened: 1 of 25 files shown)


ericwestfall:
@stack72 just wanted to see if there is anything else I need to do with this before it can be reviewed or if perhaps this implementation approach is completely off the rails. Cheers.

ericwestfall:
@stack72, I'm closing this PR in preference of #14116.

Issue #9690, opened 2016-10-28
aws_ecs_service/desired_count must be optional


BrunoBonacci:
The `desired_count` in the `aws_ecs_service` appears to be required. However the AWS api mandate this parameter only for the _creation_ of a service. For the update part of a service the `desired_count` must be optional.
- CreateService http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html
- UpdateService http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html

The reason being is that the `desired_count` is subject to automatic modification via the auto scaling policies. Mandate this value causes the service to _overwrite_ the value which was set by the auto-scaling policies.

Assume the case that you have a service which is created with a `desired_count = 3`. Then thanks to the auto scaling policies the service is scaled out to a `desired_count = 10`. At this point the terraform script still contains `desired_count = 3` and if you were to apply the changes the `desired_count` would be _forcefully reset_ to its initial value of 3.

```
~ aws_ecs_service.ecs_service
    desired_count: "10" => "3"
```

And if I try to remove the `desired_count = 3` from the script then it will set it to 0.

```
~ aws_ecs_service.ecs_service
    desired_count: "10" => "0"
```

By forcefully setting a lower number of `desired_count` it will tell the ECS service to KILL some of the excess containers to meet the desired number.
This would obviously cause some service disruption as suddenly big part of the container fleet will disappear.

The `desired_count` is required for manual scaling, but it is harmful/dangerous to use in conjunction to autoscaling. Like for the EC2 autoscaling should be entirely optional to use this value.
The solution suggested Is to introduce a new option for the `aws_ecs_service` called `initial_count` which will be passed as `desired_count` only when the resource is created and never during an update.

This is currently a big problem for those who want to use ECS in conjunction with the autoscaling policies as there is no alternative solution.
### Terraform Version
- Terraform v0.7.7
### Affected Resource(s)

Please list the resources as a list, for example:
- aws_ecs_service
### Terraform Configuration Files

``` hcl
resource "aws_ecs_service" "mongo" {
  name = "mongodb"
  cluster = "${aws_ecs_cluster.foo.id}"
  task_definition = "${aws_ecs_task_definition.mongo.arn}"

  desired_count = 3

  iam_role = "${aws_iam_role.foo.arn}"
  depends_on = ["aws_iam_role_policy.foo"]

  load_balancer {
    elb_name = "${aws_elb.foo.name}"
[...]

PR #11892, opened 2017-02-12
[WIP] - Modify aws_ecs_service to prevent destroying tasks

nhuray:
Hi everyone,

This is a Pull request to fix an annoying issue when using `aws_ecs_service` described by @BrunoBonacci here #9690.

The idea is to used `desired_count` to create or update the number of ECS service tasks, but the `desired_count` is not update in the state when reading from the real-infrastructure (during `refresh`).

WARNING: I didn't add tests because for some reasons the tests already fails in my laptop (branch `0-8-stable`)

### DIFF
--- a/builtin/providers/aws/resource_aws_ecs_service.go
+++ b/builtin/providers/aws/resource_aws_ecs_service.go
@@ -293,8 +293,6 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("task_definition", taskDefinition)
 	}
 
-	d.Set("desired_count", service.DesiredCount)
-
 	// Save cluster in the same format
 	if strings.HasPrefix(d.Get("cluster").(string), "arn:"+meta.(*AWSClient).partition+":ecs:") {
 		d.Set("cluster", service.ClusterArn)


nhuray:
Just a thought, we might do the same thing for `deployment_minimum_healthy_percent` and `deployment_maximum_percent` ...

BrunoBonacci:
@nhuray what's the reason why you would threat `deployment_minimum_healthy_percent` and `deployment_maximum_percent` in the same way? This fields aren't updated automatically like the `desired_count`, they are typically updated manually to control the deployment strategy (canary, blue/green, singleton etc)

nhuray:
That's true ... actually the issue I have with this parameters is much more to be able to calculate what I want in term of deployment strategy regarding the environment I deploy to...

For example on a `dev` environment with 2 services running the `deployment_minimum_healthy_percent ` should be 50% to have a strategy `once at a time`.

On a `prod` environment with 10 services running the `deployment_minimum_healthy_percent ` should be 10% to have a strategy `once at a time`.

For sure it might be calculated but it will be better to have something like the placement strategy and placement constraints AWS has defined recently...

So finally it's not a Terraform issue, much more an AWS issue for me !

stack72:
Hi @jmahowald 

Please can you reopen this against master? We have just killed off the 0-8-stable branch

Thanks

Paul

stack72:
@jmahowald apologies again :)

s-maj:
Hi @nhuray 
Would be possible to merge this against master? This change sounds really useful.

nhuray:
Hi @s-maj @stack72,

I recreated the pull request in the `terraform-providers` organization: https://github.com/terraform-providers/terraform-provider-aws/pull/1028

PR #11895, opened 2017-02-12
google cloud: regression - enable reading of existing project on create

zachgersh:
This is a regression - project read worked in version 0.8.4. Attempted to fix the regression in as minimal of a way as possible.

- docs say that existing projects can be used but a user
would never be able to use an existing project.
- we erroneously expected the user to set both name
and org_id when they only need to set project_id

Interesting note, the API doesn't fail if you look up a project that does not exist - this would have influenced by error check.

### DIFF
--- a/builtin/providers/google/resource_google_project.go
+++ b/builtin/providers/google/resource_google_project.go
@@ -93,52 +93,57 @@ func resourceGoogleProject() *schema.Resource {
 func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
-	var pid string
-	var err error
-	pid = d.Get("project_id").(string)
-	if pid == "" {
-		pid, err = getProject(d, config)
-		if err != nil {
-			return fmt.Errorf("Error getting project ID: %v", err)
-		}
-		if pid == "" {
-			return fmt.Errorf("'project_id' must be set in the config")
-		}
-	}
-
-	// we need to check if name and org_id are set, and throw an error if they aren't
-	// we can't just set these as required on the object, however, as that would break
-	// all configs that used previous iterations of the resource.
-	// TODO(paddy): remove this for 0.9 and set these attributes as required.
-	name, org_id := d.Get("name").(string), d.Get("org_id").(string)
-	if name == "" {
-		return fmt.Errorf("`name` must be set in the config if you're creating a project.")
-	}
-	if org_id == "" {
-		return fmt.Errorf("`org_id` must be set in the config if you're creating a project.")
-	}
-
-	log.Printf("[DEBUG]: Creating new project %q", pid)
-	project := &cloudresourcemanager.Project{
-		ProjectId: pid,
-		Name:      d.Get("name").(string),
-		Parent: &cloudresourcemanager.ResourceId{
-			Id:   d.Get("org_id").(string),
-			Type: "organization",
-		},
-	}
-
-	op, err := config.clientResourceManager.Projects.Create(project).Do()
+	pid, err := getProject(d, config)
 	if err != nil {
-		return fmt.Errorf("Error creating project %s (%s): %s.", project.ProjectId, project.Name, err)
+		return fmt.Errorf("Error getting project ID: %v", err)
 	}
 
-	d.SetId(pid)
+	if pid == "" {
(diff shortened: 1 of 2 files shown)


zachgersh:
cc @evandbrown / @paddyforan - I'd be interested to know what you all think of this regression.

paddyforan:
Thanks for the PR @zachgersh! I'll take a look at this tomorrow.

zachgersh:
Thanks @paddyforan - happy to make any changes if needed.

paddyforan:
Hey @zachgersh!

So after taking a look at the code for this, I'd love to just chat about the rationale behind this change, if that's ok. You mention reading projects that didn't have create called. The current implementation does not permit that, you're right, but should continue to work for projects that had previously been read without Create being called; it just won't allow new projects to be read that way.

Can you tell me a bit about your use-case for reading the project in but not having Terraform create it? Do you want Terraform to manage the project? If so, is there a reason `terraform import` doesn't work? Do you want Terraform to not have write access to the project? If so, what properties are you looking to read? I was going to implement a project data source, but when I went to do it, I realised the only information I'd ever want to pull out would be the project ID, and that's what you have to put in the config file to reference the project, anyways, so I didn't really see a point. Do you have a use case I missed?

Thanks!

zachgersh:
@paddyforan, really appreciate the well thought out questions now that we've had a couple of times to interact - keep up the great works.

So, rationale - we have a base project that has been setup outside of terraform and is managed by an admin team (they are the only ones with the power to create projects). I could go and import the project with `terraform import` but that would need to be done outside of my normal flow of terraforming up the rest of my infrastructure. 

The end goal of reading the project in is so that we have the ability to start attaching IAM policy data to it - in 0.8.4 where reading a project worked we would load our project in the normal `terraform apply` flow and then move on to attaching policy data to it. I'd like to be able to do that same thing without the extra step of figuring out when to call `terraform import`

Hopefully that makes sense?

paddyforan:
Hey @zachgersh! Thanks for the kind words. 😊 That sounds like a straightforward-enough thing to want to do, thanks for elaborating.

Is there a reason [`google_project_iam_policy`](https://www.terraform.io/docs/providers/google/r/google_project_iam_policy.html) doesn't work for your use case?

zachgersh:
I think that can work - I hadn't noticed that IAM policy does take a `projec_name` I can move to that and then upgrade to latest and see how we get along. It may be nice to merge this just so this regression doesn't stay in place till you can phase it out?

I could also just update the docs to note that this will ALWAYS create a new project (and add your deets about importing). I think the added clarity will be nice for all of us.

paddyforan:
Hey @zachgersh! Let me know how that works for you. That's definitely the recommended way to manage IAM policies on a project right now.

As for merging this, maybe some background would help. This originally arose from #10425, where we attempted to correct some abnormal and surprising behaviour from the google_project resource. Typically, any resource in Terraform is owned by Terraform, and Terraform creates and manages that resource. A resource like pre-0.8.5 google_project was more properly a data source, but wasn't implemented as one.

In aiming to correct this, we took a lot of pains to ensure that the _already existing_ configs/state of users that used the resource would remain intact and continue working as if nothing had changed. We decided, however, that any _new_ state would exhibit the new, Terraform-y behaviour of Terraform creating and managing the resource. This was our attempt at achieving least surprise for everyone involved.

So at the moment, I'm hesitant to merge this, just because it adds _more_ surprise to the mix (the resource behaves one way in < 0.8.5, another in 0.8.5, 0.8.6, and 0.8.7, and then back to the original way in 0.8.8 >, and then back to the second way in > 0.9. That seems really confusing to me, and I'm unclear on the value.

With regards to documentation, we did [call out](https://groups.google.com/d/msg/hashicorp-announce/uJQUdQ8Ellw/eG1Hq29uBQAJ) the change in the release notes for 0.8.5, but looking at the docs, I can totally agree we can do a better job of messaging the change. Unless you have a compelling argument for reverting #10425 and want to discuss this some more, I'll go ahead and open a PR to add better docs to the google_project resource's page on the site, documenting the behaviour a bit better.

Does that sound like a reasonable way to proceed?

zachgersh:
@paddyforan totally agree - I am going to close this out in favor of just a docs change. CC me if you'd like a review!

PR #11896, opened 2017-02-12
provider/fastly Documentation addition for fastly healthchecks

travertischio:
Healthchecks are already a part of version `0.8.6`

There is already documentation on the implementation details of the healthcheck resource, but the resource was not mentioned in the main section of the fastly cache description. This fixes that discrepancy.


stack72:
Thanks for this - pushing to the site now!

PR #11917, opened 2017-02-13
Add a realistic copy_options parameter

kevinburke:
It looks like the copy_options value was fat fingered from the
compression_format parameter - I don't believe that GZIP is a valid value for
copy_options, at least based on the documentation.

Adds a link to the documentation and adds a more realistic (and harmless) value
for the copy_options parameter.


radeksimko:
Hi @kevinburke 
thanks for the patch here, I'm also having troubles understanding the relevant docs, but it is possible that `GZIP` is actually valid value:
http://docs.aws.amazon.com/redshift/latest/dg/r_COPY_command_examples.html#r_COPY_command_examples-load-time-from-a-pipe-delimited-gzip-file

Nonetheless `delimiter '|'` is certainly a valid value and the link to docs is helpful.

PR #11920, opened 2017-02-13
Added `aws_iam_role' to list of importable AWS resources.

jaytaylor:
`aws_iam_role` isn't listed as importable in the docs but after inspecting the terraform source code it looked like it should work so I tried it out- and it works great!


stack72:
LGTM! Thanks for this :)

PR #11922, opened 2017-02-13
command: add TF_CLI_ARGS to specify additional CLI args

mitchellh:
This introduces an env var `TF_CLI_ARGS` that can be used to specify additional CLI args. 

The primary use for this is easier CI and automation integration.

The flag is parsed like a shell line, so the value `TF_CLI_ARGS="-foo 'bar baz'"` will turn into `["-foo", "bar baz"]` internally. 

The additional args are appended _before_ any args specified on the CLI _after_ the command. So, if you invoke `TF_CLI_ARGS="-force" terraform apply -input=false` then it is equivalent to specifying `terraform apply -force -input=false`. This ordering forces the CLI args to take precedent over the env var.

Updating the docs now.

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"runtime"
+	"strings"
 	"sync"
 
 	"github.com/hashicorp/go-plugin"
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -129,9 +135,45 @@ func wrappedMain() int {
...
+				"Error parsing extra CLI args from %s: %s",
+				EnvCLI, err))
+			return 1
+		}
+
+		// Find the index to place the flags. We put them exactly
+		// after the first non-flag arg.
+		idx := -1
+		for i, v := range args {
+			if v[0] != '-' {
```
jbardin: An arg can be 0 length -- even more likely in an automated system like TFE. 
`strings.HasPrefix`?

main.go
```diff
@@ -129,9 +135,45 @@ func wrappedMain() int {
...
+			}
+		}
+
+		// idx points to the exact arg that isn't a flag. We increment
+		// by one so that all the copying below expects idx to be the
+		// insertion point.
+		idx++
+
+		// Copy the args
+		newArgs := make([]string, len(args)+len(extra))
```
jbardin: Do we need to worry about non-flag arguments in the `TF_CLI_ARGS` variable? ( I can't think of a case where it matters, just tossing out the question)

mitchellh: I think we just make explicit that we insert the args right after the command. If that works, great, if that doesn't for them for normal args, then its documented.

main.go
mitchellh: Really good catch, really good. Fixing now.

PR #11942, opened 2017-02-14
Add resource_openstack_images_image_v2

yanndegat:


builtin/providers/openstack/resource_openstack_image_v2_test.go
```diff
@@ -0,0 +1,85 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/rackspace/gophercloud/openstack/imageservice/v2/images"
```
jtopjian: `s/rackspace/gophercloud/` :)

builtin/providers/openstack/provider.go
```diff
@@ -147,6 +147,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_fw_firewall_v1":                 resourceFWFirewallV1(),
 			"openstack_fw_policy_v1":                   resourceFWPolicyV1(),
 			"openstack_fw_rule_v1":                     resourceFWRuleV1(),
+			"openstack_image_v2":                       resourceImageV2(),
```
jtopjian: I started working on an Image data source and waffled over what the resource name should be. Currently, the OpenStack resources are named with the following format:

```
openstack_<service>_<resource>_<service api version>
```

With that in mind, I was going to use the following for the Image data source:

```
openstack_images_image_v2
```

`images_image` looks a little goofy, but it makes sense in the context of something like:

```
openstack_images_imagedata_v2
openstack_images_artifact_v2
```

So I'd recommend changing the name to `openstack_images_image_v2` here. Thoughts?

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"fmt"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/imagedata"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
```
jtopjian: nit: Check the other resource for how imports are ordered. Standard library imports are ordered first, then a blank line, then github/gophercloud, blank link, then github/hashicorp

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+				Required: true,
+				ForceNew: false,
+			},
+
+			"id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
```
jtopjian: This should be removed since `id` shouldn't be a visible attribute. `d.SetId(image.ID)` effectively sets the resource's "internal" id.

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	if images.ImageVisibility(v) == images.ImageVisibilityPublic {
+		return images.ImageVisibilityPublic, nil
+	}
+	if images.ImageVisibility(v) == images.ImageVisibilityShared {
+		return images.ImageVisibilityShared, nil
+	}
+
+	return "", fmt.Errorf("Error unkown image visibility %q", v)
+}
```
jtopjian: nit: I think a `switch` statement might be a better approach. See [here](https://github.com/jtopjian/terraform/blob/openstack-data-source-image/builtin/providers/openstack/data_source_openstack_images_image.go#L239) for an example.

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+
+func resourceImageV2Create(d *schema.ResourceData, meta interface{}) error {
```
jtopjian: nit: To better align with the other OpenStack resources, the function name should be something like `resourceImagesImageV2Create`

Also, Create, Read, Update, and Delete should follow the schema. Supporting functions (validation and such) should follow underneath.

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack image client: %s", err)
+	}
+
+	img, err := images.Get(imageClient, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "image")
+	}
+
+	log.Printf("[DEBUG] Retrieved Image %s: %+v", d.Id(), img)
```
jtopjian: nit: Can you use `%#v` here to align with the other resources?

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+
+			"metadata": &schema.Schema{
+				Type:     schema.TypeMap,
+				Computed: true,
+			},
+
+			"created_date": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
jtopjian: I think this should be called `created_at` to match the Glance API.

builtin/providers/openstack/resource_openstack_image_v2.go
```diff
@@ -0,0 +1,494 @@
...
+
+			"last_update": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
jtopjian: I think this should be called `updated_at` to match the Glance API.

jtopjian:
@yanndegat Also, I started work on an Image Data Source which you can see [here](https://github.com/jtopjian/terraform/blob/openstack-data-source-image/builtin/providers/openstack/data_source_openstack_images_image.go). It's currently blocked by a Gophercloud bug. We'll have to make sure our attribute names match for consistency. Since you've opened this PR before I was able to open the PR for the data source, we'll use the result of this as the source of truth and I'll modify the data source accordingly.

yanndegat:
thanks a lot for your reactivity and your review,

i've pushed the changes you requested. i've also updated the doc and sorted the attributes alphabetically (either in the docs & in the source), hoping i'm not wrong about this.

yanndegat:
the travis check that failed is a go1.8rc3 build which sigquit. don't know why

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+		Protected:       &protected,
+	}
+
+	if v := d.Get("visibility").(string); v != "" {
+		vis := resourceImagesImageV2VisibitlityFromString(v)
+		if vis == "" {
+			return fmt.Errorf("Error unkown image visibility %q", v)
+		}
+		createOpts.Visibility = &vis
+	}
```
jtopjian: A few notes about this one:

You can safely omit this entire block because you provided a default value in the schema. Since there is a default value, `v` will never be an empty string.

Instead, just add this to the above `CreateOpts` like so:

```
Visibility: resourceImagesImageV2VisibilityFromString(d.Get("visibility").(string))
```

Or, add this above `CreateOpts`:

```
visibility := resourceImagesImageV2VisibilityFromString(d.Get("visibility").(string))
createOpts := &images.CreateOpts{
    ...
    Visibility: visibility,
}
```

Also, there's no need to return an error if the string is empty. This is for two reasons:

1. You provided a `ValidateFunc`, so an empty string will never be valid and this part of the code will never be reached.
2. In [Gophercloud](https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/requests.go#L75) the `omitempty` means to ignore an empty string.

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+		} else {
+			log.Printf("[DEBUG] File exists %s", filename)
+			return filename, nil
+		}
+	} else {
+		return "", fmt.Errorf("Error in config. no file specified")
+	}
+}
+
+func resourceImagesImageV2Read(d *schema.ResourceData, meta interface{}) error {
```
jtopjian: This function should be placed right after `resourceImagesImageV2Create`

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+	config := meta.(*Config)
+	imageClient, err := config.imageV2Client(d.Get("region").(string))
```
jtopjian: ```imageClient, err := config.imageV2Client(GetRegion(d))```

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+	d.Set("size_bytes", img.SizeBytes)
+	d.Set("metadata", img.Metadata)
+	d.Set("created_at", img.CreatedAt)
+	d.Set("update_at", img.UpdatedAt)
+	d.Set("id", img.ID)
+
+	return nil
+}
+
+func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error {
```
jtopjian: This function is correctly placed below `resourceImagesImageV2Read`, but make sure to move this up when moving `Read`.

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+
+	_, err = images.Update(imageClient, d.Id(), updateOpts).Extract()
+	if err != nil {
+		return fmt.Errorf("Error updating image: %s", err)
+	}
+
+	return resourceImagesImageV2Read(d, meta)
+}
+
+func resourceImagesImageV2Delete(d *schema.ResourceData, meta interface{}) error {
```
jtopjian: This function is correctly placed below `resourceImagesImageV2Update`, but make sure to move this up when moving `Update`.

The rest of the functions (validation, uploading, etc) are then placed below `Delete`. 

Other Terraform resources might have a different order, but this has been the common pattern for the OpenStack resources.

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+	d.Set("created_at", img.CreatedAt)
+	d.Set("update_at", img.UpdatedAt)
+	d.Set("id", img.ID)
+
+	return nil
+}
+
+func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	imageClient, err := config.imageV2Client(d.Get("region").(string))
```
jtopjian: ```imageClient, err := config.imageV2Client(GetRegion(d))```

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+	if err != nil {
+		return fmt.Errorf("Error updating image: %s", err)
+	}
+
+	return resourceImagesImageV2Read(d, meta)
+}
+
+func resourceImagesImageV2Delete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	imageClient, err := config.imageV2Client(d.Get("region").(string))
```
jtopjian: ```imageClient, err := config.imageV2Client(GetRegion(d))```

builtin/providers/openstack/resource_openstack_images_image_v2_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+var testAccImagesImageV2_basic = `
+  resource "openstack_images_image_v2" "foo" {
+      name   = "Rancher TerraformAccTest"
+      image_source_url = "https://releases.rancher.com/os/latest/rancheros-openstack.img"
+      container_format = "bare"
+      disk_format = "qcow2"
+  }`
```
jtopjian: If you have time, maybe add a test using "tags" and another that uses "properties". Totally optional, though. I can always add those later.

yanndegat: added tags acc test but not properties. It seems that providers can disable this attribute on the API, and the provider i'm using seems to have done so. The test i've tried on properties fails and i don't know how to handle it so i've deleted it. let me know if you see a way to test it correctly

jtopjian: @yanndegat Not you at all. It looks like there's a [bug](https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/requests.go#L102) in Gophercloud that is not encoding the map to json correctly.

Second, there is no [properties](https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/results.go#L66) attribute in the response JSON. The properties are mixed in with all of the other keys. Ugh.

So those two things need fixed in order for properties to work.

Since neither creating nor reading properties works at the moment, how do you feel about just removing `properties` entirely from this resource until it is fixed upstream?

builtin/providers/openstack/resource_openstack_images_image_v2.go
```diff
@@ -0,0 +1,488 @@
...
+	case "community":
+		return images.ImageVisibilityCommunity
+	}
+
+	return ""
+}
+
+func resourceImagesImageV2Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	imageClient, err := config.imageV2Client(d.Get("region").(string))
```
jtopjian: Don't forget this one :)

builtin/providers/openstack/resource_openstack_images_image_v2_test.go
yanndegat: @jtopjian done

jtopjian:
@yanndegat Thank you for all the work you've put into this and for contributing it. This looks great!

yanndegat:
welli needed it ;)

 thank you for the review

PR #11952, opened 2017-02-15
provider/aws: add bucket name to delete error notification

netjunki:
* aws_s3_bucket.roster_archive: Error deleting S3 Bucket: BucketNotEmpty: The bucket you tried to delete is not empty
	    status code: 409, request id: A439DD5AEF8E8834

Which bucket is trying to be deleted? New message should be:

    * aws_s3_bucket.roster_archive: Error deleting S3 Bucket: BucketNotEmpty: The bucket you tried to delete is not empty 'mybucket'
	    status code: 409, request id: A439DD5AEF8E8834

### DIFF
--- a/builtin/providers/aws/resource_aws_s3_bucket.go
+++ b/builtin/providers/aws/resource_aws_s3_bucket.go
@@ -980,7 +980,7 @@ func resourceAwsS3BucketDelete(d *schema.ResourceData, meta interface{}) error {
 				return resourceAwsS3BucketDelete(d, meta)
 			}
 		}
-		return fmt.Errorf("Error deleting S3 Bucket: %s", err)
+		return fmt.Errorf("Error deleting S3 Bucket: %s '%s'", err, d.Get("bucket"))
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_s3_bucket.go
```diff
@@ -980,7 +980,7 @@ func resourceAwsS3BucketDelete(d *schema.ResourceData, meta interface{}) error {
 				return resourceAwsS3BucketDelete(d, meta)
 			}
 		}
-		return fmt.Errorf("Error deleting S3 Bucket: %s", err)
+		return fmt.Errorf("Error deleting S3 Bucket: %s '%s'", err, d.Get("bucket"))
```
stack72: this needs to be d.Get("bucket").(string) - it needs to have the cast

Also, instead of '%s' you can use %q

netjunki:
@stack72 how's that?

stack72:
LGTM! Thanks for making the changes here :)

PR #11953, opened 2017-02-15
the invalid variable syntax error message was misleading. changed tex…

netjunki:
…t a little bit
From:

    * invalid variable syntax: "image". If this is part of inline `template` parameter then you must escape the interpolation with two dollar signs. For example: ${a} becomes $${a}.

to:

    * invalid variable syntax: "image". Did you mean 'var.image'? If this is part of inline `template` parameter then you must escape the interpolation with two dollar signs. For example: ${a} becomes $${a}.

### DIFF
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -303,10 +303,10 @@ func (i *Interpolater) valueSimpleVar(
 	// relied on this for their template_file data sources. We should
 	// remove this at some point but there isn't any rush.
 	return fmt.Errorf(
-		"invalid variable syntax: %q. If this is part of inline `template` parameter\n"+
+		"invalid variable syntax: %q. Did you mean 'var.%s'? If this is part of inline `template` parameter\n"+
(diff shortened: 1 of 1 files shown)


catsby:
Hey @netjunki can you give us some context on what would trigger this? I'd like to verify that suggesting `var.%s` is the most likely scenario to address here

manbearwiz:
I have run into this a number of times when I wrote `foo = "${bar}"` when the correct syntax would be `    foo = "${var.bar}"`. In this case, the error message about an "inline template parameter" was misleading.

netjunki:
@catsby the case which @manbearwiz mentioned is the exact issue I was facing (we work together and both encountered this about a week apart).

netjunki:
@catsby can we wrap this one up? any other suggestions?

netjunki:
@catsby Anything else you need here?

catsby:
Thanks!

PR #11954, opened 2017-02-15
provider/google: Add google_compute_zones data source

radeksimko:
I wanted to build a GKE cluster and I saw hard-coded zones in the example which is what prompted me to write this data source in the 1st place.

There are some HCL changes that would help making this more elegant/useful though, specifically slicing lists via colon, e.g.

```hcl
zone = "${data.google_compute_zones.available.names[0]}"
additional_zones = [
    "${data.google_compute_zones.available.names[1:]}"
]
```

--------

### Test plan

```
make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccGoogleComputeZones_basic'
2017/02/15 07:33:15 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccGoogleComputeZones_basic -timeout 120m
=== RUN   TestAccGoogleComputeZones_basic
--- PASS: TestAccGoogleComputeZones_basic (3.05s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	3.079s
```

### DIFF
--- a/builtin/providers/google/data_source_google_compute_zones.go
+++ b/builtin/providers/google/data_source_google_compute_zones.go
@@ -1,0 +1,80 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"sort"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	compute "google.golang.org/api/compute/v1"
+)
+
+func dataSourceGoogleComputeZones() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleComputeZonesRead,
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"names": {
+				Type:     schema.TypeList,
+				Computed: true,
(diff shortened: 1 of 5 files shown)


builtin/providers/google/data_source_google_compute_zones.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func dataSourceGoogleComputeZonesRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	region := config.Region
+	if r, ok := d.GetOk("region"); ok {
+		region = r.(string)
+	}
+	projectBaseUrl := "https://www.googleapis.com/compute/v1/projects/" + config.Project
```
stack72: should this be done using fmt.Sprintf rather than concatenation? I am not sure which is the right way to do it TBH

radeksimko: Yeah, sure, why not... I will change it

radeksimko:
@stack72 Updated

stack72:
LGTM!

PR #11963, opened 2017-02-15
provider/azurerm: Remove location argument from loadbalancer_backend_address_pool

lvdh:
Applying a Terraform execution plan containing a loadbalancer_backend_address_pool with a location argument throws a warning:

```
Warnings:

  * azurerm_lb_backend_address_pool.test: "location": [DEPRECATED] location is no longer used

  No errors found. Continuing with 1 warning(s).
```

Removing the (required) location argument from the documentation as it is deprecated as per the warning.

(Feedbak is welcome, by all means. Apologies for misunderstanding the contribution guidelines if I missed something.)


stack72:
Thanks for picking that omission up! We just deprecated those fields recently

Paul

PR #11973, opened 2017-02-15
Update state.Locker

jbardin:
This is to update the state.Locker after some discussion over the RFC. The major difference with the new interface is the addition of a lock ID returned from Lock and required by Unlock. Extra metadata can be passed in via the LockInfo type, which is mostly filled out by the NewLockInfo constructor. The new types are as follows:

```
type Locker interface {
	Lock(info *LockInfo) (string, error)
	Unlock(id string) error
}

type LockInfo struct {
	// Unique ID for the lock. NewLockInfo provides a random ID, but this may
	// be overridden by the lock implementation. The final value of ID will be
	// returned by the call to Lock.
	ID string

	// Terraform operation, provided by the caller.
	Operation string
	// Extra information to store with the lock, provided by the caller.
	Info string

	// user@hostname when available
	Who string
	// Terraform version
	Version string
	// Time that the lock was taken.
	Created time.Time

	// Path to the state file when applicable. Set by the Lock implementation.
	Path string
}

type LockError struct {
	Info *LockInfo
	Err  error
}
```

The LockInfo also provides a pretty-printed String output, and a Marshal method which returns json.

The LockError's Error() method returns a multi-line string with the error followed by the Info string output for human-readable errors. The new error output looks like:

```
Error acquiring the state lock: resource temporarily unavailable
Lock Info:
  ID:        e2850786-48fc-a122-43cc-6a6d282a3bb1
  Path:      state.tfstate
  Operation: apply
  Who:       jbardin@mbp.lan
  Version:   0.9.0
  Created:   2017-02-15 19:12:14.823173423 +0000 UTC
  Info: 
```


backend/local/backend_local.go
```diff
@@ -29,13 +28,6 @@ func (b *Local) context(op *backend.Operation) (*terraform.Context, state.State,
 		return nil, nil, errwrap.Wrapf("Error loading state: {{err}}", err)
 	}
 
-	if op.LockState {
-		err := clistate.Lock(s, op.Type.String(), b.CLI, b.Colorize())
-		if err != nil {
-			return nil, nil, errwrap.Wrapf("Error locking state: {{err}}", err)
-		}
-	}
```
mitchellh: I think this makes a lot more sense :+1:

command/meta_backend_migrate.go
```diff
@@ -24,17 +24,23 @@ import (
...
+
+	lockIDOne, err := clistate.Lock(opts.One, lockInfoOne, m.Ui, m.Colorize())
 	if err != nil {
 		return fmt.Errorf("Error locking source state: %s", err)
 	}
-	defer clistate.Unlock(opts.One, m.Ui, m.Colorize())
+	defer clistate.Unlock(opts.One, lockIDOne, m.Ui, m.Colorize())
+
+	lockInfoTwo := state.NewLockInfo()
+	lockInfoTwo.Operation = "migration source state"
```
mitchellh: should be "destination"

state/state.go
```diff
@@ -42,9 +60,120 @@ type StatePersister interface {
...
+}
+
+// Marshal returns a string json representation of the LockInfo
+func (l *LockInfo) Marshal() []byte {
+	js, err := json.Marshal(l)
+	if err != nil {
+		panic(err)
+	}
+	return js
+}
```
mitchellh: Can we implement https://golang.org/pkg/encoding/#TextMarshaler instead? 

While we're at it, probably worth doing https://golang.org/pkg/encoding/json/#Marshaler and Unmarshaler.

This will cause those things to just work with other stdlib packages.

mitchellh:
Oh, another thing: the unlock failure error message should be updated to include the lock ID.

state/state.go
jbardin: I skipped the MarshalJSON so I could just call json.Marshal (otherwise it recurses), and since we control the data it shouldn't be able to error. I think only the encoding packages use TextMarshaler, and I wanted separate Text and JSON formats.

I can switch to MarshalJSON if you want, but I don't see a reason to use TextMarshaler.

mitchellh: Got it got it. Let's just leave it for now, I had reasons yesterday (unrelated) but I can't remember atm.

jbardin:
Doesn't the unlock failure message include the complete LockInfo, or did I miss a case somewhere?

mitchellh:
I missed it in the diff :(

LGTM

jbardin:
updated the migration lock info -- I'll merge once travis clears

PR #11978, opened 2017-02-15
state/remote: export ClientLocker, test for implementation

mitchellh:
This adds unit tests (that will fail at compile time) if various structs
don't implement the right interfaces for locking

### DIFF
--- a/state/remote/remote.go
+++ b/state/remote/remote.go
@@ -2,6 +2,8 @@ package remote
 
 import (
 	"fmt"
+
+	"github.com/hashicorp/terraform/state"
 )
 
 // Client is the interface that must be implemented for a remote state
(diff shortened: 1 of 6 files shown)


state/remote/remote.go
```diff
@@ -13,6 +15,15 @@ type Client interface {
 	Delete() error
 }
 
+// ClientLocker is an optional interface that allows a remote state
+// backend to enable state lock/unlock.
+type ClientLocker interface {
+	Client
+
```
jbardin: would it make sense to just embed `state.Locker`?

mitchellh: That's a fair point since the APIs are the same right now. Will do.

Issue #11991, opened 2017-02-16
`cloudflare_record` allows proxied = true on TXT domains


jleclanche:
Reproduce:

```hcl
resource "cloudflare_record" "example_txt_domain" {
	domain = "example.com"
	name = "txt"
	type = "TXT"
	value = "whatever"
	proxied = true
}
```

`terraform plan` will happily accept this. Applying it will fail however:

```
* cloudflare_record.example_txt_domain: Failed to create record: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1004,\"message\":\"DNS Validation Error\",\"error_chain\":[{\"code\":9004,\"message\":\"This kind of record cannot be proxied\"}]}],\"messages\":[],\"result\":null}"
```

**NOTE**: There are other record types which cannot be proxied (such as MX). I haven't tested them and I can't find an exhaustive list.

jleclanche:
The following types can't be proxied:

* `MX`
* `LOC`
* `SRV`
* `SPF`
* `TXT`
* `NS`

The following types can be proxied:

* `A`
* `AAAA`
* `CNAME`

omeroner:
My tf file;

resource "cloudflare_record" "ankara" {
    domain = "ankara.net"
    name = "deneme"
    value = "10.10.10.12"
    type = "A"
    proxied = true
    ttl = 120
}

we have a problem. 

cloudflare_record.ankara: Modifying...
  proxied: "false" => "true"
Error applying plan:

1 error(s) occurred:

* cloudflare_record.ankara: Failed to update CloudFlare Record: error from makeRequest: HTTP status 400: content "{\"success\":false,\"errors\":[{\"code\":1004,\"message\":\"DNS Validation Error\",\"error_chain\":[{\"code\":9041,\"message\":\"Invalid proxied mode. Record cannot be proxied\"}]}],\"messages\":[],\"result\":null}"

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.

how you can help me!

grubernaut:
Hi @omeroner!

Looks like the error you returned was directly from a cloudflare API request, and not a Terraform error.
Did this just recently start throwing these errors? I can reproduce both via Terraform and via the Cloudflare Web UI.

cemo:
gentle ping @OJFord. Are you aware of this problem please?

[...]

PR #11993, opened 2017-02-16
provider/cloudflare: Fix record validation

grubernaut:
Previously we only validated that the cloudflare record provided was a valid record type. However, a record can be of a valid type, and still not be proxied, making it an invalid record type.

The main downside to having to check for whether or not the record type is proxied or not during validation, is that it relies on having two schema keys populated. This means that we can only catch the improper record type during `apply` time, instead of `plan` time.

```
$ go test -v -run "TestValidateRecordType" ./builtin/providers/cloudflare
=== RUN   TestValidateRecordType
--- PASS: TestValidateRecordType (0.00s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/cloudflare     0.004s
```
Fixes: #11991

### DIFF
--- a/builtin/providers/cloudflare/resource_cloudflare_record.go
+++ b/builtin/providers/cloudflare/resource_cloudflare_record.go
@@ -18,51 +18,50 @@ func resourceCloudFlareRecord() *schema.Resource {
 		SchemaVersion: 1,
 		MigrateState:  resourceCloudFlareRecordMigrateState,
 		Schema: map[string]*schema.Schema{
-			"domain": &schema.Schema{
+			"domain": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
 
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
 
-			"hostname": &schema.Schema{
+			"hostname": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"type": &schema.Schema{
-				Type:         schema.TypeString,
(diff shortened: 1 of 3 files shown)


stack72:
Build fails:

```
go vet .
builtin/providers/cloudflare/resource_cloudflare_record.go:98: result of fmt.Errorf call not used
exit status 1
```

jleclanche:
@grubernaut If I'm reading this code correctly, this is a regression and now the record type doesn't even get validated at plan time. :/

grubernaut:
@jleclanche We currently have no way to validate a supplied schema item during plan time. So we can either validate the record type at plan time, and validate the prefixed record type at apply time, or perform both at the same time during apply time.

jleclanche:
@grubernaut By apply time, do you mean that the apply will fail before anything is actually applied or during the application? If the latter, it's no better than letting it fail on its own.

Plan-time validation is a lot more useful; Whoever lets Terraform handle critical infrastructure will be using a plan+apply cycle to verify things look good (and plan will be on CI).

grubernaut:
@jleclanche yes it will fail before it actually touches any real-world infrastructure. 

I fully agree that plan-time validation is much more useful, however we are limited by the interface that the validation functions are given, and only have access to a single schema element during plan-time validation. Which means, that it is currently impossible to validate whether or not a record type is valid based on the proxy value during plan time. We _can_ however, validate that a record type is valid during plan time, then validate if the record type is valid with respect to the proxy value during apply time.

jleclanche:
Following up in #12021

Issue #7636, opened 2016-07-14
Terraform removes non-managed project metadata


cblecker:
### Terraform Version

Terraform v0.6.16
### Affected Resource(s)
- google_compute_project_metadata
### Terraform Configuration Files

``` hcl
variable "region" {
  default = "us-east1"
}

variable "project_name" {
  description = "The ID of the Google Cloud project"
}

provider "google" {
  region      = "${var.region}"
  project     = "${var.project_name}"
  credentials = ""
}

resource "google_compute_project_metadata" "tf-testing-metadata" {
  metadata {
    tf-testing-1 = "test metadata variable 1 created by TF"
    tf-testing-2 = "test metadata variable 2 created by TF"
  }
}
```
### Debug Output

https://gist.github.com/cblecker/881acb52b2a3af9b194973ddd673fc08
### Panic Output

Not applicable
### Expected Behaviour

Any project metadata, including project-wide SSH keys created outside of Terraform should remain as part of the project should the Terraform-managed infrastructure be modified or destroyed.
### Actual Behaviour

The API call that Terraform uses to remove it's metadata from the project seems to clear all metadata from the project, including the very important sshKeys metadata key, which can result in loss of access to instances in the project that may not be managed by Terraform.
### Steps to Reproduce
1. Add non-Terraform managed metadata to a project: `gcloud compute project-info add-metadata --metadata testing-nontf-1=testing`
2. Verify metadata in the project: `gcloud compute project-info describe`
3. Add Terraform managed metadata to project: `terraform apply`
4. Verify metadata in the project: `gcloud compute project-info describe`
5. Destroy Terraform managed objects: `terraform destroy`
6. Verify metadata in the project: `gcloud compute project-info describe`
### Important Factoids

Not applicable
### References

Not applicable

cblecker:
https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_project_metadata.go#L190

cblecker:
I did a bit of further testing on this one and came up with some interesting results.

So if you create two different metadata resources, they conflict and overwrite each other on initial apply.
Something like:
```hcl
variable "region" {
  default = "us-east1"
}

variable "project_name" {
  description = "The ID of the Google Cloud project"
}

provider "google" {
  region      = "${var.region}"
  project     = "${var.project_name}"
  credentials = ""
}

resource "google_compute_project_metadata" "tf-testing-metadata" {
  metadata {
[...]

PR #12002, opened 2017-02-16
provider/google: Check for errors when deleting project metadata.

paddyforan:
Our delete operation for google_compute_project_metadata didn't check an
error when making the call to delete metadata, which led to a panic in
our tests. This is also probably indicative of why our tests
failed/metadata got left dangling.


cblecker:
Does this have any impact on #7636?

stack72:
@paddyforan ?

danawillow:
LGTM, thanks @paddyforan!

@cblecker, I don't think it will, no (paddy can confirm though)

paddyforan:
This does not affect #7636, @cblecker. This just fixes a test panic and allows us to debug a bit better.

I'll take a look at #7636, which is a bit tricky.

PR #12013, opened 2017-02-16
provider/aws: Refactor snapshot tests to always delete the snapshot first

stack72:
//cc @radeksimko


builtin/providers/aws/resource_aws_db_instance_test.go
```diff
@@ -780,12 +784,12 @@ resource "aws_db_instance" "snapshot" {
 	parameter_group_name = "default.mysql5.6"
 
 	copy_tags_to_snapshot = true
-	final_snapshot_identifier = "foobarbaz-test-terraform-final-snapshot-2"
+	final_snapshot_identifier = "foobarbaz-test-terraform-final-snapshot-%d"
```
grubernaut: This won't make the snapshot identifier too long will it?

PR #12014, opened 2017-02-16
documentation: add missing allowed_uses in certificate signing

cu12:
While I was using the `tls` provider's `self_signed_cert` resource for creating my CA cert I realized that although `crl_signing` is in the code, it's missing from the documentation. This PR aims to fix this.


apparentlymart:
Awesome! Thanks, @cu12!

PR #12026, opened 2017-02-16
build: Bump Vagrant to go1.8

cblecker:
Same as #12022, but for Vagrantfile 🎉


stack72:
Nice one! Thanks @cblecker :)

PR #11843, opened 2017-02-10
provider/fastly Allows for conditional settings across fastly

travertischio:
In Fastly:
* Allows user to set `cache_condition` for Gzip
* Allows user to set `cache_condition`, `request_condition` and `response_condition` for Header
* Adds optional `response_condition` setting for S3Logging
* Adds optional `response_condition` setting for Papertrail

- [x] Tests included
- [x] Website documentation included

#7554

PR #12028, opened 2017-02-17
provider/fastly Fixes logging response condition documentation

travertischio:
In #11843 conditions were added across the fastly service. In the documentation for the website it should say `response_condition` for s3 and papertrail logging. Instead the documentation says `request_condition`. This PR fixes this documentation change for both logging services related to fastly caches.


stack72:
LGTM thanks for this @travertischio :)

Issue #12030, opened 2017-02-17
output command fails when state is remote


ChrisParis:
### Terraform Version
`Terraform v0.9.0-dev (d8232314cca6f03297a7f02798ec26c8eff4d644)`

### Terraform Configuration Files
Using an S3 backend configured like:
```hcl
terraform {
  backend "s3" {
    name = "my-state"
    bucket = "my-terraform"
    key = "network/my.tfstate"
    region = "us-east-1"
  }
}
```

### Expected Behavior
`terraform output` shows the outputs.

### Actual Behavior
`terraform output` reports `The module root could not be found. There is nothing to output.`

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `terraform output`

### References

Pull request with fix: https://github.com/hashicorp/terraform/pull/12031

mitchellh:
Fixed by #12031

PR #12031, opened 2017-02-17
RefreshState in output command

ChrisParis:
Fixes output command when state is remote (#12030).


mitchellh:
LGTM. This needs a test though so we don't make this mistake again in the future. I'll add that post-merge.

PR #12039, opened 2017-02-17
aws/codebuild: Fixed documentation: Optional/Required

rickard-von-essen:
Also extended the example to show how have two `environment_variable`s.


stack72:
LGTM thanks for this @rickard-von-essen :)

PR #12041, opened 2017-02-17
Terraform environments

jbardin:
This PR adds the concept of local "environments" to Terraform, which can be created changed and deleted with the `env` command:

```
Usage: terraform env [options] [NAME]

  By default env will list all configured environments. If NAME is provided,
  env will change into to that named environment.


Options:

  -new=name      Create a new environment.
  -delete=name   Delete an existing environment,

  -state=path    Used with -new to copy a state file into the new environment.
  -force         Used with -delete to remove a non-empty environment.
```

The env command as presented here is a single command with flags for the `new` and `delete` operations. These could be refactored into subcommands if we chose, but are submitted here as a single command for simplicity.

### DIFF
--- a/command/env_command.go
+++ b/command/env_command.go
@@ -1,0 +1,306 @@
+package command
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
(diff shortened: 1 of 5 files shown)


jbardin:
closing this -- going to finish along with the enhanced backend support.

PR #12062, opened 2017-02-17
Add owner_account option to aws_redshift_cluster

jklukas:
See OwnerAccount in http://docs.aws.amazon.com/redshift/latest/APIReference/API_RestoreFromClusterSnapshot.html

I'm very new to terraform's code base, so it's unclear to me whether an acceptance test is needed and how to approach adding one (or updating an existing test).


stack72:
Hi @jklukas 

Thanks for the work here - i manually merged this as I made this small change

9bb678d76cb827ebee1bc8cc5496d15a3717e2b1

This should only be sent when it's specified - otherwise we would be sending an empty string

Paul

jklukas:
Thanks for the quick review and merge!

Issue #11971, opened 2017-02-15
Mechanism to allow rancher provider have a configurable timeout


jmahowald:
### Terraform Version
v0.8.6

### Affected Resource(s)
rancher_stack

I love the rancher provider that was added in GH-9173. My issue is that the rancher provider for me frequently but not always has a timeout.  Obviously this is dependent on my environment, but I feel like the timeout should be configurable in some way.  The underlying rancher go client does have a timeout setting, but I'd need to find a way to fit that in. I wanted to submit a pull request against this, but I don't yet know how to best tackle.   The only choices I can see are either through an environment variable setting or an additional setting on the provider.   If either seems acceptable, I'll send in a pull

Thanks,

stack72:
Hi @jmahowald 

Closing - as explained in #12070 - the new timeout features in Terraform 0.9 will take care of this :)

Paul

PR #12070, opened 2017-02-18
Adding ability to set rancher timeouts

jmahowald:
As per GH-11971 since rancher servers don't have anyone enforcing SLAs like Amazon, it is easily possible that a timeout can occur based on environmental conditions.  The rancher go api has this as a config option, and I wanted to make that available to the provider.

### DIFF
--- a/builtin/providers/rancher/config.go
+++ b/builtin/providers/rancher/config.go
@@ -2,6 +2,7 @@ package rancher
 
 import (
 	"log"
+	"time"
 
 	rancherClient "github.com/rancher/go-rancher/client"
 	"github.com/raphink/go-rancher/catalog"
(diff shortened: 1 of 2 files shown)


stack72:
@raphink / @johnrengelman thoughts here gentlemen?

johnrengelman:
Seems innocuous to me. How will this work in conjunction with the new resource timeouts from here https://github.com/hashicorp/terraform/pull/12311?

stack72:
Agreed - @jmahowald 

Thanks for the work here - as @johnrengelman points out, there is a much bigger timeout refactor coming in 0.9 (via #12311)

Therefore, I am going to close this - it will conflict with that work and this work will allow us to set create, update and delete timeouts easily across an entire provider

Thanks for all the work here

Paul

jmahowald:
Was just going to suggest that I could redo it so that it matched that model.  It removes the boilerplate changes.   But if it'll take a while to settle, I'll hold off

stack72:
@jmahowald the new 0.9 stuff is coming *really* soon:tm: ;)

jmahowald:
Not sure if there's a branch that fixes it, but I would contend the referenced  fix only changes the mechanism of how to express timeouts, but that the way to implement those is still dependent on the provider, and thus the logic to set the timeout of the underlying rancher client connection would still need to be done.  Shouldn't be too hard to try a fix with that mechanism, though I prefer to wait until I have a working 0.9 release to try a new PR.

Issue #12000, opened 2017-02-16
Openstack volume attachment not working


michalmedvecky:
### Terraform Version

Terraform v0.9.0-dev (89acb5811a25e59e6a9d55cf5a2429846e3b5982+CHANGES)

### Affected Resource(s)

- openstack_blockstorage_volume_attach_v2

### Terraform Configuration Files
```hcl
resource "openstack_blockstorage_volume_v2" "tftestvol" {
  name = "tf-test-volume"
  description = "Delete me whenever you need"
  size = 100
}
resource "openstack_blockstorage_volume_attach_v2" "tftestmount" {
  instance_id = "${openstack_compute_instance_v2.testinstance.id}"
  volume_id = "${openstack_blockstorage_volume_v2.tftestvol.id}"
  device = "/dev/vdd"
  attach_mode = "rw"
}
```

### Expected Behavior

Volume should be attached to the instance

### Actual Behavior
Volume is somehow semi-attached and cannot be seen by the instance

### Steps to Reproduce
1. `terraform apply`

### Important Factoids

I have two volumes in the same tenants, one of them created/attached by TF and the other one via Horizon.

This is what I see:

```
$ openstack volume list
+--------------------------------------+--------------------+--------+------+-----------------------------------------------------------+
| ID                                   | Display Name       | Status | Size | Attached to                                               |
+--------------------------------------+--------------------+--------+------+-----------------------------------------------------------+
| 913e680e-0066-4f28-ba12-edf1ec57ea68 | tf-test-volume     | in-use |  100 | Attached to mytestinstance.com on /dev/vdd  |
| 6b8ac3dd-a141-449c-adda-3d1d91525b47 | test               | in-use |  100 | Attached to mytestinstance.com on /dev/vdb  |
+--------------------------------------+--------------------+--------+------+-----------------------------------------------------------+
```

`tf-test-volume` is TF created/managed
`test` is Horizon created/attached

But when I run

```
$ nova volume-attachments mytestinstance
+--------------------------------------+----------+--------------------------------------+--------------------------------------+
| ID                                   | DEVICE   | SERVER ID                            | VOLUME ID                            |
+--------------------------------------+----------+--------------------------------------+--------------------------------------+
| 6b8ac3dd-a141-449c-adda-3d1d91525b47 | /dev/vdb | aaba3514-d079-45db-a79c-dfdd13825046 | 6b8ac3dd-a141-449c-adda-3d1d91525b47 |
[...]

PR #12071, opened 2017-02-18
provider/openstack: Redesign openstack_blockstorage_volume_attach_v2

jtopjian:
The current design of openstack_blockstorage_volume_attach_v2 does
not correctly implement the Block Storage API attachment call. It
was only partially implemented, only marking volumes as being
attached, while never actually attaching them.

This redesign is a closer alignment to how creating attachments
to a standalone Block Storage service works.

For creating attachments specifically in the case of OpenStack
Compute instances, the openstack_compute_volume_attach_v2 resource
is required.

For #12000


jtopjian:
So this is embarrassing. I originally made this resource in parallel to `openstack_compute_volume_attach_v2` and I must have gotten my testing mixed up. None of the acceptance tests actually _log in_ to the created virtual machines to confirm a volume was attached (it would be quite difficult to do, but something to perhaps look into in the future), so I must have fumbled when I did a manual verification. Since then, always thinking this was verified at some point. Ouch.

As reported in #12000, no volume is actually being attached to the instance. Therefore, if anyone is using this resource, it's not actually doing anything, and thus, IMO, open to either two things:

1. total removal
2. attempt to fix it for what it should actually be used for.

I'm going with 2. I have actually tested this one across two OpenStack clouds (and did some iSCSI work, as noted in the updated docs), so I'm _positive_ this resource actually _does something_ this time. However, it's worth noting that this functionality is a bit cutting-edge (and mostly undocumented, as can humorously be seen [here](http://lists.openstack.org/pipermail/openstack-dev/2016-January/083684.html)). I'm OK with that. I'd rather try adding something where a user could do some possibly cool stuff and report any fixes required than entirely yanking this resource. 

But if it's safer to just remove this resource, I'm OK with that, too, and I apologize for the blunder on this one.

michalmedvecky:
For my purposes, `openstack_compute_volume_attach_v2` works fine so as someone who reported the issue, I am fine with the removal of `openstack_blockstorage_volume_attach_v2`.

jtopjian:
@michalmedvecky Yeah, removing this won't actually hurt anyone, so I'm not opposed to it. Since it's already been added, though, I'm interested to see if users find this useful when applied in the way it was supposed to be used.

I've updated the docs to mention that this resource is experimental and might be removed in the future, which will help in the case when it's best to just throw in the towel. 

@stack72 Thoughts?

stack72:
Now that we have handled backwards incompatibility - this LGTM! Thanks for tall the work on fixing this resource up @jtopjian

PR #12080, opened 2017-02-18
website: Fixes data entry errors on /d/aws_db_instance

jcardinal:
Corrected so \<h1\> shows correct topic and intro text is applicable. This applies to shipping versions of terraform.


stack72:
LGTM Thanks for this :)

PR #12082, opened 2017-02-18
provider/aws: Data source aws_caller_identity exposes resource from ARN

alanctgardner:
I was unable to find a data source that provides the IAM user name and ID for the access key currently being used. This is useful for environments where many users are creating resources and we want to attribute resources to the user who created them by adding a tag, for instance.

I'd be happy to break this out into a new data source - it seemed like there were already several sources for information about the access key in use.

### DIFF
--- a/builtin/providers/aws/auth_helpers.go
+++ b/builtin/providers/aws/auth_helpers.go
@@ -21,7 +21,7 @@ import (
 	"github.com/hashicorp/go-cleanhttp"
 )
 
-func GetAccountInfo(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (string, string, error) {
+func GetAccountInfo(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (string, string, string, error) {
 	// If we have creds from instance profile, we can use metadata API
 	if authProviderName == ec2rolecreds.ProviderName {
 		log.Println("[DEBUG] Trying to get account ID via AWS Metadata API")
@@ -30,7 +30,7 @@ func GetAccountInfo(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string)
 		setOptionalEndpoint(cfg)
 		sess, err := session.NewSession(cfg)
 		if err != nil {
-			return "", "", errwrap.Wrapf("Error creating AWS session: {{err}}", err)
+			return "", "", "", errwrap.Wrapf("Error creating AWS session: {{err}}", err)
 		}
 
 		metadataClient := ec2metadata.New(sess)
@@ -38,7 +38,7 @@ func GetAccountInfo(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string)
 		if err != nil {
 			// This can be triggered when no IAM Role is assigned
 			// or AWS just happens to return invalid response
-			return "", "", fmt.Errorf("Failed getting EC2 IAM info: %s", err)
+			return "", "", "", fmt.Errorf("Failed getting EC2 IAM info: %s", err)
 		}
 
 		return parseAccountInfoFromArn(info.InstanceProfileArn)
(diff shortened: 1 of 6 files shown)


radeksimko:
Hi @alanctgardner 
thanks for the contribution.
A few questions:

> This is useful for environments where many users are creating resources and we want to attribute resources to the user who created them by adding a tag, for instance.

1. Are you aware that each time a different team member executes terraform plan/apply, they will effectively "re-own all resources" as the data source will return current identity & update all references? Is that intended?
1. Have you tested this with different auth methods? It may not always be IAM user executing terraform - specifically we already support EC2 Instance Role, Assumed IAM Role, IdP-federated IAM Role (e.g. SAML) and it would be shame we broke the data source for any one of these use cases.
1. Was there any specific reason you decided to re-implement the logic here instead of modifying the [existing](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/auth_helpers.go#L24-L84) [code](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/config.go#L261-L265)?

I think that **ARN of a role/user/identity** is something we could reliably obtain in all cases and it can universally represent the identity of API caller, although I'm not sure if that's any helpful for your use case (see my 1st question).

alanctgardner:
@radeksimko Thanks for the quick response. I'm pretty new to terraform, I appreciate you taking the time to give feedback:

1) This is fine for my use-case, I realize the tag would be re-computed.

2 + 3) I've reworked it to extract the resource itself from the ARN, which works for my use case but should also support the other auth methods you mentioned.

alanctgardner:
@radeksimko Just wanted to follow up and see if you have further feedback on this PR. Thanks!

scouten:
I'd like to add a bump to this PR. I just ran into a similar use case and the ability to extract the user name would have been very helpful. Any hope of getting this updated and merged?

scouten:
FWIW I would have preferred the version initially submitted (838f880447be899a9d87fb2828f4f0f94e2e3195).

scouten:
Our use case is to provide individual user sandboxes on AWS that prevent inadvertent impact on other users' resources. Consider this fragment of an IAM policy:

```
        {
            "Sid": "ForceInstanceOpWithTag",
            "Action": [
                "ec2:StartInstances",
                "ec2:StopInstances",
                "ec2:RebootInstances",
                "ec2:TerminateInstances"
            ],
            "Effect": "Allow",
            "Resource": "arn:aws:ec2:us-west-2:1234556789012:instance/*",
            "Condition": {
                "StringEquals": {
                    "ec2:ResourceTag/UserName": "${aws:username}"
                }
            }
        },
```

I want to have the TF script be able to generate a corresponding resource tag on instance creation. (Due to the sandbox model, we are intentionally not placing our `.tfstate` files in version control.)

alanctgardner:
@scouten It looks like this already got fixed on April 3rd: e97900aef28a4d3b2f3df30ee9558dfc8f730503

Can you use `${aws:user_id}`?

scouten:
No joy. With this `.tf` file …

```
provider "aws" {
  region = "us-west-2"
}

# TODO: Remove example
  ami = "ami-7ac6491a"
  instance_type = "t2.micro"
  tags {
    UserName = "${aws:user_id}"
  }
}
```

… I get the following failure:

```
$ terraform apply
Failed to load root config module: Error loading .../path/to/amazon.tf: Error reading config for aws_instance[example]: parse error at 1:6: expected "}" but found ":"
```

alanctgardner:
Whoops, so based on the examples I think it should be:

```
provider "aws" {
  region = "us-west-2"
}

  ami = "ami-7ac6491a"
  instance_type = "t2.micro"
  tags {
    UserName = "${data.aws_caller_identity.current.user_id}"
  }
}

```

scouten:
Oh so close. Out of the following information …

```
$ aws iam get-user
{
    "User": {
        "UserName": "fred-bob",
        "Path": "/",
        "CreateDate": "2017-04-01T12:34:56Z",
        "UserId": "ABCD1234...",
        "Arn": "arn:aws:iam::123456789012:user/fred-bob"
    }
}
```

… the syntax you suggested gave me `"ABCD1234..."`. I was hoping to get `"fred-bob"`.

scouten:
That said, I might be able to tweak my IAM rules to use the user ID instead of the user name. Not as pretty, but at least functional …

alanctgardner:
Try parsing the name out of the ARN instead:

```
provider "aws" {
  region = "us-west-2"
}

  ami = "ami-7ac6491a"
  instance_type = "t2.micro"
  tags {
    UserName = "${element(split(data.aws_caller_identity.current.arn, ':'), 5)}"
  }
}
```

scouten:
@alanctgardner I wound up using a slightly different syntax based on a comment in the Gitter thread. The key line here was:

```
    UserName = "${element(split("/",data.aws_caller_identity.current.arn),1)}"
```

The syntax you gave expanded out to a single colon. In any case, thanks very much and I'm off to the races!

Issue #11868, opened 2017-02-10
Openstack API error messages are not propagated to the user


ruoso:
Running with terraform 0.8.6, I had a configuration that would just give me the output:

```
* openstack_compute_instance_v2.terraform: Error creating OpenStack server: Invalid request due to incorrect syntax or missing required parameters.
```

Even enabling trace wouldn't give me any more useful output to indicate what was wrong. After a lot of suffering I finally decided to run terraform with mitmproxy to be able to see if there was anything useful in the requests/responses themselves, and it turns out that the openstack server was returning a very useful error message that allowed me to solve the problem right away (flavor disk too small for image).

Please propagate the errors from the openstack API to the user, or at least add debugging output with the responses such that we have some hope of troubleshooting.

jtopjian:
@ruoso Agreed. This is an issue with the underlying OpenStack library that Terraform uses. This is a known issue and will hopefully be resolved soon. I'll leave this Issue open as a placeholder and will update it if there's any news.

ruoso:
Maybe a short term answer is to have gophercloud just have an env variable for dumping the json for the requests and responses... But I guess that'd be an issue on that repo, not here.

jtopjian:
That should be possible to do within Terraform. I will investigate this.

jtopjian:
@ruoso I've just submitted #12089 which enables this functionality. Feel free to test it out and let me know if you have any feedback :)

stack72:
Closed via #12089

PR #12089, opened 2017-02-19
provider/openstack: Enable HTTP Logging

jtopjian:
This commit adds the ability to log all requests and responses
between Terraform and the OpenStack cloud. To enable, set the
OS_DEBUG environment variable to 1.

This implementation is based off of the [Rackspace CLI](https://github.com/rackspace/rack/blob/1.3/auth/clients.go#L309).

For #11868

### DIFF
--- a/builtin/providers/openstack/config.go
+++ b/builtin/providers/openstack/config.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net/http"
+	"os"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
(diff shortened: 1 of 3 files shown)


stack72:
LGTM! Thanks :)

ihr:
@jtopjian apologies for the question, but when I set `OS_DEBUG=1 terraform apply ... ` I don't get any debug only the `openstack_compute_instance_v2.terraform: Error creating OpenStack server: Invalid request due to incorrect syntax or missing required parameters.` What am I missing?

jtopjian:
@ihr You have to combine `OS_DEBUG=1` with `TF_LOG=DEBUG` :)

ihr:
Thank you @jtopjian , indeed `TF_LOG=DEBUG OS_DEBUG=1 terraform apply ...` does the magic.

PR #12092, opened 2017-02-19
Move AWS Config to be alphabetically correct

bodgit:
Noticed the recently added AWS Config resources are not quite in the right place in the sidebar for the documentation. This PR just shuffles the links to the right place.


radeksimko:
Good catch, thanks!

PR #12094, opened 2017-02-19
website/docs fix #5448 - vm page was changed

### LINKED ISSUE #5448: vSphere provider tutorial fails
Currently in the vSphere provider tutorial the `name` attribute is set to a name with an underscore in it.

However this causes VMWare customization to fail when the node is brought up. 

The error returned from VMWare is: 
```
vsphere_virtual_machine.web: A specified parameter was not correct.
spec.identity.hostName
```

Searching around I finally figured out what it meant via this doc, specifically in the `Resolution` section:
https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=2009820

Terraform Version:
```
terraform -v
Terraform v0.6.12
```

vSphere Version:
```
VMWare vSphere v5.5 U2
```

Really just bringing it up so if someone searches for the issue they can find the resolution more easily, as it really doesn't have anything to do with Terraform itself.

ewypych:
Hello!

According to issue #5448 I made more changes. "vsphere_virtual_machine" contains an underscore in the "name" var, and that is wrong. Also I added information about underscores and name length. I think it is an important information and can save the time for newbies. :)

If you think it is ok, please merge it. I do not see any additional places with this problem in the vSphere Provider. 

Thanks in advance!

Emil

### DIFF
--- a/website/source/docs/providers/vsphere/r/virtual_machine.html.markdown
+++ b/website/source/docs/providers/vsphere/r/virtual_machine.html.markdown
@@ -15,7 +15,7 @@ modify, and delete virtual machines.
 
 ```
 resource "vsphere_virtual_machine" "web" {
-  name   = "terraform_web"
+  name   = "terraform-web"
   vcpu   = 2
   memory = 4096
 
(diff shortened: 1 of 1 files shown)


stack72:
LGTM! thanks

PR #12096, opened 2017-02-19
provider/openstack: Updates to openstack_images_image_v2 resource

jtopjian:
This commit has a few more fixes to the recently added
openstack_images_image_v2 resource:

* tags were changed to a Set because the OpenStack Image API does
not seem to respect ordering.
* The visibility argument was fixed.
* Acceptance tests for all updatable fields has been implemented.
* Documentation updates, including a new entry in the sidebar.


jtopjian:
I apologize for the post-commits on this new resource. I should have tested this harder before I initially merged it.

PR #12097, opened 2017-02-20
provider/openstack: Image Data Source

jtopjian:
This commit adds the `openstack_images_image_v2` data source which
is able to query the Image Service v2 API for a specific image.


stack72:
LGTM :)

PR #12099, opened 2017-02-20
provider/openstack: Don't allow floating IP and port

jtopjian:
This commit adds a check to prevent a user from specifying both
a floating IP and a port on a specific network. While this
configuration is currently allowed, the Port will be chosen and
applying the configuration again will show a state mismatch. This
attempts to prevent such a misconfiguration.

Fixes #10548


stack72:
LGTM!

Issue #12103, opened 2017-02-20
provider/aws: Invalid protocol aws_network_acl_rule


MattiasGees:
### Terraform Version
v0.8.7

### Affected Resource(s)
- aws_network_acl_rule

### Terraform Configuration Files
```hcl
resource "aws_network_acl" "nat" {
  vpc_id     = "${var.vpc_id}"
  subnet_ids = ["${var.nat_subnets}"]
}

resource "aws_network_acl_rule" "nat_ingress_vrrp_nat" {
  network_acl_id = "${aws_network_acl.nat.id}"
  rule_number    = "1110"
  egress         = false
  protocol       = 112
  rule_action    = "allow"
  cidr_block     = "10.20.20.0/24"
}
```

### Debug Output
```
* aws_network_acl_rule.nat_ingress_vrrp_nat: Invalid Protocol 112 for rule 1110
```

### Expected Behavior
That the `terraform plan` completes without giving any errors. 

### Actual Behavior
`terraform plan` gives errors at the end of the plan to say that 112 is an invalid protocol. While at the seem time this is saved in the terraform state file with the correct protocol. It is also correctly deployed on the AWS console
I looked it up in the protocol numbers and `112` is present in that list http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `terraform plan`

stack72:
Closed via #12107

PR #12107, opened 2017-02-20
Add VRRP to allowed protocols in network ACL rules

MattiasGees:
This will add VRRP to the allowed protocols list in ACL entries. This is a fix for  #12103. I saw that a lot of the protocols defined in https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml are not in this list. Is there a reason not to? 

If needed I can add more of the protocols defined in iana.


stack72:
LGTM! Thanks for this @MattiasGees :) I think if there are some useful protocols you would need added, then send them

Issue #9799, opened 2016-11-02
provider/openstack: Remove default egress rule from Security Group on creation


gpaggi:
Hi there,

### Terraform Version
Terraform v0.7.8

### Affected Resource(s)
openstack_networking_secgroup_v2
openstack_networking_secgroup_rule_v2

### Terraform Configuration Files
```hcl
resource "openstack_networking_secgroup_v2" "_def_secgroup" {
  region = "${var.region}"
  name = "_def_secgroup"
  description = "Default security group"
}

resource "openstack_networking_secgroup_rule_v2" "_def_secgroup_rule_1" {
  direction = "egress"
  ethertype = "IPv4"
  protocol = "tcp"
  security_group_id = "${openstack_networking_secgroup_v2._def_secgroup.id}"
  remote_ip_prefix = "10.0.0.0/8"
}
```

### Expected Behavior
Terraform should create a security group with one egress rule limiting TCP traffic to only 10.0.0.0/8.

### Actual Behavior
On security group creation, a default egress ANY:ANY is added by Neutron next to the one managed by TF.
This makes it impossible to manage egress rules from TF, since they are overridden by the default egress rule.

```
➜  terraform-dev neutron security-group-show _def_secgroup                                                                                                                                     
+----------------------+--------------------------------------------------------------------+
| Field                | Value                                                              |
+----------------------+--------------------------------------------------------------------+
| description          | Default security group                                             |
| id                   | 1b9a92a6-2972-4717-8a03-a14b2acdb197                               |
| name                 | _def_secgroup                                                      |
| security_group_rules | {                                                                  |
|                      |      "remote_group_id": null,                                      |
|                      |      "direction": "egress",                                        |
|                      |      "remote_ip_prefix": "0.0.0.0/0",                              |
|                      |      "protocol": "any",                                            |
|                      |      "ethertype": "IPv4",                                          |
|                      |      "port_range_max": 65535,                                      |
|                      |      "security_group_id": "1b9a92a6-2972-4717-8a03-a14b2acdb197",  |
|                      |      "port_range_min": 0,                                          |
|                      |      "tenant_id": "f35f2d7d893d431b8a105242da8de971",              |
|                      |      "id": "37659e18-eb70-4504-b140-096fe4b72de1"                  |
|                      | }                                                                  |
|                      | {                                                                  |
|                      |      "remote_group_id": null,                                      |
|                      |      "direction": "egress",                                        |
[...]

PR #12119, opened 2017-02-20
provider/openstack: Toggle Creation of Default Security Group Rules

### LINKED ISSUE #12102: 0.8.6 is not backwards compatible for OpenStack security groups
Version: >=0.8.6

### Affected Resource(s)
openstack_networking_secgroup_v2
openstack_networking_secgroup_rule_v2

```hcl
resource "openstack_networking_secgroup_v2" "secgroup" {
  region      = "CustomRegion"
  name        = "terraform-test-group"
}
```

Before 0.8.6 Terraform the [default OpenStack behaviour](https://wiki.openstack.org/wiki/Neutron/SecurityGroups) was implemented. The created security group did contain two egress rules allowing all outgoing IPv4 and IPv6 traffic.

Executing the same script with >=0.8.6 creates an empty security group.

That means, you need to create two additional security group rules.
```hcl
resource "openstack_networking_secgroup_rule_v2" "secgroup_rule_v4" {
  direction = "egress"
  ethertype = "IPv4"
  security_group_id = "${openstack_networking_secgroup_v2.secgroup.id}"
}

resource "openstack_networking_secgroup_rule_v2" "secgroup_rule_v6" {
  direction = "egress"
  ethertype = "IPv6"
  security_group_id = "${openstack_networking_secgroup_v2.secgroup.id}"
}
```

Executing this with >=0.8.6 creates again the expected security group with its rules.

Executing this with <0.8.6 breaks because the rules already exist.

Having a running system where the security group was created the old way, and running an update with the new version including the explicit egress rule fails.

It would be great, if the [breaking change](https://github.com/hashicorp/terraform/pull/11466) could be reverted since it is not backwards compatible as @jtopjian suggested.
Creating the security group entity is not breaking, but you need to introduce new egress rules which are breaking.

Alternatively to removing the change completely would be to introduce a flag in the security group resource. Without the flag, the sec group should still have the default egress rules. This would be backwards compatible. But people could specify the flag in order to get a "clean/empty/secure by default" security group as I guess this is what you wanted to achieve with the change.

Suggested example:
```hcl
resource "openstack_networking_secgroup_v2" "secgroup" {
  region      = "CustomRegion"
  name        = "terraform-test-group"
  empty      = "true"
}
```
default of `empty` being `false`

jtopjian:
This commit modifies the behavior implemented in #9799 by enabling
the user to be able to toggle the creation of the default security
group rules.

Fixes #12102

non-commit note: This effectively changes the behavior introduced in #9799 to be opt-in instead of dictated, which I agree is a better solution. I tried to make the new argument intuitive as well as provide documentation about how to recreate the default rules. Caveats are mentioned for OpenStack environments that have deviated from the default behavior.

Between #9799 and now, there will be three sets of `openstack_networking_secgroup_v2`'s in the wild, however, no changes should happen to existing deployments since there is not a default value specified and an explicit value of `true` must be given for the rules to be deleted.

In addition, no destructive behavior is expected (and by destructive, I mean the loss of data and stateful services). If someone chooses to add this argument with a value of `true` or `false` to their configuration, then the security group and the rules connected to it will be recreated. All resources that consume security groups are able to handle security group updates without being destroyed. If such a thing happens, it is a bug and should be reported.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_secgroup_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_secgroup_v2.go
@@ -46,6 +46,11 @@ func resourceNetworkingSecGroupV2() *schema.Resource {
 				ForceNew: true,
 				Computed: true,
 			},
+			"delete_default_rules": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+			},
 		},
 	}
 }
@@ -71,11 +76,14 @@ func resourceNetworkingSecGroupV2Create(d *schema.ResourceData, meta interface{}
 		return err
 	}
 
-	// Remove the default rules
-	for _, rule := range security_group.Rules {
-		if err := rules.Delete(networkingClient, rule.ID).ExtractErr(); err != nil {
-			return fmt.Errorf(
-				"There was a problem deleting a default security group rule: %s", err)
+	// Delete the default security group rules if it has been requested.
+	deleteDefaultRules := d.Get("delete_default_rules").(bool)
+	if deleteDefaultRules {
+		for _, rule := range security_group.Rules {
+			if err := rules.Delete(networkingClient, rule.ID).ExtractErr(); err != nil {
+				return fmt.Errorf(
+					"There was a problem deleting a default security group rule: %s", err)
+			}
 		}
 	}
 
--- a/builtin/providers/openstack/resource_openstack_networking_secgroup_v2_test.go
+++ b/builtin/providers/openstack/resource_openstack_networking_secgroup_v2_test.go
@@ -23,7 +23,7 @@ func TestAccNetworkingV2SecGroup_basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckNetworkingV2SecGroupExists(
 						"openstack_networking_secgroup_v2.secgroup_1", &security_group),
-					testAccCheckNetworkingV2SecGroupRuleCount(&security_group, 0),
+					testAccCheckNetworkingV2SecGroupRuleCount(&security_group, 2),
 				),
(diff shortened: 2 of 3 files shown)


Kiemes:
@jtopjian I see that you are using boolean value for `delete_default_rules`. If I would like to set the value via a terraform variable this wouldn't be possible, would it?
As far as I can read from the [documentation](https://www.terraform.io/docs/configuration/variables.html), variables cannot be from type `bool`. Or is type `bool` being inferred by you setting the default to `false`?
Just to make sure because I have not stubbled upon any other boolean variables in Terraform so far.
https://www.terraform.io/docs/configuration/syntax.html however suggests that using boolean values directly is allowed. Is just the documentation out of sync?

jtopjian:
@Kiemes Booleans can be used. :)

In the link you referenced, check out the section that reads:

> When passing boolean-like variables as parameters to resource configurations that expect boolean values, they are converted consistently

In this case, the following works just fine:

```hcl
variable "ddr" {                                     
    default = true                                   
}                                                    
                                                     
resource "openstack_networking_secgroup_v2" "test" {
  name = "test"                                     
  description = "test"                              
  delete_default_rules = "${var.ddr}"                
}                                                    
```

You can test this out with any other parameter that is currently of type boolean, of which there are currently 25 in the OpenStack provider (not counting this new one).

However, I would imagine you will want to just omit `delete_default_rules` entirely so you don't run into the situation you described yesterday.

Kiemes:
@jtopjian Thanks for the explanation. I actually expected you to do the right thing :) Just wanted to make sure.

And yes, we are going to use the default value `false` for `delete_default_rules`.

stack72:
LGTM :)

jtopjian:
@stack72 Thank you!

PR #12141, opened 2017-02-21
provider/pagerduty: Import support for service integrations

heimweh:
This PR adds support for importing service integrations as well as fixes an issue where `vendor` would never be set in the state.

Since we need two IDs when importing a service integration (the `serviceID` and the `serviceIntegrationID`) the import just tries to locate a service with the matching service integration ID, ~~but it would be even better to be explicit here and maybe have something like: `terraform import pagerduty_service_integration.main <serviceID>_<serviceIntegrationID>`~~ 
Update: Since it's only possible to have one service integration for one service at a time, this should be a safe to do.

This PR also removes the unnecessary use of custom imports for the other PagerDuty resources.
Since we're not doing anything fancy with those we can just use `schema.ImportStatePassthrough` instead.

Any feedback is more than welcome.

### DIFF
--- a/builtin/providers/pagerduty/import_pagerduty_service_integration_test.go
+++ b/builtin/providers/pagerduty/import_pagerduty_service_integration_test.go
@@ -1,0 +1,28 @@
+package pagerduty
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccPagerDutyServiceIntegration_import(t *testing.T) {
+	resourceName := "pagerduty_service_integration.foo"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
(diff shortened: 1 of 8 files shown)


heimweh:
Import test passing:

```bash

$ make testacc TEST=./builtin/providers/pagerduty TESTARGS='-run=TestAccPagerDutyServiceIntegration*'                                                                                               20:14:34
2017/02/23 20:14:46 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/pagerduty -v -run=TestAccPagerDutyServiceIntegration* -timeout 120m
=== RUN   TestAccPagerDutyServiceIntegration_import
--- PASS: TestAccPagerDutyServiceIntegration_import (17.20s)
=== RUN   TestAccPagerDutyServiceIntegration_Basic
--- PASS: TestAccPagerDutyServiceIntegration_Basic (26.01s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/pagerduty	43.238s
```

stack72:
LGTM Thanks for this @heimweh :)

PR #12147, opened 2017-02-21
provider/google: Check all fields in GKE tests instead of just that the resource exists

danawillow:
@paddyforan, I'd love your thoughts on this approach. One the one hand, yay checking more stuff! On the other hand, it's a long function that itself could be buggy, and requires that people actually keep it up-to-date.

### DIFF
--- a/builtin/providers/google/provider.go
+++ b/builtin/providers/google/provider.go
@@ -241,7 +241,7 @@ func getNetworkName(d *schema.ResourceData, field string) (string, error) {
 	return "", nil
 }
 
-func getNetworkNameFromString(network string) (string, error) {
+func getNetworkNameFromSelfLink(network string) (string, error) {
 	if strings.HasPrefix(network, "https://www.googleapis.com/compute/") {
 		// extract the network name from SelfLink URL
 		networkName := network[strings.LastIndex(network, "/")+1:]
--- a/builtin/providers/google/provider.go
+++ b/builtin/providers/google/provider.go
@@ -236,7 +236,7 @@ func getNetworkLink(d *schema.ResourceData, config *Config, field string) (strin
 func getNetworkName(d *schema.ResourceData, field string) (string, error) {
 	if v, ok := d.GetOk(field); ok {
 		network := v.(string)
-		return getNetworkNameFromString(network)
+		return getNetworkNameFromSelfLink(network)
 	}
 	return "", nil
 }
--- a/builtin/providers/google/provider.go
+++ b/builtin/providers/google/provider.go
@@ -236,17 +236,20 @@ func getNetworkLink(d *schema.ResourceData, config *Config, field string) (strin
 func getNetworkName(d *schema.ResourceData, field string) (string, error) {
 	if v, ok := d.GetOk(field); ok {
 		network := v.(string)
-
-		if strings.HasPrefix(network, "https://www.googleapis.com/compute/") {
-			// extract the network name from SelfLink URL
-			networkName := network[strings.LastIndex(network, "/")+1:]
-			if networkName == "" {
-				return "", fmt.Errorf("network url not valid")
-			}
-			return networkName, nil
-		}
-
-		return network, nil
+		return getNetworkNameFromString(network)
 	}
 	return "", nil
+}
+
+func getNetworkNameFromString(network string) (string, error) {
+	if strings.HasPrefix(network, "https://www.googleapis.com/compute/") {
+		// extract the network name from SelfLink URL
+		networkName := network[strings.LastIndex(network, "/")+1:]
+		if networkName == "" {
+			return "", fmt.Errorf("network url not valid")
+		}
+		return networkName, nil
+	}
+
+	return network, nil
 }
--- a/builtin/providers/google/resource_container_cluster_test.go
+++ b/builtin/providers/google/resource_container_cluster_test.go
@@ -20,7 +20,7 @@ func TestAccContainerCluster_basic(t *testing.T) {
 			resource.TestStep{
 				Config: testAccContainerCluster_basic,
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckContainerClusterExists(
+					testAccCheckContainerCluster(
 						"google_container_cluster.primary"),
 				),
 			},
@@ -37,10 +37,8 @@ func TestAccContainerCluster_withAdditionalZones(t *testing.T) {
 			resource.TestStep{
 				Config: testAccContainerCluster_withAdditionalZones,
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckContainerClusterExists(
+					testAccCheckContainerCluster(
 						"google_container_cluster.with_additional_zones"),
-					testAccCheckContainerClusterAdditionalZonesExist(
-						"google_container_cluster.with_additional_zones", 2),
 				),
 			},
 		},
@@ -56,7 +54,7 @@ func TestAccContainerCluster_withVersion(t *testing.T) {
 			resource.TestStep{
 				Config: testAccContainerCluster_withVersion,
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckContainerClusterExists(
+					testAccCheckContainerCluster(
 						"google_container_cluster.with_version"),
 				),
 			},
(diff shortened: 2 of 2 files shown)


danawillow:
I think both ways make sense, and I don't have a strong opinion either way. Here's an argument for doing it the way I have now:

This way is a bit simpler since each test gets to make the exact same call (rather than each one having to do it based on which values were set in that particular test), which I think avoids a certain amount of developer laziness that the other approach could have ("eh, I can just go ahead and check that the resource exists instead of checking each field" which is pretty much what we have now). We could hypothetically have one function that calls all the rest, but then I'm not sure that gives us an advantage over what I have now.

For the point of reusable functions- that's definitely a good practice overall, is there an opportunity for them to be reused at all? If there were, I'd be all for it.

paddyforan:
> This way is a bit simpler since each test gets to make the exact same call (rather than each one having to do it based on which values were set in that particular test), which I think avoids a certain amount of developer laziness that the other approach could have ("eh, I can just go ahead and check that the resource exists instead of checking each field" which is pretty much what we have now).

That's fair, and I thought about that. I can honestly see this one both ways. On the one hand, there's a use case for distinguishing between "exists" and "is correct" (test robustness, so test that only need something to exist but don't need its values to be exact still pass when there's a bug in setting one of the values), but also if we're just using the same collection of functions all the time, they may as well be one function to cut down on human error.

> For the point of reusable functions- that's definitely a good practice overall, is there an opportunity for them to be reused at all? If there were, I'd be all for it.

In not-this-case, the general place I see them getting reused is Update checks... but this case doesn't have an Update.

Honestly, I'll defer to your judgement in this case, as you're a bit closer to the context on this one and are better positioned to make a judgement call like that. Also, they're tests, so we can refactor down the line without significant hardship, if necessary. :D

danawillow:
For now I think it makes sense to just do the one function. We can always change it later. I added another commit to make it a bit more compact if you want to take a look.

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -135,53 +133,124 @@ func testAccCheckContainerClusterDestroy(s *terraform.State) error {
...
+			{"zone", cluster.Zone},
+			{"cluster_ipv4_cidr", cluster.ClusterIpv4Cidr},
+			{"description", cluster.Description},
+			{"endpoint", cluster.Endpoint},
+			{"instance_group_urls", cluster.InstanceGroupUrls},
+			{"logging_service", cluster.LoggingService},
+			{"monitoring_service", cluster.MonitoringService},
+			// TODO(danawillow): Add this back in. Currently this field is saved via the config instead of from the API response,
+			// and the config may contain the network name or self_link, whereas the API only returns the self_link.
+			// {"network", cluster.Network},
```
paddyforan: Is there a way to derive the network name from the self_link? If so, could we just normalize on it in the test (i.e., convert both to the network name) so we could actually test it?

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -135,53 +133,124 @@ func testAccCheckContainerClusterDestroy(s *terraform.State) error {
...
 
-func testAccCheckContainerClusterExists(n string) resource.TestCheckFunc {
+func testAccCheckContainerCluster(n string) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
-		rs, ok := s.RootModule().Resources[n]
-		if !ok {
-			return fmt.Errorf("Not found: %s", n)
-		}
-
-		if rs.Primary.ID == "" {
```
paddyforan: May be a good idea to still check this, I'm unsure of what the consequences for not checking it could be.

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -135,53 +133,124 @@ func testAccCheckContainerClusterDestroy(s *terraform.State) error {
...
 	}
 }
 
+func getResourceAttributes(n string, s *terraform.State) (map[string]string, error) {
+	rs, ok := s.RootModule().Resources[n]
+	if !ok {
+		return nil, fmt.Errorf("Not found: %s", n)
+	}
+
+	if rs.Primary.ID == "" {
```
paddyforan: Oh, I see, you moved it down here. This feels like an odd place to put it, since it's not being returned with the Attributes?

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -135,53 +133,124 @@ func testAccCheckContainerClusterDestroy(s *terraform.State) error {
...
+	if tf != gcp {
+		return fmt.Sprintf("Cluster has mismatched %s.\nTF State: %+v\nGCP State: %+v", attr, tf, gcp)
+	}
+	return ""
+}
+
+func checkListMatch(attributes map[string]string, attr string, gcpList []string) string {
+	num, err := strconv.Atoi(attributes[attr+".#"])
+	if err != nil {
+		return fmt.Sprintf("error in number conversion for attribute %s", attr)
```
paddyforan: Would be helpful to return what the error is.

builtin/providers/google/resource_container_cluster_test.go
danawillow: It doesn't really have an obvious place it belongs. The flow of this section of code in the original (and in most resources) seems to be leading towards getting the Attributes, but returning an error if certain preconditions aren't met- the resource isn't there, or has an empty id. From that point on, the only part of the ResourceState that gets used is the Attributes.

builtin/providers/google/resource_container_cluster_test.go
danawillow: Done

builtin/providers/google/resource_container_cluster_test.go
danawillow: Done

danawillow:
Do you think it makes sense (for this PR) to leave these helper methods in the container cluster test code or should I go ahead and move it somewhere common now?

builtin/providers/google/provider.go
```diff
@@ -270,17 +270,20 @@ func getNetworkLink(d *schema.ResourceData, config *Config, field string) (strin
...
+}
 
-		if strings.HasPrefix(network, "https://www.googleapis.com/compute/") {
-			// extract the network name from SelfLink URL
-			networkName := network[strings.LastIndex(network, "/")+1:]
-			if networkName == "" {
-				return "", fmt.Errorf("network url not valid")
-			}
-			return networkName, nil
+func getNetworkNameFromString(network string) (string, error) {
```
paddyforan: Nit: Would it be more helpful to call this `getNetworkNameFromSelfLink` or something similar? The `string` part is implied by the type signature, but what's in the string isn't apparent at first glance.

builtin/providers/google/resource_container_cluster_test.go
paddyforan: Fair enough! My first thought was that the `getResourceAttributes` function would just return the attributes and do nothing else, but in looking at it... that doesn't make much sense, does it? Withdrawn. :)

paddyforan:
> Do you think it makes sense (for this PR) to leave these helper methods in the container cluster test code or should I go ahead and move it somewhere common now?

I'd say leave them there in the name of getting this merged, and let's have a conversation at some point about extracting out common helper methods (possibly as a fix-it friday task).

builtin/providers/google/provider.go
danawillow: Yes, that is a much better name. Done.

stack72:
Hi @danawillow / @paddyforan 

Since the last change (https://github.com/hashicorp/terraform/pull/12147/commits/ad1fb23ac4e390183a29e9454baf6e3c0fcec434), we now have a failing build here

P.

stack72:
ping @danawillow / @paddyforan

paddyforan:
Thanks @stack72. We'll take care of it. :)

paddyforan:
Closed and reopened due to Travis testing the wrong commit. Hopefully this forces an update to use the latest commit.

danawillow:
Thanks @stack72, I was OOO all last week but I'm glad to be finally getting this change in now.

Merging.

PR #12155, opened 2017-02-22
command: refresh state in old commands for backend

### LINKED ISSUE #12148: Experimental remote backend feature seems to break "taint" and "state" commands
### Terraform Version
`Terraform v0.9.0-dev`
`commit: 757b83aeebffa9c19ea3982ed62124511b910ff2`

### Expected Behavior
When using a remote backend (in my case, it is consul) commands such as `taint`, and `state` should work without having to pass a "-state" flag. They should also work if you do pass a state flag.

### Actual Behavior
`terraform state list` output:

```
$ terraform state list
[WARN] /Users/tomwganem/go/bin/terraform-provider-softlayer overrides an internal plugin for softlayer-provider.
No state file was found!

State management commands require a state file. Run this command
in a directory where Terraform has been run or use the -state flag
to point the command to a specific state location.
```
`terraform state list -state terraform.tfstate.backup` output:

```
$ terraform state list -state terraform.tfstate.backup
[WARN] /Users/tomwganem/go/bin/terraform-provider-softlayer overrides an internal plugin for softlayer-provider.
No state file was found!

State management commands require a state file. Run this command
in a directory where Terraform has been run or use the -state flag
to point the command to a specific state location.
```

`terraform taint` output
```
$ terraform taint chef_environment.environment
[WARN] /Users/tomwganem/go/bin/terraform-provider-softlayer overrides an internal plugin for softlayer-provider.
The state is empty. The most common reason for this is that
an invalid state file path was given or Terraform has never
 been run for this infrastructure. Infrastructure must exist
for it to be tainted.
```

### Steps to Reproduce
1. set a backend for your terraform project
```
terraform {

  backend "consul" {
    path =  "files/dev/terraform"
  }
}
```

and have any number of resources

```
resource "chef_environment" "environment" {
  name                     = "${var.mesos_config["chef_environment_name"]}"
  override_attributes_json = "${data.template_file.override.rendered}"
  default_attributes_json  = "${data.template_file.default.rendered}"
  cookbook_constraints     = "${module.cookbook_versions.berksfile}"
}
```

2. `terraform init`

```
$ terraform init
[WARN] /Users/tomwganem/go/bin/terraform-provider-softlayer overrides an internal plugin for softlayer-provider.
Downloading modules (if any)...
Get: file:///Users/tomwganem/Development/tf-files/aspera-files
Get: file:///Users/tomwganem/Development/as-mesos-?terrterraform/softlayer
Get: file:///Users/tomwganem/Development/cookbook-aspera-files
Initializing the backend...

should now work.

```

3. `terraform apply`

```
module.aspera-files.chef_environment.environment: Creating...
...
```

4. `terraform state list ` or `terraform taint ${resource}`. Specifying a state file doesn't matter

```
$ terraform state list -state  terraform.tfstate.backup
[WARN] /Users/tomwganem/go/bin/terraform-provider-softlayer overrides an internal plugin for softlayer-provider.
No state file was found!

State management commands require a state file. Run this command
in a directory where Terraform has been run or use the -state flag
to point the command to a specific state location.
```

mitchellh:
Fixes #12148 

The Backend API doesn't automatically `RefreshState` for you like the old Meta State API did. We have to manually do this across all commands.

I considered making it a requirement to do a refresh but we call State() multiple times on a backend and the point of _not_ refreshing it was so that calling State() was cheap. I think its worth it to keep it that way.

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -207,6 +207,7 @@ func (c *Context) Graph(typ GraphType, opts *ContextGraphOpts) (*Graph, error) {
 		opts = &ContextGraphOpts{Validate: true}
 	}
 
+	log.Printf("[INFO] terraform: building graph: %s", typ)
 	switch typ {
 	case GraphTypeApply:
 		return (&ApplyGraphBuilder{
--- a/backend/local/backend.go
+++ b/backend/local/backend.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/backend"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/state"
@@ -111,11 +110,6 @@ func (b *Local) State() (state.State, error) {
 	var s state.State = &state.LocalState{
 		Path:    b.StatePath,
 		PathOut: b.StateOutPath,
-	}
-
-	// Load the state as a sanity check
-	if err := s.RefreshState(); err != nil {
-		return nil, errwrap.Wrapf("Error reading local state: {{err}}", err)
 	}
 
 	// If we are backing up the state, wrap it
--- a/command/meta_backend_test.go
+++ b/command/meta_backend_test.go
@@ -103,6 +103,9 @@ func TestMetaBackend_emptyWithDefaultState(t *testing.T) {
 	if err != nil {
 		t.Fatalf("bad: %s", err)
 	}
+	if err := s.RefreshState(); err != nil {
+		t.Fatalf("err: %s", err)
+	}
 	if actual := s.State().String(); actual != testState().String() {
 		t.Fatalf("bad: %s", actual)
 	}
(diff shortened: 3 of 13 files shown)


terraform/context.go
```diff
@@ -207,6 +207,7 @@ func (c *Context) Graph(typ GraphType, opts *ContextGraphOpts) (*Graph, error) {
 		opts = &ContextGraphOpts{Validate: true}
 	}
 
+	log.Printf("[INFO] terraform: building graph: %s", typ)
```
mitchellh: Just added this while we're here since the debug log was unclear what graph was being built. The old graph builders used to do this.

jbardin:
If `Backend.State` is not guaranteed to load the state, which is how it's documented in the interface comments, I think we should make the `Local` backend follow this as strictly as possible too. For example, if I remove the automatic "sanity check" `RefreshState` call from `Local.State()`, we get more state command test failures (though some may simply be failures in tests themselves, I haven't checked). 

Having the most commonly used Local backend instance follow the spec as strictly as possible helps ensure that we don't get any divergence with remote backends which are inherently less tested.

mitchellh:
@jbardin This is a really good idea, doing it now.

mitchellh:
@jbardin Done. The actual number of test failures was "0" in the sense that the failures introduced were actually in the tests themselves. The logic is sound after this PR. Pushed the last commit!

PR #12156, opened 2017-02-22
command/state: mv and rm -backup works

### LINKED ISSUE #12154: `-backup` option for `terraform state mv` seems not to work correctly
I'm not sure it's a spec or bug, but I think this behavior is different from document's specification.
https://www.terraform.io/docs/commands/state/mv.html#_backup_path

### Terraform Version
Terraform v0.8.7

### Affected Resource(s)
- My example is `aws_vpc` but I think it's core issue.

```hcl
provider "aws" {
  region = "us-east-1"
  max_retries = 3
}

resource "aws_vpc" "example" {
  cidr_block = "10.1.0.0/16"
  tags { Name = "example" }
}
# I want to move the above resource to the module
# module "vpc" {
#  source = "./vpc"
#}

```

### Debug Output
```
sishimura% TF_LOG=debug terraform state mv -backup=mybackup.tfstate aws_vpc.example module.vpc.aws_vpc.example                                                                                                                               [kamatama-test](master)
2017/02/22 12:44:55 [INFO] Terraform version: 0.8.7  c1df88ac4633b61b83802bbb10c343c20d3035b5
2017/02/22 12:44:55 [INFO] CLI args: []string{"/Users/shinichi/IdeaProjects/tfenv/versions/0.8.7/terraform", "state", "mv", "-backup=mybackup.tfstate", "aws_vpc.example", "module.vpc.aws_vpc.example"}
2017/02/22 12:44:55 [DEBUG] Detected home directory from env var: /Users/shinichi
2017/02/22 12:44:55 [DEBUG] Detected home directory from env var: /Users/shinichi
2017/02/22 12:44:55 [DEBUG] Attempting to open CLI config file: /Users/shinichi/.terraformrc
2017/02/22 12:44:55 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2017/02/22 12:44:55 [DEBUG] Detected home directory from env var: /Users/shinichi
2017/02/22 12:44:55 [DEBUG] plugin: waiting for all plugin processes to complete...
Moved aws_vpc.example to module.vpc.aws_vpc.example
```
### Panic Output
No panic

### Expected Behavior
`terraform.tfstate` should be updated and `mybackup.tfstate` should be generated as a backup file.

### Actual Behavior
`mybackup.tfstate` was generated as a new state file and `mybackup.tfstate.1487735095.backup` was generated as a backup file

### Steps to Reproduce
1. `terraform apply`
2. `terraform state mv -backup=mybackup.tfstate aws_vpc.example module.vpc.aws_vpc.example`

### Important Factoids
Nothing

### References
Nothing

mitchellh:
Fixes #12154

The "-backup" flag before for "state *" CLI had some REALLY bizarre behavior:
it would change the _destination_ state and actually not create any
additional backup at all (the original state was unchanged and the
normal timestamped backup still are written). Really weird.

This PR makes the -backup flag work as you'd expect with one caveat:
we'll _still_ create the timestamped backup file. The timestamped backup
file helps make sure that you always get a backup history when using
these commands. We don't want to make it easy for you to overwrite a
state with the `-backup` flag.

This can be backported.

### DIFF
--- a/command/state_rm.go
+++ b/command/state_rm.go
@@ -16,9 +16,8 @@ type StateRmCommand struct {
 func (c *StateRmCommand) Run(args []string) int {
 	args = c.Meta.process(args, true)
 
-	var backupPath string
 	cmdFlags := c.Meta.flagSet("state show")
-	cmdFlags.StringVar(&backupPath, "backup", "", "backup")
+	cmdFlags.StringVar(&c.Meta.backupPath, "backup", "-", "backup")
 	cmdFlags.StringVar(&c.Meta.statePath, "state", DefaultStateFilename, "path")
 	if err := cmdFlags.Parse(args); err != nil {
 		return cli.RunResultHelp
@@ -75,7 +74,8 @@ Options:
   -backup=PATH        Path where Terraform should write the backup
                       state. This can't be disabled. If not set, Terraform
                       will write it to the same path as the statefile with
-                      a backup extension.
+                      a backup extension. This backup will be made in addition
+                      to the timestamped backup.
 
   -state=statefile    Path to a Terraform state file to use to look
                       up Terraform-managed resources. By default it will
--- a/command/state_meta.go
+++ b/command/state_meta.go
@@ -17,9 +17,6 @@ type StateMeta struct{}
 // in the way that backups are done. This configures backups to be timestamped
 // rather than just the original state path plus a backup path.
 func (c *StateMeta) State(m *Meta) (state.State, error) {
-	// Disable backups since we wrap it manually below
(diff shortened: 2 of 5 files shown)


command/state_rm.go
```diff
@@ -16,9 +16,8 @@ type StateRmCommand struct {
 func (c *StateRmCommand) Run(args []string) int {
 	args = c.Meta.process(args, true)
 
-	var backupPath string
```
mitchellh: This is so bizarre cause this var was never actually used for anything. Its like I just forgot.

PR #12162, opened 2017-02-22
Improve description of consul_catalog_entry

blalor:
The other resources for this provider are similarly misleading (they read like data sources).

Also adding undocumented properties of the resource.


stack72:
LGTM - thanks for the changes here!

PR #12164, opened 2017-02-22
provider/azurerm: Auto base64encode virtual_machine custom data

stack72:
Reported by @sethvargo - we auto encode for AWS, we should follow a
similar pattern for Azure.

In order to escape the double encoding, we check that it's not already
encoded before encoding

### DIFF
--- a/website/source/docs/providers/azurerm/r/virtual_machine.html.markdown
+++ b/website/source/docs/providers/azurerm/r/virtual_machine.html.markdown
@@ -266,7 +266,7 @@ For more information on the different example configurations, please check out t
 * `computer_name` - (Required) Specifies the name of the virtual machine.
 * `admin_username` - (Required) Specifies the name of the administrator account.
 * `admin_password` - (Required) Specifies the password of the administrator account.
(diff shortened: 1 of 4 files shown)


website/source/docs/providers/azurerm/r/virtual_machine.html.markdown
```diff
@@ -266,7 +266,7 @@ For more information on the different example configurations, please check out t
 * `computer_name` - (Required) Specifies the name of the virtual machine.
 * `admin_username` - (Required) Specifies the name of the administrator account.
 * `admin_password` - (Required) Specifies the password of the administrator account.
-* `custom_data` - (Optional) Specifies a base-64 encoded string of custom data. The base-64 encoded string is decoded to a binary array that is saved as a file on the Virtual Machine. The maximum length of the binary array is 65535 bytes.
+* `custom_data` - (Optional) Specifies a base-64 encoded string of custom data. The base-64 encoded string is decoded to a binary array that is saved as a file on the Virtual Machine. The maximum length of the binary array is 65535 bytes. Terraform will auto base64encode this string for you.
```
sethvargo: I think we should update this to be a bit clearer. I recommend:

> Specifies custom data to supply to the machine. On linux-based systems, this can be used as a cloud-init script. On other systems, this will be copied as a file on disk. Internally, Terraform will base64 encode this value before sending it to the API. The maximum length of the binary array is 65535 bytes.

PR #12165, opened 2017-02-22
Small doc updates

directionless:
Changed the comment about configuring the provider

Replaced the `data` example, with a correct `resource` example.


stack72:
LGTM thanks for the work here

PR #12173, opened 2017-02-22
providers/terraform: remote state data source supports backends

mitchellh:
This changes the `terraform_remote_state` data source to support backends rather than going through the legacy remote state interface. 

To support this change without duplicating the list of backends, I introduce a new package `backend/init` which simply provides an API for looking up backend initializers. The `command` package was changed to use this as well. This means that we have a uniform way to lookup and initialize a backend.

### DIFF
--- a/backend/init/init.go
+++ b/backend/init/init.go
@@ -1,0 +1,55 @@
+// Package init contains the list of backends that can be initialized and
+// basic helper functions for initializing those backends.
+package init
+
+import (
+	"sync"
+
+	"github.com/hashicorp/terraform/backend"
+
+	backendlegacy "github.com/hashicorp/terraform/backend/legacy"
+	backendlocal "github.com/hashicorp/terraform/backend/local"
+	backendconsul "github.com/hashicorp/terraform/backend/remote-state/consul"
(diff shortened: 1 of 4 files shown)


backend/init/init.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"sync"
+
+	"github.com/hashicorp/terraform/backend"
+
+	backendlegacy "github.com/hashicorp/terraform/backend/legacy"
+	backendlocal "github.com/hashicorp/terraform/backend/local"
+	backendconsul "github.com/hashicorp/terraform/backend/remote-state/consul"
+	backendinmem "github.com/hashicorp/terraform/backend/remote-state/inmem"
```
jbardin: not to block this PR, but I wonder if we need some general refactoring/cleanup regarding around the backend packages. some of these import blocks are getting messy.

mitchellh: I don't think so since its no different than our `internal_plugin_list.go` file. That file happens to be generated now (it used to not!), but its just like this. The full list of built-in things has to be enumerated _somewhere_ and `backend/init` is where that's happening. I think as long as we do this only in one place we're good to go, cause its going to happen somewhere anyways.

PR #12182, opened 2017-02-22
Environments 

jbardin:
This starts support for "environments" in the backends.

Environments, aka "named states", are optionally supported by Backend implementations. Because they are optional, the corresponding method set has been split out into its own interface:

```
// MultiState is an interface that a backend can implement to allow changing
// between named states depending on the configured environment.
type MultiState interface {
	// States returns a list of configured named states and the current state.
	States() ([]string, string, error)

	// ChangeState changes to the named state. If the named state doesn't exist
	// it will be created.
	ChangeState(name string) error

	// DeleteState removes the named state if it exists. If the current state is
	// deleted, the backend should change to the default state. It is an error
	// to delete the default state.
	DeleteState(name string) error
}
```

This diverges slightly from the RFC in a few ways. 
- There's no `name` argument added to the State method. This would change a _lot_ of code, and since the backend must already know about the current state, it is mostly superfluous. The primary drawback is that it requires a call to `ChangeState` to fetch a particular state.
- `States()` returns the list of known states, and the current state. The alleviates the `state.State` from having to know about its name.
- Added `DeleteState`.

An "environment" corresponds to a named state. Environments can be manipulated through the new `terraform env` command. This is implemented as a single command with flags for convenience, but could be refactored into subcommands if it improves the UI.

```
Usage: terraform env [options] [NAME]

  By default env will list all configured environments. If NAME is provided,
  env will change into to that named environment.


Options:

  -new=name      Create a new environment.
  -delete=name   Delete an existing environment,

  -state=path    Used with -new to copy a state file into the new environment.
  -force         Used with -delete to remove a non-empty environment.
```

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -27,13 +27,22 @@ type Backend interface {
 	// to load the state. If the state.State is a state.Locker, it's up to the
 	// caller to call Lock and Unlock as needed.
 	State() (state.State, error)
+}
 
+// MultiState is an interface that a backend can implement to allow changing
+// between named states depending on the configured environment.
+type MultiState interface {
 	// States returns a list of configured named states and the current state.
 	States() ([]string, string, error)
 
-	// ChangeState changes to the named state. If this doesn't exist it'll be
-	// created.
+	// ChangeState changes to the named state. If the named state doesn't exist
+	// it will be created.
 	ChangeState(name string) error
+
+	// DeleteState removes the named state if it exists. If the current state is
+	// deleted, the backend should change to the default state. It is an error
+	// to delete the default state.
+	DeleteState(name string) error
 }
 
 // Enhanced implements additional behavior on top of a normal backend.
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -12,6 +12,8 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
+const DefaultStateName = "default"
+
 // Backend is the minimal interface that must be implemented to enable Terraform.
 type Backend interface {
 	// Ask for input and configure the backend. Similar to
@@ -25,6 +27,13 @@ type Backend interface {
 	// to load the state. If the state.State is a state.Locker, it's up to the
 	// caller to call Lock and Unlock as needed.
 	State() (state.State, error)
+
+	// States returns a list of configured named states and the current state.
+	States() ([]string, string, error)
+
+	// ChangeState changes to the named state. If this doesn't exist it'll be
+	// created.
+	ChangeState(name string) error
 }
 
 // Enhanced implements additional behavior on top of a normal backend.
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -26,23 +26,21 @@ type Backend interface {
 	// not be loaded locally: the proper APIs should be called on state.State
 	// to load the state. If the state.State is a state.Locker, it's up to the
 	// caller to call Lock and Unlock as needed.
-	State() (state.State, error)
-}
+	//
+	// If the named state doesn't exist it will be created. The "default" state
+	// is always assumed to exist.
+	State(name string) (state.State, error)
 
-// MultiState is an interface that a backend can implement to allow changing
-// between named states depending on the configured environment.
-type MultiState interface {
-	// States returns a list of configured named states and the current state.
-	States() ([]string, string, error)
-
-	// ChangeState changes to the named state. If the named state doesn't exist
-	// it will be created.
-	ChangeState(name string) error
-
-	// DeleteState removes the named state if it exists. If the current state is
-	// deleted, the backend should change to the default state. It is an error
+	// DeleteState removes the named state if it exists. It is an error
 	// to delete the default state.
+	//
+	// DeleteState does not prevent deleting a state that is in use. It is the
+	// responsibility of the caller to hold a Lock on the state when calling
+	// this method.
 	DeleteState(name string) error
+
+	// States returns a list of configured named states.
+	States() ([]string, error)
 }
 
 // Enhanced implements additional behavior on top of a normal backend.
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -123,6 +123,9 @@ type Operation struct {
 	// If LockState is true, the Operation must Lock any
 	// state.Lockers for its duration, and Unlock when complete.
 	LockState bool
+
+	// Environment is the named state that should be loaded from the Backend.
+	Environment string
 }
 
 // RunningOperation is the result of starting an operation.
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -6,6 +6,7 @@ package backend
 
 import (
 	"context"
+	"errors"
 
 	"github.com/hashicorp/terraform/config/module"
 	"github.com/hashicorp/terraform/state"
@@ -13,6 +14,9 @@ import (
 )
 
 const DefaultStateName = "default"
+
+// Error value to return when a named state operation isn't supported
+var ErrNamedStatesNotSupported = errors.New("named states not supported")
 
 // Backend is the minimal interface that must be implemented to enable Terraform.
 type Backend interface {
--- a/backend/local/backend.go
+++ b/backend/local/backend.go
@@ -2,7 +2,13 @@ package local
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
 	"sync"
 
 	"github.com/hashicorp/terraform/backend"
@@ -11,6 +17,14 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/mitchellh/cli"
 	"github.com/mitchellh/colorstring"
+)
+
+const (
+	DefaultEnvDir          = "terraform.tfstate.d"
+	DefaultEnvFile         = "environment"
+	DefaultStateFilename   = "terraform.tfstate"
+	DefaultDataDir         = ".terraform"
+	DefaultBackupExtension = ".backup"
 )
 
 // Local is an implementation of EnhancedBackend that performs all operations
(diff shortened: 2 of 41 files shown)


jbardin:
@mitchellh, 

After discussing with @apparentlymart, we agree that subcommands are a cleaner solution than command flags and flag-specific flags. The awkwardness seems to be around the bare `env` command taking an argument _or_ a subcommand, which makes it less intuitive and has the unfortunate possibility of subcommand and environment name collision (without also supporting the non-obvious `--` end of options delimiter). 

How about the following command set?

```
- env: print's usage
   |- list: list environments
   |- new: create new env, and change to that env
   |- select: change to the named env
   |- delete: remove the selected env
```

I'm not opposed to `list` being the default command for `env`, but I do like the immediacy of printing the usage for new users, rather than requiring the `-h` option.

apparentlymart:
This subcommand-based interface makes good sense to me. Perhaps a reasonable compromise on the bare `terraform env` command would be for it to produce both status information and usage, to help explain the context:

```
Current Environment is production

Subcommands:
  list: list all environments
  select <name>: select the given environment for subsequent operations
  new <name>: create a new environment and select it
  delete <name>: delete the given environment
```

This way the output tells us where we are *and* where we can go from here. I presume here that running e.g. `terraform env new` or `terraform env new -h` would provide a more detailed version of the usage of that command, so we don't need to overwhelm the user with all the flag details in the initial output.

Some of the subcommand flags could be progressively discovered through good error messages. For example:

```
$ terraform env delete production
The environment "production" contains resources.
Deleting it will cause Terraform to lose track of these resources.

Either select this environment and destroy the resources first, or use the -force
option to delete Terraform's record of these resources without destroying them
first.

After using -force it will be necessary to use "terraform import" to bring
these resources back under Terraform's management.
```

The `-state` one isn't really a response to an error, so there's not such a good path to discovering that one, but presumably importing an existing local state as an environment is a rare enough situation that it's fine to make that one be discoverable only by explicitly looking in the docs/usage, just like the `-state` options on many other Terraform commands.

jbardin:
This last commit splits the the env command into subcommands. I used apparentlymart's suggestion and have `env` print the current environment along with the usage text.

jbardin:
Pushed with the new subcommands, but I'm retracting the current env output from the bare `env` command. Because a config directory could be provided as an argument, there could again be a conflict with subcommands and paths. It's not impossible to add, but I'm leaning towards simplicity for now.

jonmorehouse:
@jbardin this is a really interesting concept; you mentioned an RFC in your original description is that something shareable here? I'd personally love to read a bit more about the backstory of this and intended use cases etc.

Thanks!

jbardin:
@jonmorehouse,

Sorry, I don't have anything I can share, but at this point this PR probably contains more relevant info than the original RFC ;). 

The immediate use case for most developers is replacing the workflow of having multiple state files manually managed and specified at runtime, to a workflow of using environments to manage those state files in a less error-prone way. 

I think the only missing concept here is that "Enhanced Backends" will have the ability to extend the idea of an "environment" further, as they handle the Terraform operations directly. What this means precisely will depend on the backend implementation.

backend/local/backend.go
```diff
@@ -61,6 +79,10 @@ type Local struct {
 	schema *schema.Backend
 	opLock sync.Mutex
 	once   sync.Once
+
+	// workingDir is where the State* paths should be relative to.
+	// This is currently only used for tests.
+	workingDir string
```
mitchellh: I'd avoid this and use `testChdir`. We actually USED to have a field like this but I've tried to make helpers to make sure we're executing code identical as much as possible (we're not perfect yet).

backend/backend.go
```diff
@@ -27,6 +29,22 @@ type Backend interface {
 	State() (state.State, error)
 }
 
+// MultiState is an interface that a backend can implement to allow changing
+// between named states depending on the configured environment.
+type MultiState interface {
```
mitchellh: I'd prefer to make this required for all states (make it part of Backend).

For legacy remote state we can return an error in the actual `remote-state.Backend` implementation instead.

The reason being is that going forward I don't want this to be optional and introduce more optional interfaces. It is difficult now due to backwards compatibility but if we force all new remote implementations to support it and only error for legacy then we can slowly fix them over one by one as we convert to the new API.

I believe that should be fully backwards compatible and simplifies all the checking.

backend/backend.go
```diff
@@ -27,6 +29,22 @@ type Backend interface {
...
+	// States returns a list of configured named states and the current state.
+	States() ([]string, string, error)
```
mitchellh: The backend shouldn't be responsible for "current" state since that should be local to the user (i.e. with the same remote backend on two different machines one user can point to one env and another can point to another. The backend itself has no concept of "current").

The "current" state should be stored locally in the terraform cache dir (.terraform so it is VCS ignored).

backend/local/backend.go
```diff
@@ -97,6 +119,150 @@ func (b *Local) Configure(c *terraform.ResourceConfig) error {
...
+			return b.States()
+		} else {
+			return nil, "", ErrEnvNotSupported
+		}
+	}
+
+	// the listing always start with "default"
+	envs := []string{backend.DefaultStateName}
+
+	current, err := b.currentStateName()
```
mitchellh: Same as above, "current" state should be a TF core thing and backends should just be told directly what state to load. This will simplify backend implementation a lot and shift that complexity into core but at least its shared.

backend/local/backend.go
```diff
@@ -97,6 +119,150 @@ func (b *Local) Configure(c *terraform.ResourceConfig) error {
...
+		return errors.New("cannot delete default state")
+	}
+
+	_, current, err := b.States()
+	if err != nil {
+		return err
+	}
+
+	// if we're deleting the current state, we change back to the default
+	if name == current {
```
mitchellh: Deleting the current state isn't allowed (as per the spec, we have to show an error and tell the user to switch to another state first). This protects the user by making it so they 1.) must have another state to go to and 2.) can't easily delete states.

backend/remote-state/backend.go
```diff
@@ -46,6 +47,14 @@ func (b *Backend) Configure(rc *terraform.ResourceConfig) error {
 	return b.Backend.Configure(rc)
 }
 
+func (b *Backend) States() ([]string, string, error) {
+	return []string{backend.DefaultStateName}, backend.DefaultStateName, nil
```
mitchellh: As per the comments above: this is where I'd move the "multiple states not supported" errors.

backend/local/backend.go
jbardin: Yeah, I agree. This was solely to check be able to differentiate 2 backends in a test, where I should have made a test implementation instead.

backend/local/backend.go
jbardin: Oh yeas, I missed that and figured it OK since state was protected if it's not empty. Easy change.

backend/backend.go
jbardin: I like to err on the side of smaller interfaces (isn't there some tao of Go quote about small interfaces? 😉 ) Though if we do want to enforce this it's not too unwieldy and not nearly the largest in terraform.

backend/backend.go
jbardin: I like to err on the side of smaller interfaces (isn't there some tao of Go quote about small interfaces? 😉 ) Though if we do want to enforce this it's not too unwieldy and not nearly the largest in terraform.

backend/backend.go
jbardin: Ok, this makes sense, and I clears up part of what I wanted to cover in this first review. The original spec called for a remote ChangeState method, which implied the backend also managed the current state. Deferring to the backend was much less intrusive in the terraform code, not even accounting for possible synchronization issues between the local environment and backend.  Removing the backend from being aware of "current" at all makes this work much better.

backend/backend.go
mitchellh: It may be, but I also think an interface should represent the full set of functionality required to make it useful. In the context of Terraform, I believe that is loading AND changing states, so splitting it into two required interface seems more of a mental burden than having one required interface. 

So I think it should be part of the one interface.

backend/backend.go
mitchellh: Ah I can see how that is confusing. Sorry!

Given that API, maybe the behavior should be to remove `ChangeState` and just have `State(name)`, `DeleteState(name)`, `States() []string error`?

jbardin: Yes, that's what I was about to implement. I can't think of a reason to require a separate `CreateState(string)` method, when `State` could handle it.

mitchellh: Agreed, and state existence when switching to it can live in the CLI itself. I think thats still important to avoid a typo when switching states.

backend/backend.go
jbardin: Yup, advocating the alternative if only to keep ourselves in line ;)

command/env_delete.go
```diff
@@ -0,0 +1,139 @@
...
+	hasResources := sMgr.State().HasResources()
+
+	if hasResources && !force {
+		c.Ui.Error(fmt.Sprintf(envNotEmpty, delEnv))
+		return 1
+	}
+
+	// Lock the state if we can
+	lockInfo := state.NewLockInfo()
+	lockInfo.Operation = "env new"
```
mitchellh: "env delete" probably

command/state_meta.go
```diff
@@ -23,8 +25,9 @@ func (c *StateMeta) State(m *Meta) (state.State, error) {
 		return nil, err
 	}
 
+	env := c.Env()
```
mitchellh: May want to be using `m.Env()` here. Practically I don't think it makes a difference currently but maybe one day we may pass in a differently configured meta here?

jbardin: Ah, ok. This was the reason for your question above, in that StateMeta was calling the embedded Meta.Env method, and embedding Meta in StateMeta was just the cleanest way to ensure all State commands also had the same method. 

Looking at this though, I think we can also just drop the Meta param, since `StateMeta` would now have the correctly configured Meta for the command embedded.

jbardin: Oh, actually there is no more Meta.State with the new backends, so this structure probably isn't needed at all.

jbardin: I finally see what the second Meta is for now. I'm going to split the Meta back out just to alleviate the confusion, since `StateMeta` doesn't it need.

jbardin:
Fixed lock reason for `env delete` and split Meta back out of StateMeta since it turns out StateMeta doesn't need the Env method itself.

PR #12188, opened 2017-02-23
provider/aws: Lambda DeadLetterConfig support

### LINKED ISSUE #10630: DeadLetterConfig support for AWS Lambda
According to the [AWS docs](http://docs.aws.amazon.com/lambda/latest/dg/dlq.html) , it is possible to specify dead letter queues for AWS Lambda function. If Terraform could support this parameter, we could directly send the failed events to the dead letter queue rather than manually publishing it ourselves as part of the Lambda function

apparentlymart:
This feature allows sending a notification to either an SQS queue or an SNS topic when an error occurs running an AWS Lambda function.

This fixes #10630.

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -58,6 +58,22 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"dead_letter_config": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				MinItems: 0,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"target_arn": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateArn,
+						},
+					},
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_lambda_function.go
```diff
@@ -58,6 +58,21 @@ func resourceAwsLambdaFunction() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				MinItems: 0,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"target_arn": {
+							Type:     schema.TypeString,
+							Required: true,
```
tomelliff: Is it worth adding `ValidateFunc: validateArn,` here?

tomelliff:
I was just starting to write a pull request for this for some Lambda work I'm doing right now so that's amazingly good timing :)

builtin/providers/aws/resource_aws_lambda_function.go
apparentlymart: Oh yes, good idea!

pulveni1:
Hi
Do you plan to make this feature available soon?
Thanks
Nico

PR #12189, opened 2017-02-23
aws/provider: what was the value of the cluster id which caused the problem

netjunki:
Found tracking down the exact source of this error:

    aws_elasticache_cluster.Cluster: "cluster_id" must contain from 1 to 20 alphanumeric characters or hyphens

So I've extended the error messages to include the value of the problematic cluster id:

  aws_elasticache_cluster.Cluster: "cluster_id" ("thisisaverylongclusterid") must contain from 1 to 20 alphanumeric characters or hyphens

Added this data to the other errors as well since I was in there. :-)


josephholsten:
Looks good to me.

PR #12190, opened 2017-02-23
provider/openstack: openstack_compute_floatingip_associate_v2 resource

jtopjian:
This commit adds the openstack_compute_floatingip_associate_v2
resource which specifically handles associating a floating IP
address to an instance. This can be used instead of the existing
floating_ip options in the openstack_compute_instance_v2 resource.


stack72:
LGTM! Thanks :)

PR #12192, opened 2017-02-23
Added missing attribute refrence

viktorvoltaire:


stack72:
Hi @viktorvoltaire 

thanks for taking the time here to make the update to the documentation. One of the things that isn't that clear with our documentation is that parameters that get passed to a resource are already available as reference attributes for other resources

Therefore, as lambda_function takes a function_name, we can already use that function_name parameter in the lambda_permission resource

Hope this helps

Thanks

Paul

PR #12197, opened 2017-02-23
Profitbricks primary nic fix

jasminSPC:
A fix that handles how `primary_nic` is set. 
Removed some unused functions while fixing it


stack72:
Thanks for this :)

Issue #11844, opened 2017-02-10
aws_ecs_service placement_strategy documentation update


mdkent:
Quick one here: the aws_ecs_service placement_strategy response from the API was lowercased in #11644, but the documentation [here](https://www.terraform.io/docs/providers/aws/r/ecs_service.html#placement_strategy) says that `instanceId` is a valid value, which doesn't work when it gets lowercased. Suggest updating the docs to `instanceid`, or stop lowercasing this value.

### Terraform Version
0.8.6.

### Affected Resource(s)
- aws_ecs_service

### Terraform Configuration Files
```hcl
resource "aws_ecs_service" "service_test" {
  name            = "test"
  cluster         = "${aws_ecs_cluster.ecs.id}"
  desired_count   = "${var.main_cluster["autoscale_min"]}"
  task_definition = "${aws_ecs_task_definition.task_test.arn}"

  deployment_minimum_healthy_percent = 50
  deployment_maximum_percent = 100

  placement_strategy {
    type  = "spread"
    field = "instanceId"
  }
}
```

### Plan Output
```
-/+ module.ecs.aws_ecs_service.service_test
    cluster:                             "arn:aws:ecs:us-east-1:178556630403:cluster/test-ecs" => "arn:aws:ecs:us-east-1:178556630403:cluster/test-ecs"
    deployment_maximum_percent:          "100" => "100"
    deployment_minimum_healthy_percent:  "50" => "50"
    desired_count:                       "10" => "10"
    name:                                "test" => "test"
    placement_strategy.#:                "1" => "1"
    placement_strategy.1676812570.field: "instanceid" => "" (forces new resource)
    placement_strategy.1676812570.type:  "spread" => "" (forces new resource)
    placement_strategy.3946258308.field: "" => "instanceId" (forces new resource)
    placement_strategy.3946258308.type:  "" => "spread" (forces new resource)
    task_definition:                     "arn:aws:ecs:us-east-1:178556630403:task-definition/test:8" => "arn:aws:ecs:us-east-1:178556630403:task-definition/test:8"
```

### Expected Behavior
Nothing, idempotent plan.

### Actual Behavior
terraform wants to recreate the service on each run.

### Steps to Reproduce
1. `terraform push`

### References
- #11644

mdkent:
Ugh, this goes a bit deeper, using `instanceid` on a newly created service results in

```
* aws_ecs_service.service_large_logspout: InvalidParameterException: spread field 'instanceid' is invalid.
[...]

Issue #11644, opened 2017-02-02
ECS Service Placement Strategy Forcing Service Rebuild


hikerspath:
We are observing a rebuild of ECS services that are making use of the new placement_strategy options.  Basically seems to be a case-test issue if not something more.  The app only accepts lowercase field values as your [documentation](https://www.terraform.io/docs/providers/aws/r/ecs_service.html) was updated to state (https://github.com/hashicorp/terraform/pull/11338).  However there appears to be a case mis-match that is causing the service to rebuild on a static (no changes to apply) run of terraform apply.

### Terraform Version

```
$ terraform -v
Terraform v0.8.5
```

### Affected Resource:

- aws_ecs_service

### Terraform Configuration:

```hcl
resource "aws_ecs_service" "test" {
  name            = "test-ecs-service"
  cluster         = "${aws_ecs_cluster.main.id}"
  task_definition = "${aws_ecs_task_definition.test.arn}"
  desired_count   = 1
  iam_role        = "${aws_iam_role.ecs_service.name}"

  placement_constraints {
    type = "memberOf"
    expression = "attribute:ecs.availability-zone in [${var.availability_zones}]"
  }

  placement_strategy {
    type = "binpack"
    field = "memory"
  }

  load_balancer {
    target_group_arn = "${aws_alb_target_group.test.id}"
    container_name   = "${data.template_file.task_definition.vars.container_name}"
    container_port   = "REDACTED"
  }

  depends_on = [
    "aws_iam_role_policy.ecs_service",
    "aws_alb_listener.front_end",
  ]
}
```

### Debug:

```
-/+ aws_ecs_service.test
    cluster:                                     "arn:aws:ecs:eu-west-1:__REDACTED__:cluster/test-ecs-cluster" => "arn:aws:ecs:eu-west-1::__REDACTED__:cluster/test-ecs-cluster"
    deployment_maximum_percent:                  "200" => "200"
    deployment_minimum_healthy_percent:          "100" => "100"
    desired_count:                               "1" => "1"
    iam_role:                                    "test-ecs-role" => "test-ecs-role"
    load_balancer.#:                             "1" => "1"
    load_balancer.547002511.container_name:      "oauth2-proxy" => "oauth2-proxy"
    load_balancer.547002511.container_port:      ":__REDACTED__" => ":__REDACTED__"
    load_balancer.547002511.elb_name:            "" => ""
[...]

PR #12199, opened 2017-02-23
Fix AWS ECS placement strategy spread fields

lwilliams-oats:
Fix for #11844 

AWS API requires ECS placement strategies "field" attribute to be
"memory" or "cpu" (lowercase) when type=bin, but these read back as
"MEMORY" and "CPU" (uppercase) respectively.

PR #11565 (which fixed separately reported #11644) deals with this by
always lowering the case of the resource received from the API, but this
breaks for other "field" values (e.g. "instanceId" -> "instanceid").

This PR only lowers the case of the returned resource when field
"MEMORY" or "CPU". Haven't checked if any other fields need this
treatment.

### DIFF
--- a/builtin/providers/aws/resource_aws_ecs_service.go
+++ b/builtin/providers/aws/resource_aws_ecs_service.go
@@ -357,7 +357,13 @@ func flattenPlacementStrategy(pss []*ecs.PlacementStrategy) []map[string]interfa
 	for _, ps := range pss {
 		c := make(map[string]interface{})
 		c["type"] = *ps.Type
-		c["field"] = strings.ToLower(*ps.Field)
+		c["field"] = *ps.Field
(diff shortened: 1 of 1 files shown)


stack72:
Hi @jmahowald 

Please can you reopen this against master? We have just killed off the 0-8-stable branch

Thanks

Paul

jmahowald:
@stack72 I think maybe you cut and paste from the comment on issue #12142 as I had not seen this PR before . . .

stack72:
@jmahowald sorry! I had 2 tabs open at once - I am not a good multitasker

stack72:
@lwilliams-oats

Please can you reopen this against master? We have just killed off the 0-8-stable branch

Thanks

Paul

PR #12201, opened 2017-02-23
[WIP] Introducing new IBM Cloud provider

ashishth09:
We are adding IBM Cloud provider.
There are two resources being managed by this provider
1. SSH Key 
2. Virtual Guest


builtin/providers/ibmcloud/data_source_ibmcloud_infra_ssh_key.go
```diff
@@ -0,0 +1,110 @@
...
+func dataSourceIBMCloudInfraSSHKeyRead(d *schema.ResourceData, meta interface{}) error {
+	sess := meta.(ClientSession).SoftLayerSession()
+	service := services.GetAccountService(sess)
+
+	label := d.Get("label").(string)
+	mostRecent := d.Get("most_recent").(bool)
+
+	keys, err := service.
+		Filter(filter.Build(filter.Path("sshKeys.label").Eq(label))).
+		Mask("id,label,key,fingerprint,notes,createDate").
```
ashishth09: We actually don't need to provide them, but doing so saves some data transfer over the wire.
It is well supported at the API server itself and hence we are leveraging that. In this case we might not save much but in other cases when payload may be huge it helps us bring only what we need from the server

builtin/providers/ibmcloud/resource_ibmcloud_infra_virtual_guest.go
```diff
@@ -0,0 +1,1061 @@
...
+				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					memoryInMB := float64(v.(int))
+
+					// Validate memory to match gigs format
+					remaining := math.Mod(memoryInMB, 1024)
+					if remaining > 0 {
+						suggested := math.Ceil(memoryInMB/1024) * 1024
+						errors = append(errors, fmt.Errorf(
+							"Invalid 'ram' value %d megabytes, must be a multiple of 1024 (e.g. use %d)", int(memoryInMB), int(suggested)))
```
ashishth09: I think we can do that. So the API call expects the memory to be provided in MB when we create the virtual guest. It is only when we update the virtual guest that value is required in GB and when that moment comes we take benefit of it having already validated in the schema. Just convert the MB/1024

builtin/providers/ibmcloud/resource_ibmcloud_infra_virtual_guest.go
```diff
@@ -0,0 +1,1061 @@
...
+		_, err = virtual.UpgradeVirtualGuest(sess, &result, upgradeOptions)
+		if err != nil {
+			return fmt.Errorf("Couldn't upgrade virtual guest: %s", err)
+		}
+
+		// Wait for softlayer to start upgrading...
+		_, err = WaitForUpgradeTransactionsToAppear(d, meta)
+
+		// Wait for upgrade transactions to finish
+		_, err = WaitForNoActiveTransactions(d, meta)
```
ashishth09: It is is intentional in the sense that when we destroy the virtual guest for example, it will not allow saying that there are some Active transactions.

ashishth09:
Hi @tombuildsstuff ,

Thanks for the review comments. I have addressed them in the new commit.

builtin/providers/ibmcloud/resource_ibmcloud_infra_virtual_guest.go
ashishth09: The API server also responds back with memory in MB when there is read operation, so we simply put back that in the schema, that way it is more consistent.

builtin/providers/ibmcloud/resource_ibmcloud_infra_virtual_guest.go
```diff
@@ -0,0 +1,1061 @@
...
+			"primaryIpAddress,primaryBackendIpAddress,privateNetworkOnlyFlag," +
+			"hourlyBillingFlag,localDiskFlag," +
+			"userData[value],tagReferences[id,tag[name]]," +
+			"datacenter[id,name,longName]," +
+			"primaryNetworkComponent[networkVlan[id]," +
+			"primaryVersion6IpAddressRecord[subnet,guestNetworkComponentBinding[ipAddressId]]," +
+			"primaryIpAddressRecord[subnet,guestNetworkComponentBinding[ipAddressId]]]," +
+			"primaryBackendNetworkComponent[networkVlan[id]," +
+			"primaryIpAddressRecord[subnet,guestNetworkComponentBinding[ipAddressId]]]",
+	).GetObject()
```
ashishth09: Although similar one we saw for the ssh key in the remarks above and the answer holds true here as well. Still worth the explanation as in this case, in addition to bringing less data, we also need to bring all required data with just single API call. We are able to pull in [_Relational_](http://sldn.softlayer.com/reference/datatypes/SoftLayer_Virtual_Guest) properties of Virtual Guest  also in the same API call which will not be possible without masking and getting to multiple API calls. Again this is supported at the SoftLayer API level and hence we are using that. I think [SDK](http://github.com/softlayer/softlayer-go) provides this high level API and as there can be so many combinations of fields that a user may want, leaves it to the client to ask what it needs.

builtin/providers/ibmcloud/provider.go
```diff
@@ -0,0 +1,81 @@
...
+				Description: "The Bluemix Region (for example 'ng').",
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{"BM_REGION", "BLUEMIX_REGION"}, "ng"),
+			},
+			"softlayer_timeout": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "The timeout (in seconds) to set for any SoftLayer API calls made.",
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{"SL_TIMEOUT", "SOFTLAYER_TIMEOUT"}, 60),
+			},
+			"softlayer_account_number": {
```
ashishth09: The current provider will cut across the IBM Cloud portfolio and it just helps to point out that this particular account number is related to SoftLayer.

builtin/providers/ibmcloud/provider.go
```diff
@@ -0,0 +1,81 @@
...
+				Description: "The password for the IBM ID.",
+				DefaultFunc: schema.EnvDefaultFunc("IBMID_PASSWORD", ""),
+			},
+			"region": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The Bluemix Region (for example 'ng').",
+				DefaultFunc: schema.MultiEnvDefaultFunc([]string{"BM_REGION", "BLUEMIX_REGION"}, "ng"),
+			},
+			"softlayer_timeout": {
```
ashishth09: With the prefix we are just making sure that it is for the softlayer API. We will add bluemix API and might have a timeout for that set differently

builtin/providers/ibmcloud/resource_ibmcloud_infra_ssh_key.go
```diff
@@ -0,0 +1,229 @@
...
+			editKey = true
+		}
+
+		if editKey {
+			_, err = service.Id(id).EditObject(&slKey)
+			return err
+		}
+
+		return nil
+	} // End of "Import"
```
ashishth09: Right. The only reason to keep it here is to deal with cases where the same key is attempted to be created twice due to some reason which will not be allowed actually and the API will simply report error that the key already exists. In such case we simply update the notes and label if any without erring.

ashishth09:
Hi @tombuildsstuff 

Thanks for reviewing and providing the comments. I have addressed them in the new commits. 
Regarding the filters/masks we will  look into the possibility of moving these as we go on with subsequent PRs.

jvrplmlmn:
@ashishth09 do you have any update on this? Is this effort still being pursued?

ashishth09:
@jvrplmlmn , we are waiting for someone to look into this.

teamterraform:
Hello, we apologize that this pull request review fell by the wayside. We are actively cleaning up Terraform issues and pull requests, and given Terraform's [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10), we will no longer be accepting pull requests for providers into the core repository. We encourage any potential provider to reach out through the [provider development program](https://www.terraform.io/guides/terraform-provider-development-program.html) to become an official provider. Thank you for your ongoing contributions to our community.

PR #12210, opened 2017-02-23
flatmap: mark computed list as a computed value in Expand

### LINKED ISSUE #12183: Panic on apply after destroy
Hi there,

### Terraform Version

```
Terraform v0.9.0-dev (9ae0e077f1a43956077f8e0ac7f706f1fb5f3edc)
```

### Affected Resource(s)

- core

### Panic Output

https://gist.github.com/sethvargo/9ca97d49f4e754607906bf63db54446b

### Expected Behavior

Terraform should not have crashed.

### Actual Behavior

Terraform crashed.

### Steps to Reproduce

1. Create resources
2. Destroy resources
3. Run apply

### Steps to Fix

1. Delete local statefile

### Important Factoids

- Using Azure, got an error on destroy, but subsequent destroy succeeded.

mitchellh:
Fixes #12183

The fix is in flatmap for this but the entire issue is a bit more
complex. Given a schema with a computed set, if you reference it like
this:

    lookup(attr[0], "field")

And "attr" contains a computed set within it, it would panic even though
"field" is available. There were a couple avenues I could've taken to
fix this:

1.) Any complex value containing any unknown value at any point is
entirely unknown.

2.) Only the specific part of the complex value is unknown.

I took route 2 so that the above works without any computed (since
"name" is not computed but something else is). This may actually have an
effect on other parts of Terraform configs, however those similar
configs would've simply crashed previously so it shouldn't break any
pre-existing configs.

Given the fix, I don't recommend this be backported.

### DIFF
--- a/builtin/providers/test/provider.go
+++ b/builtin/providers/test/provider.go
@@ -8,7 +8,8 @@ import (
 func Provider() terraform.ResourceProvider {
 	return &schema.Provider{
 		ResourcesMap: map[string]*schema.Resource{
-			"test_resource": testResource(),
+			"test_resource":         testResource(),
+			"test_resource_gh12183": testResourceGH12183(),
 		},
 		DataSourcesMap: map[string]*schema.Resource{
 			"test_data_source": testDataSource(),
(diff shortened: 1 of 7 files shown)


mitchellh:
Hah, there were a number of places where this could go wrong :P Wanted to test each

PR #12214, opened 2017-02-23
Hash custom_data in state storage

sethvargo:
This PR changes custom_data to be hashed in the statefile instead of plain-textified. This reduces the size of the statefile and is more in line with how other providers behave.

This also switches to helpers for b64.

/cc @stack72

### DIFF
--- a/builtin/providers/azurerm/resource_arm_virtual_machine.go
+++ b/builtin/providers/azurerm/resource_arm_virtual_machine.go
@@ -2,7 +2,6 @@ package azurerm
 
 import (
 	"bytes"
-	"encoding/base64"
 	"fmt"
 	"log"
 	"net/http"
@@ -296,9 +295,10 @@ func resourceArmVirtualMachine() *schema.Resource {
 						},
 
 						"custom_data": {
-							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Type:      schema.TypeString,
+							Optional:  true,
+							Computed:  true,
+							StateFunc: userDataStateFunc,
 						},
 					},
 				},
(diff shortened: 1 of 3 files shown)


builtin/providers/azurerm/resource_arm_virtual_machine.go
```diff
@@ -1047,12 +1040,7 @@ func expandAzureRmVirtualMachineOsProfile(d *schema.ResourceData) (*compute.OSPr
 	}
 
 	if v := osProfile["custom_data"].(string); v != "" {
-		if isBase64Encoded(v) {
-			log.Printf("[WARN] Future Versions of Terraform will automatically base64encode custom_data")
-		} else {
-			v = base64Encode(v)
-		}
-
+		v = base64Encode(v)
```
sethvargo: Helper does this check now

builtin/providers/azurerm/resource_arm_virtual_machine.go
```diff
@@ -887,14 +887,7 @@ func flattenAzureRmVirtualMachineOsProfile(osProfile *compute.OSProfile) []inter
...
 	if osProfile.CustomData != nil {
-		var data string
-		if isBase64Encoded(*osProfile.CustomData) {
-			decodedData, _ := base64.StdEncoding.DecodeString(*osProfile.CustomData)
-			data = string(decodedData)
-		} else {
-			data = *osProfile.CustomData
-		}
-		result["custom_data"] = data
+		result["custom_data"] = *osProfile.CustomData
```
sethvargo: This I'm not sure about. I tried this with b64 and non-b64 encoded data, and it worked both times. I'm not sure what this code was attempting to do?

stack72:
Looks awesome @sethvargo 

This is going to be a 0.9 only feature just incase it causes any upgrade issues ;)

PR #12219, opened 2017-02-23
fix sorting of module resources during state mv

### LINKED ISSUE #12019: terraform 0.8.7 (and below) `terraform state mv` reorders counted resource if count > 10
Terraform v0.8.7 and below.
Moving a counted resource (resource with multiple instances) with `terraform state mv` reorders the resources at the destination address if the resource count is greater than 10.
Attached ZIP file containing `terraform.tfstate`: [tfstate.zip](https://github.com/hashicorp/terraform/files/781161/tfstate.zip)
Run:
```
terraform state mv module.peerings.aws_vpc_peering_connection.src module.peerings.aws_vpc_peering_connection.dest
```
Note that **aws_vpc_peering_connection.src.1** != **aws_vpc_peering_connection.dest.1**.
Attached also updated test code `state_mv_test.go`: [state_mv_test.go.txt](https://github.com/hashicorp/terraform/files/781170/state_mv_test.go.txt) that shows the problem:
```
go test -v -run="TestStateMv_stateOutNew_largeCount"
```

jbardin:
Module resource were being sorted lexically by name by the state filter.
If there are 10 or more resources, the order won't match the index
order, and resources will have different indexes in their new location.

Sort the FilterResults by index numerically when the names match.

Clean up the module String output for visual inspection by sorting
Resource name parts numerically when they are an integer value.

Fixes #12019

### DIFF
--- a/command/state_mv_test.go
+++ b/command/state_mv_test.go
@@ -370,6 +370,184 @@ func TestStateMv_stateOutNew_count(t *testing.T) {
 	testStateOutput(t, backups[0], testStateMvCount_stateOutOriginal)
 }
 
+// Modules with more than 10 resources were sorted lexically, causing the
+// indexes in the new location to change.
+func TestStateMv_stateOutNew_largeCount(t *testing.T) {
+	state := &terraform.State{
+		Modules: []*terraform.ModuleState{
+			&terraform.ModuleState{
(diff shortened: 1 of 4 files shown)


jbardin:
needed a minor fix of the output test in another test.

Issue #10984, opened 2017-01-01
Google Image Family doesn't detect when "latest" image changes.


ldelossa:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

terraform --version
Terraform v0.7.13
(Not using latest to avoid create_timeout bug)

### Affected Resource(s)
Please list the resources as a list, for example:
- google_compute_instance_template
- google_compute_instance_group_manager

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "google_compute_instance_template" "sentry_web" {
  depends_on = ["google_compute_instance.sentry-redis-01",
                "google_compute_instance.sentry-redis-02",
                "google_compute_instance.sentry-postgresql-01",
                "consul_keys.sentry_web"]

  name_prefix = "player-sentry-${var.cluster_name}-web-${var.image_version}-"
  description = "Sentry Web Template for env: ${var.cluster_name} version: ${var.image_version}"
  region = "us-east1"

  machine_type = "n1-standard-2"
  disk {
    source_image = "sentry-web-${var.cluster_name}"
    auto_delete = true
    boot = true
  }

  tags = ["sentry", "sentry-web"]

  network_interface {
    subnetwork = "us-east1-b"
    access_config = {
      // Empty to default to ephemeral public ip
    }
  }

  lifecycle {
    create_before_destroy = true
  }
}

resource "google_compute_instance_template" "sentry_web_test" {
  depends_on = ["google_compute_instance.sentry-redis-01",
                "google_compute_instance.sentry-redis-02",
                "google_compute_instance.sentry-postgresql-01",
                "consul_keys.sentry_web"]

  name_prefix = "player-sentry-${var.cluster_name}-web-${var.image_version}-test"
  description = "Sentry Web Template for env: ${var.cluster_name} version: ${var.image_version}"
  region = "us-east1"

  machine_type = "n1-standard-2"
  disk {
    source_image = "sentry-web-${var.cluster_name}"
    auto_delete = true
    boot = true
  }

  tags = ["sentry", "sentry-web"]

  network_interface {
    subnetwork = "us-east1-b"
    access_config = {
[...]

PR #12223, opened 2017-02-24
provider/google: upgrade our image resolution logic

paddyforan:
Prior to now, GCP images could be specified as:

* The URL of the image, which would be passed straight through to the API
* `{project}/{name}`, which would be resolved as an image in the named project, and if found, passed as a URL to the API. If not found, it would be interpreted as an image family in the named project, and if found, resolved to the latest image in that family, which would be passed through to the API as a URL.
* `{name}`, which would be resolved as an image in the authenticated project, and if found, passed as a URL to the API. If not found, it would be interpreted as an image family in the authenticated project, and if found, resolved to the latest image in that family, which would be passed through to the API as a URL. If still not found, it would be checked against a static map of GCP-provided projects, and if it had a match, would be interpreted as an image in that project. If _that_ image was found, it would be passed through to the API as a URL. Finally, if it _still_ wasn't found, it would be interpreted as a family in the GCP-provided project, and if it was found, if would be resolved to the latest image in that family, which would be passed through to the API as a URL.

This code was used uniformly across all our disk input modes (as a sub-resource of instances, as standalone disks, as instance templates), but it had some shortcomings:

* Not all API inputs were valid inputs for Terraform, which feels unintuitive, especially as there's no real reason they couldn't be supported. [Valid API inputs](https://cloud.google.com/compute/docs/reference/latest/disks#sourceImage):
  * `projects/{project}/global/images/family/{family}`
  * `projects/{project}/global/images/{image}`
  * `global/images/{image}`
  * `global/images/family/{family}`
* Families would get resolved to images, meaning things like instance templates would get frozen to a specific image, even though the user asked for them to point to the most recent image for the family at disk spinup time.
* The freezing of images happened at disk spinup time, meaning Terraform had no idea whether a newer image was available, and couldn't prompt the user to apply the change.
* In the case of an image and a family conflicting in names, there was no way to tell Terraform which to use. While this shouldn't happen, it'd be nice to have the ability to workaround it.

This PR updates the resolveImage helper in two ways:
* First, it hews closer to the API input. URLs are still passed as URLs, but everything else gets converted to the API equivalent that most closely matches the user intent. For example, `{project}/family` specifies the project, but not an image, so it gets converted to `projects/{project}/global/images/family/{family}`. `family/{family}` does not specify a project or an image, so it gets converted to `global/images/family/{family}`, rather than interpolating the authenticating project in.
* The following inputs are now accepted:
  * `family/{family}`
  * `projects/{project}/global/images/{image}`
  * `{family}`
  * `{image}`
  * `{URL to image}`
  * `global/images/{image}`
  * `global/images/family/{family}`
  * `{project}/{image}`
  * `{project}/{family}`

I also added an acceptance test that checks that each of those inputs gets converted to the API values we expect, including for images and families provided by GCP.

This change was prompted by #10984 and is half the solution to it. The other half is to provide a data source which will return the latest image in a family, which will allow users to pin disks to the latest image at `terraform apply`-time, and use `terraform apply` to roll out updates to the disk image.


builtin/providers/google/image.go
```diff
@@ -2,96 +2,186 @@ package google
...
+	for k, v := range imageMap {
+		if strings.Contains(name, k) {
+			builtInProject = v
+			break
+		}
+	}
+	switch {
+	case resolveImageLink.MatchString(name): // https://www.googleapis.com/compute/v1/projects/xyz/global/images/xyz
+		return name, nil
+	case resolveImageProjectImage.MatchString(name): // projects/xyz/global/images/xyz
```
danawillow: This and the following three checks all follow the same format. It might be a bit cleaner to loop over a slice of structs containing the regex, the expected length, and the result URL (although if we can't do a similar thing elsewhere it might look out of place- use your judgment)

builtin/providers/google/image.go
```diff
@@ -2,96 +2,186 @@ package google
...
 )
 
-// If the given name is a URL, return it.
-// If it is of the form project/name, search the specified project first, then
-// search image families in the specified project.
-// If it is of the form name then look in the configured project, then hosted
-// image projects, and lastly at image families in hosted image projects.
-func resolveImage(c *Config, name string) (string, error) {
+const (
+	resolveImageProjectRegex = "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?" // TODO(paddy): this isn't based on any documentation; we're just copying the image name restrictions. Need to follow up with @danawillow and/or @evandbrown and see if there's an actual limit to this
```
paddyforan: So my thought on this is that I'd like to try to keep my future options open and not trap myself into a forced backwards incompatibility. In this case, that means accepting the smallest set of strings I can while still accepting every valid string.

On the other hand, however, the typical Terraform approach is to avoid client-side validation, as it is a headache keeping it in sync with the server-side validation.

So I could see both sides on this one. Do you have a preference either way?

builtin/providers/google/image.go
```diff
@@ -2,96 +2,186 @@ package google
...
+	resolveImageFamilyRegex  = "[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?" // TODO(paddy): this isn't based on any documentation; we're just copying the image name restrictions. Need to follow up with @danawillow and/or @evandbrown and see if there's an actual limit to this
```
paddyforan: Yeah, I saw that. I just didn't know which specific part of RFC 1035 it meant. Does it mean it must be a valid domain? A valid subdomain?

builtin/providers/google/image.go
paddyforan: I thought of doing that, but wasn't super convinced of the benefit--the indirection didn't make it feel any clearer, and because the format of the code would be essentially the same (define a case, find the matches, check the lengths, handle errors, build a string) it seemed to me like it would be adding more code, not removing code.

I'm happy to reconsider on this one, though, if you have any suggestions or pointers.

builtin/providers/google/image.go
```diff
@@ -2,96 +2,186 @@ package google
...
+			builtInProject = v
+			break
+		}
+	}
+	switch {
+	case resolveImageLink.MatchString(name): // https://www.googleapis.com/compute/v1/projects/xyz/global/images/xyz
+		return name, nil
+	case resolveImageProjectImage.MatchString(name): // projects/xyz/global/images/xyz
+		res := resolveImageProjectImage.FindStringSubmatch(name)
+		if len(res)-1 != 2 { // subtract one, index zero is the entire matched expression
```
paddyforan: I'm mostly just sanity-checking to make sure that if the regex string changes and the logic handling it isn't updated, or some other type of human error, it's an error, not a crash. I'm open to removing the code, if you think that's overdoing it. :) In terms of a helper function, you mean just something like replacing

```go
if len(res)-1 != 2 {
    return "", fmt.Errorf("Expected %d project image regex matches, got %d for %s", 2, len(res)-1, name)
}
```

with

```go
if err := checkRegexMatches(res, 2); err != nil {
    return "", err
}
```

?

I could see a case for that, and how it would clean up some code.

builtin/providers/google/image.go
danawillow: Not really, though I don't 100% agree with your logic. If we let the server handle it, then we also don't have to worry about backwards incompatibility (that's Google's problem). Sure, we could try to predict it by being extra strict now just in case, but we can't really tell in what way the Google API would become stricter. The win by doing client-side validation is more so that problems show up faster to the user, which I think is a valid win. If you do keep it this way though, let's get that regex changed to the actual requirements, since the ones for projects are stricter than for images.

builtin/providers/google/image.go
danawillow: Just took a closer look and it's the same validation as images, so this looks like the right regex to use.

builtin/providers/google/image.go
danawillow: I took a look to see if I could quickly do my own suggestion and it looks like you're right that it would add rather than remove code.

builtin/providers/google/image.go
danawillow: I don't think it's necessary, but I'm not opposed to having it there either. I do like your helper function though- I think one of the reasons this part of the code feels messy to me is because all of the repeated `-1`s and their associated `// subtract one, index zero is the entire matched expression` comments, so this would help add more whitespace into that section.

builtin/providers/google/image.go
paddyforan: > If we let the server handle it, then we also don't have to worry about backwards incompatibility (that's Google's problem).

Where this comes into play is that we have several custom shorthands for this. The one I'm thinking of specifically is `{project}/{image}` and `{project}/{family}`; if, in the future, we wanted to do something like `{family}/{image}` for whatever reason, we'd find ourselves in a tough spot, as we wouldn't be able to tell between `{project}/{image}` and `{family}/{image}`. But the more I'm thinking about this, the more I'm agreeing with you; we try to avoid doing too much client-side validation because it can be brittle, and in this case, guarding against the vagaries of the future doesn't feel like it's yielding enough concrete benefits.

My proposal: Let's accept anything that matches `/[a-zA-Z0-9_-]*/` for projects, images, and families. Sound like a good way forward?

paddyforan:
Latest push should address all feedback. Unit test still runs:

```
=== RUN   TestAccComputeImage_resolveImage
--- PASS: TestAccComputeImage_resolveImage (93.62s)
PASS
```

builtin/providers/google/image.go
```diff
@@ -47,6 +47,13 @@ func resolveImageFamilyExists(c *Config, project, name string) (bool, error) {
 	}
 }
 
+func sanityTestRegexMatches(expected int, got []string, regexType, name string) error {
+	if len(got)-1 != expected { // subtract one, index zero is the entire matched expression
+		return fmt.Errorf("Expected %d %s regex matches, got %d for %s", 2, regexType, len(got)-1, name)
```
danawillow: Shouldn't this `2` be `expected`?

paddyforan: Should be addressed now.

paddyforan:
Updated docs to document the new acceptable inputs and updated the copy/paste error you noticed. Still look good? :)

website/source/docs/providers/google/r/compute_disk.html.markdown
```diff
@@ -37,9 +37,11 @@ The following arguments are supported:
     encoded in [RFC 4648 base64](https://tools.ietf.org/html/rfc4648#section-4)
     to encrypt this disk.
 
-* `image` - (Optional) The image from which to initialize this disk. Either the
-    full URL, a contraction of the form "project/name", or just a name (in which
-    case the current project is used).
+* `image` - (Optional) The image from which to initialize this disk. This can be
```
danawillow: Can each of the image choices be code font?

website/source/docs/providers/google/r/compute_instance.html.markdown
```diff
@@ -117,12 +117,11 @@ the type is "local-ssd", in which case scratch must be true).
 * `disk` - The name of the existing disk (such as those managed by
     `google_compute_disk`) to attach.
 
-* `image` - The image from which to initialize this
-    disk. Either the full URL, a contraction of the form "project/name", the
-    name of a Google-supported
-    [image family](https://cloud.google.com/compute/docs/images#image_families),
-    or simple the name of an image or image family (in which case the current
-    project is used).
+* `image` - The image from which to initialize this disk. This can be
```
danawillow: (here too)

website/source/docs/providers/google/r/compute_instance_template.html.markdown
```diff
@@ -176,8 +176,12 @@ The `disk` block supports:
 * `disk_name` - (Optional) Name of the disk. When not provided, this defaults
     to the name of the instance.
 
-* `source_image` - (Required if source not set) The name of the image to base
-    this disk off of. Accepts same arguments as a [google_compute_instance image](https://www.terraform.io/docs/providers/google/r/compute_instance.html#image).
+* `source_image` - (Required if source not set) The image from which to
+    initialize this disk. This can be one of: the image self_link,
```
danawillow: (here too)

danawillow:
Still looks good! Feel free to merge when ready (after the code font changes if you agree with that comment)

paddyforan:
Updated! Once Travis goes green, I'll merge.

PR #12228, opened 2017-02-24
provider/google: add migration notes for projects

paddyforan:
Update the docs for `google_project` with @zachgersh's suggestions
(#11895) to properly communicate the changes that took place in 0.8.5.
While they don't break any current configs or state, the new behaviour
should be called out for people who were using the old behaviour and are
adding new projects to their configs/state.

I also took this opportunity to update google_project_iam_policy with a
note to users letting them know that there be dragons.

Prompted by #11895.


danawillow:
LGTM, merge whenever you feel like it.

PR #12236, opened 2017-02-24
Vendor fixes

jbardin:
Most of the updates here are matching the commits of some sub-packages with those of their parent. 

Update the `golang.org/x/crypto/curve25519` package, and manually add the generic source that govendor won't add.

Manually fix the hashes for the `golang.org/x/crypto/curve25519` and `github.com/scaleway/scaleway-cli/pkg/api` packages so `govendor status` passes. 

Make the `govendor status` check mandatory in travis.


mitchellh:
LGTM if everything works :P

PR #12239, opened 2017-02-24
provider/openstack: Rename provider to loadbalancer_provider

### LINKED ISSUE #12237: openstack_lb_loadbalancer_v2: "provider" attribute fails validation because of namespace collision with "Provider"
```
resource "openstack_lb_loadbalancer_v2" "hero_edge" {
  region        = "${var.region}"
  vip_subnet_id = "${var.vip_subnet_id}"
  provider      = "a10networks_ext"
}
```

Upon attempting to plan:
```
» terraform plan xx-98
2 error(s) occurred:

* module web: provider alias must be defined by the module or a parent: a10networks_ext
```

Obviously, this isn't referring to the `provider` attribute listed here:
https://www.terraform.io/docs/providers/openstack/r/lb_loadbalancer_v2.html

And instead is referring to the Provider during validation:
https://github.com/hashicorp/terraform/blob/f2c16b7ba2b01b233623097e479825efe1ead3c8/config/module/validate_provider_alias.go#L69-L71

### Terraform Version
Terraform v0.8.7

### Affected Resource
- openstack_lb_loadbalancer_v2

```
resource "openstack_lb_loadbalancer_v2" "hero_edge" {
  region        = "${var.region}"
  vip_subnet_id = "${var.vip_subnet_id}"
  provider      = "a10networks_ext"
}
```

### Debug Output
https://gist.github.com/rcreasey/f79c160cd14a3c1f5110567efb4070d1

### Expected Behavior
This "provider" attribute should be applied to the resource and not validated as a "Provider" alias.

### Actual Behavior
What actually happened?

### Steps to Reproduce
1. Define a `openstack_lb_loadbalancer_v2` resource and specify the `provider` attribute.
2. Attempt to run `terraform plan`

### Important Factoids
This seems to be a failure in the validation code of Terraform, and not related to any specific Openstack API.

jtopjian:
This commit renames provider to loadbalancer_provider in the
openstack_lb_loadbalancer_v2 resource.

It also changes security_group_ids to Computed so default
security groups are added to the state correctly.

Fixes #12237

### DIFF
--- a/builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
+++ b/builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
@@ -75,7 +75,7 @@ func resourceLoadBalancerV2() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"provider": &schema.Schema{
+			"loadbalancer_provider": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
@@ -85,6 +85,7 @@ func resourceLoadBalancerV2() *schema.Resource {
 			"security_group_ids": &schema.Schema{
 				Type:     schema.TypeSet,
 				Optional: true,
+				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
 			},
@@ -108,7 +109,7 @@ func resourceLoadBalancerV2Create(d *schema.ResourceData, meta interface{}) erro
 		VipAddress:   d.Get("vip_address").(string),
 		AdminStateUp: &adminStateUp,
 		Flavor:       d.Get("flavor").(string),
-		Provider:     d.Get("provider").(string),
+		Provider:     d.Get("loadbalancer_provider").(string),
 	}
 
 	log.Printf("[DEBUG] Create Options: %#v", createOpts)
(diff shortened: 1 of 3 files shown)


builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
```diff
@@ -75,7 +75,7 @@ func resourceLoadBalancerV2() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"provider": &schema.Schema{
+			"loadbalancer_provider": &schema.Schema{
```
jtopjian: @stack72 I can definitely do a deprecation thing here.

Just curious, though: would `provider` have worked during any previous release? Was it only recent that the error reported in #12237 would happen? If so, that would definitely help my sanity :)

builtin/providers/openstack/resource_openstack_lb_loadbalancer_v2.go
```diff
@@ -100,6 +108,11 @@ func resourceLoadBalancerV2Create(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
 	}
 
+	var lbProvider string
+	if v, ok := d.GetOk("loadbalancer_provider"); ok {
+		lbProvider = v.(string)
+	}
+
```
jtopjian: @stack72 I couldn't decide on leaving this part as-is or adding the following:

```go
if v, ok := d.GetOk("provider"); ok && lbProvider != "" {
        lbProvider = v.(string)
}
```

The above would set `lbProvider` if `provider` was set and the new preferred `loadbalancer_provider` wasn't set. However, since having an argument of `provider will error out anyway, is it worth adding the above block of code in?

rcreasey:
FYI, I've validated this actually correctly sets the `openstack_lb_loadbalancer_v2` provider attribute against an actual OpenStack Neutron API.

Issue #12137, opened 2017-02-21
AWS `s3_object` resource now requires `s3:GetObjectTagging` permission


ljfranklin:
Starting with Terraform version 0.8.7, the `s3_object` resource now requires the `s3:GetObjectTagging` permission even if no tags are specified in your TF configuration files. Probably introduced in [this PR](https://github.com/hashicorp/terraform/pull/11344). The error message (shown below) is unfortunately not helpful as it doesn't mention what new permission you need.

Could the new s3 tagging code be updated to swallow 403 errors when trying to retrieve the tags for an object, or update docs somewhere to list this as a required permission?

Thanks!

### Terraform Version
0.8.7

### Affected Resource(s)
- s3_object

### Error

```
aws_s3_bucket_object.s3_object: AccessDenied: Access Denied
  	status code: 403, request id: 599D98A31E927E6E
```

ewbankkit:
Swallowing the error seems a bit "dangerous" to me - If a `403` was swallowed and assumed no tags then for example `terraform plan` could show changes when there shouldn't be.
I agree that changes to IAM permissions should be recorded and communicated somehow.
I'll have the same problem with https://github.com/hashicorp/terraform/pull/12243.

radeksimko:
Hi @ljfranklin 
I understand this has caused inconvenience to mindful users like you that are very strict about crafting IAM policies - sorry for that. Can you explain further why is it dangerous/risky or not possible in your environment to add **read** permission for reading tags?

I would agree that requiring new write or even delete permissions or resource-agnostic permissions in general (not scoped to resources, like S3 bucket name) would be undesirable, but I'm struggling to understand the problem or context here specifically.

I agree with @ewbankkit that ignoring 403 error would cause confusions to users that do want to use tags and just might forgot to setup the right policies.

> The error message (shown below) is unfortunately not helpful as it doesn't mention what new permission you need.

I agree, unfortunately that's all we get from the AWS API, many errors are as brief as this one. That said I opened https://github.com/hashicorp/terraform/pull/12759 - feedback welcomed there.

ljfranklin:
> Can you explain further why is it dangerous/risky or not possible in your environment to add read permission for reading tags?

No problem with adding it, the main issue was it took a bit of time to figure out exactly which new permission was required to avoid the error. But your PR seems to have given a much nicer error now, so I think we're good to close this out.

Thanks!

PR #12243, opened 2017-02-24
provider/aws: Add KMS key tag support

ewbankkit:
Add support for AWS KMS [key tagging](http://docs.aws.amazon.com/kms/latest/developerguide/tagging-keys.html).
Acceptance test:
```
make testacc TEST=./builtin/providers/aws/ TESTARGS='-run=TestAccAWSKmsKey_tags'
```
Additional KMS API actions:
* `ListResourceTags`
* `TagResource`
* `UntagResource`

### DIFF
--- a/builtin/providers/aws/resource_aws_kms_key.go
+++ b/builtin/providers/aws/resource_aws_kms_key.go
@@ -80,6 +80,7 @@ func resourceAwsKmsKey() *schema.Resource {
 					return
 				},
 			},
+			"tags": tagsSchema(),
 		},
 	}
 }
@@ -97,6 +98,9 @@ func resourceAwsKmsKeyCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 	if v, exists := d.GetOk("policy"); exists {
 		req.Policy = aws.String(v.(string))
+	}
+	if v, exists := d.GetOk("tags"); exists {
+		req.Tags = tagsFromMapKMS(v.(map[string]interface{}))
 	}
 
 	var resp *kms.CreateKeyOutput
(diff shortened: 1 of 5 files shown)


ewbankkit:
Added acceptance test and rebased.

stack72:
Hi @ewbankkit 

What do you suggest we do about the permissions issue?

Paul

ewbankkit:
@stack72 I would say just note in the CHANGELOG. The documentation doesn't currently state which IAM permissions are required for which resources and I'm sure this situation (additional attributes imply additional permissions) happens relatively frequently.
Maybe once you get to `1.0` we'll have to worry more about such changes.

builtin/providers/aws/resource_aws_kms_key.go
```diff
@@ -170,6 +174,14 @@ func resourceAwsKmsKeyRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.Set("enable_key_rotation", krs.KeyRotationEnabled)
 
+	tagList, err := conn.ListResourceTags(&kms.ListResourceTagsInput{
+		KeyId: metadata.KeyId,
+	})
+	if err != nil {
+		return nil
```
radeksimko: As mentioned in https://github.com/hashicorp/terraform/issues/12137#issuecomment-287066725 we should not be ignoring the error here. Instead to avoid confusions you can do something similar to what I did in https://github.com/hashicorp/terraform/pull/12759 - i.e. make the error message descriptive enough.

ewbankkit: Agreed. I'll try and fix that today.

ewbankkit:
@radeksimko Made change as suggested (and rebased). I think the initial `return nil` was a typo for `return err` but still agree on better error message.

PR #12245, opened 2017-02-24
Always quote identifiers when encoding HCL

### LINKED ISSUE #11977: terraform `push` fails where `plan` doesn't
Hi there,

I'm running into an issue when trying to use the following variables :

```
variable "foo" {
  type = "map"
  default = {
    "0.0.0.0/0" = "i-123456"
  }
}
```

If I do a `terraform plan`, no issue. But I have issues when I want to push :

```
An error has occurred while serializing the variables for uploading:
At 2:1: expected: IDENT | STRING | ASSIGN | LBRACE got: FLOAT
```

### Terraform Version
v0.8.6

### Steps to Reproduce

The issue is easy to reproduce by changing `command/test-fixtures/push-tfvars/main.tf` and running `make test TEST=./command`

I think this is a HCL parsing problem but I was not sure where to open the issue.

Let me know if you need more details.

jbardin:
HCL identifiers may need to be quoted, so always quote them to be safe.

fixes #11977

### DIFF
--- a/command/hcl_printer.go
+++ b/command/hcl_printer.go
@@ -94,7 +94,7 @@ func (e *encodeState) encodeMap(m map[string]interface{}) error {
 	for i, k := range sortedKeys(m) {
 		v := m[k]
 
-		e.WriteString(k + " = ")
+		e.WriteString(fmt.Sprintf("%q = ", k))
 		err := e.encode(v)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


jbardin:
ha, the failure is fixed in another PR ;) re-running tests until they pass for now

Issue #9121, opened 2016-09-29
google_compute_target_pool is not idempotent when using URLs as values from google_compute_instance


jsierles:
Hi there,

Terraform v0.7.4
google_compute_target_pool

``` hcl
resource "google_compute_target_pool" "loadbalancer" {
  name = "target-pool-loadbalancer-${var.env}"
  instances = ["${google_compute_instance.loadbalancer.*.self_link}"]
  health_checks = ["${google_compute_http_health_check.loadbalancer.name}"]
}
```
### Plan output

```
~ google_compute_target_pool.loadbalancer
    instances.0: "europe-west1-d/loadbalancer-0-staging" => "https://www.googleapis.com/compute/v1/projects/myproject-com/zones/europe-west1-d/instances/loadbalancer-0-staging"
    instances.1: "europe-west1-d/loadbalancer-1-staging" => "https://www.googleapis.com/compute/v1/projects/myproject-com/zones/europe-west1-d/instances/loadbalancer-1-staging"
```
### Expected Behavior

This value should not be changed on every plan/apply run.
### Actual Behavior

On every `plan` and `apply`, this resource is modified without makes changes to the configuration.

jsierles:
If this is a bug, I'd be happy to look at applying a fix if someone could clarify what the bug might be.

danawillow:
Sorry for the long delay! We're trying to get through some of our backlog of issues now, and I found this one. I'm not sure if you're still having the issue, but here's why it's happening.

The code is written so that an instance doesn't need to exist before it's added to the target pool. So instead of specifying the full URL, the author of the config can just specify the zone/name of the instance. After the resource is created, the state file stores that same zone/name key so that terraform doesn't think there's a diff. Unfortunately, this ends up meaning that if you specify the full URL, terraform will think there's a diff.

A solution would be to do a d.Get("instances") in the read method to check if the currently set values are URLs, and if so, don't do the conversion away from them. I'm not going to work on this now, but if anyone wants to take it I'm pretty sure that would work.

An easy workaround is to just convert to the zone/name syntax by doing something like:
`instances = ["${formatlist("%s/%s", google_compute_instance.loadbalancer.*.zone, google_compute_instance.loadbalancer.*.name)}"]`

Hope that helps!

jsierles:
Thanks, will try that.

queeno:
I've just been hit by this bug today, so it is still a problem in the latest terraform `v0.8.4`. it would be nice to have it fixed.

[...]

Issue #9051, opened 2016-09-26
google_compute_forwarding_rule isn't idempotent anymore (issue with a single port in port_range).


sbadia:
### Terraform Version
- `v0.7.4` (I don't remember but this bug occur since 0.7.x)
### Affected Resource(s)
- google_compute_forwarding_rule
### Output

The resource google_compute_forwarding_rule isn't idempotent anymore across run.

```
-/+ google_compute_forwarding_rule.ipsec-foobar1-gw1-rule-udp4500
    ip_address:  "1.2.3.4" => "1.2.3.4"
    ip_protocol: "UDP" => "UDP"
    name:        "ipsec-foobar1-gw1-rule-udp4500" => "ipsec-foobar1-gw1-rule-udp4500"
    port_range:  "4500-4500" => "4500" (forces new resource)
    project:     "foobar-cloud" => "<computed>"
    region:      "europe-west1" => "<computed>"
    self_link:   "https://www.googleapis.com/compute/v1/projects/foobar-cloud/regions/europe-west1/forwardingRules/ipsec-foobar1-gw1-rule-udp4500" => "<computed>"
    target:      "https://www.googleapis.com/compute/v1/projects/foobar-cloud/regions/europe-west1/targetVpnGateways/ipsec-foobar1-gw1" => "https://www.googleapis.com/compute/v1/projects/foobar-cloud/regions/europe-west1/targetVpnGateways/ipsec-foobar1-gw1"

-/+ google_compute_forwarding_rule.ipsec-foobar1-gw1-rule-udp500
    ip_address:  "1.2.3.4" => "1.2.3.4"
    ip_protocol: "UDP" => "UDP"
    name:        "ipsec-foobar1-gw1-rule-udp500" => "ipsec-foobar1-gw1-rule-udp500"
    port_range:  "500-500" => "500" (forces new resource)
    project:     "foobar-cloud" => "<computed>"
    region:      "europe-west1" => "<computed>"
    self_link:   "https://www.googleapis.com/compute/v1/projects/foobar-cloud/regions/europe-west1/forwardingRules/ipsec-foobar1-gw1-rule-udp500" => "<computed>"
    target:      "https://www.googleapis.com/compute/v1/projects/foobar-cloud/regions/europe-west1/targetVpnGateways/ipsec-foobar1-gw1" => "https://www.googleapis.com/compute/v1/projects/foobar-cloud/regions/europe-west1/targetVpnGateways/ipsec-foobar1-gw1"
```
[...]

PR #12258, opened 2017-02-26
Update google-two-tier example to be idempotent

gobengo:
The google-two-tier example in master adds or changes 2 things on each re-apply. This PR makes it idempotent.

* `google_compute_forwarding_rule.default.port_range = "80"` would save okay, but then would save to state file as "80-80", so it would always think a change was needed
* `google_compute_target_pool.default.instances` items would resolve to full URLs, but saved in state as just `{zone}/{instance.name}`.
  * This PR uses a [workaround](https://github.com/hashicorp/terraform/issues/9121#issuecomment-274910248) mentioned in #9121

### DIFF
--- a/examples/google-two-tier/main.tf
+++ b/examples/google-two-tier/main.tf
@@ -17,14 +17,14 @@ resource "google_compute_http_health_check" "default" {
 
 resource "google_compute_target_pool" "default" {
   name          = "tf-www-target-pool"
-  instances     = ["${google_compute_instance.www.*.self_link}"]
+  instances     = ["${formatlist("%s/%s", var.region_zone, google_compute_instance.www.*.name)}"]
   health_checks = ["${google_compute_http_health_check.default.name}"]
 }
 
 resource "google_compute_forwarding_rule" "default" {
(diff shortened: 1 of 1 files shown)


paddyforan:
Thanks for opening this @gobengo! I'm tempted to just get #9121 and #9051 fixed, instead of updating the examples to work around the bugs. What do you think?

gobengo:
I'm sort of new to terraform, but the workarounds don't seem too hacky.

Fixing bugs can take time or be deprioritized for plenty of good reasons.
Might as well have idempotent examples until then. PRs addressing those
issues can always change the examples back to the best right way.

apparentlymart:
Hello @gobengo, and thanks for working on this!

If you or someone else has the time and motivation to apply the same changes to [the Google Cloud provider repository](https://github.com/terraform-providers/terraform-provider-google) and open a new PR there, the maintainers of that provider should be able to review and merge it. On the other hand, it looks like at least one of the issues that @paddycarver mentioned has been closed in the mean time, so I'm not sure if this change is still required.

gobengo:
@apparentlymart makes sense. thanks!

PR #12265, opened 2017-02-26
provider/openstack: Update Development Documentation

jtopjian:
This commit removes the bundled devstack script and updates the
documentation to point to the latest build scripts used for
testing. It also mentions that development should be possible on
any OpenStack environment.


stack72:
LGTM thanks :)

Issue #12257, opened 2017-02-26
openstack_lb_member_v1 additional status for F5 provider


beeva-germanmoya:
### Terraform Version
Terraform v0.8.7 

### Affected Resource(s)
- [openstack_lb_member_v1](https://www.terraform.io/docs/providers/openstack/r/lb_member_v1.html)

### Expected Behavior
Accept `CREATED` or `DOWN` as a valid and correct status for the resource.

### Actual Behavior
`apply` command fail because doesn't recognize F5 additional states for the lb_members

### Steps to Reproduce
Use an openstack_lb_member_v1 in a openstack_lb_pool_v1 with the F5 provider

### Important Factoids
This case apply for OpenStack LBaaSv1 using the F5 provider

### References
- [F5Networks/f5-openstack-lbaasv1](https://github.com/F5Networks/f5-openstack-lbaasv1/blob/master/agent/f5/oslbaasv1agent/drivers/bigip/icontrol_driver.py#L67)

jtopjian:
@beeva-germanmoya Thanks for pointing this out. This is a pretty easy fix. I'm just marking it as an enhancement since it's related to getting a vendor plugin working.

jtopjian:
@beeva-germanmoya Check out #12266. If you're able test it out by compiling from source, let me know if you run into any problems with it.

stack72:
Closed via #12266

PR #12266, opened 2017-02-26
provider/openstack: Add Additional Targets for LBaaS v1 Member

jtopjian:
This commit adds CREATED and DOWN as additional valid targets
for creating LBaaS v1 members.

For #12257

The typo in the test file must have been some `vi` key mash that happened post-test / pre-commit. I know that test passed last time I ran it.


stack72:
LGTM thanks :)

Issue #12255, opened 2017-02-26
DigitalOcean infinite apply loop (forces new resource)


MartinsThiago:
Hi there,

### Terraform Version
v0.8.7

### Affected Resource(s)
- digitalocean_droplet

### Terraform Configuration Files
```hcl
resource digitalocean_droplet stage_server {
  image = "22995881" //ubuntu 16.04 with docker 1.13.1
  name = "test"
  region = "nyc1"
  size = "512mb"
}
```

### Expected Behavior
After a successful `terraform apply`, running `terraform plan` should show
```
Plan: 0 to add, 0 to change, 0 to destroy.
```

### Actual Behavior
It shows
```
Plan: 1 to add, 0 to change, 1 to destroy.
```

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

### Possible solution
I think I found where it can be fixed ([See code](https://github.com/hashicorp/terraform/blob/master/builtin/providers/digitalocean/resource_digitalocean_droplet.go#L253-L257))
What happens is that digitalocean_droplet.**image** accepts an **id** or **slug** ([See in docs](https://www.terraform.io/docs/providers/do/r/droplet.html#image))... the specific image I am using offers both options

As the code a mentioned above prefers **slug** over **id**, terraform saves the state container image "ubuntu-16.04"... when running `terraform plan` it displays

![image](https://cloud.githubusercontent.com/assets/8419829/23336300/59291d18-fbaa-11e6-8263-f5d2a3d57388.png)

Despite of just have created the resource, terraform thinks it needs to be updated

**PS: Simply inverting the if code, to prefer ids over slugs won't solve the problem**

snowsky:
I reproduced this issue. I think you are right for the solution. +1

snowsky:
I think this is by design. It will always force to recreate this resource, [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/digitalocean/resource_digitalocean_droplet.go#L29).

However, we can still use computed (true) instead of ForceNew(true) to let terraform to decide if update when apply.

roidelapluie:
I have proposed a fix in #13879

PR #12269, opened 2017-02-27
DigitalOcean infinite apply loop (forces new resource)

snowsky:


snowsky:
@stack72 it failed at some point which is not related to digitalocean. How could I rebuild the PR on TravisCI?

snowsky:
#12255

PR #12273, opened 2017-02-27
Fix emr_cluster docs

mbode:
List of applications was broken. Also added Apache Flink as valid application.


stack72:
LGTM Thanks for this :)

PR #12281, opened 2017-02-27
provider/google: add local ssd count support for container clusters

mrooding:
Support the `NodeConfig.localSsdConfig` property for `container_cluster` resources.

I have:
- Updated the `container/v1` and `gensupport` vendor dependencies
- Updated the website documentation
- Ran tests
- Formatted the code

### DIFF
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -231,6 +231,22 @@ func resourceContainerCluster() *schema.Resource {
 							},
 						},
 
+						"local_ssd_count": &schema.Schema{
+							Type:     schema.TypeInt,
+							Optional: true,
+							Computed: true,
+							ForceNew: true,
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+								value := v.(int)
+
+								if value < 0 {
+									errors = append(errors, fmt.Errorf(
+										"%q cannot be negative", k))
+								}
+								return
+							},
+						},
+
 						"oauth_scopes": &schema.Schema{
 							Type:     schema.TypeList,
 							Optional: true,
@@ -388,6 +404,10 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 
 		if v, ok = nodeConfig["disk_size_gb"]; ok {
 			cluster.NodeConfig.DiskSizeGb = int64(v.(int))
+		}
+
+		if v, ok = nodeConfig["local_ssd_count"]; ok {
+			cluster.NodeConfig.LocalSsdCount = int64(v.(int))
 		}
 
 		if v, ok := nodeConfig["oauth_scopes"]; ok {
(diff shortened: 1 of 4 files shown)


danawillow:
Looks good overall. I recently made a change to the resource_container_cluster tests that needs updating in your PR, can you add something [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_container_cluster_test.go#L177)?

mrooding:
@danawillow I've updated the PR as requested.

builtin/providers/google/resource_container_cluster.go
```diff
@@ -231,6 +231,22 @@ func resourceContainerCluster() *schema.Resource {
...
+								value := v.(int)
+
+								if value < 0 {
+									errors = append(errors, fmt.Errorf(
+										"%q cannot be negative", k))
+								}
+								return
+							},
+						},
+
```
danawillow: Can you also update the `flattenClusterNodeConfig` method?

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -191,6 +191,7 @@ func testAccCheckContainerCluster(n string) resource.TestCheckFunc {
 			{"subnetwork", cluster.Subnetwork},
 			{"node_config.0.machine_type", cluster.NodeConfig.MachineType},
 			{"node_config.0.disk_size_gb", strconv.FormatInt(cluster.NodeConfig.DiskSizeGb, 10)},
+			{"node_config.0.local_ssd_count", strconv.FormatInt(cluster.NodeConfig.LocalSsdCount, 0)},
```
danawillow: The 0 here should be 10 (it's the base the number should be in)

builtin/providers/google/resource_container_cluster_test.go
```diff
@@ -339,6 +340,7 @@ resource "google_container_cluster" "with_node_config" {
 	node_config {
 		machine_type = "g1-small"
```
danawillow: The machine type here will need to be `n1-standard-1` (tests fail with "local SSDs require machine type n1-standard-1 or larger")

builtin/providers/google/resource_container_cluster.go
mrooding: 👍

mrooding:
I've updated everything as requested. The test that you spoke about, is that something that I can run myself next time? Will save us both some time if I can weed out this stuff myself next time.

danawillow:
Thanks! I'll merge this later today.

For running tests, you'll need a GCP project with enough quota to run the tests, and a service account with the right permissions (I have mine at project owner for simplicity). Make sure you download the private key for the service account and put it somewhere where you can find it.

Here's the command I use once that's all set up:
```
$ export GOOGLE_CREDENTIALS=$(cat [path-to-keyfile-json])
$ make testacc TEST=./builtin/providers/google GOOGLE_PROJECT=$PROJECT GOOGLE_REGION=$REGION GOOGLE_XPN_HOST_PROJECT=ignore TESTARGS='-run=TestAccContainerCluster_withNodeConfig'
```
The stuff after `run` is a regex (kind of), so you can run all the cluster tests for example with just `TestAccContainerCluster`.

Note that sometimes when tests fail they leave dangling resources, so you might want to periodically check and make sure you're not getting charged for resources you aren't actually using.

Let me know if you have any questions!

PR #12314, opened 2017-02-28
provider/aws: more details on which s3 bucket had an error

netjunki:
Before:
    * data.aws_s3_bucket_object.redshift_reader: Failed getting S3 object: NotFound: Not Found
    status code: 404, request id: B7B4BBDFBB9EE1FA

After:
    * data.aws_s3_bucket_object.redshift_reader: Failed getting S3 object: NotFound: Not Found
    status code: 404, request id: B7B4BBDFBB9EE1FA "s3://mybucket/lambda.zip"

### DIFF
--- a/builtin/providers/aws/data_source_aws_s3_bucket_object.go
+++ b/builtin/providers/aws/data_source_aws_s3_bucket_object.go
@@ -132,7 +132,7 @@ func dataSourceAwsS3BucketObjectRead(d *schema.ResourceData, meta interface{}) e
 	log.Printf("[DEBUG] Reading S3 object: %s", input)
 	out, err := conn.HeadObject(&input)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


stack72:
Hi @netjunki 

Thanks for the work here - I agree that some of our error message lack any real detail in this - this is not idea. I suggest we do something different here though. You have gone for the address as follows:

```
Failed getting S3 object: NotFound: Not Found
status code: 404, request id: B7B4BBDFBB9EE1FA "s3://mybucket/lambda.zip"
```

I am worried about that s3 address format - that may not be correct depending on region etc. Therefore, I suggest something like this:

```
Failed getting S3 object: NotFound: Not Found
status code: 404, request id: B7B4BBDFBB9EE1FA. Bucket: `mybucket`, Object: `lambda.zip`
```

This will point out both parts of the problem

What do you think?

Paul

netjunki:
@stack72 Totally reasonable change. Pushed. :-)

PR #12317, opened 2017-03-01
provider/google: fix url map test and update logic

danawillow:
Happened to randomly stumble across this. The logic for checking if the "test" field change had two moments where it was looking at the wrong field, and the test that was checking the update function wasn't actually doing an update (it was just creating/destroying resources).

(Side note- is there a function or something in a test that can be used to make sure something was updated and not recreated?)


stack72:
LGTM nice catch @danawillow :)

stack72:
@danawillow with regards to finding out if something is recreated - there is no built in function - I had to write one for DigitalOcean droplets

https://github.com/hashicorp/terraform/blob/master/builtin/providers/digitalocean/resource_digitalocean_droplet_test.go#L404

P.

PR #12321, opened 2017-03-01
Add auto_upgrade_minor_version documentation

momamene:
`auto_upgrade_minor_version` argument in `azurerm_virtual_machine_extension` is already implemented but not documented.

Reference
  - https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_virtual_machine_extension.go#L59
  - https://docs.microsoft.com/en-us/rest/api/compute/extensions/extensions-add-or-update


stack72:
LGTM - thanks for taking the time to help us out :)

PR #12324, opened 2017-03-01
No instance security groups and interface

sgran:
The comment makes this clear, but the code was still adding security
groups to the instance definition.  This moves that assignment to a code
path that evaluates whether an interface definition is being created.

Signed-off-by: Stephen Gran <stephen.gran@piksel.com>


gotoadmin:
Hello, 

When could we expect resolution of the issue ?

sgran:
there looks to be a very similar patch that has gone into master - https://github.com/hashicorp/terraform/commit/93c4730de74c0fec28aa616dcfd096661d9f5832

Dunno if it's been released yet, though.

sgran:
It looks like a similar patch set has gone in.  Closing this.

PR #12338, opened 2017-03-01
Circonus Provider

sean-:
Add a Circonus provider.


builtin/providers/circonus/resource_circonus_check_http.go
```diff
@@ -0,0 +1,381 @@
...
+		config.SubmissionURL:    struct{}{},
+	}
+
+	for k := range swamp {
+		if _, ok := whitelistedConfigKeys[k]; ok {
+			delete(c.Config, k)
+		}
+
+		if _, ok := whitelistedConfigKeys[k]; !ok {
+			panic(fmt.Sprintf("PROVIDER BUG: API Config not empty: %#v", swamp))
```
catsby: I admit I don't recall all the specifics of our video chat the other week, but I don't believe `panic` is appropriate here. Do you recall if we covered that? 

In generally I default to logging unexpected things and continuing on (or perhaps erring). I feel a panic would be a very bad thing for operators to receive.

builtin/providers/circonus/resource_circonus_check_http.go
```diff
@@ -0,0 +1,381 @@
...
+
+	if err := d.Set(checkHTTPAttr, schema.NewSet(hashCheckHTTP, []interface{}{httpConfig})); err != nil {
+		return errwrap.Wrapf(fmt.Sprintf("Unable to store check %q attribute: {{err}}", checkHTTPAttr), err)
+	}
+
+	return nil
+}
+
+// hashCheckHTTP creates a stable hash of the normalized values
+func hashCheckHTTP(v interface{}) int {
```
catsby: same question as before, regarding custom hash functions for Sets

builtin/providers/circonus/resource_circonus_check_http_test.go
```diff
@@ -0,0 +1,180 @@
...
+		CheckDestroy: testAccCheckDestroyCirconusCheckBundle,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCirconusCheckHTTPConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("circonus_check.jezebel", "active", "true"),
+					resource.TestCheckResourceAttr("circonus_check.jezebel", "collector.#", "1"),
+					resource.TestCheckResourceAttr("circonus_check.jezebel", "collector.2388330941.id", "/broker/1"),
+					resource.TestCheckResourceAttr("circonus_check.jezebel", "http.#", "1"),
+					// resource.TestCheckResourceAttr("circonus_check.jezebel", "http.4213422905.auth_method", ""),
```
catsby: Are these commented out to be restored later, or are they not testable?

builtin/providers/circonus/resource_circonus_check_http_test.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+const testAccCirconusCheckHTTPConfig = `
+variable "http_check_tags" {
+  type = "list"
+  default = [ "app:circonus", "app:jezebel", "lifecycle:unittests", "source:circonus" ]
+}
+
+resource "circonus_metric" "status_code" {
+  name = "code"
```
catsby: are metric names unique? Will any of these tests fail due to being ran concurrently with other tests in our CI? If so, we'll need constants like this test config to be made into methods and have randomized names inserted. See https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_autoscaling_policy_test.go#L18 for an example

builtin/providers/circonus/resource_circonus_check_caql.go
```diff
@@ -0,0 +1,89 @@
...
+var checkCAQLDescriptions = attrDescrs{
+	checkCAQLQueryAttr: "The query definition",
+}
+
+var schemaCheckCAQL = &schema.Schema{
+	Type:     schema.TypeSet,
+	Optional: true,
+	MaxItems: 1,
+	MinItems: 1,
+	Set:      hashCheckCAQL,
```
sean-: Discussed: `hashCheckCAQL()` trims the whitespace of strings before hashing the value because the API provider is doing the same.  For attributes that frequently have HEREDOC input, this is an important usability requirement.

builtin/providers/circonus/descriptions.go
```diff
@@ -0,0 +1,29 @@
...
+func init() {
+	// NOTE(sean@): needs to be completed
+	accountDescription = map[string]string{
+		accountContactGroupsAttr: "Contact Groups in this account",
+		accountInvitesAttr:       "Outstanding invites attached to the account",
+		accountUsageAttr:         "Account's usage limits",
+		accountUsersAttr:         "Users attached to this account",
+	}
+
+	// NOTE(sean@): needs to be completed
```
sean-: There are missing descriptions that need to be populated, but I'm going to punt on this for now and let the vendor populate these values.  Most of our providers don't seem to have this value populated so I didn't consider it mandatory.  LMK if this is considered mandatory or not.

builtin/providers/circonus/resource_circonus_check_http.go
```diff
@@ -0,0 +1,381 @@
...
+func checkAPIToStateHTTP(c *circonusCheck, d *schema.ResourceData) error {
+	httpConfig := make(map[string]interface{}, len(c.Config))
+
+	// swamp is a sanity check: it must be empty by the time this method returns
+	swamp := make(map[config.Key]string, len(c.Config))
+	for k, v := range c.Config {
+		swamp[k] = v
+	}
+
+	saveStringConfigToState := func(apiKey config.Key, attrName schemaAttr) {
```
sean-: This closure can be made into a function if necessary but it felt like the code was easier to read with a closure vs a helper function with an opaque use.

As for what the closure does, it stages a value into the map that is saved into the state store.  The purpose of this closure is to catch any unsupported or unanticipated drift in the configuration block that comes back from the API vs what is supported by the resource.

The `config` block is a json hash of arbitrary key/value pairs that changes based on the check type.

LMK.

builtin/providers/circonus/resource_circonus_check_http.go
sean-: Same answer as above, trimming whitespace for some descriptions is required.

builtin/providers/circonus/resource_circonus_check_http_test.go
sean-: Everything is testable.  :)  These are commented out because the ROI on testing every individual return parameter is pretty low.

In this case, we'd need to put together a valid end-to-end test and viable endpoint.  There are a number of permutations and combinations for how these values could be tested, but we don't have an endpoint that we could easily rig up.

builtin/providers/circonus/resource_circonus_check_http_test.go
sean-: Nope, these are local and won't conflict with anything.  Internally `circonus_metrics` are actually random UUIDs.  To be safe, I checked and don't see any metrics with the same `name`.  To the best of my knowledge, these tests can be run in parallel without issue (that's how I've been testing this code to date).  I think if there were two concurrent instances of `make test` running for this provider it could cause a problem, but I'm not sure it's worth solving.  At present, a full test run looks like (run only a few seconds ago):

```
2017/03/03 16:10:38 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/circonus -v  -timeout 120m
=== RUN   TestAccDataSourceCirconusAccount
--- PASS: TestAccDataSourceCirconusAccount (2.03s)
=== RUN   TestAccDataSourceCirconusCollector
--- PASS: TestAccDataSourceCirconusCollector (1.88s)
=== RUN   Test_MetricChecksum
--- PASS: Test_MetricChecksum (0.00s)
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccCirconusCheckCAQL_basic
--- PASS: TestAccCirconusCheckCAQL_basic (4.99s)
=== RUN   TestAccCirconusCheckCloudWatch_basic
--- PASS: TestAccCirconusCheckCloudWatch_basic (5.78s)
=== RUN   TestAccCirconusCheckHTTP_basic
--- PASS: TestAccCirconusCheckHTTP_basic (5.75s)
=== RUN   TestAccCirconusCheckHTTPTrap_basic
--- PASS: TestAccCirconusCheckHTTPTrap_basic (8.14s)
=== RUN   TestAccCirconusCheckICMPPing_basic
--- PASS: TestAccCirconusCheckICMPPing_basic (5.51s)
=== RUN   TestAccCirconusCheckJSON_basic
--- PASS: TestAccCirconusCheckJSON_basic (9.56s)
=== RUN   TestAccCirconusCheckMySQL_basic
--- PASS: TestAccCirconusCheckMySQL_basic (5.06s)
=== RUN   TestAccCirconusCheckPostgreSQL_basic
--- PASS: TestAccCirconusCheckPostgreSQL_basic (5.42s)
=== RUN   TestAccCirconusCheckTCP_basic
--- PASS: TestAccCirconusCheckTCP_basic (6.34s)
=== RUN   TestAccCirconusContactGroup_basic
--- PASS: TestAccCirconusContactGroup_basic (4.15s)
=== RUN   TestAccCirconusGraph_basic
--- PASS: TestAccCirconusGraph_basic (8.15s)
=== RUN   TestAccCirconusMetric_basic
--- PASS: TestAccCirconusMetric_basic (0.02s)
=== RUN   TestAccCirconusMetric_tagsets
--- PASS: TestAccCirconusMetric_tagsets (0.05s)
=== RUN   TestAccCirconusStreamGroup_basic
--- PASS: TestAccCirconusStreamGroup_basic (3.64s)
=== RUN   TestAccCirconusTrigger_basic
--- PASS: TestAccCirconusTrigger_basic (7.97s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/circonus       84.459s
```

catsby: > I think if there were two concurrent instances of make test running for this provider it could cause a problem

I believe this is how our CI is setup, each test is ran in isolation and possibly in parallel, is that correct @stack72 ?

builtin/providers/circonus/resource_circonus_rule_set.go
```diff
@@ -0,0 +1,859 @@
...
+func (rs *circonusRuleSet) Update(ctxt *providerContext) error {
+	_, err := ctxt.client.UpdateRuleSet(&rs.RuleSet)
+	if err != nil {
+		return errwrap.Wrapf(fmt.Sprintf("Unable to update rule set %s: {{err}}", rs.CID), err)
+	}
+
+	return nil
+}
+
+func (rs *circonusRuleSet) Validate() error {
```
stack72: Should we just kill this?

builtin/providers/circonus/resource_circonus_check_http_test.go
sean-: The race condition here would be *if* the same test is running at the same time and a unique resource name overlaps.  The tests are short lived so that race condition is on the order of 5s.  If that becomes an issue I can randomize easily enough but I think there's a low ROI on that work atm.

builtin/providers/circonus/resource_circonus_rule_set.go
sean-: If possible I want to leave it as a stub because I know there are a few client-side checks that we could provide to validate the inputs.  For instance:

```
  // For an 'on absence' rule this is the number of seconds the metric must not
  // have been collected for, and should not be lower than either the period or
  // timeout of the metric being collected.
```

As we flex this code path I plan on adding checks here where necessary because we can provide a more rich error message that provides context vs a `400` response code with a sometimes cryptic error message.

builtin/providers/circonus/resource_circonus_check_http_test.go
catsby: My concern was if for example `TestAccCirconusCheckMySQL_basic` could **not** be ran at the same time as say `TestAccCirconusTrigger_basic`. If that doesn't introduce conflict then we should be fine, especially if tests are ran serially by default 👍

builtin/providers/circonus/data_source_circonus_account.go
```diff
@@ -0,0 +1,269 @@
...
+
+	var cid string
+
+	var a *api.Account
+	var err error
+	if v, ok := d.GetOk(accountIDAttr); ok {
+		cid = v.(string)
+	}
+
+	if v, ok := d.GetOk(accountCurrentAttr); ok {
```
stack72: ok, after speaking to @sean- - my comment here isn't correct, this isn't being added to a struct to therefore, we don't need a default value - therefore, this is correct!

builtin/providers/circonus/resource_circonus_check_http_test.go
sean-: OH!  Yeah, then in this case we're 100% a-okay.

sean-: I took the preventative step of fully randomizing all resource names.  I do not believe this was necessary but just in case we make a radical change to our test runners in the future and go fully parallel on all tests, we're now covered.

PR #12340, opened 2017-03-01
return 0 from resourceCountMax with a nil Resource

### LINKED ISSUE #12253: Incorrect output value syntax check on lists causes panics
### Terraform Version
Terraform v0.9.0-dev (a7078ad89702b32cf12b4dba3052f80007f6a3e1)

### Affected Resource

terraform core

```hcl
variable "aws_access_key" {}
variable "aws_secret_key" {}

provider "aws" {
  access_key = "${var.aws_access_key}"
  secret_key = "${var.aws_secret_key}"
  region     = "eu-central-1"
}

output "account_id" {
  value = ["${aws_caller_identity.current.*.account_id}"]
}
```

### Panic Output
I get a SIGSEGV panic when I run a `terraform plan` or `terraform apply`

https://gist.github.com/nanoz/5e97febc225076dede93510e810ff641

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
In case you didn't notice, I forgot "data." in the value. It is only reproducible on list values, either on data sources or resources. I can't reproduce the bug on terraform 0.8.7.

jbardin:
If a resource can't be found when looking up an interpolated variable in the config, make sure an error is returned with a nil Resource.

Fixes #12253

### DIFF
--- a/terraform/interpolate.go
+++ b/terraform/interpolate.go
@@ -698,6 +698,10 @@ func (i *Interpolater) resourceCountMax(
 	// from the state. Plan and so on may not have any state yet so
 	// we do a full interpolation.
 	if i.Operation != walkApply {
+		if cr == nil {
+			return 0, nil
+		}
+
 		count, err := cr.Count()
 		if err != nil {
 			return 0, err
--- a/terraform/interpolate_test.go
+++ b/terraform/interpolate_test.go
@@ -876,6 +876,31 @@ func TestInterpolator_sets(t *testing.T) {
 		interfaceToVariableSwallowError(set))
 }
 
+// When a splat reference is made to a resource that is unknown, we should
+// return an error.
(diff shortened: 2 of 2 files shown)


mitchellh:
Some failing tests. I have a sneaking suspicion this fix is going to be more complicated than this though. In the past, blanket "not found" errors have caused various operations to fail (apply works but plan fails, for example) and has required more... nuance.

jbardin:
Ah you're right. I was relying on travis to do the full test suite for me ;) I'll bounce back to this in a bit.

jbardin:
The terraform tests were a couple cases where the test state didn't match the test fixture.

The repl tests broke because they rely on not having a config present. The update allow returning a nil resource without an error, and adds a nil check in the one codepath were it's missing.

PR #12350, opened 2017-03-01
command/state-push: support pushing from stdin

mitchellh:
Allow state push from stdin.

This makes it easy to script together a `state pull | state push` combination when working with different environments (requires different working directories so its not as simple, but still no intermediate files).

### DIFF
--- a/command/state_push.go
+++ b/command/state_push.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


mitchellh:
Tests all pass locally with `make test`, not sure what these Travis failures are since their logs aren't loading atm. :(

josephholsten:
oh, the terrible things my team will do with `jq(1)` now!

PR #12352, opened 2017-03-01
backend/atlas: convert to new style

mitchellh:
This converts the "atlas" backend to the new style. 

It doesn't yet contain any of the new work we still have planned prior to 0.9 final, this still uses the old API and just moves the logic over to the new helper/schema framework. I wanted to get at least this in for 0.9-beta2.

This is backwards compatible with the prior "atlas" backend without any changes.


backend/atlas/backend.go
```diff
@@ -0,0 +1,196 @@
...
+	b.once.Do(b.init)
+	return b.schema.Validate(c)
+}
+
+func (b *Backend) Configure(c *terraform.ResourceConfig) error {
+	b.once.Do(b.init)
+	return b.schema.Configure(c)
+}
+
+func (b *Backend) State() (state.State, error) {
```
jbardin: Doesn't this need to implement the new `backend.Backend` interface to be used?

mitchellh:
@jbardin Yep :) just pushed all those changes pu.

PR #12354, opened 2017-03-01
plugin: bump protocol version

mitchellh:
We changed the type of InstanceState.Meta and this will break
pre-existing plugins. They simply have to recompile to work.


mitchellh:
(Discussed with jbardin in Slack)

PR #12359, opened 2017-03-02
aws/provider: redshift cluster exercise invalid password char path

netjunki:
A quick test of the invalid password character code branch.


stack72:
LGTM! Thanks :)

PR #12364, opened 2017-03-02
state/azure: add environment option for non-public cloud usage

pmcatominey:
Acceptance tests passed for Public and German cloud environments with the exception of `TestAzureClientLease` which is broken due to an SDK bug.


stack72:
LGTM thanks for this @pmcatominey :)

PR #12366, opened 2017-03-02
website: document state environments

mitchellh:
Title is self-explanatory!


shanielh:
If I understand correct, there will be no interpolation variable `environment` or something that I may use to configure specific differences between the environments?

mitchellh:
@shanielh Actually we planned for that but not sure it'll make it in prior to 0.9. It'll come shortly after though.

mitchellh:
@shanielh Its going to make it in: https://github.com/hashicorp/terraform/pull/12660

shanielh:
@mitchellh Awesome, what about support of the [terraform remote state provider](https://www.terraform.io/docs/providers/terraform/d/remote_state.html)?

PR #12367, opened 2017-03-02
backend/consul: support named states

mitchellh:
This adds support for named states (environments) for Consul.

Most importantly: this adds the `backend.TestBackend` function that will automatically test all the functionality of named states if the backend reports it supports it. Hurray!

Named states for Consul work by appending "-env:NAME" to the given path for any non-default state. This enables it to continue working with previous versions of Terraform that didn't use environments.

For a sanity check for reviewers: I ran tests locally with `CONSUL_HTTP_ADDR` set and saw the API calls coming into my Consul logs. Travis won't run them so I wanted to make it clear I saw it actually work. :)

### DIFF
--- a/backend/remote-state/consul/backend.go
+++ b/backend/remote-state/consul/backend.go
@@ -6,67 +6,78 @@ import (
 
 	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/terraform/backend"
-	"github.com/hashicorp/terraform/backend/remote-state"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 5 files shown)


apparentlymart:
Is there some way we could allow an entirely different Consul server to be used (with the same key path) in each environment? I run a separate Consul cluster per environment and I assume that isn't an entirely original idea on my part, but I must admit I don't see a great way to make that work within the environment arch since there would be no way within consul itself to enumerate all the available environments, without the help of some other discovery system. Perhaps I would need a new backend that layers on top of others to list environments some other way but still use the state implementation from the consul backend?

PR #12368, opened 2017-03-02
terraform: run Gosched in test so other goroutines run

mitchellh:
Starting with Go 1.8 betas, we've periodically received SIGQUITs on our
tests in Travis. The stack trace looks like this:

https://gist.github.com/mitchellh/abf09b0980f8ea01269f8d9d6133884d

The tests are timing out! This is a test that hasn't been touched really
in a very long time and has always passed. I've **reproduced this
locally** by setting `GOMAXPROCS=1` and running the test. By yielding
the scheduler in the hot loop, it now passes almost instantly every
time.

My only theory is that in Go 1.8 the scheduler changed in some way
that it now sometimes doesn't schedule the other goroutine. This
is not in any way a bug in Go, of course, it is just old behavior we
depended on. Gosched fixes this.

Perhaps the test can be written in a different way, but this gets tests
passing and I think will fix our periodic errors.

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


jbardin:
A busy loop can't be preempted, which isn't a problem for the other goroutines on other threads, but it poses a problem for the garbage collector, because all threads need to be preempted to for the STW phase. 

My first guess is this wasn't a truly busy loop before, with the method call triggering a scheduling point, but is now being inlined completely (the generated machine code around this much more efficient in go1.8 -- the atomic operations are emitted directly as assembly rather than a CALL).

The new GC may change this behavior too, since it may be more aggressive trying to get a lock for the guaranteed shorter pauses.

Regardless, if you want a spin-lock in Go you usually need Gosched to make it work, or poll with a sleep. 

and that was a rather long LGTM ;)

PR #12369, opened 2017-03-02
provider/azurerm: Allow blob copy between storage accounts

rrudduck:
Currently, unless the blob is publically accessible, specifying the source_uri in the azurerm_storage_blob resource does not work. This change allows the source for a blob to be in any storage account to which the service principal has access.

Side node: I'm still having some issues with acceptance tests running correctly when creating storage accounts. I did validate this change however against our production subscription and it worked as designed.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_storage_blob.go
+++ b/builtin/providers/azurerm/resource_arm_storage_blob.go
@@ -205,7 +205,7 @@ func resourceArmStorageBlobCreate(d *schema.ResourceData, meta interface{}) erro
 			return fmt.Errorf("Source storage account %q not found", sourceAccount)
 		}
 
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
Hi @rrudduck

Thanks for your contribution! Apologies for the delay in reviewing this.

Whilst this PR looks in a pretty good shape, and the `azurerm_storage_blob` resource currently supports copying from a public URL; after chatting with some of the other maintainers we came to a conclusion that this functionality might be better within a tool such as [AZCopy](https://azure.microsoft.com/en-gb/blog/azcopy-5-1-release/) - which actually already [supports this functionality](https://docs.microsoft.com/en-us/azure/storage/storage-use-azcopy#blob-copy).

As such - I'm going to close this PR for the moment, but I'd like to thank you again for the contribution here.

Thanks!

PR #3584, opened 2015-10-21
provider/kubernetes: Pods Resource CRUD operations implemented

lwander:
The issue I was having in the old PR is actually a Terraform core issue, see #3575 

Currently the Update operation mirrors the Create operation, this is done to have Terraform updates mirror those of `kubectl`. 

I will add more acceptance tests soon, but I'm not sure how in depth the documentation should be.

@radeksimko @sparkprime

PR #3471, opened 2015-10-09
provider/kubernetes: Pods resource spec is finished

lwander:
@radeksimko 

No need to merge this yet, I just wanted to check with you about my approach to separating out the resource specs. Since you mentioned that there will be a lot sub-resource duplication, I hoisted the schemas into `gen_kubernetes_schemas.go`. Already it's easier to reuse shared sub-resources like this. 

Another thing - ever attribute is marked as `optional` as a way to keep the `spec` attribute for users who want it.

PR #3453, opened 2015-10-08
[WIP] Add Kubernetes provider

radeksimko:
Resubmitting #3132 from `hashicorp` branch since there will be a lot more work involved in this and @lwander volunteered to help out. :wink: 

Feel free to submit PRs against this new branch.

PR #3132, opened 2015-08-31
Add Kubernetes provider

radeksimko:
### Status (TL;DR)

 - [x] `kubernetes_limit_range` resource + acceptance test
 - [x] `kubernetes_namespace` resource + acceptance test
 - [x] `kubernetes_persistent_volume` resource + acceptance test
 - [x] `kubernetes_persistent_volume_claim` resource + acceptance test
 - [x] `kubernetes_pod` resource + acceptance test
 - [x] `kubernetes_replication_controller` resource + acceptance test
 - [x] `kubernetes_resource_quota` resource + acceptance test
 - [x] `kubernetes_service` resource + acceptance test
 - [x] Docs - up to date
 - [x] Example
 - [ ] Decide between `KUBERNETES_CLIENT_CERTIFICATE` / `KUBERNETES_CLIENT_CERTIFICATE_PATH` etc.
 - [ ] Support `kubernetes_namespace.finalizers`
 - [ ] Detect default/undefined fields and ignore these when generating diffs
  - [ ] `kubernetes_limit_range`
  - [ ] `kubernetes_persistent_volume`
  - [ ] `kubernetes_pod`
  - [ ] `kubernetes_replication_controller`
  - [ ] `kubernetes_service`

### Config ENV vars (w/ and w/out `_PATH`)

I think the config options should stay as are (accepting keys & crts contents as strings) to provide direct compatibility w/ `google_container_engine`, yet allowing people to use `file()`. This is however less convenient for ENV variables since keys & certs are typically multi-line strings and that's a bit tricky to handle in ENV vars. Hence I guess `_PATH` with some extra logic to read files wins here, but I'll leave this open for a while.

### `schema.CompareFunc` - Why

Creation / deletion of all resources works aok, but the problem is that K8S server returns specs with some extra attributes that we have not defined when sending it to the API originally and these are often not predictable (e.g. node IP address).

I experimented with a few ideas. I decided not to strip attributes inside state (via `StateFunc`), because I still think that user should be able to reference the real spec from server elsewhere. Hence I created `schema.CompareFunc` and I'm now trying to come up with a meaningful/universal way of stripping attributes that user has not defined in the given spec, so that we can only compare what **has** been defined, yet still save the original spec into state.

### Mind the test suite

It will also demonstrate the issue with spec as described above.
```
$ make testacc TEST=./builtin/providers/kubernetes 2>/dev/null
```

cc @sparkprime @lwander

PR #12372, opened 2017-03-02
kubernetes: Add provider + namespace resource

radeksimko:
I intentionally made this PR as minimal as possible (provider + 1 resource + relevant tests + docs) to make it easier for review. Other resources and potentially data sources may follow.

### Test plan

```
make testacc TEST=./builtin/providers/kubernetes
2017/03/02 11:54:13 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/kubernetes -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccKubernetesNamespace_basic
--- PASS: TestAccKubernetesNamespace_basic (6.66s)
=== RUN   TestAccKubernetesNamespace_importBasic
--- PASS: TestAccKubernetesNamespace_importBasic (5.78s)
=== RUN   TestAccKubernetesNamespace_generatedName
--- PASS: TestAccKubernetesNamespace_generatedName (5.63s)
=== RUN   TestAccKubernetesNamespace_importGeneratedName
--- PASS: TestAccKubernetesNamespace_importGeneratedName (6.14s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/kubernetes	24.288s
```

### History Lesson

There have been a couple of PRs with K8S provider in the past:

 - https://github.com/hashicorp/terraform/pull/3132
 - https://github.com/hashicorp/terraform/pull/3453
 - https://github.com/hashicorp/terraform/pull/3471
 - https://github.com/hashicorp/terraform/pull/3584

which were closed (unmerged) for a number of reasons, TL;DR:

 - representing object specs as JSON string turned out to have some significant flaws - https://github.com/hashicorp/terraform/pull/3132#issuecomment-137571095
 - some resources have significantly complex & nested schemas which in turn makes maintenance of the TF schema & relevant docs difficult - https://github.com/hashicorp/terraform/pull/3471

Both mentioned things are being addressed in https://github.com/radeksimko/terraform-gen, although we might not be able to address these as well as I'd hope, mainly because the source SDK structs aren't concrete enough for our schema (e.g. validation, list vs set, hashing mechanisms etc.). But at least folks can get going quickly from zero, generate skeleton of schema + docs and then modify and do the rest manually.


paddyforan:
Spent a couple days reviewing this in depth and reading up on the history, and it gets a :+1: from me. :)

radeksimko:
Thanks Paddy, I appreciate the extra time invested into the review!

paddyforan:
My pleasure! Sorry for the delay, I got confused and wanted to understand done background before signing off on it.

builtin/providers/kubernetes/provider.go
```diff
@@ -0,0 +1,171 @@
...
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/mitchellh/go-homedir"
+	kubernetes "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_5"
```
pires: Shouldn't client-go be used instead?

radeksimko: I really wanted to use that library when starting the work on K8S provider, but it was practically impossible to use it as a vendored dependency due to some upstream issues/conflicts which I described in https://github.com/hashicorp/terraform/pull/12230#issue-210035666

We might switch to that library at some point in the future to reduce the number of vendored LOC and because that's the library that should be used for that purpose, but the main goal is to get initial coverage of all stable (v1) k8s resources for now so re-vendoring is not currently on my radar.

I hope that's understandable 😃

pires: Totally understandable. Keep up!

PR #12375, opened 2017-03-02
provider/statuscake: use default status code list when updating test

andrebeat:
This PR fixes an issue whenever a StatusCake test is updated which would clear the list of status codes considered as errors.

According to the StatusCake API documentation (https://github.com/DreamItGetIT/statuscake/blob/master/tests.go#L93), when updating a test definition the `StatusCodes` definition is always updated. Since currently there's no support for setting the list of status codes in terraform this would replace the list with an empty list.

This PR fixes the issue by always providing the default list of status codes when updating a test.


stack72:
LGTM! Thanks for the work on this :)

PR #12389, opened 2017-03-02
Implement filter function

maikzumstrull:
This change adds an interpolation function `filter()` that allows selecting only the elements of a list that match given criteria. An example use is given in the documentation: After creating several instances spread out across several availability zones (using `count = <some number>` and `zone = "${element(var.some_list_of_zones, count.index)}"`), create lists of only the instances in this set that are in a particular zone.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -65,6 +65,7 @@ func Funcs() map[string]ast.Function {
 		"distinct":     interpolationFuncDistinct(),
 		"element":      interpolationFuncElement(),
 		"file":         interpolationFuncFile(),
+		"filter":       interpolationFuncFilter(),
 		"floor":        interpolationFuncFloor(),
 		"format":       interpolationFuncFormat(),
 		"formatlist":   interpolationFuncFormatList(),
@@ -638,6 +639,44 @@ func appendIfMissing(slice []string, element string) []string {
 		}
 	}
 	return append(slice, element)
+}
+
+// interpolationFuncFilter implements the "filter" function that
+// removes elements from a list
+func interpolationFuncFilter() ast.Function {
+	return ast.Function{
+		ArgTypes:   []ast.Type{ast.TypeList, ast.TypeList, ast.TypeString},
+		ReturnType: ast.TypeList,
+		Callback: func(args []interface{}) (interface{}, error) {
+			var list []string
+
+			argument, _ := args[0].([]ast.Variable)
+			keys, _ := args[1].([]ast.Variable)
+			searchkey, _ := args[2].(string)
+
+			if len(argument) != len(keys) {
+				return nil, fmt.Errorf("length of lists should be equal")
+			}
+
+			for i, element := range argument {
+				if element.Type != ast.TypeString {
+					return nil, fmt.Errorf(
+						"only works for flat lists, this list contains elements of %s",
+						element.Type.Printable())
+				}
+				if keys[i].Type != ast.TypeString {
+					return nil, fmt.Errorf(
+						"only works for flat lists, this list contains elements of %s",
+						keys[i].Type.Printable())
+				}
+				if keys[i].Value.(string) == searchkey {
+					list = append(list, element.Value.(string))
+				}
+			}
+
+			return stringSliceToVariableValue(list), nil
+		},
+	}
 }
 
 // interpolationFuncJoin implements the "join" function that allows
--- a/config/interpolate_funcs_test.go
+++ b/config/interpolate_funcs_test.go
@@ -883,6 +883,43 @@ func TestInterpolateFuncDistinct(t *testing.T) {
 	})
 }
 
+func TestInterpolateFuncFilter(t *testing.T) {
+	testFunction(t, testFunctionConfig{
+		Cases: []testFunctionCase{
+			// normal usage
+			{
+				`${filter(list("a", "b", "c"), list("ref1", "ref2", "ref3"), "ref2")}`,
+				[]interface{}{"b"},
+				false,
+			},
+			// zero case
+			{
+				`${filter(list(), list(), "nope")}`,
+				[]interface{}{},
+				false,
+			},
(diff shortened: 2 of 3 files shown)


config/interpolate_funcs.go
```diff
@@ -640,6 +641,44 @@ func appendIfMissing(slice []string, element string) []string {
...
+func interpolationFuncFilter() ast.Function {
+	return ast.Function{
+		ArgTypes:   []ast.Type{ast.TypeList, ast.TypeList, ast.TypeString},
+		ReturnType: ast.TypeList,
+		Callback: func(args []interface{}) (interface{}, error) {
+			var list []string
+
+			argument, _ := args[0].([]ast.Variable)
+			keys, _ := args[1].([]ast.Variable)
+			searchkey, _ := args[2].(string)
```
radeksimko: Is there any specific reason you decided to silence the errors here? What if the user misreads or just doesn't read the docs, or just simply makes a mistake in providing the right argument types?

config/interpolate_funcs.go
```diff
@@ -640,6 +641,44 @@ func appendIfMissing(slice []string, element string) []string {
...
+
+			for i, element := range argument {
+				if element.Type != ast.TypeString {
+					return nil, fmt.Errorf(
+						"only works for flat lists, this list contains elements of %s",
+						element.Type.Printable())
+				}
+				if keys[i].Type != ast.TypeString {
+					return nil, fmt.Errorf(
+						"only works for flat lists, this list contains elements of %s",
```
radeksimko: Can we be a bit more specific here in the error message so it's more clear _which_ list (e.g. `the list of keys contains %s`)?

config/interpolate_funcs.go
```diff
@@ -640,6 +641,44 @@ func appendIfMissing(slice []string, element string) []string {
...
+						keys[i].Type.Printable())
+				}
+				if keys[i].Value.(string) == searchkey {
+					list = append(list, element.Value.(string))
+				}
+			}
+
+			return stringSliceToVariableValue(list), nil
```
radeksimko: Was it an intention here to just return an empty list? (genuine question - not suggesting it should be changed!)
If it was - that's fine - but I think we should document that in the docs, e.g. `No match will result in empty list` - just to avoid questions whether or not it will error out in such cases.

config/interpolate_funcs.go
alvelcom: Type checking in this case is enforced by ast.Functions's ArgType, so if user put a wrong argument, he would get an error like that:

```
* google_compute_instance_group.kube_minion[0]: At column 3, line 1: filter: argument 1 should be type list, got type string in:

${filter("test",
      google_compute_instance.kube_minion.*.zone,
      element(var.cluster_zones, count.index))}
```

config/interpolate_funcs.go
alvelcom: Will add, thanks.

config/interpolate_funcs.go
alvelcom: That was intended, I'll document it.

apparentlymart:
Hi @maikzumstrull!

Sorry to dive in here, but @radeksimko and I were chatting about this and agreed that the name `filter` feels a little too generic for this function, which supports only a very specific use-case of filtering.

We chatted about a few different formulations of this and landed on one that we thought would be easier to understand for those seeing it in code or in the docs. I'll copy your use-case from your documentation as an example:

```
matchingkeys(map(aws_instance.example.*.id, aws_instance.example.*.availability_zone), list("us-west-2a")))
```

This differs from your formulation in two ways, aside from the name changing:

* Rather than providing two lists, we instead provide a map. Users are likely to already be familiar with the idea of a map providing pairs of corresponding keys and values, so this avoids the need to re-explain that idea just for the sake of this function. An implication of this is that it's not possible to have the same key twice in the input map, but I think the main use-cases for this don't require that.
* The argument to filter by is a list of strings rather than just a single string. This means you can potentially search for, say, `list("us-west-2a", "us-west-2b")` to get keys matching both, which hopefully means more use-cases can be met by this function with only a small amount of additional code overhead in the since-item case.

I think this different signature and name would make this function easier to understand/explain, and also leave open the more-generic function name `filter` in case we have a more generic list/map filtering function in future.

How does that seem to you, @maikzumstrull? If that seems like a reasonable change but you don't have time to update the code, I'm happy to tweak it on your behalf. If you see important use-cases that aren't supported by this new formulation please let me know!

maikzumstrull:
Hi Radek and Martin,

sorry for the confusion, but the author of the code is actually @alvelcom. We filed the PR through my account because he didn't have access to his while traveling. The commits have the correct email address.

He has access to his account now, so maybe we should just close this and refile for clarity?

As for your proposal: I find the map-based syntax very confusing, in a "what is that map object doing there?" way. I don't like the proposed list-based one all that much either, though - it's kind of required by the Terraform type model because there is no directly accessible "list of things that have attributes" type.

The idea of allowing for multiple matches seems potentially useful. I had also been thinking if there might be away to allow relations (return things where foo < 5), but again, the syntax and type system probably doesn't express it well.

radeksimko:
> He has access to his account now, so maybe we should just close this and refile for clarity?

@maikzumstrull up to you two

I'm leaving the review to @apparentlymart as he has more insight into this specific problem.

alvelcom:
Hi @apparentlymart

Thanks for your comments, I've fixed the issues you've pointed out. Since I don't own this PR, I've created a new one: https://github.com/hashicorp/terraform/pull/13847. Here is list of major changes:

* renamed to more specific name;
* use a list of strings instead of just string as a third argument.

I haven't updated this function for using maps, because I don't think it commits for readability too much–in both cases if you hadn't seen this function before, you would need to check the documentation. In the same time I believe the example in docs is quite explanatory.

Thanks!

apparentlymart:
Thanks for the new PR, @alvelcom. I'll close this one!

PR #12396, opened 2017-03-03
provider/pagerduty: Remove deprecated name_regex field 

heimweh:
This PR removes the deprecated field `name_regex` as well as the old behaviour of looking up vendors.

Existing configurations using the `name_regex` field will render:

```bash
1 error(s) occurred:

* data.pagerduty_vendor.test: "name_regex": [REMOVED] Use `name` instead. This attribute will be removed in a future version
```

Not sure if this is the best way to proceed so any feedback is more than welcome :)


stack72:
Just a quick note here that this is *not* being back ported to 0.8.x - this is only going to be a 0.9 series inclusion

stack72:
LGTM :)

PR #12397, opened 2017-03-03
provider/google: retry storage creation/deletion on rate limiting

paddyforan:
Our GCP storage tests are really flaky right now due to rate limiting.
In theory, this could also impact Terraform users that are
deleting/creating large numbers of Google Cloud Storage buckets at once.

To fix, I'm detecting the specific error code that GCP returns when it's
a rate limit error, and using that with resource.Retry to try the
request again.

### DIFF
--- a/builtin/providers/google/resource_storage_bucket.go
+++ b/builtin/providers/google/resource_storage_bucket.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"time"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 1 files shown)


radeksimko:
I wish the GCloud SDK just retried automatically in such cases and even more intelligently (by using the `Retry-After` header), like AWS SDK does it:

https://github.com/aws/aws-sdk-go/blob/master/aws/request/retryer.go#L26-L42

paddyforan:
@radeksimko thanks, agreed. I'd open a bug or PR on the repo, but it's an auto-generated library. :-/

@danawillow or @evandbrown, what's the best way to get this fixed upstream in the SDK?

PR #12401, opened 2017-03-03
[azurerm/doc] - Adds exposed attributes to the docs. 

mzupan:
These attributes are exposed but missing in the docs.


stack72:
Thanks for this @mzupan :) LGTM!

PR #12402, opened 2017-03-03
[azurerm/docs] - adding missing subnet attributes

mzupan:
These attributes are exposed but missing from the docs.


stack72:
LGTM thanks for this :)

PR #12409, opened 2017-03-03
Updated list of available services

GuiSim:
http://docs.aws.amazon.com/general/latest/gr/aws-ip-ranges.html#aws-ip-download
> Valid values: AMAZON | EC2 | CLOUDFRONT | ROUTE53 | ROUTE53_HEALTHCHECKS | S3


stack72:
LGTM! thanks for this

PR #12411, opened 2017-03-03
provider/google: Cloud router resource

drebes:
This adds support for Cloud Router, which allows BGP announcement of routes through Cloud VPN tunnels.

It's still missing documentation, I'll provide that once the code is reviewed.

It adds the following resources:

- `google_compute_router`
- `google_compute_router_interface`
- `google_compute_router_peer`

It also modifies `google_compute_vpn_tunnel` to support Cloud Router.

A full test is provided on resource_compute_router_peer_test.go

```
$ make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccComputeRouter'
2017/03/03 15:38:06 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccComputeRouter -timeout 120m
=== RUN   TestAccComputeRouterInterface_import
--- PASS: TestAccComputeRouterInterface_import (197.04s)
=== RUN   TestAccComputeRouterPeer_import
--- PASS: TestAccComputeRouterPeer_import (218.85s)
=== RUN   TestAccComputeRouter_import
--- PASS: TestAccComputeRouter_import (83.30s)
=== RUN   TestAccComputeRouterInterface_basic
--- PASS: TestAccComputeRouterInterface_basic (378.11s)
=== RUN   TestAccComputeRouterPeer_basic
--- PASS: TestAccComputeRouterPeer_basic (410.42s)
=== RUN   TestAccComputeRouter_basic
--- PASS: TestAccComputeRouter_basic (96.71s)
=== RUN   TestAccComputeRouter_noRegion
--- PASS: TestAccComputeRouter_noRegion (62.47s)
=== RUN   TestAccComputeRouter_networkLink
--- PASS: TestAccComputeRouter_networkLink (84.34s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	1531.337s
```


builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+		},
+	})
+}
+
+func testAccCheckComputeRouterPeerDestroy(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		config := testAccProvider.Meta().(*Config)
+		project := config.Project
+
+		routersService := compute.NewRoutersService(config.clientCompute)
```
paddycarver: `config.clientCompute.Routers`?

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+					"google_compute_router_peer.foobar"),
+			},
+		},
+	})
+}
+
+func testAccCheckComputeRouterPeerDestroy(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		config := testAccProvider.Meta().(*Config)
+		project := config.Project
```
paddycarver: Would love to see this support the `getProject` checks the same way the resource does, to avoid nasty bugs down the road.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+		routersService := compute.NewRoutersService(config.clientCompute)
+
+		for _, rs := range s.RootModule().Resources {
+			if rs.Type != "google_compute_router_peer" {
+				continue
+			}
+
+			region := rs.Primary.Attributes["region"]
```
paddycarver: Would love to see this support the `getRegion` checks the same way the resource does, to avoid nasty bugs down the road.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+			name := rs.Primary.Attributes["name"]
+			routerName := rs.Primary.Attributes["router"]
+
+			router, err := routersService.Get(project, region, routerName).Do()
+
+			if err != nil {
+				return fmt.Errorf("Error Reading Router %s: %s", routerName, err)
+			}
+
+			var peerExists bool = false
```
paddycarver: `= false` is unnecessary here.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+			var peers []*compute.RouterBgpPeer = router.BgpPeers
+			for _, peer := range peers {
```
paddycarver: Why not just range over `router.BgpPeers`?

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+				if peer.Name == name {
```
paddycarver: Couldn't this just return an error now, obviating the need for `peerExists`?

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		config := testAccProvider.Meta().(*Config)
+		name := rs.Primary.Attributes["name"]
+		routerName := rs.Primary.Attributes["router"]
+		region := rs.Primary.Attributes["region"]
```
paddycarver: Would really like to see this honor the same logic as `getRegion`, just like the resource.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+		project := config.Project
```
paddycarver: Would like to see this honor the same logic as `getProject`, just like the resource.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+		routersService := compute.NewRoutersService(config.clientCompute)
```
paddycarver: `config.clientCompute.Routers`?

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+		router, err := routersService.Get(project, region, routerName).Do()
+
+		if err != nil {
+			return fmt.Errorf("Error Reading Router %s: %s", routerName, err)
+		}
+
+		var peerExists bool = false
```
paddycarver: This `= false` is unnecessary.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+		var peers []*compute.RouterBgpPeer = router.BgpPeers
+		for _, peer := range peers {
```
paddycarver: Why not just range over `router.BgpPeers`?

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+			if peer.Name == name {
+				peerExists = true
```
paddycarver: If you just return the nil here, we don't need `peerExists` anymore.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+		if !peerExists {
+			return fmt.Errorf("Peer %s not found for router %s", name, router.Name)
+		}
+
+		return nil
+	}
+}
+
+var testAccComputeRouterPeer_basic = fmt.Sprintf(`
```
paddycarver: Would love to see this as a function with parameters passed in.

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,263 @@
...
+  region  = "${google_compute_router.foobar.region}"
+  ip_address = "169.254.3.2"
+  asn = 65515
+  interface = "${google_compute_router_interface.foobar.name}"
+}`, acctest.RandString(10), acctest.RandString(10), acctest.RandString(10),
+	acctest.RandString(10), acctest.RandString(10), acctest.RandString(10),
+	acctest.RandString(10), acctest.RandString(10), acctest.RandString(10),
+	acctest.RandString(10), acctest.RandString(10))
+
+var testAccComputeRouterPeer_keepRouter = fmt.Sprintf(`
```
paddycarver: Would love to see this as a function, with parameters passed in.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+			},
+		},
+	})
+}
+
+func testAccCheckComputeRouterDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
+	project := config.Project
+
+	routersService := compute.NewRoutersService(config.clientCompute)
```
paddycarver: `config.clientCompute.Routers`?

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+				Check: testAccCheckComputeRouterExists(
+					"google_compute_router.foobar"),
+			},
+		},
+	})
+}
+
+func testAccCheckComputeRouterDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
+	project := config.Project
```
paddycarver: Would like to see this support the same `getProject` logic the resource does, to avoid future bugs.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	routersService := compute.NewRoutersService(config.clientCompute)
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "google_compute_router" {
+			continue
+		}
+
+		region := rs.Primary.Attributes["region"]
```
paddycarver: Would like to see this support the same `getRegion` logic the resource does, to avoid future bugs.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		config := testAccProvider.Meta().(*Config)
+		name := rs.Primary.Attributes["name"]
+		region := rs.Primary.Attributes["region"]
```
paddycarver: Would like to see this support the same `getRegion` logic the resource does.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+		project := config.Project
```
paddycarver: Would like to see this support the same `getProject` logic the resource does.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+		routersService := compute.NewRoutersService(config.clientCompute)
```
paddycarver: `config.clientCompute.Routers`?

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+		if err != nil {
+			return fmt.Errorf("Error Reading Router %s: %s", name, err)
+		}
+
+		return nil
+	}
+}
+
+var testAccComputeRouter_basic = fmt.Sprintf(`
```
paddycarver: Would love to see this as a function instead of a variable.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+        name = "router-test-%s"
+	region = "${google_compute_subnetwork.foobar.region}"
+	network = "${google_compute_network.foobar.name}"
+        bgp { 
+           asn = 64514
+       }
+}
+`, acctest.RandString(10), acctest.RandString(10), acctest.RandString(10))
+
+var testAccComputeRouter_noRegion = fmt.Sprintf(`
```
paddycarver: Would love to see this as a function instead of a variable.

builtin/providers/google/resource_compute_router_test.go
```diff
@@ -0,0 +1,176 @@
...
+resource "google_compute_router" "foobar" {
+        name = "router-test-%s"
+	network = "${google_compute_network.foobar.name}"
+        bgp {
+           asn = 64514
+       }
+}
+`, acctest.RandString(10), acctest.RandString(10), acctest.RandString(10))
+
+var testAccComputeRouter_networkLink = fmt.Sprintf(`
```
paddycarver: Would love to see this as a function instead of a variable.

builtin/providers/google/resource_compute_vpn_tunnel.go
```diff
@@ -77,6 +78,7 @@ func resourceComputeVpnTunnel() *schema.Resource {
 				Type:     schema.TypeSet,
 				Optional: true,
 				ForceNew: true,
+				Computed: true,
```
paddycarver: For posterity, why does this need to be computed?

builtin/providers/google/resource_compute_vpn_tunnel_test.go
```diff
@@ -154,6 +173,67 @@ resource "google_compute_vpn_tunnel" "foobar" {
 	acctest.RandString(10), acctest.RandString(10), acctest.RandString(10),
 	acctest.RandString(10), acctest.RandString(10))
 
+var testAccComputeVpnTunnel_router = fmt.Sprintf(`
```
paddycarver: Would be great to turn this into a function.

builtin/providers/google/resource_compute_vpn_tunnel_test.go
```diff
@@ -154,6 +173,67 @@ resource "google_compute_vpn_tunnel" "foobar" {
...
+resource "google_compute_forwarding_rule" "foobar_udp4500" {
+	name = "tunnel-test-%s"
+	region = "${google_compute_forwarding_rule.foobar_udp500.region}"
+	ip_protocol = "UDP"
+	port_range = "4500-4500"
+	ip_address = "${google_compute_address.foobar.address}"
+	target = "${google_compute_vpn_gateway.foobar.self_link}"
+}
+resource "google_compute_router" "foobar"{
+	name = "tunnel-test-router"
```
paddycarver: Doesn't this have to be unique? If so, for resiliency, I'd like to have it randomized so we don't have collisions if the test runner fails and leaves dangling resources.

drebes:
Hi, thanks for the feedback! I'll need some time to go through all the comments but will definitely do it.

paddycarver:
Awesome, thanks! Feel free to tag off or ask for help or clarification if we can assist somehow. :)

builtin/providers/google/resource_compute_router.go
```diff
@@ -0,0 +1,251 @@
...
+
+			return nil
+		}
+
+		return fmt.Errorf("Error Reading Router %s: %s", name, err)
+	}
+
+	d.Set("self_link", router.SelfLink)
+
+	// if we don't have a network (when importing), set it to the URI returned from the server
```
drebes: When the user specifies the network just by name, and we use getNetworkLink to get the URL to be passed to the API, setting it back on read will cause a delta that triggers router recreation. Any way to avoid that?

builtin/providers/google/resource_compute_router_peer.go
```diff
@@ -0,0 +1,290 @@
...
+		log.Printf("[DEBUG] Router %s/%s had no peer %s already", region, routerName, peerName)
+	}
+
+	return nil
+}
+
+func resourceComputeRouterPeerImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	parts := strings.Split(d.Id(), "/")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("Invalid router specifier. Expecting {region}/{router}")
```
drebes: I wanted to say that we expect the user to define it as <region_name>/<router_name>. How about:
"Invalid router specifier. Please define it as <region_name>/<router_name>"

builtin/providers/google/resource_compute_vpn_tunnel.go
drebes: When a tunnel is created attached to a router, it will have a default policy (local and remote traffic selectors) of "0.0.0.0/0". We already had local_traffic_selector as computed, but would not get a default remote_traffic_selector from the server until now. With router based tunnels, we get them, so the need for the Computed.

drebes:
Hi @paddycarver , I've implemented most of the changes, and replied to the comments where just clarification was needed.

Sorry for the large PR, creating a router without peers and interfaces is not functional so I thought it made more sense to send all as a single PR. Will send smaller, incremental ones in the future.

What did you mean by "Exists" support? Can you point me to an example?

I still need to add some docs, will work on it from now.

drebes:
Done with the docs, just missing feedback on the points replied above.

paddycarver:
Resolved merge conflicts, will review now. :)

builtin/providers/google/resource_compute_router_peer.go
```diff
@@ -0,0 +1,291 @@
...
+
+		if peer.Name == peerName {
+			peerFound = true
+			continue
+		} else {
+			newPeers = append(newPeers, peer)
+		}
+	}
+
+	if !peerFound {
```
paddycarver: nit: probably not worth updating, but I think `if len(newPeers) == len(router.BgpPeers)` would work as well, here. Which would mean `peerFound` isn't necessary. Small detail, not big enough to block merge.

builtin/providers/google/resource_compute_router_peer.go
```diff
@@ -0,0 +1,291 @@
...
+
+	for _, peer := range router.BgpPeers {
+
+		if peer.Name == peerName {
+			d.SetId(fmt.Sprintf("%s/%s/%s", region, routerName, peerName))
+			d.Set("interface", peer.InterfaceName)
+			d.Set("peer_ip_address", peer.PeerIpAddress)
+			d.Set("peer_asn", peer.PeerAsn)
+			d.Set("advertised_route_priority", peer.AdvertisedRoutePriority)
+			d.Set("ip_address", peer.IpAddress)
```
paddycarver: Would love to see project and region persisted here, in case they change in the provider config.

builtin/providers/google/resource_compute_router_peer.go
paddycarver: I think I'm missing something. Why are we expecting three parts (line 281) if we're expecting only a single `/` as input? Where in that expecting string is the "name" portion that's being set in state below?

builtin/providers/google/resource_compute_router_peer_test.go
```diff
@@ -0,0 +1,310 @@
...
+	subnet := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	address := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	gateway := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	espRule := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	udp500Rule := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	udp4500Rule := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	router := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	tunnel := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	iface := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
+	peer := fmt.Sprintf("router-peer-test-%s", acctest.RandString(10))
```
paddycarver: Same thing here.

builtin/providers/google/resource_compute_vpn_tunnel.go
paddycarver: Thanks!

builtin/providers/google/resource_compute_vpn_tunnel_test.go
```diff
@@ -32,6 +32,35 @@ func TestAccComputeVpnTunnel_basic(t *testing.T) {
...
+func TestAccComputeVpnTunnel_router(t *testing.T) {
+	network := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	subnet := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	address := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	gateway := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	espRule := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	udp500Rule := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	udp4500Rule := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	router := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
+	tunnel := fmt.Sprintf("router-interface-test-%s", acctest.RandString(10))
```
paddycarver: Same thing here.

paddycarver:
> What did you mean by "Exists" support? Can you point me to an example?

We have an Exists lifecycle method (like Read, Update, Destroy) that runs before some of the other methods, allowing Terraform to check if a resource with the given ID exists or not.

>    // Exists is a function that is called to check if a resource still
>    // exists. If this returns false, then this will affect the diff
>    // accordingly. If this function isn't set, it will not be called. It
>    // is highly recommended to set it. The *ResourceData passed to Exists
>    // should _not_ be modified.

(from https://godoc.org/github.com/hashicorp/terraform/helper/schema#Resource)

An example of it being used is here: https://github.com/hashicorp/terraform/blob/2a36e75b9e26cf3a6e1e4eedc30a79517ee35d75/builtin/providers/google/resource_container_node_pool.go#L167-L191

It's not _necessary_, but is nice to have. At this point in the review of this resource, I'd be happy to accept it without Exists, and iterate Exists into... existence later. But if you feel the urge, there's the info. :)

Thanks for your patience and your hard work on this.

builtin/providers/google/provider.go
```diff
@@ -5,12 +5,16 @@ import (
...
 
+	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
 )
 
+// Global MutexKV
+var mutexKV = mutexkv.NewMutexKV()
```
drebes: Interfaces and Peers are not first class objects in the API, but attributes of the routers. As they don't have their own create/delete methods, but instead are created/destroyed with  updates to the router that hosts them, the mutex is to ensure that no concurrent access on the router due to different peers/interfaces breaks the read -> add/remove -> write create/delete logic.

drebes:
Thanks for the clarification on the Exists support. Still not clear to me is when the function is called during the resource lifecycle, but I can check that. If that would give me reasonable confidence that I can get rid of the multiple checks for existence I do on read/create/delete, I see the value in adding it.

The other changes should be implemented, and the mutex question I've answered above.

builtin/providers/google/provider.go
drebes: BTW, that's the same reason I test deletion of peers/interfaces by using a second test step with similar config. (with the omitted resource, but same testID). If I just tested destroy, this would delete the hosted router, which of course would delete the peer/interface, but not test the proper deletion of the peer/interface only.

PR #12413, opened 2017-03-03
Sort Validation warning and errors

### LINKED ISSUE #10434: Random error output ordering
### Terraform Version
0.7.11

### Affected Resource(s)
All.

### Debug Output

Running the same command twice produces different set of errors each time without changing the configuration.

```
/data/env-dev # terraform plan -out terra.plan
var.rancher_command
  Enter a value: echo

fix these before continuing.
Errors:

  * At column 3, line 1: lookup: argument 1 should be type map, got type string in:

${lookup(element(var.cloudwatch_alerts, count.index), "statistic")}
  * At column 3, line 1: lookup: argument 1 should be type map, got type string in:

${lookup(element(var.cloudwatch_alerts, count.index), "comparison_operator")}
  * At column 3, line 1: lookup: argument 1 should be type map, got type string in:

${lookup(element(var.cloudwatch_alerts, count.index), "statistic")}
/data/env-dev # terraform plan -out terra.plan
var.rancher_command
  Enter a value: echo

fix these before continuing.

Errors:

  * At column 3, line 1: lookup: argument 1 should be type map, got type string in:

${lookup(element(var.cloudwatch_alerts, count.index), "evaluation_periods")}
  * At column 3, line 1: lookup: argument 1 should be type map, got type string in:

${lookup(element(var.cloudwatch_alerts, count.index), "alarm_description")}
  * At column 3, line 1: lookup: argument 1 should be type map, got type string in:

${lookup(element(var.cloudwatch_alerts, count.index), "alarm_description")}
```
### Expected Behavior
List either all errors or a selection of the same errors every time.

### Actual Behavior
Random list of errors is displayed on each run.

### Steps to Reproduce
As above.

jbardin:
Make the UI output consistent

fixes #10434

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -665,6 +665,12 @@ func (c *Context) Validate() ([]string, []error) {
 
 	// Return the result
 	rerrs := multierror.Append(errs, walker.ValidationErrors...)
+
+	sort.Strings(walker.ValidationWarnings)
+	sort.Slice(rerrs.Errors, func(i, j int) bool {
+		return rerrs.Errors[i].Error() < rerrs.Errors[j].Error()
+	})
+
 	return walker.ValidationWarnings, rerrs.Errors
 }
 


jbardin:
I didn't manage to get a test fixture to return more than one error, rather than a single multierror.

PR #12414, opened 2017-03-03
Skip tag operations on cloudwatch logs in govcloud partition.

jmcarp:
AWS GovCloud doesn't support tag operations on cloudwatch log groups:

```
± aws logs list-tags-log-group --log-group-name some-log-group
An error occurred (UnknownOperationException) when calling the ListTagsLogGroup operation:

± aws logs tag-log-group --log-group-name some-log-group --tags Name=test
An error occurred (UnknownOperationException) when calling the TagLogGroup operation:
```

This patch skips adding, removing, and listing tags on cloudwatch log groups when the partition is `aws-us-gov`.

cc @LinuxBozo

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudwatch_log_group.go
+++ b/builtin/providers/aws/resource_aws_cloudwatch_log_group.go
@@ -91,11 +91,13 @@ func resourceAwsCloudWatchLogGroupRead(d *schema.ResourceData, meta interface{})
 		d.Set("retention_in_days", lg.RetentionInDays)
 	}
 
-	tags, err := flattenCloudWatchTags(d, conn)
-	if err != nil {
-		return err
+	if meta.(*AWSClient).partition != "aws-us-gov" {
+		tags, err := flattenCloudWatchTags(d, conn)
+		if err != nil {
+			return err
+		}
+		d.Set("tags", tags)
 	}
-	d.Set("tags", tags)
 
 	return nil
 }
@@ -152,7 +154,7 @@ func resourceAwsCloudWatchLogGroupUpdate(d *schema.ResourceData, meta interface{
 		}
 	}
 
-	if d.HasChange("tags") {
+	if meta.(*AWSClient).partition != "aws-us-gov" && d.HasChange("tags") {
 		oraw, nraw := d.GetChange("tags")
 		o := oraw.(map[string]interface{})
 		n := nraw.(map[string]interface{})


apparentlymart:
Thanks for this, @jmcarp! I'm going to leave it to another Terraform maintainer to review/test since I don't have a govcloud account to try it with, but I do have a question:

Do you know that this limitation is intentional on the part of AWS? I wonder if this is a bug on their end that could be fixed, rather than adding this special case to Terraform. It seems like a rather arbitrary restriction, but perhaps this sort of additional constraint is normal for GovCloud?

jmcarp:
I couldn't find any mention of this limitation in the GovCloud docs, but that's not unusual for GovCloud limitations. I completely agree that this should be fixed on AWS and not terraform, but in my experience, it can take a while for GovCloud to catch up with the main partition. I'll check with AWS support and report back here if I learn anything.

apparentlymart:
Thanks @jmcarp! If this is indeed something that isn't going to get fixed for a while then working around it in Terraform would make sense, though perhaps we should make Terraform fail with an error message if any tags are provided in config in this case so that Terraform isn't just silently ignoring them, which would be particularly confusing if GovCloud later got support and it looked like a Terraform bug that the tags would not apply.

jmcarp:
@apparentlymart @paddyforan: here's what I heard from AWS support:

> The tagging issue is a known limitation that the service team is actively working. They expect this feature to be available early Q2.

My team doesn't have an urgent need to update to 0.8.x, so I don't have a strong preference for (a) fixing this patch up and reverting in a few months or (b) doing nothing and waiting for AWS to fix. WDYT?

realflash:
I would dearly love access to the local data provider so I vote for patch it now, please. Who knows when Amazon will actually deliver this functionality, if they ever do? The number of services in govcloud is about 20% of those in other partitions; being crippled is practically part of its raison d'etre.

radeksimko:
Hi 👋 
I think it's ok to put workaround in place to make things work for "special" regions.

That said given that there are no guarantees/plans about what is/isn't support in any given region I think it would be better if we just silently ignored `UnknownOperationException` error code for `ListTagsLogGroup` + perhaps logged a `DEBUG` message mentioning that we ignored it, just for potential future debugging purposes.

We can keep `(Un)TagLogGroup` as these **should error out** when the user actually does try to define tags in region which doesn't support it.

This is to avoid the need to fix this again and cut release when Amazon starts supporting it in any specific region.

@jmcarp What do you think?

paddycarver:
Hi! This issue has been open for over two weeks and waiting response from the community without any activity, so I'm going to close it out. We do this to try to keep our attention on issues that are actively impacting the community, it's not a comment on the issue itself. If there is still interest in this issue, please feel free to reply, and we'll re-open the issue. Thanks!

realflash:
What sort of response do you want? I said "I would dearly love access to the local data provider so I vote for patch it now, please. Who knows when Amazon will actually deliver this functionality, if they ever do? The number of services in govcloud is about 20% of those in other partitions; being crippled is practically part of its raison d'etre."

realflash:
I'm just waiting for it to be fixed - we're stuck on 0.7.2 and not able to progress our usage until it is. If there's something we can contribute to the decision please say so.

@kevinbrooke

LinuxBozo:
I think the solution proposed by @radeksimko works all around.

grubernaut:
Hey all, I'm going to re-open this and we're going to begin working on a fix for this. 

I agree with @radeksimko's solution, and think it's the best path forward in the long-run. However, this patch will also solve the issue in the immediate term. Either way, we should fix this issue. I apologize for the confusion here, it should not have been closed.

PR #12418, opened 2017-03-03
Add minio remote state backend

jrandall:
Add a remote state backend for minio, using the minio-go client (https://github.com/minio/minio-go). 

In addition to supporting the Minio object store, this backend also supports other S3-compatible object stores that use the V2 AWS auth, which are _not_ supported by the existing s3 backend (which uses aws-sdk-go). 

In particular, it should work with both Ceph Object Store (radosgw) and Riak CS, neither of which works with aws-sdk-go.


jrandall:
I would ideally like to include locking as well, but minio itself can't really support locking because S3 only promises to be eventually consistent. I'd ideally like to have the flexibility to use this `minio` remote state backend to store the state, but use another backend (such as the existing lock implementations in the consul or dynamodb backends, or potentially others that provide locking only) to lock it. Is there a plan to be able to mix-and-match backends in this way? If not, perhaps one option would be to add configuration options to `minio` that specify which backend to use for locking, and have it implement locking methods that just pass-through to the other backend, although that would add quite a bit of complexity and the configuration would need to somehow include all of the config for the locking backend as well. It seems easier to address that in the core, such that the backend configuration can support having one backend for state storage and (potentially) another for locking, such that locking operations are simply delegated to the configured backend.

jrandall:
I see there was some discussion in #11187 between @jbardin and @apparentlymart about splitting up the storage and locking functionality of the backends. Given that there are some storage backends (such as this one) that can't usefully support locking themselves (one might argue the "s3" backend is also one of those, and has just arbitrary been combined with the dynamodb locking backend) I think it makes a lot of sense to split up and separately configure the storage and locking backends.

mytototo:
Any news about this feature? I'd really like to store my Terraform states in Minio so I don't need to run both Consul and Minio.

wingZero21:
Is it possible to get this pulled in. 

It would be great to get the alternatives for S3 as we have applications we would like to add state files for internal only and this would fix this problem without having to use AWS S3 as the backend for the state files.

jrandall:
I believe the minio state backend in this PR does work with the version of terraform it was developed against. I haven't had a chance to update it as we non longer need it since we ended up just keeping state in consul as we also needed locking. I continue to think it would make sense for terraform to separate out state storage and state locking backends so that we could keep state in minio and use consul for locking, but this PR could probably be fairly easily updated to be mergeable as a state storage backend with no locking.

jrandall:
I've updated the PR so it is ready to merge and all tests (including tests on the new minio remote state backend) are passing. Hopefully a maintainer will notice and review.

lfarnell:
@jrandall Can you explain your rationale why this needs to be a separate backend vs replacing the aws-sdk code with the minio sdk?

jrandall:
@lfarnell the main reason is that the AWS backend also includes dynamodb locking, so that backend seems to be intended for people using actual Amazon services, whereas this backend would be for other S3-like services. 

If locking and remote state storage backends were separated, I would no longer have a good reason that the minio-based client should not just replace the aws-sdk based client as the main client for all S3-like remote state backends.

teamterraform:
We are sorry to have let this PR sit for so long. Since this PR targets a part of the code base that is no longer used to manage remote state, we will close this PR. If this is a feature you are still interested in, please open a Github issue using the Feature Request template so we can discuss the design.

Thank you!

PR #12421, opened 2017-03-03
provider/aws: fix aws_route53_zone force_destroy behavior

### LINKED ISSUE #12407: provider/aws: aws_route53_zone with force_destroy fails to delete zone on destroy
### Terraform Version

```
$ terraform version
Terraform v0.8.8
```

### Affected Resource(s)

- aws_route53_zone

```hcl
provider "aws" {
  region = "us-east-1"
}

resource "aws_route53_zone" "public" {
  name          = "myexample.com."
  force_destroy = true
}

output "zone_id" {
  value = "${aws_route53_zone.public.id}"
}
```

### Debug Output

https://gist.github.com/ryane/ba584cdb3023b4f02dfe9cb4dfb2225a

### Expected Behavior

It is my understanding that setting `force_destroy = true` should tell Terraform to delete all records in the zone when it destroys the resource even if it contains records that are not managed by Terraform.

### Actual Behavior

Destroying fails with this error:

```
* aws_route53_zone.public: HostedZoneNotEmpty: The specified hosted zone contains non-required resource record sets  and so cannot be deleted.
        status code: 400, request id: e466d633-0014-11e7-a009-8d1bd2e9a272
```

### Steps to Reproduce

1. Run `terraform apply`. This should complete successfully.
2. Create a record in the zone that is outside of Terraform control. For example:

create.json:

```json
{
  "Comment": "Create an A record",
  "Changes": [
    {
      "Action": "UPSERT",
      "ResourceRecordSet": {
        "Name": "test.myexample.com.",
        "Type": "A",
        "TTL": 300,
        "ResourceRecords": [
          {
            "Value": "8.8.8.8"
          }
        ]
      }
    }
  ]
}
```

```
$ aws route53 change-resource-record-sets \
   --hosted-zone-id $(terraform output zone_id) \
   --change-batch file://create.json
```

3. Run `terraform destroy`

```
$ terraform destroy -force
aws_route53_zone.public: Refreshing state... (ID: Z273D4PE6ATWJH)
aws_route53_zone.public: Destroying...
Error applying plan:

1 error(s) occurred:

* aws_route53_zone.public: HostedZoneNotEmpty: The specified hosted zone contains non-required resource record sets  and so cannot be deleted.
        status code: 400, request id: e466d633-0014-11e7-a009-8d1bd2e9a272

```

ryane:
The conditional to ignore the deletion of NS and SOA records can fail to
match if the hostedZoneName already ends with a ".". When that happens,
terraform tries to delete those records which is not supported by AWS
and results in a 400 bad request. This fixes the conditional so that it
will work whether or not hostedZoneName ends with a ".".

fixes #12407

### DIFF
--- a/builtin/providers/aws/resource_aws_route53_zone.go
+++ b/builtin/providers/aws/resource_aws_route53_zone.go
@@ -300,7 +300,7 @@ func deleteAllRecordsInHostedZoneId(hostedZoneId, hostedZoneName string, conn *r
 		changes := make([]*route53.Change, 0)
 		// 100 items per page returned by default
 		for _, set := range sets {
-			if *set.Name == hostedZoneName+"." && (*set.Type == "NS" || *set.Type == "SOA") {
+			if strings.TrimSuffix(*set.Name, ".") == strings.TrimSuffix(hostedZoneName, ".") && (*set.Type == "NS" || *set.Type == "SOA") {
 				// Zone NS & SOA records cannot be deleted
 				continue
 			}
--- a/builtin/providers/aws/resource_aws_route53_zone_test.go
+++ b/builtin/providers/aws/resource_aws_route53_zone_test.go
@@ -414,7 +414,7 @@ resource "aws_route53_zone" "main" {
 
 const testAccRoute53ZoneConfig_forceDestroy = `
 resource "aws_route53_zone" "destroyable" {
-	name = "terraform.io"
+	name = "terraform.io."
 	force_destroy = true
 }
 `


radeksimko:
Hi @ryane 
this is useful bugfix, thanks.

I appreciate the test modification to cover the fixed behaviour. It would be better if we covered both cases though - with and without trailing dot, to make sure we don't introduce a similar bug in the future that would break zones w/out dot.

I will send a separate patch to reflect this.

-----

FYI: For future contributions it usually helps both the reviewer and yourself (I think 😃 ) when the PR is raised from a separate branch - not `master`. It's much easier to make any necessary changes without having to force-push into `master` and it allows you to raise multiple different PRs in isolation. 😉

PR #12424, opened 2017-03-03
Fix error message in Azure state backend

ashb:
The error led me to try adding `resource_group` but the code wanted `resource_group_name`. This fixes the error message to match the code.


stack72:
LGTM Thanks for the work here :)

Issue #11806, opened 2017-02-09
`depends_on` always triggers data source read


queeno:
Hi there,

### Terraform Version
Terraform v0.8.4

### Affected Resource(s)
- data.template_file

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
data "template_file" "hello" {
    template = "template"
    depends_on = ["null_resource.hello"]
}

resource "null_resource" "hello" {}
```

### Debug Output
https://gist.github.com/queeno/f198c93760f7c60e5102e7fd5873ad1f

### Expected Behavior
`terraform plan` shouldn't re-read the data source.

### Actual Behavior
`terraform plan` re-reads the data source.

```
❯ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but
will not be persisted to local or remote state storage.

null_resource.hello: Refreshing state... (ID: 3983604783330107355)

The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed. Cyan entries are data sources to be read.

Note: You didn't specify an "-out" parameter to save this plan, so when
"apply" is called, Terraform can't guarantee this is what will execute.

<= data.template_file.hello
    rendered: "<computed>"
    template: "template"


Plan: 0 to add, 0 to change, 0 to destroy.
```

### Steps to Reproduce
`terraform plan`

### Important Factoids
When using `depends_on` in `template_file`, `terraform plan` always seems to re-read the data source. If the data source is used by an instance's `user-data`, terraform plans to change the instance's user-data. `terraform apply`, however, doesn't produce any change.

If `depends_on` is not used, then the data source is not re-read.

apparentlymart:
Hi @queeno! Thanks for this issue.

I seem to remember this being done intentionally to fix a bug a while ago. `depends_on` presents a rather-awkward situation for data sources because they are defined as being refreshed early when they have no computed data, but `depends_on` means Terraform can't tell what they might be depending on.

There is possibly something more refined we could do here to make this support more cases, but unfortunately I think right now this is working as designed and we don't really have a better strategy in mind.

queeno:
Hi @apparentlymart 

Thanks for your very quick response!

[...]

PR #11081, opened 2017-01-06
command/plan: add flag to display diffs only

gechr:
Adds an optional new `-diff` flag to `terraform plan` to view only attributes which have changed. Useful when you're making a quick change and don't want to trawl through 100s of lines to pinpoint what's actually changing.

### Before
```
› terraform plan
[...]
~ aws_security_group.main
    ingress.#:                            "2" => "3"
    ingress.1969601585.cidr_blocks.#:     "0" => "1"
    ingress.1969601585.cidr_blocks.0:     "" => "0.0.0.0/0"
    ingress.1969601585.from_port:         "0" => "13714"
    ingress.1969601585.protocol:          "" => "udp"
    ingress.1969601585.security_groups.#: "0" => "0"
    ingress.1969601585.self:              "false" => "false"
    ingress.1969601585.to_port:           "0" => "13714"
    ingress.1971169986.cidr_blocks.#:     "1" => "1"
    ingress.1971169986.cidr_blocks.0:     "124.29.202.146/32" => "124.29.202.146/32"
    ingress.1971169986.from_port:         "443" => "443"
    ingress.1971169986.protocol:          "tcp" => "tcp"
    ingress.1971169986.security_groups.#: "0" => "0"
    ingress.1971169986.self:              "false" => "false"
    ingress.1971169986.to_port:           "443" => "443"
    ingress.3256548365.cidr_blocks.#:     "1" => "1"
    ingress.3256548365.cidr_blocks.0:     "124.29.202.146/32" => "124.29.202.146/32"
    ingress.3256548365.from_port:         "22" => "22"
    ingress.3256548365.protocol:          "tcp" => "tcp"
    ingress.3256548365.security_groups.#: "0" => "0"
    ingress.3256548365.self:              "false" => "false"
    ingress.3256548365.to_port:           "22" => "22"


Plan: 0 to add, 1 to change, 0 to destroy.
```

### After
```george • mac • /tmp/tmp.NGdGnpbrW0 • master
› terraform plan -diff
[...]
~ aws_security_group.main
    ingress.#:                            "2" => "3"
    ingress.1969601585.cidr_blocks.#:     "0" => "1"
    ingress.1969601585.cidr_blocks.0:     "" => "0.0.0.0/0"
    ingress.1969601585.from_port:         "0" => "13714"
    ingress.1969601585.protocol:          "" => "udp"
    ingress.1969601585.to_port:           "0" => "13714"


Plan: 0 to add, 1 to change, 0 to destroy.
```

PR #12432, opened 2017-03-03
Don't display plan diffs with only data reads

jbardin:
If the only action in a plan is to read data, possibly because the data
source depends on a value to be computed, we don't need to display that
in the plan output.

The add change and destroy counts must also be 0 to supress output.
These should match the plan diff, but check them as an added failsafe.

Possible fix for issues like #11806?

### DIFF
--- a/backend/local/backend_plan.go
+++ b/backend/local/backend_plan.go
@@ -102,8 +102,15 @@ func (b *Local) opPlan(
 		return
 	}
 
-	// Record state
-	runningOp.PlanEmpty = plan.Diff.Empty()
+	// the hook counts and EmptyExceptData have to match in order to hide
+	// diff output.
+	anyChanges := (countHook.ToAdd +
+		countHook.ToRemoveAndAdd +
+		countHook.ToChange +
+		countHook.ToRemove +
+		countHook.ToRemoveAndAdd)
+
+	runningOp.PlanEmpty = plan.Diff.EmptyExceptData() && anyChanges == 0
 
 	// Save the plan to disk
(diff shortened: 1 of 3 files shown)


josephholsten:
We're assuming users are comfortable just using `terraform state show` for seeing what data has changed? Seems fine by me.

apparentlymart:
@jbardin the usual symptom of the problem is that the data resource is used in a value for a `ForceNew` attribute and so Terraform believes it needs to replace the resource on each run.

This change wouldn't address that since it is just a cosmetic update to the UI.

The main UI issue I've noted here is that users don't generally understand that when a data read appears in a plan this means that there is some dependency that needs to be resolved. It would perhaps be helpful to try to describe the reason why a data source read is deferred to the apply phase so that users can understand what they need to do in order to avoid that; right now we give no feedback at all that a delayed read is an unusual case that can be avoided.

jbardin:
@apparentlymart,

Yes, this was just a cosmetic update (also changing the return code to go along with empty output). Maybe there is some way we can better document _"are you sure you want `depends_on` here?"_. 

This wasn't intended to replace a possible fix where we can determine that there is actually no change.

apparentlymart:
I think having a note under the diff when it is caused by `depends_on` could be a good temp fix here, but unfortunately I don't think the diff renderer has enough info to do that right now. :(

gechr:
Though not directly related to this PR, any chance we can bundle diff-related PR https://github.com/hashicorp/terraform/pull/11081 into whatever release this makes it into?

PR #12434, opened 2017-03-03
provider/google: add location to storage tests.

paddyforan:
Add location to storage tests that need it, which fixes the failing
TestAccStorageStorageClass test.

Test before patch:

```
=== RUN   TestAccStorageStorageClass
Error creating bucket tf-test-acc-bucket-1345402779410348481: googleapi: Error 400: The combination of locationConstraint and storageClass you provided is not supported for your project, invalid--- FAIL: TestAccStorageStorageClass (5.18s)
        testing.go:265: Step 2 error: Error applying: 1 error(s) occurred:

                * google_storage_bucket.bucket: 1 error(s) occurred:

                * google_storage_bucket.bucket: googleapi: Error 400: The combination of locationConstraint and storageClass you provided is not supported for your project, invalid
FAIL
```

Test after patch:

```
=== RUN   TestAccStorageStorageClass
--- PASS: TestAccStorageStorageClass (7.72s)
PASS
```

### DIFF
--- a/builtin/providers/google/resource_storage_bucket_test.go
+++ b/builtin/providers/google/resource_storage_bucket_test.go
@@ -68,7 +68,7 @@ func TestAccStorageStorageClass(t *testing.T) {
 		CheckDestroy: testAccGoogleStorageDestroy,
 		Steps: []resource.TestStep{
 			{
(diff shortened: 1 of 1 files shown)


paddyforan:
Thanks for the review! Agree string bits are kinda gross... but I wanted to make sure a config without the location attribute specified at all would still work, and couldn't think of a better way to do it. :)

PR #12444, opened 2017-03-04
Rename fields wrongly named in docs

pvcnt:
There is not azurerm_container_service.linux_profile.ssh_keys argument, as stated in the examples, but an ssh_key field. Arguments reference was correct.


stack72:

pvcnt:
Thanks!

PR #12445, opened 2017-03-05
document update: terraform get -u flag doesn't work

brettski:
The -u flag didn't work for me and based on documentation should be -update, which did work.  Makes getting started docs more accurate.


stack72:
LGTM! Thanks for this :)

PR #12476, opened 2017-03-06
Don't reuse the ResourceProvider between tests

jbardin:
[DO NOT MERGE]

This is testing out an idea that requires a change to helper/schema, which should be done through another PR.

The arm provider was configured once in the test init, but if the StopContext was ever canceled any further tests would fail.  Using a ResourceProviderFactory ensures that the arm provider is re-configured between tests.


jbardin:
closing this for now to implement the framework for a fix in the helper packages.

PR #12480, opened 2017-03-06
Website: Add Mega Nav

jasoncostello:
🚧 WIP. Don't merge until final middleman-hashicorp update

HashiCorp is adding a high-level element to provide navigation and visibility to OSS project sites. We like to call it Mega Nav :zap:. Mega Nav is pulled in from [middleman-hashicorp](https://github.com/hashicorp/middleman-hashicorp) and is positioned above the local nav in the site. It looks like this: 

### Desktop Inactive
screen shot 2017-03-06 at 2 09 12 pm

### Desktop Active
screen shot 2017-03-06 at 1 57 46 pm

### Mobile Inactive
screen shot 2017-03-06 at 2 16 59 pm

### Mobile Active
screen shot 2017-03-06 at 2 17 13 pm

note: There is one small section of `.mega-nav` css overrides in `global.scss`. I'll follow up in another PR after changes are made globally to the mega nav component in middleman-hashicorp to remove when they are no longer needed.

cc/ @pearkes @hashicorp/design


website/Gemfile
```diff
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "middleman-hashicorp", "0.3.4"
+gem "middleman-hashicorp", "0.3.11"
```
sethvargo: @jasoncostello can you use 0.3.12 here please? Also, please update `packer.json` to use 0.3.12.

sethvargo: Oh, also `Makefile` too 😦 

Sorry - I wish we could unify it a bit more.

jasoncostello:
@sethvargo yep, I'll run through and make sure all OSS sites ref 0.3.12. looping back through asap

PR #12483, opened 2017-03-06
add "name" to exported attributes

mumblez:
missing but a valid exported attribute


stack72:
LGTM thanks for the addition here @mumblez

Issue #4336, opened 2015-12-16
Error creating backend service


yosssi:
I got the following error when I try to create a backend service:

```
google_compute_backend_service.my-backend-service: Creating...
  backend.#:                               "" => "1"
  backend.499478644.balancing_mode:        "" => "UTILIZATION"
  backend.499478644.capacity_scaler:       "" => "1"
  backend.499478644.description:           "" => ""
  backend.499478644.group:                 "" => "https://www.googleapis.com/replicapool/v1beta2/projects/myproject-dev/zones/asia-east1-a/instanceGroupManagers/gke-myproject-api-27da66d4-group"
  backend.499478644.max_rate:              "" => ""
  backend.499478644.max_rate_per_instance: "" => ""
  backend.499478644.max_utilization:       "" => "0.8"
  description:                             "" => "xxxx"
  fingerprint:                             "" => "<computed>"
  health_checks.#:                         "" => "1"
  health_checks.2686024285:                "" => "https://www.googleapis.com/compute/v1/projects/myproject-dev/global/httpHealthChecks/myi-http-basic-check"
  name:                                    "" => "my-backend-service"
  port_name:                               "" => "http"
  protocol:                                "" => "HTTP"
  self_link:                               "" => "<computed>"
  timeout_sec:                             "" => "<computed>"

Error applying plan:

1 error(s) occurred:

* google_compute_backend_service.my-backend-service: Error creating backend service: googleapi: Error 400: Invalid value for field 'resource.backends[0]': ''.  Invalid group url, invalid

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

Following is my template:

```
resource "google_compute_backend_service" "my-backend-service" {
    name = "my-backend-service"
    description = "xxxx"
    port_name = "http"
    protocol = "HTTP"
    backend {
      group = "${element(google_container_cluster.my_cluster.instance_group_urls, 1)}"
    }
    health_checks = ["${google_compute_http_health_check.my-http-basic-check.self_link}"]
}
```

I don't know why this kind of error occurs and how to fix it.

What is the reason of this error and how can I fix it?

Thanks.

phinze:
I'm not familiar offhand with this API, but perhaps @lwander can shed some light on this?

lwander:
[...]

PR #12486, opened 2017-03-07
provider/google: fix container instance group URLs

paddyforan:
Google Container Engine's cluster API returned instance group manager
URLs when it meant to return instance group URLs. See #4336 for details
about the bug.

While this is undeniably an upstream problem, this PR:

* detects the error, meaning it will work as expected when the API is
  fixed.
* corrects the error by requesting the instance group manager, then
  retrieving its instance group URL, and using that instead.
* adds a test that exercises the error and the solution, to ensure it is
  functioning properly.

Unit test without fix:

```
=== RUN   TestAccContainerCluster_backend
--- FAIL: TestAccContainerCluster_backend (303.57s)

                * google_compute_backend_service.my-backend-service: 1 error(s) occurred:

                * google_compute_backend_service.my-backend-service: Error creating backend service: googleapi: Error 400: Invalid value for field 'resource.backends[0].group': 'https://www.googleapis.com/compute/v1/projects/hc-terraform-testing/zones/us-central1-c/instanceGroupManagers/gke-terraform-test-evnwj-default-pool-68b6a1a7-grp'. The value is not a valid resourceView nor instanceGroup URL., invalid
FAIL
```

Unit test with fix:

```
=== RUN   TestAccContainerCluster_backend
--- PASS: TestAccContainerCluster_backend (298.37s)
```

### DIFF
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -13,6 +13,10 @@ import (
 	"google.golang.org/api/googleapi"
 )
 
+var (
+	instanceGroupManagerURL = regexp.MustCompile("^https://www.googleapis.com/compute/v1/projects/([a-z][a-z0-9-]{5}(?:[-a-z0-9]{0,23}[a-z0-9])?)/zones/([a-z0-9-]*)/instanceGroupManagers/([^/]*)")
+)
+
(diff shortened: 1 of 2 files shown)


paddyforan:
Going to go ahead and merge this one now. :)

PR #12498, opened 2017-03-07
Add schema.Provider.TestReset to reset StopContext between tests

jbardin:
In normal operation, a provider is only created once and only ever canceled once, but in testing a ResourceProvider may be initialized and reused throughout multiple tests. 

This poses a problem when `Stop()` has been called on a provider and the StopContext has been used by the provider. 

This adds a `TestReset` method to `schema.Provider` to reset its internal state, currently only consisting of the internal Context. `Provider.TestReset` will then call Provider.MetaReset if it's not nil, which can be used to reset any state stored in the meta interface value which is especially important if it stored a StopContext value. 

The TestReset method will be automatically called after PreCheck in resource.Test.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -22,6 +22,12 @@ import (
 
 const TestEnvVar = "TF_ACC"
 
+// TestProvider is implemented by scheme.Provider to allow resetting the
+// provider state between tests.
+type TestProvider interface {
+	TestReset() error
+}
+
 // TestCheckFunc is the callback type used with acceptance tests to check
 // the state of a resource. The state passed in is the latest state known,
 // or in the case of being after a destroy, it is the last known state when
@@ -216,13 +222,9 @@ func Test(t TestT, c TestCase) {
 		c.PreCheck()
 	}
 
-	// Build our context options that we can
-	ctxProviders := c.ProviderFactories
-	if ctxProviders == nil {
-		ctxProviders = make(map[string]terraform.ResourceProviderFactory)
-		for k, p := range c.Providers {
-			ctxProviders[k] = terraform.ResourceProviderFactoryFixed(p)
-		}
+	ctxProviders, err := testProviderFactories(c)
+	if err != nil {
+		t.Fatal(err)
 	}
 	opts := terraform.ContextOpts{Providers: ctxProviders}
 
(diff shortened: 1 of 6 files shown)


jbardin:
This is still a little awkward for provider, going to push an updated version...

stack72:
Hi @jbardin 

I am still getting the errors using this branch...

```
% make testacc TEST=./builtin/providers/azurerm TESTARGS='-run=TestAccAzureRMStorageAccount_'
2017/03/08 14:41:26 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMStorageAccount_ -timeout 120m
=== RUN   TestAccAzureRMStorageAccount_importBasic
--- FAIL: TestAccAzureRMStorageAccount_importBasic (68.12s)
	testing.go:267: Step 0 error: Error applying: 1 error(s) occurred:

		* azurerm_storage_account.testsa: 1 error(s) occurred:

		* azurerm_storage_account.testsa: Error creating Azure Storage Account 'unlikely23exst2accttnew': storage.AccountsClient#Create: Failure sending request: StatusCode=0 -- Original Error: Put https://management.azure.com/subscriptions/c0a607b2-6372-4ef3-abdb-dbe52a7b56ba/resourceGroups/testAccAzureRMSA-5659809402542612052/providers/Microsoft.Storage/storageAccounts/unlikely23exst2accttnew?api-version=2016-01-01: net/http: request canceled
=== RUN   TestAccAzureRMStorageAccount_basic
--- FAIL: TestAccAzureRMStorageAccount_basic (75.23s)
	testing.go:267: Step 0 error: Error applying: 1 error(s) occurred:

		* azurerm_storage_account.testsa: 1 error(s) occurred:

		* azurerm_storage_account.testsa: Error creating Azure Storage Account 'unlikely23exst2acct9sn2': storage.AccountsClient#Create: Failure sending request: StatusCode=0 -- Original Error: Put https://management.azure.com/subscriptions/c0a607b2-6372-4ef3-abdb-dbe52a7b56ba/resourceGroups/testAccAzureRMSA-5846784653319619894/providers/Microsoft.Storage/storageAccounts/unlikely23exst2acct9sn2?api-version=2016-01-01: net/http: request canceled
````

**Comment by jbardin** [comment]
@mitchellh: that last commit is the important one, and effects all uses of StopContext, not just in tests. Azure had problems simply because it was the first to make use of the new StopContext.

File: helper/resource/testing.go [hunk 1]
```diff
@@ -22,6 +22,12 @@ import (
 
 const TestEnvVar = "TF_ACC"
 
+// TestProvider is implemented by scheme.Provider to allow resetting the
+// provider state between tests.
```
mitchellh: Nitpick but I'd reword the comment to say that this can be implemented by any ResourceProvider to have custom reset functionality, and note when TestReset is called. You can also make a note that helper/schema implements this.

helper/resource/testing.go
```diff
@@ -333,6 +335,43 @@ func Test(t TestT, c TestCase) {
...
+		}
+	}
+
+	// reset the providers if needed
+	for k, pf := range ctxProviders {
+		// we can ignore any errors here, if we don't have a provider to reset
+		// the error will be handled later
+		p, _ := pf()
+		if p, ok := p.(TestProvider); ok {
+			err := p.TestReset()
```
mitchellh: Do we have to do this for all providers or can we just do it as part of the constructed factory above (line 355)?

terraform/resource_provider_mock.go
```diff
@@ -56,6 +56,8 @@ type MockResourceProvider struct {
 	ReadDataDiffFn                 func(*InstanceInfo, *ResourceConfig) (*InstanceDiff, error)
 	ReadDataDiffReturn             *InstanceDiff
 	ReadDataDiffReturnError        error
+	TestResetCalled                bool
+	TestResetError                 error
```
mitchellh: Seems messy to add this interface here since the core doesn't care about it all. Could we modify the test in helper/resource to add it to a custom provider to verify it is called?

helper/resource/testing.go
jbardin: ah yes, I often forget that schema.Provider technically isn't the only ResourceProvider ;)

helper/resource/testing.go
jbardin: we could, but we need to range over both `c.Providers` and `c.ProviderFactories` above, and here I do the check once, rather than repeat it in both loops.

terraform/resource_provider_mock.go
jbardin: Makes sense. Removed the fields from terraform, and wrapped the mock provider with a TestProvider implementation.

helper/resource/testing.go
mitchellh: Ah I see. Yeah it is probably fine anyways. :+1:

PR #12507, opened 2017-03-07
nested lists in variables panic during validation

### LINKED ISSUE #12450: crash when variable default value is nested list
### Terraform Version
```shell
$ terraform --version
Terraform v0.8.8
```

### Affected Resource(s)
- variable

```hcl
variable "nested_list_as_variable" {
    default = [["nested_list_crashes_things"]]
}
```

### Debug Output
https://gist.github.com/ivanlei/d4542e0da239aa3a757d6a47806b19a0

### Panic Output
https://gist.github.com/ivanlei/a632a61276a4f0dae729e16e0341622e

### Expected Behavior
terraform parses cleanly or reports error

### Actual Behavior
crash

### Steps to Reproduce
1. `terraform apply`

jbardin:
Verify that we have enough containers in the stack to look for a map in
replaceCurrent.

This won't guarantee that nested containers work in all cases (they don't), but we shouldn't panic when validating the config. 

Fixes #12450

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -201,6 +201,12 @@ func TestConfigValidate_table(t *testing.T) {
 			true,
 			"cannot contain interp",
 		},
+		{
+			"nested types in variable default",
+			"validate-var-nested",
+			true,
+			"ERROR",
+		},
 	}
 
 	for i, tc := range cases {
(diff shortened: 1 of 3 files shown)


mitchellh:
Oh weird.

Issue #12510, opened 2017-03-07
Terraform crashes during azurerm_container_service provisioning


maticko:
### Terraform Version
0.8.8

### Affected Resource(s)
- azurerm_container_service

### Terraform Configuration Files
```hcl
resource "azurerm_container_service" "kubernetes" {
  name = "${var.env}-kubernetes"
  resource_group_name = "${var.location}"
  location = "${var.location}"
  orchestration_platform = "Kubernetes"

  master_profile {
    count = 1
    dns_prefix = "${var.env}master${count.index}"
  }

  linux_profile {
    admin_username = "kuberentes"

    ssh_key {
      key_data = "ssh-rsa key"
    }
  }

  agent_pool_profile {
    name = "default"
    count = 1
    dns_prefix = "${var.env}agent${count.index}"
    vm_size = "Standard_A0"
  }

  service_principal {
    client_id = "${var.client_id}"
    client_secret = "${var.client_secret}"
  }

  diagnostics_profile {
    enabled = false
  }

  tags {
    environment = "${var.env}"
  }
}
```

### Debug Output
regular debug output contains lots of secrets. let me know if the panic output is not enough.  
https://gist.github.com/maticko/a470bb15e71d5eb7f6c032251e0a702e

### Panic Output
https://gist.github.com/maticko/a470bb15e71d5eb7f6c032251e0a702e

### Expected Behavior
Resources should be created and terraform should store them in state file

### Actual Behavior
application crashed during provisioning and didn't save anything to a state file. Resources were created in the account. 

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Nothing specific, standard azure account

brandontosch:
Looks like a simple pointer referencing problem. I'll try and get a PR open for the fix shortly.

stack72:
Closed via #12516

PR #12516, opened 2017-03-08
Bug Fix: Terraform crashes during azurerm_container_service provisioning

brandontosch:
Fixed pointer to pointer referencing issue causing crash while provisioning azurerm_container_service  (#12510).

Also updated tests to use East US region since resources required for tests were not available in West US. All container service tests are now passing.


stack72:
LGTM thanks for this :)

brandontosch:
Sure, my pleasure :)

PR #12532, opened 2017-03-08
provider/azurerm: Add support for managed availability sets.

rrudduck:
This PR enables specifying an availability set as being managed. An availability set must be managed in order to have VMs with managed disks as members.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_availability_set.go
+++ b/builtin/providers/azurerm/resource_arm_availability_set.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/arm/compute"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -65,6 +66,13 @@ func resourceArmAvailabilitySet() *schema.Resource {
 				},
 			},
 
+			"managed": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
@@ -82,6 +90,7 @@ func resourceArmAvailabilitySetCreate(d *schema.ResourceData, meta interface{}) 
 	updateDomainCount := d.Get("platform_update_domain_count").(int)
 	faultDomainCount := d.Get("platform_fault_domain_count").(int)
 	tags := d.Get("tags").(map[string]interface{})
+	managed := d.Get("managed").(bool)
 
 	availSet := compute.AvailabilitySet{
 		Name:     &name,
@@ -91,6 +100,13 @@ func resourceArmAvailabilitySetCreate(d *schema.ResourceData, meta interface{}) 
 			PlatformUpdateDomainCount: azure.Int32(int32(updateDomainCount)),
 		},
 		Tags: expandTags(tags),
+	}
+
+	if managed == true {
+		n := "Aligned"
+		availSet.Sku = &compute.Sku{
+			Name: &n,
+		}
 	}
 
 	resp, err := availSetClient.CreateOrUpdate(resGroup, name, availSet)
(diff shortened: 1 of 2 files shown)


builtin/providers/azurerm/resource_arm_availability_set.go
```diff
@@ -93,6 +102,13 @@ func resourceArmAvailabilitySetCreate(d *schema.ResourceData, meta interface{})
 		Tags: expandTags(tags),
 	}
 
+	if managed == true {
+		n := "Aligned"
```
pmcatominey: Could you clarify this behaviour? The normal approach is to expose the SKU options on on the resource schema as special behaviour can put us in tricky situations with future API updates.

builtin/providers/azurerm/resource_arm_availability_set_test.go
```diff
@@ -259,3 +280,18 @@ resource "azurerm_availability_set" "test" {
...
+    name = "acctestRG-%d"
+    location = "West US"
+}
+resource "azurerm_availability_set" "test" {
+    name = "acctestavset-%d"
+    location = "West US"
+    resource_group_name = "${azurerm_resource_group.test.name}"
+    platform_update_domain_count = 10
+    platform_fault_domain_count = 1
+	managed = true
```
pmcatominey: Nitpick: formatting

stack72:
Tests pass @pmcatominey :)

```
% make testacc TEST=./builtin/providers/azurerm TESTARGS='-run=TestAccAzureRMAvailabilitySet_'
2017/03/08 20:06:49 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMAvailabilitySet_ -timeout 120m
=== RUN   TestAccAzureRMAvailabilitySet_importBasic
--- PASS: TestAccAzureRMAvailabilitySet_importBasic (70.50s)
=== RUN   TestAccAzureRMAvailabilitySet_basic
--- PASS: TestAccAzureRMAvailabilitySet_basic (68.69s)
=== RUN   TestAccAzureRMAvailabilitySet_disappears
--- PASS: TestAccAzureRMAvailabilitySet_disappears (67.36s)
=== RUN   TestAccAzureRMAvailabilitySet_withTags
--- PASS: TestAccAzureRMAvailabilitySet_withTags (63.83s)
=== RUN   TestAccAzureRMAvailabilitySet_withDomainCounts
--- PASS: TestAccAzureRMAvailabilitySet_withDomainCounts (66.06s)
=== RUN   TestAccAzureRMAvailabilitySet_managed
--- PASS: TestAccAzureRMAvailabilitySet_managed (68.72s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/azurerm	405.220s
```

builtin/providers/azurerm/resource_arm_availability_set.go
rrudduck: According to the documentation here: https://docs.microsoft.com/en-us/rest/api/compute/availabilitysets/availabilitysets-create it's not a full sku object. It's just sku.name that is used.

sku.name	Specifies whether the availability set is managed or not. Possible values are: Aligned or Classic. An Aligned availability set is managed, Classic is not.

I thought the easier change was to specify managed or not as a boolean since that is how it's represented in the portal. Additionally in the sdk (https://github.com/Azure/azure-sdk-for-go/blob/master/arm/compute/models.go#L415) there is a property for "managed" but it does not appear in the REST docs and does not appear to be used when I tested it.

rrudduck: I should also not that there did not appear to be a constant for that value defined in the SDK either.

https://github.com/Azure/azure-sdk-for-go/search?q=Aligned&type=Code&utf8=%E2%9C%93

rrudduck:
Formatting should be resolved.

builtin/providers/azurerm/resource_arm_availability_set.go
pmcatominey: Thanks for the clarification!

PR #12533, opened 2017-03-08
helper/schema: Rename Timeout resource block to Timeouts

catsby:
Pluralize configuration argument name to better represent that there is
one block for many timeouts. Also adds a `timeouts` block into the `db_instance` basic test, just to have them present in an acceptance test.

Example:
```
resource "example" "name" {
  name = "thing"

  timeout {
    create = "10m"
    update = "5m"
  }
}
```

Becomes:

```
resource "example" "name" {
  name = "thing"

  timeouts {
    create = "10m"
    update = "5m"
  }
}
```

This is a follow up change to the following:

- https://github.com/hashicorp/terraform/commit/aa3677cd896809718f26d066216ded14673a7961
- https://github.com/hashicorp/terraform/pull/12503

cc @mitchellh 

------

This is intended for `v0.9` only, as Timeouts were not back ported to `v0.8.x` series

### DIFF
--- a/builtin/providers/aws/resource_aws_db_instance_test.go
+++ b/builtin/providers/aws/resource_aws_db_instance_test.go
@@ -622,6 +622,10 @@ resource "aws_db_instance" "bar" {
 	backup_retention_period = 0
 
 	parameter_group_name = "default.mysql5.6"
+
+	timeouts {
+		create = "30m"
+	}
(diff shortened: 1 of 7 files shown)


mitchellh:
May want to consider making the key a constant in the file. :)

PR #12537, opened 2017-03-08
Add coalescelist interpolation function

Pryz:
This PR add a `coalescelist` function in order to avoid having things like : 

```
${split(",", length(var.subnets) != 0 ? join(",",var.subnets) : join(",",module.foo.subnets))}
```

We will do : 

```
${coalescelist(var.subnets, module.foo.subnets)}
```


Pryz:
Any change of having this PR merged ? :)

Pryz:
Thanks !

PR #12538, opened 2017-03-08
fix bool vs Compatibool ref used in flattenBackends

jonoden:
The `gofastly.Compatibool` reference being used in `flattenBackends` was causing a type mismatch, as `gofastly.Backend` uses bool (It's `UpdateBackendInput` and `CreateBackendInput` that use `gofastly.Compatibool`).

Problem during importing a backend:
```
2017/03/07 20:56:50 [DEBUG] plugin: terraform-provider-fastly: 2017/03/07 20:56:50 [WARN] Error setting Backends for (xxxxxxxxxxxxxxxxxxxx): backend.0.ssl_check_cert: '' expected type 'bool', got unconvertible type '*fastly.Compatibool'
```

struct used in `flattenBackends(backendList []*gofastly.Backend) []map[string]interface{} `
https://github.com/hashicorp/terraform/blob/master/vendor/github.com/sethvargo/go-fastly/backend.go#L9

Test run:
``` 
$ make testacc TEST=./builtin/providers/fastly TESTARGS='-run=TestResourceFastlyFlattenBackend'
2017/03/08 15:53:32 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/fastly -v -run=TestResourceFastlyFlattenBackend -timeout 120m
=== RUN   TestResourceFastlyFlattenBackend
--- PASS: TestResourceFastlyFlattenBackend (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/fastly	0.019s
```


catsby:
Thank you @jonoden ! Sorry for the delay here, I had trouble reproducing at first but the problem was on my end. I've confirmed this patch is a fix, thanks again!

Issue #12506, opened 2017-03-07
aws_dms_replication_task creation causing crash


rgravlin-carecloud:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
~$ terraform -v
Terraform v0.8.8

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_dms_replication_task

### Terraform Configuration Files
```hcl
resource "aws_dms_replication_subnet_group" "test-dms-replication-subnet-group-tf" {
  replication_subnet_group_description = "Replication subnet group"
  replication_subnet_group_id = "test-dms-replication-subnet-group-tf"
  subnet_ids = [
    "removed",
    "removed",
    "removed"
  ]
}

# Create replication instance
resource "aws_dms_replication_instance" "test-dms-replication-instance-tf" {
  allocated_storage = 50
  apply_immediately = true
  auto_minor_version_upgrade = false
  availability_zone = "us-east-1a"
  engine_version = "1.9.0"
  multi_az = false
  preferred_maintenance_window = "sun:10:30-sun:14:30"
  publicly_accessible = false
  replication_instance_class = "dms.t2.micro"
  replication_instance_id = "test-dms-replication-instance-tf"
  replication_subnet_group_id = "${aws_dms_replication_subnet_group.test-dms-replication-subnet-group-tf.id}"
  tags {
    Name = "DMS_replication"
  }
  vpc_security_group_ids = [
    "removed"
  ]
}

# Create DMS Endpoint
resource "aws_dms_endpoint" "test-dms-source-endpoint-tf" {
  database_name = "database_dev"
  endpoint_id = "test-dms-source-endpoint-tf"
  endpoint_type = "source"
  engine_name = "postgres"
  extra_connection_attributes = ""
  password = "password"
  port = 5432
  server_name = "1.2.3.4"
  ssl_mode = "require"
  tags {
    Name = "DMS_Source"
  }
  username = "postgres"
}

# Create DMS Endpoint
resource "aws_dms_endpoint" "test-dms-target-endpoint-tf" {
  database_name = "database_dev"
  endpoint_id = "test-dms-target-endpoint-tf"
  endpoint_type = "target"
  engine_name = "postgres"
  extra_connection_attributes = ""
  password = "password"
  port = 5432
  server_name = "1.2.3.4"
  ssl_mode = "require"
[...]

PR #12539, opened 2017-03-08
provider/aws: Prevent aws_dms_replication_task panic

stack72:
Fixes: #12506

When a replication_task cdc_start_time was specified as an int, it was
causing a panic as the conversion to a Unix timestampe was expecting a
string

It's important to know that the schema was incorrect here - the strip was required for the Parse of the time not an int

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAwsDmsReplicationTaskBasic'
2017/03/08 22:55:29 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAwsDmsReplicationTaskBasic -timeout 120m
=== RUN   TestAccAwsDmsReplicationTaskBasic
--- PASS: TestAccAwsDmsReplicationTaskBasic (1089.77s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	1089.802s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_dms_replication_task.go
+++ b/builtin/providers/aws/resource_aws_dms_replication_task.go
@@ -27,7 +27,7 @@ func resourceAwsDmsReplicationTask() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"cdc_start_time": {
-				Type:     schema.TypeInt,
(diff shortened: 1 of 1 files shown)


stack72:
<3 @grubernaut

Issue #11266, opened 2017-01-18
Support IAM roles for cloudformation stacks


telepath:
We would like to switch to terraform, but we still have many cloudformation templates. We also have a role concept, so we have a CloudFormation role that is allowed to create the stacks, and the stack is run with a specific role given in the stack configuration.
Currently, terraform cannot configure cloudformation templates to run as a specific role.

### Terraform Version
Terraform v0.8.4

### Affected Resource(s)
- aws_cloudformation_stack

### Expected Behavior
I can add the role_arn parameter to the cloudformation stacks terraform config, and terraform creates the stack with the given IAM role.

### Actual Behavior
role_arn is not supported by terraform.

PR #12547, opened 2017-03-08
provider/aws: Add support to set iam_role_arn on cloudformation Stack

stack72:
Fixes: #11266

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCloudFormation_basic'
2017/03/09 01:39:16 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCloudFormation_basic -timeout 120m
=== RUN   TestAccAWSCloudFormation_basic
--- PASS: TestAccAWSCloudFormation_basic (89.38s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	89.416s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudformation_stack.go
+++ b/builtin/providers/aws/resource_aws_cloudformation_stack.go
@@ -22,12 +22,12 @@ func resourceAwsCloudFormationStack() *schema.Resource {
 		Delete: resourceAwsCloudFormationStackDelete,
 
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-			"template_body": &schema.Schema{
+			"template_body": {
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/resource_aws_cloudformation_stack.go
```diff
@@ -385,6 +393,10 @@ func resourceAwsCloudFormationStackUpdate(d *schema.ResourceData, meta interface
 		input.StackPolicyURL = aws.String(d.Get("policy_url").(string))
 	}
 
+	if d.HasChange("iam_role_arn") {
+		input.RoleARN = aws.String(d.Get("iam_role_arn").(string))
```
catsby: does this work when removing the role?

stack72: I don't believe so - it's the same bug as always AFAICT - when we stop managing something with Terraform, Terraform doesn't recognise that it has "changed"

Issue #10208, opened 2016-11-18
Feature Request: Add support for CreateVPCAssociationAuthorization AWS API


rginev:
Hey there,

Terraform does have [route53_zone_association](https://www.terraform.io/docs/providers/aws/r/route53_zone_association.html) resource, but it works for private zones and VPCs in the same AWS account. 
If you want to associate a VPC that you created with one AWS account with a private hosted zone that you created with a different account you had to ask AWS support to create authorisation manually. 
Now they added an [API](http://docs.aws.amazon.com/Route53/latest/APIReference/API_CreateVPCAssociationAuthorization.html) for this. It would be great to have this implemented in terraform e.g. new resource.

Thanks in advance!

### References
- [Associating an Amazon VPC and a Private Hosted Zone That You Created with Different AWS Accounts](http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-associate-vpcs-different-accounts.html)

bigkraig:
Terraform will also need to be modified on how it detects if a zone is associated. I get errors that I am unable to access the resource (after it takes 30+ seconds to create the association) but when I run a aws cli, I see that it made the association successfully.

druidsbane:
This would definitely be useful as I've just run into this issue myself.

RyanJarv:
So the issue right now is there doesn't seem to be a way to check what zones are associated with a VPC without access to the zone. You can verify that the change completed successfully but if the association was ever modified outside of terraform the account that created the association wouldn't be able to detect this.

FransUrbo:
Duplicate of #12465

opetch:
I've created the following module which by no means solves this issue but provides me with an alternative in the meantime.
https://github.com/opetch/terraform-aws-cli-resource

bevanbennett:
Just as a note... when dealing with cross-account aws_route53_zone_association resources:
1) To CREATE the association the resource provider needs to point to the owner of the VPC
After creation terraform will generate the following error:
* aws_route53_zone_association.allowlocal: AccessDenied: User: arn:aws:iam::111111111111111:user/terraform_user is not authorized to access this resource
	status code: 403, request id: REDACTED
2) To VERIFY the association the resource provider needs to be switched to the owner of the ZONE

This should not effect the ability to run create-vpc-association-authorization, which needs to be executed as the owner of the ZONE. Our current process is:
* manually run create-vpc-association-authorization from the CLI
[...]

PR #12553, opened 2017-03-09
[WIP] provider/aws: aws_route53_vpc_association_authorization resource

RyanJarv:
# Description
Adds the aws_route53_vpc_association_authorization resource which allows
you to create cross account route53 zone associations.
# Related Issues
https://github.com/hashicorp/terraform/issues/10208
# TODO

- [ ] Fix up aws_route53_zone_association to allow for cross account usage
- [ ] Documentation
- [ ] Cross account testing

### DIFF
--- a/builtin/providers/aws/resource_aws_route53_vpc_association_authorization_test.go
+++ b/builtin/providers/aws/resource_aws_route53_vpc_association_authorization_test.go
@@ -1,0 +1,242 @@
+package aws
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+func TestAccAWSRoute53VpcAssociationAuthorization_basic(t *testing.T) {
+	var zone route53.HostedZone
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckRoute53VPCAssociationAuthorizationDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccRoute53VPCAssociationAuthorizationConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRoute53VPCAssociationAuthorizationExists("aws_route53_vpc_association_authorization.peer", &zone),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAWSRoute53VPCAssociationAuthorization_region(t *testing.T) {
+	var zone route53.HostedZone
+
+	// record the initialized providers so that we can use them to
+	// check for the instances in each region
+	var providers []*schema.Provider
+	providerFactories := map[string]terraform.ResourceProviderFactory{
+		"aws": func() (terraform.ResourceProvider, error) {
+			p := Provider()
+			providers = append(providers, p.(*schema.Provider))
+			return p, nil
+		},
+	}
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: providerFactories,
+		CheckDestroy:      testAccCheckRoute53VPCAssociationAuthorizationDestroyWithProviders(&providers),
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccRoute53ZoneAssociationRegionConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRoute53VPCAssociationAuthorizationExistsWithProviders("aws_route53_vpc_association_authorization.peer", &zone, &providers),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckRoute53VPCAssociationAuthorizationDestroy(s *terraform.State) error {
+	return testAccCheckRoute53VPCAssociationAuthorizationDestroyWithProvider(s, testAccProvider)
+}
+
+func testAccCheckRoute53VPCAssociationAuthorizationDestroyWithProviders(providers *[]*schema.Provider) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		for _, provider := range *providers {
+			if provider.Meta() == nil {
+				continue
+			}
+			if err := testAccCheckRoute53VPCAssociationAuthorizationDestroyWithProvider(s, provider); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+func testAccCheckRoute53VPCAssociationAuthorizationDestroyWithProvider(s *terraform.State, provider *schema.Provider) error {
+	conn := provider.Meta().(*AWSClient).r53conn
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_route53_vpc_association_authorization" {
+			continue
+		}
+
+		zone_id, vpc_id := resourceAwsRoute53ZoneAssociationParseId(rs.Primary.ID)
+
+		req := route53.ListVPCAssociationAuthorizationsInput{HostedZoneId: aws.String(zone_id)}
+		res, err := conn.ListVPCAssociationAuthorizations(&req)
+		if err != nil {
+			return err
+		}
+
+		exists := false
+		for _, vpc := range res.VPCs {
+			if vpc_id == *vpc.VPCId {
+				exists = true
+			}
+		}
+
+		if exists {
+			return fmt.Errorf("VPC association authorization for zone %v with %v still exists", zone_id, vpc_id)
+		}
+	}
+	return nil
+}
+
+func testAccCheckRoute53VPCAssociationAuthorizationExists(n string, zone *route53.HostedZone) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		return testAccCheckRoute53VPCAssociationAuthorizationExistsWithProvider(s, n, zone, testAccProvider)
+	}
+}
+
+func testAccCheckRoute53VPCAssociationAuthorizationExistsWithProviders(n string, zone *route53.HostedZone, providers *[]*schema.Provider) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		for _, provider := range *providers {
+			if provider.Meta() == nil {
+				continue
+			}
+			if err := testAccCheckRoute53VPCAssociationAuthorizationExistsWithProvider(s, n, zone, provider); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+func testAccCheckRoute53VPCAssociationAuthorizationExistsWithProvider(s *terraform.State, n string, zone *route53.HostedZone, provider *schema.Provider) error {
+	rs, ok := s.RootModule().Resources[n]
+	if !ok {
+		return fmt.Errorf("Not found: %s", n)
+	}
+
+	if rs.Primary.ID == "" {
+		return fmt.Errorf("No VPC association authorization ID is set")
+	}
+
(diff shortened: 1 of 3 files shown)


RyanJarv:
Not sure if it is possible to create cross account tests yet, saw that it was an issue in https://github.com/hashicorp/terraform/pull/11505 as well.

RyanJarv:
Right now getting this working with aws_route53_zone_association is blocked by a reliable way to retrieve a list of hosted zones associations when you don't have access to the zone.

builtin/providers/aws/resource_aws_route53_vpc_association_authorization_test.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+resource "aws_route53_vpc_association_authorization" "peer" {
+    zone_id = "${aws_route53_zone.foo.id}"
+    vpc_id  = "${aws_vpc.peer.id}"
+}
+
+resource "aws_route53_zone_association" "foobar" {
+	zone_id = "${aws_route53_zone.foo.id}"
+	vpc_id  = "${aws_vpc.peer.id}"
```
FransUrbo: It is needed/required that the association comes **shortly** after the authorisation. I'm not fully clear how long, but I did some tests using the `aws` command to do the authorisation and then run TF to do the association. It can't have been more than half a minute or so between the two, but very often the association failed. Making sure the time between the two was less than half a minute (a few seconds perhaps, five-ten is a guesstimate), then it worked.

So I suggest a `auth_id = "${aws_route53_vpc_association_authorization.peer.id}"` to "link" the two resources in a way that TF creates the authorisation "just before" doing the association. I'm not sure a `depends_on` is enough.

RyanJarv: I don't think I ran into this but if that's true I feel like it should be fixed by AWS (or at least documented so we understand the issue). If we end up tying the two resources together it would prevent you from using it when you don't have access to both account. This is also why I mentioned the blocker above, if we wanted to require access to both accounts we could work around it. I feel like that would be a fairly big limitation though considering a multi account setup is usually for isolation.

I personally don't need this resource anymore, I would still like to finish it up if I get time but right now I'm not sure when that will be. If anyone is interested in this feel free to take over.

FransUrbo:
I've applied this PR onto latest master, rebuilt and installed. Now I'm trying to modify my manifest to use `aws_route53_vpc_association_authorization`, but I can't seem to get it working:

```
resource "aws_route53_vpc_association_authorization" "local2remote" {
  vpc_id        = "${var.remote_vpc_id}"
  zone_id       = "${aws_route53_zone.local_zone.id}"
}

resource "aws_route53_zone_association" "local2remote" {
  vpc_id        = "${var.remote_vpc_id}"
  zone_id       = "${aws_route53_zone.local_zone.id}"
}
```

Gives me:
```
module.validation-elasticsearch.aws_route53_vpc_association_authorization.local2remote: Creating...
  vpc_id:     "" => "vpc-remote"
  vpc_region: "" => "<computed>"
  zone_id:    "" => "LOCAL_ZONE"
[...]
module.validation-elasticsearch.aws_route53_zone_association.local2remote: Creating...
  vpc_id:     "" => "vpc-remote"
  vpc_region: "" => "<computed>"
  zone_id:    "" => "LOCAL_ZONE"
[...]
module.validation-elasticsearch.aws_route53_vpc_association_authorization.local2remote: Creation complete (ID: LOCAL_ZONE:vpc-remote)
[...]
* module.validation-elasticsearch.aws_route53_zone_association.local2remote: 1 error(s) occurred:

* aws_route53_zone_association.local2remote: NotAuthorizedException: The VPC: vpc-remote has not authorized to associate with your hosted zone.
```

And if I add a `provider` for the remote zone in the `aws_route53_vpc_association_authorization` resource, I get:

```
* module.validation-elasticsearch.aws_route53_vpc_association_authorization.local2remote: aws_route53_vpc_association_authorization.local2remote: AccessDenied: User: arn:aws:iam::REMOTE_USER_ID:user/turbo is not authorized to access this resource
```
I'm the admin, with access to everything (as far as I know)...

builtin/providers/aws/resource_aws_route53_vpc_association_authorization_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+resource "aws_route53_zone" "foo" {
+	name = "foo.com"
+	vpc_id = "${aws_vpc.foo.id}"
+}
+
+resource "aws_route53_vpc_association_authorization" "peer" {
+    zone_id = "${aws_route53_zone.foo.id}"
+    vpc_id  = "${aws_vpc.peer.id}"
+	region  = "us-east-2"
```
FransUrbo: Is that supposed to fail?

```
* module.mymodule.aws_route53_vpc_association_authorization.local2remote: : invalid or unknown key: region
```

builtin/providers/aws/resource_aws_route53_vpc_association_authorization_test.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+resource "aws_route53_zone_association" "foobar" {
+	zone_id = "${aws_route53_zone.foo.id}"
+	vpc_id  = "${aws_vpc.peer.id}"
+	region  = "us-east-2"
```
FransUrbo: Ditto:

```
* module.mymodule.aws_route53_zone_association.local2remote: : invalid or unknown key: region
```

builtin/providers/aws/resource_aws_route53_vpc_association_authorization_test.go
asedge: Should be using option `vpc_region` according to the code: https://github.com/hashicorp/terraform/pull/12553/files#diff-4d2ea30d2ceccbd84af4d620b8533ab3R32

FransUrbo:
@RyanJarv Any news on getting this finished?

RyanJarv:
I'll try to take a look at this again soon. Right now I'm thinking we would want to try to implement this using access to both accounts but just make sure we can easily break it up later if/when AWS supports verifying the Route53 server for a VPC from the client account only.

wyardley:
@RyanJarv I guess the way providers work in Terraform, it's not possible to use an account that can assume the role of the other account for this? Also, I guess this wouldn't help in the case where allowing that access isn't feasible?

FransUrbo:
@RyanJarv @wyardley I've successfully created the association authorisation (`aws_route53_vpc_association_authorization`) using a `provider` for "the other account", but what's missing now is to have `aws_route53_zone_association` understand that this is done. I guess. Because that still fails.

RyanJarv:
@wyardley That's a good question, will have to look into that. For the case where access to the other account isn't available I was thinking that it would lead to situations where the client doesn't know the real state of the route53 association, although thinking about this some more this may not be as much of an issue as I thought. Wondering if it's enough to just have it stored in the state file and accept that you may have to make manual changes to the state in certain cases, like if the authorization expired before the association was made (not sure if that's possible), or if the association was removed on the route53 side and we wanted to recreate the authorization (assuming the first one isn't valid anymore). Would be curious if any one else has any thoughts on this or if other resources have the same issue (this is the first one I've worked on).

@FransUrbo Cool, yeah sorry didn't get a chance to see how you did this before. Do you mind making a PR to my branch or just pasting a link to something you have that is working?

wyardley:
@RyanJarv I guess part of the issue is philosophical, I don't know tf well enough to know whether using assume role is something that's typically done. As a user, I personally would find it easier to grant those permissions, use the same account for auth, and provide some way of providing some detail (an account number or ARN, presumably) that would let a single provider handle the association end-to-end. Or maybe just let you specify a second AWS provider and account number in the same resource?

But I don't know if setting up a second provider and using two separate resources is a more 'Terraformish' solution.

In my limited testing using the CLI to do similar associations, if there's already a pending association that wasn't authorized, that can cause issues. So may be worth making sure to test what happens if the authorization is created multiple times without the association being made.

RyanJarv:
@wyardley 
> I guess part of the issue is philosophical, I don't know tf well enough to know whether using assume role is something that's typically done. As a user, I personally would find it easier to grant those permissions, use the same account for auth, and provide some way of providing some detail (an account number or ARN, presumably) that would let a single provider handle the association end-to-end.

The assume role part is already handled by the AWS provider (you can just specify the role and source account instead of keys directly), so the main difference here would be just having a single resource handle the functions of route53_vpc_association_authorization and route53_vpc_association. From my experience the provider connection info seems to be kinda transparent to the provider so we would have to look into if managing two different contexts in the same resource is something that makes sense. This also wouldn't be consistent with how it's handled with the vpc_peering_accepter resource but other then those to things I think this would work.

FransUrbo:
If I understand the documentation for *CreateVPCAssociationAuthorization* correctly, it NEVER (!!) expires the request... It say several times in the documentation that "if you want to delete the authorisation request, delete it by ....".

FransUrbo:
@RyanJarv There's no need for a PR. TF will take care of all of this automatically. I just put a `provider` in the relevant resources:

```
resource "aws_route53_vpc_association_authorization" "validation-elasticsearch" {
  vpc_id        = "${aws_vpc.validation-elasticsearch.id}"
  zone_id       = "${var.r53_zoneid}"
  provider      = "aws.core"
}

resource "aws_route53_zone_association" "validation-elasticsearch" {
  vpc_id        = "${aws_vpc.validation-elasticsearch.id}"
  zone_id       = "${var.r53_zoneid}"
  provider      = "aws.core"
}
```

The first one succeeded just fine, but not the second one. Maybe because TF didn't understand the auth request (or didn't find it, not sure what).

RyanJarv:
Ok so if the authorization never expires I think the only gotcha to what we have now is if the authorization is modified out side of terraform. Assuming that's the only issue I think ideally we would want to detect that but probably fine to just document it for now.

When I made this originally we didn't have a way to do cross account unit testing which is what that last check box is about, not sure if that's changed since.

FransUrbo:
@RyanJarv I think "they" do that for VPC Peering. I think I remember I created an issue about that, and when it was updated (the issue/PR), there where tests for cross-account VPC peering added. Can't remember the issue number though..

colinmorelli:
@RyanJarv According to the AWS docs, deleting the authorization does not delete the association. So, as long as the parent account with the DNS zone can detect if the authorization is still alive (which it can by using the `ListVPCAssociationAuthorizations` API), and the child account with the association can detect if its association is still alive (which it can), shouldn't everything be fine?

If the authorization is deleted outside of terraform, the parent account's terraform plan will reflect the need to re-add it. The child account will rightfully show nothing because the zone will still be associated.

If the association is deleted outside of terraform, the parent account's terraform plan will show nothing, but the child account will show the need to re-authorize.

antonosmond:
Is this likely to progress? I would be really happy to have a vpc_association_authorization resource! I already use the peering resources cross account and have a module that takes advantage of being able to create an additional provider to accept the connection in the other account. It sounds like this would work the same way. Currently I have to use a provisioner which calls the AWS CLI and I'd really like to be able to remove that dependency. 😄

RyanJarv:
@antonosmond Curious, would pushing this forward in the current state with documentation and/or a warning about the incompatibilities of aws_route53_vpc_association_authorization and aws_route53_zone_association address your use case? (see https://github.com/terraform-providers/terraform-provider-aws/issues/384)

antonosmond:
@RyanJarv if we could create the authorization & association on first apply using this then we would probably make it a module which we can target once to set it up and wouldn't need to re-run it so the fact that subsequent apply's would fail wouldn't really matter. Obviously it'd be great if they worked, but I think it'd still be useful to us in it's current state.

RyanJarv:
@antonosmond Correct me if I'm wrong but I think you may be confused about the current state. This works however there would be no way to currently use aws_route53_zone_association with it. If you did use this the association would need to be external to terraform because the aws_route53_zone_association resource makes the assumption the VPC and Zone are owned by the same account.

antonosmond:
Aaaa ok thanks for the clarification @RyanJarv. I think it'd still be useful - I guess I could use a provisioner which does the association so if the authorization was already done by terraform it'd be one less thing for the provisioner to do.

RyanJarv:
Opened a PR for this on the new terraform-providers repo: https://github.com/terraform-providers/terraform-provider-aws/pull/2005

PR #12554, opened 2017-03-09
store endpoint password in hash in state file

arcadiatea:


iceycake:
Do we have any update on this pull request?

catsby:
Hey @arcadiatea thanks for the PR! I wonder, is hashing the password considered (or aiming to be considered) "secure" here? Some of our other resources allow you to encrypt values with a PGP key (see [Lightsail Key Pair](https://www.terraform.io/docs/providers/aws/r/lightsail_key_pair.html)), I wonder if that would be a better approach? I've asked @stack72 to also review as he as some more context here. Thanks!

arcadiatea:
Hi @catsby @stack72, 
We don't need decryption function here. Lightsail only available for us-east-1, which may not help for ticketmaster's case. It is just for solving the problem today we have clean password stored in our tfstat file at backend (s3 bucket). It is explore to multiple teams can see.  Any reason or objection for using hash? Is hash better then encryption since it is not reversible?
Thanks!

Ninir:
@catsby this one has been migrated to the AWS repository. Can you close it please?

grubernaut:
Closing in favor of: https://github.com/terraform-providers/terraform-provider-aws/pull/1083, thanks!

Issue #11135, opened 2017-01-10
Fail to create PagerDuty Schedule with Weekly Restriction


koebane:
Hi there,

I am trying to create a pager duty schedule that utilizes weekly restrictions.  An error gets returned because the PagerDuty API requires start_day_of_week parameter for weekly restrictions but terraform doesn't allow for that parameter according to https://www.terraform.io/docs/providers/pagerduty/r/schedule.html.

Even though the terraform API doesn't acknowledge the start_day_of_week parameter, I tried including it as well.

### Terraform Version
Terraform v0.8.3

### Affected Resource(s)
pagerduty_schedule

### Terraform Configuration Files
```hcl
variable "user_id" {
  type = "map"
  default = {
    user1 = "PD85ZRB"
    user2 = "PQP8L2Y"
    user3  = "P3DX8AA"
  }
}

provider "pagerduty" {
    token = "6TksacesvefcwWp83zxiT"
}

resource "pagerduty_schedule" "my_schedule" {
  name = "${var.name_prefix} Production Weekly Schedule Rotation"
  time_zone = "America/New_York"

  layer {
    name                          = "Evening and Weekend Shift"
    rotation_turn_length_seconds  = 604800 # weekly
    start                         = "2017-01-10T17:00:00-06:00"
    rotation_virtual_start        = "2017-01-06T17:00:00-06:00"
    users                         = [ "${var.user_id["user1"]}", "${var.user_id["user2"]}"]

    restriction {
      type              = "weekly_restriction"
      start_time_of_day = "17:00:00"
      duration_seconds  =  57600  
    }
    restriction {
      type              = "weekly_restriction"
      start_time_of_day = "17:00:00"
      duration_seconds  =  57600  
    }
    restriction {
      type              = "weekly_restriction"
      start_time_of_day = "17:00:00"
      duration_seconds  =  57600  
    }
    restriction {
      type              = "weekly_restriction"
      start_time_of_day = "17:00:00"
      duration_seconds  =  57600  
    }
    restriction {
      type              = "weekly_restriction"
      start_time_of_day = "17:00:00"
      duration_seconds  =  604800  
    }
  }

  layer {
    name                          = "Day and Work Shifts"
    rotation_turn_length_seconds  = 86400 # Daily
    start                         = "2017-01-10T11:54:16-06:00"
    rotation_virtual_start        = "2017-01-10T09:00:00-06:00"
[...]

PR #12559, opened 2017-03-09
provider/pagerduty: Improve errors

heimweh:
This PR should resolve the issue described in: https://github.com/hashicorp/terraform/issues/11135.

Previously when receiving an error from the PagerDuty API the client library couldn't decode the JSON response thus making it very difficult for the user to know what actually went wrong. 
With this fix in place, we should now be able to see what actually went wrong.

Previously:
```bash
* pagerduty_schedule.my_schedule: Response did not contain formatted error: Could not decode JSON response: json: cannot unmarshal object into Go value of type []string. HTTP response code: 400. Raw response: &{Status:400 Bad Request StatusCode:400 Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[Access-Control-Allow-Methods:[GET, POST, PUT, DELETE, OPTIONS] Access-Control-Allow-Origin:[*] Access-Control-Allow-Headers:[Authorization, Content-Type] X-Ua-Compatible:[IE=Edge,chrome=1] Date:[Tue, 10 Jan 2017 18:51:55 GMT] Connection:[keep-alive] Status:[400 Bad Request] Access-Control-Max-Age:[1728000] Cache-Control:[no-cache] X-Request-Id:[70b9b14be473c73684a3c69e478671f0] Server:[nginx] Content-Type:[application/json; charset=utf-8]] Body:0xc4204edc40 ContentLength:-1 TransferEncoding:[chunked] Close:false Uncompressed:false Trailer:map[] Request:0xc4205e7860 TLS:0xc4201b56b0}
```

With the fix in place:
```bash
* pagerduty_schedule.my_schedule: Failed call API endpoint. HTTP response code: 400. Error: &{3001 Invalid Schedule map[rotation_layers.weekly_restrictions.duration_seconds:[Please use a shorter duration or just turn the recurrence off.]]}
```

### DIFF
--- a/vendor/github.com/PagerDuty/go-pagerduty/client.go
+++ b/vendor/github.com/PagerDuty/go-pagerduty/client.go
@@ -37,9 +37,9 @@ type APIReference struct {
 }
 
 type errorObject struct {
-	Code   int      `json:"code,omitempty"`
-	Mesage string   `json:"message,omitempty"`
-	Errors []string `json:"errors,omitempty"`
+	Code    int         `json:"code,omitempty"`
+	Message string      `json:"message,omitempty"`
+	Errors  interface{} `json:"errors,omitempty"`
 }
 
 // Client wraps http client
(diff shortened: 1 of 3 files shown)


stack72:
LGTM thanks for this :)

PR #5212, opened 2016-02-19
Add AWS direct connect virtual interface resources

mhlias:
Hello,

This PR is adding support for DirectConnect virtual interfaces as seen in the documentation here: http://docs.aws.amazon.com/directconnect/latest/UserGuide/createhostedvirtualinterface.html

I decided to not create separate resources for public and private virtual interfaces as they only have minor differences at creation time and from there they are described as the same resource by the API. So I am handling them with parameters instead. I am not sure if that is a good way of doing it or not, so any feedback on that would be welcome.

I also have two types of resources. One for interfaces that will be created on the same AWS account where the direct connect connection has been created and one for when the virtual interface will be created for usage on a separate AWS account to the one that has the direct connect connection.
I provide an example in the documentation on how to use the latter case.

I wasn't sure how to handle some metaparameters that are used for internal logic and not directly for the resource itself. Also how to have required parameters or not based on another parameter's value. e.g. interface_type private or public require 2 different things, the former requires a virtual_gateway_id and the latter requires a route_filter_prefixes. I would appreciate some feedback on that.

I added some testing but it is not easy to run as it requires a pre-existing direct connect connection which requires an offline procedure to get it set up. Please provide some feedback on how testing should be for this kind of resource.

Best,

Ilias.

PR #12567, opened 2017-03-09
Add initial Direct Connect resources

bodgit:
This is based in part on the work started in #5212 however I've just adapted the single resource I need:

* `aws_directconnect_public_virtual_interface_confirm`

I've updated it with the missing attributes and import functionality and also renamed the resource slightly so it fits better with the naming convention of other resources, (IMHO!).

Sadly this has the same sticking point in that testing it the same way as most other AWS resources doesn't really work as a) it involves more than one AWS account and b) Direct Connect infrastructure quite often involves some manual operator involvement or can take some time to be verified by AWS staff.

The [directconnectiface package](https://docs.aws.amazon.com/sdk-for-go/api/service/directconnect/directconnectiface/) gives the ability to mock the API so I could test the Terraform resource without having to create anything physical/virtual however I'm struggling to work out how I can slot it in to the testing framework, it looks like I need to make my own `AWSClient` with a mocked  `dcconn`.

I can't see any existing resources that I can copy or use for inspiration so I'm asking for some guidance here. I'd really like to get this resource added as it's the last outstanding piece I have left and maybe once the tests are figured out it'll help with adding the remaining Direct Connect resources.


akumria:
[bikeshed]: It really would be nice if the resource was named 'aws_dx_...'

Comparing the two PRs, I generally prefer PR#5212 since cover both the public and private virtual interface confirmation.

bodgit:
I'll close this in preference to terraform-providers/terraform-provider-aws#876.

PR #12583, opened 2017-03-10
Add note for terraform remote config

Technowise:
Added note on required account type for successfully running "terraform remote config" command given in this guide.


mitchellh:
I just made this a lot clearer on master. We'll be shipping 0.9 tomorrow and you can check it out then!

Issue #12566, opened 2017-03-09
provider/aws: Add support for ACM certificates in API Gateway


MrGossett:
### Affected Resource(s)
- api_gateway_domain_name

AWS has [announced](https://aws.amazon.com/about-aws/whats-new/2017/03/amazon-api-gateway-integrates-with-aws-certificate-manager-acm/) that API Gateway Custom Domains can now use certificates from Amazon Certificate Manager.

PR #12592, opened 2017-03-10
provider/aws: Add support for ACM certificates to api_gateway_domain_name

stack72:
Fixes: #12566

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAPIGatewayDomainName_'         ✹
2017/03/10 19:32:31 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAPIGatewayDomainName_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayDomainName_basic
--- PASS: TestAccAWSAPIGatewayDomainName_basic (54.06s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	54.091s
```

Unfortunately, due to ACM not being automatable, we can't add a test for this!

### DIFF
--- a/builtin/providers/aws/resource_aws_api_gateway_domain_name.go
+++ b/builtin/providers/aws/resource_aws_api_gateway_domain_name.go
@@ -21,33 +21,41 @@ func resourceAwsApiGatewayDomainName() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 
+			//According to AWS Documentation, ACM will be the only way to add certificates
+			//to ApiGateway DomainNames. When this happens, we will be deprecating all certificate methods
+			//except certificate_arn. We are not quite sure when this will happen.
 			"certificate_body": {
 				Type:     schema.TypeString,
 				ForceNew: true,
-				Required: true,
+				Optional: true,
 			},
 
 			"certificate_chain": {
 				Type:     schema.TypeString,
 				ForceNew: true,
-				Required: true,
+				Optional: true,
 			},
 
 			"certificate_name": {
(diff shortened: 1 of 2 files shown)


stack72:
Requires PR #12591

Ninir:
@stack72 Hi Paul,

Even though it has been merged, have a few questions on this:

As the CLI already deprecates it (http://docs.aws.amazon.com/apigateway/api-reference/link-relation/domainname-create/), what do you think of deprecating it also to be consistent?

Also, don't you think we should mark `certificate_arn` as conflicting with `certificate_body`, `certificate_chain`, `certificate_name`, `certificate_private_key`?

Thanks for the work on this, really helpful :)

radeksimko:
Re conflicting - I agree - would you mind sending a PR for this? :)

Regarding deprecation - we did have a brief conversation with Paul about this - I was firstly suggesting the same thing as you, but then we both realized that ACM doesn't offer a good automation mechanisms (which is also why we don't have any ACM resources in Terraform yet) - therefore we have no reasonable alternatives to offer to the user, unless the alternative is "go ahead and manually verify domain in ACM and then come back", which I think is suboptimal.

Therefore we're hoping AWS will eventually support DNS based verification prior to removing the deprecated certificate handling and then we'd actually deprecate or remove those fields here.

Ninir:
Hi @radeksimko !

Sure, here it is: https://github.com/hashicorp/terraform/pull/12649
Sounds good for the next steps on that, thanks for the inputs!

PR #12595, opened 2017-03-10
Fixes #11258

### LINKED ISSUE #11258: Iterating through an array of users in aws_opsworks_user_profile fails.
Hi there,

i found a problem within the resource **aws_opsworks_user_profile** 
It seems like you cannot properly iterate with count through an array of users.
The initial `terraform apply` works perfectly. It creates all users in the correct way. Then you change one user within the array and terraform tries to modify the already existing user with a new username which is not possible in opsworks. I would expect a delete operation for the user that does not exist anymore and the creation of the new user.

### Terraform Version
0.8.4 (probably lower versions are effected as well)

### Affected Resource(s)
- aws_opsworks_user_profile

```
variable "aws_access_key" {} # reading from terraform.tfvars file
variable "aws_secret_key" {} # reading from terraform.tfvars file
variable "aws_region" {} # reading from terraform.tfvars file
variable "aws_iam_user_arn" { default = ["arn:aws:iam::0000000000000:user/testuNM27", "arn:aws:iam::0000000000000:user/testuNM28", "arn:aws:iam::0000000000000:user/testuNM29"]}
variable "aws_opsworks_stack_id" { default = "xxxxxx-xxxx-xxxx-xxxxx-xxxxxxxxx" }
variable "aws_iam_user_ssh" { default = ["testuNM27", "testuNM28", "testuNM29"]}


##############

provider "aws" { access_key = "${var.aws_access_key}" secret_key = "${var.aws_secret_key}" region = "${var.aws_region}" }


resource "aws_opsworks_user_profile" "my_profile" {
  count = "${length(var.aws_iam_user_arn)}"
  ssh_username = "${element(var.aws_iam_user_ssh, count.index)}"
  user_arn = "${element(var.aws_iam_user_arn, count.index)}"
  allow_self_management = true
}

```

### Expected Behavior
 Delete and Create operation instead of modify.

### Actual Behavior

Modify operation - which will cause the AWS API to send an error, because this operation is not possible .
```

~ aws_opsworks_user_profile.my_profile.0
    ssh_username: "testuNM27" => "testuNM25"
    user_arn:     "arn:aws:iam::000000000000:user/testuNM27" => "arn:aws:iam::000000000000:user/testuNM25"


```

### Steps to Reproduce
1. Use code snippet above
2. Initial `terraform apply`
3. Change one user in both arrays (e.g. testuNM27" => "testuNM25" )
4. `terraform apply `

Samangan:
Bug: #11258
If the user_arn changes on the opsworks user profile resource we should delete it and recreate it instead of attempting to modify it incorrectly and causing the terraform apply to fail.

I also added an acceptance test that will trigger this case.

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_user_profile.go
+++ b/builtin/providers/aws/resource_aws_opsworks_user_profile.go
@@ -26,6 +26,7 @@ func resourceAwsOpsworksUserProfile() *schema.Resource {
 			"user_arn": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 
 			"allow_self_management": &schema.Schema{
--- a/builtin/providers/aws/resource_aws_opsworks_user_profile_test.go
+++ b/builtin/providers/aws/resource_aws_opsworks_user_profile_test.go
@@ -14,14 +14,14 @@ import (
 
 func TestAccAWSOpsworksUserProfile(t *testing.T) {
 	rName := fmt.Sprintf("test-user-%d", acctest.RandInt())
-	roleName := fmt.Sprintf("tf-ops-user-profile-%d", acctest.RandInt())
+	updateRName := fmt.Sprintf("test-user-%d", acctest.RandInt())
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckAwsOpsworksUserProfileDestroy,
 		Steps: []resource.TestStep{
 			resource.TestStep{
-				Config: testAccAwsOpsworksUserProfileCreate(rName, roleName),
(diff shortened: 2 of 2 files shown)


radeksimko:
Thanks for the PR, I was able to reproduce the issue. Even though it's not entirely clear from the [docs](http://docs.aws.amazon.com/opsworks/latest/APIReference/API_UpdateUserProfile.html) I understand the IAM User ARN is used as a unique identifier for the user profile and there's always 1-1 mapping between IAM user and the opsworks profile.

Also thanks for the test! 👍 

```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSOpsworksUser -timeout 120m
=== RUN   TestAccAWSOpsworksUserProfile
--- PASS: TestAccAWSOpsworksUserProfile (29.85s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	29.880s
```

PR #12597, opened 2017-03-10
Give keys and values map functions an option to disable lexical sorting of keys

Zordrak:
This is an alternative option for https://github.com/hashicorp/terraform/pull/12598


Zordrak:
This is an alternative option for https://github.com/hashicorp/terraform/pull/12598

Zordrak:
https://blog.golang.org/go-maps-in-action#TOC_7.

*(╯°□°）╯︵ ┻━┻*

PR #12598, opened 2017-03-10
Add specific and independent unsortedkeys and unsortedvalues functions

Zordrak:
For retrieving keys and values from maps without any lexical sorting. This prevents recreation of swathes of resources when using count to generate many resources based on the contents of a map.

This is an alternative option for https://github.com/hashicorp/terraform/pull/12597


Zordrak:
This is an alternative option for:
https://github.com/hashicorp/terraform/pull/12597

Zordrak:
https://blog.golang.org/go-maps-in-action#TOC_7.

*(╯°□°）╯︵ ┻━┻*

PR #12601, opened 2017-03-11
provider/aws: catch invalid lambda function timeout in plan

netjunki:
Encountered the following error after a terraform apply failed:

    * aws_lambda_function.LambdaFunction: Error modifying Lambda Function Configuration cifull1-Redshift-Persistence: ValidationException: 1 validation error detected: Value '600' at 'timeout' failed to satisfy constraint: Member must have value less than or equal to 300
	status code: 400, request id: e8d3f902-05d8-11e7-9dff-9b2b32ac011c

This patch adds a validator to the timeout value to ensure it is under during 300 during plan.

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -83,9 +83,10 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				ValidateFunc: validateRuntime,
 			},
 			"timeout": {
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  3,
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      3,
+				ValidateFunc: validateTimeout,
 			},
 			"publish": {
 				Type:     schema.TypeBool,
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_lambda_function.go
```diff
@@ -586,3 +587,14 @@ func validateRuntime(v interface{}, k string) (ws []string, errors []error) {
 	}
 	return
 }
+
+func validateTimeout(v interface{}, k string) (ws []string, errors []error) {
+	timeout := v.(int)
+
+	if timeout > 300 {
+		errors = append(errors, fmt.Errorf(
+			"Lambda timeout can't exceed 300. Currently set to %d.",
```
stack72: Hi @netjunki 

I can't find any mention of the 300 timeout in the AWS docs?

Paul

netjunki: I couldn't either. All I know is that I get an error back from them with a value. I'm guessing with both have accounts with AWS do you want to open up a support request to ask them about it? Or shall I?

stack72: Yeah looks like it's an account specific thing - I therefore think adding this validation may not be the best thing to do.

Thoughts?

Paul

netjunki: Drop it. And maybe one day we'll have an easy way to query all limits in aws. :-)

stack72: ok sounds good to me - i love all the PRs you are sending so please do continue :)

We love it when community members help to make the product better!

Paul

netjunki: Scratching my itches and getting familiar with the code base. :-)

Issue #12599, opened 2017-03-10
Deleted AWS IAM SAML Identity Providers are not recreated


jfautley:
If an AWS IAM Identity Provider entry is created by Terraform and subsequently deleted by a non-Terraform process, the state refresh fails with a 404 error. This causes the entire Terraform plan/apply operation to fail.

### Terraform Version
0.8.8 (also present on 0.7.13)

### Affected Resource(s)
- aws_iam_saml_provider

### Terraform Configuration Files
```hcl
resource "aws_iam_saml_provider" "default" {
    name = "TEST-FED"
    saml_metadata_document = "${file("./SAML.xml")}"
}
```

### Debug Output
Trimmed debug output is here: https://gist.github.com/jfautley/c1a19c09b2644dbc7179512d172ce225

### Expected Behavior
The deleted resource should have been re-created.

### Actual Behavior
Terraform exited with an error, and the output below.

```
aws_iam_saml_provider.default: Refreshing state... (ID: arn:aws:iam::1111111111:saml-provider/TEST-FED)
Error refreshing state: 1 error(s) occurred:

* aws_iam_saml_provider.default: NoSuchEntity: Manifest not found for arn arn:aws:iam::111111111111:saml-provider/TEST-FED
	status code: 404, request id: 282212e4-05e9-11e7-8649-d3eeaa03ac0e
```

### Steps to Reproduce
1. Obtain a valid SAML metadata XML definition file
2. Execute the above Terraform configuration example (`terraform apply`)
3. Delete the SAML provider from the AWS console (or via the API)
4. Execute a `terraform apply` or `terraform plan` operation and observe the failure
  (optional)
5. Delete the local Terraform state file, and re-run your plan/apply (which will succeed)

### Important Factoids
This is only an issue because of the presence of the resource in the Terraform state file. If the state file is removed/hacked to delete references to the relevant resource, a `terraform apply` operation completes successfully.

stack72:
Hi @jfautley 

Sorry for the issue here - I just opened a PR that will fix this - it will be part of the next release of Terraform

Paul

PR #12602, opened 2017-03-11
provider/aws: Refresh iam saml provider from state on 404

stack72:
Fixes: #12599

Before this patch:

```
% terraform plan

aws_iam_saml_provider.salesforce: Refreshing state... (ID: arn:aws:i...rce-test)

* aws_iam_saml_provider.salesforce: aws_iam_saml_provider.salesforce: NoSuchEntity: Manifest not found for arn arn:aws:iam::187416307283:saml-provider/tf-salesforce-test
	status code: 404, request id: fc32c7f8-0631-11e7-8e1f-29a8c10edf64
```

After this patch:

```
% terraform plan                                                                                  ✚ ✭
[WARN] /Users/stacko/Code/go/bin/terraform-provider-aws overrides an internal plugin for aws-provider.

aws_iam_saml_provider.salesforce: Refreshing state... (ID: arn:aws:i...rce-test)

+ aws_iam_saml_provider.salesforce
    arn:                    "<computed>"
    name:                   "tf-salesforce-test"
```

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_saml_provider.go
+++ b/builtin/providers/aws/resource_aws_iam_saml_provider.go
@@ -2,10 +2,12 @@ package aws
 
 import (
 	"fmt"
+	"log"
 	"regexp"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
 
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_iam_saml_provider.go
```diff
@@ -70,6 +72,11 @@ func resourceAwsIamSamlProviderRead(d *schema.ResourceData, meta interface{}) er
 	}
 	out, err := iamconn.GetSAMLProvider(input)
 	if err != nil {
+		if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == "NoSuchEntity" {
```
radeksimko: it's not exactly EC2 error in this context, but that's really a nitpick 😄

stack72: You are correct - fixed now :) Will merge on green

Issue #9979, opened 2016-11-08
provider/openstack: openstack_compute_instance_v2 export availability_zone attribute


ChiefAlexander:
Would it be possible to add availability_zone to the list of exported attributes for the openstack provider?

jtopjian:
Good idea. I will look into this.

ChiefAlexander:
Done via https://github.com/hashicorp/terraform/pull/12610

PR #12610, opened 2017-03-11
provider/openstack: Set Availability Zone in Instances

jtopjian:
This configures the openstack_compute_instance_v2 resource
to set the Availability Zone of the instance resource.

For #9979


stack72:
LGTM! :)

PR #12612, opened 2017-03-12
English sentence correction

goodshailesh:
English sentence correction. Corrected following:
"map values are merged and all are values are overridden"
to
"map values are merged and all other values are overridden"


stack72:
LGTM thanks for this :)

Issue #9952, opened 2016-11-08
openstack_compute_floatingip_v2 always recreates FIP when using pool id instead pool name


Kiemes:
### Terraform Version
0.7.9

### Affected Resource(s)
openstack_compute_floatingip_v2

### Terraform Configuration Files
```hcl
resource "openstack_compute_floatingip_v2" "lifecycle_floating_ip" {
  region = "${var.region_name}"
  pool   = "${var.ext_net_id}"
}
```

### Expected/Actual Behavior
We know that according to the [documentation](https://www.terraform.io/docs/providers/openstack/r/compute_floatingip_v2.html) the pool value should be a pool name and not the pool id as we used it by mistake. When using the pool id, the FIP is created correctly as the [OpenStack API](http://developer.openstack.org/api-ref/compute/?expanded=create-allocate-floating-ip-address-detail) also allows a pool id (even if not mentioned explicitly). But it is recreated on every run of `terraform  apply` and not  just kept as it is, when it's already there.

Expected behavior would be that there either is an error message when a non-existent name is used (such as the pool id), or that the state check works with the id as well, and the floating IP is not recreated when it already exists.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` with a configuration containing a floating IP  and the pool referenced with the id
2. `terraform apply` on the same configuration

jtopjian:
Thank you for reporting this. This is definitely not expected behavior.

I have a few questions:

1. Does this happen when you use the `openstack_networking_floatingip_v2` resource? If your OpenStack environment uses Neutron (which I imagine it would nowadays), both resources interact with Neutron an the same floating IP pool in the end.

2. Can you tell me more about your OpenStack environment (what release, what Neutron driver, etc) as well as provide a full Terraform configuration of how this is being triggered?

Thanks again! Hopefully we can sort this out.

beyhan:
@jtopjian 
Hi, we verified 1):
When using `openstack_networking_floatingip_v2`, the expected error message is actually raised.
```
openstack_networking_floatingip_v2.lifecycle_floating_ip: Creating...
  address:   "" => "<computed>"
  fixed_ip:  "" => "<computed>"
  pool:      "" => "90e5af93-b59b-46ab-8a3e-0febe1a6b9ad"
  port_id:   "" => "<computed>"
  region:    "" => <redacted>
  tenant_id: "" => "<computed>"
Error applying plan:

1 error(s) occurred:

[...]

PR #12615, opened 2017-03-12
provider/openstack: Add network_id to Network data source

jtopjian:
This commit adds the ability to search for a network by its ID. This
is useful for doing ID-to-name translations.

For #9952

### DIFF
--- a/builtin/providers/openstack/data_source_openstack_networking_network_v2.go
+++ b/builtin/providers/openstack/data_source_openstack_networking_network_v2.go
@@ -17,6 +17,10 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 		Read: dataSourceNetworkingNetworkV2Read,
 
 		Schema: map[string]*schema.Schema{
+			"network_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


builtin/providers/openstack/data_source_openstack_networking_network_v2.go
```diff
@@ -56,6 +60,7 @@ func dataSourceNetworkingNetworkV2Read(d *schema.ResourceData, meta interface{})
 	networkingClient, err := config.networkingV2Client(GetRegion(d))
 
 	listOpts := networks.ListOpts{
+		ID:       d.Get("network_id").(string),
```
jtopjian: Yep, Gophercloud will ignore zero-value parameters in the GET URL. For example, here's the URL that's built for searching by name

```GET http://10.1.12.81:9696/v2.0/networks?name=tf_test_network&status=ACTIVE```

and for searching by network_id:

```GET http://10.1.12.81:9696/v2.0/networks?id=d1f4958c-dc7b-4726-8b3a-1a748b131e74&status=ACTIVE```

`tenant_id` is also unset for both, which doesn't get passed through :)

stack72: 👍 In that case, LGTM!

jtopjian: Thanks, Paul!

PR #12619, opened 2017-03-12
provider/dnsimple: Don't prompt the user for email address

stack72:
As part of the new changes to the DNSimple provider, we changed to use
the new API version. This requires a token and *not* email address

In order for backwards compatibility - we kept the email address in the
schema but we had the default as nil, meaning that Terraform was
prompting the user for it, they would enter it, then Terraform would
error out due to using a combination of token and email address

This commit makes the default email address an empty String. This means
we don't prompt the user


stack72:
//cc @weppos

Issue #12376, opened 2017-03-02
aws_instances data source


mattlqx:
I would find it useful to have an `aws_instances` data source that would be able to execute a filter and return the instance IDs of multiple instances. Presently the `aws_instance` data source says:

```
NOTE: If more or less than a single match is returned by the search, Terraform will fail. Ensure that your search is specific enough to return a single Instance ID only.
```

My use-case is that have a Terraform repo that has some common shared infrastructure (like HTTP hosts that hand-out 301s for whatever URL they're given) and I want to attach that to multiple AWS ALBs that are in separate Terraform repos (per application). I would like to do a search on those instances based on tags and be able to use the result list in an `aws_alb_target_group_attachment`. I'd really like to be able to have an `aws_alb_target_group` data source, but this is an acceptable (and useful in more ways) alternative.

Right now, the only real way I can get those instances is to hardcode the instance IDs in individual `aws_instance` data sources. I can work out some outputs where the hosts are created, and work them into being imported into the application Terraforms but this additional layer increases complexity and can become stale or out-of-sync if not done correctly. It's better just to hardcode the IDs for now.

vaygr:
I second this.

PR #12624, opened 2017-03-12
WIP: implements aws_instances

artburkart:
I'm not sure if it's acceptable to create an unfinished PR, but I hit a wall and am unsure what to do to resolve it. This PR is in response to the request for an "aws_instances" data source in Issue #12376. My approach was to make the returned data have a field called "instances", which would be a list of instances with the same format as the one returned in the "aws_instance" data source. 

I wanted to be able to do something like this: https://gist.github.com/artburkart/16f88b5856861d164dae4b1aeefc0724

Because I wanted it to be similar to the "aws_route_table" data source which lets you do this: https://gist.github.com/artburkart/d5dd81302179493b2ab2c23fbfcb84b8

Everything seems to work. The multiple instances are getting queried and the data sources are getting populated, but for some reason, I can't get them to attach to the "aws_instances" ResourceData object. I have a feeling the problem is in the code here: https://github.com/artburkart/terraform/blob/e7f03452ffd58fc953c0915d77e5b8551e1f70f5/builtin/providers/aws/data_source_aws_instance.go#L315-L334

In this code I tried to follow the pattern laid out in [data_source_aws_route_table](https://github.com/hashicorp/terraform/blob/aa4676e6222921f06dd451af571512679ef1dea6/builtin/providers/aws/data_source_aws_route_table.go#L146-L148), but something has gone awry. Whenever I perform a `terraform apply` with the configs I've provided in this PR, I get this:

```
data.aws_instances.boxes: Refreshing state...
data.aws_instances.boxes: Refreshing state...
Error applying plan:

1 error(s) occurred:

* data.template_file.boxes: Resource 'data.aws_instances.boxes' does not have attribute 'instances.0.subnet_id' for variable 'data.aws_instances.boxes.instances.0.subnet_id'
```

For some reason, `data.aws_instances.boxes.instances` does not exist, and I'm not exactly sure why not. I'm going to keep pecking at it, but if someone could offer some guidance, it would be much appreciated. Once I figure this bug out, I'll add all the tests and docs that generally accompany a new data source.

Cheers

### DIFF
--- a/builtin/providers/aws/data_source_aws_instance.go
+++ b/builtin/providers/aws/data_source_aws_instance.go
@@ -9,193 +9,182 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-func dataSourceAwsInstance() *schema.Resource {
-	return &schema.Resource{
-		Read: dataSourceAwsInstanceRead,
+func awsInstanceSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"tags": dataSourceTagsSchema(),
+		"ami": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"instance_type": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"instance_state": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"availability_zone": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"tenancy": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"key_name": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"public_dns": {
+			Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


artburkart:
Well, I've noticed an indication of a bug. Some code is getting double-executed. I think, once I get to the bottom of that, I may have a working solution.

cemo:
eager to see :) What is the state of this issue?

artburkart:
I'm still debugging my code, will keep you posted.

artburkart:
I've figured out my problem. Everything I had was more-or-less correct, except I didn't realize resources and data sources *must* have at least one invocation of `d.SetId()`. I didn't do this, and so every time it got to the state writing portion of execution, it would see that no id was set and throw a fit.

I should have a working copy with tests and docs soon.

artburkart:
I'm closing this PR in favor of a different one I made. You can find it here: 
https://github.com/hashicorp/terraform/pull/12856

PR #12659, opened 2017-03-13
provider/google: Remove deprecated project fields.

paddyforan:
Remove the deprecated fields from google_project, and drop all the logic
that went into supporting them. Tests still pass after one minor change.

### DIFF
--- a/builtin/providers/google/resource_google_project.go
+++ b/builtin/providers/google/resource_google_project.go
@@ -1,7 +1,6 @@
 package google
 
 import (
-	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


builtin/providers/google/resource_google_project_test.go
```diff
@@ -214,35 +214,16 @@ resource "google_project" "acceptance" {
 `, pid)
 }
 
-func testAccGoogleProjectImportExistingWithIam(pid string) string {
```
danawillow: While you're deleting uncalled functions, want to also delete `testAccGoogleProjectImportExisting`? (or if you think that should actually be made into a test, maybe add a todo)

builtin/providers/google/resource_google_project.go
```diff
@@ -242,6 +172,7 @@ func resourceGoogleProjectRead(d *schema.ResourceData, meta interface{}) error {
 func prefixedProject(pid string) string {
 	return "projects/" + pid
 }
+
 func resourceGoogleProjectUpdate(d *schema.ResourceData, meta interface{}) error {
```
paddyforan: I *think* it's OK, because it's not creating a new resource, then trying to assign that resource, which leaves the possibility of the second call failing and leaving the resource dangling. But I'd have to check my notes on that (SetPartial leaves me fuzzy as to when exactly it's _necessary_, so I don't want to be loudly wrong here.) Want to chat about it on Thursday?

danawillow: Sure! And that's definitely not a blocker for getting this PR merged.

paddyforan:
@danawillow just pruned the dead function. While I wait for Travis, and have you here, still look good?

danawillow:
Yup!

PR #12661, opened 2017-03-13
Change example value to lower case in ECS service

endofcake:
According to the validator, the only acceptable values are 'cpu' and 'memory' (both lower case).
https://github.com/hashicorp/terraform/blob/73f4acf0415925399c521982e2d48225ab611a52/builtin/providers/aws/validators.go#L722:L725
Updating the docs so that this example works out of the box.


stack72:
LGTM - thanks for this :)

PR #12664, opened 2017-03-14
Check for ErrDefault500 when creating/deleting pool member

kars7e:
If LBaaSV2 member creation/deletion is attempted while LB resource is
busy (for example it's in PENDING_UPDATE state), Neutron will return 500
error. Currently This being caught by type assertion to
ErrUnexpectedResponseCode, however error 500 has a dedicated type
ErrDefault500, which makes the current type assertion to miss it and
eventually the error slips out through the check. This patch
changes the type assertion to explicitly check for ErrDefault500.

Also similar check has been added when member is deleted.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_lb_member_v2.go
+++ b/builtin/providers/openstack/resource_openstack_lb_member_v2.go
@@ -124,15 +124,20 @@ func resourceMemberV2Create(d *schema.ResourceData, meta interface{}) error {
 		log.Printf("[DEBUG] Attempting to create LBaaSV2 member")
 		member, err = pools.CreateMember(networkingClient, poolID, createOpts).Extract()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


kars7e:
I was hitting following error:

openstack_lb_member_v2.k8s-member-1: Error deleting OpenStack LBaaSV2 Member: Internal Server Error

When deleting a member (or similar, respective error when creating it). I was able to track it down to the fact 500 error had its own concrete type, and assertion with that type was required.

I've tested this patch in our production environment and it works, no more errors when deleting/creating pool members!

jtopjian:
@kars7e Nice catch - thank you! I'll do a quick test of this tonight.

jtopjian:
LGTM!

@stack72 This checks out on my end. :)

stack72:
LGTM - thanks @kars7e / @jtopjian

PR #12668, opened 2017-03-14
provider/google: remove deprecated account_file field.

paddyforan:
Remove the shims for the long-deprecated account_file field in the
Google provider.


builtin/providers/google/provider.go
```diff
@@ -21,7 +21,7 @@ func Provider() terraform.ResourceProvider {
 				Optional:     true,
 				DefaultFunc:  schema.EnvDefaultFunc("GOOGLE_ACCOUNT_FILE", nil),
 				ValidateFunc: validateAccountFile,
```
danawillow: Do you still need the ValidateFunc if the field itself is Removed? If not then you can remove the function entirely.

paddyforan: Cool, let's do that. :deleting code dance: Thanks!

paddyforan:
Check failed on unrelated test, booping it and trying again.

Issue #12673, opened 2017-03-14
Can't create Azure Elastic Pool sql db


holmesb:
### Terraform Version
v0.8.8

### Affected Resource(s)
azurerm_sql_database

### Terraform Configuration Files
```provider "azurerm" {
  subscription_id = "${var.subscription_id}"
  client_id       = "${var.client_id}"
  client_secret   = "${var.client_secret}"
  tenant_id       = "${var.tenant_id}"
}
resource "azurerm_resource_group" "main_rg" {
  name     = "test_resource_group"
  location = "West Europe"
}
resource "azurerm_sql_server" "sql_server" {
    name = "testserverepfail"
    resource_group_name = "${azurerm_resource_group.main_rg.name}"
    location = "West Europe"
    version = "12.0"
    administrator_login = "user1"
    administrator_login_password = "secretpassword1!"
}
resource "azurerm_sql_database" "test_db" {
    name = "test_db"
    resource_group_name = "${azurerm_resource_group.main_rg.name}"
    location = "West Europe"
    server_name = "testserverepfail"
    elastic_pool_name = "test_ep"
    requested_service_objective_name = "ElasticPool"
```

### Debug Output
"azurerm_sql_database.test_db: Error creating SQL Database: 40862 (400) - Elastic pool name must be specified if the elastic pool service objective is specified."

### Expected Behavior
Elastic pool database is created

### Actual Behavior
Elastic pool database fails to be created

### Steps to Reproduce
- Create elastic pool with name "test_ep" on sql server name "testserverepfail" (or names of your choosing - you'll need to change above code).
- `terraform apply`

PR #12678, opened 2017-03-14
[WIP] provider/azurerm: Add support for elastic_pool_name to sql_database

stack72:
Fixes: #12673

Unfortunately, this ElasticPool has to actually be created so we need to
implement the ElasticPool resource


stack72:
Closed in favour of #14099 :)

dominik-lekse:
Hi @stack72,

we need this pull request since #14099 only contains the elastic pool resource. This pull request allow to create a database as part of an elastic pool. I did not include the changes from #12678 in #14099.

After merging #14099, we could continue working on this. In particular, we need to add a test case for creating a sql database as part of an elastic pool.

PR #4166, opened 2015-12-04
core: support HTTP basic auth in consul remote state

phinze:
Closes #1663

PR #12679, opened 2017-03-14
Add basic auth to consul provider

commarla:
This PR adds a basic auth option to the consul provider.

Most of the code is copied from this https://github.com/hashicorp/terraform/pull/4166


stack72:
Hi @commarla 

Paul

PR #12684, opened 2017-03-14
Include a consul server for tests

jbardin:
This uses a local consul test server for backend unit tests, as well as the provider acceptance tests.

The backend tests weren't being run automatically anywhere, and the provider ACC tests relied on the consul public demo server, which wasn't always available.


jbardin:
This will now always run the backend consul tests in travis, but makes it opt-in locally so contributors aren't required to have consul installed.

jbardin:
Unfortunate result of pulling in `consul/testutil`. We can probably come up with something better in the long run, but this is self-contained and works for now.

Issue #12674, opened 2017-03-14
Cannot destroy instance when reclaim_instance is enabled in Openstack Nova.


cloudrkt:
Hi,

I cannot destroy an instance created on Openstack when the status is soft_deleted. Can terraform force delete an instance?

### Terraform Version

Terraform v0.8.8

### Affected Resource(s)
- openstack instance 

### Terraform Configuration Files
```hcl

resource "openstack_compute_floatingip_v2" "terraform" {
  pool       = "${var.pool}"
  depends_on = ["openstack_networking_router_interface_v2.terraform"]
}

resource "openstack_compute_keypair_v2" "terraform" {
  name       = "terraform"
  public_key = "${file("${var.ssh_key_file}.pub")}"
}

resource "openstack_networking_network_v2" "terraform" {
  name           = "terraform"
  admin_state_up = "true"
}

resource "openstack_networking_subnet_v2" "terraform" {
  name            = "terraform"
  network_id      = "${openstack_networking_network_v2.terraform.id}"
  cidr            = "10.0.0.0/24"
  ip_version      = 4
  dns_nameservers = ["8.8.8.8", "8.8.4.4"]
}

resource "openstack_networking_router_v2" "terraform" {
  name             = "terraform"
  admin_state_up   = "true"
  external_gateway = "${var.default_public_gateway_id}"
}

resource "openstack_networking_router_interface_v2" "terraform" {
  router_id = "${openstack_networking_router_v2.terraform.id}"
  subnet_id = "${openstack_networking_subnet_v2.terraform.id}"
}

resource "openstack_compute_secgroup_v2" "terraform" {
  name        = "terraform"
  description = "Security group for the Terraform example instances"
}

resource "openstack_compute_instance_v2" "terraform" {
  availability_zone = "NL1"
  name              = "terraform"
  image_name        = "${var.image}"
  flavor_name       = "${var.flavor}"
  key_pair          = "${openstack_compute_keypair_v2.terraform.name}"
  security_groups   = ["${openstack_compute_secgroup_v2.terraform.name}"]
  floating_ip       = "${openstack_compute_floatingip_v2.terraform.address}"

  network {
    uuid = "${openstack_networking_network_v2.terraform.id}"
  }
}
```

### Debug Output

https://gist.github.com/cloudrkt/4ecf84692f79ffa71120b4037eeb6db5

### Expected Behavior
Should delete the instance when using terraform destroy

### Actual Behavior
Error message when deleting instance

```
[...]

PR #12689, opened 2017-03-14
provider/openstack: Force Deletion of Instances

### LINKED ISSUE #5104: Openstack - Interesting things happen when soft delete is enabled
Openstack has featured labelled "soft-delete", it enables cloud admins to restore a VM for a certain amount of time before the VM is fully removed.

It functions very different to standard delete, for a start, the VM is in a state of "soft-deleted" and all volumes, VIP, security groups associated with this VM are still seen to be in use.

This causes a great mess for terraform, even if you modify the code to see a soft-deleted we'd have to do some polling for who knows how long to delete the instance and associated items.

How would people feel about using ForceDelete as a default for openstack instance destroys? This would mean soft delete is bypassed for terraform and terraform can function as normal in an environment where soft delete is enabled.

I've recently added support in gophercloud for this https://github.com/rackspace/gophercloud/commit/370c4300029d713ff36f3377c3e98e5ad73b1c4d so it wouldn't be too hard to fit it in to our current model.

Thoughts?

Thanks,
Ian.

jtopjian:
This commit adds the `force_delete` argument, enabling instances
to be forcefully deleted.

For #12674
Fixes #5104

/cc @cloudrkt

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
@@ -322,6 +322,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"force_delete": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


stack72:
Hey @jtopjian 

Be aware that, if this resource supports import, then there will be a diff in the test as this value isn't set to state :)

P.

jtopjian:
@stack72 Ahh thanks. Instances aren't able to be imported yet (hopefully soon, though!)

What would be the best way to handle this in the future, though?

stack72:
if you look at the custom import function on aws_db_instance you can see how we set the defaults :)

jtopjian:
Do you mean how the fields listed in `import_aws_db_instance_test.go` are ignored or the `DiffSuppressFunc`?

stack72:
Look at this

```
Importer: &schema.ResourceImporter{
			State: resourceAwsDbInstanceImport,
		},
```

jtopjian:
Ah - gotcha. Thank you! OK, that's really good to know. Might apply to some other areas, too. :)

PR #12699, opened 2017-03-14
add gzip to consul backend tests

jbardin:
just making sure to cover the gzip codepath in tests

### DIFF
--- a/backend/remote-state/consul/backend_test.go
+++ b/backend/remote-state/consul/backend_test.go
@@ -47,3 +47,18 @@ func TestBackend(t *testing.T) {
 	// Test
 	backend.TestBackend(t, b)
 }
+
+func TestBackend_gzip(t *testing.T) {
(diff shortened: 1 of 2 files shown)


backend/remote-state/consul/backend_test.go
```diff
@@ -42,6 +42,7 @@ func TestBackend(t *testing.T) {
 	b := backend.TestBackendConfig(t, New(), map[string]interface{}{
 		"address": srv.HTTPAddr,
 		"path":    fmt.Sprintf("tf-unit/%s", time.Now().String()),
+		"gzip":    true,
```
mitchellh: Can we test both gzip and not-gzip? (so new test `TestBackend_gzip`)

jbardin: Sure. I did it over in the client tests too, but it doesn't hurt to have it here as well.

Issue #12608, opened 2017-03-11
Race condition in multiple Scaleway IP creation


zarbis:
### Terraform Version
Terraform v0.8.8

### Affected Resource(s)
- scaleway_ip

### Terraform Configuration Files
```hcl
resource "scaleway_ip" "public_ip" {
  count = "5"
}
```

### Debug Output
https://gist.github.com/zarbis/cf530e2451b5c11a428d920526672d35

### Expected Behavior
Five Scaleway IPs created.

### Actual Behavior
Only one random IP created. Index of successful IP changes every time.
While `terraform apply -parallelism=1` always gives expected behavior.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

nicolai86:
I'll take a look later today. Thank you for the bug report!

nicolai86:
I've looked into this and it looks like a Scaleway API issue. The issue goes away once I put a mutex in place. I've raised the issue in the official SDK and will provide an hack which limits concurrency inside the Scaleway Provider for this specific case…

nicolai86:
closing as it has been fixed /w #12707

PR #12707, opened 2017-03-14
provider/scaleway: work around API concurrency issue

nicolai86:
when creating IPs concurrently the Scaleway API starts to return 500 internal
server errors.

since the error goes away when limiting concurrent requests, as well as the fact that the golang net/http client is safe for concurrent use,
I'm assuming this is an API error on Scaleways side.

this CS introduces a workaround so terraform does not crash.
the work around needs to be removed once Scaleway fixes their API

I've added a test case which reproduces the error mentioned in #12608:

```
make testacc TEST=./builtin/providers/scaleway TESTARGS='-run=TestAccScalewayIP_Count'
2017/03/14 22:24:15 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/scaleway -v -run=TestAccScalewayIP_Count -timeout 120m
=== RUN   TestAccScalewayIP_Count
--- FAIL: TestAccScalewayIP_Count (2.73s)
    testing.go:268: Step 0 error: Error applying: 1 error(s) occurred:

        * scaleway_ip.base[1]: 1 error(s) occurred:

        * scaleway_ip.base.1: {"message": "Internal error", "type": "server_error"}
FAIL
exit status 1
FAIL    github.com/hashicorp/terraform/builtin/providers/scaleway    2.742s
```

and the mutex fixes this:

```
make testacc TEST=./builtin/providers/scaleway TESTARGS='-run=TestAccScalewayIP_Count'
2017/03/14 22:34:56 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/scaleway -v -run=TestAccScalewayIP_Count -timeout 120m
=== RUN   TestAccScalewayIP_Count
--- PASS: TestAccScalewayIP_Count (5.17s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/scaleway    5.182s
```

### DIFF
--- a/builtin/providers/scaleway/resource_scaleway_ip.go
+++ b/builtin/providers/scaleway/resource_scaleway_ip.go
@@ -2,6 +2,7 @@ package scaleway
 
 import (
 	"log"
+	"sync"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/scaleway/scaleway-cli/pkg/api"
@@ -30,8 +31,12 @@ func resourceScalewayIP() *schema.Resource {
 	}
 }
 
+var mu = sync.Mutex{}
+
 func resourceScalewayIPCreate(d *schema.ResourceData, m interface{}) error {
 	scaleway := m.(*Client).scaleway
+	mu.Lock()
+	defer mu.Unlock()
 	resp, err := scaleway.NewIP()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


nicolai86:
\cc @stack72 not what I was looking for, but I'd choose stability over performance while the issue is under investigation.

stack72:
LGTM! you rock @nicolai86 :)

PR #12725, opened 2017-03-15
Revert "Revert "provider/alicloud: change create ecs postpaid instance API""

stack72:
Reverting the revert


jbardin:
nope, this doesn't fix the extra reverts issue, they're not here

PR #12726, opened 2017-03-15
fix another hot lop in tests

jbardin:
Found another test spinlock.
Slow it down to prevent it from blocking the runtime scheduler.


mitchellh:
Nice, LGTM

PR #12729, opened 2017-03-15
provider/google: drop the account file.

paddyforan:
This was already marked as removed, but the way the provider handled it,
people were still being prompted for input anyways. This removes it from
the provider entirely, so people won't be prompted for input.


danawillow:
Alternative: should terraform core change so that people stop getting prompted for removed fields?

(I'm ok with this change but it seems to me that if best practice is to mark the field as removed instead of actually removing it, then we should try to adhere to that if possible)

PR #12733, opened 2017-03-15
Revert "Have the consul provider use a local test server"

jbardin:
This reverts commit 9529bd3bf040e5c7d62468aa560f98f145dabc45.

Pulling the consul `testutil.TestServer` out of the provider tests for now. The consul testutil package uses a `*testing.T` for errors and control flow, and returns a server that needs to be stopped. The provider however is sometimes configured multiple times per test, and there's often no good point to insert the testing instance or clean up the server. 

It's probably easier to provide another hook at some point in the consul testutil package rather than refactor all the tests here to fit.


jbardin:
Yeah. The backend unit tests however will continue to use this setup, so they will still run in travis.

PR #12737, opened 2017-03-15
Update environments.html.md

sprutner:
kind of confusing, example right below is different than quoted command.

### DIFF
--- a/website/source/docs/state/environments.html.md
+++ b/website/source/docs/state/environments.html.md
@@ -28,7 +28,7 @@ it cannot ever be deleted. If you've never explicitly used environments, then
 you've only ever worked on the "default" environment.
 
 Environments are managed with the `terraform env` set of commands. To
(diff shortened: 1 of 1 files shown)


paddyforan:
Hi! Thanks for reporting the issue. It looks like this is backwards, though; `new` is correct, `create` is not.

```
$ terraform env create
Usage: terraform env

Subcommands:

    list      List environments.
    select    Select an environment.
    new       Create a new environment.
    delete    Delete an existing environment.
```

```
$ terraform env new
Expected a single argument: NAME.

Usage: terraform env new [OPTIONS] NAME [DIR]

  Create a new Terraform environment.


Options:

    -state=path    Copy an existing state file into the new environment.
```

radeksimko:
Fixed via https://github.com/hashicorp/terraform/pull/12781

PR #12745, opened 2017-03-15
[WIP] provider/aws: Return information about failed AMI filter

NoUseFreak:


NoUseFreak:
This issue occurred when using modules, it was not possible to show what module failed. As the full path is now printed this is no longer needed.

PR #12748, opened 2017-03-16
Added schema to expose public IPv6

willemvv:
Exposing public IPv6 for scaleway provider based on #9217


nicolai86:
\cc @stack72

PR #12750, opened 2017-03-16
Add navigation for aws_instance documentation

jasonlotito:
aws_instance documentation is currently available on the site; however, a link is not provided via the navigation under the data sources section.  This adds that link.


stack72:
LGTM! Thanks forthis :)

PR #12772, opened 2017-03-16
Don't interpolate multivariables during walkInput

### LINKED ISSUE #12742: Terraform 0.9.0 regression
I just updated Terraform from v0.8.8 to v0.9.0 and I think that I have found a regression in `module` behavior. The original code that I was having issues with is quite large, but I think that I have managed to produce a minimum working example, which I have pushed to a separate repository to make it easier to reproduce. The code can be found at https://github.com/joshuaspence/terraform-0.9.0-regression.

On Terraform 0.8.8 running `terraform plan` runs fine, but in Terraform 0.9.0 I get the following error:

```
Error asking for user input: 1 error(s) occurred:

* module.vpc.module.jumphost.var.vpc_zone_identifier: strconv.ParseInt: parsing "${var.availability_zones_count}": invalid syntax
```

jbardin:
Another unfortunate instance of an `i.Operation ==` check in the interpolator.

We don't need these expanded for input, and if the multivar's count is a
variable, it may be known but not available during the input walk.

Fixes #12742

### DIFF
--- a/terraform/context_input_test.go
+++ b/terraform/context_input_test.go
@@ -658,3 +658,30 @@ func TestContext2Input_hcl(t *testing.T) {
 		t.Fatalf("bad: \n%s", actualStr)
 	}
 }
+
+// adding a list interpolation in fails to interpolate the count variable
+func TestContext2Input_submoduleTriggersInvalidCount(t *testing.T) {
+	input := new(MockUIInput)
+	m := testModule(t, "input-submodule-count")
+	p := testProvider("aws")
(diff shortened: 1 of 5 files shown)


SerCeMan:
Hey! Thanks for the fix 👍 ! When it's going to be released?

Also, we encountered this issue only on MacOS machines, on linux everything works fine.

mitchellh:
0.9.1 should come out tomorrow barring any unforseen issues

PR #12779, opened 2017-03-16
provider/google: add support for GKE node pool autoscaling

### LINKED ISSUE #8605: [GKE] Support autoscaling

danawillow:
Fixes #8605.

Note that I decided not to include the "enabled" boolean in the terraform state, and instead just assert that if the autoscaling field exists in a terraform config then autoscaling is enabled. Let me know if you notice any gotchas with that approach, or if you think it makes more sense just to go with what the API has.

### DIFF
--- a/builtin/providers/google/resource_container_node_pool.go
+++ b/builtin/providers/google/resource_container_node_pool.go
@@ -55,6 +55,27 @@ func resourceContainerNodePool() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+
+			"autoscaling": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"min_node_count": &schema.Schema{
+							Type:     schema.TypeInt,
+							Required: true,
+							ForceNew: true,
+						},
+						"max_node_count": &schema.Schema{
+							Type:     schema.TypeInt,
+							Required: true,
+							ForceNew: true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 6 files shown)


danawillow:
Ah, looks like you can update autoscaling but I didn't notice because it's part of the clusters API call instead of the nodepools one. I just sent an email to the GKE team to check that I'm understanding things correctly overall; I'll update again once I've gotten a clear answer.

Liberski:
@danawillow @paddycarver Any update on this?

danawillow:
No updates right now, but it's still on our radar. A few higher-priority things came up but I'll try to finish this up soonish.

munnerz:
This is pretty important for a project I'm working on at the moment.

Would it be possible to get it merged in as-is at the moment? (with ForceNew still enabled). I may be able to spend some time working on this if there's anything extra needed doing.

danawillow:
Hey @munnerz, we're actually in a code freeze at the moment so this won't be able to get merged until next week at the earliest (I'm not sure how long the freeze will last for).

munnerz:
Hey again - I'm not sure where to check on the status of code freeze so sorry to keep bothering you. Again, is there anything I can do to move this along? 😄

danawillow:
Ah whoops! So the code freeze was a part of the provider split that happened (we're now in https://github.com/terraform-providers/terraform-provider-google). Let me port this over there and finish it up right now.

danawillow:
New PR at https://github.com/terraform-providers/terraform-provider-google/pull/157

PR #12785, opened 2017-03-16
command: apply needs to look at the right field for backend state

mitchellh:
Plans were properly encoding backend configuration but the apply was
reading it from the wrong field. :( This meant that every apply from a
plan was applying it locally with backends.

This needs to get released ASAP.


mitchellh:
Tests passssssss

PR #12794, opened 2017-03-17
provider/openstack: Add timeout support for Compute resources

jtopjian:
This commit adds support for specifying timeout values for the
openstack_compute_* family of resources.


jtopjian:
@stack72 Let me know if I missed anything.

stack72:
@jtopjian LGTM!

PR #12795, opened 2017-03-17
backend/atlas: accept ATLAS_ADDRESS env var for address

mitchellh:
This is required for on-premise TFE.


ryanuber:
:heart_eyes_cat: :+1:

PR #12805, opened 2017-03-17
provider/powerdns: add support for set-ptr option

anuriq:
According to PowerDNS API specification:
set-ptr: If set to true, the server will find the matching reverse zone and create a PTR there. Existing PTR records are replaced. If no matching reverse Zone, an error is thrown. Only valid in client bodies, only valid for A and AAAA types. Not returned by the server. Only valid for the Authoritative server.
See https://doc.powerdns.com/md/httpapi/api_spec/

### DIFF
--- a/builtin/providers/powerdns/client.go
+++ b/builtin/providers/powerdns/client.go
@@ -86,6 +86,7 @@ type Record struct {
 	Content  string `json:"content"`
 	TTL      int    `json:"ttl"` // For API v0
 	Disabled bool   `json:"disabled"`
+	SetPtr   bool   `json:"set-ptr"`
 }
 
 type ResourceRecordSet struct {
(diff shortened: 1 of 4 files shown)


linuxus:
Hello @anuriq has this PR made it into the latest change. I've just downloaded the provider for TF v0.10 and it looks this is not merged. Any plan to have this included soon?
Thanks.

anuriq:
Hi @linuxus 
I am creator of this PR, but have no connection to Hashicorp or Terraform developers. 
By the way, plugin system changed, and now each of them has its own repo. I'll add new PR to the repo of powerdns plugin ASAP.

linuxus:
Thanks Azat for putting this into a new PR on the new split out provider
model. I think it just makes sense to have PTR records support added as
most of DNS A/AAAA records do require a reverse pointer as well.

Abdi

anuriq:
Moved to https://github.com/terraform-providers/terraform-provider-powerdns/pull/4

PR #12807, opened 2017-03-17
provider/openstack: Add fixed_ips computed attribute to port resource

jtopjian:
This commit adds the `fixed_ips` attribute to the
openstack_networking_port_v2 resource. This contains all of the port's
Fixed IPs returned by the Networking v2 API.

/cc @sebastien-prudhomme

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_port_v2.go
+++ b/builtin/providers/openstack/resource_openstack_networking_port_v2.go
@@ -98,6 +98,11 @@ func resourceNetworkingPortV2() *schema.Resource {
 					},
 				},
 			},
+			"fixed_ips": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
 			"allowed_address_pairs": &schema.Schema{
 				Type:     schema.TypeSet,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


jtopjian:
I'm not 100% happy with the name `fixed_ips` since it's quite close to the existing `fixed_ip`. Marking this WIP until either I think of a better name or cannot think of one at all.

I don't want to use `ip_addresses` since that implies _all_ types of IPs, such as Floating IPs.

Suggestions are appreciated! :)

jtopjian:
Alright, having mulled it over for a few hours, I think `fixed_ips` is the safest name.

ready for review.

sebastien-prudhomme:
Really no way with the Terraform language to use the **fixed_ip** attributes?

jtopjian:
@sebastien-prudhomme Not to my knowledge. Because users can specify a Fixed IP in _any_ order and because Neutron will return those IPs in a different order, ordered lists can't be used. The "hash" form is the only way I know of. 

This proposed feature will resume the previous method you were using, you would just need to add an "s" to the lines you were using it. :)

jtopjian:
I've been testing this with a bunch of scenarios and the change to `TypeSet` has also "broken" cases where someone specifies a `fixed_ip` with no `ip_address` (meaning, they want DHCP to issue an address). A crash does not happen, but there will be a state change upon refresh.

It seems like there's a conflict satisfying all port/IP scenarios:

* With `TypeList`, when a user specifies `fixed_ip`s out of alphabetical/numerical order, they will see a state change upon refresh.
* With `TypeSet`, when a user specifies a single `fixed_ip` (which is probably the vast majority of users), they cannot easily reference the IP address of that IP, whether static or DHCP.
* With `TypeSet`, when a user specifies a `fixed_ip` with no `ip_address`, they will see a state change upon refresh.
* With `TypeSet`, there's no way, at the moment, to specify multiple `fixed_ip`s with DHCP, since the hashing won't be able to accurately identify the `fixed_ip` upon first `apply`.

It looks like `TypeSet` has the bulk of the problems, but really, all but the first are edge cases. However, they still need to be accounted for.

Back to the drawing board on this one...

jtopjian:
I've created issue #13052 to better track this. I'm going to close this PR.

PR #12811, opened 2017-03-17
provider/google: Add support node config for GKE node pool

aadidenko:


danawillow:
Thanks @aadidenko! I'm wondering if it actually makes sense to separate some of the NodeConfig stuff out both from NodePool and Cluster, maybe into its own file (or just have helper methods inside one of the existing ones), since they'll always have the same set of fields. What do you think?

meekrosoft:
@danawillow I think having it in the NodePool makes sense - I would want the machines in a given pool to be homogeneous intuitively.  Is there a case for them being a mix?

danawillow:
@meekrosoft I think you misunderstood my comment, let me try to rephrase. In the GCP API, NodeConfig is its [own type](https://cloud.google.com/container-engine/reference/rest/v1/NodeConfig). It then gets referenced from both [cluster](https://cloud.google.com/container-engine/reference/rest/v1/projects.zones.clusters) and [NodePool](https://cloud.google.com/container-engine/reference/rest/v1/projects.zones.clusters.nodePools). So rather than duplicating the logic in two places, I was just suggesting that we only declare it once and then call it from both places. Does that make more sense?

aadidenko:
@danawillow I agree with you, I will prepare the changes

Timer:
Friendly ping, would love to see this completed & merged. 😄
Let me know if there's anything I can do to help!

aadidenko:
I pushed the changes to https://github.com/terraform-providers/terraform-provider-google/pull/184

grubernaut:
Thanks for migrating this @aadidenko, closing original.

PR #12812, opened 2017-03-17
fix race in hook ui PreApply test

jbardin:
Fix a race in the PreApply test, and make the PreApply background task
actually concealable.


jbardin:
@mitchellh:  re-worked this to on need a small addition to synchronize the test using the existing UiHook

PR #12829, opened 2017-03-17
Fix ECS field inconsistency

KensoDev:
ECS has some inconsistency in the way it sends down
field names and the way it expects them in the api.

`MEMORY` and `CPU` are expected capitalized and `instanceId` should be sent as
is.

The previous fixed lowercased everything, meaning `instanceId` is being
saved as `instanceid` and causing a change in the resource.

PRs and issues

* https://github.com/hashicorp/terraform/pull/12199
* https://github.com/hashicorp/terraform/issues/11644
* https://github.com/hashicorp/terraform/issues/11844


stack72:
Closed via #12998

Sorry I didn't see this one before merging the other - thanks so much for the contribution!

PR #12838, opened 2017-03-18
Joyent cloud uses user-data for some of their smart datacenter config…

zero-below:
In Joyent's images, at least for ubuntu, user-data is used for joyent infrastructure and cloud-init needs to be uploaded to cloud-init:user-data instead of just user-data.

See notes in: https://docs.joyent.com/public-cloud/instances/virtual-machines/images/linux/ubuntu-certified#cloud-init-examples

`Cloud-init data is provided as metadata with the cloud-init:user-data key. Joyent boot-time data is supplied as metadata with the user-data and user-script keys.`

I've built and tested this in my own infrastructure in joyent, with v0.9.0. This successfully placed the key onto the machine, and a basic cloud-config template was applied.

```  cloud_config = <<EOF
#cloud-config
write_files:
-   content: test123
    owner: root:root
    path: /tmp/testfile123
    permissions: '0644'
EOF
```

```
ubuntu@dev-infra-01:~$ cat /tmp/testfile123 
test123ubuntu@dev-infra-01:~$
```

### DIFF
--- a/builtin/providers/triton/resource_machine.go
+++ b/builtin/providers/triton/resource_machine.go
@@ -25,6 +25,7 @@ var (
 		"user_script":          "user-script",
 		"user_data":            "user-data",
 		"administrator_pw":     "administrator-pw",
+		"cloud_config":         "cloud-init:user-data",
 	}
 )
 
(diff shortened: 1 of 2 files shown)


zero-below:
Honestly, it would probably be better to have some sort of a meta-data block, and then each key/val can be set as metadata. But this was sufficient for the main use case for now.

zero-below:
Travis seems to have died on a test of the consul remote (which isn't touched in this commit). The triton portions of the test succeeded.

PR #12839, opened 2017-03-18
Fix go vet error

cblecker:
```
go vet .
builtin/providers/aws/resource_aws_kms_key.go:181: arg metadata.KeyId for printf verb %s of wrong type: *string
exit status 1
```

Fix vet error introduced in #12243


stack72:
LGTM! thanks for this :)

PR #10760, opened 2016-12-15
Upgrade DNSimple provider to API v2

weppos:
A couple of days ago [DNSimple announced the next major release](https://blog.dnsimple.com/2016/12/api-v2-stable/) of the API. Among the various improvements, API v2 supports the following new features:

1. OAuth
1. Multi-accounts

In the last months, few users contacted us in particular about the second point, as they created a separate account (e.g. for the company) and they wanted to be able to use it via Terraform.

This PR upgrades the current DNSimple provider to API v2, and replaces @pearkes's Go client with the [official DNSimple Go client](https://github.com/dnsimple/dnsimple-go).

I tried to follow the instructions in the CONTRIBUTING.md file, but I'd be happy if someone can review the changes and provide a feedback. Some notes:

- I removed @pearkes's client from the vendor folder, and added the DNSimple one in a separate commit, as requested in the CONTRIBUTING file.
- I upgraded the provider to v2
- I needed to access some config inside the resources, hence I added a Client wrapper (similar approaches were adopted e.g. in the GitHub provider and others) in https://github.com/hashicorp/terraform/commit/ba723c5d07f6ba9a49315e6d0415fc4c0949ce2a. Please let me know if there is a preferred way.

Open questions:

- API v1 and API v2 are not compatible. The API token is different, as well the auth mechanism. I've tried to add a warning if the user had the previous token configured. Is there a better way?
- Some names in the resource no longer really reflect the API structure. For e.g, we replaced "domain" with "zone", that it will be more relevant also when we will expose reverse zones. Any objection to change them as part of this major change?
- There are certain resource fields that Terraform defines as String, although they are actually integers (e.g. TTL, Priority). I've seen a very limited usage of non-String types, even in other providers. Is there a recommended approach? Should I leave them String?

The documentation is not updated yet, I'll work on it, including feedback from this PR (if any). I will probably highlight API v2 support. Does it make sense to provide an API v1 to v2 Terraform upgrade section?

PR #12843, opened 2017-03-18
Allow dnsimple_record.priority attribute to be set

yob:
This is an updated version of PR #9128. Unlike the original PR, this one isn't blocked by an upstream change (thanks to #10760) and is hopefully ready for review.

Some DNS record types (like MX) allow a priority to specified, and the ability to do so is important in many environments.

This diff makes `dnsimple_record.priority` optional, allowing it to be used in terraform configs like so:

```
    resource "dnsimple_record" "mx1" {
        domain = "example.com"
        name = ""
        value = "mx1.example.com"
        type = "MX"
        priority = "1"
    }

    resource "dnsimple_record" "mx2" {
        domain = "example.com"
        name = ""
        value = "mx2.example.com"
        type = "MX"
        priority = "2"
    }
```

An earlier issue (https://github.com/hashicorp/terraform/issues/1058) proposed this feature , but was eventually closed due to inactivity.

### DIFF
--- a/builtin/providers/dnsimple/resource_dnsimple_record.go
+++ b/builtin/providers/dnsimple/resource_dnsimple_record.go
@@ -58,6 +58,7 @@ func resourceDNSimpleRecord() *schema.Resource {
 			"priority": {
 				Type:     schema.TypeString,
 				Computed: true,
+				Optional: true,
 			},
 		},
 	}
@@ -74,6 +75,10 @@ func resourceDNSimpleRecordCreate(d *schema.ResourceData, meta interface{}) erro
 	}
 	if attr, ok := d.GetOk("ttl"); ok {
 		newRecord.TTL, _ = strconv.Atoi(attr.(string))
+	}
+
+	if attr, ok := d.GetOk("priority"); ok {
+		newRecord.Priority, _ = strconv.Atoi(attr.(string))
 	}
 
 	log.Printf("[DEBUG] DNSimple Record create configuration: %#v", newRecord)
@@ -140,6 +145,10 @@ func resourceDNSimpleRecordUpdate(d *schema.ResourceData, meta interface{}) erro
 	}
 	if attr, ok := d.GetOk("ttl"); ok {
 		updateRecord.TTL, _ = strconv.Atoi(attr.(string))
+	}
+
+	if attr, ok := d.GetOk("priority"); ok {
+		updateRecord.Priority, _ = strconv.Atoi(attr.(string))
 	}
 
 	log.Printf("[DEBUG] DNSimple Record update configuration: %#v", updateRecord)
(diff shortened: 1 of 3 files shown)


yob:
@stack72 @weppos - I'd welcome any feedback on this :)

stack72:
This LGTM! The only nit I have is the conversion of string -> Int. Is there a chance this could error out? If so, then we should follow up with a PR on that to make sure we catch the errors

```
% make testacc TEST=./builtin/providers/dnsimple                                                                                            ✭
2017/03/18 13:03:49 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/dnsimple -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccDNSimpleRecord_Basic
--- PASS: TestAccDNSimpleRecord_Basic (1.80s)
=== RUN   TestAccDNSimpleRecord_CreateMxWithPriority
--- PASS: TestAccDNSimpleRecord_CreateMxWithPriority (1.22s)
=== RUN   TestAccDNSimpleRecord_Updated
--- PASS: TestAccDNSimpleRecord_Updated (2.02s)
=== RUN   TestAccDNSimpleRecord_UpdatedMx
--- PASS: TestAccDNSimpleRecord_UpdatedMx (2.16s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/dnsimple	7.213s
```

yob:
> This LGTM!

Cheers.

> The only nit I have is the conversion of string -> Int. Is there a chance this could error out?

I'm not yet familiar enough with terraform's internals or the DNSimple API to be sure. I just followed the precedent set by the TTL attribute (which is also an int) in the existing code.

I notice that @weppos raised the same issue in #10760:

> There are certain resource fields that Terraform defines as String, although they are actually integers (e.g. TTL, Priority). I've seen a very limited usage of non-String types, even in other providers. Is there a recommended approach? Should I leave them String?

I'll be happy to open a follow up PR if you can offer any guidance on the best way forward.

weppos:
Thanks @yob. Yes, I'd definitely like to be interested in understanding more about String vs Int usage in Terraform.

I've seen that most fields are everywhere defined as `schema.TypeString`. In fact, even the DNSimple provider had it. Any guideline @stack72 ?

stack72:
Hi @weppos / @yob 

So the good news (I have just been informed by @radeksimko) is that the " " are optional - therefore

this:

```
resource "dnsimple_record" "foobar" {
	domain = "hashicorptest.com"

	name = ""
	value = "mx.example.com"
	type = "MX"
	ttl = "3600"
	priority = "5"
}
```

is equivalent to:

```
resource "dnsimple_record" "foobar" {
	domain = "hashicorptest.com"

	name = ""
	value = "mx.example.com"
	type = "MX"
	ttl = 3600
	priority = 5
}
```

So we should be able to change the schema easily enough :)

Paul

PR #12860, opened 2017-03-19
[docs][tfstate]Align document to the general recommendation

SergK:
Terraform also puts some state into the `terraform.tfstate` file
by default. This state file is extremely important...
This will mean that any potential secrets
stored in the state file, will not be checked into version control

Related issue #516

### DIFF
--- a/website/source/docs/state/environments.html.md
+++ b/website/source/docs/state/environments.html.md
@@ -114,7 +114,7 @@ aren't any more complex than that. Terraform wraps this simple notion with
 a set of protections and support for remote state.
 
 For local state, Terraform stores the state environments in a folder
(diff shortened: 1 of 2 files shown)


website/source/docs/state/environments.html.md
```diff
@@ -114,7 +114,7 @@ aren't any more complex than that. Terraform wraps this simple notion with
 a set of protections and support for remote state.
 
 For local state, Terraform stores the state environments in a folder
-`terraform.state.d`. This folder should be committed to version control
+`terraform.state.d`. This folder should not be committed to version control
```
mitchellh: It actually should be, this is not a typo! If you're using local-only state, it should be committed to version control. If you don't want state in version control then you should be using remote backends.

PR #12864, opened 2017-03-19
index(list, var.element) throws error when element not found

### LINKED ISSUE #12066: index(list, var.element) throws error when element not found
### Terraform Version
```
% tf --version
Terraform v0.8.7
```

### Affected Resource(s)
- core

This is a core issue.

```
variable "list" {
  default = [ "one", "two", "three" ]
}

data "null_data_source" "indextest" {
  inputs = {
    result = "${index(var.list, "four")}"
  }
}
```

### Debug Output
I don't think that's necessary

### Expected Behavior
index should return -1.

### Actual Behavior
index threw an error that halted execution:
```
fix these before continuing.

Errors:

  * index: Could not find 'four' in '[{Variable (TypeString): one} {Variable (TypeString): two} {Variable (TypeString): three}]' in:

${index(var.list, "four")}
```

### Steps to Reproduce
1. `terraform apply`
2. ???
3. ...

### Important Factoids
This is dumb bug.

### References
- couldn't find anything relevant

snowsky:
Fixes #12066

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -594,7 +594,7 @@ func interpolationFuncIndex() ast.Function {
 					return index, nil
 				}
 			}
-			return nil, fmt.Errorf("Could not find '%s' in '%s'", needle, haystack)
+			return -1, nil
 		},
 	}
 }


snowsky:
The test case failed because it is expecting an error instead of -1 and without error.

teamterraform:
Hi @snowsky! Thanks for working on this, and sorry for the very long delay in responding.

As you can see in #12066, we chose to resolve this instead by making the conditional operator work as an explicit expression of the same default of `-1` you implemented here. You can see over there a more detailed rationale, but the short form is that we prefer to make this behavior explicit in the configuration rather than implicit in Terraform's implementation.

With that said, we're going to close out this PR now. Sorry again that it took so long to respond to it, and thanks for working on this change.

PR #12870, opened 2017-03-19
Add a substring interpolation function

### LINKED ISSUE #1431: Add substring between built-in interpolation functions
I can see this being useful in following situation:

``` ruby
resource "aws_subnet" "private" {
  count = "${var.number_of_azs}"

  vpc_id = "${aws_vpc.default.id}"
  cidr_block = "${element(split(",", lookup(var.subnet_cidrs, "private")), count.index)}"
  availability_zone = "${element(split(",", lookup(var.aws_avalability_zones, var.aws_region)), count.index)}"

  tags {
    Name = "${substr(-1, 1, element(split(",", lookup(var.aws_avalability_zones, var.aws_region)), count.index))}-private"
  }
}
```

so the important bit:

```
${substr(-1, 1, element(split(",", lookup(var.aws_avalability_zones, var.aws_region)), count.index))}
```

if the `element` returns e.g. `eu-west-1a` then turns into `a`, which means I can get

```
tags {
  Name = "a-private"
}
```

joshuaspence:
Fixes #1431. Adds a new `substr` interpolation function which can be used to truncate a string.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -88,6 +88,7 @@ func Funcs() map[string]ast.Function {
 		"slice":        interpolationFuncSlice(),
 		"sort":         interpolationFuncSort(),
 		"split":        interpolationFuncSplit(),
+		"substr":       interpolationFuncSubstr(),
 		"timestamp":    interpolationFuncTimestamp(),
 		"title":        interpolationFuncTitle(),
 		"trimspace":    interpolationFuncTrimSpace(),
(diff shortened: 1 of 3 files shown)


joshuaspence:
```
> make test TEST='./config' TESTARGS='-run=TestInterpolateFuncSubstr'
2017/03/22 08:50:59 Generated command/internal_plugin_list.go
go test -i ./config || exit 1
echo ./config | \
	xargs -t -n4 go test -run=TestInterpolateFuncSubstr -timeout=30s -parallel=4
go test -run=TestInterpolateFuncSubstr -timeout=30s -parallel=4 ./config 
ok  	github.com/hashicorp/terraform/config	0.011s
```

dforsber-p:
`substr: 'offset + length' cannot be larger than the length of the string`

grubernaut:
Hey @dforsber-p, I'm assuming this is an issue you encountered with the `substring()` interpolation function. Would you mind opening a detailed issue on [Terraform](https://github.com/hashicorp/terraform) with the problem you're seeing? 
Thanks!

PR #12873, opened 2017-03-20
Add DigitalOcean snapshot Data Source.

andrewsomething:
This PR adds a data source for DigitalOcean snapshots similar to the aws_ami data source. This allows for referencing user created snapshots when creating a new Droplet or volume.

### DIFF
--- a/builtin/providers/digitalocean/data_source_digitalocean_snapshot.go
+++ b/builtin/providers/digitalocean/data_source_digitalocean_snapshot.go
@@ -1,0 +1,218 @@
+package digitalocean
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"sort"
+	"time"
+
+	"github.com/digitalocean/godo"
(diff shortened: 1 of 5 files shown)


andrewsomething:
I'm realizing I'm a bit unclear on the acceptance tests for data sources. I can run them locally, but they're looking at resources on my account. Is there some way I should be creating the resources first, or will the values be updated to reflect resources on the testing account?

stack72:
Hi @andrewsomething 

Yes, the way we usually do it is like this:

https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/data_source_aws_autoscaling_groups_test.go#L15

Notice i have a config that rolls out the actual resources as a first step, then a follow up that adds the datasource

Paul

andrewsomething:
@stack72 thanks for the pointer! Though, I guess I'll have to withdraw this for now. Looks like I need to implement a resource to take snapshots before I can implement the data source.

PR #12226, opened 2017-02-24
provider/alicloud: change create ecs postpaid instance API

demonwy:
Mainly: change create ecs postpaid instance API form createInstance to runInstances, support BusinessInfo and userdata

IMPROVEMENTS:

resource/alicloud_instance: change create ecs postpaid instance API form createInstance to runInstances, support BusinessInfo [#80](https://github.com/alibaba/terraform-provider/pull/80)
resource/alicloud_instance: change ecs parameter zoneId from required to optional [#74](https://github.com/alibaba/terraform-provider/pull/74)
resource/alicloud_instance: support userdata [#71](https://github.com/alibaba/terraform-provider/pull/71)

BUG FIXES:

resource/alicloud_security_group_rule: fix security group egress rule delete failed [#79](https://github.com/alibaba/terraform-provider/pull/79)
data resource/alicloud_images: data alicloud_images filter all images not only the first page [#78](https://github.com/alibaba/terraform-provider/pull/78)

PR #12879, opened 2017-03-20
provider/alicloud: change create ecs postpaid instance API (#12226)

stack72:
* change create ecs postpaid instance API form createInstance to runInstances, support BusinessInfo and userdata

* update sdk vendor


stack72:
Closed via #12913

PR #12883, opened 2017-03-20
provider/aws: Deprecate the usage of Api Gateway Key Stages in favor of Usage Plans

Ninir:
## Description
Due to a change in the APIGateway API itself, API keys needs to be associated to a usage plan rather than setting `stage_keys` on the API Key itself.

This adds a warning in the documentation, plus a deprecation on the api key resource at the `stage_key`-level.

Related Documentation: http://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-api-keys.html

## Related issues
N/A


stack72:
LGTM!

PR #12891, opened 2017-03-20
nil checks when assigning to param map in resourceAwsSsmDocumentRead

dcjohnson:
Fields in the Parameters struct were being set as nil. Those fields were marked as optional in AWS anyway. To prevent a panic when the nil pointer is dereferenced I wrapped the assignments with a nil check.


stack72:
LGTM! Thanks for this :)

PR #12894, opened 2017-03-20
helper/acctest: Add NewSSHKeyPair function

jen20:
Many cloud services prevent duplicate key pairs with different names. Among them are Digital Ocean, Joyent Triton and Packet. Consequently, if tests leave dangling resources it is not enough to simply randomise the name, the entire key material must be regenerated.

This commit adds a helper method that returns a new randomly generated key pair, where the public key material is formatted in OpenSSH "authorized keys" format, and the private key material is PEM encoded.

cc @stack72

### DIFF
--- a/helper/acctest/random.go
+++ b/helper/acctest/random.go
@@ -1,8 +1,18 @@
 package acctest
 
 import (
+	"bufio"
+	"bytes"
+	crand "crypto/rand"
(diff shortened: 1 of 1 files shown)


stack72:
Watched this being made - it's going to be a great addition for our helper library

PR #12903, opened 2017-03-20
random_pet resource

jtopjian:
Just a quick, simple random name generator. If, for some reason, this isn't worthwhile to add, I'd be happy to make it a dedicated plugin. :)


stack72:
Looks like a useful addition :)

```
% make testacc TEST=./builtin/providers/random
2017/03/27 12:49:40 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/random -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccResourceID
--- PASS: TestAccResourceID (0.04s)
=== RUN   TestAccResourcePet_basic
--- PASS: TestAccResourcePet_basic (0.02s)
=== RUN   TestAccResourcePet_length
--- PASS: TestAccResourcePet_length (0.04s)
=== RUN   TestAccResourcePet_prefix
--- PASS: TestAccResourcePet_prefix (0.03s)
=== RUN   TestAccResourcePet_separator
--- PASS: TestAccResourcePet_separator (0.02s)
=== RUN   TestAccResourceShuffle
--- PASS: TestAccResourceShuffle (0.04s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/random	0.203s
```

SydOps:
@jtopjian 

If I add `count = 5`, what will happen? still get the random name for each instance?

jtopjian:
@SydOps You _should_. It should act like the other random providers which support `count`.

SydOps:
@jtopjian 

Seems not. 

```
provider "aws" {
  region = "us-west-2"
}

data "aws_ami" "ubuntu" {
  most_recent = true

  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
  }

  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }

  owners = ["099720109477"] # Canonical
}

  count = 3
  ami           = "${data.aws_ami.ubuntu.id}"
  instance_type = "t2.micro"

  tags {
    Name =  "web-server-${random_pet.server.id}"
  }
}

resource "random_pet" "server" {
  keepers = {
    # Generate a new pet name each time we switch to a new AMI id
    ami_id = "${data.aws_ami.ubuntu.id}"
  }
}

output "key" {
  value ="${random_pet.server.id}"
}
```

All ec2 instances have same name. Can we have different names for each instance? Or this is not your design.

jtopjian:
@SydOps This does indeed work with `count` like the other resources. You just need to use the `element` function when using `count` just as you would with any other reference to a multi-count resource. Check the docs on `count` and `element` :)

PR #12932, opened 2017-03-21
Updates AMI and instance_type

deiga:
The given identifiers are no longer vlaid


catsby:
Thanks for pointing this out, @deiga ! Those AMIs match to eu-west I think ,and the guide is meant for us-east. The AMI is used in some other pages as well, so I opened #13057 to continue this work. Thanks!

PR #12939, opened 2017-03-21
add `-force-copy` option to init command

### LINKED ISSUE #12921: terraform init - local to remote(s3) state, no way to force yes
### Terraform Version

terraform -v
Terraform v0.9.1

### Affected Resource(s)
- terraform init
- backend s3

main.tf

First Run to establish a local terraform.tfstate
```hcl
#terraform {
#  backend "s3" {
#    bucket = "S3bucketname"
#    key    = "s3key"
#    region = "s3region"
#  }
#}

provider "aws" {}

resource "aws_route53_zone" "testzonecom" {
  name = "testzone.com"
}

resource "aws_route53_record" "test" {
  zone_id = "${aws_route53_zone.testzonecom.zone_id}"
  name = "test"
  type = "A"
  ttl = "300"
  records = [
    "127.0.0.1"
  ]
}
```

Second run to initialize backend

```hcl
terraform {
  backend "s3" {
    bucket = "S3bucketname"
    key    = "s3key"
    region = "s3region"
  }
}

provider "aws" {}

resource "aws_route53_zone" "testzonecom" {
  name = "testzone.com"
}

resource "aws_route53_record" "test" {
  zone_id = "${aws_route53_zone.testzonecom.zone_id}"
  name = "test"
  type = "A"
  ttl = "300"
  records = [
    "127.0.0.1"
  ]
}
```

### Expected Behavior
When remotely executing terraform init there should be a flag to force yes.

### Actual Behavior
Console sits at prompt waiting for user to type yes. 

```
Initializing the backend...
Do you want to copy state from "local" to "s3"?
  Pre-existing state was found in "local" while migrating to "s3". No existing
  state was found in "s3". Do you want to copy the state from "local" to
  "s3"? Enter "yes" to copy and "no" to start with an empty state.

  Enter a value: 
```

-force flag does not work
```
terraform init -force
flag provided but not defined: -force
```

### Steps to Reproduce
1. create a configuration without a remote backend
2. run `terraform apply`
3. modify configuration to use remote backend 
4. run `terraform init`

### Workaround

```
yes yes | terraform init
```

jbardin:
The `-force-copy` option will suppress confirmation for copying state
data.

Modify some tests to use the option, making sure to leave coverage of
the Input code path.

fixes #12921

### DIFF
--- a/command/init.go
+++ b/command/init.go
@@ -21,11 +21,14 @@ type InitCommand struct {
 func (c *InitCommand) Run(args []string) int {
 	var flagBackend, flagGet bool
 	var flagConfigExtra map[string]interface{}
+
 	args = c.Meta.process(args, false)
 	cmdFlags := c.flagSet("init")
 	cmdFlags.BoolVar(&flagBackend, "backend", true, "")
 	cmdFlags.Var((*variables.FlagAny)(&flagConfigExtra), "backend-config", "")
 	cmdFlags.BoolVar(&flagGet, "get", true, "")
+	cmdFlags.BoolVar(&c.forceInitCopy, "force-copy", false, "suppress prompts about copying state data")
+
 	cmdFlags.Usage = func() { c.Ui.Error(c.Help()) }
 	if err := cmdFlags.Parse(args); err != nil {
 		return 1
@@ -224,6 +227,10 @@ Options:
                        input was required.
 
   -no-color            If specified, output won't contain any color.
+
+  -force-copy          Suppress prompts about copying state data. This is
+                       equivalent to providing a "yes" to all confirmation
+                       prompts.
 
 `
 	return strings.TrimSpace(helpText)
--- a/command/init_test.go
+++ b/command/init_test.go
@@ -270,9 +270,6 @@ func TestInit_backendUnset(t *testing.T) {
 			t.Fatalf("err: %s", err)
 		}
 
-		// Run it again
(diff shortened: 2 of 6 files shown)


command/init.go
```diff
@@ -225,6 +228,10 @@ Options:
 
   -no-color            If specified, output won't contain any color.
 
+  -force-copy          Suppress prompts about copying state data. This is
+                       equivalent to providing a "yes" to all confirmation
+					   prompts.
```
mitchellh: I think this is a stray tab char.

jbardin:
@mitchellh, thanks for the confirmation, I wasn't sure if that was the plan for the input ID and was going to look into it later ;)

ianpackard:
The documentation for init command is missing this flag.

s-pk:
Is there any way to force a no...?

nbering:
@s-pk If you use `-input=false` and a prompt is required it will error out instead of prompting.

tonglil:
@nbering but I don't want to error out, I don't want to copy.

nbering:
@tonglil I think the context of my comment may have been lost. I suspect the person I was replying to deleted their account, because their comment is missing and that name's profile comes up with a 404.

Despite that, I think what you're looking for is the `-reconfigure` flag, which ignores the existing backend configuration, so it won't prompt and will just set up the new backend ignoring the old configuration.

PR #12940, opened 2017-03-21
Adding import of AWS IAM server certificate.

mrcopper:
Adding import, as well as test and docs changes. Need to change create to use an ID that would match import for the tests, but should be inconsequential.


builtin/providers/aws/resource_aws_iam_server_certificate.go
```diff
@@ -116,15 +120,15 @@ func resourceAwsIAMServerCertificateCreate(d *schema.ResourceData, meta interfac
...
+	_, err := conn.UploadServerCertificate(createOpts)
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok {
 			return fmt.Errorf("[WARN] Error uploading server certificate, error: %s: %s", awsErr.Code(), awsErr.Message())
 		}
 		return fmt.Errorf("[WARN] Error uploading server certificate, error: %s", err)
 	}
 
-	d.SetId(*resp.ServerCertificateMetadata.ServerCertificateId)
+	d.SetId(sslCertName)
```
stack72: Hi @mrcopper 

Thanks for the work here - we can't just change the ID i'm afraid - we would need to migrate the users state otherwise everyone would show that changes are required

When we can do is to is to leave the Id as is and change the test to make sure it passes the correct value to the custom import func

The test would then look as follows:

```
func TestAccAWSIAMServerCertificate_importBasic(t *testing.T) {
 +	resourceName := "aws_iam_server_certificate.test_cert"
 +
 +		CheckDestroy: testAccCheckIAMServerCertificateDestroy,
 +			resource.TestStep{
 +				Config: testAccIAMServerCertConfig,
 +			},
 +
 +			resource.TestStep{
 +				ImportState:       true,
 +				ImportStateVerify: true,
                                ImportStateId: "the id from the config for the name",
 +					"private_key"},
 +			},
 +		},
 +	})
 +}
```
This would then allow the users to specify the name value for import

Hope this helps

Paul

mrcopper: Is there a way to get that resource ID? It looks to me like it is set by AWS and I am not sure how to retrieve it.

mrcopper: Alternatively, could I pull out the test, put the setId in Create back the way it was and resubmit?

stack72:
Hi @mrcopper 

I just added the tests to support the awesome work you have done - this way we can 100% make sure it is part of 0.9.2 release

Thanks for all your work here - I was able to preserve your commit (squashed into a single commit to keep the log in a way that makes it more logical)

https://github.com/hashicorp/terraform/pull/13065

Thanks again

Paul

mrcopper:
Thanks for all the help Paul. Glad to contribute.

PR #12945, opened 2017-03-21
kubernetes: Ignore internal K8S annotations in config_map + use PATCH

radeksimko:
This is something you'd unlikely noticed in the context of config map since config maps aren't annotated internally by default, but some other resources which I'm going to PR very soon are and I'd like to have the code in place and reviewed in isolation before adding more LOC.

### Why ignoring internal annotations

Each K8S object can have annotations which can be defined by the user, but more often are used by tools that interact w/ the K8S API - also by kubelet internally. Kubelet is maintaining annotations like these:

 - `pv.kubernetes.io/bound-by-controller = yes`
 - `pv.kubernetes.io/bind-completed = yes`

[et cetera](https://kubernetes.io/docs/api-reference/labels-annotations-taints/). [The convention and expectation](https://github.com/kubernetes/kubernetes/blob/v1.5.3/pkg/api/v1/types.go#L186-L191) from a K8S API consumer is to preserve any existing ones that it doesn't manage - therefore Terraform should not be deleting any existing annotations (which is what it does now).

#### Should Terraform have dedicated annotations prefix, e.g. `terraform.io/`?

I don't know, maybe? I'm not sure where is the right balance between allowing people to do anything, exclusively w/ Terraform VS allowing them to use Terraform alongside other tools with may be annotating resources too.

Looking at some other tools the approach really differs:

 - https://fabric8.io/guide/annotations.html (strictly separated)
 - https://www.projectcalico.org/calico-network-policy-comes-to-kubernetes/ (generic, i.e. `policy`)

Feedback welcomed. If we decide to have our own namespace, it would mean BC, but it's early enough in the life of the K8S provider, so should be pretty painless for most users.

### Why `PATCH`

While `Update()`s seem to be working fine, it's merely because kubelet is watching for any changes in the cluster and re-associates any annotations as necessary (you're able to watch how this happens in a matter of seconds via `kubectl get -w`). We do not want Terraform to stamp over other APIs or Kubelet annotations in any way.

### DIFF
--- a/builtin/providers/kubernetes/patch_operations.go
+++ b/builtin/providers/kubernetes/patch_operations.go
@@ -1,0 +1,135 @@
+package kubernetes
+
+import (
+	"encoding/json"
+	"reflect"
+	"sort"
+	"strings"
+)
+
+func diffStringMap(pathPrefix string, oldV, newV map[string]interface{}) PatchOperations {
+	ops := make([]PatchOperation, 0, 0)
+
+	pathPrefix = strings.TrimRight(pathPrefix, "/")
+
+	// This is suboptimal for adding whole new map from scratch
+	// or deleting the whole map, but it's actually intention.
+	// There may be some other map items managed outside of TF
+	// and we don't want to touch these.
+
+	for k, _ := range oldV {
+		if _, ok := newV[k]; ok {
+			continue
+		}
+		ops = append(ops, &RemoveOperation{Path: pathPrefix + "/" + k})
+	}
+
+	for k, v := range newV {
+		newValue := v.(string)
+
+		if oldValue, ok := oldV[k].(string); ok {
+			if oldValue == newValue {
+				continue
+			}
+
+			ops = append(ops, &ReplaceOperation{
+				Path:  pathPrefix + "/" + k,
+				Value: newValue,
+			})
+			continue
+		}
+
+		ops = append(ops, &AddOperation{
+			Path:  pathPrefix + "/" + k,
+			Value: newValue,
+		})
+	}
+
+	return ops
+}
+
+type PatchOperations []PatchOperation
+
+func (po PatchOperations) MarshalJSON() ([]byte, error) {
+	var v []PatchOperation = po
+	return json.Marshal(v)
+}
+
+func (po PatchOperations) Equal(ops []PatchOperation) bool {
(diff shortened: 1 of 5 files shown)


builtin/providers/kubernetes/patch_operations.go
```diff
@@ -0,0 +1,135 @@
+package kubernetes
+
+import (
+	"encoding/json"
+	"reflect"
+	"sort"
+	"strings"
+)
+
```
stack72: Nit pick - should this file really be called patch_operations? It seems there are patch and replace operation funcs within :)

builtin/providers/kubernetes/resource_kubernetes_config_map.go
```diff
@@ -73,19 +75,22 @@ func resourceKubernetesConfigMapRead(d *schema.ResourceData, meta interface{}) e
...
 
-	metadata := expandMetadata(d.Get("metadata").([]interface{}))
 	namespace, name := idParts(d.Id())
-	// This is necessary in case the name is generated
-	metadata.Name = name
 
-	cfgMap := api.ConfigMap{
-		ObjectMeta: metadata,
-		Data:       expandStringMap(d.Get("data").(map[string]interface{})),
+	ops := patchMetadata("metadata.0.", "/metadata/", d)
```
stack72: Are we guaranteed to have `metadata.0.` ?

builtin/providers/kubernetes/patch_operations.go
radeksimko: So the struct type is called `PatchOperation` and the word _patch_ comes from PATCH HTTP method, or more specifically from [JSON Patch](https://tools.ietf.org/html/rfc6902) - so I think it's the right name.

I think eventually we could use these helpers cross providers, but it depends on the exact implementation of the SDK.

builtin/providers/kubernetes/resource_kubernetes_config_map.go
radeksimko: Good question, I believe we are:

```go
&schema.Schema{
		Type:        schema.TypeList,
		Required:    true,
		MaxItems:    1,
```

stack72: In that case, LGTM

Issue #6535, opened 2016-05-08
Support (and test functionality) for AWS alternatives?


radeksimko:
It seems that more people are asking for support of AWS-like alternatives - some supported/provided directly by AWS, some not, e.g.:
- [Eucalyptus](https://github.com/eucalyptus/eucalyptus) (EC2 and probably a lot more)
- [Hologram](https://github.com/AdRoll/hologram) (Fake EC2 metadata API) - related: https://github.com/hashicorp/terraform/pull/5014
- [Local DynamoDB](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) and/or [Fake Dynamo](https://github.com/ananthakumaran/fake_dynamo)  - related: https://github.com/hashicorp/terraform/pull/6533
- [Fake SQS](https://github.com/iain/fake_sqs)
- [Fake SNS](https://github.com/yourkarma/fake_sns)
- [Fake S3](https://github.com/jubos/fake-s3)
- [EB Local](http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb3-local.html)

The obvious issue is that alternatives won't be as much battle-tested as real APIs hence some false positives/negatives may occur. If we decide to **run nightly acceptance tests against such endpoints**, these should be completely separated from current setup and fixing bugs in those should probably has lower priority than real endpoints - i.e. we don't want to break integration with real AWS API to make something work with an alternative.

I still think it's worth supporting these though because it can lower the bar for potential contributors wanting to contribute tested code as they will be able to test it against local endpoints which is not going to cost 💰 💰  in most cases. Yes, there might be false positives/negatives, but it's IMO still better than untested code.

I personally think that developers _should_ have access to a real AWS environment and be able to create real resources (which may cost money, yes - but that's about trust and potential IAM limitations and/or account limits). There will always be services which cannot be "mocked" locally - e.g. ECS, Lambda, CloudFront, EFS, RDS, ... and alternative implementations will inevitably lack behind the real official API.

I'm opening this issue to track interest and also to gather opinions from other maintainers.

RsrchBoy:
This would be nice -- for remote state, if nothing else.  e.g. I'm trying to use DreamHost Objects (S3) with the S3 remote state backend and failing miserably (and confusingly), even while other tools (e.g. s3cmd) work w/o issue.

vrenjith:
[...]

PR #12951, opened 2017-03-22
Feature/allow holo clients

### LINKED ISSUE #12704: AWSClient.partition and AWSClient.accountid not set on clients using AdRoll's hologram
### Summary

Just a quick summary (though it's not in your template). When running with EC2 credentials, [GetCredentials](https://github.com/hashicorp/terraform/blob/20cf1516a40e6a9ce2900e3157f25813bb3bb932/builtin/providers/aws/config.go#L182) is called, which recognizes the fact that I have something that looks like EC2 instance metadata creds available and injects the EC2RoleProvider into the credential chain. Then, [GetAccountInfo](https://github.com/hashicorp/terraform/blob/20cf1516a40e6a9ce2900e3157f25813bb3bb932/builtin/providers/aws/config.go#L262-L268) is called. It sees that the credential provider is the EC2RoleProvider and tries to get the instance profile ARN [out of the metadata service](https://github.com/hashicorp/terraform/blob/20cf1516a40e6a9ce2900e3157f25813bb3bb932/builtin/providers/aws/auth_helpers.go#L37) by calling `metadataClient.IAMInfo()` out of the AWS Go SDK. That method merely tries [calling the iam/info endpoint](https://github.com/aws/aws-sdk-go/blob/05de8bee0d7efef0507aeffccb4e65c2592eb174/aws/ec2metadata/api.go#L91), which is an endpoint that Hologram doesn't implement. The result is that metadataClient.IAMInfo returns an error to GetAccountInfo, which then returns the error to Client, which then silently eats the error and skips setting the accountid and partition. This can be seen in the debug logs below where the aws_partition data source is getting set to the empty string and the aws_caller_identity data source raises an error as neither are getting set.

The net result of this is that anybody using AdRoll's hologram to get instance credentials locally won't be able to use the aws_caller_identity or aws_partition data sources. We could probably deal with that, but then #11359 changed up a number of resources to depend on the *AWSClient.partition being set to dynamically generate the ARN. The one we're running in to is sns_topic_policy -- trying to destroy an SNS topic policy goes down a code path which [dynamically injects](https://github.com/hashicorp/terraform/blob/1527eae6b77ead08f36f645bea26dd0dd7dc2c2d/builtin/providers/aws/resource_aws_sns_topic_policy.go#L140) *AWSClient.partition into the ARN regex, and when the partition is empty, that regex match fails, which causes the delete to fail. In other words, for anbody who's using Hologram, they'll be able to _create_ an SNS topic policy with Terraform, but not _destroy_ it.

There's an additional subtlety/regression here. The call to GetAccountInfo is guarded by a check to see if SkipRequestingAccountId is set. If so, then the call to GetAccountInfo is skipped and *AWSClient.partition isn't set. This means that if anybody has SkipRequestingAccountId set, they also will see the same behavior -- they'll be able to create SNS topic policies but not destroy them, whereas prior to #11359 being merged, they would be able to do the same. I'd also guess that the whole point of the call to getAccountIdFromSnsTopicArn in the SNS topic provider is precisely to support the use case where SkipRequestingAccountId is set. I suspect that similar regressions will occur with the other resources touched in #11359, but sns_topic_policy is the only one I've examined in detail.

Not sure what the right path is here -- I feel like if the call to `metadataClient.IAMInfo` errors out, the code should just fall through to sts:GetCallerIdentity. However, that addresses our use case, but not the regression where setting SkipRequestingAccountId would still cause this to not be set. I believe the partition can be inferred from the region, so maybe try to infer it from the region?

### Terraform Version

$ terraform -v
Terraform v0.8.8

### Affected Resource(s)

- sns_topic_policy
- data_aws_caller_identity
- data_aws_partition

among others.

```hcl
provider "aws" {
  region = "us-east-1"
}
data "aws_partition" "current" { }
```

### Debug Output

```bash
$ TF_LOG=TRACE TF_LOG_PATH=tf.log terraform plan

data.aws_partition.current: Refreshing state...
data.aws_caller_identity.current: Refreshing state...

* data.aws_caller_identity.current: No AWS Account ID is available to the provider. Please ensure that
skip_requesting_account_id is not set on the AWS provider.
```

Can provide a full log if you want, doesn't seem necessary here. The relevant bits:

```
2017/03/14 16:12:09 [DEBUG] plugin: terraform: aws-provider (internal) 2017/03/14 16:12:09 [DEBUG] Reading Caller Identity.
2017/03/14 16:12:09 [DEBUG] plugin: terraform: aws-provider (internal) 2017/03/14 16:12:09 [DEBUG] No Account ID available, failing
2017/03/14 16:12:09 [DEBUG] plugin: terraform: aws-provider (internal) 2017/03/14 16:12:09 [DEBUG] Reading Partition.
2017/03/14 16:12:09 [DEBUG] plugin: terraform: aws-provider (internal) 2017/03/14 16:12:09 [DEBUG] Setting AWS Partition to .
2017/03/14 16:12:09 [ERROR] root: eval: *terraform.EvalReadDataApply, err: data.aws_caller_identity.current: No AWS Account ID is available to the provider. Please ensure that
skip_requesting_account_id is not set on the AWS provider.
2017/03/14 16:12:09 [ERROR] root: eval: *terraform.EvalSequence, err: data.aws_caller_identity.current: No AWS Account ID is available to the provider. Please ensure that
skip_requesting_account_id is not set on the AWS provider.
2017/03/14 16:12:09 [ERROR] root: eval: *terraform.EvalOpFilter, err: data.aws_caller_identity.current: No AWS Account ID is available to the provider. Please ensure that
skip_requesting_account_id is not set on the AWS provider.
2017/03/14 16:12:09 [ERROR] root: eval: *terraform.EvalSequence, err: data.aws_caller_identity.current: No AWS Account ID is available to the provider. Please ensure that
skip_requesting_account_id is not set on the AWS provider.
2017/03/14 16:12:09 [TRACE] [walkRefresh] Exiting eval tree: data.aws_caller_identity.current
```

### Expected Behavior
In this case, nothing, but the account ID and partition should have been available to me, and TF shouldn't have errored out.

### Actual Behavior
TF errored out

### Steps to Reproduce

1. Run `terraform apply` with the above TF on a machine running AdRoll's Hologram

### Important Factoids

Using AdRoll's hologram for credentials.

### References

- GH-11359

joelthompson:
This fixes #12704 by falling through on caller checks in case IAMInfo fails, and it refactors some of the testing code to support an added test for this condition in a cleaner fashion.

NOTE: This doesn't address potential regressions when SkipRequestingAccountId check is set, only when using something like AdRoll's Hologram.


radeksimko:
Hi @joelthompson 
I'm happy for Terraform to support alternative implementations of the API but I'm not too inclined to break any existing functionality for users that are using the real AWS APIs (most likely majority) in favour of that.

Without going too much into the implementation details (for now) - is there any specific reason Hologram can't implement the `iam/info` endpoint?

Thanks.

joelthompson:
I haven't asked them, but one simple answer is that the `iam/info` endpoint returns the arn of the instance profile, while when using hologram, there might not actually be an instance profile (just an assumed role for which you're getting credentials). Sure, an instance profile could be fabricated, but that seems like the dirtier solution than just falling through to `sts:GetCallerIdentity` as I'm proposing here.

And this shouldn't break any existing functionality for any users. It just adds a fall back in case anything existing fails.

I'm also curious why you labeled this an enhancement but the issue it fixes, #12704, was labeled (correctly, IMHO) as a bug.

copumpkin:
Can we get this merged? Not having it is causing me and a few others quite a bit of pain, since I need to compile my own Terraform with this patch

akazakov:
👍

joelthompson:
@radeksimko -- are there any plans to address this?

builtin/providers/aws/auth_helpers.go
```diff
@@ -35,29 +35,32 @@ func GetAccountInfo(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string)
...
+		if err == nil {
+			return parseAccountInfoFromArn(info.InstanceProfileArn)
+		}
+		// We can end up here if there's an issue with the instance metadata service
+		// or if we're getting credentials from AdRoll's Hologram (in which case IAMInfo will
+		// error out). In any event, if we can't get account info here, we should try
+		// the other methods available.
+		// If we have creds from something that looks like an IAM instance profile, but
+		// we were unable to retrieve account info from the instance profile, it's probably
+		// a safe assumption that we're not an IAM user
```
radeksimko: I don't think this is a safe assumption we can make. The metadata API may be unavailable or broken and we'd like to inform the user in such case rather than falling through.

joelthompson:
I've fixed the merge conflict, and I'd really appreciate a look at this, thanks! :)

builtin/providers/aws/auth_helpers.go
joelthompson: Why?

As a user who has experienced occasional errors that might or might not be attributable to the instance metadata service being flaky, I can say with confidence that I have never once said, "I'm glad that my process has died due to the instance metadata service returning an error instead of trying to recover through some alternative means." These types of errors are fairly infrequent and highly inconsistent and flaky (i.e., the worst kind to debug), so I'd always prefer something try to fall through.

Also, if you assume that AWS will never lie to you via the instance metadata service (which is a different assumption than the instance metadata service being always available), and that AWS will never let you map an instance profile to an IAM role in a different account from the instance profile, then there is NEVER a case where falling through to GetCallerIdentity is wrong. The worst case is that you'll get a different error because a node doesn't have network connectivity to the STS API endpoints (which seems _extremely_ unlikely to me to be a problem because that node will need access to AWS API endpoints to make AWS API calls as part of TF provisioning), and the real error will be as extremely hard to track down as it is today (because errors here are basically swallowed up silently). Even if you invalidate the assumptions I'm making here, I would argue that GetCallerIdentity is a better source of truth than the EC2 instance metadata service because GetCallerIdentity tells you information about the IAM principal actually making the calls, not making a number of inferences.

And in the case of the particular bug I'm trying to fix, you're not even actually informing users of the error. Instead, you're letting obscure errors like being unable to destroy an SNS topic policy propagate to the end user. So really the choice is between letting very obscure and hard-to-understand error messages make their ways to users, or to try to recover first before letting those very same error messages make their way to users. It seems pretty clear to me that the latter is the right behavior.

radeksimko:
Hi Joel,
I really appreciate the time you spent implementing the PR and then waiting for the feedback. 🙈 

I admit you're right in your arguments about near zero chance of breaking things.

Nonetheless I discussed this with another maintainer this week and we agreed that the best place to fix this is upstream (that is Hologram).

Hologram aims to imitate the real API (using their own words) and it does already expose certain fake data, like instance ID, AZ name, DNS server hostname, role name, etc.:
https://github.com/AdRoll/hologram/blob/master/agent/metadata_service.go#L94-L123

So exposing fake IAM instance role is in line with the current implementation.

In the worst case scenario the name of the instance profile could be configurable the same way [as the the role](https://github.com/AdRoll/hologram/blob/master/config/server.json#L16) and default to something sensible - e.g. role name. Per AWS docs:

> An instance profile can contain only one role, and this limit cannot be increased.

This 1-1 mapping makes you name your instance profile according to the role name that is assigned to it anyway.

> I would argue that GetCallerIdentity is a better source of truth than the EC2 instance metadata service because GetCallerIdentity tells you information about the IAM principal actually making the calls, not making a number of inferences.

That's true and the API method didn't exist when we were initially implementing this. It was carefully added later on as the last option. The expectation here however is that if we're running on EC2 we expect the metadata API to be available. If it is just flaky it's up to the upstream SDK to deal with such problem, like [it already does with many other APIs](https://github.com/aws/aws-sdk-go/blob/master/aws/request/retryer.go#L26-L52). All the API methods may fail for number of different reasons, and yes - sometimes it does make sense to just fall through to another method. This however means the current implementation is already quite complex and if we can fix this problem elsewhere, we should - merely to avoid increased complexity.

TL;DR We don't think Terraform is the right place to fix this.

joelthompson:
Hi @radeksimko -- technical arguments aside, just to be clear, I'm not trying to add a new feature. I'm trying to fix a regression introduced in #11359 and which I reported in #12704. Are you saying that you're OK with introducing regressions and breaking existing customers, then refusing a PR to fix the regressions? Because that's what's going on here, and it's a pretty poor way to treat customers.

I'd be open to other ways of fixing #12704 (e.g., I _think_ setting a default partition of `aws` if `GetAccountInfo` fails would work, but I haven't tested it, and adding special cases increases complexity); if you can suggest an alternate implementation that you'd be willing to merge, I would probably be happy to write the code for it and open a PR.

> The expectation here however is that if we're running on EC2 we expect the metadata API to be available.

Sure, but the reverse isn't true -- just because you have credentials at 169.254.169.254 doesn't necessarily mean you're running on an EC2 instance.

> This however means the current implementation is already quite complex and if we can fix this problem elsewhere, we should - merely to avoid increased complexity.

I don't think my code increases the complexity -- it keeps the actual number of lines of code the same, except in the testing where I add a test for this case. In fact, this seems like the _least_ complex way of fixing the regression in #12704.

radeksimko:
Hi @joelthompson 
I took some time to discuss this with other team members, once again - to avoid any potential biases and make sure we collectively agree on the solution of this problem.

I came to the conclusion that it's 👌  to pull this in after the discussion with a few tweaks:

 - We'd like to avoid loosing errors if things do fail - if we do _return_ an error in any stage (sts, iam, metadata API) it should contain all errors - we can use [go-multierror](https://github.com/hashicorp/go-multierror) to collect those errors.
 - To make testing code a bit more readable would you mind avoiding function with 3 bool arguments? e.g. we could expose each route as constant and then pass slice of those constants as a single argument. That way it will be clearer what endpoints are being used for any given test without looking at the implementation.

```go
awsTs := awsEnv(generateMetadataApiRoutes([]*endpoint{
    instanceIdEndpoint,
    iamInfoEndpoint,
    securityCredentialsEndpoint,
}))
```

After those 2 things are addressed I'm happy to merge it.

----

With all that said as maintainers we cannot guarantee Terraform will keep working with a 3rd party tool like Hologram in the future, we might introduce some other dependencies on the metadata API structure in some other contexts and that is why we agreed that the problem _should_ be solved upstream either way. If you want to avoid any similar disruption in the future we'd highly recommend sending PR to Hologram.

joelthompson:
Thanks @radeksimko! I'll update this PR in a couple days with your tweaks, as they seem like great additions.

As best as I understand your position on future changes, you're saying that future changes may break hologram clients because you don't explicitly test against hologram nor do you explicitly consider these clients, that you're willing to accept PRs which don't increase the code complexity to support hologram to fix previous breaks, but that you don't want to bend over backwards to support hologram. If so, that seems completely reasonable to me, thanks!

I'd also like to add that none of what we've discussed fixes the symptoms of the issue where `skip_requesting_account_id` is set. I believe that's still a regression, but one I'd be happy to work with you on fixing :)

radeksimko:
> future changes may break hologram clients because you don't explicitly test against hologram nor do you explicitly consider these clients, that you're willing to accept PRs which don't increase the code complexity to support hologram to fix previous breaks, but that you don't want to bend over backwards to support hologram

Yes, you're right in all points here. Just for completeness we did discuss official support & testing in the past: https://github.com/hashicorp/terraform/issues/6535

joelthompson:
OK, I think I've addressed your feedback, thanks!

joelthompson:
Looks like I need to move to the terraform-providers/terraform-provider-aws repo, so I've opened a new PR in terraform-providers/terraform-provider-aws#17 Please confirm that's the right place to continue this discussion, and then feel free to close this.

PR #12955, opened 2017-03-22
provider/kubernetes: ConfigMap metadata name did not comply to the regex rules

mrooding:
According to the [Kubernetes documentation](https://kubernetes.io/docs/user-guide/identifiers/#names) the name cannot contain an underscore. I've updated the ConfigMap documentation to comply to the naming regex.


radeksimko:
Thanks 👍

PR #8692, opened 2016-09-06
New users should know it's possible to save credentials outside of tf…

joeshockman:
… configs

Terraform will automatically search for AWS API credentials or Instance Profile Credentials. I wish I'd known that when I first read these docs.
Saving credentials outside of tf config files is a much better plan for situations where config files end up in source control and or where multiple people collaborate. Making this information available early will allow new users to set up a much more secure and robust plan for deploying terraform at scale and in production environments.

PR #12968, opened 2017-03-22
New users should know it's possible to save credentials outside of tf…

radeksimko:
Closes #8692

Thanks @joeshockman

cc @cblecker


cblecker:
LGTM!

PR #12970, opened 2017-03-22
Add name to aws_iam_user doc attributes reference

phoolish:
This seems to be missing from the list of attributes you can reference.


radeksimko:
Hi @phoolish 
we generally take it as given that all arguments are also exposed as attributes. An extensive list would make a very long list for some resources.

I can see that `name` is something people would look for often in this context, so I'm happy to merge this to make it clearer 👍

PR #12982, opened 2017-03-22
alicloud: simplify validators

josephholsten:


builtin/providers/alicloud/validators.go
```diff
@@ -2,26 +2,17 @@ package alicloud
...
 func validateInstancePort(v interface{}, k string) (ws []string, errors []error) {
-	value := v.(int)
-	if value < 1 || value > 65535 {
-		errors = append(errors, fmt.Errorf(
-			"%q must be a valid instance port between 1 and 65535",
-			k))
-		return
-	}
-	return
+	return validation.IntBetween(1, 65535)(v, k)
```
pmcatominey: Is there a reason the validation functions are wrapped rather than being used directly in the schema? Seems to create a lot of redundant functions.

josephholsten: mostly because there are a ton of tests, and I wanted to be certain that all of these worked correctly before inlining.

speaking of which, any thoughts on how to test the validation functions after I inline them? seems overkill to actually create full HCL test data for each variant.

pmcatominey: Makes sense.

Where the validation functions are used in the azurerm provider there are no tests, we trust that the validation package itself is tested properly.

PR #12988, opened 2017-03-22
test changes to travis

jbardin:
Now that we've found the test busy-loops, travis should be able to run
them again in a single batch.

Increase timeout to 60s. The consul backend tests come too close to that
and timout every time they stall a little.


mitchellh:
The issue that caused this was actually RAM issues: we didn't have enough RAM to compile all the tests at once.

jbardin:
LOL, there it goes and fails

PR #12992, opened 2017-03-23
provider/aws: Update IAM Role Policy Attachment Acctests

grubernaut:
Leaked resources may prevent this resource from correctly passing acceptance tests. Seeding the policy names with random integer suffixes allows tests to pass regardless of resource leaks.

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSRolePolicyAttachment_basic'
2017/03/22 19:58:58 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSRolePolicyAttachment_basic -timeout 120m
=== RUN   TestAccAWSRolePolicyAttachment_basic
--- PASS: TestAccAWSRolePolicyAttachment_basic (31.98s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    31.989s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_iam_role_policy_attachment_test.go
+++ b/builtin/providers/aws/resource_aws_iam_role_policy_attachment_test.go
@@ -15,9 +15,9 @@ import (
 func TestAccAWSRolePolicyAttachment_basic(t *testing.T) {
 	var out iam.ListAttachedRolePoliciesOutput
 	rInt := acctest.RandInt()
-	testPolicy := fmt.Sprintf("test-policy-%d", rInt)
-	testPolicy2 := fmt.Sprintf("test-policy2-%d", rInt)
-	testPolicy3 := fmt.Sprintf("test-policy3-%d", rInt)
+	testPolicy := fmt.Sprintf("tf-acctest-%d", rInt)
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_iam_role_policy_attachment_test.go
```diff
@@ -7,30 +7,35 @@ import (
...
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 )
 
 func TestAccAWSRolePolicyAttachment_basic(t *testing.T) {
 	var out iam.ListAttachedRolePoliciesOutput
+	rInt := acctest.RandInt()
+	testPolicy := fmt.Sprintf("test-policy-%d", rInt)
+	testPolicy2 := fmt.Sprintf("test-policy2-%d", rInt)
+	testPolicy3 := fmt.Sprintf("test-policy3-%d", rInt)
```
radeksimko: Would you mind changing this to something like `tf-acc-test-%d` so it's more obvious - in case of a leak - that those resources come from an acceptance test?

Issue #12808, opened 2017-03-17
provider/circonus: Change check fields from TypeSet to TypeList


radeksimko:
The main reason for this change is the following test failure:

```
=== RUN   TestAccCirconusCheckCloudWatch_basic
--- FAIL: TestAccCirconusCheckCloudWatch_basic (3.22s)
    testing.go:268: Step 0 error: Check failed: Check 5/191 error: circonus_check.rds_metrics: Attribute 'cloudwatch.1637274235.dimmensions.%' not found
```

but generally speaking if there's `TypeSet` which is limited to max. 1 item there's (almost) never a reason to keep it as set as there's nothin to sort - no benefit from the hash.

Such change will require state migration also, because some users may already have existing resources in the state under a hash and we'll need to migrate to the new schema (i.e. TypeList, idx = `0`).

cc @sean-

sean-:
@radeksimko I'm working on fixing this test so that it passes testing in TC.  Yesterday we added this to TC, fixing the TC checks is next.

FWIW, the reason for the `TypeSet` here is schema enforcement on arbitrary attributes.  I didn't use a list because the key/value pairs shouldn't have a set order, therefore a `TypeSet` was used.

radeksimko:
@sean- I believe you can enforce schema in `TypeList` too - it allows you to specify `Elem` exactly the same way as `TypeSet` does. The only difference is that you have numerical zero-based indexes, instead of `Set` function calculating the hash/index.

See [this example](https://github.com/hashicorp/terraform/blob/3ac05a539bc4d02b1925e0704d0714d2e8723bcc/builtin/providers/aws/data_source_aws_alb_listener.go#L39-L54).

> because the key/value pairs shouldn't have a set order

that would be the right decision under normal circumstances - all the reasoning you mentioned makes sense, but if there's ever going to be max. of 1 item, there is nothing to sort and therefore no reason to deal with ordering.

PR #12994, opened 2017-03-23
Circonus testing maintenance

sean-:
Consolidate all testing tags to `lifecycle:unittest`.

No functional change, just lots of CRC updates in the tests.

(`source:unittest` || `lifecycle:unittests`) -> `lifecycle:unittest`


builtin/providers/circonus/resource_circonus_check_cloudwatch_test.go
```diff
@@ -23,192 +23,192 @@ func TestAccCirconusCheckCloudWatch_basic(t *testing.T) {
...
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.metric.3550163941", "TransactionLogsDiskUsage"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.metric.2231806695", "TransactionLogsGeneration"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.metric.335777904", "WriteIOPS"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.metric.3894876280", "WriteLatency"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.metric.1569904650", "WriteThroughput"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.namespace", "AWS/RDS"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.version", "2010-08-01"),
-					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.1637274235.url", "https://monitoring.us-east-1.amazonaws.com"),
+					resource.TestCheckResourceAttr("circonus_check.rds_metrics", "cloudwatch.88938937.dimmensions.%", "1"),
```
radeksimko: If I understand correctly this is making it pass with the AWS credentials with have in TeamCity, but isn't entirely solving https://github.com/hashicorp/terraform/issues/12808 ? Not suggesting it should be fixed in this PR, just checking I understand the motivation behind the hash change 😃

builtin/providers/circonus/resource_circonus_check_httptrap_test.go
```diff
@@ -86,7 +86,7 @@ resource "circonus_check" "consul" {
   period = "60s"
 
   collector {
-    id = "/broker/1286"
+    id = "/broker/2110"
```
radeksimko: Probably unrelated to this PR, but where does this path and `consul-server-10-151-2-8` come from and what's the significance of these? i.e. would Circonus accept any hostname & path? 🤔

builtin/providers/circonus/resource_circonus_check_cloudwatch_test.go
sean-: Correct.  I'm going to make a sweep for #12808 soon, but yeah, this PR isn't addressing that.  This PR is just chasing some tag drift that I noticed from the tests and chasing the checksum drift as a result.

re: #12808 I didn't realize until you brought it up that there was a special case for `TypeList` with a `MaxItems=1`, but yeah, I'm definitely going to make that conversion.  Chasing these checksums is tedious.  ;)

builtin/providers/circonus/resource_circonus_check_httptrap_test.go
sean-: This collector `id` is specific to the TF test account.  This isn't actually a "path", it's what's called the `CID` and is actually a unique key that happens to look like a URI and can be injected directly into future API calls.  Which is to say, that CID belongs to the collector that's dedicated to the TF test account.

radeksimko: I see, thanks for the explanation, it is - as I said - unrelated to this PR, but I think it would be worthwhile turning this into ENV variable at some point, so that the tests can run with any account.

PR #13023, opened 2017-03-23
NS1: Fix zone value in example

nnalam:
As experienced during testing, the value of the zone needs to be the zone name, and not the NS1 internal ID.

@pashap


stack72:
LGTM! Thanks for this @nnalam :)

Paul

pashap:
thanks @nnalam

Issue #12842, opened 2017-03-18
Terraform 0.9.x "forgot" about OpsWorks resources


rafaelmagu:
Hi there,

I'd like to report an issue with the upgrade to Terraform 0.9.0. It seems to be ignoring the existing OpsWorks resources I have in my state files, and it is attempting to re-create them.

### Terraform Version
```
Terraform v0.9.0
```

### Affected Resource(s)
- aws_opsworks_stack
- aws_opsworks_nodejs_app_layer
- aws_opsworks_instance
- aws_opsworks_application

### Terraform Configuration Files
```hcl
resource "aws_opsworks_stack" "admin" {
  name                          = "admin"
  region                        = "${var.region}"
  service_role_arn              = "${var.iam_role}"
  default_instance_profile_arn  = "${var.iam_profile}"
  vpc_id                        = "${var.vpc}"
  default_subnet_id             = "${var.subnet_a}"
  default_os                    = "Ubuntu 14.04 LTS"
  configuration_manager_version = "11.10"
  default_ssh_key_name          = "${var.key_name}"
  hostname_theme                = "Wild_Cats"
}
```

### Expected Behavior
After running `terraform init` to migrate the remote state from 0.8.x to 0.9.x, it correctly imported all the resources, including the ones above. A subsequent `terraform plan` should have yielded no plan changes, as `terraform state list` shows the existing resources.

### Actual Behavior
A `terraform plan` offered to re-create all OpsWorks resources. Furthermore, `terraform import` did not work, stating imported resources would collide with existing ones.

### Steps to Reproduce
1. `terraform init`
1. `terraform plan`

### Important Factoids
Replicated across three different AWS accounts, using three different environments (with shared modules).

rafaelmagu:
```
$ terraform import module.admin.aws_opsworks_stack.admin long-id-goes-here
module.admin.aws_opsworks_stack.admin: Importing from ID "long-id-goes-here"...
module.admin.aws_opsworks_stack.admin: Import complete!
  Imported aws_opsworks_stack (ID: long-id-goes-here)
Error importing: 1 error(s) occurred:

* module.admin.aws_opsworks_stack.admin (import id: long-id-goes-here): Can't import module.admin.aws_opsworks_stack.admin, would collide with an existing resource.

Please remove or rename this resource before continuing.
```

rafaelmagu:
Issue verified to exist in 0.9.1 too.

catsby:
[...]

PR #12688, opened 2017-03-14
provider/aws: Remove restriction on running aws_opsworks_* on us-east-1

stack72:
Fixes: #11824
Fixes: #9826

A full set of acceptance tests for *all* OpsWorks tests are below. There is a single issue, opsworks_instance isn't ordering the layer_ids so it's non-deterministic which can come back. This introduces a false positive.

The opsworks tests are running both against us-east-1 for some and us-west-2 for others. This must be marked as a backwards incompatibility in 0.9 just incase people are using the opsworks region parameter to try and override the provider region

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSOpsworksApplication'                                                              2 ↵ ✹
==> Checking that code complies with gofmt requirements...
go generate $(go list ./... | grep -v /terraform/vendor/)
2017/03/14 17:11:36 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSOpsworksApplication -timeout 120m
=== RUN   TestAccAWSOpsworksApplication
--- PASS: TestAccAWSOpsworksApplication (164.29s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	164.319s
```

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSOpsworksCustomLayer'                                                                  ✹
==> Checking that code complies with gofmt requirements...
go generate $(go list ./... | grep -v /terraform/vendor/)
2017/03/14 17:16:18 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSOpsworksCustomLayer -timeout 120m
=== RUN   TestAccAWSOpsworksCustomLayerImportBasic
--- PASS: TestAccAWSOpsworksCustomLayerImportBasic (105.98s)
=== RUN   TestAccAWSOpsworksCustomLayer
--- PASS: TestAccAWSOpsworksCustomLayer (92.68s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	198.684s
```

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSOpsworksPermission'                                                                   ✹
==> Checking that code complies with gofmt requirements...
go generate $(go list ./... | grep -v /terraform/vendor/)
2017/03/14 17:23:28 Generated command/internal_plugin_list.go
[...]

PR #13024, opened 2017-03-23
provider/aws: Fix issue upgrading to Terraform v0.9+ with AWS OpsWorks Stacks

catsby:
Fix an issue when upgrading from Terraform < 0.9 to 0.9+, when we added
support for the regional endpoints in OpsWorks Stacks. OpsWorks Stacks
can only be managed via the endpoint with which they were created, not
where the stack resides:

    Stacks can only be accessed or managed within the endpoint in which they are created

- http://docs.aws.amazon.com/opsworks/latest/APIReference/Welcome.html

Prior to https://github.com/hashicorp/terraform/pull/12688, all OpsWorks stacks were created via `us-east-1` endpoint, which was the only way to create them until very recent. Amazon has since added regional endpoints, and deprecated the approach of all-stacks-go-thru-us-east.

When we removed the default of `us-east-1` in Terraform v0.9.0, any `opsworks_stack` resources created in a version of Terraform < 0.9  began querying the API with whatever region the Provider had been configured, not necessarily `us-east-1`, the region they were created with. 

Any user in this situation would be presented with Terraform trying to create a new opsworks stack, because it was unable to find it due to the region diff. See https://github.com/hashicorp/terraform/issues/12842 .

Here, we add a `stack_endpoint` computed attribute, where we'll store the region in which the stack is found. On the next UPDATE or REFRESH action, Terraform will attempt to find the stack in the default Provider region. If not found, it will attempt to find it in `us-east-1`. If it's found in east, we'll save that as the `stack_endpoint`. 

Future UPDATE, REFRESH, and DELETE operations will check this computed attribute and get the connection needed, if the `stack_endpoint` is found.

The acceptance test `TestAccAWSOpsWorksStack_classic_endpoints` should cover this. It first creates a stack in `us-east-1`, using us-east-1 region. In the second step, we switch to `us-west-2` and preform a plan-only check of the configuration. Without the ability to check `us-east-1` after checking the updated Provider region, this test step would fail to find the stack created and offer to create it, failing the step because it produces a non-empty plan.

### DIFF
--- a/builtin/providers/aws/resource_aws_opsworks_stack.go
+++ b/builtin/providers/aws/resource_aws_opsworks_stack.go
@@ -3,14 +3,17 @@ package aws
 import (
 	"fmt"
 	"log"
+	"os"
 	"strings"
 	"time"
 
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/opsworks"
 )
 
@@ -183,6 +186,11 @@ func resourceAwsOpsworksStack() *schema.Resource {
 				Computed: true,
 				Optional: true,
 			},
+
+			"stack_endpoint": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -254,6 +262,13 @@ func resourceAwsOpsworksSetStackCustomCookbooksSource(d *schema.ResourceData, v 
 
 func resourceAwsOpsworksStackRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*AWSClient).opsworksconn
+	var conErr error
+	if v := d.Get("stack_endpoint").(string); v != "" {
+		client, conErr = opsworksConnForRegion(v, meta)
+		if conErr != nil {
+			return conErr
+		}
+	}
 
 	req := &opsworks.DescribeStacksInput{
 		StackIds: []*string{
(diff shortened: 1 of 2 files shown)


stack72:
LGTM! Thanks for the fix :)

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSOpsWorksStack_'                ✭
2017/03/28 12:24:47 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSOpsWorksStack_ -timeout 120m
=== RUN   TestAccAWSOpsWorksStack_classic_endpoints
--- PASS: TestAccAWSOpsWorksStack_classic_endpoints (45.10s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	45.121s```

GuiSim:
I'm facing this issue right now! Thanks for the fix.
How can I know which version of Terraform has this fix? I'm running on 0.9.2.

If this fix is in 0.9.2, I might have a closely related issue that affects module.phobos.aws_opsworks_custom_layer and aws_opsworks_instance. It appears that Terraform can't see layers and instances created with classic endpoint using a version of Terraform older than 0.9. It will attempt to recreate them and fail.

Let me know if you'd like me to fill a new issue or if this is supposed to be fixed by this @catsby !

Thanks!

catsby:
This fix should apply to `opsworks_stacks` only. In my research that was the only resource that had this limitation, according to the docs I read:

> Stacks can only be accessed or managed within the endpoint in which they are created

It's entirely possible things that belong to this stack also have that limitation? If you discover this is the case, please open a new GitHub issue.

Thanks!

rafaelmagu:
For the record, this does not fix the problem with existing stacks. It continues to think my stack doesn't exist, but won't allow me to import saying that the stack already exists.

GuiSim:
Judging by @rafaelmagu 's last comment, I think that this bug is not resolved and should be reopened.

nabeken:
@catsby I think `Stacks` is just used to clarify since they have another `AWS OpsWorks for Chef Automate` so `Stacks` does not mean a stack in OpsWorks Stacks.

Actually, **All** API requests should go to regional endpoint, which means all TF opsworks resources should contact to it.

nabeken:
@catsby Filed an issue https://github.com/hashicorp/terraform/issues/13483

PR #13045, opened 2017-03-24
provider/scaleway: work around parallel request limitation

nicolai86:
according to the official scaleway support, requests within the same session, as identified by the token, can not be parallelized.

While I do not know for sure that this is a write-only limitation, I've
implemented it as a write-only limitation for now: all `Create` and `Destroy` actions are now serialized for the Scaleway provider.

Previously scaleway resources created in parallel would produce a 500 internal server error, e.g.:

```
resource "scaleway_ip" "test_ip" {
  count = 2
}
```

this no longer happens, neither on create nor on destroy

### DIFF
--- a/builtin/providers/scaleway/provider.go
+++ b/builtin/providers/scaleway/provider.go
@@ -1,10 +1,14 @@
 package scaleway
 
 import (
+	"sync"
+
 	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
+
+var mu = sync.Mutex{}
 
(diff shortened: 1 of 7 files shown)


nicolai86:
Still waiting on the test results…

nicolai86:
ping @stack72 ready for your review, sir 👍

stack72:
Hi @nicolai86 

AFAICT, this uses a Mutex in a normal way except for the IP which we need to explicitly unlock. is this correct?

Paul

stack72:
LGTM :)

nicolai86:
@stack72 correct. There was one additional area where I need to unlock manually b/c the create action relies on the Update action, but yes.

nicolai86:
❤️ @stack72

Issue #12892, opened 2017-03-20
Interpolation error when referencing multiple sources


srhaber:
Interpolation seems to fail when referencing values from different types of sources.  

In the example below, the ingress block contains: 
* Hard-coded value
* Interpolated variable
* Interpolated resource attribute  

When both interpolations are present, the Terraform plan fails.  If either one is removed, the plan succeeds.

This bug was introduced in version 0.9.1.

### Terraform Version
0.9.1

### Affected Resource(s)
- aws_security_group

This example use `aws_security_group` to demonstrate this issue.  However, it's very possible this is a core bug.

### Terraform Configuration Files
```hcl
variable "cidr_block" {
  default = "172.10.0.0/16"
}

resource "aws_eip" "eip" { }

resource "aws_security_group" "allow_all" {
  name        = "allow_all"
  description = "Allow all inbound traffic"

  ingress {
    from_port   = 0
    to_port     = 0
    protocol    = "-1"
    cidr_blocks = [
      "192.168.0.0/16",
      "${var.cidr_block}",
      "${aws_eip.eip.public_ip}/32"
    ]
  }
}
```

### Debug Output
https://gist.github.com/srhaber/0c356008b1b782466a9e6e4368d5e312

### Expected Behavior

Output from `terraform plan` using 0.9.0:

```
+ aws_eip.eip
    allocation_id:     "<computed>"
    association_id:    "<computed>"
    domain:            "<computed>"
    instance:          "<computed>"
    network_interface: "<computed>"
    private_ip:        "<computed>"
    public_ip:         "<computed>"
    vpc:               "<computed>"

+ aws_security_group.allow_all
    description:                            "Allow all inbound traffic"
    egress.#:                               "1"
    egress.482069346.cidr_blocks.#:         "1"
    egress.482069346.cidr_blocks.0:         "0.0.0.0/0"
    egress.482069346.from_port:             "0"
    egress.482069346.ipv6_cidr_blocks.#:    "0"
    egress.482069346.prefix_list_ids.#:     "0"
    egress.482069346.protocol:              "-1"
    egress.482069346.security_groups.#:     "0"
    egress.482069346.self:                  "false"
    egress.482069346.to_port:               "0"
    ingress.#:                              "1"
    ingress.~2281976595.cidr_blocks.#:      "<computed>"
    ingress.~2281976595.from_port:          "0"
    ingress.~2281976595.ipv6_cidr_blocks.#: "0"
[...]

PR #13046, opened 2017-03-24
reify the list values before validation

jbardin:
If the list was marked as computed, all values will be raw config
values. Fetch the individual keys from the config to get any known
values before validating.

#12892

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -1219,6 +1219,13 @@ func (m schemaMap) validateList(
 	for i, raw := range raws {
 		key := fmt.Sprintf("%s.%d", k, i)
 
+		// Reify the key value from the ResourceConfig.
+		// If the list was computed we have all raw values, but some of these
+		// may be known in the config, and aren't individually marked as Computed.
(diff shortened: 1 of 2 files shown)


jbardin:
Hmm, that map in the test fixture schema looks little wonky (it was just there for good measure, not actually part of the bug). I'll fix that up.

jbardin:
computed primitive values are skipped completely, so ValidateFunc should never have to deal with the raw interpolation strings or know about UnknownValue.

mitchellh:
@jbardin An issue we ran into early then was the exists/not-exists check. For example: if an entire list is computed, will helper/schema properly see that some value may exist there in the future (and therefore pass validation)?

jbardin:
@mitchellh, In this case the list is marked as computed (which is why we had all raw values), which would skip ValidateFunc for the list if it had one (I confirmed that). The individual elements are still validated independently. 

If you're referring to Resource.ExistsFunc, that's called from a different code path

mitchellh:
That works great.

Yeah not ExistsFunc that is to check if a resource exists via state (so never computed values) I think.

PR #13053, opened 2017-03-24
Handling missing resources

jasminSPC:
Implemented handling of status code 404


stack72:
LGTM! Thanks for this - the tests that are failing are for different reasons....

PR #13061, opened 2017-03-25
provider/openstack: Adding all_metadata attribute

jtopjian:
This commit adds the all_metadata attribute which contains all
instance metadata returned from the OpenStack Compute API. This
is useful for gaining access to metadata which was applied
out of band of Terraform.

This commit also stops setting the original metadata attribute,
effectively making metadata an input and all_metadata the output
to reference all data from.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_instance_v2.go
@@ -335,6 +335,10 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"all_metadata": &schema.Schema{
+				Type:     schema.TypeMap,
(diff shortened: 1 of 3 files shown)


builtin/providers/openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -554,7 +558,7 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
 		})
 	}
 
-	d.Set("metadata", server.Metadata)
+	d.Set("all_metadata", server.Metadata)
```
stack72: is this just a string?

jtopjian: `server.Metadata` is defined as `map[string]string` in Gophercloud.

PR #13062, opened 2017-03-25
provider/openstack: Deprecating Instance Volume attribute

jtopjian:
This commit deprecates the volume attribute in the
openstack_compute_instance_v2 resource. It's recommended to either
use the block_device attribute or the openstack_compute_volume_attach_v2
resource.

I've been planning to do this for quite a while now. There are a few good reasons for doing this:

1. The `volume` attribute requires a BlockStorage client to be mixed in with the Compute resource. Right now it's hard-coded for the v1 API.
2. `block_device` provides a wealth of volume attachment capabilities and strictly uses the Compute API.
3. The `openstack_compute_volume_attach_v2` resource has been around for a while now and has been reported to work as it should. It provides the flexibility and decoupled design that `volume` is missing.


stack72:
LGTM! :) Thanks

PR #13063, opened 2017-03-25
provider/openstack: Decprecating Instance Floating IP attribute

jtopjian:
This commit deprecates the floating_ip attributes from the
openstack_compute_instance_v2 resource. It is recommended to use
either the openstack_compute_floatingip_associate resource or
configure an openstack_networking_port_v2 resource with a floating
IP.

Another deprecation that I've had planned for a while. The biggest benefit here is that the eventual removal of this feature will clean up _a lot_ of code in the instance resource. 

The two recommended alternatives provide much better flexibility with floating IPs.


stack72:
LGTM! Thanks for this :)

PR #13066, opened 2017-03-25
terraform output adds commas when printing lists

### LINKED ISSUE #13025: terraform output adds commas when printing lists
### Issue
terraform is adding a comma `,` after list items displayed in the normal `terraform output` or scoped `terraform output my_list` commands.

the `-json` view adds a comma as well, but is expected

### Terraform Version
0.9.1

### Affected Resource(s)
outputs, list-type

```hcl
output "my_list" {
  value = [
    "item_one",
    "item_two"
  ]
}
```

### Expected Behavior
What should have happened?

```
$ terraform apply

Outputs:

my_list = [
    item_one
    item_two
]
```
```
$ terraform output my_list
item_one
item_two
```

note that json is correct, as comma is required to be valid json
```
$ terraform output -json my_list
{
    "sensitive": false,
    "type": "list",
    "value": [
        "item_one",
        "item_two"
    ]
}
```

### Actual Behavior
What actually happened?

note the commas present in the normal and scoped outputs
```
$ terraform apply

Outputs:

my_list = [
    item_one,
    item_two
]
```
```
$ terraform output my_list
item_one,
item_two
```
```
$ terraform output -json my_list
{
    "sensitive": false,
    "type": "list",
    "value": [
        "item_one",
        "item_two"
    ]
}
```

### Steps to Reproduce
1. `terraform apply`
2. `terraform output`
3. `terraform output my_list`
4. `terraform output -json my_list`

Since `terraform output <scoped>` is meant to be machine readable specifically with something like bash, adding the extra commas is bad

snowsky:
Fixes #13025

### DIFF
--- a/command/output.go
+++ b/command/output.go
@@ -172,9 +172,7 @@ func formatListOutput(indent, outputName string, outputList []interface{}) strin
 		keyIndent = "    "
 	}
 
-	lastIdx := len(outputList) - 1
-
-	for i, value := range outputList {
+	for _, value := range outputList {
 		switch typedValue := value.(type) {
 		case string:
 			outputBuf.WriteString(fmt.Sprintf("\n%s%s%s", indent, keyIndent, value))
@@ -184,10 +182,6 @@ func formatListOutput(indent, outputName string, outputList []interface{}) strin
 		case map[string]interface{}:
 			outputBuf.WriteString(fmt.Sprintf("\n%s%s", indent,
 				formatNestedMap(indent+keyIndent, typedValue)))
-		}
-
-		if lastIdx != i {
-			outputBuf.WriteString(",")
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


teamterraform:
Hi all! Sorry for the long silence here.

The formatting of these has changed in Terraform 0.12, but in a different way than discussed above. The intent of the default output presentation was to resemble how the value would be written in the Terraform language, although in earlier versions that was not true for a number of reasons, including that any nested strings were not given in quotes.

The change for Terraform 0.12 was to more faithfully produce a Terraform-language-like output format, which includes quoted strings. But in line with that goal, it still produces trailing commas on list elements because that is required in the Terraform language syntax. It does not produce them for map elements because the Terraform language does not require commas in that situation.

Consistency with the input language is the main design goal here, so we don't intend to remove the commas.

Issue #12495, opened 2017-03-07
Crash and remove resources on updating openstack_lb_pool_v1 with new monitor


carlpett:
I had accidentally used the wrong monitor for my pool when creating it. I changed the `monitor_ids` property and ran plan. Plan reports that the value will be changed in-place. Run apply. Terraform crashes, and the pool is lost from the state in the process (still exists, but Terraform does not know about it). Running apply again will re-create all members and a new pool.

### Terraform Version
Terraform v0.8.6

### Affected Resource(s)
- openstack_lb_pool_v1 

### Panic Output
https://gist.github.com/carlpett/e79e0e570a0ee1edcd5e335d5d7eaa5e

### Expected Behavior
Monitor is changed.

### Actual Behavior
Crash, existing pool is lost.

### Steps to Reproduce
Create a pool with a monitor. Apply. Change to another pool. Apply.

sebastien-prudhomme:
I don't really know if it's related: got a crash on "terraform plan" with v0.9.0-beta2 when creating a new openstack_lb_monitor_v2 ressource (not a openstack_lb_pool_v1)

```hcl
resource "openstack_lb_monitor_v2" "test" {
  type           = "HTTP"
  delay          = 20
  timeout        = 10
  max_retries    = 3
  expected_codes = 200
  url_path       = "/v1/status/leader"
  pool_id        = "bf2ab482-42fd-4741-b301-723601169585"
}
```
crash.log:  https://gist.github.com/sebastien-prudhomme/6604df7c75a88b43ba47f0f2eacebd15

jtopjian:
@carlpett Sorry for the delay on this one. I actually had this fixed soon after you reported it but must have got side tracked and never opened a PR for it. #13069 will resolve this issue.

@sebastien-prudhomme Your crash looks to have been a bug with the new "timeout" feature in the beta2 release. I'm able to reproduce your crash with beta2 and can confirm that any release after beta2 is fine. Let me know if you have any questions though.

Thank you both for reporting the issues!

carlpett:
@jtopjian Nice work!

stack72:
Closed via #13069

PR #13069, opened 2017-03-25
provider/openstack: Fix monitor_id typo in LBaaS v1 Pool

jtopjian:
For #12495


stack72:
LGTM! Thanks for this :)

PR #10875, opened 2016-12-21
provider/openstack: 409 Response on Member Create

jtopjian:
This commit accounts for a 409 response when a LBaaS v2 member is
being created. Rather than error out, this should be considered a
pending state.

Fixes #10852

PR #13074, opened 2017-03-26
provider/openstack: 409/500 Response on Pool Create

jtopjian:
This commit accounts for a 409 or 500 response when a LBaaS v2 pool is
being created. Rather than error out, this should be considered a
pending state.

Just hit this by random chance. The acceptance test would fail once every 10 times or so. Pretty much the exact same as #10875.


stack72:
LGTM :)

PR #13075, opened 2017-03-26
provider/openstack: Don't log the catalog

jtopjian:
The OS_DEBUG feature has worked out great, but frequent logging of
the service catalog during client initialization can make logging
very chatty. This commit omits the service catalog in the logs.


stack72:
LGTM! Thanks for this :)

PR #13080, opened 2017-03-27
Add `basename` and `dirname` functions

joshuaspence:
Adds `basename` and `dirname` interpolation. I want to add a `stack` tag to our infrastructure, the value of which is set to `${basename(path.cwd)}`. We currently use `${replace(path.cwd, "/^.+\\//", "")}` instead, but this is extremeley unreadable. The existance of a `basename` function would be very useful for this use case.

I don't have an immediate use case for a `dirname` function, but it seemed reasonable to add it as well.

### DIFF
--- a/config/interpolate_funcs.go
+++ b/config/interpolate_funcs.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


ewbankkit:
Use of the [`path`](https://golang.org/pkg/path/) package or [`filepath`](https://golang.org/pkg/path/filepath/) package? It seems like **filepath** is correct here.

apparentlymart:
Agreed with @ewbankkit that it probably makes sense to use `filepath` here, for correct portable behavior for Terraform on all platforms -- particularly on Windows.

Fortunately the two are API compatible to the extend that matters here. Would you mind switching that, @joshuaspence, or do you think there's a good reason to use the generic `path` package?

joshuaspence:
Yep done. I'm not really a golang programmer, so I didn't know about `filepath`.

PR #13086, opened 2017-03-27
Rancher: error when no api_url is provided

raphink:
Since `api_url` is optional (which is required to get `config` to work), it is possible to end up with a provider that has no API URL provided, which should error.

### DIFF
--- a/builtin/providers/rancher/provider.go
+++ b/builtin/providers/rancher/provider.go
@@ -2,6 +2,7 @@ package rancher
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/url"
 	"os"
@@ -87,7 +88,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 			return config, err
 		}
 
-		if apiURL == "" {
+		if apiURL == "" && config.URL != "" {
 			u, err := url.Parse(config.URL)
 			if err != nil {
 				return config, err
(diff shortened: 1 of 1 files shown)


stack72:
Hi @raphink 

Thanks for the work here - the build fails with the following errors:

```
2017/03/27 11:47:52 Generated command/internal_plugin_list.go
go test -i $(go list ./... | grep -v '/terraform/vendor/' | grep -v '/builtin/bins/') || exit 1
# github.com/hashicorp/terraform/builtin/providers/rancher
builtin/providers/rancher/provider.go:91: config.URL undefined (type interface {} is interface with no methods)
builtin/providers/rancher/provider.go:92: config.URL undefined (type interface {} is interface with no methods)
builtin/providers/rancher/provider.go:100: config.AccessKey undefined (type interface {} is interface with no methods)
builtin/providers/rancher/provider.go:104: config.SecretKey undefined (type interface {} is interface with no methods)
builtin/providers/rancher/provider.go:118: no new variables on left side of :=
builtin/providers/rancher/provider.go:118: config.GlobalClient undefined (type interface {} is interface with no methods)
make: *** [test] Error 1

```

raphink:
My bad, just fixed it @stack72. The build still fails, but it's not my code :wink:

raphink:
There, rebased and now Travis is :green_heart:

stack72:
LGTM!

Issue #13084, opened 2017-03-27
Add ARN and UserID to AWS_CALLER_IDENTITY data source


eredi93:
I need to get the arn of the current user to add it in a AWS policy.
AWS_CALLER_IDENTITY data source should return `Arn` and `UserID` as per API docs http://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

PR #13092, opened 2017-03-27
provider/aws: Update calling_identity

grubernaut:
Updates `aws_caller_identity` data source to actually include the correct attributes from the `GetCallerIdentity` API function.

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCallerIdentity_basic'
2017/03/27 09:26:13 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCallerIdentity_basic -timeout 120m
=== RUN   TestAccAWSCallerIdentity_basic
--- PASS: TestAccAWSCallerIdentity_basic (12.74s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    12.767s
```

Fixes: #13084

### DIFF
--- a/builtin/providers/aws/data_source_aws_caller_identity.go
+++ b/builtin/providers/aws/data_source_aws_caller_identity.go
@@ -50,8 +50,8 @@ func dataSourceAwsCallerIdentityRead(d *schema.ResourceData, meta interface{}) e
 	}
 
 	log.Printf("[DEBUG] Setting AWS Account ID to %s.", *res.Account)
-	d.Set("account_id", *res.Account)
-	d.Set("arn", *res.Arn)
-	d.Set("user_id", *res.UserId)
+	d.Set("account_id", res.Account)
+	d.Set("arn", res.Arn)
+	d.Set("user_id", res.UserId)
 	return nil
 }
--- a/builtin/providers/aws/data_source_aws_caller_identity.go
+++ b/builtin/providers/aws/data_source_aws_caller_identity.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/sts"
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/data_source_aws_caller_identity.go
```diff
@@ -17,24 +18,40 @@ func dataSourceAwsCallerIdentity() *schema.Resource {
...
 			"skip_requesting_account_id is not set on the AWS provider.")
 	}
 
-	log.Printf("[DEBUG] Setting AWS Account ID to %s.", client.accountid)
-	d.Set("account_id", meta.(*AWSClient).accountid)
-
+	log.Printf("[DEBUG] Setting AWS Account ID to %s.", *res.Account)
+	d.Set("account_id", *res.Account)
+	d.Set("arn", *res.Arn)
+	d.Set("user_id", *res.UserId)
```
apparentlymart: A pattern I've seen elsewhere is to just pass in the pointer without dereferencing it, since `d.Set` will automatically unwrap it safely, and thus we avoid crashes if for some reason AWS ends up returning `nil` for any of these.

website/source/docs/providers/aws/d/caller_identity.html.markdown
```diff
@@ -32,4 +40,6 @@ There are no arguments available for this data source.
 
 ## Attributes Reference
 
-`account_id` is set to the ID of the AWS account.
+`account_id` - The AWS Account ID number of the account that owns or contains the calling entity.
```
apparentlymart: I guess this wasn't a list before because there was only one item, but our usual style is for the exported attributes to be a bullet list using `*`.

Issue #11232, opened 2017-01-17
AWS SES support for email sending configuration


ksperling:
#6771 / #5387 added AWS SES support for email receiving functionality.

It would be great if sending functionality could be provisioned via terraform as well, specifically creation of domain and email identities and configuration sets.

There are a number of up-votes for this on https://github.com/hashicorp/terraform/issues/6771#issuecomment-250381602 but since that's a closed issue I thought it's better to open a separate one.

PR #13098, opened 2017-03-27
provider/aws: Implement aws_ses_domain_identity

dougneal:
Adding the capability to [add domain identities](http://docs.aws.amazon.com/cli/latest/reference/ses/verify-domain-identity.html) to Amazon SES and capture the domain verification code.

Tasks:
- [x] Implement CRUD functions
- [x] Implement import
- [x] Complete test coverage
- [x] Update documentation

Should (partially) satisfy https://github.com/hashicorp/terraform/issues/11232.

### DIFF
--- a/builtin/providers/aws/resource_aws_ses_domain_identity.go
+++ b/builtin/providers/aws/resource_aws_ses_domain_identity.go
@@ -1,0 +1,99 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ses"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsSesDomainIdentity() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesDomainIdentityCreate,
+		Read:   resourceAwsSesDomainIdentityRead,
+		Delete: resourceAwsSesDomainIdentityDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"domain": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"verification_token": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsSesDomainIdentityCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	domainName := d.Get("domain").(string)
+
+	createOpts := &ses.VerifyDomainIdentityInput{
+		Domain: aws.String(domainName),
+	}
+
+	_, err := conn.VerifyDomainIdentity(createOpts)
+	if err != nil {
+		return fmt.Errorf("Error requesting SES domain identity verification: %s", err)
+	}
+
+	d.SetId(domainName)
+
+	return resourceAwsSesDomainIdentityRead(d, meta)
+}
+
+func resourceAwsSesDomainIdentityRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	domainName := d.Id()
+	d.Set("domain", domainName)
+
+	readOpts := &ses.GetIdentityVerificationAttributesInput{
+		Identities: []*string{
+			aws.String(domainName),
+		},
+	}
+
+	response, err := conn.GetIdentityVerificationAttributes(readOpts)
+	if err != nil {
+		log.Printf("[WARN] Error fetching identity verification attributes for %s: %s", d.Id(), err)
+		return err
+	}
+
+	verificationAttrs, ok := response.VerificationAttributes[domainName]
+	if !ok {
+		log.Printf("[WARN] Domain not listed in response when fetching verification attributes for %s", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("verification_token", verificationAttrs.VerificationToken)
+	return nil
+}
+
+func resourceAwsSesDomainIdentityDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	domainName := d.Get("domain").(string)
+
+	deleteOpts := &ses.DeleteIdentityInput{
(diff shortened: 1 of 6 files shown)


dougneal:
Build failure appears to be caused by unrelated code.

builtin/providers/aws/resource_aws_ses_domain_identity_test.go
```diff
@@ -0,0 +1,103 @@
...
+			return fmt.Errorf("SES Domain Identity %s not found in AWS", domain)
+		}
+
+		return nil
+	}
+}
+
+const testAccAwsSESDomainIdentityConfig = `
+resource "aws_ses_domain_identity" "test" {
+	domain = "example.com"
```
radeksimko: Would you mind randomizing this domain, so that we can safely run this test multiple times in parallel?
Have a look at how we do it in other tests, e.g. [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_alb_test.go#L50-L62)

I think for the purpose of this test it doesn't matter if we're ever going to be able to finish the verification process (i.e. own the domain or manage DNS for it), so I'd go for something like `%d.terraformtesting.com`.

builtin/providers/aws/resource_aws_ses_domain_identity.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	response, err := conn.GetIdentityVerificationAttributes(readOpts)
+	if err != nil {
+		log.Printf("[WARN] Error fetching identity verification attributes for %s: %s", d.Id(), err)
+		return err
+	}
+
+	verificationAttrs := response.VerificationAttributes[domainName]
+	if verificationAttrs == nil {
```
radeksimko: I think more idiomatic way to check the existence of a key in a map would be something like
```go
verificationAttrs, ok := response.VerificationAttributes[domainName]
if !ok {
```

website/source/docs/providers/aws/r/ses_domain_identity.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+layout: "aws"
+page_title: "AWS: ses_domain_identity"
+sidebar_current: "docs-aws-resource-ses-domain-identity"
+description: |-
+  Provides an SES domain identity resource
+---
+
+# aws\_ses\_domain_identity
+
+Provides an SES domain identity resource
```
radeksimko: Would you mind attaching some links and/or details about what steps are expected from the user to actually finish the verification of the domain? What do you think about adding a Route53 DNS record to the example + mentioning the WHOIS records need to point to the zone, or eventually delegate to it? It may sound like an obvious thing, but I think it's good to explain to what extent do we automate this process and help the user along the way.

http://docs.aws.amazon.com/ses/latest/DeveloperGuide/dns-txt-records.html

website/source/docs/providers/aws/r/ses_domain_identity.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `verification_token` - A code which when added to the domain as a TXT record
+  will signal to SES that the owner of the domain has authorised SES to act on
+  their behalf. The domain identity will be in state "verification pending"
+  until this is done. Find out more about verifying domains in Amazon SES in
+  the [AWS SES docs](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/verify-domains.html).
```
radeksimko: Oh, I see, you have the explanation here actually, that's good 👍  but I still think we should include an example of a DNS record, like Route53 record.

dougneal:
Thanks for the feedback @radeksimko, I've addressed all points. I reordered the documentation slightly to put the example at the end including a Route 53 resource to add the verification token.

builtin/providers/aws/resource_aws_ses_domain_identity_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if response.VerificationAttributes[domain] != nil {
+			return fmt.Errorf("SES Domain Identity %s still exists. Failing!", domain)
+		}
+
+		// Verify the error is what we want
+		_, ok := err.(awserr.Error)
+		if !ok {
+			return err
+		}
```
radeksimko: Probably a nitpick, but these 4 line of code are redundant since we always return from _any_ error above (which _is_ the right thing to do btw. as there's nothing like `NotFound` error in SES API). Would you mind removing these - just to avoid confusions?

builtin/providers/aws/resource_aws_ses_domain_identity.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	verificationAttrs, ok := response.VerificationAttributes[domainName]
+	if !ok {
+		log.Printf("[WARN] Domain not listed in response when fetching verification attributes for %s", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("verification_token", *verificationAttrs.VerificationToken)
```
radeksimko: I'm not sure why I didn't notice this one before 🙀 , but we tend to avoid pointer dereferencing in `Set()` - the function has [a built-in functionality](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L153-L169) to do this safely for all primitive data types and we have been bitten many times in the past by doing this in the resource code.

The API (for any weird reason) may not return the token here, in which case we'd be dereferencing pointer to `nil` which would in turn cause crash.

builtin/providers/aws/resource_aws_ses_domain_identity_test.go
dougneal: Sure. I actually lifted most of the code from [here|https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_ses_receipt_rule_test.go#L70] so there is probably some further cleanup to be done across the SES code...

builtin/providers/aws/resource_aws_ses_domain_identity.go
dougneal: Cool, so should I just pass the pointer in?

dougneal:
All done!

builtin/providers/aws/resource_aws_ses_domain_identity_test.go
radeksimko: Good catch, I think that is slightly different situation as you may get errors, but we should be scoping it down to specific error code(s): http://docs.aws.amazon.com/ses/latest/APIReference/API_DescribeReceiptRule.html#API_DescribeReceiptRule_Errors

PR #13099, opened 2017-03-27
helper/resource: Allow unknown "pending" states

jen20:
Sometimes when waiting on a target state, the set of valid states through which a value will transition is unknown. This commit adds support for an empty Pending slice and will treat any states that are not the target as valid provided the timeouts are not exceeded.


stack72:
LGTM :)

PR #13105, opened 2017-03-27
disallow env names that aren't url-safe

jbardin:
Environment names can be used in a number of contexts, and should be
properly escaped for safety. Since most state names are store in path
structures, and often in a URL, use `url.PathEscape` to check for
disallowed characters.

### DIFF
--- a/command/env_command.go
+++ b/command/env_command.go
@@ -1,6 +1,9 @@
 package command
 
-import "strings"
+import (
+	"net/url"
+	"strings"
+)
 
 // EnvCommand is a Command Implementation that manipulates local state
 // environments.
@@ -37,6 +40,13 @@ Subcommands:
 
 func (c *EnvCommand) Synopsis() string {
 	return "Environment management"
+}
+
+// validEnvName returns true is this name is valid to use as an environment name.
+// Since most named states are accessed via a filesystem path or URL, check if
+// escaping the name would be required.
+func validEnvName(name string) bool {
+	return name == url.PathEscape(name)
 }
 
 const (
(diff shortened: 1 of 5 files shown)


jbardin:
@mitchellh, This is technically a breaking change, since someone could be use path separators in their state name already. I feel like it's better to do something like this as early as possible, rather than wait for a major release.

jbardin:
Actually, a version that could make use of `/` in a name never was released, but there could possibly some other reserved/unicode chars in an existing local env name.

jbardin:
Updated this to simply reject invalid names, which should make the result much more obvious for users.

command/env_command_test.go
```diff
@@ -103,6 +103,44 @@ func TestEnv_createAndList(t *testing.T) {
...
+
+	if code := listCmd.Run(nil); code != 0 {
+		t.Fatalf("bad: %d\n\n%s", code, ui.ErrorWriter)
+	}
+
+	actual := strings.TrimSpace(ui.OutputWriter.String())
+	expected := "* default"
+
+	if actual != expected {
+		t.Fatalf("\nexpcted: %q\nactual:  %q", expected, actual)
```
apparentlymart: little typo here

command/env_command.go
```diff
@@ -39,6 +42,13 @@ func (c *EnvCommand) Synopsis() string {
 	return "Environment management"
 }
 
+// validEnvName returns true is this name is valid to use as an environment name.
+// Since most named states are accessed via a filesystem path or URL, check if
+// escaping the name would be required.
+func validEnvName(name string) bool {
+	return name == url.PathEscape(name)
```
mitchellh: This is clever!

PR #13110, opened 2017-03-27
Don't Validate if we have an execution plan

jbardin:
The plan file should contain all data required to execute the apply
operation. Validation requires interpolation, and the `file()`
interpolation function may fail if the module files are not present.
This is the case currently with how TFE executes plans.


jbardin:
HAHA, a wild unit test appears!
Unfortunately there's easy way to do a full end-to-end test of this in the command package.

PR #13113, opened 2017-03-28
environment docs

jbardin:
List backends supporting environments.
Note character restrictions for environment names.


jbardin:
@apparentlymart, yes that might be useful for reference. I just updated the initial S3 implementation in #13104 to match that of consul too.

mitchellh:
I've also been putting what a backend supports on the backend type pages!

PR #13126, opened 2017-03-28
Added "$" back into aws_iam_policy_attachment example.

bisoldi:
The aws_iam_policy_attachment example is missing the "$" from roles, groups, users properties, so I simply added it back in.


apparentlymart:
Thanks, @bisoldi! Not sure what happened there...

PR #13137, opened 2017-03-28
Allow escaped interpolation-like sequences in variable defaults

### LINKED ISSUE #2909: No way to escape a dollar in template variable.
Terraform 0.6.1

In a template variable section, I need to define a variable like this:

```
resource "template_file" "foo_template" {
    filename = "foo.yaml"
    vars     = {
        TF_FOO = "bar-${baz}"
    }
}
```

That is - literally, where I have ${TF_FOO} in my foo.yaml, I want it replaced, literally with bar-${baz}

Now this obviously doesn't work, because terraform treats ${baz} as a variable and rightly says it doesn't exist.

So then I tried

```
TF_FOO="bar-$${baz}"
```

No luck - this gets written into the template as ... literally bar-$${baz}

So then I tried

```
TF_FOO="bar-\${baz}"
```

No luck - terraform treats this as a variable again.

There is no way to get terraform to write a literal ${baz} as a variable replacement in a terraform file.

### LINKED ISSUE #13001: Can't use anything that looks like interpolation in variable defaults
### Terraform 
Terraform v0.9.1

```hcl
variable "constraint" {
  default = {
    attribute = "$${cpu.numcores}"
    operator  = ">="
    value     = "4"
  }
}

output "out" {
  value = "${var.constraint}"
}
```

### Debug Output
https://gist.github.com/ORBAT/8a26a1647599ac7fb6fd3daa77ecb2c2

### Expected Behavior
```

Outputs:

out = {
  attribute = ${cpu.numcores}
  operator = >=
  value = 4
}
```

### Actual Behavior
```
1 error(s) occurred:

* module root: 1 error(s) occurred:

* Variable 'constraint': cannot contain interpolations
```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids

Also tried `\${...}` and `${...}`.

apparentlymart:
The variable validator assumes that any AST node it gets from an interpolation walk is an indicator of an interpolation. Unfortunately, back in f223be15 we changed the interpolation walker to emit a LiteralNode
as a way to signal that the result is a literal but not identical to the input due to escapes.

The existence of this issue suggests a bit of a design smell in that the interpolation walker interface at first glance appears to skip over all literals, but it actually emits them in this one situation. In the long run we should perhaps think about whether the abstraction is right here, but this is a shallow, tactical change that fixes #13001.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -285,8 +285,15 @@ func (c *Config) Validate() error {
 		}
 
 		interp := false
-		fn := func(ast.Node) (interface{}, error) {
-			interp = true
+		fn := func(n ast.Node) (interface{}, error) {
+			// LiteralNode is a literal string (outside of a ${ ... } sequence).
+			// interpolationWalker skips most of these. but in particular it
+			// visits those that have escaped sequences (like $${foo}) as a
+			// signal that *some* processing is required on this string. For
+			// our purposes here though, this is fine and not an interpolation.
+			if _, ok := n.(*ast.LiteralNode); !ok {
+				interp = true
+			}
 			return "", nil
 		}
 
--- a/command/test-fixtures/validate-valid/main.tf
+++ b/command/test-fixtures/validate-valid/main.tf
@@ -1,3 +1,8 @@
+variable "var_with_escaped_interp" {
+  # This is here because in the past it failed. See Github #13001
+  default = "foo-$${bar.baz}"
+}
+
 resource "test_instance" "foo" {
     ami = "bar"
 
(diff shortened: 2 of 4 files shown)


config/config.go
```diff
@@ -285,8 +285,10 @@ func (c *Config) Validate() error {
 		}
 
 		interp := false
-		fn := func(ast.Node) (interface{}, error) {
-			interp = true
+		fn := func(n ast.Node) (interface{}, error) {
+			if _, ok := n.(*ast.LiteralNode); !ok {
```
mitchellh: Comment above here please explaining what this is looking for. :) Its obvious for us since we know the HIL AST.

PR #13147, opened 2017-03-29
provider/aws: Mark password fields as sensitive

radeksimko:
Related to https://github.com/hashicorp/terraform/pull/13132 and https://github.com/hashicorp/terraform/pull/13145

These came up when I `grep`ped through aws provider for `password"` and `key"`.


radeksimko:
No, this won't trigger any diff because `Sensitive` doesn't affect state at all, it's purely a UI feature which affects whether or not is the attribute value displayed in `plan` or `apply` output.

dougneal:
@radeksimko @stack72 in relation to directory services passwords would you consider merging https://github.com/hashicorp/terraform/pull/9848 as an additional measure? As far as I can tell there is no use of storing the password unhashed once it's been set initially!

PR #13148, opened 2017-03-29
provider/google: Mark GKE pass as sensitive

radeksimko:
Related to https://github.com/hashicorp/terraform/pull/13132 and https://github.com/hashicorp/terraform/pull/13145


radeksimko:
As mentioned in the previous PR - it won't have any significant (meaning it doesn't break things) effect on the user as `Sensitive` doesn't affect whether or how is the value stored in the state. The value remains as is (plaintext). It only affects whether it's displayed in the `plan` or `apply` output.

PR #13165, opened 2017-03-29
add Rehash to terraform.BackendState

### LINKED ISSUE #13107: Partial s3 backend config errors out when running plan on v0.9.1
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.9.1

### Affected Resource(s)
- backend

```hcl
terraform {
  backend "s3" {
    bucket = "XXXXXXX"
    region = "us-west-2"
  }
}
```

### Debug Output
https://gist.github.com/JschuttTableau/b2f4e6fe32e683cce9d7b4ee86512b1d

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
* terraform init with s3 with a -backend-config option for the key name passes
* terraform plan works as normal

### Actual Behavior
What actually happened?
* terraform init with s3 with a -backend-config option for the key name passes
* terraform plan errors out with  "Backend reinitialization required. Please run "terraform init"

### Steps to Reproduce
1. Create a s3 backend config but missing the key parameter
2. terraform init -backend=true -input=true -backend-config key=mystate.tfstate
3. terraform plan

### Important Factoids
We originally had all the options set in the terraform:backend config, but we want to parameterize this to support multiple environments. 

### References
- GH-1234

### LINKED ISSUE #13149: terraform init reconfigures backend every time
Hi there,

### Terraform Version
0.9.2

### Affected Resource(s)
- Terraform core - init command

```hcl
terraform {
  required_version = "0.9.2"

  backend "s3" {
  }
}
```

### Expected Behavior
Terraform init detects that I have already reconfigured my backend to use S3 and doesn't do it on every run.

### Actual Behavior
```
Initializing the backend...
Backend configuration changed!

Terraform has detected that the configuration specified for the backend
has changed. Terraform will now reconfigure for this backend. If you didn't
intend to reconfigure your backend please undo any changes to the "backend"
section in your Terraform configuration.

should now work.

```

### Steps to Reproduce
1. Run Terraform init
```
REGION="eu-west-1"

terraform init \
    -get=true \
    -input=false \
    -force-copy \
    -backend=true \
    -backend-config "bucket=some-bucket" \
    -backend-config "key=system-development" \
    -backend-config "region=$REGION"
```
2. Terraform init states that the back-end is successfully reconfigured
3. Run Terraform init again (same command as in step 1)
4. Terraform init again states that the back-end was reconfigured

### Important Factoids
I'm running Terraform as part of a CI/CD pipeline. For this reason Terraform init is run every time we run any Terraform command to make sure that the state is set-up correctly. In our CI/CD pipeline we also don't know whether an environment already exists up front.

jbardin:
This method mirrors that of config.Backend, so we can compare the
configuration of a backend read from a config vs that of a backend read
from a state. This will prevent init from reinitializing when using
`-backend-config` options that match the existing state.

After determining that there is no drift in the config, check that the stored state has the correct hash for the configuration. If an option was removed from the config but supplied via the CLI, the effective hash is correct, but the stored backend state needs to be updated and no migration needs to be done. 

fixes #13149
fixes #13107

### DIFF
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -20,6 +20,7 @@ import (
 	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/terraform/config"
 	"github.com/mitchellh/copystructure"
+	"github.com/mitchellh/hashstructure"
 	"github.com/satori/go.uuid"
 )
 
@@ -799,6 +800,32 @@ type BackendState struct {
 // Empty returns true if BackendState has no state.
 func (s *BackendState) Empty() bool {
 	return s == nil || s.Type == ""
+}
+
+// Rehash returns a unique content hash for this backend's configuration
+// as a uint64 value.
+// The Hash stored in the backend state needs to match the config itself, but
+// we need to compare the backend config after it has been combined with all
+// options.
+// This function must match the implementation used by config.Backend.
+func (s *BackendState) Rehash() uint64 {
+	if s == nil {
+		return 0
+	}
+
+	// Use hashstructure to hash only our type with the config.
+	code, err := hashstructure.Hash(map[string]interface{}{
+		"type":   s.Type,
+		"config": s.Config,
+	}, nil)
+
+	// This should never happen since we have just some basic primitives
+	// so panic if there is an error.
+	if err != nil {
+		panic(err)
+	}
+
+	return code
 }
 
 // RemoteState is used to track the information about a remote
--- a/terraform/state.go
+++ b/terraform/state.go
@@ -20,7 +20,6 @@ import (
 	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/terraform/config"
 	"github.com/mitchellh/copystructure"
-	"github.com/mitchellh/hashstructure"
 	"github.com/satori/go.uuid"
 )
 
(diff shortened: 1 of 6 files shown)


jbardin:
@mitchellh: this version now does a Rehash on the BackendState too. There was also the inverse drift problem which I added a fix for, where there is no change in the effective config but we need to update the hash.

terraform/state.go
```diff
@@ -801,6 +802,32 @@ func (s *BackendState) Empty() bool {
...
+		"config": s.Config,
+	}, nil)
+
+	// This should never happen since we have just some basic primitives
+	// so panic if there is an error.
+	if err != nil {
+		panic(err)
+	}
+
+	return code
```
mitchellh: I'd prefer here instead of copy/pasting the config code to actually generate a `config.Backend` and call Rehash. That way we only need to ensure that Rehash is bug free in one place. Although, if the hash ever changes its a big problem. :) But at least it limits that to one code path.

jbardin:
Updated with requested changes

manuwelakanade:
@jbardin :  I was facing https://github.com/hashicorp/terraform/issues/13149, and I've also added comment about the issue with locks. Is there a workaround which can be used in 0.9.2 which will not cause this issue?

jbardin:
@manuwelakanade,

Sorry about that, That's another issue which I meant to open, but was in the process of narrowing down the cause. I'll get a fix out, or at least get a specific issue filed today. 

In the meantime you could remove the `.terraform/terraform.tfstate` file (this should only contain the backend state), and run `tf init` again with the correct config. As always, make sure you have backups of the actual state file.

PR #13166, opened 2017-03-29
provider/opc: Updating the Website

tombuildsstuff:
Updating the Website files to match the provider name


mbfrahry:
Did we decide oracleopc was the way to go? It seems redundant with opc standing for oracle public cloud

tombuildsstuff:
> Did we decide oracleopc was the way to go? It seems redundant with opc standing for oracle public cloud

After chatting internally we're going to merge this with this name for now

PR #13175, opened 2017-03-29
provider/aws: Add support for aws_lightsail_static_ip

radeksimko:
Related: https://github.com/hashicorp/terraform/issues/10447

I'll submit `aws_lightsail_static_ip_attachment` too and that should hopefully allow us to close the mentioned issue and satisfy Lightsail users for a while.

There's a potential for a few more resources based on the current API, but I'd close that issue anyway, otherwise it will have no end. People can always submit PRs for those if they wish.

 - `aws_lightsail_domain_entry`
 - `aws_lightsail_public_port`

### Test plan

```
make testacc TEST=./builtin/providers/aws TESTARGS='-run=AWSLightsailStaticIp'
2017/03/29 21:23:40 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=AWSLightsailStaticIp -timeout 120m
=== RUN   TestAccAWSLightsailStaticIp_basic
--- PASS: TestAccAWSLightsailStaticIp_basic (19.12s)
=== RUN   TestAccAWSLightsailStaticIp_disappears
--- PASS: TestAccAWSLightsailStaticIp_disappears (13.50s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	32.662s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_lightsail_static_ip.go
+++ b/builtin/providers/aws/resource_aws_lightsail_static_ip.go
@@ -1,0 +1,98 @@
+package aws
+
+import (
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/lightsail"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsLightsailStaticIp() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsLightsailStaticIpCreate,
+		Read:   resourceAwsLightsailStaticIpRead,
+		Delete: resourceAwsLightsailStaticIpDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ip_address": {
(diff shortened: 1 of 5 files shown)


builtin/providers/aws/resource_aws_lightsail_static_ip.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return nil
+}
+
+func resourceAwsLightsailStaticIpDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).lightsailconn
+
+	name := d.Get("name").(string)
+	log.Printf("[INFO] Deleting Lightsail Static IP: %q", name)
+	out, err := conn.ReleaseStaticIp(&lightsail.ReleaseStaticIpInput{
```
stack72: If the static IP is in use by an instance, will this release API call work? Or will it fail?

radeksimko: Good question, I just tested this situation and the releasing just succeeded. Admittedly I'd expect and like it to fail (and implement something like `force_destroy (bool)` to override it), but there isn't much we can do about this as that's how the API works.

stack72: In that case, LGTM :)

PR #13180, opened 2017-03-29
[WIP] vet should fail in travis

jbardin:


jbardin:
Leaving this alone for now. It's a mystery

Issue #13163, opened 2017-03-29
github_issue_label is not idempotent


sethvargo:
### Terraform Version

`Terraform v0.9.3-dev (51081678a45fdb4276f0a914c848d6a661879ab8+CHANGES)`

### Affected Resource(s)

- `github_issue_label`

### Terraform Configuration Files

```hcl
resource "github_issue_label" "test_repo" {
  repository = "tf-test"
  name       = "enhancement"
  color      = "FF00FF"
}
```

### Error Output

```
Error applying plan:

1 error(s) occurred:

* github_issue_label.test_repo: 1 error(s) occurred:

* github_issue_label.test_repo: POST https://api.github.com/repos/sethvargo/tf-test/labels: 422 Validation Failed [{Resource:Label Field:name Code:already_exists Message:}]
```

### Expected Behavior

Background: the label "enhancement" already exists. Terraform should have updated the existing "enhancement" label to the new color.

### Actual Behavior

Terraform throws an error.

### Steps to Reproduce

1. `terraform apply`

PR #13182, opened 2017-03-29
Handle the case when issue labels already exist

sethvargo:
This fixes GH-13163

I'm not sure how to write an acceptance test for this, since I'd have to create the issue outside of the acceptance test first. Any suggestions or examples of where we do that already?

### DIFF
--- a/builtin/providers/github/resource_github_issue_label_test.go
+++ b/builtin/providers/github/resource_github_issue_label_test.go
@@ -32,6 +32,13 @@ func TestAccGithubIssueLabel_basic(t *testing.T) {
 					testAccCheckGithubIssueLabelAttributes(&label, "bar", "FFFFFF"),
 				),
 			},
+			{
+				Config: testAccGitHubIssueLabelExistsConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGithubIssueLabelExists("github_issue_label.test", &label),
+					testAccCheckGithubIssueLabelAttributes(&label, "enhancement", "FF00FF"),
+				),
+			},
 		},
 	})
 }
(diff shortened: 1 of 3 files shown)


radeksimko:
Existing acceptance tests are passing, it would be nice to add another one covering this use case.
The code otherwise looks good, assuming conflicts are resolved.

> I'd have to create the issue outside of the acceptance test first

How about we just create `github_repository` as part of the acceptance test which will have the mentioned default labels and then try modified one of the default labels we expect to exist there?

builtin/providers/github/resource_github_issue_label_test.go
```diff
@@ -134,3 +141,16 @@ resource "github_issue_label" "test" {
   color      = "FFFFFF"
 }
 `, testRepo)
+
+var testAccGitHubIssueLabelExistsConfig string = fmt.Sprintf(`
+// Create a repository which has the default labels
+resource "github_repository" "test" {
+  name = "tf-acc-repo-label-abc1234"
```
radeksimko: I think we'll need to randomize this, but apparently the other tests here aren't randomized yet either, so I'll do it in a separate PR.

website/source/docs/providers/github/r/issue_label.html.markdown
```diff
@@ -6,16 +6,24 @@ description: |-
   Provides a GitHub issue label resource.
 ---
 
-# github\_issue_label
+# github_issue_label
```
radeksimko: I never really understood the point of the `\` escaping here in this context 🤔

builtin/providers/github/resource_github_issue_label_test.go
sethvargo: Yea, I tried that. I used random_id, but random_id doesn't get compiled 😦

website/source/docs/providers/github/r/issue_label.html.markdown
sethvargo: It does nothing... Trust me lol. This is identical.

builtin/providers/github/resource_github_issue_label_test.go
radeksimko: So we usually generate the randomness within the test via helpers (https://github.com/hashicorp/terraform/tree/master/helper/acctest), intentionally to not couple providers like `random` with other ones.

PR #13188, opened 2017-03-30
provider/aws Add support for aws_subnet_ids as a data source

### LINKED ISSUE #9843: provider/aws: Subnets Data Source
A data source that provides a list of all available subnets in the vpc would be helpful

cwood:
Fixes #9843

### DIFF
--- a/builtin/providers/aws/data_source_aws_subnet_ids.go
+++ b/builtin/providers/aws/data_source_aws_subnet_ids.go
@@ -1,0 +1,60 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsSubnetIDs() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsSubnetIDsRead,
+		Schema: map[string]*schema.Schema{
+			"vpc_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"ids": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+		},
+	}
+}
+
+func dataSourceAwsSubnetIDsRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	req := &ec2.DescribeSubnetsInput{}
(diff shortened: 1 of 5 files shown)


apparentlymart:
Hi @cwood! Thanks for working on this.

We didn't initially implement these "plural" data sources because Terraform's core currently makes it hard to wrangle lists of complex objects like this. This is why, for example, the `aws_availability_zones` data source returns a list of names rather than a list of complex objects.

I certainly agree that being able to list subnets matching constraints would be useful, but I'm not sure that this particular design would be very usable with today's language limitations. For example, there isn't a good way to express "all of the id attribute values from the members of this list" in the interpolation language.

Perhaps in the short term it'd be useful to have an `aws_subnet_ids` resource that, similarly to `aws_availability_zones`, just returns a list of strings that are the ids of the matched subnet, which could then be used with the existing `aws_subnet` data source to get the result I think you were looking for here:

```hcl
data "aws_subnet_ids" "example" {
  vpc_id = "${var.vpc_id}"
}

data "aws_subnet" "example" {
  count = "${length(data.aws_subnet_ids.example.ids)}"
  id = "${aws_subnet_ids.example.ids[count.index]}"
}

output "subnet_cidr_blocks" {
  value = ["${data.aws_subnet.example.*.cidr_block}"]
}
```

What do you think?

cwood:
That's honestly fine with me. I only want subnet_ids anyway and to use a vpc_id

Derek-Ashmore:
++1 -- Thanks for considering this.  The lack of either aws_subnets or aws_subnet_ids is causing a lot of code bloat for me.

apparentlymart:
Thanks for the update @cwood. Looking good now!

The only other thing (which I'm sorry I didn't catch on the first pass) is that we should probably have an acceptance test for this. Acceptance tests for data sources can be kinda weird, but I think we could get away here with something that applies a config that creates a VPC and a couple subnets and then uses your data source to retrieve the subnets from that VPC and verifies that it gets back the two subnet ids we expect.

(It looks like Travis caught some compile-time issues, too...)

cwood:
I am working on the compiling issues right now. Ill work on an acceptance test later. Is there a way I can run only the acceptance test for this data source and not the whole collection.

apparentlymart:
There's some long details on this here:
https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#running-an-acceptance-test

...but in summary: you can use the `TESTARGS` variable to set the `-run` argument on the test runner, which allows you to run only the tests whose names match the given regular expression. Usually when running acceptance tests we'll run specifically the test we're working on, since acceptance tests are often relatively slow.

PR #13217, opened 2017-03-30
Possible correction regarding remote state files

fillup:
Documentation has "... people have to remove state files..." when I believe it should say "...people have to remote state files...".


fillup:
Yes, good suggestion, thanks!

PR #13222, opened 2017-03-30
Fix some backend locking issues

### LINKED ISSUE #13215: can't re-initialize an s3 backend with invalid lock_table
If an s3 backend was initialized with an incorrect `lock_table` value, you can't fix the backend with the `init` command.

If the `lock_table` value was supplied via `-backend-config`, `init` doesn't do anything, but all write commands will fail. Init can be triggered again by supplying `-backend-config=lock_table=""`, but it will still fail to get a lock.

Removing the `lock_table` value from the config will cause `init` to try a migration, but the fail to get a lock.

A manual solution of providing the `-lock` flag may be sufficient for this edge-case.

jbardin:
Skip copying of states during migration if they are the same state. This
can happen when trying to reconfigure a backend's options, or if the
state was manually transferred. This can fail unexpectedly with locking
enabled.

Honor the `-input` flag for all confirmations (the new test hit some
more). Also unify where we reference the Meta.forceInitCopy and transfer
the value to the existing backendMigrateOpts.force field.

fixes #13215

### DIFF
--- a/backend/remote-state/s3/client.go
+++ b/backend/remote-state/s3/client.go
@@ -113,8 +113,7 @@ func (c *RemoteClient) Lock(info *state.LockInfo) (string, error) {
 		return "", nil
 	}
 
-	stateName := fmt.Sprintf("%s/%s", c.bucketName, c.path)
-	info.Path = stateName
+	info.Path = c.lockPath()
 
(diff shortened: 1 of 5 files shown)


jbardin:
rebased on master

PR #13223, opened 2017-03-30
provider/azurerm: Setting an optional field on Service Bus Topic

tombuildsstuff:
The field `duplicate_detection_history_time_window` is defaulted at the API to `00:10:00`

Due a bug in the current version of the Azure SDK, this value wasn't being surfaced correctly, however we're going to need to make this optional by default

(also fixing the logging to output the error's)

### DIFF
--- a/builtin/providers/azurerm/resource_arm_servicebus_topic.go
+++ b/builtin/providers/azurerm/resource_arm_servicebus_topic.go
@@ -173,7 +173,7 @@ func resourceArmServiceBusTopicRead(d *schema.ResourceData, meta interface{}) er
 
 	resp, err := client.Get(resGroup, namespaceName, name)
 	if err != nil {
-		return fmt.Errorf("Error making Read request on Azure ServiceBus Topic %s: %s", name, err)
+		return fmt.Errorf("Error making Read request on Azure ServiceBus Topic %s: %+v", name, err)
 	}
 	if resp.StatusCode == http.StatusNotFound {
 		d.SetId("")
--- a/builtin/providers/azurerm/resource_arm_servicebus_topic.go
+++ b/builtin/providers/azurerm/resource_arm_servicebus_topic.go
@@ -55,6 +55,7 @@ func resourceArmServiceBusTopic() *schema.Resource {
 			"duplicate_detection_history_time_window": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 
 			"enable_batched_operations": {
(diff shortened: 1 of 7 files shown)


tombuildsstuff:
Tests pass:

```
$ envchain azurerm make testacc TEST=./builtin/providers/azurerm TESTARGS='-run=TestAccAzureRMServiceBusTopic'
2017/03/31 20:25:40 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMServiceBusTopic -timeout 120m
=== RUN   TestAccAzureRMServiceBusTopic_importBasic
--- PASS: TestAccAzureRMServiceBusTopic_importBasic (339.46s)
=== RUN   TestAccAzureRMServiceBusTopic_basic
--- PASS: TestAccAzureRMServiceBusTopic_basic (320.62s)
=== RUN   TestAccAzureRMServiceBusTopic_update
--- PASS: TestAccAzureRMServiceBusTopic_update (349.12s)
=== RUN   TestAccAzureRMServiceBusTopic_enablePartitioningStandard
--- PASS: TestAccAzureRMServiceBusTopic_enablePartitioningStandard (476.61s)
=== RUN   TestAccAzureRMServiceBusTopic_enablePartitioningPremium
--- PASS: TestAccAzureRMServiceBusTopic_enablePartitioningPremium (643.32s)
=== RUN   TestAccAzureRMServiceBusTopic_enableDuplicateDetection
--- PASS: TestAccAzureRMServiceBusTopic_enableDuplicateDetection (340.50s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/azurerm	2469.647s
```

builtin/providers/azurerm/resource_arm_servicebus_topic.go
```diff
@@ -55,6 +55,7 @@ func resourceArmServiceBusTopic() *schema.Resource {
 			"duplicate_detection_history_time_window": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
```
stack72: So we are not setting the default value? We are just getting it from Azure API?

tombuildsstuff: Yeah - unfortunately setting this to `Optional: true, Default: "00:30:00"` leaves a diff in the Import test:
```
TF_ACC=1 go test ./builtin/providers/azurerm/ -v -run=TestAccAzureRMServiceBusTopic_import -timeout 120m
=== RUN   TestAccAzureRMServiceBusTopic_importBasic
--- FAIL: TestAccAzureRMServiceBusTopic_importBasic (349.48s)
	testing.go:280: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) {
		}


		(map[string]string) (len=1) {
		 (string) (len=39) "duplicate_detection_history_time_window": (string) (len=8) "00:30:00"
		}

FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/azurerm	349.497s
```

stack72:
LGTM! :shipit:

PR #13248, opened 2017-03-31
kubernetes PatchOperations tests reliability

zimbatm:
This fixes a flakey test where a map is turned into a list but doesn't
always produce the same result because go maps don't guarantee
predictable ordering.

PatchOperations.Equal sorts the left-hand side before applying the
comparison.

Can be applied against 0.9.2 as well


radeksimko:
Hi @zimbatm 
thanks for the contribution. I did notice the flaky test few days ago and I _believe_ I fixed it: https://github.com/hashicorp/terraform/commit/587b50fc432f2459f8261ed8bed89dc852a8858e

Are you still experiencing flakiness on current master (which includes the linked commit)?

The PR only swaps arguments, but then we sort _both_ slices in the `Equal` function, so I don't believe this would make any difference - but feel free correct me if I'm wrong. 🤔

zimbatm:
Ah never mind :) I developed this against 0.9.2.

Issue #12233, opened 2017-02-24
aws_api_gateway_integration resources not getting updated properly.


naikajah:
### Terraform Version
 Terraform v8.7

### Affected Resource(s)
aws_api_gateway_integration

### Terraform Configuration 
My terraform configuration for aws_api_gateway_integration

```
resource "aws_api_gateway_integration" "testMockGET" {
  rest_api_id             = "${aws_api_gateway_rest_api.test.id}"
  resource_id             = "${aws_api_gateway_resource.testMock.id}"
  type                    = "MOCK"
  uri                     = "/test"
  http_method             = "${aws_api_gateway_method.testGET.http_method}"
}
```
### Expected Behavior
After the first plan and apply created the resource in AWS. The subsequent plan and apply shouldn't modify the resource unless there is any change in the resource definition.

### Actual Behavior
After the first plan and apply creates the resource properly all the subsequent plan shows that there is an update even though there are none and apply updates the URI parameter with the same value again and again.

```
~ module.TEST.aws_api_gateway_integration.testMockGET
 uri: "" => "/test"
```
This is then leading to https://github.com/hashicorp/terraform/issues/9877

### Steps to Reproduce

1. terraform plan original definition
2. terraform apply original definition
3. terraform plan/apply

After looking at the repo, I see that for api_gateway_integration resource we have same method which is resourceAwsApiGatewayIntegrationCreate where as for other resources e.g. api_gateway_method we have an resourceAwsApiGatewayMethodUpdate which takes care of checking if the value has changed.
 
Create: resourceAwsApiGatewayIntegrationCreate,
Update: resourceAwsApiGatewayIntegrationCreate,

Shouldn't there be an Update: resourceAwsApiGatewayIntegrationUpdate for Integration resource that checks for hasChanged().
could this be a problem here?

stack72:
Closed via #13249

Issue #10223, opened 2016-11-18
aws_api_gateway_integration request_parameters JSON?!


ianaz:
Hi there,

More examples for the "request_parameters" would be veeery useful :) maybe also how to send a custom json object to the lambda. I'm probably doing something wrong but I can't bring it to work

### Terraform Version
v0.7.11

### Affected Resource(s)
- aws_api_gateway_integration

### Terraform Configuration Files
```hcl

resource "aws_api_gateway_integration" "community_integration" {
  rest_api_id = "${var.apigateway_id}"
  resource_id = "${aws_api_gateway_resource.api_gateway_res_community_comments.id}"
  http_method = "${aws_api_gateway_method.method_get_comments.http_method}"
  credentials = "${var.lambdarole_arn}"

  type = "AWS"
  integration_http_method = "POST"
  uri = "arn:aws:apigateway:${var.region}:lambda:path/2015-03-31/functions/arn:aws:lambda:${var.region}:${var.account_id}:function:${aws_lambda_function.lambda_community.function_name}/invocations"
  depends_on = [
    "aws_api_gateway_method.method_get_comments"]
request_parameters = { 
"integration.request.header.X-Some-Other-Header" = "method.request.header.X-Some-Header"
 }
}
```

### Debug Output
https://gist.github.com/ianaz/80891b04b939aa16611ea45f4b4649e1

### Expected Behavior
I expect Terraform to create the integration request

### Actual Behavior
I get the error specified

Even by following [your example ](https://www.terraform.io/docs/providers/aws/r/api_gateway_integration.html#request_parameters) I get this error.

ianaz:
Had to fix with

request_parameters {
    "integration.request.header.X-Some-Other-Header" = "'method.request.header.X-Some-Header'"
  }

(note the double quote + single quote). 
Definitely a bug

stack72:
Closed via #13249

PR #13249, opened 2017-03-31
provider/aws: Added API Gateway integration update

Ninir:
## Description
This adds the update part to the API Gateway integration. 

As per [the documentation](https://docs.aws.amazon.com/apigateway/api-reference/link-relation/integration-update/#remarks), everything excepted `request_templates` & `request_parameters` can't be added nor removed, for the attributes we have at time.
As they are not marked as `Required`, we need to flag them to be recreated on change, thus using the `ForceNew` attribute.

I reworked the acc tests so that they are clearer and handle this workflow, since the API to Create is different from the Update one (Series of PatchOperations):
1. Creation of resources using the 
2. Update: Removal + Additions + Replaces
3. Removal & state checks

## Related issues
* https://github.com/hashicorp/terraform/issues/12233
* https://github.com/hashicorp/terraform/issues/10223

## Tests
```bash
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAPIGatewayIntegration_basic'
2017/03/31 17:51:35 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAPIGatewayIntegration_basic -timeout 120m
=== RUN   TestAccAWSAPIGatewayIntegration_basic
--- PASS: TestAccAWSAPIGatewayIntegration_basic (96.88s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	96.913s
```

### DIFF
--- a/website/source/docs/providers/aws/r/api_gateway_integration.html.markdown
+++ b/website/source/docs/providers/aws/r/api_gateway_integration.html.markdown
@@ -3,12 +3,12 @@ layout: "aws"
 page_title: "AWS: aws_api_gateway_integration"
 sidebar_current: "docs-aws-resource-api-gateway-integration"
 description: |-
-  Provides an HTTP Method Integration for an API Gateway Resource.
+  Provides an HTTP Method Integration for an API Gateway Integration.
 ---
 
 # aws\_api\_gateway\_integration
 
-Provides an HTTP Method Integration for an API Gateway Resource.
+Provides an HTTP Method Integration for an API Gateway Integration.
 
 ## Example Usage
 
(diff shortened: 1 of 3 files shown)


website/source/docs/providers/aws/r/api_gateway_integration.html.markdown
```diff
@@ -37,6 +37,10 @@ resource "aws_api_gateway_integration" "MyDemoIntegration" {
   http_method = "${aws_api_gateway_method.MyDemoMethod.http_method}"
   type        = "MOCK"
 
+  request_parameters = {
+    "integration.request.header.X-Authorization" = "'static'"
+  }
```
Ninir: This fixes the second issue

stack72:
This is fantastic! Thanks so much for taking the time to work on this - LGTM!

PR #13255, opened 2017-03-31
deps: Update github.com/joyent/triton-go/authentication

jen20:
This commit allows private key material to be used with the Triton provider, which is necessary for running acceptance tests in the HashiCorp CI environment.


stack72:
LGTM! Just tested this locally and seeing all tests pass

Issue #12758, opened 2017-03-16
provider/ns1/record SPF record creation misbehavior


parky:
### Terraform Version
`Terraform v0.9.0`
same with v0.8.8

### Affected Resource(s)
Please list the resources as a list, for example:
- ns1_record

### Terraform Configuration Files
```hcl
provider "ns1" {
  apikey = "${var.ns1_apikey}"
}
resource "ns1_zone" "tld" {
  zone = "terraform.example"
}
resource "ns1_record" "txt" {
  zone              = "${ns1_zone.tld.zone}"
  domain            = "${ns1_zone.tld.zone}"
  type              = "TXT"
  ttl               = 86400
  use_client_subnet = "true"
  answers = {
    answer = "v=DKIM1; k=rsa; p=XXXXXXXX"
  }
}
resource "ns1_record" "spf" {
  zone              = "${ns1_zone.tld.zone}"
  domain            = "${ns1_zone.tld.zone}"
  type              = "SPF"
  ttl               = 86400
  use_client_subnet = "true"
  answers = {
    answer = "v=DKIM1; k=rsa; p=XXXXXXXX"
  }
}
```

### Debug Output
https://gist.github.com/parky/f2f1c9491c459ff87c168441e843d5d9

### Panic Output
No

### Expected Behavior
`host -t SPF terraform.example dns1.p03.nsone.net` should return
terraform.example has SPF record "v=DKIM1; k=rsa; p=XXXXXXXX"

### Actual Behavior
`host -t SPF terraform.example dns1.p03.nsone.net` returns
terraform.example has SPF record "v=DKIM1;" "k=rsa;" "p=XXXXXXXX"

### Steps to Reproduce
1. `terraform apply`
2. `host -t SPF terraform.example dns1.p03.nsone.net`

### Important Factoids
Inserting SPF record lead to response where there are quotes between all parts delimited with space instead of an answer which has quotes only at the beginning and at the end.
This does not happen when the record is inserted via the web interface or via the API like:
`curl -X PUT -H 'X-NSONE-Key: myNS1apikey' -d '{"zone":"terraform.example", "domain":"terraform.example", "type":"SPF", "answers":[{"answer":["v=DKIM1; k=rsa; p=XXXXXXXX"]}]}' https://api.nsone.net/v1/zones/terraform.example/terraform.example/SPF`

radeksimko:
Closed via #13260

PR #13260, opened 2017-03-31
provider/ns1: No splitting answer on spf records.

pashap:
References https://github.com/hashicorp/terraform/issues/12758

Previously, provider was incorrectly splitting spf answers by spaces. 

Tested via `host -t SPF terraform.example dns1.p03.nsone.net` and received the expected output

### DIFF
--- a/builtin/providers/ns1/resource_record.go
+++ b/builtin/providers/ns1/resource_record.go
@@ -236,7 +236,7 @@ func resourceDataToRecord(r *dns.Record, d *schema.ResourceData) error {
 			var a *dns.Answer
 			v := answer["answer"].(string)
 			switch d.Get("type") {
-			case "TXT":
+			case "TXT", "SPF":
 				a = dns.NewTXTAnswer(v)
 			default:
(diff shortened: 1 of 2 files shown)


radeksimko:
Hi @pashap 
thanks for submitting the PR.

Would you mind attaching an acceptance test too? The test would normally go into `builtin/providers/ns1/resource_record.go` - It should be fairly easy to take one of the existing tests and modify according to our needs here for this bugfix, but let me know if you need further help in that area.

Checking the API response struct (via `testAccCheckRecordAnswerRdata`) and/or state (via `resource.TestCheckResourceAttr`) is sufficient - i.e. we don't need to wait for the DNS record to propagate nor use the DNS protocol in this context generally.

pashap:
Hi @radeksimko, thanks for pointing that out. Ill avoid PRs in the future without acceptance tests, they are very easy to write in terraform :)

I just pushed a commit with the SPF record acc test, please let me know if it looks good whenever you get a chance

Issue #9198, opened 2016-10-03
azure RM crash


spuder:
I'm suddenly getting the following terraform crash. I've successfully used this exact terraform config dozens of time over the last few days, so I'm at a loss of why terraform would suddenly start crashing
### Terraform Version

Terraform v0.7.4
### Steps to Reproduce
1. `terraform plan`
### Terraform Configuration Files

``` hcl
provider "azurerm" {
  subscription_id = "${var.subscription_id}"
  client_id = "${var.client_id}"
  client_secret = "${var.client_secret}"
  tenant_id = "${var.tenant_id}"
}

resource "azurerm_resource_group" "artifactory" {
    name = "artifactory"
    location = "West US"
}

resource "azurerm_virtual_network" "artifactory" {
    name = "artifactory-vnet"
    address_space = ["10.105.0.0/16"]
    location = "West US"
    resource_group_name = "${azurerm_resource_group.artifactory.name}"
}

resource "azurerm_subnet" "artifactory-backend" {
    name = "backend"
    resource_group_name = "${azurerm_resource_group.artifactory.name}"
    virtual_network_name = "${azurerm_virtual_network.artifactory.name}"
    address_prefix = "10.105.0.0/24"
}

resource "azurerm_subnet" "artifactory-gateway" {
    name = "GatewaySubnet"
    resource_group_name = "${azurerm_resource_group.artifactory.name}"
    virtual_network_name = "${azurerm_virtual_network.artifactory.name}"
    address_prefix = "10.105.10.0/27"
}

resource "azurerm_public_ip" "artifactory" {
  name = "artifactoryPublicIp1"
  location = "West US"
  resource_group_name = "${azurerm_resource_group.artifactory.name}"
  public_ip_address_allocation = "static"
}

resource "azurerm_network_interface" "artifactory" {
    name = "artifactory-nic"
    location = "West US"
    resource_group_name = "${azurerm_resource_group.artifactory.name}"

    ip_configuration {
        name = "artifactory-ipconfig"
        subnet_id = "${azurerm_subnet.artifactory-backend.id}"
        private_ip_address_allocation = "dynamic"
        public_ip_address_id = "${azurerm_public_ip.artifactory.id}"
    }
}

resource "azurerm_network_security_group" "artifactory" {
    name = "artifactory-nsg"
    location = "West US"
    resource_group_name = "${azurerm_resource_group.artifactory.name}"
}

resource "azurerm_network_security_rule" "artifactory_allow_https" {
[...]

PR #13261, opened 2017-03-31
provider/azurerm: Fixing crash when importing Local Network Gateways #13301

whiskeyjay:
Originally reported by @partamonov in #9198, I have created a dedicated issue for this specific crash #13301. 

The plug-in crashes if "localNetworkGateways" is not found in the id parameter. The fix is to verify the parameter before proceeding.

Also the "import" example in the documentation is wrong, "localNetworkGateways" should be case sensitive. The current example actually causes the plugin to crash due to the bug.

### DIFF
--- a/builtin/providers/azurerm/resource_arm_local_network_gateway.go
+++ b/builtin/providers/azurerm/resource_arm_local_network_gateway.go
@@ -101,6 +101,9 @@ func resourceArmLocalNetworkGatewayRead(d *schema.ResourceData, meta interface{}
 		return err
 	}
 	name := id.Path["localNetworkGateways"]
+	if name == "" {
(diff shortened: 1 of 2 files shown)


whiskeyjay:
Test:

```
make testacc TEST=./builtin/providers/azurerm TESTARGS='-run=TestAccAzureRMLocalNetworkGateway_importBasic'
2017/03/31 14:41:05 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/azurerm -v -run=TestAccAzureRMLocalNetworkGateway_importBasic -timeout 120m
=== RUN   TestAccAzureRMLocalNetworkGateway_importBasic
--- PASS: TestAccAzureRMLocalNetworkGateway_importBasic (64.68s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/azurerm    64.709s
```

PR #13270, opened 2017-04-01
Google address importability

JDiPierro:
I wanted this at work last week so I figured I'd take a stab at adding it in. :)

### DIFF
--- a/builtin/providers/google/resource_compute_address.go
+++ b/builtin/providers/google/resource_compute_address.go
@@ -14,7 +14,12 @@ func resourceComputeAddress() *schema.Resource {
 		Create: resourceComputeAddressCreate,
 		Read:   resourceComputeAddressRead,
 		Delete: resourceComputeAddressDelete,
-
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				d.Set("name", d.Id())
+				return []*schema.ResourceData{d}, nil
+			},
+		},
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
--- a/builtin/providers/google/resource_compute_global_address.go
+++ b/builtin/providers/google/resource_compute_global_address.go
@@ -14,7 +14,12 @@ func resourceComputeGlobalAddress() *schema.Resource {
 		Create: resourceComputeGlobalAddressCreate,
 		Read:   resourceComputeGlobalAddressRead,
 		Delete: resourceComputeGlobalAddressDelete,
-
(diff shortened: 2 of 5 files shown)


builtin/providers/google/resource_compute_address.go
```diff
@@ -14,7 +14,12 @@ func resourceComputeAddress() *schema.Resource {
 		Create: resourceComputeAddressCreate,
 		Read:   resourceComputeAddressRead,
 		Delete: resourceComputeAddressDelete,
-
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
paddyforan: Any reason to do a custom implementation instead of [`schema.ImportStatePassthrough`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#ImportStatePassthrough)?

builtin/providers/google/resource_compute_global_address.go
```diff
@@ -14,7 +14,12 @@ func resourceComputeGlobalAddress() *schema.Resource {
 		Create: resourceComputeGlobalAddressCreate,
 		Read:   resourceComputeGlobalAddressRead,
 		Delete: resourceComputeGlobalAddressDelete,
-
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
paddyforan: Same thing here: why not use [`schema.ImportStatePassthrough`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#ImportStatePassthrough)?

builtin/providers/google/resource_compute_address.go
JDiPierro: Google's API doesn't provide a "name" field, so the custom importer is copying the id over to name. I was thinking of making a `IdToNameImporter` that could be used generically as I saw this done in one other place but I wasn't sure where to place that.

builtin/providers/google/resource_compute_global_address.go
JDiPierro: Same as above: I tried actually using this to import a resource after making the change and the "name" field wasn't filled out so this copies ID to name.

builtin/providers/google/resource_compute_address.go
paddyforan: Oh, sorry! Sloppy reading on my part. Your solution totally works, though ideally I'd say we should just fix the resource to use the ID when reading... No matter, doesn't matter for this PR. :)

paddyforan:
I'm running another set of tests at the moment, then I'll run these tests and merge it. I just want to let 0.9.3 get out the door first, instead of doing a last-minute merge. :)

PR #13290, opened 2017-04-03
Changed output type of ips variable of ip_block ProfitBricks resource

jasminSPC:
ProfitBricks `ip_block` property `ips` now returns a list of strings instead of just a string.


stack72:
LGTM! :)

PR #13291, opened 2017-04-03
provider/aws: Migrate aws_dms_* resources away from AWS waiters

stack72:
The AWS waiter package has changed location in the 1.8.0 version of the
SDK. DMS will need to mitigate a breaking change because of this

Between @radeksimko and myself, we think that we should migrate the DMS
resources to using the Terraform state refresh func pattern that is used
across the entire of the AWS provider. DMS is the *only* resource that
currently uses the AWS waiters, so the LOE to migrate is pretty low

### DIFF
--- a/builtin/providers/aws/resource_aws_dms_endpoint.go
+++ b/builtin/providers/aws/resource_aws_dms_endpoint.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/private/waiter"
 	dms "github.com/aws/aws-sdk-go/service/databasemigrationservice"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
@@ -168,6 +167,7 @@ func resourceAwsDmsEndpointRead(d *schema.ResourceData, meta interface{}) error 
 	})
 	if err != nil {
 		if dmserr, ok := err.(awserr.Error); ok && dmserr.Code() == "ResourceNotFoundFault" {
+			log.Printf("[DEBUG] DMS Replication Endpoint %q Not Found", d.Id())
 			d.SetId("")
 			return nil
 		}
@@ -283,11 +283,6 @@ func resourceAwsDmsEndpointDelete(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	waitErr := waitForEndpointDelete(conn, d.Get("endpoint_id").(string), 30, 20)
-	if waitErr != nil {
-		return waitErr
-	}
-
 	return nil
 }
 
@@ -309,37 +304,4 @@ func resourceAwsDmsEndpointSetState(d *schema.ResourceData, endpoint *dms.Endpoi
 	d.Set("username", endpoint.Username)
 
 	return nil
-}
-
-func waitForEndpointDelete(client *dms.DatabaseMigrationService, endpointId string, delay int, maxAttempts int) error {
-	input := &dms.DescribeEndpointsInput{
-		Filters: []*dms.Filter{
-			{
-				Name:   aws.String("endpoint-id"),
-				Values: []*string{aws.String(endpointId)},
-			},
-		},
-	}
-
-	config := waiter.Config{
-		Operation:   "DescribeEndpoints",
-		Delay:       delay,
-		MaxAttempts: maxAttempts,
-		Acceptors: []waiter.WaitAcceptor{
-			{
-				State:    "success",
-				Matcher:  "path",
-				Argument: "length(Endpoints[]) > `0`",
-				Expected: false,
-			},
-		},
-	}
-
-	w := waiter.Waiter{
-		Client: client,
-		Input:  input,
-		Config: config,
-	}
-
-	return w.Wait()
 }
--- a/builtin/providers/aws/resource_aws_dms_replication_instance.go
+++ b/builtin/providers/aws/resource_aws_dms_replication_instance.go
@@ -3,11 +3,12 @@ package aws
 import (
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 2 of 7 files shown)


stack72:
https://ci.hashicorp.engineering/viewLog.html?buildId=30005&tab=buildResultsDiv

builtin/providers/aws/resource_aws_dms_endpoint.go
```diff
@@ -283,11 +282,6 @@ func resourceAwsDmsEndpointDelete(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	waitErr := waitForEndpointDelete(conn, d.Get("endpoint_id").(string), 30, 20)
-	if waitErr != nil {
-		return waitErr
-	}
-
```
catsby: Is it OK to no longer wait for deletion here? Waiting for Delete is usually a lesser concern than say waiting for creation, but it's still caused dependency issues in tear-down before

builtin/providers/aws/resource_aws_dms_replication_instance.go
```diff
@@ -174,12 +175,28 @@ func resourceAwsDmsReplicationInstanceCreate(d *schema.ResourceData, meta interf
 		return err
 	}
 
-	err = waitForInstanceCreated(conn, d.Get("replication_instance_id").(string), 30, 20)
+	//err = waitForInstanceCreated(conn, d.Get("replication_instance_id").(string), 30, 20)
+	//if err != nil {
+	//	return err
+	//}
```
catsby: dead code, can be removed?

builtin/providers/aws/resource_aws_dms_replication_instance.go
```diff
@@ -287,6 +304,23 @@ func resourceAwsDmsReplicationInstanceUpdate(d *schema.ResourceData, meta interf
 			return err
 		}
 
+		d.SetId(d.Get("replication_instance_id").(string))
```
catsby: Do we need re-set ID here, or is this just copy/pasta?

builtin/providers/aws/resource_aws_dms_replication_instance.go
```diff
@@ -355,68 +399,35 @@ func resourceAwsDmsReplicationInstanceSetState(d *schema.ResourceData, instance
...
-	}
+		if v.ReplicationInstances != nil {
+			log.Printf("[DEBUG] DMS Replication Instance status for instance %s: %s", d.Id(), *v.ReplicationInstances[0].ReplicationInstanceStatus)
+		}
 
-	w := waiter.Waiter{
-		Client: client,
-		Input:  input,
-		Config: config,
+		return v, *v.ReplicationInstances[0].ReplicationInstanceStatus, nil
```
catsby: This can panic if `v.ReplicationInstances` is `nil`, yeah? Not sure how likely that is, but seems worth guarding

builtin/providers/aws/resource_aws_dms_replication_instance.go
```diff
@@ -174,12 +175,28 @@ func resourceAwsDmsReplicationInstanceCreate(d *schema.ResourceData, meta interf
...
+	//	return err
+	//}
+
+	d.SetId(d.Get("replication_instance_id").(string))
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"creating"},
+		Target:     []string{"available"},
+		Refresh:    resourceAwsDmsReplicationInstanceStateRefreshFunc(d, meta),
+		Timeout:    d.Timeout(schema.TimeoutCreate),
```
catsby: `schema.TimeoutCreate` isn't defined, are you just leveraging the system level 20 minute default?

builtin/providers/aws/resource_aws_dms_replication_task.go
```diff
@@ -235,12 +255,26 @@ func resourceAwsDmsReplicationTaskDelete(d *schema.ResourceData, meta interface{
 
 	_, err := conn.DeleteReplicationTask(request)
 	if err != nil {
+		if dmserr, ok := err.(awserr.Error); ok && dmserr.Code() == "ResourceNotFoundFault" {
+			d.SetId("")
```
catsby: We should log with `[DEBUG]` that the resource `d.Id()` wasn't found and is being removed from state, so users can at least see in the debug log what/why this is happening

builtin/providers/aws/resource_aws_dms_endpoint.go
stack72: This isn't an issue luckily :) This wasn't necessary luckily

builtin/providers/aws/resource_aws_dms_replication_instance.go
stack72: Yessir!

builtin/providers/aws/resource_aws_dms_replication_instance.go
stack72: Copy/pasta issue :)

builtin/providers/aws/resource_aws_dms_replication_task.go
stack72: Done :)

builtin/providers/aws/resource_aws_dms_replication_instance.go
stack72: Caught above :)

stack72:
@catsby changes made as requested :)

builtin/providers/aws/resource_aws_dms_endpoint.go
catsby: 👍

builtin/providers/aws/resource_aws_dms_replication_instance.go
catsby: 👍

builtin/providers/aws/resource_aws_dms_replication_instance.go
catsby: I _swear_ that was `!=` above... did you rebase? ;)

builtin/providers/aws/resource_aws_dms_replication_task.go
catsby: I don't see it

stack72:
Last log line added, @catsby - will merge on green :)

PR #13293, opened 2017-04-03
Contentful provider

lneves75:
Adding a new provider for Contentful, a content as a service platform with a collection of APIs that help developers manage, integrate, and deliver digital content across all types of clients, devices, and platforms.

This provider manages resources such as:

1. Spaces,
2. Content types
3. Webhooks
4. Content delivery API tokens
5. Locales


lneves75:
@radeksimko any chance we get a review soonish ? If not, any estimation for when would it be possible ?

PR #13311, opened 2017-04-04
fix integration between ALB and ECS

KensoDev:
For our ECS service definition we have this snippet at the
`load_balancer`.

The `target_group_arn` is being pupulated by an external service that
returns the arn based on a simple string from our microservices list.

If the arn changed, this would not cause a recreation of the service and
leaving a dangling pointer to an arn that does not exist anymore.

```
  load_balancer {
    target_group_arn  = "${lookup(var.target_group_mapping, element(values(var.microservices), count.index))}"
    container_name    = "${element(values(var.microservices), count.index)}"
    container_port    = "${var.container_port}"
  }
```

The fix is adding another field to the set that's creating the ELB/ALB
definition. From looking into the git history seems this code was
created prior to ALB thus not having this field available at the time.

Service is being recreated as expected, no other services are affected
(expected behavior)

### DIFF
--- a/builtin/providers/aws/resource_aws_ecs_service.go
+++ b/builtin/providers/aws/resource_aws_ecs_service.go
@@ -510,6 +510,8 @@ func resourceAwsEcsServiceDelete(d *schema.ResourceData, meta interface{}) error
 func resourceAwsEcsLoadBalancerHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
+
+	buf.WriteString(fmt.Sprintf("%s-", m["target_group_arn"].(string)))
 	buf.WriteString(fmt.Sprintf("%s-", m["elb_name"].(string)))
 	buf.WriteString(fmt.Sprintf("%s-", m["container_name"].(string)))
 	buf.WriteString(fmt.Sprintf("%d-", m["container_port"].(int)))
--- a/builtin/providers/aws/resource_aws_ecs_service.go
+++ b/builtin/providers/aws/resource_aws_ecs_service.go
@@ -511,6 +511,8 @@ func resourceAwsEcsLoadBalancerHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
 
+	// There are 2 types of load balancers: ELB/ALB
+	// In case of ELB `elb_name` is expected and for ALB `target_group_arn` is.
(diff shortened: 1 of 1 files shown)


builtin/providers/aws/resource_aws_ecs_service.go
```diff
@@ -510,6 +510,8 @@ func resourceAwsEcsServiceDelete(d *schema.ResourceData, meta interface{}) error
 func resourceAwsEcsLoadBalancerHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
+
+	buf.WriteString(fmt.Sprintf("%s-", m["target_group_arn"].(string)))
```
stack72: Hey @KensoDev 

Thanks for the work here. I think a very slightly better way to handle this would be to check if `target_group_arn` is actually set before adding it to the hash

I suggest the following:

```
if m["target_group_arn"] != nil {
		buf.WriteString(fmt.Sprintf("%s-", m["target_group_arn"].(string)))
	}
```

This means that the hash will not change for users who have not got that value set

Thoughts?

Paul

KensoDev: I can make that change. I followed the convention with elb_name which is not set if it's an ALB. In which case you will get empty and the hash should not change.

Anyway, I will verify and see what can be done to harden this a bit.

KensoDev:
@stack72 Done! 
LMK if there's anything else required to merge this in.

Right now we are using a patched version in order to bypass this bug.

stack72:
Hi @KensoDev 

Thanks for all the work and being so fast on the feedback changes. I made 1 further change here - ee9c84cf7

We decided this was a better way to handle it to stop anyone from being affected with their existing configs

Paul

KensoDev:
Thanks @stack72 

I have to say I don't fully understand the effect this will have on previous configurations, thank you for being on top of this.

stack72:
Hey @KensoDev 

So, if a user has a current state with a hash of x, and we change how that has is made, then the new hash, y will tell terraform that a change is needed

We discussed this in the team and realised we didn't want to trigger any bad experience on people :)

But your change works and will allow people's updates when the target_group changes :)

Paul

KensoDev:
Got it.
That makes sense actually. since this is forcing a new resource by default.

Thanks for the explanation

PR #13312, opened 2017-04-04
Add `min_size` and `max_size` ASG fields

CaseyLeask:
These are required fields
```
$ terraform plan
2 error(s) occurred:

* aws_autoscaling_group.bar: "max_size": required field is not set
* aws_autoscaling_group.bar: "min_size": required field is not set
```


stack72:
LGTM! Thanks for this

Issue #13334, opened 2017-04-04
Cannot refresh when openstack_compute_volume_attach_v2 has been manually deleted


dannytrigo:
### Terraform Version
v0.9.2

### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_compute_volume_attach_v2

### Expected Behavior
I have a volume, an instance and I attach the volume to the instance with openstack_compute_volume_attach_v2.  I then manually delete the instance.
I would expect that the next refresh will detect that the instance is gone, and be removed from the state, and that the openstack_compute_volume_attach_v2 resource is also gone as the attachment is to an instance that has been deleted.

### Actual Behavior
Instead of the openstack_compute_volume_attach_v2 just being removed from the state as it no longer exists, Terraform exits with the error:

* module.int.module.minions.openstack_compute_volume_attach_v2.openshift_node_vol_1[2]: openstack_compute_volume_attach_v2.openshift_node_vol_1.2: Resource not found

A refresh/plan/apply cannot be performed without manually removing the attach resource from the state file.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Manually delete the instance
3. `terraform refresh`

stack72:
Closed via #13342

PR #13342, opened 2017-04-04
Do not error if compute volume attachment is missing

dannytrigo:
https://github.com/hashicorp/terraform/issues/13334


jtopjian:
Thanks for this. I'll test this out later this evening :)

jtopjian:
@stack72 This looks good to me!

@dannytrigo Thanks again for this.

PR #13344, opened 2017-04-04
aws provider: specify that only elastic load balancers can be defined for `load_balancers`

brandongalbraith:
`load_balancer` description leaves a bit of ambiguity as to whether an ALB name can be provided; it cannot!


catsby:
Thanks!

PR #13348, opened 2017-04-04
provider/aws: Fix KMS Key reading with Exists method

### LINKED ISSUE #13322: aws_kms_key : cannot move between regions because of incomplete state
### Terraform Version

0.92

### Affected Resource(s)
- aws_kms_key

### Expected Behavior

* Changed the provider on an existing `aws_kms_key` resource to one with a different region
* Expected `terraform apply` to cope with removing key from region 1 and recreating in region 2

### Actual Behavior

Terraform composites an ARN for the key from the provider region and the stored ID value in the state.

Therefore if you take an action that changes the region of the resource, it cannot be found during state discovery and you get

```
* aws_kms_key.healthcheck: aws_kms_key.healthcheck: NotFoundException: Key 'arn:aws:kms:us-east-1:<censored>:key/<censored>' does not exist
	status code: 400, request id: 2928add4-1927-11e7-85b2-41bf5eb3228a
```

- key is actually in `eu-west-2`, not `us-east-1` as in the ARN in the error message

### Steps to Reproduce

Create an `aws_kms_key` resource in one region, then use a provider attribute to change the region of that resource and re-apply.

### Suggested resolution

Store the full ARN of key resources in the state instead of compositing them from ID and region.

### Workaround

? Manually create a new key and edit the state file to point to that one instead

catsby:
Fixes #13322 by checking if the key Exists and offering to recreate if
not found, or pending delete

### DIFF
--- a/builtin/providers/aws/resource_aws_kms_key.go
+++ b/builtin/providers/aws/resource_aws_kms_key.go
@@ -136,13 +136,6 @@ func resourceAwsKmsKeyRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	resp, err := conn.DescribeKey(req)
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
-			if awsErr.Code() == "NotFoundException" {
-				log.Printf("[WARN] Removing KMS key %s: Key not found", d.Id())
-				d.SetId("")
-				return nil
-			}
-		}
 		return err
 	}
 	metadata := resp.KeyMetadata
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_kms_key.go
```diff
@@ -134,6 +136,13 @@ func resourceAwsKmsKeyRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	resp, err := conn.DescribeKey(req)
 	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() == "NotFoundException" {
+				log.Printf("[WARN] Removing KMS key %s: Key not found", d.Id())
```
catsby: Yeah that makes sense, thanks, removed.

Issue #13263, opened 2017-03-31
Support TreatMissingData for aws_cloudwatch_metric_alarm


grimm26:
### Terraform Version
I'm running 0.8.8

### Affected Resource(s)
aws_cloudwatch_metric_alarm

I want to be able to specify the TreatMissingData parameter as specified at https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricAlarm.html

grimm26:
FYI, I just started to try my hand at generating a PR for this and found that aws-cli doesn't even currently support this parameter.

PR #13358, opened 2017-04-05
provider/aws: Add support for treat_missing_data to cloudwatch_metric_alarm

stack72:
Fixes: #13263

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSCloudWatchMetricAlarm'
2017/04/05 08:51:06 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSCloudWatchMetricAlarm -timeout 120m
=== RUN   TestAccAWSCloudWatchMetricAlarm_importBasic
--- PASS: TestAccAWSCloudWatchMetricAlarm_importBasic (23.93s)
=== RUN   TestAccAWSCloudWatchMetricAlarm_basic
--- PASS: TestAccAWSCloudWatchMetricAlarm_basic (27.81s)
=== RUN   TestAccAWSCloudWatchMetricAlarm_treatMissingData
--- PASS: TestAccAWSCloudWatchMetricAlarm_treatMissingData (43.39s)
=== RUN   TestAccAWSCloudWatchMetricAlarm_extendedStatistic
--- PASS: TestAccAWSCloudWatchMetricAlarm_extendedStatistic (26.80s)
=== RUN   TestAccAWSCloudWatchMetricAlarm_missingStatistic
--- PASS: TestAccAWSCloudWatchMetricAlarm_missingStatistic (5.95s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	127.899s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_cloudwatch_metric_alarm.go
+++ b/builtin/providers/aws/resource_aws_cloudwatch_metric_alarm.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsCloudWatchMetricAlarm() *schema.Resource {
@@ -95,6 +96,12 @@ func resourceAwsCloudWatchMetricAlarm() *schema.Resource {
 				Optional:      true,
 				ConflictsWith: []string{"statistic"},
 			},
+			"treat_missing_data": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validation.StringInSlice([]string{"breaching", "notBreaching", "ignore", "missing"}, true),
+			},
 		},
 	}
 }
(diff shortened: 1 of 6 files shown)


builtin/providers/aws/resource_aws_cloudwatch_metric_alarm.go
```diff
@@ -95,6 +96,12 @@ func resourceAwsCloudWatchMetricAlarm() *schema.Resource {
 				Optional:      true,
 				ConflictsWith: []string{"statistic"},
 			},
+			"treat_missing_data": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
```
radeksimko: I'm wondering if `Default: "missing"` would be more suitable here 🤔  That would make people put that field if it drifts from the default value, rather than leave it out and let anyone mess with that outside of Terraform.

radeksimko: btw. new `Default` would probably also require state migration - not sure what's the best option here, but generally speaking I'd try to use `Default` over `Compute` (if there's an option) where possible. I think it makes the behaviour more predictable.

radeksimko: but it's certainly not a blocker - just some food for thought.

w32-blaster: The default in web console is also "missing" so I think it make sense to keep "missing"

stack72: let me think more about this one

PR #13396, opened 2017-04-05
DOCS: Updated list of triggers for codedeploy_deployment_group

nwalke:
Might be better just to refer the user to the AWS docs since these aren't specified by Terraform at all.

If there's any thoughts on that, just let me know and I'll make the change.


radeksimko:
> Might be better just to refer the user to the AWS docs since these aren't specified by Terraform at all.

Yes, you're absolutely right. I think we could mention a few examples and then include a link to the official docs. Would you mind making that change? Thanks!

nwalke:
Should be good to go!

PR #13399, opened 2017-04-05
P aws tests

mbfrahry:
Two more aws tests now pass. 
```
make testacc TEST=./builtin/providers/aws/ TESTARGS='-run=TestAccAWSEcsDataSource_ecsTaskDefinition'
2017/04/05 08:57:18 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws/ -v -run=TestAccAWSEcsDataSource_ecsTaskDefinition -timeout 120m
=== RUN   TestAccAWSEcsDataSource_ecsTaskDefinition
--- PASS: TestAccAWSEcsDataSource_ecsTaskDefinition (21.97s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	21.997s
```

```
make testacc TEST=./builtin/providers/aws/ TESTARGS='-run=TestAccAWSAutoScalingGroup_ALB_TargetGroups'
2017/04/05 15:56:20 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws/ -v -run=TestAccAWSAutoScalingGroup_ALB_TargetGroups -timeout 120m
--- PASS: TestAccAWSAutoScalingGroup_ALB_TargetGroups (116.55s)
=== RUN   TestAccAWSAutoScalingGroup_ALB_TargetGroups_ELBCapacity
--- PASS: TestAccAWSAutoScalingGroup_ALB_TargetGroups_ELBCapacity (363.60s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	480.180s```

### DIFF
--- a/builtin/providers/aws/resource_aws_autoscaling_group.go
+++ b/builtin/providers/aws/resource_aws_autoscaling_group.go
@@ -446,7 +446,6 @@ func resourceAwsAutoscalingGroupRead(d *schema.ResourceData, meta interface{}) e
 	d.Set("health_check_type", g.HealthCheckType)
 	d.Set("launch_configuration", g.LaunchConfigurationName)
 	d.Set("load_balancers", flattenStringList(g.LoadBalancerNames))
-	d.Set("target_group_arns", flattenStringList(g.TargetGroupARNs))
 
 	if err := d.Set("suspended_processes", flattenAsgSuspendedProcesses(g.SuspendedProcesses)); err != nil {
 		log.Printf("[WARN] Error setting suspended_processes for %q: %s", d.Id(), err)
--- a/builtin/providers/aws/resource_aws_autoscaling_group_test.go
+++ b/builtin/providers/aws/resource_aws_autoscaling_group_test.go
@@ -445,15 +445,6 @@ func TestAccAWSAutoScalingGroup_ALB_TargetGroups(t *testing.T) {
 						"aws_autoscaling_group.bar", "target_group_arns.#", "1"),
 				),
 			},
-
-			resource.TestStep{
-				Config: testAccAWSAutoScalingGroupConfig_ALB_TargetGroup_pre,
-				Check: resource.ComposeAggregateTestCheckFunc(
-					testAccCheckAWSAutoScalingGroupExists("aws_autoscaling_group.bar", &group),
-					resource.TestCheckResourceAttr(
-						"aws_autoscaling_group.bar", "target_group_arns.#", "0"),
-				),
-			},
 		},
 	})
 }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: builtin/providers/aws/resource_aws_autoscaling_group.go [hunk 1]
```diff
@@ -446,7 +446,6 @@ func resourceAwsAutoscalingGroupRead(d *schema.ResourceData, meta interface{}) e
 	d.Set("health_check_type", g.HealthCheckType)
 	d.Set("launch_configuration", g.LaunchConfigurationName)
 	d.Set("load_balancers", flattenStringList(g.LoadBalancerNames))
-	d.Set("target_group_arns", flattenStringList(g.TargetGroupARNs))
```
radeksimko: Can you explain the removal here, please? I think it would break things in the sense that Terraform wouldn't be able to detect drifts for this field from the state anymore.

builtin/providers/aws/resource_aws_autoscaling_group_test.go
```diff
@@ -445,15 +445,6 @@ func TestAccAWSAutoScalingGroup_ALB_TargetGroups(t *testing.T) {
...
 			},
-
-			resource.TestStep{
-				Config: testAccAWSAutoScalingGroupConfig_ALB_TargetGroup_pre,
-				Check: resource.ComposeAggregateTestCheckFunc(
-					testAccCheckAWSAutoScalingGroupExists("aws_autoscaling_group.bar", &group),
-					resource.TestCheckResourceAttr(
-						"aws_autoscaling_group.bar", "target_group_arns.#", "0"),
-				),
-			},
```
radeksimko: I'm not sure if it's a good idea just to remove a failing test here. 🤔 

Looking at the failure

```
--- FAIL: TestAccAWSAutoScalingGroup_ALB_TargetGroups (38.42s)
    testing.go:280: Step 3 error: Check failed: 1 error(s) occurred:
        
        * Check 2/2 error: aws_autoscaling_group.bar: Attribute 'target_group_arns.#' expected "0", got "1"
```
I suspect this may be caused by the fact that we don't randomize ALB target group names and these in turn remain attached to ASGs whilst other tests are running in parallel. i.e. how do you feel about randomizing the names there instead?

builtin/providers/aws/resource_aws_autoscaling_group.go
mbfrahry: This was duplicated here https://github.com/hashicorp/terraform/blob/1f13ff22a3f4961782e3c4793510a4d966169e9a/builtin/providers/aws/resource_aws_autoscaling_group.go#L453 so I removed the redundant piece

radeksimko: I see, I didn't notice the second occurrence before. That's good then 👍

builtin/providers/aws/resource_aws_autoscaling_group_test.go
radeksimko: As discussed on Slack this is failing because at the core/schema level we don't handle `Optional`+`Computed` Sets very well when removing last items from them.

It's not great, but I'm ok with merging this then 👍 We'll resolve the core bug in a separate PR eventually.

PR #13402, opened 2017-04-05
provider/google: Add node_pool field in resource_container_cluster

danawillow:
As @paddyforan and I discussed offline, even though nodepools could be their own resource, it makes more sense to keep track of them inside the cluster resource. Each nodepool belongs to exactly one cluster, some of the update methods on nodepools are actually part of the clusters API, and the creation of a cluster without a specified nodepool will create a default one that we might want terraform to be able to manage.

For this PR, I've just included the minimum functionality. In follow-ups I'll add the ability to add and remove nodepools from a cluster and add more fields to the node_pool schema.

### DIFF
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"regexp"
 
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"google.golang.org/api/container/v1"
 	"google.golang.org/api/googleapi"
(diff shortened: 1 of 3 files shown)


imkira:
@danawillow  It would be nice to have the ability to configure as one would do with the default pool otherwise you get something like:

```
googleapi: Error 400: It's invalid to specify both cluster.node_config and a node pool. Please only provide a node pool.., badRequest
```

Maybe adding `config` (like `node_config` under `google_container_cluster` resource) to the `node_pool` object would be nice:

https://godoc.org/google.golang.org/api/container/v1#NodePool

danawillow:
@imkira Yup, that's exactly the plan!

PR #13413, opened 2017-04-06
Updated provider documentation to highlight limitations of interpolat…

nicholasjackson:
Small update to the provider documentation highlighting the limitations of the interpolation syntax.


website/source/docs/configuration/providers.html.md
```diff
@@ -117,3 +117,14 @@ KEY {
...
+## Interpolation
+Providers support [interpolation syntax](/docs/configuration/interpolation.html) allowing dynamic configuration at run time.
+
+```
+provider "aws" {
+	region = "${var.aws_region}"
+}
+```
+
+Only [variables](/docs/configuration/variables) and [remote state](/docs/state/remote.html) are supported, it is not possible to use the output from a resource, module or data source in the interpolation syntax for a provider.
```
radeksimko: > and [remote state](/docs/state/remote.html) are supported

Not sure what exactly do you mean by this. 🤔 

Also I'd try to be a bit more positive in the wording - i.e. `are supported` **`at this point`** (or _currently_). There's at least one open issue about this and it's something we eventually want to support:
https://github.com/hashicorp/terraform/issues/4149

nicholasjackson: Thanks, I was just browsing some of the other docs I think this should be a note also.  

I was trying to make a note that you can't at the moment use outputs in the syntax, I was not really sure how to write this I will do another update after lunch.

website/source/docs/configuration/providers.html.md
```diff
@@ -117,3 +117,14 @@ KEY {
 	CONFIG
 }
 ```
+
+## Interpolation
+Providers support [interpolation syntax](/docs/configuration/interpolation.html) allowing dynamic configuration at run time.
+
+```
```
sethvargo: tag this as hcl so it'll syntax highlight (when we enable that)

website/source/docs/configuration/providers.html.md
```diff
@@ -117,3 +117,14 @@ KEY {
...
+provider "aws" {
+	region = "${var.aws_region}"
```
sethvargo: two spaces (looks like a tag)

website/source/docs/configuration/providers.html.md
sethvargo: This should be `/docs/configuration/variables.html` (missing `.html`)

sethvargo: I agree with @radeksimko - I think we can change the tone a bit here, and do some education in the process. Consider:

Because providers are one of the first things loaded when Terraform parses the graph, it is not possible to use the output from modules, resources, or data sources as inputs to the provider. At this time, only [variables](link) and [remote state](link) may be used in an interpolation inside a `provider` stanza.

website/source/docs/configuration/providers.html.md
nicholasjackson: I have update this in the latest commit.

website/source/docs/configuration/providers.html.md
nicholasjackson: Good catch, I need to update my editor settings, I normally always use 2 spaces in place of a tab and not sure why this did not get picked up for the markdown.

website/source/docs/configuration/providers.html.md
nicholasjackson: I have updated the comments to update the description, also ran some tests this morning to confirm that remote state works as I thought.  Not sure if it is worth adding the example?
```hcl
data "terraform_remote_state" "nomad" {
  backend = "s3"

  config {
    bucket  = "nic-terraform-state"
    key     = "examples/nomad/terraform.state"
    region  = "eu-west-1"
    profile = "hashicorp"
  }
}

provider "nomad" {
  address = "${format("http://%s:4646",data.terraform_remote_state.nomad.alb_nomad)}"
}
```

PR #13422, opened 2017-04-06
providers/aws: Update ElasticTranscoderPreset to have default for MaxFrameRate

catsby:
Updates `video.max_frame_rate` attribute to supply a default of `30`, matching our Docs and AWS docs:

- http://docs.aws.amazon.com/elastictranscoder/latest/developerguide/create-preset.html

Fixes test `TestAccAWSElasticTranscoderPreset_basic`

### DIFF
--- a/builtin/providers/aws/resource_aws_elastic_transcoder_preset.go
+++ b/builtin/providers/aws/resource_aws_elastic_transcoder_preset.go
@@ -213,6 +213,7 @@ func resourceAwsElasticTranscoderPreset() *schema.Resource {
 						"max_frame_rate": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
+							Default:  "30",
 							ForceNew: true,
 						},
 						"max_height": &schema.Schema{


stack72:
@catsby as this now has a default, and is marked as ForceNew: true, will this cause any issues for some users? Do we need a State Migration for those who don't have it set?

catsby:
@stack72 The API defaults it to `30`. Any user omitting `max_frame_rate` is getting a `ForceNew` diff . Any user specifying `max_frame_rate` should see no change.

stack72:
@catsby nice :) LGTM then

svanzoest:
This broke https://github.com/hashicorp/terraform/pull/11340, I can no longer set a preset with a fixed frame rate.

dcosson:
I also ran into this problem and can't set a fixed frame rate.

Is the fix as simple as reverting this single line?

I can't tell since there's not really enough context here explaining what the purpose of this PR was. Why should terraform ever have its own opinion on a default value if AWS specifically allows that value to be omitted?

svanzoest:
@dcosson applying https://github.com/terraform-providers/terraform-provider-aws/issues/695 will address this issue. I think the only reason it got added by terraform as it made the tests easier.

dcosson:
Got it, thanks for updating with the more in-depth comment.

PR #3764, opened 2015-11-05
Google Cloud: BigQuery DataSet and Table resources

coffeepac:
BigQuery is a managed database-like service.  It supports SQL as
a query language and nearly infinite scale.  This PR adds support
to manage tables and datasets.

PR #13436, opened 2017-04-06
provider/google: BigQuery Dataset

heimweh:
This PR adds support for BigQuery datasets.

Heavily inspired by (https://github.com/hashicorp/terraform/pull/3764) https://github.com/22Acacia/terraform-provider-googlebigquery.

Big thanks to @coffeepac :star: for submitting the initial PR. This PR might also lack some of the functionality in the initial PR and I'm quite new to GCP so any feedback is more than welcome 😺

### DIFF
--- a/builtin/providers/google/config.go
+++ b/builtin/providers/google/config.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
+	bigquery "google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/cloudbilling/v1"
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/compute/v1"
@@ -42,6 +43,7 @@ type Config struct {
 	clientSqlAdmin        *sqladmin.Service
 	clientIAM             *iam.Service
 	clientServiceMan      *servicemanagement.APIService
+	clientBigQuery        *bigquery.Service
 }
 
 func (c *Config) loadAndValidate() error {
@@ -168,6 +170,13 @@ func (c *Config) loadAndValidate() error {
 		return err
 	}
 	c.clientBilling.UserAgent = userAgent
+
+	log.Printf("[INFO] Instantiating Google Cloud BigQuery Client...")
+	c.clientBigQuery, err = bigquery.New(client)
+	if err != nil {
+		return err
+	}
+	c.clientBigQuery.UserAgent = userAgent
 
 	return nil
 }
--- a/builtin/providers/google/config.go
+++ b/builtin/providers/google/config.go
@@ -13,7 +13,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
-	bigquery "google.golang.org/api/bigquery/v2"
+	"google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/cloudbilling/v1"
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/compute/v1"
--- a/builtin/providers/google/resource_bigquery_dataset.go
+++ b/builtin/providers/google/resource_bigquery_dataset.go
@@ -99,13 +99,6 @@ func resourceBigQueryDataset() *schema.Resource {
 				},
 			},
 
-			// Skip destroying the resource
-			"retain_on_delete": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  false,
-			},
-
 			// Labels: [Experimental] The labels associated with this dataset. You
 			// can use these to organize and group your datasets. You can set this
 			// property when inserting or updating a dataset.
@@ -278,12 +271,6 @@ func resourceBigQueryDatasetUpdate(d *schema.ResourceData, meta interface{}) err
 
 func resourceBigQueryDatasetDelete(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
-
-	if d.Get("retain_on_delete").(bool) {
-		log.Printf("[WARN] Removing BigQuery dataset ID %q with `retain_on_delete` set. Please delete this dataset manually.", d.Id())
-		d.SetId("")
-		return nil
-	}
 
 	log.Printf("[INFO] Deleting BigQuery dataset: %s", d.Id())
 
--- a/builtin/providers/google/resource_bigquery_dataset.go
+++ b/builtin/providers/google/resource_bigquery_dataset.go
@@ -1,0 +1,298 @@
+package google
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"google.golang.org/api/bigquery/v2"
+	"google.golang.org/api/googleapi"
+)
+
+func resourceBigQueryDataset() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceBigQueryDatasetCreate,
+		Read:   resourceBigQueryDatasetRead,
+		Update: resourceBigQueryDatasetUpdate,
+		Delete: resourceBigQueryDatasetDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			// DatasetId: [Required] A unique ID for this dataset, without the
+			// project name. The ID must contain only letters (a-z, A-Z), numbers
+			// (0-9), or underscores (_). The maximum length is 1,024 characters.
+			"dataset_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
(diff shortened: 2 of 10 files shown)


kurtmaile:
Cool exciting thanks so much for contributing this PR - will this be in next release?

builtin/providers/google/resource_bigquery_dataset_test.go
```diff
@@ -0,0 +1,118 @@
...
+func testAccCheckBigQueryDatasetDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "google_bigquery_dataset" {
+			continue
+		}
+
+		_, err := config.clientBigQuery.Datasets.Get(
+			config.Project,
```
danawillow: Optional style nitpick: I actually think this would look nicer on just one line

builtin/providers/google/resource_bigquery_dataset_test.go
```diff
@@ -0,0 +1,118 @@
...
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		config := testAccProvider.Meta().(*Config)
+
+		found, err := config.clientBigQuery.Datasets.Get(
```
danawillow: likewise

website/source/docs/providers/google/r/bigquery_dataset.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+
+    Possible values include `EU` and `US`. The default value is `US`.
+
+    Changing this forces a new resource to be created.
+
+* `default_table_expiration_ms` - (Optional) The default lifetime of all
+    tables in the dataset, in milliseconds.
+
+    The minimum value is 3600000
+    milliseconds (one hour).
```
danawillow: nit- move this line to the previous

website/source/layouts/google.erb
```diff
@@ -233,6 +233,17 @@
       </li>
     </ul>
     </li>
+    <li<%= sidebar_current("docs-google-bigquery") %>>
+    <a href="#">Google BigQuery Resources</a>
```
danawillow: Let's also keep this list alphabetical

builtin/providers/google/resource_bigquery_dataset.go
```diff
@@ -0,0 +1,297 @@
...
+					value := v.(int)
+					if value < 3600000 {
+						errors = append(errors, fmt.Errorf("%q cannot be shorter than 3600000", k))
+					}
+
+					return
+				},
+			},
+
+			// Skip destroying the resource
```
coffeepac: My use case for this is that I may want to create a bigquery output and keep it around after I destroyed the processing framework.  My use case at the time was I was creating and destroying a big processing framework on a regular interval and wanted to keep the outputs but I had to create the outputs (big query resources) so I was creating the bq resources in TF.

heimweh:
Hi @danawillow,

that makes a lot of sense, sorry about that :/
I'd be more than happy to split this PR into two, if that'll help reviewing :)

Also, thanks for the feedback. I'll look into this asap.

builtin/providers/google/resource_bigquery_dataset.go
```diff
@@ -0,0 +1,297 @@
...
+			// removed before the table expires, or if you provide an explicit
+			// expirationTime when creating a table, that value takes precedence
+			// over the default expiration time indicated by this property.
+			"default_table_expiration_ms": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(int)
+					if value < 3600000 {
+						errors = append(errors, fmt.Errorf("%q cannot be shorter than 3600000", k))
```
heimweh: `%q` in this case is the field name. Example error:
```bash
* google_bigquery_dataset.foo: "dataset_id" cannot be greater than 1,024 characters
```

I can change it to `%v` if that makes more sense.

builtin/providers/google/resource_bigquery_dataset.go
heimweh: To us, this was kept as an option to make Terraform forget a dataset rather than actually destroying it, but since this behavior can be accomplished with something like `taint` instead I would be totally okay with removing this option.

builtin/providers/google/resource_bigquery_dataset.go
```diff
@@ -0,0 +1,297 @@
...
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	dataset, err := resourceDataset(d, meta)
+	if err != nil {
+		return err
+	}
+
```
heimweh: Makes sense, I added more log statements.

builtin/providers/google/resource_bigquery_dataset.go
heimweh: Another alternative is to keep this option but to be explicit in the documentation and/or add a field warning that this should be used with caution.

builtin/providers/google/resource_bigquery_dataset.go
heimweh: I added a better error message.

`%q` in this case is the field name: `"default_table_expiration_ms"`.

The actual error would look like this:

```bash
$ terraform plan
1 error(s) occurred:

* google_bigquery_dataset.foo: "default_table_expiration_ms" cannot be shorter than 3600000 milliseconds (one hour)
```

heimweh:
@danawillow I'm splitting this PR into two separate PRs.

This PR should now only contain the `dataset` resource and vendors BigQuery (added as a separate commit).

I'll open up another PR for the `table` resource and add a link here when it's up.

builtin/providers/google/config.go
```diff
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
+	bigquery "google.golang.org/api/bigquery/v2"
```
heimweh: That's correct, not sure why that ended up being named. I'll push a fix for that straight away.

builtin/providers/google/resource_bigquery_dataset.go
heimweh: I merely used it while testing BigQuery so I don't really have a valid use case here.
Maybe this option can be in a possible follow up PR instead and we'll leave this option out for now?

coffeepac: Use Case:
 a TF job is running on a schedule and it saves state off to atlas.  this state file will exist forever but the resources will age out at periodic intervals, ie one day, one week, etc.  The resource we would want to be deleted are the processing resources (dataflow jobs) but we would want the output resources to exist forever (big query datasets).  we could leave the output resources in the config file forever but that will get excessively verbose.

danawillow:
The new changes need reformatting (you can run `make fmt`), but other than that this is good to go!

builtin/providers/google/resource_bigquery_dataset.go
danawillow: Thanks @coffeepac, that makes sense. I'm still inclined to have that change in a follow-up PR to make it easier for other people to comment on and offer opinions and whatnot, also that way we can get this one nice and checked in.

coffeepac: makes sense.  its an unusual option for TF.

heimweh:
@danawillow should be fixed now :) thank you so much for reviewing.

kurtmaile:
Hi all,  Any plans to provide a resource for bigquery views?  Would be super helpful!!

Thanks for your hard work on these core resources.....

builtin/providers/google/resource_bigquery_dataset.go
```diff
@@ -0,0 +1,285 @@
...
+			// deleted automatically. If a table's expirationTime is modified or
+			// removed before the table expires, or if you provide an explicit
+			// expirationTime when creating a table, that value takes precedence
+			// over the default expiration time indicated by this property.
+			"default_table_expiration_ms": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(int)
+					if value < 3600000 {
```
mdpilam: Is it possible to define the default expiration as "Never"? I know that's the default, but we have a Terraform script that we want Never for most datasets, but 30days for some and we can't specify "Never" or 0 - even though the documentation says setting "0" in the CLI will set it to Never: https://cloud.google.com/bigquery/docs/updating-datasets#table-expiration

danawillow: if you don't have the ability to not set it, could you conditionally set it to null? https://www.terraform.io/docs/configuration/expressions.html#null

mdpilam: I tried and got an error: 

> Unknown token: 42:17 IDENT null

danawillow: are you using terraform 0.12? if not, then null won't work.

mdpilam: unfortunately not; it's in our plan to upgrade

Issue #12291, opened 2017-02-27
Potential Race Condition When Apply Google Project IAM Policy


zachgersh:
### Terraform Version
`0.85+`

### Affected Resource(s)
- google_projec_iam_policy

### Terraform Configuration Files
None provided. Totally can work up an example if needed.

### Expected Behavior
I should be able to remove/add members to the policy

### Actual Behavior
I get random permission denied errors due to the fact that the policy has become out of date (a member was added or removed before the policy was destroyed).

Spoke to @evandbrown a bit about this and I am unsure if restoring a saved copy of the previous policy that is stored in your tfstate file is the right way to go. It would seem that anyone with multiple environments / policy changes in the same `project` would end up hitting this problem.

### Steps to Reproduce
1. create two different policies in the same GCP project
2. start to teardown one and then begin tearing down the other
3. watch it fail

cc @paddyforan as well whose been looking over a bunch of my GCP changes.

kkallday:
@madamkiwi and I have been able to reproduce this issue:

template 1: https://gist.github.com/kkallday/5d9127f37b33ab660869a3f4011492ee#file-template-1-tf
template 2: https://gist.github.com/kkallday/5d9127f37b33ab660869a3f4011492ee#file-template-2-tf

1. `terraform apply -state=some-dir` with template 1
1.  `terraform apply -state=some-OTHER-dir` with template 2
1. `terraform destroy -state=some-dir`
1. `terraform destroy -state=some-OTHER-dir` and you will receive the following error:

```
google_service_account.opsman_service_account: Refreshing state... (ID: projects/...ount.com)
data.google_iam_policy.opsman_policy: Refreshing state...
google_project_iam_policy.opsman_iam_policy: Refreshing state... (ID: YOUR-PROJECT-ID)
google_project_iam_policy.opsman_iam_policy: Destroying... (ID: YOUR-PROJECT-ID)
Error applying plan:

1 error(s) occurred:

* google_project_iam_policy.opsman_iam_policy (destroy): 1 error(s) occurred:

[...]

PR #13439, opened 2017-04-06
fixes issue where IAM policies get deleted due to race condition

kkallday:
This fixes the issue outlined in #12291. We wanted to write a test for this code however, we had trouble getting the test to pass without our changes. Please let us know what we can do to get this PR merged as it has prevented us from moving to versions greater than 0.8.4.

[#136756369]

Signed-off-by: Kalai Wei <kawei@pivotal.io>


paddycarver:
This should be resolved by terraform-providers/terraform-provider-google#171. If you continue to have issues or problems with it, please open issues in the new terraform-providers/terraform-provider-google repository. Thanks!

PR #13441, opened 2017-04-07
Fix validation of the `name_prefix` parameter of the `aws_alb` resource

joshuaspence:
This parameter is being validated using the wrong validation function, which means that we are incorrectly disallowing a `name_prefix` value ending with a dash.


stack72:
LGTM!

PR #12398, opened 2017-03-03
provider/google: Add support for attaching/detaching a disk to/from an instance

danawillow:
Fixes #6678.

This currently has a few limitations, but I think it's good enough to handle the most common uses. The limitations are:
- Trying to attach/detach an identical disk (I tried to write a test that exposes this but couldn't get one that was accepted, so maybe this is an impossible scenario)
- Trying to detach a disk and delete it at the same time (the dependencies get resolved in the wrong order and the disk can't be deleted because it's still attached)

cc @evandbrown in case you're interested since you had looked into this at one point
also @jevonearth

PR #13443, opened 2017-04-07
provider/google: add attached_disk field to google_compute_instance

danawillow:
This adds the minimum functionality for an `attached_disk` field.

This PR is an alternative to #12398. Instead of allowing attaching/detaching of all disks and having to compare them field-by-field to determine whether a disk is the same, this field would allow only persistent disks that have been defined elsewhere in terraform to be attached/detached. This encourages better practices with terraform actually managing state. For the other solution, if a persistent disk that was created via the disks field is detached, then the disk disappears from terraform's knowledge. With a separate field, we can only allow detaching disks that terraform knows about.

Actually allowing attaching/detaching will come in a follow-up PR. This one was kept small for the sake of clarity and ease of reviewing.

### DIFF
--- a/builtin/providers/google/resource_compute_instance.go
+++ b/builtin/providers/google/resource_compute_instance.go
@@ -854,45 +854,51 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
 	disksCount := d.Get("disk.#").(int)
 	attachedDisksCount := d.Get("attached_disk.#").(int)
 	disks := make([]map[string]interface{}, 0, disksCount)
-	attached_disks := make([]map[string]interface{}, 0, attachedDisksCount)
+	attachedDisks := make([]map[string]interface{}, 0, attachedDisksCount)
 
 	if expectedDisks := disksCount + attachedDisksCount; len(instance.Disks) != expectedDisks {
 		return fmt.Errorf("Expected %d disks, API returned %d", expectedDisks, len(instance.Disks))
 	}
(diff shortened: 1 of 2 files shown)


builtin/providers/google/resource_compute_instance.go
```diff
@@ -791,24 +851,48 @@ func resourceComputeInstanceRead(d *schema.ResourceData, meta interface{}) error
...
-			"type":                    d.Get(fmt.Sprintf("disk.%d.type", i)),
-			"scratch":                 d.Get(fmt.Sprintf("disk.%d.scratch", i)),
-			"auto_delete":             d.Get(fmt.Sprintf("disk.%d.auto_delete", i)),
-			"size":                    d.Get(fmt.Sprintf("disk.%d.size", i)),
-			"device_name":             d.Get(fmt.Sprintf("disk.%d.device_name", i)),
-			"disk_encryption_key_raw": d.Get(fmt.Sprintf("disk.%d.disk_encryption_key_raw", i)),
-		}
-		if disk.DiskEncryptionKey != nil && disk.DiskEncryptionKey.Sha256 != "" {
-			di["disk_encryption_key_sha256"] = disk.DiskEncryptionKey.Sha256
+		if i < disksCount {
```
danawillow: Done.

(technically we're still relying on the fact they're returned in the same order by mixing disk.[property] and d.Get([property]) calls, but your way does certainly look less hacky)

PR #13469, opened 2017-04-07
provider/aws: Fix aws_ami_launch_permission refresh when AMI disappears

phinze:
Launch permissions are implicitly nuked if an AMI is removed for any
reason - Terraform should not error on refresh in this case, but rather
just see the launch permissions as gone and react appropriately.


phinze:
```
=== RUN   TestAccAWSAMILaunchPermission_Basic
--- PASS: TestAccAWSAMILaunchPermission_Basic (347.24s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    347.261s
```

Issue #13460, opened 2017-04-07
Mklink fails on Terraform Get "Local NTFS Shares are required..."


rorychatt:
Hi there,

I am having an issue with running "Terraform Get" on my work machine running Windows 10. This error does not occur on my home machine also running Windows 10.

I think I have narrowed it down to an issue with the path being passed to the go-getter modules, where the destination is being passed as a relative URL rather than fully qualified. This thesis is based on isolation of the go-getter module, where passing a relative URL recreates this same error, but fully qualified works. Code Snippets below.

I will try make a pull request in the next few days when I work out what function call is passing the relative URL.

### Terraform Version
Terraform v0.9.2

### Affected Resources / Commands
terraform get

### Terraform Configuration Files
```
module "resource_group" {
  source   = "../../modules/terraform-azure-modules/resource_group"
  name     = "${var.name}"
  location = "${var.location}"
}
```

### Debug Output
Debug N/A as it fails on Terraform Get command. I recieve an error stating that the mklink fails.

```
Error loading modules: error downloading 'file://C:/Repositories/Terraform/terraform-azure-boilerplate/modules/terraform-azure-modules/resource_group': failed to run mklink .terraform\modules\23503c23fef1c91de45f2822d4a281bf C:\Repositories\Terraform\terraform-azure-boilerplate\modules\terraform-azure-modules\resource_group: exit status 1 "Local NTFS volumes are required to complete the operation.\r\n"
```

### Expected Behavior
A junction link should be created between my file paths when running Terraform Get

### Actual Behavior
The Mklink fails, with the windows output:
`"Local NTFS volumes are required to complete the operation.\r\n"`

Running mklink manually works. i.e.

```
C:\Repositories\Terraform\terraform-azure-boilerplate\aus_east> mklink /J .terraform\modules\23503c23fef1c91de45f2822d4a281bf C:\Repositories\Terraform\terraform-azure-boilerplate\modules\terraform-azure-modules\resource_group
Junction created for .terraform\modules\23503c23fef1c91de45f2822d4a281bf <<===>> C:\Repositories\Terraform\terraform-azure-boilerplate\modules\terraform-azure-modules\resource_group
```

### Steps to Reproduce
This error only appearing on my work machine (Win 10 Enterprise). I do not recieve this error on my home Windows 10 Machine.  I haven't been able to isolate the difference that is causing the error, as I can run the same command, from exactly the same path, with the same Terraform Version, and only have it work on my work machine.

1. terraform get

[...]

PR #13482, opened 2017-04-08
Resolve sporadic error with Terraform Get on Windows with Relative Paths

rorychatt:
Terraform Get now passes a full Path, rather than Relative Path, to the Go-Getter application when referencing local terraform modules. This resolves issue seen in #13460 .


rorychatt:
Forgot to commit in make fmt - resubmitting.

PR #13484, opened 2017-04-08
Resolve Sporadic Error with Terraform get with NTFS errors due to relative path

rorychatterton:
Terraform Get now passes a full Path, rather than Relative Path, to the Go-Getter application when referencing local terraform modules. This resolves issue seen in #13460 .

### DIFF
--- a/command/get.go
+++ b/command/get.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+	"path/filepath"
 
 	"github.com/hashicorp/terraform/config/module"
 )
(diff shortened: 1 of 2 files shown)


rorychatt:
Just want to check - is there anything else you need from me before you can pull this into the next release?

rorychatt:
I have finally found the environment issue that was triggering this error. My companies Windows SOE has the following registry key set in Windows:

"\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Command Processor\Autorun" to change the directory to HOME.

registry key

This is an (unfortunately) common setting in larger enterprises to deal with legacy apps that don't know how to honour $PATH/$HOMEPATH. 

In this case, it's a fairly small code fix that remediates the issue, but also a use case that is due to external environmental variables.

mildwonkey:
Hi! I am going to close this PR, as the original issue (#13460) was closed with a different PR (#13482) and there have been several other changes to the code base since this was opened.
If there is still an outstanding issue, please open a new PR. Thank you!

Issue #13487, opened 2017-04-09
[tf0.9.2/master] Crash during terraform plan's refresh operation


netjunki:
### Terraform Version
Tested with both terraform 0.9.2 and build from master branch (07ff095755010653fa4274c11d0da853c34b7996)

### Affected Resource(s)
Multiple but the crash specifically seems to be triggered by aws_elasticache_replication_group in particular.

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
These are large and complex and can't really be run without a ton of additional infrastructure to execute them. I'll see if I can produce a simpler config to repro this specific issue.

### Panic Output
The full crash.log is over 1 million lines... so I scooped out just the stack trace.

```
2017/04/08 21:38:39 [DEBUG] plugin: terraform: panic: runtime error: index out of range
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_listener.sgp", waiting for: "aws_alb_target_group.sgp_default"
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_listener.studentdiscovery", waiting for: "aws_alb_target_group.studentdiscovery_default"
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_listener.rostering", waiting for: "aws_alb_target_group.rostering_default"
2017/04/08 21:38:39 [DEBUG] plugin: terraform:
2017/04/08 21:38:39 [DEBUG] plugin: terraform: goroutine 17852 [running]:
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_target_group.bomservice", waiting for: "aws_alb.rms"
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_listener.session", waiting for: "aws_alb_target_group.session_default"
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_listener.core", waiting for: "aws_alb_target_group.core_default"
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_listener.dashboard", waiting for: "aws_alb_target_group.dashboard_default"
2017/04/08 21:38:39 [DEBUG] dag/walk: vertex "aws_alb_target_group.prefportal", waiting for: "aws_alb.core"
[...]

PR #13488, opened 2017-04-09
[provider/aws] check that we actually have NodeGroupMembers

netjunki:
This fixes the crash I reported in #13487 but I don't quite understand why (other than the fact that the array obviously was empty but we try to access element 0 of it... but terraform plan does complete successfully after this change. Unclear if this is leaving the system in a weird state though.


catsby:
Thanks!

PR #13493, opened 2017-04-09
provider/google: Changed network argument in google_compute_instance_group as optional

tmshn:
If you want to add an instance group to a backend service, the `network` filed of the instance group must be set.
Formerly this field is set based on the network where instances are in (this is computed by Google API). It means; if you create the empty instance group, the `network` field will be empty, and you cannot add it to a backend service.

In this PR, I changed `network` field in `google_compute_instance_group` from the computed attribute to the configurable optional argument.
Now user can specify network the group is in and add it to a backend service.

### DIFF
--- a/builtin/providers/google/resource_compute_instance_group.go
+++ b/builtin/providers/google/resource_compute_instance_group.go
@@ -66,7 +66,9 @@ func resourceComputeInstanceGroup() *schema.Resource {
 
 			"network": &schema.Schema{
 				Type:     schema.TypeString,
+				Optional: true,
 				Computed: true,
+				ForceNew: true,
 			},
 
 			"project": &schema.Schema{
@@ -126,6 +128,10 @@ func resourceComputeInstanceGroupCreate(d *schema.ResourceData, meta interface{}
 
 	if v, ok := d.GetOk("named_port"); ok {
 		instanceGroup.NamedPorts = getNamedPorts(v.([]interface{}))
+	}
+
+	if v, ok := d.GetOk("network"); ok {
+		instanceGroup.Network = v.(string)
 	}
 
 	log.Printf("[DEBUG] InstanceGroup insert request: %#v", instanceGroup)
--- a/builtin/providers/google/resource_compute_instance_group_test.go
+++ b/builtin/providers/google/resource_compute_instance_group_test.go
@@ -14,6 +14,12 @@ import (
 func TestAccComputeInstanceGroup_basic(t *testing.T) {
 	var instanceGroup compute.InstanceGroup
 	var instanceName = fmt.Sprintf("instancegroup-test-%s", acctest.RandString(10))
+	projs := []string{
+		"GOOGLE_PROJECT",
+		"GCLOUD_PROJECT",
+		"CLOUDSDK_CORE_PROJECT",
+	}
+	var network = fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/global/networks/default", multiEnvSearch(projs))
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -21,12 +27,16 @@ func TestAccComputeInstanceGroup_basic(t *testing.T) {
 		CheckDestroy: testAccComputeInstanceGroup_destroy,
 		Steps: []resource.TestStep{
 			resource.TestStep{
-				Config: testAccComputeInstanceGroup_basic(instanceName),
+				Config: testAccComputeInstanceGroup_basic(instanceName, network),
 				Check: resource.ComposeTestCheckFunc(
 					testAccComputeInstanceGroup_exists(
 						"google_compute_instance_group.basic", &instanceGroup),
 					testAccComputeInstanceGroup_exists(
 						"google_compute_instance_group.empty", &instanceGroup),
+					testAccComputeInstanceGroup_exists(
+						"google_compute_instance_group.empty-with-network", &instanceGroup),
+					testAccComputeInstanceGroup_network(
+						"google_compute_instance_group.empty-with-network", network, &instanceGroup),
 				),
 			},
 		},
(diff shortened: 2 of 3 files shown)


builtin/providers/google/resource_compute_instance_group.go
```diff
@@ -128,6 +130,10 @@ func resourceComputeInstanceGroupCreate(d *schema.ResourceData, meta interface{}
 		instanceGroup.NamedPorts = getNamedPorts(v.([]interface{}))
 	}
 
+	if v, ok := d.GetOk("network"); ok {
+		instanceGroup.Network = v.(string)
```
danawillow: In other terraform resources, we have the network parameter be the name of the network and then do a lookup. For consistency, I think we should do that here too. See https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_compute_instance.go#L479 for an example.

builtin/providers/google/resource_compute_instance_group_test.go
```diff
@@ -201,7 +211,34 @@ func testAccComputeInstanceGroup_named_ports(n string, np map[string]int64, inst
 	}
 }
 
-func testAccComputeInstanceGroup_basic(instance string) string {
+func testAccComputeInstanceGroup_network(n string, network string, instanceGroup *compute.InstanceGroup) resource.TestCheckFunc {
```
danawillow: nitpick: can you rename this to something like `testAccComputeInstanceGroup_networkExists`? Check functions (in my opinion) should be more boolean-like in name to make it clear that it's a check and not a config.

builtin/providers/google/resource_compute_instance_group_test.go
```diff
@@ -245,7 +282,22 @@ func testAccComputeInstanceGroup_basic(instance string) string {
...
 		}
-	}`, instance, instance, instance)
+	}
+
+	resource "google_compute_instance_group" "empty-with-network" {
+		description = "Terraform test instance group empty-with-network"
+		name = "%s-empty-with-network"
+		network = "%s"
+		zone = "us-central1-c"
+			named_port {
```
danawillow: nit: indentation

tmshn:
@danawillow Hey, sorry for the late reaction, but I fixed the points you reviewed! 👍  So could you check again?

Actually the first point, the one about the `network` parameter to take name rather than URL, will be backward incompatible; users who expect `network` argument to be URL should change its usage.
Is it OK? I'm happy to do extra work if needed.

danawillow:
Oh, good point about backwards incompatibility- I think you're right, sorry to lead you down the wrong path.

tmshn:
@danawillow  No problem! I dropped the last commit (and bit modified the next last one).

builtin/providers/google/resource_compute_instance_group_test.go
```diff
@@ -14,19 +14,32 @@ import (
 func TestAccComputeInstanceGroup_basic(t *testing.T) {
 	var instanceGroup compute.InstanceGroup
 	var instanceName = fmt.Sprintf("instancegroup-test-%s", acctest.RandString(10))
+	var networkName = "default"
+	projs := []string{
+		"GOOGLE_PROJECT",
+		"GCLOUD_PROJECT",
+		"CLOUDSDK_CORE_PROJECT",
+	}
+	var networkUrl = fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/global/networks/%s", multiEnvSearch(projs), networkName)
```
danawillow: This test doesn't feel very basic anymore. Mind creating a new test for this? I also think creating the network from within the test would be nicer than relying on the default network and having to read the project from the environment.

tmshn:
@danawillow Divided the test for the network 👍

builtin/providers/google/resource_compute_instance_group_test.go
```diff
@@ -201,6 +227,44 @@ func testAccComputeInstanceGroup_named_ports(n string, np map[string]int64, inst
 	}
 }
 
+func testAccComputeInstanceGroup_hasCorrectNetwork(n_ig string, n_nw string, instanceGroup *compute.InstanceGroup) resource.TestCheckFunc {
```
danawillow: style nit: I'm not a huge fan of mixing types of cases- let's leave everything camelCase

builtin/providers/google/resource_compute_instance_group_test.go
```diff
@@ -365,3 +429,40 @@ func testAccComputeInstanceGroup_outOfOrderInstances(instance string) string {
...
+		}
+
+		network_interface {
+			network = "${google_compute_network.ig_network.name}"
+		}
+	}
+
+	resource "google_compute_instance_group" "with_instance" {
+		description = "Terraform test instance group"
+		name = "%[1]s-with-insntance"
```
danawillow: typo: instance

builtin/providers/google/resource_compute_instance_group_test.go
tmshn: Okay, but what does "everything" mean? "Every function names" including the ones I didn't modify? Or just "every letter" of this function name?

builtin/providers/google/resource_compute_instance_group_test.go
tmshn: Oops, sorry! Fixed -> 4aa065c

builtin/providers/google/resource_compute_instance_group_test.go
danawillow: oh I just meant the variables (n_ig, etc.)

tmshn: Oh sorry, yeah, I got it 😅

tmshn: Fixed 💪 28f0d5e

danawillow:
```
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccComputeInstanceGroup -timeout 120m
=== RUN   TestAccComputeInstanceGroup_basic
--- PASS: TestAccComputeInstanceGroup_basic (136.93s)
=== RUN   TestAccComputeInstanceGroup_update
--- PASS: TestAccComputeInstanceGroup_update (165.93s)
=== RUN   TestAccComputeInstanceGroup_outOfOrderInstances
--- PASS: TestAccComputeInstanceGroup_outOfOrderInstances (116.40s)
=== RUN   TestAccComputeInstanceGroup_network
--- PASS: TestAccComputeInstanceGroup_network (198.64s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	618.011s
```

Issue #13376, opened 2017-04-05
provider/digitalocean: Add monitoring attribute


skovorodkin:
### Terraform Version
v0.9.2

### Affected Resource(s)
- digitalocean_droplet

### Terraform Configuration Files
```hcl
provider "digitalocean" {
  token = "TOKEN"
}
resource "digitalocean_droplet" "droplet-1" {
  image = "ubuntu-16-04-x64"
  name = "droplet-01"
  region = "nyc3"
  size = "512mb"
  monitoring = true
}
```

### Expected Behavior
A new droplet should be created with Monitoring enabled.

### Actual Behavior
```
1 error(s) occurred:

* digitalocean_droplet.droplet-1: : invalid or unknown key: monitoring
```

### Steps to Reproduce
1. `terraform plan`

---

DigitalOcean API allows to enable Monitoring while [creating a new droplet](https://developers.digitalocean.com/documentation/v2/#create-a-new-droplet).

Godo already [has](https://github.com/hashicorp/terraform/blob/master/vendor/github.com/digitalocean/godo/droplets.go#L215) the necessary field.

PR #13494, opened 2017-04-10
provider/digitalocean: Add monitoring option to digitalocean droplets

joshpurvis:
This adds the monitoring option described [in the DO API](https://developers.digitalocean.com/documentation/v2/#create-a-new-droplet) which was added to `godo` on [this commit](https://github.com/digitalocean/godo/commit/7c9bf156916c9a1bf90936231549388231dd5d4e).

Caveats:
As far as I can tell, the current DO API doesn't allow you to turn off monitoring or read the monitoring state. 

Due to this, I wasn't able to come up with a way to properly confirm that the option was applied on the other end after state is read. 

Also, I assumed forcing new on a new option would be a bad idea in this case, so reverting the flag to false will have no effect on existing resources.

I'm new to go-- feel free to let me know if this needs some work.

### DIFF
--- a/builtin/providers/digitalocean/resource_digitalocean_droplet.go
+++ b/builtin/providers/digitalocean/resource_digitalocean_droplet.go
@@ -107,6 +107,11 @@ func resourceDigitalOceanDroplet() *schema.Resource {
 				Optional: true,
 			},
 
+			"monitoring": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
 			"ipv4_address": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -167,6 +172,10 @@ func resourceDigitalOceanDropletCreate(d *schema.ResourceData, meta interface{})
 
 	if attr, ok := d.GetOk("private_networking"); ok {
 		opts.PrivateNetworking = attr.(bool)
+	}
+
+	if attr, ok := d.GetOk("monitoring"); ok {
+		opts.Monitoring = attr.(bool)
 	}
 
 	if attr, ok := d.GetOk("user_data"); ok {
--- a/builtin/providers/digitalocean/resource_digitalocean_droplet_test.go
+++ b/builtin/providers/digitalocean/resource_digitalocean_droplet_test.go
@@ -294,6 +294,27 @@ func TestAccDigitalOceanDroplet_PrivateNetworkingIpv6(t *testing.T) {
 	})
 }
 
+func testAccDigitalOceanDroplet_Monitoring(t *testing.T) {
+	var droplet godo.Droplet
+	rInt := acctest.RandInt()
+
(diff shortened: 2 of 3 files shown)


lfarnell:
I'm new to Terraform but after reviewing this PR I think you may have missed somethings. You should probably add a check for the state of monitoring on the [read](https://github.com/hashicorp/terraform/blob/master/builtin/providers/digitalocean/resource_digitalocean_droplet.go#L232) and add a check for update as well as the [update](https://github.com/hashicorp/terraform/blob/master/builtin/providers/digitalocean/resource_digitalocean_droplet.go#L317) much like how private_networking is done. Well done though otherwise!

skovorodkin:
Closes #13376.

stack72:
Hi @joshpurvis 

Thanks so much for the contribution here - in it's current format, I don't think I can merge it. There are a few reasons for this:

1. If we cannot update the value, then we need to set the parameter `ForceNew: true` on the schema value. If we don't do this, then if a user disables monitoring in their config then they will believe something has changed when it hasn't

2. We are not able to set the value back to state in the Read func as you pointed out. This will break import functionality and when a user sets the value, it will then force a new droplet

You pointed both of these things out of course, but I think we may need DigitalOcean to provide a fix for #2 and we can easily set #1 ourselves :)

What do you think?

Paul

joshpurvis:
Yep I tend to agree. I've created a support ticket with Digitalocean.

lfarnell:
> Hi Luke, 
> 
> Thanks for contacting us today. Sorry for the confusion with all of this. Monitoring management through the API is not fully implemented yet as you have seen. There isn't even a documentation section on it in the API docs at this stage. Our monitoring team is working on getting that fully implemented into the API soon as well as the documentation for it. Currently you are able to implement it but as you have seen this doesn't show up in the features. We do hope to see that available soon. 
> 
> Let us know if you have any questions. 

This is what DO responded with to me after I looked into it.

lfarnell:
@joshpurvis It looks like DO has included the required field in the features block

```json
"features": [
    "backups",
    "private_networking",
    "ipv6",
    "monitoring"
],
```

Might also have to move this PR to the new repository as well

grubernaut:
Hello @joshpurvis, and all! 

Thank you for the contribution, we would love to be able to support this in the DigitalOcean provider!

However, with the migration of all of our providers to their new GitHub organizations (https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10), the DigitalOcean provider is now located at: https://github.com/terraform-providers/terraform-provider-digitalocean. 

As such, I'm going to close this PR, as we cannot merge it in it's current state. If you wouldn't mind migrating this PR over to the new repository, I will happily review it for you (Just tag me as a requested reviewer). 

Thanks again, and sorry for the inconvenience!

PR #13499, opened 2017-04-10
Remove Terraform logo from README

ajlanghorn:
The README used to show the Terraform logo, but since the rebrand this
appears to have disappeared. I'm not sure how valuable it is to show the
image in the README, so I've removed the image from the README to stop a
404 rather than create a new image and have to continue to maintain that in future.


stack72:
Hi @ajlanghorn 

Thanks for the work here - we discussed this internally and are actually going to update the logo so that it doesn't 404 - we would rather do that then remove it completely

Thanks again for pointing this out

Paul

ajlanghorn:
@stack72 Yeah, that was my other thought, but then I couldn't get a decent image together. So, that's fine. As long as the 404 is gone then that's me happy :)

stack72:
Understand completely, we are sorting out the svg right now :)

PR #13509, opened 2017-04-10
Update syntax on some providers

sethvargo:


website/source/docs/providers/chef/r/data_bag.html.markdown
```diff
@@ -6,7 +6,7 @@ description: |-
   Creates and manages a data bag in Chef Server.
 ---
 
-# chef\_data\_bag
+# chef_data_bag
```
apparentlymart: hmm we used to do this because markdown would interpret the underscores as emphasis. Is that no longer true?

sethvargo: It actually never did that. I don't know where that belief came from, but that's literally never happened 😄 . Here's an example of it working properly as-is: https://www.terraform.io/docs/providers/terraform/d/remote_state.html

catsby: Seem so 
screen shot 2017-04-10 at 11 36 02 am

I welcome this change 😄

apparentlymart: Fair enough... I thought I remembered trying that once and seeing it italic it, but probably what happened  instead was that I saw the escapes and thought "oh, that must be because it italicizes them" and then never actually questioned it. So okay! Looking forward to not having to insert backslashes anymore.

sethvargo: Yea, it only happens when there are spaces, so _foo_ is italic, but foo_bar_baz is not. See, even github got it right!

PR #13515, opened 2017-04-10
provider/google: add support for networkIP in compute instance templates

danawillow:
cc @danisla

### DIFF
--- a/builtin/providers/google/resource_compute_instance_template.go
+++ b/builtin/providers/google/resource_compute_instance_template.go
@@ -197,6 +197,12 @@ func resourceComputeInstanceTemplate() *schema.Resource {
 							Computed: true,
 						},
 
+						"network_ip": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+							ForceNew: true,
+						},
+
 						"subnetwork": &schema.Schema{
 							Type:     schema.TypeString,
 							Optional: true,
@@ -462,7 +468,9 @@ func buildNetworks(d *schema.ResourceData, meta interface{}) ([]*compute.Network
 		var iface compute.NetworkInterface
 		iface.Network = networkLink
 		iface.Subnetwork = subnetworkLink
-
+		if v, ok := d.GetOk(prefix + ".network_ip"); ok {
+			iface.NetworkIP = v.(string)
+		}
 		accessConfigsCount := d.Get(prefix + ".access_config.#").(int)
 		iface.AccessConfigs = make([]*compute.AccessConfig, accessConfigsCount)
 		for j := 0; j < accessConfigsCount; j++ {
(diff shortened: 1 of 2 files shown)


danawillow:
@paddyforan- for some reason I couldn't get the test to act the way I wanted it to. My original test func looked something like this:
```go
    "google_compute_instance_template.foobar", "network_interface.0.network_ip",
    instanceTemplate.Properties.NetworkInterfaces[0].NetworkIP),
```
But that was giving me segfaults (during compilation) because at that point `instanceTemplate.Properties` was nil even though the previous function accesses it just fine (`panic: runtime error: invalid memory address or nil pointer dereference`). Any idea what I'm doing wrong?

paddyforan:
Hmmm. Any chance you could push the not-working code somewhere, @danawillow?

danawillow:
Sure, it's at https://github.com/danawillow/terraform/tree/networkip-testfail

paddycarver:
It's not awesome, but I found the problem. With this patch, the test passes: https://github.com/hashicorp/terraform/commit/5ad0b99997450cc94c54815e1181abe4f3fa3d49

Essentially, the problem was that you were accessing the variable before it was set, because it was being passed into a function that was set as a value in a struct, so all of the functions were called. Which normally would be fine. But because these functions _return functions_, and it's the _returned_ function of a previous value that would set the variable, it was being accessed before it was set.

By moving the access into the _returned_ function, it correctly does not get called until after the variable is set.

Hopefully that made at least a little sense. My brain hurts.

Embedded diff, for posterity in case the branch gets deleted:

<details>

```diff
diff --git a/builtin/providers/google/resource_compute_instance_template_test.go b/builtin/providers/google/resource_compute_instance_template_test.go
index 9020e7b..e5e4322 100644
--- a/builtin/providers/google/resource_compute_instance_template_test.go
+++ b/builtin/providers/google/resource_compute_instance_template_test.go
@@ -69,9 +69,9 @@ func TestAccComputeInstanceTemplate_networkIP(t *testing.T) {
 					testAccCheckComputeInstanceTemplateExists(
 						"google_compute_instance_template.foobar", &instanceTemplate),
 					testAccCheckComputeInstanceTemplateNetwork(&instanceTemplate),
-					resource.TestCheckResourceAttr(
+					testAccCheckComputeInstanceTemplateNetworkIP(
 						"google_compute_instance_template.foobar", "network_interface.0.network_ip",
-						instanceTemplate.Properties.NetworkInterfaces[0].NetworkIP),
+						&instanceTemplate),
 				),
 			},
 		},
@@ -266,6 +266,13 @@ func testAccCheckComputeInstanceTemplateNetwork(instanceTemplate *compute.Instan
 	}
 }
 
+func testAccCheckComputeInstanceTemplateNetworkIP(name, key string, instanceTemplate *compute.InstanceTemplate) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		ip := instanceTemplate.Properties.NetworkInterfaces[0].NetworkIP
+		return resource.TestCheckResourceAttr(name, key, ip)(s)
+	}
+}
+
 func testAccCheckComputeInstanceTemplateNetworkName(instanceTemplate *compute.InstanceTemplate, network string) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		for _, i := range instanceTemplate.Properties.NetworkInterfaces {
```

</details>

danawillow:
Ok, I added a change pretty similar to what you suggested (I decided to partially keep what I had because why not do a three way check of config vs state vs gcp). Ready for review.

PR #13523, opened 2017-04-11
provider/openstack: Fixing names for import tests

jtopjian:
Noticed the names weren't following the same pattern as the other tests, so they weren't being included when I would run a group-by-group test run.

Fixed and all pass.


stack72:
LGTM! thanks for this

PR #13525, opened 2017-04-11
provisioner/chef: make sure we add new Chef-Vault clients as clients

### LINKED ISSUE #9137: Problem with vault_json implementation for chef provisioner
### Terraform Version

0.7.4
### Affected Resource(s)

n/a
### Affected Provisioner(s)

chef
### Expected Behavior

When using the `vault_json` argument, the behavior should mimic the `--bootstrap-vault-file`/`--bootstrap-vault-json` feature of `knife bootstrap`.
### Actual Behavior

The provisioned node is added as an admin not a client of the vault item(s).

Provisioned node is visible in `knife vault show VAULT ITEM -p admins` and not `knife vault show VAULT ITEM -p clients`.

Privileged operations such as `knife vault update VAULT ITEM -S "" -k /etc/chef/client.pem` are possible on the provisioned node.
### Steps to Reproduce
1. `terraform apply`
### Important Factoids

Chef Vault has two roles of access; client (ro) and admin (rw). Adding the provisioned node as an admin allows it to both modify the vault item contents and manage it's access controls.

As far as I'm aware, there is not functionality to efficiently add a single node to client level access for a vault using `knife vault` from the command line. You have basically two options:
- create a client that satisfies the `search_query` for the item and run `knife vault refresh`
- use `knife bootstrap` with one of the above mentioned options

The first option has the downside that it can be very slow due to the core mechanisms in chef-vault for search and returning large node objects. In production environments, it's not uncommon for this operation to take multiple minutes if run on or near the chef server and even an hour or more if run from a developer workstation. There are long standing chef-vault issues surrounding this behavior.

The second is much faster since it only checks that the node being bootstrapped satisfies the search query and inserts it without refreshing the rest of the item's access. The current implementation in terraform for the chef provisioner doesn't use `knife bootstrap` so this would require significant changes.

Before this feature was available in terraform my workaround was:
1. use the chef provisioner with an empty run list
2. use a `local-exec` with custom code to mimic the insert of `knife bootstrap`
3. use a `local-exec` to update the run list
4. use a `remote-exec` to run chef on the node

This is all necessary since the `search_query` will generally be something that is more complex than just `name:foo` and so the node object must exist with the required tags, attributes, environment etc. which doesn't fit well in the `knife client create` ... `chef-client` workflow.

Due to that complexity I was quite excited to see this support in 0.7.4 but I'm afraid that it's not working as intended.
### References

This feature was added in #8577 
https://github.com/chef/chef-vault/blob/master/KNIFE_EXAMPLES.md
https://docs.chef.io/knife_bootstrap.html#options
https://github.com/hashicorp/terraform/blob/968472a63e4542a2c5e2693c4489da4776dac996/builtin/provisioners/chef/resource_provisioner.go#L548

Hopefully this makes sense (and is correct), let me know if I can provide more information. I'd be happy to help work on the refactor after it's decided upon.

Thanks!

svanharmelen:
This is possible using the newly released Chef-Vault 3.0 gem. Before we could only add new clients as admins.

Fixes #9137

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -579,7 +579,7 @@ func (p *Provisioner) configureVaultsFunc(
 
 		for vault, items := range p.vaults {
 			for _, item := range items {
-				updateCmd := fmt.Sprintf("%s vault update %s %s -A %s -M client %s",
+				updateCmd := fmt.Sprintf("%s vault update %s %s -C %s -M client %s",
 					knifeCmd,
 					vault,
 					item,
--- a/builtin/provisioners/chef/resource_provisioner_test.go
+++ b/builtin/provisioners/chef/resource_provisioner_test.go
@@ -269,7 +269,7 @@ func TestResourceProvider_configureVaults(t *testing.T) {
 
 			Commands: map[string]bool{
 				fmt.Sprintf("%s install chef-vault", linuxGemCmd): true,
-				fmt.Sprintf("%s vault update vault1 item1 -A nodename1 -M client -c %s/client.rb "+
+				fmt.Sprintf("%s vault update vault1 item1 -C nodename1 -M client -c %s/client.rb "+
 					"-u bob --key %s/bob.pem", linuxKnifeCmd, linuxConfDir, linuxConfDir): true,
 			},
 		},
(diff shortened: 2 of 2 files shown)


svanharmelen:
@radeksimko if you could have a quick look? This is a very small fix, but we had to wait a long time to get my changes to Chef-Vault released in a new gem... But that happened yesterday, so we can now fix/improve this one 😀

svanharmelen:
Thanks for the review @radeksimko!

Issue #13518, opened 2017-04-10
keep_on_remove functionality of vsphere_virtual_machine resource is not documented


madrover:
### Terraform Version
Terraform v0.9.2
### Affected Resource(s)
- vsphere_virtual_machine

### Issue

The **keep_on_remove** functionality for the **vsphere_virtual_machine** resource is not documented at https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#disks

### References
- provider/vsphere: Add support for keep_on_remove to vsphere_virtual_machine (#7169)

PR #13536, opened 2017-04-11
provider/vsphere: Add keep_on_remove to docs

grubernaut:
Fixes: #13518


radeksimko:
I think the name is not great considering what it really does (keeps **disks** on removal of the _VM_), but that's not something to discuss in this PR which is just updating docs :)

Issue #11588, opened 2017-02-01
panic: empty structure "initial_lifecycle_hook." has entry "initial_lifecycle_hook.78676173.default_result"


pagemedias:
### Terraform Version
0.8.5

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_autoscaling_group

### Terraform Configuration Files
```hcl
resource "aws_autoscaling_group" "ecs_host_autoscaling_group" {
  # Required
  launch_configuration = "${aws_launch_configuration.host_launch_conf.name}"
  max_size = "${var.max_hosts}"
  min_size = "${var.min_hosts}"

  # Other
  name = "${aws_launch_configuration.host_launch_conf.name_prefix}autoscaling"
  vpc_zone_identifier = "${var.host_subnets}"
  initial_lifecycle_hook {
    name                   = "hook_ec2_instance_terminating"
    default_result         = "CONTINUE"
    heartbeat_timeout      = 3600
    lifecycle_transition   = "autoscaling:EC2_INSTANCE_TERMINATING"
    notification_target_arn = "${aws_sns_topic.ecs_host_autoscaling_group_lifecyclehook_notis.arn}"
    role_arn                = "${aws_iam_role.ecs_host_autoscaling_group_role.arn}"
  }

  depends_on = ["aws_launch_configuration.host_launch_conf","aws_sns_topic.ecs_host_autoscaling_group_lifecyclehook_notis"]
}
```

### Panic Output
panic: empty structure "initial_lifecycle_hook." has entry "initial_lifecycle_hook.78676173.default_result"
https://gist.github.com/pagemedias/b90b584ffc147be399437d53776332df

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create terraform config of ASG with `initial_lifecycle_hook` attribute
2. `terraform apply`
3. Remove `initial_lifecycle_hook` attribute from ASG
4. `terraform apply`

### References
Came across this while reporting... https://github.com/hashicorp/terraform/issues/11587

mitchellh:
Hey @jbardin can you take a look at this since this was your context and fix awhile back?

Issue #11320, opened 2017-01-20
Panic on aws_s3_bucket resource


grubernaut:
### Terraform Version
`Terraform v0.8.5-dev (2c708557821593cf4e248199b5f6c0c672ff3fb5+CHANGES)`

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_s3_bucket

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "aws_s3_bucket" "bucket" {
  bucket = "testing-bucket-panic"
  acl = "private"
  lifecycle_rule {
    id = "id1"
    prefix = "path1/"
    enabled = true
    expiration {
      days = 365
    }
    transition {
      days = 30
      storage_class = "STANDARD_IA"
    }
    transition {
      days = 60
      storage_class = "GLACIER"
    }
  }
  lifecycle_rule {
    id = "id2"
    prefix = "path2/"
    enabled = true
    expiration {
      date = "2016-01-12"
    }
  }
}
```

### Debug Output
https://gist.github.com/grubernaut/fc871de63e719895d0732321594d1dd8

### Panic Output
https://gist.github.com/grubernaut/4d8012cc5831cbe71cebbc5477d947d7

### Expected Behavior
Terraform should have been able to handle the change from `private` to `public-read` on the same bucket.

### Actual Behavior
Panicked

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Comment out all of the lifecycle blocks, and change `acl` to `public-read`
3. `terraform plan` or `terraform apply`

### Important Factoids
First started failing acceptance tests after January 18th

Issue #11170, opened 2017-01-12
aws_security_group egress/ingress rule cidr_block changes cause crash


justicel:
Terraform v0.8.3

### Affected Resource(s)
- aws_security_group

### Terraform Configuration Files
```hcl

variable "alb_cidr_blocks" {
  type        = "list"
  default     = ["192.168.1.0/24", "192.168.2.0/24"]
  description = "List of DGN Subnets to allow bulk access to host(s)"
}

resource "aws_security_group" "alb" {
  provider    = "aws.rancher_${var.region}"
  name        = "${var.alb_name}-rancher-alb"
  description = "Rancher ALB"
  vpc_id      = "${var.vpc}"

  ingress {
    from_port   = 443
    to_port     = 443
    protocol    = "tcp"
    cidr_blocks = ["${var.alb_cidr_blocks}"]
  }
}

//Rancher ALB Security Group Ingress Rules
output "alb_rancher_sg_ingress" {
  value = "${aws_security_group.alb.ingress}"
}

//Rancher ALB Security Group Egress Rules
output "alb_rancher_sg_egress" {
  value = "${aws_security_group.alb.egress}"
}

```

### Panic Output
https://gist.github.com/justicel/c54ef9bc668e474bb8d9071c33d86e58

### Expected Behavior
Changing alb_cidr_blocks from a list of two elements, to three elements (e.g: ["192.168.1.0/24", "192.168.2.0/24", "192.168.3.0/24"]) should allow a terraform plan and terraform apply to run.

### Actual Behavior
Terraform crashes if the length of the array is changed, and you are outputting the 'ingress' and 'egress' values of the resource. If you REMOVE the output for these two items, the plan commences fine.

### Steps to Reproduce
1. `terraform plan`
2. Change length of array for ingress/egress rule cidr blocks
3 `terraform plan`

PR #13541, opened 2017-04-11
core: fix a crash by suggesting a different approach to solve #11170

svanharmelen:
For readability I've splitted the changes into two commits. The first commit only reverts the current solution, the second commit shows my take on solving the problem.

The main reason I started to look into this particular issue was that I encountered a panic caused by this line: https://github.com/hashicorp/terraform/blob/master/terraform/state.go#L1737. But when I took a closer look I also noticed that the current diff output seems less correct then it did before.

But first for reference here is a list with links that give you a nice timeline you can walk through to get up to speed:

1. On 16-12-2016 some work was done to properly expand sets as lists from a flatmap ([link](https://github.com/hashicorp/terraform/commits/master/flatmap/expand.go))
2. On 10-01-2017 version 0.8.3 was released containing this new code ([link](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#083-january-10-2017))
3. On 12-01-2017 issue #11170 was filled describing a crash caused by the new code ([link](https://github.com/hashicorp/terraform/issues/11170))
4. On 17-01-2017 a patch was merged to fix issue #11170 ([link](https://github.com/hashicorp/terraform/pull/11245/files))
5. On 20-01-2017 issue #11320 was filled describing a crash caused by the new code ([link](https://github.com/hashicorp/terraform/issues/11320))
6. On 20-01-2017 a patch was merged to fix issue #11320 ([link](https://github.com/hashicorp/terraform/pull/11321))
7. On 29-01-2017 a patch was added as the initial patch caused some acceptance tests to fail ([link](https://github.com/hashicorp/terraform/pull/11498/files))
8. On 01-02-2017 issue #11588 was filled describing a crash caused by the new code ([link](https://github.com/hashicorp/terraform/issues/11588))
9. On 07-02-2017 a patch was merged to fix issue #11588 ([link](https://github.com/hashicorp/terraform/pull/11757))
10. Not filled, but after updating our systems we started to see panics as described above as well...

To reproduce the error you can use this config:

```
provider "cloudstack" {
  api_url = "xxx"
  api_key = "xxx"
  secter_key = "xxx"
}

resource "cloudstack_network_acl_rule" "default" {
  acl_id = "979a3d17-98cf-489d-885f-289a3888b6b6"
  managed = true

  rule {
      cidr_list = ["10.0.0.0/8"]
      protocol  = "tcp"
      ports     = ["90", "1000-2000"]
    }
}
```

First run a `terraform apply` and then update the config to contain:

```
provider "cloudstack" {
  api_url = "xxx"
  api_key = "xxx"
  secter_key = "xxx"
}

resource "cloudstack_network_acl_rule" "default" {
  acl_id = "979a3d17-98cf-489d-885f-289a3888b6b6"
  managed = true
}
```

After this a `terraform plan/apply` will cause the crash/panic to occur.

The `cloudstack_network_acl_rule` resource contains a `rule` block which is a set. In that set there is a computed `schema.TypeMap` field called `uuids` ([link](https://github.com/hashicorp/terraform/blob/master/builtin/providers/cloudstack/resource_cloudstack_network_acl_rule.go#L84-L87)). Now since this field is computed it is not taken into account when creating the diff and so any entries in the map will not be marked as `NewRemoved` (which is expected and perfectly fine if you ask me).

As the current solution doesn't seem to fix this corner case I started looking into how to add a fix for this. When doing so (after adding a simple patch to test some things) I noticed that the diff was looking a bit funny/different:

```
~ cloudstack_network_acl_rule.default
    rule.#:                              "1" => "0"
    rule.990414728.action:               "allow" => ""
    rule.990414728.cidr_list:            "" => ""
    rule.990414728.cidr_list.#:          "1" => "0"
    rule.990414728.cidr_list.3423270202: "10.0.0.0/8" => ""
    rule.990414728.icmp_code:            "" => ""
    rule.990414728.icmp_type:            "" => ""
    rule.990414728.ports:                "" => ""
    rule.990414728.ports.#:              "2" => "0"
    rule.990414728.ports.1209010669:     "1000-2000" => ""
    rule.990414728.ports.1770303465:     "90" => ""
    rule.990414728.protocol:             "tcp" => ""
    rule.990414728.traffic_type:         "ingress" => ""
    rule.990414728.uuids:                "" => ""
```

Mainly notice the lines `rule.990414728.cidr_list: "" => ""`, `rule.990414728.ports: "" => ""` and `rule.990414728.uuids: "" => ""`. These lines should not actually exist in the diff as they are describing the container (set, map, list) field but not an actual value.

Besides these lines I would like to argue that all lines that have the value `"" => ""` should not be in the diff as they clearly do not change...

Before version 0.8.3 and again after adding the patch in this PR the diff looks like this:

```
~ cloudstack_network_acl_rule.default
    rule.#:                              "1" => "0"
    rule.990414728.action:               "allow" => ""
    rule.990414728.cidr_list.#:          "1" => "0"
    rule.990414728.cidr_list.3423270202: "10.0.0.0/8" => ""
    rule.990414728.ports.#:              "2" => "0"
    rule.990414728.ports.1209010669:     "1000-2000" => ""
    rule.990414728.ports.1770303465:     "90" => ""
    rule.990414728.protocol:             "tcp" => ""
    rule.990414728.traffic_type:         "ingress" => ""
```

Which (IMHO) resembles the expected diff much better.

So... To make a long story short (:wink:), I believe the suggested patch in this PR fixes both the crash/panic I encountered and restores the diff to again be inline with what one would expect, showing only the things that are going to change.

Please let me know your thoughts on this one... Thx!


jbardin:
This looks great! There's nothing I like more than removing code to fix things. 

My only question is if we should put the helper schema changes back in, or were those part of the problem with the odd diffs? 

IIRC one of the issues in this whole chain of events was a nested data structure would get "removed" by only having a count marked as 0. But remain in the diff, causing an diff when there shouldn't be. The schema changes were intended to have the diff be more precise, rather than rely on the expansion cleaning things up where it didn't have the context to get it right in all cases. 

Let me try to come up with a test case for that, but I think it might look like:

	Map: map[string]string{
		"empty_map_of_sets.%":         "0",
		"empty_map_of_sets.set1.#":    "0",
		"empty_map_of_sets.set1.1234": "x",
	},

Where we expect an empty map, but instead we get a map with a single empty `set1`.

jbardin:
@svanharmelen,

I think all we may need to do here is expand this same logic to maps as well. I made another PR so we can view my commit https://github.com/hashicorp/terraform/pull/13628/commits/7b39d951d465dbfd7db6248f419f74169f7a2af6 (I can't target your remote branch, so the that full PR has your commits too).

It required fixing up some test fixtures, but from what I can see those were technically incorrect to begin with.

svanharmelen:
I'm offline the rest of the day, but will follow up on this one tomorrow morning! Thx!

PR #13563, opened 2017-04-12
provider/openstack: Ignore fixed_ip when importing ports

jtopjian:
I missed this one in the logs when fixing up the import test fails from yesterday.

@stack72 Does this make sense to do? `fixed_ip` is meant to be declared by the user and not meant to be set by the API at all. I tested doing an actual import then making the relevant configuration to match the state and it worked fine.


stack72:
LGTM!

PR #13571, opened 2017-04-12
Add support for K8s pod

ashishth09:
I was just testing it and looks default volumes are created as part of Pod creation which is leading to the below error.

```
2017/04/12 17:58:48 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/kubernetes -v -run TestAccKubernetesPod_basic -timeout 120m
=== RUN   TestAccKubernetesPod_basic
--- FAIL: TestAccKubernetesPod_basic (3.24s)
		
		DIFF:
		
		UPDATE: kubernetes_pod.test
		  spec.0.containers.0.resources.#:                "1" => "0"
		  spec.0.containers.0.resources.0.limits:         "" => ""
		  spec.0.containers.0.resources.0.requests:       "" => ""
		  spec.0.containers.0.volume_mounts.#:            "1" => "0"
		  spec.0.containers.0.volume_mounts.0.mount_path: "/var/run/secrets/kubernetes.io/serviceaccount" => ""
		  spec.0.containers.0.volume_mounts.0.name:       "default-token-ndgzz" => ""
		  spec.0.containers.0.volume_mounts.0.read_only:  "true" => "false"
		  spec.0.containers.0.volume_mounts.0.sub_path:   "" => ""
		  spec.0.volumes.#:                               "1" => "0"
		  spec.0.volumes.0.aws_elastic_block_store:       "" => ""
		  spec.0.volumes.0.azure_disk:                    "" => ""
		  spec.0.volumes.0.azure_file:                    "" => ""
		  spec.0.volumes.0.ceph_fs:                       "" => ""
		  spec.0.volumes.0.cinder:                        "" => ""
		  spec.0.volumes.0.fc:                            "" => ""
		  spec.0.volumes.0.flex_volume:                   "" => ""
		  spec.0.volumes.0.flocker:                       "" => ""
		  spec.0.volumes.0.gce_persistent_disk:           "" => ""
		  spec.0.volumes.0.glusterfs:                     "" => ""
		  spec.0.volumes.0.host_path:                     "" => ""
		  spec.0.volumes.0.iscsi:                         "" => ""
		  spec.0.volumes.0.name:                          "default-token-ndgzz" => ""
		  spec.0.volumes.0.nfs:                           "" => ""
		  spec.0.volumes.0.persistent_volume_claim:       "" => ""
		  spec.0.volumes.0.photon_persistent_disk:        "" => ""
		  spec.0.volumes.0.quobyte:                       "" => ""
		  spec.0.volumes.0.rbd:                           "" => ""
		  spec.0.volumes.0.vsphere_volume:                "" => ""
		
		STATE:
		
		kubernetes_pod.test:
		  ID = default/tf-acc-pod-test-wm4gasjx7y
		  metadata.# = 1
		  metadata.0.annotations.% = 0
		  metadata.0.generate_name = 
		  metadata.0.generation = 0
		  metadata.0.labels.% = 1
		  metadata.0.labels.app = pod_label
		  metadata.0.name = tf-acc-pod-test-wm4gasjx7y
		  metadata.0.namespace = default
		  metadata.0.resource_version = 336409
		  metadata.0.self_link = /api/v1/namespaces/default/pods/tf-acc-pod-test-wm4gasjx7y
		  metadata.0.uid = ae63549f-1f7b-11e7-a206-22422a00791b
		  spec.# = 1
		  spec.0.active_deadline_seconds = 0
		  spec.0.containers.# = 1
		  spec.0.containers.0.args.# = 0
		  spec.0.containers.0.command.# = 0
		  spec.0.containers.0.image = nginx:1.7.9
		  spec.0.containers.0.image_pull_policy = IfNotPresent
		  spec.0.containers.0.name = cname
		  spec.0.containers.0.ports.# = 1
		  spec.0.containers.0.ports.0.container_port = 9080
		  spec.0.containers.0.ports.0.host_ip = 
		  spec.0.containers.0.ports.0.host_port = 0
		  spec.0.containers.0.ports.0.name = 
		  spec.0.containers.0.ports.0.protocol = TCP
		  spec.0.containers.0.resources.# = 1
		  spec.0.containers.0.volume_mounts.# = 1
		  spec.0.containers.0.volume_mounts.0.mount_path = /var/run/secrets/kubernetes.io/serviceaccount
		  spec.0.containers.0.volume_mounts.0.name = default-token-ndgzz
		  spec.0.containers.0.volume_mounts.0.read_only = true
		  spec.0.containers.0.volume_mounts.0.sub_path = 
		  spec.0.containers.0.working_dir = 
		  spec.0.dns_policy = ClusterFirst
		  spec.0.host_ipc = false
		  spec.0.host_network = false
		  spec.0.hostname = 
		  spec.0.image_pull_secrets.# = 0
		  spec.0.node_name = minikube
		  spec.0.node_selector.% = 0
		  spec.0.restart_policy = Always
		  spec.0.service_account_name = default
		  spec.0.subdomain = 
		  spec.0.termination_grace_period_seconds = 30
		  spec.0.volumes.# = 1
		  spec.0.volumes.0.aws_elastic_block_store.# = 0
		  spec.0.volumes.0.azure_disk.# = 0
		  spec.0.volumes.0.azure_file.# = 0
		  spec.0.volumes.0.ceph_fs.# = 0
		  spec.0.volumes.0.cinder.# = 0
		  spec.0.volumes.0.fc.# = 0
		  spec.0.volumes.0.flex_volume.# = 0
		  spec.0.volumes.0.flocker.# = 0
		  spec.0.volumes.0.gce_persistent_disk.# = 0
		  spec.0.volumes.0.glusterfs.# = 0
		  spec.0.volumes.0.host_path.# = 0
		  spec.0.volumes.0.iscsi.# = 0
		  spec.0.volumes.0.name = default-token-ndgzz
		  spec.0.volumes.0.nfs.# = 0
		  spec.0.volumes.0.persistent_volume_claim.# = 0
		  spec.0.volumes.0.photon_persistent_disk.# = 0
		  spec.0.volumes.0.quobyte.# = 0
		  spec.0.volumes.0.rbd.# = 0
		  spec.0.volumes.0.vsphere_volume.# = 0
FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/kubernetes	3.310s
```

### DIFF
--- a/builtin/providers/kubernetes/resource_kubernetes_pod.go
+++ b/builtin/providers/kubernetes/resource_kubernetes_pod.go
@@ -1,0 +1,195 @@
+package kubernetes
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgApi "k8s.io/apimachinery/pkg/types"
+	api "k8s.io/kubernetes/pkg/api/v1"
+	kubernetes "k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
+)
+
+func resourceKubernetesPod() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceKubernetesPodCreate,
+		Read:   resourceKubernetesPodRead,
+		Update: resourceKubernetesPodUpdate,
+		Delete: resourceKubernetesPodDelete,
+		Exists: resourceKubernetesPodExists,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"metadata": namespacedMetadataSchema("pod", true),
+			"spec": {
+				Type:        schema.TypeList,
+				Description: "Spec of the pod owned by the cluster",
+				Required:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: podSpecFields(),
+				},
+			},
+		},
+	}
+}
+func resourceKubernetesPodCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
+	spec, err := expandPodSpec(d.Get("spec").([]interface{}))
+	if err != nil {
+		return err
+	}
+
+	spec.AutomountServiceAccountToken = ptrToBool(false)
+
+	pod := api.Pod{
+		ObjectMeta: metadata,
+		Spec:       spec,
+	}
+
+	log.Printf("[INFO] Creating new pod: %#v", pod)
+	out, err := conn.CoreV1().Pods(metadata.Namespace).Create(&pod)
+
+	if err != nil {
+		return err
+	}
+	log.Printf("[INFO] Submitted new pod: %#v", out)
+
+	d.SetId(buildId(out.ObjectMeta))
+
+	stateConf := &resource.StateChangeConf{
+		Target:  []string{"Running"},
+		Pending: []string{"Pending"},
+		Timeout: 5 * time.Minute,
+		Refresh: func() (interface{}, string, error) {
+			out, err := conn.CoreV1().Pods(metadata.Namespace).Get(metadata.Name, metav1.GetOptions{})
+			if err != nil {
+				log.Printf("[ERROR] Received error: %#v", err)
+				return out, "Error", err
+			}
+
+			statusPhase := fmt.Sprintf("%v", out.Status.Phase)
+			log.Printf("[DEBUG] Pods %s status received: %#v", out.Name, statusPhase)
+			return out, statusPhase, nil
+		},
+	}
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return err
+	}
+	log.Printf("[INFO] Pod %s created", out.Name)
+
+	return resourceKubernetesPodRead(d, meta)
+}
+
+func resourceKubernetesPodUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+	namespace, name := idParts(d.Id())
+	ops := patchMetadata("metadata.0.", "/metadata/", d)
+	if d.HasChange("spec") {
+		specOps, err := patchPodSpec("/spec", "spec.0.", d)
+		if err != nil {
+			return err
+		}
+		ops = append(ops, specOps...)
+	}
+	data, err := ops.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("Failed to marshal update operations: %s", err)
+	}
+
+	log.Printf("[INFO] Updating  pod %s: %s", d.Id(), ops)
+
+	out, err := conn.CoreV1().Pods(namespace).Patch(name, pkgApi.JSONPatchType, data)
+	if err != nil {
+		return err
+	}
+	log.Printf("[INFO] Submitted updated pod: %#v", out)
+
+	d.SetId(buildId(out.ObjectMeta))
+	return resourceKubernetesPodRead(d, meta)
+}
+
+func resourceKubernetesPodRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+	namespace, name := idParts(d.Id())
+
+	log.Printf("[INFO] Reading pod %s", name)
+	pod, err := conn.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		log.Printf("[DEBUG] Received error: %#v", err)
+		return err
+	}
+	log.Printf("[INFO] Received pod: %#v", pod)
+
+	err = d.Set("metadata", flattenMetadata(pod.ObjectMeta))
+	if err != nil {
+		return err
+	}
+
+	podSpec, err := flattenPodSpec(pod.Spec)
+	if err != nil {
+		return err
+	}
+
+	err = d.Set("spec", podSpec)
+	if err != nil {
+		return err
+	}
+	return nil
+
+}
+
+func resourceKubernetesPodDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+	namespace, name := idParts(d.Id())
+	log.Printf("[INFO] Deleting pod: %#v", name)
+	err := conn.CoreV1().Pods(namespace).Delete(name, nil)
+	if err != nil {
+		return err
+	}
+
+	err = resource.Retry(1*time.Minute, func() *resource.RetryError {
+		out, err := conn.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			if statusErr, ok := err.(*errors.StatusError); ok && statusErr.ErrStatus.Code == 404 {
+				return nil
+			}
+			return resource.NonRetryableError(err)
+		}
+
+		log.Printf("[DEBUG] Current state of pod: %#v", out.Status.Phase)
+		e := fmt.Errorf("Pod %s still exists (%s)", name, out.Status.Phase)
+		return resource.RetryableError(e)
+	})
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[INFO] Pod %s deleted", name)
+
+	d.SetId("")
+	return nil
+}
+
+func resourceKubernetesPodExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	conn := meta.(*kubernetes.Clientset)
+
+	namespace, name := idParts(d.Id())
+	log.Printf("[INFO] Checking pod %s", name)
+	_, err := conn.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		if statusErr, ok := err.(*errors.StatusError); ok && statusErr.ErrStatus.Code == 404 {
+			return false, nil
+		}
+		log.Printf("[DEBUG] Received error: %#v", err)
+	}
+	return true, err
+}
--- a/builtin/providers/kubernetes/schema_pod_spec.go
+++ b/builtin/providers/kubernetes/schema_pod_spec.go
@@ -1,0 +1,373 @@
+package kubernetes
+
+import "github.com/hashicorp/terraform/helper/schema"
+
+func podSpecFields() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"active_deadline_seconds": {
+			Type:         schema.TypeInt,
+			Optional:     true,
+			ValidateFunc: validatePositiveInteger,
+			Description:  "Optional duration in seconds the pod may be active on the node relative to StartTime before the system will actively try to mark it failed and kill associated containers. Value must be a positive integer.",
+		},
+		"container": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "List of containers belonging to the pod. Containers cannot currently be added or removed. There must be at least one container in a Pod. Cannot be updated. More info: http://kubernetes.io/docs/user-guide/containers",
+			Elem: &schema.Resource{
+				Schema: containerFields(),
+			},
+		},
+		"dns_policy": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Default:     "ClusterFirst",
+			Description: "Set DNS policy for containers within the pod. One of 'ClusterFirst' or 'Default'. Defaults to 'ClusterFirst'.",
+		},
+		"host_ipc": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Use the host's ipc namespace. Optional: Default to false.",
+		},
+		"host_network": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Host networking requested for this pod. Use the host's network namespace. If this option is set, the ports that will be used must be specified.",
+		},
+
+		"host_pid": {
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     false,
+			Description: "Use the host's pid namespace.",
+		},
+
+		"hostname": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Description: "Specifies the hostname of the Pod If not specified, the pod's hostname will be set to a system-defined value.",
+		},
+		"image_pull_secrets": {
+			Type:        schema.TypeList,
+			Description: "ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the images used by this PodSpec. If specified, these secrets will be passed to individual puller implementations for them to use. For example, in the case of docker, only DockerConfig type secrets are honored. More info: http://kubernetes.io/docs/user-guide/images#specifying-imagepullsecrets-on-a-pod",
+			Optional:    true,
+			Computed:    true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"name": {
+						Type:        schema.TypeString,
+						Description: "Name of the referent. More info: http://kubernetes.io/docs/user-guide/identifiers#names",
+						Required:    true,
+					},
+				},
+			},
+		},
+		"node_name": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Description: "NodeName is a request to schedule this pod onto a specific node. If it is non-empty, the scheduler simply schedules this pod onto that node, assuming that it fits resource requirements.",
+		},
+		"node_selector": {
+			Type:        schema.TypeMap,
+			Optional:    true,
+			Description: "NodeSelector is a selector which must be true for the pod to fit on a node. Selector which must match a node's labels for the pod to be scheduled on that node. More info: http://kubernetes.io/docs/user-guide/node-selection.",
+		},
+		"restart_policy": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Default:     "Always",
+			Description: "Restart policy for all containers within the pod. One of Always, OnFailure, Never. More info: http://kubernetes.io/docs/user-guide/pod-states#restartpolicy.",
+		},
+		"security_context": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			MaxItems:    1,
+			Description: "SecurityContext holds pod-level security attributes and common container settings. Optional: Defaults to empty",
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"fs_group": {
+						Type:        schema.TypeInt,
+						Description: "A special supplemental group that applies to all containers in a pod. Some volume types allow the Kubelet to change the ownership of that volume to be owned by the pod: 1. The owning GID will be the FSGroup 2. The setgid bit is set (new files created in the volume will be owned by FSGroup) 3. The permission bits are OR'd with rw-rw---- If unset, the Kubelet will not modify the ownership and permissions of any volume.",
+						Optional:    true,
+					},
+					"run_as_non_root": {
+						Type:        schema.TypeBool,
+						Description: "Indicates that the container must run as a non-root user. If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does.",
+						Optional:    true,
+					},
+					"run_as_user": {
+						Type:        schema.TypeInt,
+						Description: "The UID to run the entrypoint of the container process. Defaults to user specified in image metadata if unspecified",
+						Optional:    true,
+					},
+					"supplemental_groups": {
+						Type:        schema.TypeSet,
+						Description: "A list of groups applied to the first process run in each container, in addition to the container's primary GID. If unspecified, no groups will be added to any container.",
+						Optional:    true,
+						Elem: &schema.Schema{
+							Type: schema.TypeInt,
+						},
+					},
+					"se_linux_options": {
+						Type:        schema.TypeList,
+						Description: "The SELinux context to be applied to all containers. If unspecified, the container runtime will allocate a random SELinux context for each container. May also be set in SecurityContext. If set in both SecurityContext and PodSecurityContext, the value specified in SecurityContext takes precedence for that container.",
+						Optional:    true,
+						MaxItems:    1,
+						Elem: &schema.Resource{
+							Schema: seLinuxOptionsField(),
+						},
+					},
+				},
+			},
+		},
+		"service_account_name": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Description: "ServiceAccountName is the name of the ServiceAccount to use to run this pod. More info: http://releases.k8s.io/HEAD/docs/design/service_accounts.md.",
+		},
+		"subdomain": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: `If specified, the fully qualified Pod hostname will be "...svc.". If not specified, the pod will not have a domainname at all..`,
+		},
+		"termination_grace_period_seconds": {
+			Type:         schema.TypeInt,
+			Optional:     true,
+			Default:      30,
+			ValidateFunc: validateTerminationGracePeriodSeconds,
+			Description:  "Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period will be used instead. The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal. Set this value longer than the expected cleanup time for your process.",
+		},
+
+		"volume": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "List of volumes that can be mounted by containers belonging to the pod. More info: http://kubernetes.io/docs/user-guide/volumes",
+			Elem:        volumeSchema(),
+		},
+	}
+}
+
+func volumeSchema() *schema.Resource {
+	v := commonVolumeSources()
+
+	v["config_map"] = &schema.Schema{
+		Type:        schema.TypeList,
+		Description: "ConfigMap represents a configMap that should populate this volume",
+		Optional:    true,
+		MaxItems:    1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"items": {
+					Type:        schema.TypeList,
+					Description: `If unspecified, each key-value pair in the Data field of the referenced ConfigMap will be projected into the volume as a file whose name is the key and content is the value. If specified, the listed keys will be projected into the specified paths, and unlisted keys will not be present. If a key is specified which is not present in the ConfigMap, the volume setup will error. Paths must be relative and may not contain the '..' path or start with '..'.`,
+					Optional:    true,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"key": {
+								Type:        schema.TypeString,
+								Optional:    true,
+								Description: "The key to project.",
+							},
+							"mode": {
+								Type:        schema.TypeInt,
+								Optional:    true,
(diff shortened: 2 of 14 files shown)


ashishth09:
Hi @radeksimko , 
I made some more changes basically putting the Compute/ForceNew  attributes according to my understanding from the APIs and some testing. But my earlier problem as mentioned above remains.
So basically when a Pod is created the default secret is being made available to it via volumes 
For example , the below snippet is obtained on "describing"  a pod. 
```Volumes:
  default-token-ndgzz:
    Type:	Secret (a volume populated by a Secret)
    SecretName:	default-token-ndgzz
```
So when user inputs some specific volumes to be mounted in the "tf" file what comes back in the API read is a Pod which has 2 volumes , one the user specified and the one k8s added by default.

So, I was thinking to loop over the user input of volumes in the tf configuration and just check  it is present in the volumes obtained from the API in the Read operation phase without setting back the schema ( as schema already has the user input). 

Is there any other way to deal with such cases where some resources get attached/created over and above the user input?

builtin/providers/kubernetes/structures.go
```diff
@@ -198,3 +199,89 @@ func newStringSet(f schema.SchemaSetFunc, in []string) *schema.Set {
...
+	return att
+}
+
+func expandVolumes(volumes []interface{}) ([]api.Volume, error) {
+	if len(volumes) == 0 {
+		return []api.Volume{}, nil
+	}
+	vl := make([]api.Volume, len(volumes))
+	for i, c := range volumes {
+		v := c.(map[string]interface{})
```
radeksimko: Just for clarity it would be nice to perhaps call it `m` (as "map") rather than `v`. That will free you up that variable name below where you can use it as "value".

While "value" is not great variable name, the map here is more significant because it is used in more than 2 lines - as opposed to 3-line conditionals where it matters less.

builtin/providers/kubernetes/schema_pod_spec.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+		"volumes": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "List of volumes that can be mounted by containers belonging to the pod. More info: http://kubernetes.io/docs/user-guide/volumes",
+			Elem:        volumeSchema(),
+		},
+		"containers": {
+			Type:        schema.TypeList,
+			Optional:    true,
```
radeksimko: It looks like this field cannot be updated, so I think we'll need to mark it as such via `ForceNew: true`.

ashishth09:
Hi @radeksimko ,

I was waiting for your reply  :smile:  So during that time I had fixed most of the issues you have mentioned. Right, testing is going to be a key here. The schema has tons of fields.  In cases , I skipped some fields as you mentioned. I was just curious to see this up and running with very basic things working first. No other reasons to leave them out :wink: 

I will update the PR soon with the new changes.

Thanks for  inputs and pointers. I didn't know about https://github.com/radeksimko/terraform-gen. This is going to be a great tool to use :+1:

builtin/providers/kubernetes/schema_pod_spec.go
ashishth09: Given that deployment or replication controllers etc may use pod's schema and they would allow the containers to be updated, sort of, basically they would destroy and create a new set of containers internally. So I thought it would be better to leave it as it is. As providing ForceNew and using the same schema in deployment would lead to deployment destruction. I am not very sure how you would typically handle this one. Perhaps by not making the deployment depend on the pod schema?. This would lead to lots of code duplication though.

builtin/providers/kubernetes/structures_pod.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// Flatteners
+func flattenPodSpec(in v1.PodSpec) []interface{} {
+	att := make(map[string]interface{})
+	if in.ActiveDeadlineSeconds != nil {
+		att["active_deadline_seconds"] = *in.ActiveDeadlineSeconds
+	}
+	if in.DNSPolicy != "" {
+		att["dns_policy"] = in.DNSPolicy
```
ashishth09: Looks like the terraform core is somehow taking care of serializing the right type ? Seems like magically working. I will check this again and put the cast.

builtin/providers/kubernetes/schema_pod_spec.go
radeksimko: 🤔  I'm not sure I understand entirely, but either we need to perform a real update/patch API operation for this field or mark it as ForceNew - there is no other expectation from the user - either they want it to be updated or recreated, certainly not ignored.

I will bump `kubernetes_replication_controller` resource in priority to understand better how is it affected by pod in this context, but I won't get to it until Friday as I'm off to Google Next on Wed+Thu.

radeksimko: Some food for thought:
I wonder if it's possible to reference a previously created PodTemplate via ObjectMeta:
https://github.com/kubernetes/kubernetes/blob/v1.5.3/pkg/api/types.go#L1968

Not that we could avoid the coupling between potential `kubernetes_pod_template` and `kubernetes_pod` but at least it's more obvious for these resources and more importantly recreation wouldn't hurt if we could then just update the reference rather than the whole spec.

Recreating whole RC just to update PodSpec is indeed overkill - I agree with you here.

ashishth09:
Hi @radeksimko,
Before I add in more changes, I think I would like to get the versioning thing right.
I am following the v1.5 APIs as I see the main provider.go referred that from the very beginning. So , I assume that is the one we are supporting. For example pod of v1.5 doesn't have the _initContainers_ while v1.6 has it in addition to other fields which are not present in the v1.5.
https://kubernetes.io/docs/api-reference/v1.5/#podspec-v1

radeksimko:
Hi @ashishth09 
that's a very good point. For the initial phase (until we get all stable K8S resources covered) I'd like to stick with `1.5.3`. We can deal with 1.6+ later.

I was following the source code of vendored K8S code and that does contain `initContainers` 🤔 
https://github.com/kubernetes/kubernetes/blob/v1.5.3/pkg/api/types.go#L1776-L1777

My Google-fu allowed me to discover https://kubernetes.io/docs/concepts/workloads/pods/init-containers/ which says

> NOTE: This feature is beta in Kubernetes 1.5.

so feel free to leave this one out as we only aim to cover stable fields/resources - sorry for the confusion.

radeksimko:
@ashishth09 I just spent some time reading more about Pods & PodTemplates & RCs and you're right in the sense that containers cannot be added nor removed for any pods, whereas RCs can update the attached pod with pretty much no limitations.

The `TemplateRef` attribute was removed in the early days of K8S and wasn't added back since then: https://github.com/kubernetes/kubernetes/pull/8980 so that won't help us here.

In addition to that the schema library doesn't have a way to express this behaviour cleanly (i.e. we have not implemented "ForceNewWhen" yet). To keep things simple I'd just fall back to the API validation and let invalid updates (which are adding or removing whole containers) pass through.
It's not ideal, but better than marking everything as `ForceNew` as you mentioned. In other words leave it as is.

I will go through the PR again and see if there's anything left. Based on your recent updates I'm assuming it's ready for another round of review.

builtin/providers/kubernetes/resource_kubernetes_pod.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return err
+	}
+
+	userSpec, err := expandPodSpec(d.Get("spec").([]interface{}))
+	if err != nil {
+		return err
+	}
+
+	err = d.Set("spec", flattenPodSpec(pod.Spec, userSpec))
```
radeksimko: I'm not sure I understand this - why exactly do you need the config (`userSpec`) here? We should be able to filter out the internal volumes without it, I think - unless I'm missing something 🤔

builtin/providers/kubernetes/structures_pod.go
```diff
@@ -0,0 +1,494 @@
...
+	for _, v := range userVolume {
+		userVolumeNames[v.Name] = true
+	}
+
+	att := make([]interface{}, len(userVolume))
+	for i, v := range volumes {
+		obj := map[string]interface{}{}
+
+		if !userVolumeNames[v.Name] {
+			//Found a volume which was not provided by user in the tf configuration file and must be internal
```
radeksimko: This is a bit tricky situation here. It is a convention and expectation from the user that Terraform will detect any drifts of the config from the reality - i.e. if someone adds a volume outside of Terraform (e.g. by calling K8S API directly or just by using kubectl) we need to be able to show that to the user.

At this moment with this current implementation we'd hide it as we're expecting that the config is single point of truth.

Also we would be wiping those internal volumes during each update, which makes me think that `DiffSuppressFunc` would be good candidate for this - instead of having the filtering logic in flattener and expander.

Is there any reason you want to avoid the proposed solution I mentioned earlier?

builtin/providers/kubernetes/structures_pod.go
```diff
@@ -0,0 +1,494 @@
...
+			vl[i].AzureDisk = expandAzureDiskVolumeSource(v)
+		}
+		if v, ok := m["photon_persistent_disk"].([]interface{}); ok && len(v) > 0 {
+			vl[i].PhotonPersistentDisk = expandPhotonPersistentDiskVolumeSource(v)
+		}
+	}
+	return vl, nil
+}
+
+func expandPodTemplateSpec(p []interface{}) (v1.PodTemplateSpec, error) {
```
radeksimko: As far as I can see this function is currently unused? 👀  I'm happy for you to help with the implementation of RC 😃 , but I'd prefer if kept this PR only focused on Pod.

builtin/providers/kubernetes/structures_pod.go
```diff
@@ -0,0 +1,494 @@
...
+		containers := d.Get(prefix + "containers").([]interface{})
+		value, _ := expandContainers(containers)
+
+		for i, v := range value {
+			ops = append(ops, &ReplaceOperation{
+				Path:  pathPrefix + "/containers/" + strconv.Itoa(i) + "/image",
+				Value: v.Image,
+			})
+		}
+
```
radeksimko: I'm assuming there's still some work left to do in terms of missing fields?

builtin/providers/kubernetes/validators.go
```diff
@@ -80,3 +92,70 @@ func validateResourceList(value interface{}, key string) (ws []string, es []erro
...
+	return
+}
+
+func validateProbeTimeoutSeconds(value interface{}, key string) (ws []string, es []error) {
+	v := value.(int)
+	if v <= 0 {
+		es = append(es, fmt.Errorf("%s must be greater than 0", key))
+	}
+	return
+}
```
radeksimko: Since they're all the same, how would you feel about merging all the above validation functions into one and call it e.g. `validatePositiveInteger`?

builtin/providers/kubernetes/structures_pod.go
ashishth09: I had this  locally on my laptop last time before I had asked for help :smile:. So I am yet to implement the logic you suggested. This would take care of the _userSpec_ in the above review comment

ashishth09:
Hi @radeksimko ,

Thanks for the review comments. Right still some work to do in terms of missing fields. Basically the containers one now. I think I have Pod level one's covered. I will recheck.

I think I have those tests passing though :smile:  I am running them against minikube on my system.

```==> Checking that code complies with gofmt requirements...
2017/05/10 21:58:20 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/kubernetes -v -run TestAccKubernetesPod_basic -timeout 120m
=== RUN   TestAccKubernetesPod_basic
--- PASS: TestAccKubernetesPod_basic (3.26s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/kubernetes	3.328s
```
```
2017/05/10 21:59:30 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/kubernetes -v -run TestAccKubernetesPod_with_pod_security_context -timeout 120m
=== RUN   TestAccKubernetesPod_with_pod_security_context
--- PASS: TestAccKubernetesPod_with_pod_security_context (1.51s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/kubernetes	1.586s
```

builtin/providers/kubernetes/structures_pod.go
ashishth09: I think we can only update the _active_deadline_seconds_ and container's _image_ in a Pod.

radeksimko:
The tests are certainly & consistently failing on GKE using the test setup from https://github.com/hashicorp/terraform/blob/master/builtin/providers/kubernetes/test-infra/main.tf

See https://gist.github.com/radeksimko/ee386a97e5cd809fe44b03fe373c1265

ashishth09:
@radeksimko , Thanks for putting it out. I will take a look.

ashishth09:
@radeksimko , 

I see the following lines in the gist
```
		  spec.0.containers.0.resources.0.limits.% = 0
		  spec.0.containers.0.resources.0.requests.% = 1
		  spec.0.containers.0.resources.0.requests.cpu = 100m
```

The test case doesn't set those values. 
I hope I am running the same test as you are :thinking: 

I need to see if  there is some default values getting set internally by the k8s :smiley:  based on some rules at the cluster level itself.
I know some limit ranges can be set at the namespace level. So I need to check that also for future cases to see if they would come back during the read operation as default values set.

For requests I am unsure though. I don't think k8s will set any default value for the requests as it should come from the user. 
Anyway, I will  look into this in some detail. At this point it looks like the prepped cluster rules might be different for both of us like limit ranges etc, which might  help find subtle bugs , which is good :smile:

ashishth09:
Hi @radeksimko ,

So I created a limit range at the cluster level on some namespace and ran the test on the same namespace. I was able to re-create the issue. Now the problem is how to solve it :smiley: 

 The resources can have limits and requests field which can get default values if some one has created a limits resource on the cluster.


Containers resources can't be updated for a Pod. So no one can add/remove resources for a container outside terraform also. Hence no drift can be seen from outside as well. 

So should I just check if the input value for resources in tf file is same as what we get on reading back the pod and ignore persisting  it if it differs.

ashishth09:
Hi @radeksimko , 
I also found another case like this 

```
	DIFF:
		
		UPDATE: kubernetes_pod.test
		  spec.0.image_pull_secrets.#:      "1" => "0"
		  spec.0.image_pull_secrets.0.name: "bluemix-default-secret" => ""
		
		STATE:
		
		kubernetes_pod.test:
		  ID = default/tf-acc-test-fd11k346y7
		  metadata.# = 1
		  metadata.0.annotations.% = 0
		  metadata.0.generate_name = 
		  metadata.0.generation = 0
```
So if _image_pull_secrets_ is not defined in the user tf file but someone has defined the secret at the cluster level then we we read back we get those results which we should ignore. 

So at this point we need to handle
* _image_pull_secrets_ from pod
* resources from containers as described in the previous comment.

My first thought is to just persist them if they are there in the tf file else ignore.
Please let me know your thoughts.

radeksimko:
Hi @ashishth09 👋 
see my answers below ⬇️ 

> Containers resources can't be updated for a Pod. So no one can add/remove resources for a container outside terraform also. Hence no drift can be seen from outside as well.

The fact that a part of a resource is not updatable doesn't mean that a drift cannot occur. Someone can still recreate the resource under the same name with different attributes. More importantly we usually stick a to this convention (of always detecting any possible drift) because it's how the most basic import mechanism (passthrough) works. It just calls `Read()` which should simply populate all fields from the API without relying on any config.

See how we verify this behaviour in other resources by `grep`ing for `_importBasic` tests.

> My first thought is to just persist them if they are there in the tf file else ignore.

For all of the cases where a drift may occur because there are some internal "subresources" the preferred approach is to detect such internal subresources and filter those out without checking any user config.

In this specific case I'd suggest we look up some details about `bluemix-default-secret` secret from the K8S secrets API and verify it's a default secret (there is hopefully something special about it which differentiates it from user-defined secrets for different context).

I hope it makes sense. 😃

ashishth09:
Hi @radeksimko ,

Sure, I get your drift (pun intentional :smiley:.) I think before proceeding any further to fine tune things, I will need your review of the _image_pull_secrets_ and the _resource_requirements_ of the containers.

From what I see _image_pull_secrets_  can get the same value as you define or if you don't then if service account has those defined. So I made it ```compute: true, optional: true ```

For the container resource requirements https://kubernetes.io/docs/api-reference/v1.5/#resourcerequirements-v1 too, if there is any _limitrange_ define on the cluster then the default values from that could be assigned  as the resource requirements of the containers, else whatever use provides that is picked. So I made the individual fields ```compute: true, optional: true ```. I am not sure if this is the right way.

radeksimko:
Hah, drift 😄 

I will have a look tomorrow - I was working on `kubernetes_service_account` in the past days which has very similar challenges like pod here - it gets a default secret created and assigned to it which causes diffs.

I really struggled to reliably identify what's default (machine-managed) and what may come from the user, so I totally feel your pain.

radeksimko:
Hi @ashishth09 
just to give you a quick update - I spent some time looking around for solutions of the mentioned drift problem.

The cleanest approach which I think we should take is to enforce this new flag when creating the pod:
https://github.com/kubernetes/kubernetes.github.io/pull/2769/commits/c1b88aa4161fbe783ab75cf035d49cccd567011b#diff-2c0d882359718c4c69c67099bed2156cR38

Unfortunately it was added in 1.6 so I'll need to bump the vendored k8s repo to support this. I thought originally it could be avoided until we finish implementing all `v1` resources, but this seems like a good reason to do it sooner.

I will hold off on submitting Service Account because of the same problem and RC (because it will reuse part of the schema from your PR here). I will see if I can finish `kubernetes_horizontal_pod_autoscaler` as that would be the last `v1` resource we can cover prior to upgrading to `1.6`.

Thank you for the effort so far & for the patience. I hope it all makes sense? 😅

radeksimko:
^ that is to address the secrets drift - I haven't looked into the other problem you mentioned about limit ranges yet

ashishth09:
I think by  enforcing that flag  we are letting go of 
_You can access the API from inside a pod using automatically mounted service account credentials_   
I am not sure if this is something that user would need to use frequently, probably only when they need to debug something or may be never :smiley: So we should be fine turning auto-mounting off.


I think enforcing that flag would work for only "Mountable secrets" as seen in the output below 
```
kubectl  describe sa
Name:		default
Namespace:	default
Labels:		<none>

Image pull secrets:	myregistrykey

Mountable secrets: 	default-token-ndgzz

Tokens:            	default-token-ndgzz
```

So on a pod, I need to know whether the image_pull_secret is internal or not. I don't think we can find that out without taking user input into consideration. 

Case 1) User specifies ["myregistrykey"]. We persist it even though it matches the service account "Image pull secrets".

Case 2) User specifies ["abcde"]. That secrets exists in the same namespace but it is not part of the current service account. We need to persist it.

ashishth09:
Hi @radeksimko ,

I think I will take this opportunity to learn the correct usage of _optional_ and _compute_ both set to true on a schema paramter. My understanding is if something could be assigned by server side of the API even if user doesn't give the value for that then it should be  _optional_ +  _computed_ in the schema.

Is the Image pull secrets above not really one?. I mean If I don't provide one  in tf file then one gets assigned to if one is set at the service account level. Else it sets what the user provides. 

So I don't have to worry if something like "myregistrykey" shows up in the terraform state even if the  tf configuration file doesn't have that because I make it optional + computed.
Let me know I am missing something.

Thanks

radeksimko:
@ashishth09 you're right, but this rule only applies if the optional user input replaces the default value completely, not when the default value is a value that you add up to. There's no such thing as "partially optional" - like half of the map/list being computed and half defined by the user. If a field is `Optional` & `Computed` it just means it can be **either** computed **or** optional, but never both at the same time.

In other words - even if you specify your own image pull secrets AFAIK the kubelet will assign the default secrets anyway and you'd be then fighting with kubelet by wiping that default secret in any subsequent apply as it would keep appearing in the diff.

Correct me if I'm wrong, but I remember looking at that specific behaviour a few weeks back and it behaved that way.

-----

re my current state - I'm dealing with vendoring issues of k8s 1.6.4, but overall the resource coverage (after adding service + HPA) is now as far as we can get without this upgrade, so this upgrade is my main focus.

ashishth09:
@radeksimko ,

Sure even I got very little time in the last week to make any progress. 
I just  tried creating a secret `kubectl create secret docker-registry regsecret --docker-server=<your-registry-server> --docker-username=<your-name> --docker-password=<your-pword> --docker-email=<your-email>` and used that in the tf configuration. 

It picked up that secret only rather than the default secret defined on the service account. There was only one secret , the secret given by user. So looks like it really is optional + computed holding only one of them at a time. Thanks for confirming the definition :smiley: While testing I did find a bug also in my code, I will push that today.

While we are at it  :smiley:  (and I am being selfish to learn things from you) was wondering about how  optional + computed  fields behave . Since it is computed and someone changes the field outside terraform  we will not know the changes made to it if that field is not present in the tf file. Is that fine?  I sense some drift occurring here.

radeksimko:
> Since it is computed and someone changes the field outside terraform we will not know the changes made to it if that field is not present in the tf file.

If the field is not defined in the config then yes - it will behave exactly that way. That's why it's best to avoid Computed if it's not necessary and/or use `Default` instead where possible.

radeksimko:
FYI - just opened https://github.com/hashicorp/terraform/pull/14923 - also verified that it's _actually_ fixing our problem with pod here. Feel free to subscribe to that PR to get notified when it's merged - then you can rebase and avoid all the complex logic for mounted secrets and then we can have discussion & debugging around the limit ranges.

Also you were right that this isn't solving similar problem with service account. I have just raised a question about that at SO: https://stackoverflow.com/questions/44262350/how-to-avoid-default-secret-being-attached-to-serviceaccount but I have a feeling it may not be possible and we'll probably end up filing an issue to upstream (K8S).

radeksimko:
@ashishth09 The PR upgrading to 1.6.1 was merged and I'd like to get this PR to the finish line and eventually help you take it there next week, if possible. Let me know how you see it. 😃 

If you don't have any time next week, that's 👌  as well, just let me know and I can finish it for you (and keep the authorship of existing commits, obvs).

ashishth09:
Hi @radeksimko ,
Just changed the API references and all the tests passed :+1: . I don't want to slow this down due to my unavailability. Please take this forward. Thanks for all the support so far. Greatly appreciate :+1: 

Two things (apart from some new fields in the new API) that I feel are pending: 
a) limit ranges issues - 
b) I guess even if pods only updatable fields are , container images and `activeDeadlineSeconds` , we would need to take care of other fields in the `patchPodSpec` function. So if some one tries to update other fields he would at least get error from the API that they are not updatable.

I would be interested to know how we handle the limit range issues though. I had made them `optional` and `computed`.
Thanks

radeksimko:
I will take care of the extra updatable fields, but I'm struggling to reproduce the issue with limit ranges 🤔  It looks like the behaviour has changed in 1.6 or I forgot what's the exact configuration needed to trigger the problem.

I tried this:

```hcl
resource "kubernetes_namespace" "test" {
  metadata {
  	name = "blablah"
  }
}

resource "kubernetes_pod" "test" {
  metadata {
    name = "radek-pod-test"
    namespace = "${kubernetes_namespace.test.metadata.0.name}"
  }

  spec {
    containers {
      image = "nginx:1.7.9"
      name  = "containername"
    }
  }
  depends_on = ["kubernetes_limit_range.test"]
}

resource "kubernetes_limit_range" "test" {
  metadata {
    name = "radek-test-lr"
    namespace = "${kubernetes_namespace.test.metadata.0.name}"
  }
  spec {
    limit {
      type = "Pod"
      min {
        cpu = "1m"
        memory = "1M"
      }
    }
    limit {
      type = "Container"
      default_request {
        cpu = "50m"
        memory = "24M"
      }
    }
  }
}
```
and saw no drift.

Would you mind sharing any more details about your setup and/or your configs? Thanks

ashishth09:
@radeksimko ,

 If there is a default request/limit set at the cluster level then K8s applies that to the Pod's resource requirements. If the Pod itself defines its own requirements then K8s would apply the pod's requirements. Following this, I set limits/requests to be computed and the problem was gone.

The change I did was - https://github.com/hashicorp/terraform/blob/16ef37dd6d5908f9d5c369e2a7fa5849921b402a/builtin/providers/kubernetes/schema_container.go#L92-L143

I wanted your view on this change, if you think that's the right way to do or not :smiley:

radeksimko:
@ashishth09 As far as I can tell limit ranges are namespace-specific:
https://kubernetes.io/docs/tasks/administer-cluster/cpu-memory-limit/ and resource quotas likewise: https://kubernetes.io/docs/concepts/policy/resource-quotas/

What do you mean my cluster-level limits and how does one set such limits?

ashishth09:
@radeksimko , 

Yes you are right. So the points I mentioned above are applicable to this namespace. Any new pods created in the namespace go though this rule. If default limits are set and you have not defined  any in Pod then that is what is attached to the pod else Pod's limits are honored. This is what I observed.

I am not sure what happens when more than one default limits are defined in a namespace. If K8s do some kind of `max(limits1, limits2, ...)`,  `min(limits1, limits2, ...)` to determine what will be applied if you don't provide one. Whatever be the case, if we are sure this is optional and computed then I guess we don't need to worry?

radeksimko:
I'd just like to understand how to reproduce the problem and possibly add a regression test - that's all 😄  - hence I'm wondering how did you set such default cluster-wide limits? Is there any example with `kubectl`/YAML or anything like that you can share?

ashishth09:
No there is no cluster-wide limits. As you pasted the link  in previous comment it is on a namespace level :smiley: 
The problem came from https://gist.github.com/radeksimko/ee386a97e5cd809fe44b03fe373c1265#file-k8s-test-acc-failures-txt-L38-L40

So when I first wrote  the code I didn't know that system is applying [default limits]( https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/#if-you-dont-specify-limits-or-requests) to the Pod.
I guess in your environment there was some default limits defined in the same namespace you ran the test.

So to have similar environment at my end, I applied https://k8s.io/docs/tasks/configure-pod-container/limits.yaml to my local `minikube` environment to test my fix of setting limits as optional + computed in schema. 

So you can't reproduce the same problem now :smiley:  I was just curious to know if this was the right thing to do, or it might break in certain other conditions if at all. My worry was mainly from the fact that calculation  of limits seems to be somewhat involved it not very complicated. For example there are some conversions - 

[`For example 100m cpu is 100 milicpu, and is the same as 0.1 cpu.`](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/#understanding-cpu-and-ram-units). So I added some `DiffSuppressFunc` to handle this.

I think we would need to add some tests around this also.

ashishth09:
Just to add to that.. I had no test specifically for limits initially and still that limits issue came which you had shown in your gist. I was not setting any values in the my test configuration (there was no limits block)but k8s  set the default value and hence terraform reported the diff below.

https://gist.github.com/radeksimko/ee386a97e5cd809fe44b03fe373c1265#file-k8s-test-acc-failures-txt-L38-L40

radeksimko:
> So I added some DiffSuppressFunc to handle this.

This was certainly the right decision and I saw a test covering this, so that's 👌 

I cannot reproduce the issue with drifting limits anymore, but I will believe you as I saw that in the past too 😄  so I'll keep it as `Computed`. The only downside of making fields Computed is that Terraform will ignore any changes outside of configs if the field is not defined in the config - e.g. if someone modifies the container resources in the pod resources we won't notice, unless we have it explicitly defined in the config.

TL;DR If the field doesn't _need_ to be computed it should not be.

I have pushed the following final touches to your branch:

 - documentation
 - tests to verify pod is actually being deleted (which resulted in adding retry to `Delete` as deleting pod may take time)
 - test for import
 - removal of all volume filtering logic as it's unnecessary w/ `AutomountServiceAccountToken` set to `false`
 - modified security context in 2 tests because GKE (which is our only testing environment atm) doesn't support containers/processes running as root
 - renamed certain fields as `TypeList` (unless it's special case with `MaxItems: 1`) should be singular due to the way it's structured in the HCL:

```hcl
resource "kubernetes_pod" "test" {
  metadata {
    name = "terraform-example"
  }

  spec {
    container {
      image = "nginx:1.9.7"
      name  = "example"
    }
    container {
      image = "nginx:1.9.6"
      name  = "example-two"
    }
  }
}
```

I know there's another way to structure it w/ `[]` but it's not something we actively encourage people to do - at least not at this point.

ashishth09:
Right, with the earlier code, I think I used to do

```hcl
resource "kubernetes_pod" "test" {
  metadata {
    name = "terraform-example"
  }

  spec {
    containers = [{
      image = "nginx:1.9.7"
      name  = "example"
    },
      {
        image = "nginx:1.9.6"
        name  = "example-two"
      },
    ]
  }
}
```

PR #13574, opened 2017-04-12
providers/aws: add tags for resource_aws_autoscaling_group

s-urbaniak:
The existing "tag" field on autoscaling groups is very limited in that it
is unable to accept a dynamic list of tag entries.

Other AWS resources don't have this restriction on tags because they work
directly on the map type.

AWS autoscaling groups on the other hand have an additional field
"propagate_at_launch" which is not usable with a pure map type.

This fixes it by introducing an additional field called "tags" which
allows specifying a list of maps. This preserves the possibility to
declare tags as with the "tag" field but additionally allows to
construct lists of maps using interpolation syntax.

/cc @philips @Quentin-M @sym3tri @alexsomesan

### DIFF
--- a/builtin/providers/aws/autoscaling_tags.go
+++ b/builtin/providers/aws/autoscaling_tags.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
@@ -12,8 +13,8 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-// tagsSchema returns the schema to use for tags.
-func autoscalingTagsSchema() *schema.Schema {
+// autoscalingTagSchema returns the schema to use for the tag element.
+func autoscalingTagSchema() *schema.Schema {
 	return &schema.Schema{
 		Type:     schema.TypeSet,
 		Optional: true,
@@ -35,11 +36,11 @@ func autoscalingTagsSchema() *schema.Schema {
 				},
 			},
 		},
-		Set: autoscalingTagsToHash,
+		Set: autoscalingTagToHash,
 	}
 }
 
-func autoscalingTagsToHash(v interface{}) int {
+func autoscalingTagToHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
 	buf.WriteString(fmt.Sprintf("%s-", m["key"].(string)))
@@ -52,35 +53,74 @@ func autoscalingTagsToHash(v interface{}) int {
 // setTags is a helper to set the tags for a resource. It expects the
 // tags field to be named "tag"
 func setAutoscalingTags(conn *autoscaling.AutoScaling, d *schema.ResourceData) error {
-	if d.HasChange("tag") {
+	resourceID := d.Get("name").(string)
+	var createTags, removeTags []*autoscaling.Tag
+
+	if d.HasChange("tag") || d.HasChange("tags") {
 		oraw, nraw := d.GetChange("tag")
 		o := setToMapByKey(oraw.(*schema.Set), "key")
 		n := setToMapByKey(nraw.(*schema.Set), "key")
 
-		resourceID := d.Get("name").(string)
-		c, r := diffAutoscalingTags(
-			autoscalingTagsFromMap(o, resourceID),
-			autoscalingTagsFromMap(n, resourceID),
-			resourceID)
-		create := autoscaling.CreateOrUpdateTagsInput{
-			Tags: c,
-		}
-		remove := autoscaling.DeleteTagsInput{
-			Tags: r,
+		old, err := autoscalingTagsFromMap(o, resourceID)
+		if err != nil {
+			return err
 		}
 
-		// Set tags
-		if len(r) > 0 {
-			log.Printf("[DEBUG] Removing autoscaling tags: %#v", r)
-			if _, err := conn.DeleteTags(&remove); err != nil {
-				return err
-			}
+		new, err := autoscalingTagsFromMap(n, resourceID)
+		if err != nil {
+			return err
 		}
-		if len(c) > 0 {
-			log.Printf("[DEBUG] Creating autoscaling tags: %#v", c)
-			if _, err := conn.CreateOrUpdateTags(&create); err != nil {
-				return err
-			}
+
+		c, r, err := diffAutoscalingTags(old, new, resourceID)
+		if err != nil {
+			return err
+		}
+
+		createTags = append(createTags, c...)
+		removeTags = append(removeTags, r...)
+
+		oraw, nraw = d.GetChange("tags")
+		old, err = autoscalingTagsFromList(oraw.([]interface{}), resourceID)
+		if err != nil {
+			return err
+		}
+
+		new, err = autoscalingTagsFromList(nraw.([]interface{}), resourceID)
+		if err != nil {
+			return err
+		}
+
+		c, r, err = diffAutoscalingTags(old, new, resourceID)
+		if err != nil {
+			return err
+		}
+
+		createTags = append(createTags, c...)
+		removeTags = append(removeTags, r...)
+	}
+
+	// Set tags
+	if len(removeTags) > 0 {
+		log.Printf("[DEBUG] Removing autoscaling tags: %#v", removeTags)
+
+		remove := autoscaling.DeleteTagsInput{
+			Tags: removeTags,
+		}
+
+		if _, err := conn.DeleteTags(&remove); err != nil {
+			return err
+		}
+	}
+
+	if len(createTags) > 0 {
+		log.Printf("[DEBUG] Creating autoscaling tags: %#v", createTags)
+
+		create := autoscaling.CreateOrUpdateTagsInput{
+			Tags: createTags,
+		}
+
+		if _, err := conn.CreateOrUpdateTags(&create); err != nil {
+			return err
 		}
 	}
 
@@ -90,11 +130,12 @@ func setAutoscalingTags(conn *autoscaling.AutoScaling, d *schema.ResourceData) e
 // diffTags takes our tags locally and the ones remotely and returns
 // the set of tags that must be created, and the set of tags that must
 // be destroyed.
-func diffAutoscalingTags(oldTags, newTags []*autoscaling.Tag, resourceID string) ([]*autoscaling.Tag, []*autoscaling.Tag) {
+func diffAutoscalingTags(oldTags, newTags []*autoscaling.Tag, resourceID string) ([]*autoscaling.Tag, []*autoscaling.Tag, error) {
 	// First, we're creating everything we have
 	create := make(map[string]interface{})
 	for _, t := range newTags {
 		tag := map[string]interface{}{
+			"key":                 *t.Key,
 			"value":               *t.Value,
 			"propagate_at_launch": *t.PropagateAtLaunch,
 		}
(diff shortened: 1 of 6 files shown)


alexsomesan:
👍
I'm working on the same project as @s-urbaniak.
We are building a Kubernetes deployment tool based on terraform. The cluster itself makes use of some predefined tags to detect underlying AWS resources, but we also want to allow the users to set their own custom tags on top, which we take in as TF vars.
In this scenario we needed to feed the `tags` attribute with a merged map of our own tags + the users'
All resources we use supported that, but the auto-scaling groups.

alexsomesan:
@stack72 @radeksimko 

Sorry for pinging directly. Is there any chance to get some initial feedback on this, so that we know if it's a feasible idea?
We're betting pretty strongly on this feature as part of our project. It would be really helpful to know if this change has a chance of making it in or we'd have to plan for vendoring it somehow on our own.

radeksimko:
Hi @s-urbaniak 
thanks for the PR.

I agree that dynamic tags for ASG is a reasonable use case and we should support it. 👍 👌 

We didn't have a good support for complex data structures like list of maps in the past and the situation still isn't ideal but will improve over time.

There are some related caveats to mention:

1) We do not support schema-based validation for maps (_yet_) - so you can't perform validation for anything [other than value type (bool/int/float/string)](https://github.com/hashicorp/terraform/blob/master/helper/schema/schema.go#L1311-L1344)
2) We have been generally treating TypeMap as type suitable for maps with arbitrary keys and values, which isn't the case here as the API has very clear requirements - `key`, `value`, `propagate_at_launch`. This is also why it's very easy to cause the current code in this PR to crash (just leave out `key` or `value` in the config)
3) There is no way to enter a list of maps interactively in the CLI, although I doubt anyone would ever need to do it

Perhaps supporting schema validation for a map like this would be a solution:

```go
Schema: map[string]*Schema{
	"validatableMap": {
		Type: TypeMap,
		Elem: map[string]*Schema{
			"key": {
				Type:         TypeString,
				ValidateFunc: validateKey,
			},
			"value": {
				Type:         TypeString,
				ValidateFunc: validateValue,
			},
			"propagate_at_launch": {
				Type: TypeBool,
			},
		},
	},
},
```

but this is not something to implement on the provider/resource side - at least not now as schema doesn't support such validation and we need to figure out whether/how we're going to support deeper nesting, i.e. map of maps etc.

I need to discuss this with other maintainers which work closer to the core to see what's the best way forward before I give you any specific suggestions.

Thanks for the patience.

s-urbaniak:
Hi @radeksimko, thanks a lot for your detailed analysis!

>We do not support schema-based validation for maps (yet) - so you can't perform validation for anything other than value type (bool/int/float/string)

Agreed, it would be very nice to add maps validation. I am fixing my local branch to simply ignore invalid map values for this use case.

> We have been generally treating TypeMap as type suitable for maps with arbitrary keys and values, which isn't the case here as the API has very clear requirements - key, value, propagate_at_launch. This is also why it's very easy to cause the current code in this PR to crash (just leave out key or value in the config)

That is actually a very good catch! I am already working on fixing that very crash including tests.

> There is no way to enter a list of maps interactively in the CLI, although I doubt anyone would ever need to do it

Our use case involves parametrization via configuration files, so indeed it does not apply.

> I need to discuss this with other maintainers which work closer to the core to see what's the best way forward before I give you any specific suggestions.

That would be great to get some feedback on how to accomplish this feature! In the meanwhile I am working on fixing the mentioned edge cases.

Having said that it would be great to land this as a stop-gap solution once I fix the above edge cases :-)

s-urbaniak:
@radeksimko I fixed the review comment about the potential crashes, also incorporating invalid values into the test. Do you mind to have another look?

s-urbaniak:
@radeksimko I have been brainstorming with @alexsomesan how we can come up with a more sensible solution while still being idiomatic with the current type system and its limitations:

we leave the new attribute `tags` as a map like all the other resources. The bonus would then be that tag key/values from other AWS resources then can be immediately reused for autoscaling groups. Since the autoscaling group additionally has the information `propagate_at_launch` we could provide this information as an additional list of key values which will have this attribute set to `true`.

A sample configuration would then look like this:
```
resource "aws_autoscaling_group" "masters" {
  tag {
    key = "some"
    value = "thing"
    propagate_at_launch = true
  }

  tags = {
    foo = "value1"
    bar = "value2"
    baz = "value3"
  }

  propagate_at_launch = ["foo", "bar"]
}
```

In the example above the tag keys `some`, `foo` and `bar` would gain the setting `propagate_at_launch = true`, the tag key `baz` would gain the setting `propagate_at_launch = true`, because this key is not present in the list.

What do you think?

joshuaspence:
Nice, I like that solution.

alexsomesan:
Just pitching in to highlight the fact that this latter approach using plain maps is exactly in line with the current tagging syntax of the other resources, which also don't validate the tags map.

s-urbaniak:
Hello @radeksimko, just kindly pinging if you had the chance to reiterate on the comments above. Sorry for the noise, and thanks a lot!

builtin/providers/aws/autoscaling_tags.go
```diff
@@ -115,31 +137,83 @@ func diffAutoscalingTags(oldTags, newTags []*autoscaling.Tag, resourceID string)
...
+	}
+
+	var propagate_at_launch bool
+
+	if v, ok := attr["propagate_at_launch"].(bool); ok {
+		propagate_at_launch = v
+	}
+
+	if v, ok := attr["propagate_at_launch"].(string); ok {
+		propagate_at_launch, _ = strconv.ParseBool(v)
```
radeksimko: Is there any specific reason to ignore this error here?
Also - perhaps a nitpick - but we prefer camelCase for internal variables rather than under_score.

builtin/providers/aws/autoscaling_tags.go
```diff
@@ -115,31 +137,83 @@ func diffAutoscalingTags(oldTags, newTags []*autoscaling.Tag, resourceID string)
...
 	return result
 }
 
+func autoscalingTagFromMap(attr map[string]interface{}, resourceID string) *autoscaling.Tag {
+	if _, ok := attr["key"]; !ok {
+		return nil
+	}
+
+	if _, ok := attr["value"]; !ok {
+		return nil
```
radeksimko: Is there any reason to ignore such case as it's an invalid block of code? I think it would make more sense to just return error in case that `key`, `value` or `propagate_at_launch` are not defined.

builtin/providers/aws/resource_aws_autoscaling_group.go
```diff
@@ -244,7 +244,13 @@ func resourceAwsAutoscalingGroup() *schema.Resource {
...
 			},
 
-			"tag": autoscalingTagsSchema(),
+			"tag": autoscalingTagSchema(),
+
+			"tags": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeMap},
+			},
```
radeksimko: Is there any reason we should accept both fields and deal with all kinds of potential edge cases related to that? If not, would you mind making them mutually exclusive via `ConflictsWith`?

radeksimko:
> we leave the new attribute tags as a map like all the other resources.

I see where you're coming from with that idea, but I'm not too inclined to it.

Terraform should make users aware that ASG tags are somewhat special - i.e. require `propagate_at_launch`. I think that splitting this into 2 fields would make it more difficult to validate and manage in general. It would be too easy to forget updating one or the other field.

Also I take it as benefit if we can stick to the conventions in the upstream API - i.e. list of maps as people coming from other tools or generating HCL/JSON may find it useful.

builtin/providers/aws/autoscaling_tags.go
s-urbaniak: There have been no strong reasons, other than "ignore on error". I will introduce returning an error then.

builtin/providers/aws/autoscaling_tags.go
s-urbaniak: The same reasoning as above, "ignore on error" and keep going. I am happy to introduce returning an error and also camelCase'ing.

builtin/providers/aws/resource_aws_autoscaling_group.go
s-urbaniak: Letting `tag` and `tags` be able to merge together just makes it nice when declaring default tags using `tag` and bringing in additional `tags` in like so: https://github.com/coreos/tectonic-installer/blob/45dfc2b/modules/aws/master-asg/master.tf#L35-L47.

I have no strong opinion on letting this be mutually exclusive, but from a logical standpoint I don't see why they should be.

I let you have the final word on this one :-)

s-urbaniak: Oh, since you mentioned if I can make them mutually exclusive: please take a look at my example posted above and let me know if you stand with your opinion to make it mutually exclusive; again I am happy to do so.

s-urbaniak:
@radeksimko No worries at all for the late reply! I appreciate your effort in looking at this! Regarding my second suggestion it was just a suggested workaround. We do like the current solution better too.

I will post an update on this PR tomorrow which will address your review comments, just please let me know your final decision on the mutual exclusiveness on `tag` vs. `tags`.

builtin/providers/aws/resource_aws_autoscaling_group.go
radeksimko: I see what do you mean - static, predefined tags combined with more tags from a variable.
I think this should work just fine

```hcl
tags = [
    {
      key                 = "foo"
      value               = "bar"
      propagate_at_launch = true
    },
    {
      key                 = "lorem"
      value               = "ipsum"
      propagate_at_launch = false
    },
    "${var.additional_tags}"
  ]
```

The main reason to make those mutually exclusive is to avoid having to document and deal with edge cases like duplicate tags or even conflicting ones (which one is the point of truth) and also to avoid any potential confusion in the heads of users - i.e. I don't want users to have to think about edge cases at all.

Also we don't have an acceptance test for importing ASGs yet, but I'm pretty confident it would fail in this PR because you're assuming that there's a config available - i.e. `GetOk()` in `Read()` decides which field to populate and during import none of those fields will be available, so none will be populated.

We need to decide which field is going to be the point of truth for this context. I'd vote for the existing `tag` field to avoid breaking changes.

I'm going to create the mentioned test and cc you in the PR, just for context.

radeksimko: Actually we do have a test for it, just in a different file I was expecting and it _is_ failing on this branch:

```
=== RUN   TestAccAWSAutoScalingGroup_importBasic
--- FAIL: TestAccAWSAutoScalingGroup_importBasic (206.37s)
	testing.go:280: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) {
		}


		(map[string]string) (len=13) {
		 (string) (len=5) "tag.#": (string) (len=1) "1",
		 (string) (len=18) "tag.4237349706.key": (string) (len=7) "FromTag",
		 (string) (len=34) "tag.4237349706.propagate_at_launch": (string) (len=4) "true",
		 (string) (len=20) "tag.4237349706.value": (string) (len=6) "value1",
		 (string) (len=6) "tags.#": (string) (len=1) "2",
		 (string) (len=8) "tags.0.%": (string) (len=1) "3",
		 (string) (len=10) "tags.0.key": (string) (len=9) "FromTags1",
		 (string) (len=26) "tags.0.propagate_at_launch": (string) (len=1) "1",
		 (string) (len=12) "tags.0.value": (string) (len=6) "value2",
		 (string) (len=8) "tags.1.%": (string) (len=1) "3",
		 (string) (len=10) "tags.1.key": (string) (len=9) "FromTags2",
		 (string) (len=26) "tags.1.propagate_at_launch": (string) (len=1) "1",
		 (string) (len=12) "tags.1.value": (string) (len=6) "value3"
		}
```

radeksimko:
@s-urbaniak Thank you for a quick response. Just to add one important thing - can you make sure to run all relevant (that is all ASG) acceptance tests and check they're all passing with this PR?

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAutoScalingGroup_'
```

s-urbaniak:
@radeksimko definitely, thanks for the pointer on those tests, I already am half-through the comments locally (error handling) and will make sure those tests pass too.

s-urbaniak:
@radeksimko PTAL, I think I addressed all the review comments.

Here is the output of the acceptance test. Without deeper investigation I believe the failed `TestAccAWSAutoScalingGroup_ALB_TargetGroups` is unrelated but I need to verify this tomorrow.

```
$ make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSAutoScalingGroup_'
which: no stringer in (/home/sur/go/bin:/home/sur/go/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl)
2017/05/11 17:11:49 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSAutoScalingGroup_ -timeout 120m
=== RUN   TestAccAWSAutoScalingGroup_importBasic
--- PASS: TestAccAWSAutoScalingGroup_importBasic (305.58s)
=== RUN   TestAccAWSAutoScalingGroup_basic
--- PASS: TestAccAWSAutoScalingGroup_basic (287.32s)
=== RUN   TestAccAWSAutoScalingGroup_namePrefix
--- PASS: TestAccAWSAutoScalingGroup_namePrefix (59.83s)
=== RUN   TestAccAWSAutoScalingGroup_autoGeneratedName
--- PASS: TestAccAWSAutoScalingGroup_autoGeneratedName (50.09s)
=== RUN   TestAccAWSAutoScalingGroup_terminationPolicies
--- PASS: TestAccAWSAutoScalingGroup_terminationPolicies (143.87s)
=== RUN   TestAccAWSAutoScalingGroup_tags
--- PASS: TestAccAWSAutoScalingGroup_tags (293.79s)
=== RUN   TestAccAWSAutoScalingGroup_VpcUpdates
--- PASS: TestAccAWSAutoScalingGroup_VpcUpdates (142.76s)
=== RUN   TestAccAWSAutoScalingGroup_WithLoadBalancer
--- PASS: TestAccAWSAutoScalingGroup_WithLoadBalancer (422.11s)
=== RUN   TestAccAWSAutoScalingGroup_withPlacementGroup
--- PASS: TestAccAWSAutoScalingGroup_withPlacementGroup (181.87s)
=== RUN   TestAccAWSAutoScalingGroup_enablingMetrics
--- PASS: TestAccAWSAutoScalingGroup_enablingMetrics (215.47s)
=== RUN   TestAccAWSAutoScalingGroup_suspendingProcesses
--- PASS: TestAccAWSAutoScalingGroup_suspendingProcesses (231.12s)
=== RUN   TestAccAWSAutoScalingGroup_withMetrics
--- PASS: TestAccAWSAutoScalingGroup_withMetrics (87.56s)
--- FAIL: TestAccAWSAutoScalingGroup_ALB_TargetGroups (50.46s)
	testing.go:280: Step 0 error: Error applying: 1 error(s) occurred:
		
		* aws_launch_configuration.foobar: 1 error(s) occurred:
		
		* aws_launch_configuration.foobar: Error creating launch configuration: ValidationError: AMI cannot be described
			status code: 400, request id: 25c54697-3662-11e7-87ef-0968d0bf68c0
=== RUN   TestAccAWSAutoScalingGroup_initialLifecycleHook
--- PASS: TestAccAWSAutoScalingGroup_initialLifecycleHook (264.13s)
=== RUN   TestAccAWSAutoScalingGroup_ALB_TargetGroups_ELBCapacity
--- PASS: TestAccAWSAutoScalingGroup_ALB_TargetGroups_ELBCapacity (347.24s)
FAIL
exit status 1
FAIL	github.com/hashicorp/terraform/builtin/providers/aws	3083.209s
make: *** [Makefile:49: testacc] Error 1
```

radeksimko:
Hi @s-urbaniak 
thank you for making those changes. This is now looking good to me.

I'm just waiting for acceptance tests to finish and if all pass I'll merge it.

s-urbaniak:
@radeksimko Great news, thanks a ton!

luisdavim:
in what version of terraform will this be available? the example posted above by @s-urbaniak doesn't work with 0.11.1

bflad:
@luisdavim `aws_autoscaling_group` tagging has been available for a few months now, since at least Terraform 0.10 which split providers like AWS away from core, [announcement here](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10), and the first version of the split terraform-provider-aws 0.1.0.

You can double check your current core and provider versions with `terraform version` and upgrade following the instructions available in the [Provider Versions section of the documentation](https://www.terraform.io/docs/configuration/providers.html#provider-versions).

If you're still having trouble, I would suggest opening an issue in the [terraform-provider-aws repository](https://github.com/terraform-providers/terraform-provider-aws/).

Hope this helps!

PR #13583, opened 2017-04-12
state/remote/swift: Support Openstack request logging

fatmcgav:
PR #12089 added support for logging request and response bodies to Openstack resource requests. 
This PR extends that to the swift remote state backend. 

In order to support this, had to modify the `LogRoundTripper` struct to use uppercase field names so that they are exposed externally of the package. 

Also extended `LogRoundTripper` to log the request/response headers, as they are useful when working with Swift and container archiving/expiration which uses headers to specify, rather than the request body.

### DIFF
--- a/builtin/providers/openstack/config.go
+++ b/builtin/providers/openstack/config.go
@@ -113,8 +113,8 @@ func (c *Config) loadAndValidate() error {
 	transport := &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: config}
 	client.HTTPClient = http.Client{
 		Transport: &LogRoundTripper{
-			rt:      transport,
-			osDebug: osDebug,
+			Rt:      transport,
+			OsDebug: osDebug,
 		},
 	}
 
--- a/builtin/providers/openstack/types.go
+++ b/builtin/providers/openstack/types.go
@@ -3,7 +3,6 @@ package openstack
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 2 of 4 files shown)


jtopjian:
Nice! I did a cursory review and it looks good. I'll take a look in more detail either today or tomorrow.

fatmcgav:
Cheers... I just had a thought aswell.

Should `LogRoundTripper` be in utils.go, rather than types.go?

jtopjian:
@fatmcgav It _could_. The reason I have it in `types.go` is because it's a method to a type and it keeps everything bundled in one file. `utils.go`, IMO, is for standalone functions.

Thanks for this feature. Printing the headers is definitely useful, especially for Swift, but we should mask some of the sensitive headers printed. For example, when building an instance, this is now printed:

```
x-subject-token: gAAAAABY8DEPwc...
...
x-auth-token: gAAAAABY8DEPwc5AEjh...
```

I think masking the token with `***` would be a better option.

Other than that, looks great :smile:

fatmcgav:
Cool, makes sense on location. 

Making sensitive headers makes sense. I'll refractor the printing of the headers and add some redaction logic...

fatmcgav:
@jtopjian I just pushed https://github.com/hashicorp/terraform/pull/13583/commits/b3da6532d9b139a6897472fa75459e6426decadd which adds a `RedactHeaders` function. 

Thoughts?

jtopjian:
Awesome! Nice work :)

@stack72 This looks good to me.

fatmcgav:
No worries... Happy to squash down the latest commit if cleaner...

PR #13612, opened 2017-04-13
Update Triton API libraries to fix regression with CNS and machine tags.

sean-:


stack72:
<3 thanks @sean- :)

siepkes:
@sean What was the problem you fixed with this commit? I still experience issues with CNS tags and Terraform.

After upgrading from 0.9.0 to Terraform 0.9.4 (this upgrade includes the change from sdc-go to triton-go) using `tags.triton.cns.services` doesn't work correctly for me anymore. It seems Terraform never sees the existing `tags.triton.cns.services` machine tag. It suddenly think it needs to add it. Don't know if it is significant but there is also a second tag attached to the VM called `role` which Terraform has no problems with. Judging from the source of triton-go the `tags.triton.cns.services` get special treatment so I'm guessing Terraform never gets to see the `tags.triton.cns.services` from triton-go?

After Terraform "modifies/adds" the already existing tag to the VM (which succeeds in Triton) Terraform hangs forever in a `Still modifying...` wait loop. 

I observe this behavior with our own private Triton instance (version 20170303T182017Z).

CC @jen20

jen20:
@siepkes I'll investigate today and get this fixed - it sounds like a bug to me.

sean-:
@siepkes Previously, whenever CNS was disabled it would break TF because the value coming back was a `bool` which couldn't be deserialized into the struct.  So that was fixed, but it's possible cns services was a casualty.  TBD.  Either @jen20 or I will look into this shortly (I need to use CNS later today/Monday so if CNS services is broken, then figuring that out will be on my critical path).

Issue #11830, opened 2017-02-09
EMR Security Configuration


wjordan:
[EMR Security Configuration](http://docs.aws.amazon.com//emr/latest/ReleaseGuide/emr-encryption-enable-security-configuration.html) is a new feature released on [Sep 21 2016](https://aws.amazon.com/about-aws/whats-new/2016/09/amazon-emr-now-supports-data-encryption-for-apache-spark-tez-and-hadoop-mapreduce/) that allows a [JSON document](http://docs.aws.amazon.com//emr/latest/ReleaseGuide/emr-encryption-enable-security-configuration.html#emr-encryption-cli-parameters) to be created and associated with an EMR cluster to configure encryption keys, at-rest and in-transit encryption properties for the cluster. Security Configurations are implemented through the following new resource-oriented EMR service APIs:

- [`CreateSecurityConfiguration`](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_CreateSecurityConfiguration.html)
- [`DescribeSecurityConfiguration`](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_DescribeSecurityConfiguration.html)
- [`ListSecurityConfigurations`](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_ListSecurityConfigurations.html)
- [`DeleteSecurityConfiguration`](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_DeleteSecurityConfiguration.html)

A Security Configuration is associated with an EMR cluster through the [`SecurityConfiguration`](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_RunJobFlow.html#EMR-RunJobFlow-request-SecurityConfiguration) Request Parameter of the [`RunJobFlow`](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_RunJobFlow.html) API.

heyarnold1:
I was thinking of implementing this myself.  Before I do have either of you made any progress?

wjordan:
I have not made any progress- I created this issue mostly to follow-up on a feature request I encountered when [answering](http://stackoverflow.com/a/42142791/2518355) a question on StackOverflow, I did not have a need for it myself.

Good luck with the implementation!

catsby:
A new resource `emr_security_configuration` was added in #14080. It does not cover `ListSecurityConfigurations`, that should be a separate data source I think. 

I'd like to close this if you're ok with it 😄

[...]

PR #13620, opened 2017-04-13
enable security configurations to be attached to EMR clusters

heyarnold1:
Security configurations that have been created in the web console or through another API can be attached to EMR cluster.  There are three possible.

1. do not include argument: behaves as is
2. attache security conf. that exists in VPC: successfully attaches during build
3. attache security conf. that does not exist in VPC: produces error 
ValidationException: Security configuration with name '****' does not exist.

Resolves issue #11830

### DIFF
--- a/builtin/providers/aws/resource_aws_emr_cluster.go
+++ b/builtin/providers/aws/resource_aws_emr_cluster.go
@@ -157,6 +157,11 @@ func resourceAwsEMRCluster() *schema.Resource {
 				ForceNew: true,
 				Required: true,
 			},
+			"security_configuration": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+			},
 			"autoscaling_role": &schema.Schema{
 				Type:     schema.TypeString,
 				ForceNew: true,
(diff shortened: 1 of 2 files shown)


stack72:
Hi @heyarnold1 

Thanks for the work here - is there any way that you could add an acceptance test to show that this works as expected?

Thanks

Paul

heyarnold1:
Hi Paul,

Unlike, for example service roles, AWS does not offer a default security configuration.  Without implementing security configurations as a resource I am not sure it is possible to test.  What do you advise?

Henry

heyarnold1:
Hi @stack72 

Since terraform doesn’t provide security configuration providers yet, would you expect that to be written first before this is merged, or can we merge this ahead so that it can become available to people ?
Failing that approach, do you have any standards for mock implementation or tests that I could go for instead perhaps ?

catsby:
Hey @heyarnold1 yeah I think we'd need a separate resource for those, as requested in https://github.com/hashicorp/terraform/issues/11830

catsby:
I implemented EMR Security Configurations in #14080 , I'll merge that one or this one and add a test, in some rational order, once it's approved 😄

catsby:
#14080 was merged! all we need here really is a test that includes the new `emr_security_configuration` resource. If you can copy and existing basic test here and add that in, that will work! If not, I can probably do it next week

catsby:
I'm folding this into #14133 where I've updated the docs a tiny bit and added a test

PR #13623, opened 2017-04-13
Host based routing on ALB possible

koenijn:
Tried to use the Host based routing feature of the ALB (ELB v2.0) with Terraform 0.9.2. Worked perfectly, so updating the documentation.


stack72:
This is fantastic news - thanks for testing and adding it to our docs

Paul

PR #13628, opened 2017-04-13
[WIP] remove maps with empty counts during expand

jbardin:
Addition to #13541

When we encounter maps with empty counts, remove them from the expansion
to prevent already empty sub-elements from being retained.

Convert TestExpand to subtests for easier debugging.

Fix test fixtures with maps that were missing the "%" key. These were
passing because the maps were being inferred by the absence of a "#"
key. These should not happen in practice with properly structured
values.

### DIFF
--- a/flatmap/expand.go
+++ b/flatmap/expand.go
@@ -114,6 +114,12 @@ func expandArray(m map[string]string, prefix string) []interface{} {
 }
 
 func expandMap(m map[string]string, prefix string) map[string]interface{} {
+	// Submaps may not have a '%' key, so we can't count on this value being
+	// here. If we don't have a count, just procede as if we have have a map.
+	if count, ok := m[prefix+"%"]; ok && count == "0" {
+		return map[string]interface{}{}
+	}
+
 	result := make(map[string]interface{})
 	for k := range m {
 		if !strings.HasPrefix(k, prefix) {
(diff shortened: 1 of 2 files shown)


jbardin:
OK, I'm glad there's a test that failed with this. Besides the `, ok` which to prevent a panic, I think this is where dotted map keys are going to fail. It would nice to get a fix for that in Expand too (I can't even remember where the other attempts to parse those live in the code).

flatmap/expand.go
```diff
@@ -114,6 +114,21 @@ func expandArray(m map[string]string, prefix string) []interface{} {
...
+			panic(err)
+		}
+
+		// If the number of elements in this map is 0, then return an
+		// empty map as there is nothing to expand.
+		if num == 0 {
+			return map[string]interface{}{}
+		}
+	}
+
```
svanharmelen: I would do this simpler... As we don't really need `num` here, why not just something like this:

```
if count, ok := m[prefix+"%"]; ok && count == "0" {
  return map[string]interface{}{}
}
```

Other then that LGTM! 😀

jbardin: good point, just left from copy&paste ;)

svanharmelen:
Nice 👍  LGTM!

PR #13643, opened 2017-04-13
Add example of dependencies between resources generated with `count`

aerostitch:
Hello there.

I was quite struggling to see how resources generated with `count` could depend on each-other so I added an example.
I know there's already [this example](https://www.terraform.io/docs/configuration/interpolation.html#using-templates-with-count) showing how to do dependencies between templates and resources using count but as I had hard time finding it (as I was looking for dependencies between resources), I thought it would be helpful to have another one in the page where the interpolation page redirects us when [reading about count](https://www.terraform.io/docs/configuration/interpolation.html#count-information).

Thanks
Joseph

### DIFF
--- a/website/source/docs/configuration/resources.html.md
+++ b/website/source/docs/configuration/resources.html.md
@@ -229,6 +229,50 @@ resource "aws_instance" "app" {
 }
 ```
 
+### Dependencies between resources generated with `count`
+
+You can also have interdependencies between resources generated using count.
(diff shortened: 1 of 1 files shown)


teamterraform:

This PR relates to a pre-0.12 version of this document. I am going to close this outdated PR, but please do not hesitate to open a new PR or issue in the future, and thank you!

PR #13644, opened 2017-04-13
provider/google: Unset the id for resource_google_project if the create operation fails

danawillow:
The current way means that the project creation can fail during the async part (for example, for a quota failure) but terraform thinks it succeeded.


paddyforan:
Don't we typically detect the error and do `d.SetId("")` instead? That way if the failure is a false negative (e.g., the project was actually created) we don't lose it.

danawillow:
Yes (for future reference for anyone looking at this, the reason is because terraform runs in parallel, so some other resource might trigger a crash. If that happens, there's a good chance that this resource was in fact created successfully, so we set the id as early as we can, and then unset it if it does fail)

paddyforan:
LGTM, thanks @danawillow! Merge at will.

PR #13647, opened 2017-04-13
typo in iam_instance_profile docs

ahl:
Not sure if "This." was a memed up way to indicate that `role` is the preferred method, but it seems either unnecessary, unclear, or both. I thought it would be worth updating the example to use the preferred `role` rather than the legacy `roles`


stack72:
LGTM! Thanks for this :)

PR #13650, opened 2017-04-13
azurerm_redis_cache resource missing hostname

codekoala:
The documentation states that an ``azurerm_redis_cache`` resource has a ``hostname`` attribute, but no such attribute is available. This fixes a typo to expose the hostname as expected. Another typo in the documentation for this resource is also fixed.


stack72:
LGTM! Thanks so much for fixing this

PR #13659, opened 2017-04-14
Update cloudwatch_log_subscription_filter.html.markdown

opteemister:
Currently CloudWatch log subscription supports Lambda as a destination. And we can use `aws_cloudwatch_log_subscription_filter` resource for creating subscriptions with Lambda as a destination, but it needs some additional actions. I described them in description, but feel free to improve description if you can say the same better.
This change will help better understand abilities of using this resource.


stack72:
LGTM! Thanks for this :)

PR #13666, opened 2017-04-14
ip_set_descriptors is no longer a required parameter.

charlieoleary:
The `ip_set_descriptors` option is no longer required. IP sets can be created without this parameter, allowing the creation of empty IP-based blacklists.


stack72:
LGTM! Thanks for this :)

PR #13672, opened 2017-04-14
provider/aws: add an option to skip getting the EC2 platforms

svanharmelen:
Even through this call fails silently in case of an error (usually lack of rights), it’s still a pretty extensive call.

In our region (eu-west-1) this can take up to 3 seconds. And since we have a system that involves doing much planning with the option `-refresh=false` these additional 3 seconds are really very annoying and totally not needed.

So being able to choose to skip them would make our lives a little better :wink:

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -97,6 +97,7 @@ type Config struct {
 	Insecure         bool
 
 	SkipCredsValidation     bool
+	SkipGetEC2Platforms     bool
 	SkipRegionValidation    bool
 	SkipRequestingAccountId bool
 	SkipMetadataApiCheck    bool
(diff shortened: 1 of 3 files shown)


radeksimko:
I can understand the reasons and I'm ok with adding the option, but keep in mind that the reasons this was added in the 1st place was to make some decisions about VPC-related resources easier.

We don't actually make any decisions based on that data for now, but we might in the future and not having these details might affect those decisions.

Just FYI. 😉

svanharmelen:
@radeksimko clear... And totally forgot the website! Will of course add a few lines before merging 👍

svanharmelen:
Updated the docs, so going ahead and merging it in 👍

PR #13689, opened 2017-04-16
Fix billing_acount into billing_account in google_project.html.markdown

okumin:
Fix a typo.


stack72:
LGTM - thanks for this :)

PR #12406, opened 2017-03-03
Added other possible statuses when a dms task is successfully modified.

crossing:
When a DMS task is successfully modified, its status can be `ready`, `stopped` or `failed`. The status ends up in `ready` only when the task hasn't been running before modification.

This may fix #12206

PR #13697, opened 2017-04-16
provider/aws: Add support for aws_dms_replication_task available states

stack72:
This supersedes the work in #12406

We have moved away from the AWS waiters - it wasn't fair to ask the origin submitter to update their code when I took so long to merge it

Paul


stack72:
screen shot 2017-04-17 at 02 30 43

PR #13699, opened 2017-04-17
provider/aws: Validate GovCloud KMS ARNs

phinze:
Fixes validation errors in GovCloud when passing a KMS ARN for
`kms_key_id` in `s3_bucket_object` and `db_instance`.

The region is `us-gov-west-1` which breaks the regexp. Just added the
optional `gov-` in the right place.

### DIFF
--- a/builtin/providers/aws/validators.go
+++ b/builtin/providers/aws/validators.go
@@ -353,7 +353,7 @@ func validateArn(v interface{}, k string) (ws []string, errors []error) {
 	}
 
 	// http://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html
-	pattern := `^arn:[\w-]+:([a-zA-Z0-9\-])+:([a-z]{2}-[a-z]+-\d{1})?:(\d{12})?:(.*)$`
+	pattern := `^arn:[\w-]+:([a-zA-Z0-9\-])+:([a-z]{2}-(gov-)?[a-z]+-\d{1})?:(\d{12})?:(.*)$`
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/validators.go
```diff
@@ -353,7 +353,7 @@ func validateArn(v interface{}, k string) (ws []string, errors []error) {
 	}
 
 	// http://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html
-	pattern := `^arn:[\w-]+:([a-zA-Z0-9\-])+:([a-z]{2}-[a-z]+-\d{1})?:(\d{12})?:(.*)$`
+	pattern := `^arn:[\w-]+:([a-zA-Z0-9\-])+:([a-z]{2}-(gov-)?[a-z]+-\d{1})?:(\d{12})?:(.*)$`
```
radeksimko: I think the Chinese region will eventually need similar care as they have their own ARN formats too, but KMS is not supported there (so far at least), so there's no example of a real ARN we could test this against ¯\_(ツ)_/¯

phinze: As you can imagine - I'm fresh on these docs 😄 

The second `partition` field changes in GovCloud and China (Beijing), but we're already loose about the contents of that field. China's `region` identifier follows the "normal" format - `cn-north-1`.

In other words, we should be fine for China!

I'd love to see ARN parsing and validating not be up to a regexp, but for now, 🤷‍♂️

PR #13676, opened 2017-04-15
AWS WAF Regional

yusukegoto:
Resolving conflicts of #11263 coz original author is silent for weeks in spite of his great work.
Add implementations of change token consideration from #13656 .

PR #13705, opened 2017-04-17
provider/aws: AWS WAF Regional IPSet + ByteMatchSet support

yusukegoto:
This PR is from #13676.

test
```
resource "aws_wafregional_ipset" "ipset" {
  name = "tfIPSet"
  ip_set_descriptors {
    type = "IPV4"
    value = "192.0.7.0/24"
  }
}
```

exec
```
aws_wafregional_ipset.ipset: Creating...
  ip_set_descriptors.#:                "" => "1"
  ip_set_descriptors.4037960608.type:  "" => "IPV4"
  ip_set_descriptors.4037960608.value: "" => "192.0.7.0/24"
  name:                                "" => "tfIPSet"
aws_wafregional_ipset.ipset: Creation complete (ID: 38bf9b97-72bb-412c-bd6f-f8c5b911127b)
```

### DIFF
--- a/builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
+++ b/builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
@@ -1,0 +1,197 @@
+package aws
+
+import (
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/waf"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsWafRegionalByteMatchSet() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsWafRegionalByteMatchSetCreate,
+		Read:   resourceAwsWafRegionalByteMatchSetRead,
+		Update: resourceAwsWafRegionalByteMatchSetUpdate,
+		Delete: resourceAwsWafRegionalByteMatchSetDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"byte_match_tuples": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"field_to_match": {
+							Type:     schema.TypeSet,
+							Required: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"data": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"type": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+								},
+							},
+						},
+						"positional_constraint": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"target_string": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"text_transformation": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceAwsWafRegionalByteMatchSetCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafregionalconn
+
+	log.Printf("[INFO] Creating ByteMatchSet: %s", d.Get("name").(string))
+
+	wr := newWafRegionalRetryer(conn)
+	out, err := wr.RetryWithToken(func(token *string) (interface{}, error) {
+		params := &waf.CreateByteMatchSetInput{
+			ChangeToken: token,
+			Name:        aws.String(d.Get("name").(string)),
+		}
+		return conn.CreateByteMatchSet(params)
+	})
+
+	if err != nil {
+		return errwrap.Wrapf("[ERROR] Error creating ByteMatchSet: {{err}}", err)
+	}
+	resp := out.(*waf.CreateByteMatchSetOutput)
+
+	d.SetId(*resp.ByteMatchSet.ByteMatchSetId)
+
+	return resourceAwsWafRegionalByteMatchSetUpdate(d, meta)
+}
+
+func resourceAwsWafRegionalByteMatchSetRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafregionalconn
+
+	log.Printf("[INFO] Reading ByteMatchSet: %s", d.Get("name").(string))
+
+	params := &waf.GetByteMatchSetInput{
+		ByteMatchSetId: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetByteMatchSet(params)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "WAFNonexistentItemException" {
+			log.Printf("[WARN] WAF IPSet (%s) not found, error code (404)", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		return err
+	}
+
+	d.Set("name", resp.ByteMatchSet.Name)
+
+	return nil
+}
+
+func resourceAwsWafRegionalByteMatchSetUpdate(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[INFO] Updating ByteMatchSet: %s", d.Get("name").(string))
+
+	err := updateByteMatchSetResourceWR(d, meta, waf.ChangeActionInsert)
+	if err != nil {
+		return errwrap.Wrapf("[ERROR] Error updating ByteMatchSet: {{err}}", err)
+	}
+	return resourceAwsWafRegionalByteMatchSetRead(d, meta)
+}
+
+func resourceAwsWafRegionalByteMatchSetDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafregionalconn
+
+	log.Printf("[INFO] Deleting ByteMatchSet: %s", d.Get("name").(string))
+
+	err := updateByteMatchSetResourceWR(d, meta, waf.ChangeActionDelete)
+	if err != nil {
+		return errwrap.Wrapf("[ERROR] Error deleting ByteMatchSet: {{err}}", err)
+	}
+
+	wr := newWafRegionalRetryer(conn)
+	_, err = wr.RetryWithToken(func(token *string) (interface{}, error) {
+		req := &waf.DeleteByteMatchSetInput{
+			ChangeToken:    token,
+			ByteMatchSetId: aws.String(d.Id()),
+		}
+		return conn.DeleteByteMatchSet(req)
+	})
+	if err != nil {
+		return errwrap.Wrapf("[ERROR] Error deleting ByteMatchSet: {{err}}", err)
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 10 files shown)


builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
```diff
@@ -0,0 +1,197 @@
...
+	conn := meta.(*AWSClient).wafregionalconn
+
+	wr := newWafRegionalRetryer(conn)
+	_, err := wr.RetryWithToken(func(token *string) (interface{}, error) {
+		req := &waf.UpdateByteMatchSetInput{
+			ChangeToken:    token,
+			ByteMatchSetId: aws.String(d.Id()),
+		}
+
+		ByteMatchTuples := d.Get("byte_match_tuples").(*schema.Set)
```
radeksimko: Perhaps a nitpick, but It's a convention to keep all internal variables (valid only within the context of a function) with lowercased letter at the beginning.

builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
```diff
@@ -0,0 +1,197 @@
...
+			log.Printf("[WARN] WAF IPSet (%s) not found, error code (404)", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		return err
+	}
+
+	d.Set("name", resp.ByteMatchSet.Name)
+
```
radeksimko: I think we're missing some fields here, specifically `byte_match_tuples` and all nested fields under it. The expectation from the Terraform user is that for any resource Terraform will detect drifts from the configuration. In order to do that we need to set all the available data from the API via `d.Set()` here in Read func.

Sets can be slightly trickier to work with, I will try and fix the WAF resources first and point you there for reference instead of trying to explain how this can be done. 😅 

It's probably the cause of some of the reported WAF bugs I saw earlier, so it's worth fixing there either way.

radeksimko: btw. this would also break import functionality when it's actually implemented as it leverages `Read()` (at least the default implementation)

builtin/providers/aws/resource_aws_wafregional_byte_match_set_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSWafRegionalByteMatchSetDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSWafRegionalByteMatchSetConfig(byteMatchSet),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSWafRegionalByteMatchSetExists("aws_wafregional_byte_match_set.byte_set", &v),
```
radeksimko: It would be nice to have some checks for the real object, i.e. `v` - like we have in many other resources, but it's not a blocker for the PR. 👌

builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
```diff
@@ -0,0 +1,197 @@
...
+		Update: resourceAwsWafRegionalByteMatchSetUpdate,
+		Delete: resourceAwsWafRegionalByteMatchSetDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"byte_match_tuples": &schema.Schema{
```
radeksimko: Given how the sets and lists are represented in HCL, i.e.

```hcl
byte_match_tuples {
...
}

byte_match_tuples {
...
}
```
it would make more sense if this field name was singular - i.e. `byte_match_tuple`. I will have a look into deprecating `byte_match_tuples` in favour of singular name in WAF resources too as we seemed to overlook this in the review of that resource code there.

builtin/providers/aws/resource_aws_wafregional_ipset.go
```diff
@@ -0,0 +1,162 @@
...
+		Update: resourceAwsWafRegionalIPSetUpdate,
+		Delete: resourceAwsWafRegionalIPSetDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ip_set_descriptors": &schema.Schema{
```
radeksimko: Similar to the other schema - it would make more sense to call this `ip_set_descriptor` IMO.

builtin/providers/aws/resource_aws_wafregional_ipset.go
```diff
@@ -0,0 +1,162 @@
...
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "WAFNonexistentItemException" {
+			log.Printf("[WARN] WAF IPSet (%s) not found, error code (404)", d.Id())
+			d.SetId("")
+			return nil
+		}
+
+		return err
+	}
+
+	var IPSetDescriptors []map[string]interface{}
```
radeksimko: I don't think this will actually work 🤔  because `ip_set_descriptors` requires Set (i.e. `*schema.Set`), not list (`[]interface{}`).

builtin/providers/aws/resource_aws_wafregional_ipset.go
```diff
@@ -0,0 +1,162 @@
...
+
+	for _, IPSetDescriptor := range resp.IPSet.IPSetDescriptors {
+		IPSet := map[string]interface{}{
+			"type":  *IPSetDescriptor.Type,
+			"value": *IPSetDescriptor.Value,
+		}
+		IPSetDescriptors = append(IPSetDescriptors, IPSet)
+	}
+
+	d.Set("ip_set_descriptors", IPSetDescriptors)
```
radeksimko: Would you mind adding error check here? We tend to avoid error checking for `d.Set()` of all primitive fields merely to avoid cluttering the code, but we tend to do it for lists and sets as we were bitten by not doing it many times in the past.

builtin/providers/aws/wafregionl_token_handlers.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/waf"
+	"github.com/aws/aws-sdk-go/service/wafregional"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+type WafRegionalRetryer struct {
+	Connection *wafregional.WAFRegional
```
radeksimko: I thought my original `WafRetryer` would be useful for you here too, but I never realized it's a different service with a different name 🤦‍♂️  

Oh well, good decision to create a copy of it. 👍

builtin/providers/aws/wafregionl_token_handlers.go
```diff
@@ -0,0 +1,50 @@
...
+			}
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
+
+	return out, err
+}
+
+func newWafRegionalRetryer(conn *wafregional.WAFRegional) *WafRegionalRetryer {
```
radeksimko: I'm aware this will require some refactoring throughout all the resources, but I think it's important to only lock for a given region, so that if folks deploy regional WAF in say 2 regions, one region doesn't wait for lock of the other region as it doesn't have to.

website/source/docs/providers/aws/r/wafregional_byte_match_set.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name or description of the ByteMatchSet.
+* `byte_match_tuples` - Settings for the ByteMatchSet, such as the bytes (typically a string that corresponds with ASCII characters) that you want AWS WAF to search for in web requests.
```
radeksimko: Would you mind documenting the nested fields of `byte_match_tuples` here too? 🙂

website/source/docs/providers/aws/r/wafregional_ipset.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name or description of the IPSet.
+* `ip_set_descriptors` - (Required) The IP address type and IP address range (in CIDR notation) from which web requests originate.
```
radeksimko: Would you mind documenting the nested fields of `ip_set_descriptors` here too? 🙂

radeksimko:
FYI I will hold on reviewing the other WAF Regional PRs until we get this in as those will most likely require rebase and/or some cherry-picking of changes from this PR.

builtin/providers/aws/resource_aws_wafregional_ipset.go
radeksimko: I was wrong here, we actually we do accept `[]interface{}`, so ignore my comment here - I just wasn't aware of it. 🙈

builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
yusukegoto: I agree with this. AWS CLI's api `ByteMatchTuples` requires list but in this case we can write each tuple.
Should I change into `byte_match_tuple` only in WAF Regional? Or leave them in this time and change them later at the same time with WAF?

yusukegoto: uhm... I got the idea resource name should be singular during writing document.

builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
yusukegoto: It seems nested `map[string] interface {}` doesn't go well as others...

builtin/providers/aws/resource_aws_wafregional_byte_match_set.go
radeksimko: > Should I change into byte_match_tuple only in WAF Regional?

Yes, please.

Don't worry about other (WAF) resources - we can rename/deprecate those later - that's a job for a separate PR.

radeksimko:
Hi @yusukegoto 
thanks for the changes you've made so far in ByteMatchSet (singular, docs updates, diffing of tuples). Those are looking ok 👌  

https://github.com/hashicorp/terraform/pull/13766 was merged - would you mind looking over it 👀  and adapting the rest of the changes/fixes there to this PR too? 

Namely:

1. Two tests for ByteMatchSet (set with no tuples + changing existing tuples within a set)
2. WafRegionalRetryer now has Region field, but it needs to be actually in use, ideally as argument to `newWafRegionalRetryer`
3. Two tests for IPSet (set with no descriptors + changing existing descriptors within a set)
4. Turn `ip_set_descriptors` into a singleton - i.e. `ip_set_descriptor`
5. Allow changing existing descriptors within an existing IPSet + allow creating IPSet with no descriptor
6. Document descriptor fields for IPSet

I hope I didn't miss anything 😃 
Let me know if anything's unclear and/or if you need any help in making these changes.

yusukegoto:
@reedloden Thanks for clearing remaining tasks! And sorry for slow response.
I will be able to have much time to fix these on weekend!

yusukegoto:
@reedloden 
hi, I think I could, plz re-review them!!

reedloden:
@yusukegoto, I think you mean @radeksimko, not me. :)

yusukegoto:
@reedloden sorry for disturbing 🙇

yusukegoto:
@radeksimko Thank you very much for additional commits and care!! I'm trying to clean other WIP PRs:smile:

PR #13706, opened 2017-04-17
Add auto_minor_version_upgrade document

gotyoooo:
`auto_minor_version_upgrade` argument in `aws_rds_cluster_instance` is already implemented but document does not exist.

https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_rds_cluster_instance.go#L106

Contents are the same as `aws_db_instance`
https://github.com/hashicorp/terraform/blob/master/website/source/docs/providers/aws/r/db_instance.html.markdown


stack72:
@gotyoooo 

Thanks for this - LGTM! sorry about the omission :)

Paul

PR #13719, opened 2017-04-17
provider/github: Supports managing a team's LDAP DN in GitHub Enterprise

alindeman:


alindeman:
@stack72 @tombuildsstuff This is ready for a look and feedback. We've applied the patch to the vendored version of Terraform we use internally and have started using the `ldap_dn` property.

builtin/providers/github/resource_github_team.go
```diff
@@ -50,6 +55,17 @@ func resourceGithubTeamCreate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
+
+	if ldap_dn := d.Get("ldap_dn").(string); ldap_dn != "" {
```
stack72: I think if the design of this is an update call, then we should maybe have Create -> Update -> Read

That way, we won't have the same logic in both Create and Update

Thoughts?

Issue #13004, opened 2017-03-23
google_project_services creates `dataproc-control.googleapis.com` without it being specified


mogronalol:
### Terraform Version
`Terraform v0.9.1`

### Affected Resource(s)
Please list the resources as a list, for example:
- google_project
- google_project_services

### Terraform Configuration Files
```hcl
resource "google_project" "project" {
  name = "foo"
  project_id = "foo"
  org_id = "180710039370"
  billing_account = "006F6F-62E19A-81CF8C"
}

resource "google_project_services" "project" {
  project = "${google_project.project.id}"
  services = [
    "cloudresourcemanager.googleapis.com",
    "container.googleapis.com",
    "compute-component.googleapis.com",
    "deploymentmanager.googleapis.com",
    "dns.googleapis.com",
    "monitoring.googleapis.com",
    "replicapool.googleapis.com",
    "replicapoolupdater.googleapis.com",
    "resourceviews.googleapis.com",
    "servicemanagement.googleapis.com",
    "storage-api.googleapis.com",
    "storage-component.googleapis.com",
    "dataproc-control.googleapis.com", // Breaks when present or omitted
    "dataproc.googleapis.com"]
}
```

### Debug Output
```
google_project_services.project: Error creating services: Error enabling service "dataproc-control.googleapis.com" for project "foo": googleapi: Error 403: Service 'dataproc-control.googleapis.com' not found or permission denied., forbidden
```

### Expected Behavior
If I don't specify `dataproc-control.googleapis.com` and `dataproc.googleapis.com` then it should not assign these services. If I do specify them then it should be able to create them.

### Actual Behavior

If I don't specify `dataproc-control.googleapis.com` and `dataproc.googleapis.com`, then Google will still set them anyway. This means a subsequent `plan` or `apply` will do a diff, and think that I am removing these two services (Because they are present Google side, but not in the Terraform template, despite them never being specified). I then get a 403 when applying because it's not possible to disable `dataproc-control.googleapis.com`, which is Terraform tries to do because it sees it as being removed.

If I do specify `dataproc-control.googleapis.com` and `dataproc.googleapis.com`, then I get 403 when applying because it's not possible to enable `dataproc-control.googleapis.com` via an API call, which it Terraform tries to do. However I can run apply again, and it will work. This is because Google assigned it anyway, so a subsequent diff during plan will return no diff.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:

1. `terraform apply`
[...]

PR #13730, opened 2017-04-18
provider/google: ignore certain project services that can't be enabled directly via the api

danawillow:
When a project service is enabled, it may also end up enabling another one. We don't yet have a fix to the general problem, but this PR solves a small case of that:
Some of the services (such as dataproc-control.googleapis.com) are only enabled via these side-effects, and can't be enabled or disabled via the api. Explicitly ignore these to avoid issues like #13004.

### DIFF
--- a/builtin/providers/google/resource_google_project_services.go
+++ b/builtin/providers/google/resource_google_project_services.go
@@ -31,6 +31,14 @@ func resourceGoogleProjectServices() *schema.Resource {
 	}
 }
 
+// These services can only be enabled as a side-effect of enabling other services,
+// so don't bother storing them in the config or using them for diffing.
+var ignore = map[string]struct{}{
+	"containeranalysis.googleapis.com": struct{}{},
+	"dataproc-control.googleapis.com":  struct{}{},
+	"source.googleapis.com":            struct{}{},
+}
+
 func resourceGoogleProjectServicesCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 	pid := d.Get("project").(string)
@@ -160,7 +168,9 @@ func getApiServices(pid string, config *Config) ([]string, error) {
 		return apiServices, err
 	}
 	for _, v := range svcResp.Services {
-		apiServices = append(apiServices, v.ServiceName)
+		if _, ok := ignore[v.ServiceName]; !ok {
+			apiServices = append(apiServices, v.ServiceName)
+		}
 	}
 	return apiServices, nil
 }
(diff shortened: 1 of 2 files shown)


builtin/providers/google/resource_google_project_services.go
```diff
@@ -160,7 +168,9 @@ func getApiServices(pid string, config *Config) ([]string, error) {
 		return apiServices, err
 	}
 	for _, v := range svcResp.Services {
-		apiServices = append(apiServices, v.ServiceName)
+		if _, ok := ignore[v.ServiceName]; !ok {
+			apiServices = append(apiServices, v.ServiceName)
+		}
```
paddycarver: I could be wrong on this, but I think we could use a DiffSuppressFunc instead of modifying the state for this. It's not a huge deal, but in the unlikely event the API starts letting us control one of these services in the future, Terraform would know it's already enabled instead of thinking it's not yet. What do you think?

paddycarver: We talked about this, and there was trouble with the count part of the Set and DiffSuppressFunc. I need to ask about this.

danawillow: Yup, and just in case anyone else is wondering, here's what the test output looks like with a DiffSuppressFunc:
```
--- FAIL: TestAccGoogleProjectServices_ignoreUnenablableServices (260.05s)

		DIFF:

		UPDATE: google_project_services.acceptance
		  services.#:          "10" => "9"
		  services.133405307:  "storage-component.googleapis.com" => "storage-component.googleapis.com"
		  services.2692275209: "dataproc.googleapis.com" => "dataproc.googleapis.com"
		  services.2966512281: "deploymentmanager.googleapis.com" => "deploymentmanager.googleapis.com"
		  services.3010261123: "replicapool.googleapis.com" => "replicapool.googleapis.com"
		  services.3075019877: "replicapoolupdater.googleapis.com" => "replicapoolupdater.googleapis.com"
		  services.3077910291: "resourceviews.googleapis.com" => "resourceviews.googleapis.com"
		  services.3731214611: "compute-component.googleapis.com" => "compute-component.googleapis.com"
		  services.3740470850: "container.googleapis.com" => "container.googleapis.com"
		  services.3875785048: "storage-api.googleapis.com" => "storage-api.googleapis.com"
```

paddycarver: Outcome of the conversation is largely that DiffSuppressFunc and Sets/Lists is largely uncharted territory, and we don't know a ton about its behaviour just yet.

I think @danawillow was going to investigate more, but unless she surfaces information that can make this work cleanly, I think the implementation as it exists is as good as this is going to get, and am fine moving forward with it.

danawillow: Yeah I took a look but didn't come up with anything nice. Good to merge?

paddycarver: Good to merge. 👍

Issue #13617, opened 2017-04-13
Race condition in helper/resource/state.go creating EC2 instances that don't appear in tfstate


joelthompson:
### Terraform Version

We're experiencing the issue with 0.8.8 and 0.9.2, and it looks like the affected code is also in master.

### Affected Resource(s)

aws_instance is what we have seen but, in principle, I think the issue is in TF core and can affect really any resource type

### Terraform Configuration Files
```hcl
resource "aws_iam_instance_profile" "MyProfile" {
    name = "MyProfile"
    role = ["MyRole"]
}

resource "aws_security_group" "MySecurityGroup" {
    name_prefix = "MySecurityGroup"
    vpc_id = "vpc-XXXXXXX"
}

resource "aws_instance" "MyInstance" {
    ami = "ami-XXXXXX"
    iam_instance_profile = "${aws_iam_instance_profile.MyProfile.name},"
    instance_type = "t2.micro"
    "key_name" = "my_key_name",
    root_block_device {
        "volume_size" 20
        "volume_type" gp2
    }
    subnet_id = "subnet-XXXXXXX"
    tags {
        Name = "MyInstance"
    },
    user_data = "user_data"
    vpc_security_group_ids = [
        "${aws_security_group.MySecurityGroup.id}"
    ],
    count = 1
}
```

### Debug Output

Can't provide full debug output, sorry

### Expected Behavior

Either the instance gets launched in the given instance profile, and it shows up in the terraform state, or it doesn't get launched at all.

### Actual Behavior

Get this error:

> * aws_instance.MyInstance: Error launching source instance: InvalidParameterValue: Value (MyProfile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name                                                                                                                                                           

But the instance has been launched in the given instance profile, and it doesn't exist in the terraform state, leaving an orphaned (and untagged) resource.

### Steps to Reproduce
Basically, just
1. `terraform apply` repeatedly

### Important Factoids

[...]

PR #13732, opened 2017-04-18
Mitigate race condition leading to orphaned resources being generated

joelthompson:
This addresses #13617 by adding the notion of a grace period to `resource.StateChangeConf` which is the amount of time that we wait for a current execution of the refresh function to finish after the timeout expires but before failing on a timeout error. It also sets the grace period for creating EC2 instances to 5 seconds. This combination seems to fix the problems we're experiencing.

### DIFF
--- a/helper/resource/state.go
+++ b/helper/resource/state.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"log"
+	"sync"
 	"sync/atomic"
 	"time"
 )
@@ -25,6 +26,7 @@ type StateChangeConf struct {
 	Refresh        StateRefreshFunc // Refreshes the current state
 	Target         []string         // Target state
 	Timeout        time.Duration    // The amount of time to wait before timeout
+	TimeoutGrace   time.Duration    // The grace period to wait for the last refresh to finish before timing out
 	MinTimeout     time.Duration    // Smallest time to wait before refreshes
 	PollInterval   time.Duration    // Override MinTimeout/backoff and only poll this often
 	NotFoundChecks int              // Number of times to allow not found
@@ -71,6 +73,7 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 	}
 	var lastResult atomic.Value
 	lastResult.Store(Result{})
+	var refreshMutex sync.Mutex
 
 	doneCh := make(chan struct{})
 	go func() {
@@ -82,6 +85,12 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 		wait := 100 * time.Millisecond
 
 		for {
+			// Intentionally not deferring the Unlock
+			// if this function returns, I don't want to let the timeout
+			// code run; it should pick up on the channel closing
+			// this does mean, however, that any continue statements MUST call
+			// refreshMutex.Unlock() before continue
+			refreshMutex.Lock()
 			res, currentState, err := conf.Refresh()
 			result := Result{
 				Result: res,
@@ -100,6 +109,7 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 				if conf.ContinuousTargetOccurence == targetOccurence {
 					return
 				} else {
+					refreshMutex.Unlock()
 					continue
 				}
 			}
@@ -128,6 +138,11 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 						if conf.ContinuousTargetOccurence == targetOccurence {
 							return
 						} else {
+							// FIXME: I think this continue is buggy, it's continuing the for
+							// loop just above, not continuing on the big for loop to check
+							// the resource status?
+							// If this is fixed, then uncomment the following line
+							// refreshMutex.Unlock()
 							continue
 						}
 					}
@@ -172,6 +187,7 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 			if targetOccurence == 0 {
 				wait *= 2
 			}
+			refreshMutex.Unlock()
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


helper/resource/state.go
```diff
@@ -180,12 +196,48 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
...
+		// refresh attempts and then give it a grace period of 5 seconds (chosen
+		// somewhat arbitrarily) to finish processing the existing call before we
+		// give up. Note that we never actually release the mutex here in order to
+		// prevent the refresh function from getting called yet again.
+		// The length of the grace period is a tradeoff. If we don't have a grace
+		// period, we could leak resources. If we have an infinite grace period,
+		// then we might never timeout if conf.Refresh never finishes.
+		mutexLockCh := make(chan struct{})
+		go func() {
+			refreshMutex.Lock()
```
jbardin: While this will work correctly when the Delay and Timeout are set to sane values, there's nothing ensuring it works in all cases. I prefer to have all the serialization of the goroutines deterministic whenever possible.

helper/resource/state.go
```diff
@@ -82,6 +85,12 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 		wait := 100 * time.Millisecond
 
 		for {
+			// Intentionally not deferring the Unlock
+			// if this function returns, I don't want to let the timeout
+			// code run; it should pick up on the channel closing
+			// this does mean, however, that any continue statements MUST call
+			// refreshMutex.Unlock() before continue
```
joelthompson: Agree that it leaves the function easy to introducing new bugs, but this seemed like the one that was least likely to introduce race conditions (as I'm admittedly not terribly familiar with go concurrency and synchronization primitives, so I'm open to better ways of doing this). Basically, the two guarantees I'm attempting to write are:

1. Once the timeout has expired, `conf.Refresh` **cannot** be called again
1. The timeout error **cannot** be returned while `conf.Refresh` is still running, unless the grace period has expired.

This is also tricky because the returning inside the for loop is what triggers the closing of `doneCh` which is what allows the return value to be processed. Now I want to have some more complex behavior inside the for loop -- in some code paths, I want to close `doneCh` and in others I want to unlock `refreshMutex`. What I'm not sure of is the guarantees of go around if I close `doneCh` and unlock `refreshMutex` then what order will the `select` statements get processed?

Anyway, because I want either `conf.Refresh()` to be called or the timeout code to process, it seems like a mutex is the right primitive.

helper/resource/state.go
```diff
@@ -128,6 +138,11 @@ func (conf *StateChangeConf) WaitForState() (interface{}, error) {
 						if conf.ContinuousTargetOccurence == targetOccurence {
 							return
 						} else {
+							// FIXME: I think this continue is buggy, it's continuing the for
```
joelthompson: Yes, I'm aware of what the `continue` does. However, that's not my continue, that continue existed in the code prior to my changes. I'm simply flagging that I think there's a small bug that pre-existed in the code, and that if you wanted to fix it, you'd need to call `refreshMutex.Unlock()` (which goes back to what you're saying about this being somewhat easy to introduce bugs into, and I'm trying to guard against that).

helper/resource/state.go
joelthompson: Sorry, I don't really understand your comment.

helper/resource/state_test.go
```diff
@@ -19,6 +19,13 @@ func TimeoutStateRefreshFunc() StateRefreshFunc {
 	}
 }
 
+func SlowSuccessfulStateRefreshFunc(d time.Duration) StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		time.Sleep(d)
```
joelthompson: Makes sense, though I'm just copying what's in the other test...

helper/resource/state.go
jbardin: This was somewhat pedantic, and isn't really relevant now that I see your overall goal here. 

Just to explain, technically there's no guarantee that the first goroutine is scheduled to run before this one, therefore this `refreshMutex.Lock` could succeed first short circuiting the grace period. It's not really relevant because  things are already broken if we can't run the first goroutine before the timeout, it was just a correctness issue that has bit us in other tighter scheduling cases.

joelthompson: Ah, makes sense, good to know. Thanks for the explanation! :)

jbardin:
@joelthompson,

Thanks for the updated breakdown of the actual behavior, that helps a lot (since I haven't been able to reproduce the issue)

My next question is -- what makes this significantly different than extending the timeout another 5s? Would this possibly happen again at t == 20s, assuming a sufficiently slow api call?

joelthompson:
Yeah, I'm assuming that our dev environment is exacerbating this issue, making it hard to repro in isolation. We're hitting our environment pretty heavily, which is causing lots of Client.RequestLimitExceeded errors and we're also observing slow eventual consistency with IAM instance profiles, and I think it's the combination of these two that is causing us to hit this issue.

In terms of your question, there are a few different scenarios here, which is why I'm calling this a "mitigation" rather than completely solving the issue. If `ec2.RunInstances` gets called at t = 14.9s and takes, say, 6 seconds to complete, and succeeds, then a resource will still get leaked even with my PR. So, in this particular case, it's no different than simply extending the timeout to 20s. What it does do is give a few seconds for `ec2.RunInstances` to return.

Maybe to explain another way:

* In the current code, `ec2.RunInstances` must both start and complete before t = 15s to have its return value recognized
* If we simply extended the timeout to 20s, then `ec2.RunInstances` must both start and complete before t = 20s to have its return value recognized
* In my PR, `ec2.RunInstances` must start before t = 15s and finish before t = 20s to have its return value recognized

In the first two scenarios, `ec2.RunInstances` could be called arbitrarily close to the timeout value (e.g., t = 14.999s) and have no chance of completing before the timeout is reached. In my PR, we create a buffer.

The grace period of 5s for `aws_instance` was chosen somewhat arbitrarily, but it seems to work in our environment.

Choosing the grace period is really a balancing act. Ideally, I'd like to know if `conf.Refresh()` will ever actually halt and in what timeframe and act based upon that, but of course that's not practical here, so we have to make a tradeoff. If we choose a grace period that's too long, then we could just be wasting a ton of time. If we choose it too short, then we'll risk orphaning resources. 5s seemed to be a good, practical balance in our environment.

Does this make a little more sense?

jbardin:
Thanks,  that completely clears up the issue for me. 

Yes, there's a more correct way we could cancel the function here rather than use a mutex. The goroutine can check for cancellation immediately before either of its 2 blocking operations; the Refresh and the Sleep, that way the `wait` value isn't hampering the timeout as well. 

I ran a quick proof of concept, and doing that actually breaks a number of tests that expect the function to be aborted altogether, which is a good sign. We are also working on implementing more pervasive timeouts for providers, which will certainly affect this code path too.  I'll finish up a fix in another branch taking those into consideration.

Thanks so much for your help, even though we're not merging this, the PR was very useful!

PR #13743, opened 2017-04-18
provider/google: BigQuery Table

heimweh:
This PR adds support for BigQuery tables.

This PR is dependent on the resource and changes in #13436.

### DIFF
--- a/builtin/providers/google/resource_bigquery_table.go
+++ b/builtin/providers/google/resource_bigquery_table.go
@@ -10,7 +10,6 @@ import (
 	"github.com/hashicorp/terraform/helper/structure"
 	"github.com/hashicorp/terraform/helper/validation"
 	"google.golang.org/api/bigquery/v2"
-	"google.golang.org/api/googleapi"
 )
 
 func resourceBigQueryTable() *schema.Resource {
@@ -271,9 +270,8 @@ func resourceBigQueryTableCreate(d *schema.ResourceData, meta interface{}) error
 }
 
 func resourceBigQueryTableParseID(id string) (string, string, string) {
-	// projectID, datasetID, tableID
 	parts := strings.FieldsFunc(id, func(r rune) bool { return r == ':' || r == '.' })
-	return parts[0], parts[1], parts[2]
+	return parts[0], parts[1], parts[2] // projectID, datasetID, tableID
 }
 
 func resourceBigQueryTableRead(d *schema.ResourceData, meta interface{}) error {
@@ -285,15 +283,7 @@ func resourceBigQueryTableRead(d *schema.ResourceData, meta interface{}) error {
 
 	res, err := config.clientBigQuery.Tables.Get(projectID, datasetID, tableID).Do()
 	if err != nil {
-		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
-			log.Printf("[WARN] Removing BigQuery table %q because it's gone", tableID)
-			// The resource doesn't exist anymore
-			d.SetId("")
-
-			return nil
-		}
-
-		return err
+		return handleNotFoundError(err, d, fmt.Sprintf("BigQuery table %q", tableID))
 	}
 
 	d.Set("description", res.Description)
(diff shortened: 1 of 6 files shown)


builtin/providers/google/resource_bigquery_table.go
```diff
@@ -0,0 +1,406 @@
...
+	log.Printf("[INFO] BigQuery table %s has been created", res.Id)
+
+	d.SetId(res.Id)
+
+	return resourceBigQueryTableRead(d, meta)
+}
+
+func resourceBigQueryTableParseID(id string) (string, string, string) {
+	// projectID, datasetID, tableID
+	parts := strings.FieldsFunc(id, func(r rune) bool { return r == ':' || r == '.' })
```
danawillow: Can you point me to where this is documented? All I'm seeing for id is `An opaque ID uniquely identifying the table` which makes me think it doesn't have a guaranteed format.

website/source/docs/providers/google/r/bigquery_table.html.markdown
```diff
@@ -0,0 +1,104 @@
...
+    partitioning for this table. Structure is documented below.
+
+The `time_partitioning` block supports:
+
+* `expiration_ms` -  (Optional) Number of milliseconds for which to keep the
+    storage for a partition.
+
+* `type` - (Required) The only type supported is DAY, which will generate
+    one partition per day based on data loading time.
+## Attributes Reference
```
danawillow: nit: newline

builtin/providers/google/resource_bigquery_table.go
```diff
@@ -0,0 +1,412 @@
...
+			// key.
+			"labels": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem:     schema.TypeString,
+			},
+
+			// Schema: [Optional] Describes the schema of this table.
+			"schema": {
+				Type:         schema.TypeString,
```
heimweh: I left this as a JSON string, because I wasn't sure what would be the preferred way.
I'd be happy to replace this and add support for defining the schema in Terraform directly instead.

I'll take a look and see if I can get that to work :)

heimweh: Hi @danawillow, sorry for the delay here. 

I started working on this but since `TableSchema` (https://godoc.org/google.golang.org/api/bigquery/v2#TableFieldSchema) can have an arbitrary depth of record fields I'm not entirely sure I can get this to work with `helper/schema`.

With the current JSON string implementation, one could either define the schema inline or for more complex schemas use for example the `$(file())` interpolation function.

I'm not quite sure how to proceed here, so if you have any ideas please let me know :)

builtin/providers/google/resource_bigquery_table.go
heimweh: Good catch! 👍 I must have missed that part of the documentation. 
While it seems to work, I totally agree that it doesn't seem to be of a guaranteed format. 
I'll push a fix for this.

danawillow: Have you gotten to this yet? It _probably_ works as-is so I'm not too concerned, but it would be nice to have some confidence.
(a pretty easy fix would be to store the expected id as the id instead of the returned one)

builtin/providers/google/resource_bigquery_table.go
danawillow: Ah, I didn't realize the arbitrary nesting earlier. I think this is the best way, then. Thanks for the clarification!

builtin/providers/google/resource_bigquery_table.go
heimweh: @danawillow thanks for the quick feedback :)
I totally agree, just pushed a fix that should set the expected ID and format.

danawillow:
Thanks, LGTM! Merging now so this gets in before 0.9.5 :)

heimweh:
Awesome! :) thanks @danawillow!

PR #13772, opened 2017-04-19
Fixing TestAccAWSInstance_multipleRegions

mbfrahry:
This PR fixes `TestAccAWSInstance_multipleRegions`

```
make testacc TEST=./builtin/providers/aws/ TESTARGS='-run=TestAccAWSInstance_multipleRegions'
2017/04/19 09:12:43 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws/ -v -run=TestAccAWSInstance_multipleRegions -timeout 120m
=== RUN   TestAccAWSInstance_multipleRegions
--- PASS: TestAccAWSInstance_multipleRegions (122.17s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	122.215s
```

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -367,7 +367,10 @@ func testProviderFactories(c TestCase) (map[string]terraform.ResourceProviderFac
 	for k, pf := range ctxProviders {
 		// we can ignore any errors here, if we don't have a provider to reset
 		// the error will be handled later
-		p, _ := pf()
+		p, err := pf()
+		if err != nil {
+			return nil, err
+		}
 		if p, ok := p.(TestProvider); ok {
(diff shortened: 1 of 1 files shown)


jbardin:
@mbfrahry, this took me way too long to verify, but I think what you did here actually will be fine for the time being. 

Fixing the factory functions was preemptive on my part for future provider tests that use a factory, need to be reset,  _and_ not reconfigured during the tests. As of now, there aren't any. 

I think since tests that use resource.TestCase need to take into account that a ProviderFactory maybe called multiple times, leaving the responsibility of maintaining required state with the test is acceptable.

mbfrahry:
Can do!

PR #13787, opened 2017-04-19
Add DigitalOcean datasource digitalocean_image

roidelapluie:
Add a new data source for Digital Ocean that finds snapshots by name.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

Current status: Test panics.

TODO:
- [x] test when 2 snapshots have the same name
- [x] remove snapshots in acceptance tests
- [x] test when no snapshot matches the query
- [x] Add real world example in documentation (data digitalocean_image used in digitalocean_droplet)

### DIFF
--- a/builtin/providers/digitalocean/datasource_digitaloceal_image.go
+++ b/builtin/providers/digitalocean/datasource_digitaloceal_image.go
@@ -1,0 +1,93 @@
+package digitalocean
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceDigitalOceanImage() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDigitalOceanImageRead,
+		Schema: map[string]*schema.Schema{
+
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "name of the image",
+			},
+			// computed attributes
+			"image": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "slug or id of the image",
+			},
+			"min_disk_size": &schema.Schema{
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "minimum disk size required by the image",
+			},
+			"private": &schema.Schema{
+				Type:        schema.TypeBool,
+				Computed:    true,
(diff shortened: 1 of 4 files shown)


roidelapluie:
Current status: snapshots and droplets are not destroyed :(

roidelapluie:
current status: working, ready for review

stack72:
Hi @roidelapluie 

This LGTM! The acceptance tests pass too :) Nice to see a DO datasource :)

```
% make testacc TEST=./builtin/providers/digitalocean TESTARGS='-run=TestAccDigitalOceanImage_'                       ✭
2017/04/21 16:16:24 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/digitalocean -v -run=TestAccDigitalOceanImage_ -timeout 120m
=== RUN   TestAccDigitalOceanImage_Basic
--- PASS: TestAccDigitalOceanImage_Basic (294.89s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/digitalocean	294.913s
```

Paul

roidelapluie:
Great. Maybe the last test step could be changed by a "Cleanup" attribute at the TestCase level. This is because Terraform is not creating the snapshots in tf code, so I create them in go.

lfarnell:
@stack72 @roidelapluie After looking at the changes, the go files for this change are spelt wrong. Currently datasource_digitaloceal_image.go should be datasource_digitalocean_image.go. Sorry for commenting on a  merged PR

andrewsomething:
Just noticed this landed. I had attempted something similar in #12873 but got blocked on tests. I didn't realize this approach was an option.

The main difference between this and what  I implemented is that I used the snapshots endpoint:

https://developers.digitalocean.com/documentation/v2/#snapshots

The main advantage to that is it also allows for finding volume snapshots.

roidelapluie:
Yeah, I ended up doing something that works for me -- This datasource
can be later extended to fetch distributions etc as well.

Maybe we can have a data_digitalocean_snapshot in the future too. What
was the problem with tests? Time?

-- 
 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

andrewsomething:
In retrospect, it was a simple misunderstanding on my part. I thought it would require writing a resource to _take snapshots_ first, which didn't seem particularly useful as opposed to creating and deleting the resources directly via calls to godo.

Issue #13606, opened 2017-04-12
aws_subnet: diffs didn't match during apply.


BlueDragonX:
Hello. I'm receiving the error `Error applying plan ... diffs didn't match during apply` when attempting to execute changes for some `aws_subnet` resources.

`terraform plan` outputs:
```
~ aws_subnet.proxies.0
    tags.Name:    "ZD-UW2A-NET-FleetPublic" => "ZD-UW2A-NET-Proxies"
    tags.Network: "fleet-public" => "proxies"

~ aws_subnet.services.0
    tags.Name:    "ZD-UW2A-NET-FleetPrivate" => "ZD-UW2A-NET-Services"
    tags.Network: "fleet-private" => "services"
```

`terraform apply` outputs:
```
Error applying plan:

2 error(s) occurred:

* aws_subnet.services[0]: aws_subnet.services.0: diffs didn't match during apply. This is a bug with Terraform and should be reported as a GitHub Issue.

Please include the following information in your report:

    Terraform Version: 0.9.2
    Resource ID: aws_subnet.services.0
    Mismatch reason: diff RequiresNew; old: false, new: true
    Diff One (usually from plan): *terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff{"tags.Name":*terraform.ResourceAttrDiff{Old:"ZD-UW2A-NET-FleetPrivate", New:"ZD-UW2A-NET-Services", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}, "tags.Network":*terraform.ResourceAttrDiff{Old:"fleet-private", New:"services", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}}, Destroy:false, DestroyDeposed:false, DestroyTainted:false, Meta:map[string]interface {}(nil)}
[...]

PR #13793, opened 2017-04-20
core: Add CountBoundaryTransformer to the plan graph builder

vancluever:
This fixes interpolation issues on grandchild data sources that have
multiple instances (ie: counts). For example, baz depends on bar, which
depends on foo.

In this instance, after an initial TF run is done and state is saved,
the next refresh/plan is not properly transformed, and instead of the
graph/state coming through as data.x.bar.0, it comes through as
data.x.bar.  This breaks interpolations that rely on splat operators -
ie: data.x.bar.*.out.

### DIFF
--- a/builtin/providers/test/data_source_test.go
+++ b/builtin/providers/test/data_source_test.go
@@ -99,3 +99,59 @@ resource "test_resource" "foo" {
 		},
 	})
 }
+
+// Test that a grandchild data source that is based off of count works, ie:
+// dependency chain foo -> bar -> baz. This was failing because
+// CountBoundaryTransformer is being run during apply instead of plan, which
+// meant that it wasn't firing after data sources were potentially changing
+// state and causing diff/interpolation issues.
+//
+// This happens after the initial apply, after state is saved.
+func TestDataSource_dataSourceCountGrandChild(t *testing.T) {
+	resource.UnitTest(t, resource.TestCase{
+		Providers: testAccProviders,
+		CheckDestroy: func(s *terraform.State) error {
+			return nil
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: dataSourceCountGrandChildConfig,
+			},
+			{
+				Config: dataSourceCountGrandChildConfig,
+				Check: func(s *terraform.State) error {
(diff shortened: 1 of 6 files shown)


vancluever:
WIPing this until I figure out what is up with the tests. Guessing that moving the transformer isn't the whole story.

Duplicating the transformer reduces the amount of tests that fail in `./terraform` to `TestPlanGraphBuilder` - will look into that more.

vancluever:
OK, I think it's still important to have the transformer in the apply builder as well as the plan builder. At least one test (`TestContext2Apply_countDecreaseToOneX`) is legitimately testing a normalization scenario where a singular `aws_instance.foo.0` instance needs to be normalized to `aws_instance.foo`. 

`TestPlanGraphBuilder` I think is failing because the expected shape of the data has now changed, as the graph now contains the `meta.count-boundary` nodes due to the work the transformer does on the graph. Will add that.

Trying to also get a good test outside of the test resource unit test.

vancluever:
Alright, `TestPlanGraphBuilder` has been adjusted, the transformer has been restored to the apply graph and all tests pass now.

I've also added another context test (`TestContext2Plan_resourceNestedCount`) that covers this case. This was tricky, because it looks like that validate is the thing doing something to the state that causes it to get into this inconsistent state, not refresh like I was thinking. It should be noted I did try to add the transformer to the refresh walk too when I was trying to think about where best to put it and it didn't seem to solve the problem.

vancluever:
Thanks @jbardin! I'm still chasing down #13828 as well but I think that's not necessarily related to resource addressing (and more how the graph is being built during a refresh in general). I think this one is good to go and I can send other work for that in a separate PR.

jbardin:
Interestingly, this also has the side effect of fixing #13606.

That diff mismatch isn't caused specifically by a count boundary, but the mismatch _only_ happens on the `[0]` resource, and only if the count is interpolated from a datasource.

jbardin:
Now that this isn't going in the day before a large-ish release, I'm going to merge it so we can run it through its paces.

vancluever:
Right on! Thanks again @jbardin and glad it helped! 👍

Issue #13416, opened 2017-04-06
google_dns_managed_zone doesn't support import


hpio:
Feature request to support import on the google_dns_managed_zone resource 😄

**Terraform Version**

Terraform v0.9.2

**Affected Resource(s)**

google_dns_managed_zone

**Expected Behavior**

Should import an existing DNS zone.

**Actual Behavior**

Generates an error: resource google_dns_managed_zone doesn't support import

**Steps to Reproduce**

run terraform import resource google_dns_managed_zone.foo

hpio:
Happy to try and pick this up if someone can point me where to start?

paddyforan:
I'd love to see a PR for this, if you're willing to give it a shot! #13270 just added import for a few other resources, so it may be a good thing to model your PR after.

If not, we'll get to it, it just hasn't hit the top of our priority list yet.

paddycarver:
@JDiPierro did some great work in adding this in #13824, which just got merged, so I'm going to close this out now. Thanks!

PR #13824, opened 2017-04-20
Importability for Google DNS Managed Zone

JDiPierro:
Requested in #13416.

I removed the default description for the DNS managed zone because when I imported a manually created zone with no description it wanted to recreate it with the default terraform description. Since Description is an optional field I think it's appropriate to make it empty by default.

```
$ make test TEST=./builtin/providers/google/
2017/04/20 17:10:57 Generated command/internal_plugin_list.go
go test -i ./builtin/providers/google/ || exit 1
echo ./builtin/providers/google/ | \
		xargs -t -n4 go test  -timeout=60s -parallel=4
go test -timeout=60s -parallel=4 ./builtin/providers/google/
ok  	github.com/hashicorp/terraform/builtin/providers/google	0.027s
```

### DIFF
--- a/builtin/providers/google/resource_dns_managed_zone.go
+++ b/builtin/providers/google/resource_dns_managed_zone.go
@@ -14,7 +14,9 @@ func resourceDnsManagedZone() *schema.Resource {
 		Create: resourceDnsManagedZoneCreate,
 		Read:   resourceDnsManagedZoneRead,
 		Delete: resourceDnsManagedZoneDelete,
-
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
 		Schema: map[string]*schema.Schema{
 			"dns_name": &schema.Schema{
 				Type:     schema.TypeString,
@@ -109,6 +111,9 @@ func resourceDnsManagedZoneRead(d *schema.ResourceData, meta interface{}) error 
 	}
 
 	d.Set("name_servers", zone.NameServers)
+	d.Set("name", zone.Name)
+	d.Set("dns_name", zone.DnsName)
+	d.Set("description", zone.Description)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


builtin/providers/google/resource_dns_managed_zone.go
```diff
@@ -32,7 +34,6 @@ func resourceDnsManagedZone() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				Default:  "Managed by Terraform",
```
paddycarver: My only concern about this is that removing the default would lead Terraform to think every managed zone changed, and try to destroy then re-create it. I'm not sure what the best way to get rid of a default value is (or support imports when dealing with a default value) without causing too much backwards incompatibility. I'll check with the team and see if anyone has advice and report back.

paddycarver: So I think if you leave the Default as it is and set this to `Computed: true`, things _should_ be OK and not cause problems. What do you think?

JDiPierro: Sounds great, I'll get that in.

JDiPierro: Eek! 
`resource google_dns_managed_zone: description: Default must be nil if computed`
I'll leave the default as-is. Users will have to manually provide an empty description on imported managed zones.

JDiPierro:
Re-added the default description.

```$ make test TEST=./builtin/providers/google/
2017/05/01 20:27:58 Generated command/internal_plugin_list.go
go test -i ./builtin/providers/google/ || exit 1
echo ./builtin/providers/google/ | \
		xargs -t -n4 go test  -timeout=60s -parallel=4
go test -timeout=60s -parallel=4 ./builtin/providers/google/
ok  	github.com/hashicorp/terraform/builtin/providers/google	0.037s```

Issue #13816, opened 2017-04-20
Feature request: Export arn as atrtribute for resource aws_cloudfront_distribution


fillup:
Currently the `aws_cloudfront_distribution` resource exports the distribution `id` in format of `EDFDVBD632BHDS5`, but it would be really helpful to have the full ARN in format `arn:aws:cloudfront::1234567890:distribution/EDFDVBD632BHDS5` for use in other resources like `aws_iam_user_policy` to specify the single resource. 

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_cloudfront_distribution

### Example:
I'd like to be able to use this like:

```hcl
resource "aws_cloudfront_distribution" "ui" {
 //...
}
resource "aws_iam_user_policy" "ci_cf" {
  name = "CF"
  user = "${var.username}"

  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "cloudfront:CreateInvalidation"
      ],
      "Resource": [
        "${aws_cloudfront_distribution.ui.arn}"
      ]
    }
  ]
}
EOF
}
```

Ninir:
Hi @fillup 

Thanks for reporting this. Even if this is not part of the documentation yet, the ARN is exported, so you should be able to use `aws_cloudfront_distribution.ui.arn`.

Could you try and tell us if it's ok?

Will open a PR for that.

Thanks!

PR #13874, opened 2017-04-23
provider/aws: Updated documentation for Cloudfront

Ninir:
This adds the missing exported attribute named `ARN`, as exposed [in the code](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_cloudfront_distribution.go#L592).

Closes https://github.com/hashicorp/terraform/issues/13816


stack72:
LGTM Thanks for this @Ninir :)

PR #13887, opened 2017-04-23
provider/github: Clears branch protection from state only if error is an HTTP 404

alindeman:
It was a bit aggressive to clear an item from state due to _any_ error. In reality, we should only clear things from state and hide the error if a 404 HTTP status is returned.


stack72:
You rock @alindeman :)

PR #13898, opened 2017-04-24
provider/gitlab: add a gitlab provider

richardc:
This adds a gitlab provider and `gitlab_project` resource.  It's copied heavily from the `github` provider and `github_repository`  resource, as there is some overlap in those types/apis.

Sample usage:

```
  name = "test1"
  visibility_level = "public"
}
```

Its implemented in terms of the [go-gitlab](https://github.com/xanzy/go-gitlab) library, which provides a wrapping of the [gitlab api](https://docs.gitlab.com/ee/api/)

We have been a little selective in the properties we surface for the project resource, as not all properties are very instructive.
Notable is the removal of the `public` bool as the `visibility_level`  will take precedent if both are supplied which leads to confusing interactions if they disagree.

### DIFF
--- a/builtin/providers/gitlab/provider.go
+++ b/builtin/providers/gitlab/provider.go
@@ -1,0 +1,52 @@
+package gitlab
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+// Provider returns a terraform.ResourceProvider.
+func Provider() terraform.ResourceProvider {
+
+	// The actual provider
+	return &schema.Provider{
+		Schema: map[string]*schema.Schema{
+			"token": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_TOKEN", nil),
+				Description: descriptions["token"],
+			},
+			"base_url": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_BASE_URL", ""),
+				Description: descriptions["base_url"],
+			},
+		},
+		ResourcesMap: map[string]*schema.Resource{
+			"gitlab_project": resourceGitlabProject(),
+		},
+
+		ConfigureFunc: providerConfigure,
+	}
+}
+
+var descriptions map[string]string
+
+func init() {
+	descriptions = map[string]string{
+		"token": "The OAuth token used to connect to GitLab.",
+
+		"base_url": "The GitLab Base API URL",
+	}
+}
+
+func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	config := Config{
+		Token:   d.Get("token").(string),
+		BaseURL: d.Get("base_url").(string),
+	}
+
+	return config.Client()
+}
--- a/builtin/providers/gitlab/resource_gitlab_project.go
+++ b/builtin/providers/gitlab/resource_gitlab_project.go
@@ -1,0 +1,207 @@
+package gitlab
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabProject() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabProjectCreate,
+		Read:   resourceGitlabProjectRead,
+		Update: resourceGitlabProjectUpdate,
+		Delete: resourceGitlabProjectDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"default_branch": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"issues_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"merge_requests_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"wiki_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"snippets_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"visibility_level": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateValueFunc([]string{"private", "internal", "public"}),
+				Default:      "private",
+			},
+
+			"ssh_url_to_repo": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"http_url_to_repo": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"web_url": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGitlabProjectUpdateFromAPI(d *schema.ResourceData, project *gitlab.Project) {
+	d.Set("name", project.Name)
+	d.Set("description", project.Description)
+	d.Set("default_branch", project.DefaultBranch)
+	d.Set("issues_enabled", project.IssuesEnabled)
+	d.Set("merge_requests_enabled", project.MergeRequestsEnabled)
+	d.Set("wiki_enabled", project.WikiEnabled)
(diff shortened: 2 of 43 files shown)


richardc:
This was previously proposed in PR #9591 which I'd stalled out on.

stack72:
Hi @richardc 

What version of Gitlab do we need to be able to run the tests for this? Will the community edition suffice?

Thanks

Paul

richardc:
The community edition is what I developed against using a local [gitlab-ce](https://hub.docker.com/r/gitlab/gitlab-ce/).  I built off latest about a week ago so that gave me a snapshot that reports itself as version 9.0.5.  

I've also run the tests using the free tier of hosting at gitlab.com, which is currently reporting being version 9.1.0.

builtin/providers/gitlab/provider.go
```diff
@@ -0,0 +1,52 @@
...
+			"token": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_TOKEN", nil),
+				Description: descriptions["token"],
+			},
+			"base_url": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITLAB_BASE_URL", ""),
```
stack72: should we not have a default here? In github provider, we default to public github

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,207 @@
...
+	client := meta.(*gitlab.Client)
+	options := &gitlab.CreateProjectOptions{
+		Name: gitlab.String(d.Get("name").(string)),
+	}
+
+	if v, ok := d.GetOk("description"); ok {
+		options.Description = gitlab.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("issues_enabled"); ok {
```
stack72: as there is a default for bools like this, we can add these straight to the `CreateProjectOptions` struct

There is a known bug with d.GetOk and bool values like `false`  and ints like 0, so defaults mean it will always be set and we don't need to use d.GetOk :)

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	log.Printf("[DEBUG] create gitlab project %q", options.Name)
+
+	project, _, err := client.Projects.CreateProject(options)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%d", project.ID))
```
stack72: why are you needing to use fmt.Sprintf? d.SetId takes a string and you are formatting as as int

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,207 @@
...
+			},
+			"web_url": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGitlabProjectUpdateFromAPI(d *schema.ResourceData, project *gitlab.Project) {
```
stack72: I would rename this func to something like resourceGitlabProjectSetToState

It doesn't interact with the API at all and just sets to state

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,207 @@
...
+	log.Printf("[DEBUG] create gitlab project %q", options.Name)
+
+	project, _, err := client.Projects.CreateProject(options)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%d", project.ID))
+
+	resourceGitlabProjectUpdateFromAPI(d, project)
```
stack72: in this case, Create should call the Read func - that way we know that the project has been created

Right now, we are just setting the values the user specified back to state

Therefore, this should be `return resourceGitlabProjectRead(d, meta)`

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	log.Printf("[DEBUG] update gitlab project %s", d.Id())
+
+	project, _, err := client.Projects.EditProject(d.Id(), options)
+	if err != nil {
+		return err
+	}
+
+	resourceGitlabProjectUpdateFromAPI(d, project)
```
stack72: Updates should also call Read to make sure that the API has persisted the changes

so again, `return resourceGitlabProjectRead(d, meta)`

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return nil
+}
+
+func resourceGitlabProjectDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	log.Printf("[DEBUG] update gitlab project %s", d.Id())
```
stack72: typo in comment

stack72:
Hi @richardc 

I am going to go ahead and make these small changes - you have worked really hard on this and it would be unfair for me to ask you to make these small changes :)

I will ping you when i merge it

Thanks

Paul

builtin/providers/gitlab/resource_gitlab_project.go
richardc: project.ID is an int.

builtin/providers/gitlab/resource_gitlab_project.go
richardc: Not according to my reading of the gitlab code.  POST /projects returns a json document that is the state of the project once it has been created.  It is the same project info, so by making this tail call Read you're just creating extra unecessary api hits.
~~~
> POST /projects {:spec of project}
< project state

Then you propose an extra
> GET /projects/:id
< project state
~~~

stack72:
Hi @richardc 

I made the changes here - they took me like 10 minutes to add :) Your work is *very* much appreciated! It will be great to get this as part of Terraform 0.9.5 :)

Thanks

Paul

builtin/providers/gitlab/provider.go
richardc: Can do, and it will simplify the code flow in this path: https://github.com/richardc/terraform/blob/feature/gitlab_provider/builtin/providers/gitlab/config.go#L16

builtin/providers/gitlab/resource_gitlab_project.go
richardc: Much better name 👍

builtin/providers/gitlab/resource_gitlab_project.go
richardc: It's already doing that.   You're adding an extra api round trip, which isn't needed.

richardc:
Uh, I disagree with the restructuring to tail-call Read;  it's an extra set of api calls per operation that aren't really needed, but if it's house style then go for it.

builtin/providers/gitlab/resource_gitlab_project.go
richardc: Ouch

stack72:
@ricardclau I am looking at your response now and will update it - as i said above, you had this all working as expected, there were just a few tiny nits - but nothing crazy

I'm looking at the docs for the Create and the Update now to make sure all works as expected

P.

richardc:
OK great.  I'm excited to see this merged

stack72:
works a dream sir :) just took it for a test spin

jugatsu:
@richardc great work :+1:

hauleth:
Will it be possible to create project in group, not in user scope?

richardc:
I'll caveat my response with a mild disclaimer:  I've not looked at doing this in depth, as it wasn't my initial use case.

The API wrapped defaults to operating on the currently authenticated user, as determined by the API key configured for the provider.   If it's possible to issue an API key to a group then it should just  shake out already, otherwise extra work will be needed to figure out how the API models this, and to make the code changes.  If the changes are needed to make it work I'd consider it new functionality, so please raise a ticket and mention me in it.

https://github.com/xanzy/go-gitlab/blob/v0.5.1/projects.go#L360 https://docs.gitlab.com/ce/api/projects.html#create-project

hauleth:
I believe that `namespace_id` value determines the correct namespace needed. API for namespaces: https://docs.gitlab.com/ee/api/namespaces.html#namespaces

hauleth:
What is more it would be unbelievably more useful if we would have possibility to:

- [set build variables](https://docs.gitlab.com/ce/api/build_variables.html#create-variable) like bucket names based on other resources
- [manage labels](https://docs.gitlab.com/ce/api/labels.html#create-a-new-label) just like in GitHub provider
- [setup deploy keys](https://docs.gitlab.com/ce/api/labels.html#create-a-new-label)
- [manage boards](https://docs.gitlab.com/ce/api/boards.html#new-board-list)
- [setting project hooks](https://docs.gitlab.com/ce/api/projects.html#hooks)

I know that this is a little too much to ask someone to do work like this, but unfortunately I am not gophery enough to create such resources.

richardc:
I've got you covered for setting project hooks already #14155

I agree the rest could be useful, but you're commenting on a merged pull-request, which is why I suggest you file enhancement requests for anything else.

PR #13908, opened 2017-04-24
Rancher V2 provider updates

cloudnautique:
This PR upgrades the Rancher builtin provider to use the V2 API by default. Rancher has moved away from the client currently being used in the provider. 

Using the V2 API, the PR also fixes the `rancher_environment` resource that is broken for versions of Rancher after v1.2.x. 

This is a breaking change for versions of Rancher below 1.2.x, but Rancher is dropping support for that soon.

I tried to split the commits so that there is the Vendor updates and the code updates to make reviewing easier. 

Please let me know what/if anything needs changing.


mcanevet:
It's great to see RancherLab's team to contribute to this provider.
Thanks

onlydole:
@tombuildsstuff - Is there an ETA or verdict on when this will be merged in?

sebradloff:
Just adding a comment to echo @onlydole

builtin/providers/rancher/config.go
```diff
@@ -3,7 +3,7 @@ package rancher
 import (
 	"log"
 
-	rancherClient "github.com/rancher/go-rancher/client"
+	rancherClient "github.com/rancher/go-rancher/v2"
 	"github.com/raphink/go-rancher/catalog"
```
raphink: This might not be necessary anymore with the v2 API.

builtin/providers/rancher/provider.go
```diff
@@ -71,7 +73,7 @@ func init() {
 
 		"secret_key": "API secret used to authenticate with the rancher server",
 
-		"api_url": "The URL to the rancher API",
+		"api_url": "The URL to the rancher API, must include version uri (ie. v1 or v2-beta)",
```
raphink: I'm not sure to see the point in breaking compatibility here if you're going to use the v2 API everywhere anyway.

raphink:
Switching to the v2 API should also allow to fix the scope (user/system) parameter in `rancher_stack` resources.

builtin/providers/rancher/config.go
cloudnautique: Which part? The `rancherClient` part? I was going for the smallest change possible. And changing every reference didn't seem necessary.

builtin/providers/rancher/provider.go
cloudnautique: I don't mind appending it in the backend end for compatibility, but I think the correct setting is to include the version string. I know we have the non-beta version of V2 is over the horizon.  In general we include the API version string in both Rancher Compose and Rancher CLI sort of the convention we go for.

builtin/providers/rancher/config.go
raphink: The part that uses my fork of the lib for catalog templates. We could change that later though.

builtin/providers/rancher/provider.go
raphink: I suppose the new code won't work with v1 though, and the CATTLE_* variables still return the v1 URL.

cloudnautique: Discussed internally, we are going to go back to the original behavior and let the go-rancher client figure it out. The go-rancher client can take a v1, v2 or / url and do the right thing. I'll make it so the other client types in the provider build the URLs off of it.

raphink: Sounds good to me.

cloudnautique:
@raphink, any thoughts here?
I updated this to handle system scope.
Removed the need need for the URLs to need any versioning.
Moved to the the go-rancher/catalog

In addition to general migration to v2-beta api and support for HA K8s envs.

Ideally, this gets merged so that new stuff is built on the newer API.

@tombuildsstuff, what can I do to help shepherd this along?

raphink:
LGTM. Looking forward to rebasing my PRs :wink:

raphink:
@stack72 @grubernaut could you review this please?

SerialVelocity:
@raphink @stack72 @grubernaut Is it possible to have a look at this? Rancher hasn't worked that well in Terraform for a _really_ long time.

tombuildsstuff:
Hi @cloudnautique @mcanevet @onlydole @raphink @SerialVelocity 

Thanks for this PR - apologies for the delay in reviewing this!

In order to move forward with this PR we need to get the acceptance tests for the (current) Rancher provider hooked up to our CI server. I've started looking into this today, and I'll post an update once that's done so that we can proceed with reviewing this PR :)

Thanks!

SerialVelocity:
Things I've noticed so far:
  - Test case: Try and apply the same rancher_stack twice with `docker_compose` and `rancher_compose` set
    Current behaviour: `docker_compose` and `rancher_compose` are always listed as different
    Old behaviour: Same as current
    Expected behaviour: No changes should be necessary after applying once
  - Test case: Set scope to `system` and apply the rancher_stack twice
    Current behaviour: `scope` is set correctly but TF always thinks the scope is `user`.
    Old behaviour: Setting `scope` wouldn't actually change the scope in rancher.
    Expected behaviour: Sets `scope` correctly in rancher and says no changes on second application
  - Not possible to create a mount from TF. (feature, not a bug)

Everything I've noticed is strictly better than the last version but still buggy.

raphink:
Is there anything that can be done to help speed up this?

tombuildsstuff:
👋 hey @cloudnautique @raphink 

We run the acceptance tests for most providers nightly by either spinning up some infrastructure in a cloud provider on demand, or by launching a docker container and then running tests against it.

I've been working through trying to get both the existing acceptance tests and the acceptance tests in this PR for the Rancher provider working using the Docker image [`rancher/server`](https://hub.docker.com/r/rancher/server/) - however I'm having issues with this. Is this Docker image sufficient to run the tests, or does an agent need to be connected too? In addition - is this the best approach to take, or would you recommend something else?

Thanks!

raphink:
@tombuildsstuff i sent à message to @stack72 yesterday evening at his request, explaining what I know about running the tests.

cloudnautique:
@tombuildsstuff, I'm not sure how to kick off the tests, but for everything but the stack deployments, you should be able to run against a standalone server. Is there a slack channel or something that we can discuss more real time?

raphink:
@cloudnautique just forwarded you an email about that.

PR #13914, opened 2017-04-24
provider/kubernetes: Add support for resource_quota

radeksimko:
### Test plan

```
make testacc TEST=./builtin/providers/kubernetes TESTARGS='-run=TestAccKubernetesResourceQuota_'
```
```
=== RUN   TestAccKubernetesResourceQuota_basic
--- PASS: TestAccKubernetesResourceQuota_basic (9.02s)
=== RUN   TestAccKubernetesResourceQuota_generatedName
--- PASS: TestAccKubernetesResourceQuota_generatedName (8.44s)
=== RUN   TestAccKubernetesResourceQuota_withScopes
--- PASS: TestAccKubernetesResourceQuota_withScopes (9.43s)
=== RUN   TestAccKubernetesResourceQuota_importBasic
--- PASS: TestAccKubernetesResourceQuota_importBasic (4.34s)
```

Related upstream bug: https://github.com/kubernetes/kubernetes/issues/44539

### DIFF
--- a/builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
+++ b/builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
@@ -1,0 +1,211 @@
+package kubernetes
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	pkgApi "k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/errors"
+	api "k8s.io/kubernetes/pkg/api/v1"
+	kubernetes "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_5"
+)
+
+func resourceKubernetesResourceQuota() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceKubernetesResourceQuotaCreate,
+		Read:   resourceKubernetesResourceQuotaRead,
+		Exists: resourceKubernetesResourceQuotaExists,
+		Update: resourceKubernetesResourceQuotaUpdate,
+		Delete: resourceKubernetesResourceQuotaDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"metadata": namespacedMetadataSchema("resource quota", true),
+			"spec": {
+				Type:        schema.TypeList,
+				Description: "Spec defines the desired quota. http://releases.k8s.io/HEAD/docs/devel/api-conventions.md#spec-and-status",
+				Optional:    true,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"hard": {
+							Type:         schema.TypeMap,
+							Description:  "The set of desired hard limits for each named resource. More info: http://releases.k8s.io/HEAD/docs/design/admission_control_resource_quota.md#admissioncontrol-plugin-resourcequota",
+							Optional:     true,
+							Elem:         schema.TypeString,
+							ValidateFunc: validateResourceList,
+						},
+						"scopes": {
+							Type:        schema.TypeSet,
+							Description: "A collection of filters that must match each object tracked by a quota. If not specified, the quota matches all objects.",
+							Optional:    true,
+							ForceNew:    true,
+							Elem:        &schema.Schema{Type: schema.TypeString},
+							Set:         schema.HashString,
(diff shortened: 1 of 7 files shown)


builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
```diff
@@ -0,0 +1,211 @@
...
+				},
+			},
+		},
+	}
+}
+
+func resourceKubernetesResourceQuotaCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+
+	metadata := expandMetadata(d.Get("metadata").([]interface{}))
```
stack72: Why does `expandMetadata` not return error status whereas `expandResourceQuoteSpec` does? Is there no change metadata could not return an error?

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	log.Printf("[INFO] Creating new resource quota: %#v", resQuota)
+	out, err := conn.CoreV1().ResourceQuotas(metadata.Namespace).Create(&resQuota)
+	if err != nil {
+		return fmt.Errorf("Failed to create resource quota: %s", err)
+	}
+	log.Printf("[INFO] Submitted new resource quota: %#v", out)
+	d.SetId(buildId(out.ObjectMeta))
+
+	err = resource.Retry(1*time.Minute, func() *resource.RetryError {
```
stack72: Is 1 minute enough here?

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func resourceKubernetesResourceQuotaRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
+
+	namespace, name := idParts(d.Id())
+	log.Printf("[INFO] Reading resource quota %s", name)
+	resQuota, err := conn.CoreV1().ResourceQuotas(namespace).Get(name)
+	if err != nil {
+		log.Printf("[DEBUG] Received error: %#v", err)
```
stack72: Can this be manually deleted? If so, can we catch it here?

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
```diff
@@ -0,0 +1,211 @@
...
+	err = d.Set("spec", flattenResourceQuotaSpec(resQuota.Spec))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func resourceKubernetesResourceQuotaUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*kubernetes.Clientset)
```
stack72: Even if it's only `spec` that changes here, we are going to update the metadata - correct?

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
radeksimko: That's merely because `expandResourceQuoteSpec` has to parse "resource list" and that may error out, `expandMetadata` doesn't have to parse anything like that.
https://github.com/hashicorp/terraform/pull/13914/files#diff-886d9a6a55a7569117aaf167990fed25R223

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
radeksimko: `ops := patchMetadata("metadata.0.", "/metadata/", d)` should return empty slice of `PatchOperations` if it has no changes and if we have reached the `Update()` function it's clear that we have _something_ to update - either `spec` or `metadata`, so it is there just to initialize the slice, really.

I could change it to something like

```go
ops := make([]PatchOperations, 0, 0)
if d.HasChange("metadata") {
    ops = append(ops, patchMetadata("metadata.0.", "/metadata/", d)
}
```
if you think it's better?

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
radeksimko: That's why there's `resourceKubernetesResourceQuotaExists` 😉

builtin/providers/kubernetes/resource_kubernetes_resource_quota.go
radeksimko: It was plenty of time based on my testing - it is usually ready within a few seconds.

stack72:
Your answers sound good :) LGTM!

PR #13926, opened 2017-04-25
Load Cloudstack configuration from file

mcanevet:
This PR adds 2 new parameters to the CloudStack provider allowing to fetch credentials from a configuration file compatible with https://github.com/exoscale/cs CloudStack CLI.

With this patch, one can use this manifest to configure the CloudStack provider:
```terraform
provider "cloudstack" {
  config = "${pathexpand("~/.cloudstack.ini")}"
  region = "my_region"
}
```

### DIFF
--- a/builtin/providers/cloudstack/provider.go
+++ b/builtin/providers/cloudstack/provider.go
@@ -1,6 +1,9 @@
 package cloudstack
 
 import (
+	"fmt"
+
+	"github.com/go-ini/ini"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
@@ -11,20 +14,30 @@ func Provider() terraform.ResourceProvider {
 		Schema: map[string]*schema.Schema{
 			"api_url": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_URL", nil),
 			},
 
 			"api_key": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_KEY", nil),
 			},
 
 			"secret_key": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_SECRET_KEY", nil),
+			},
+
+			"config": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"profile": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
 			},
 
 			"http_get_only": &schema.Schema{
@@ -72,10 +85,34 @@ func Provider() terraform.ResourceProvider {
 }
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	apiURL := d.Get("api_url").(string)
+	apiKey := d.Get("api_key").(string)
+	secretKey := d.Get("secret_key").(string)
+
+	if configFile, ok := d.GetOk("config"); ok {
+		config, err := ini.Load(configFile.(string))
+		if err != nil {
+			return nil, err
+		}
+
+		section, err := config.GetSection(d.Get("profile").(string))
+		if err != nil {
+			return nil, err
+		}
+
+		apiURL = section.Key("url").String()
+		apiKey = section.Key("apikey").String()
+		secretKey = section.Key("secretkey").String()
+	}
+
+	if apiURL == "" || apiKey == "" || secretKey == "" {
+		return nil, fmt.Errorf("No api_url or api_key or secretKey provided")
+	}
+
 	config := Config{
-		APIURL:      d.Get("api_url").(string),
-		APIKey:      d.Get("api_key").(string),
-		SecretKey:   d.Get("secret_key").(string),
+		APIURL:      apiURL,
+		APIKey:      apiKey,
+		SecretKey:   secretKey,
 		HTTPGETOnly: d.Get("http_get_only").(bool),
 		Timeout:     int64(d.Get("timeout").(int)),
 	}
--- a/builtin/providers/cloudstack/provider.go
+++ b/builtin/providers/cloudstack/provider.go
@@ -13,31 +13,36 @@ func Provider() terraform.ResourceProvider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
 			"api_url": &schema.Schema{
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_URL", nil),
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("CLOUDSTACK_API_URL", nil),
+				ConflictsWith: []string{"config", "profile"},
 			},
 
 			"api_key": &schema.Schema{
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_KEY", nil),
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("CLOUDSTACK_API_KEY", nil),
+				ConflictsWith: []string{"config", "profile"},
 			},
 
 			"secret_key": &schema.Schema{
-				Type:        schema.TypeString,
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_SECRET_KEY", nil),
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("CLOUDSTACK_SECRET_KEY", nil),
+				ConflictsWith: []string{"config", "profile"},
 			},
 
 			"config": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"api_url", "api_key", "secret_key"},
 			},
 
 			"profile": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"api_url", "api_key", "secret_key"},
 			},
 
 			"http_get_only": &schema.Schema{


mcanevet:
@svanharmelen could you please review this?

svanharmelen:
@mcanevet I must admit that I'm not in favour of adding this customization.

The configuration file you refer to is an Exoscale specific config file that is not part of Apache CloudStack. There are more cloud providers out there that add their own magic sauce or add tools to standard Apache CloudStack, but of course this single provider cannot serve all those use different flavours. So my goal is to keep this provider as close the Apache CloudStack as possible.

As an example you can think about CloudMonkey. CloudMonkey also uses an ini style config file, but  with a different layout (different fields). So now we already have 2 different formats. And I'm very sure there are more, but they are all customized. So why should we take one or the other? If anything I would argue that the CloudMonkey format is probably the most official one as that is hosted as the Apache CloudStack CloudMonkey project.

So I would very much stay away from adding something like this, sorry. Luckily there are already a couple of ways you can config the provider within TF itself (directly in the file, using cmdline vars, using a tfvars file or use env variables) so I hope one of those options can make it into your workflow instead.

mcanevet:
@svanharmelen sounds legit.

However, I'm not a fan of any other way of configuring the provider:
- hardcode credentials in the manifests or using tfvars file : we definitely don't want to commit credentials and each user has its own credentials,
- Using cmdline vars or environment variables : way too dangerous, if one uses the wrong command line arguments or load the wrong environment variables, he may deploy something on the wrong region.

For the moment we are fetching credentials from pass-password datasource (https://github.com/hashicorp/terraform/pull/11381) using that kind of declaration (we could also use vault provider the same way):
```terraform
data "pass_password" "cloudstack" {
  path = "foo/bar"
}

provider "cloudstack" {
  api_url    = "${data.pass_password.cloudstack.data["api_url"]}"
  api_key    = "${data.pass_password.cloudstack.data["api_key"]}"
  secret_key = "${data.pass_password.cloudstack.data["secret_key"]}"
}
```

Using that, we don't have to commit any credentials in the manifests and there is no risk of using wrong API keys because we hardcode the location of it in a shared password store.

BUT, all user of the team use the same API keys belonging to a specific user, not its personal API keys. This is really annoying, especially when a team member leaves and closes its Cloudstack account ; or when we want to look into Cloudstack logs to see who did what ; we definitely want to use one's personal API keys...

So the idea behind this PR is to be able to hardcode the location of a file in every team member's personal workstation containing personal API keys to access Cloudstack using this declaration:

```terraform
provider "cloudstack" {
  config = "${pathexpand("~/.cloudstack.ini")}"
  region = "my_region"
}
```
Like we do for example for the AWS provider:

```terraform
provider "aws" {
  profile = "foo"
  region = "eu-west-1"
}
```

I totally agree the Exoscale's Cloustack API client CLI is not the most used, but I guess it works for every Cloudstack deployment and to be honest I knew that it cloud be criticized because of that.

Would you be agree with a PR implementing the same thing but using CloudMonkey file format?

mcanevet:
@svanharmelen I'm now using CloudMonkey's config file. Is it OK for you now?

builtin/providers/cloudstack/provider.go
```diff
@@ -1,30 +1,46 @@
...
+				Optional: true,
+				Default:  "",
+			},
+
 			"api_url": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_URL", nil),
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_URL", ""),
```
svanharmelen: Why would you change this from `nil` to `""`?

builtin/providers/cloudstack/provider.go
```diff
@@ -1,30 +1,46 @@
...
+				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_API_KEY", ""),
 			},
 
 			"secret_key": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_SECRET_KEY", nil),
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_SECRET_KEY", ""),
 			},
```
svanharmelen: Please put the new options below the `secret_key`, as the existing `api_url`, `api_key` and `secret_key` would be the main way to use this provider, and alternatively you can point to a config file. So let's order the parameters in that same way...

builtin/providers/cloudstack/provider.go
```diff
@@ -1,30 +1,46 @@
...
 
 // Provider returns a terraform.ResourceProvider.
 func Provider() terraform.ResourceProvider {
+	usr, _ := user.Current()
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
+			"config": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  fmt.Sprintf("%s/.cloudmonkey/config", usr.HomeDir),
```
svanharmelen: You should not set a default here. Doing so will cause `providerConfigure` to always use the config file, even if you specified the API URL/key/secret.

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +88,42 @@ func Provider() terraform.ResourceProvider {
...
+			apiKey = section.Key("apikey").String()
+		}
+
+		if secretKey == "" {
+			secretKey = section.Key("secretkey").String()
+		}
+	}
+
+	if apiURL == "" {
+		return &Config{}, fmt.Errorf("No api_url provided")
```
svanharmelen: Just return `nil` here instead of allocating a new empty `Config` struct.

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +88,42 @@ func Provider() terraform.ResourceProvider {
...
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	apiURL := d.Get("api_url").(string)
+	apiKey := d.Get("api_key").(string)
+	secretKey := d.Get("secret_key").(string)
+
+	if configFile := d.Get("config").(string); configFile != "" {
+		config, err := ini.Load(configFile)
+		if err != nil {
+			return config, err
```
svanharmelen: Please return a `nil` instead of the `config`.

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +88,42 @@ func Provider() terraform.ResourceProvider {
...
+		}
+
+		section, err := config.GetSection(d.Get("profile").(string))
+		if err != nil {
+			return config, err
```
svanharmelen: Please return a `nil` instead of the `config`.

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +88,42 @@ func Provider() terraform.ResourceProvider {
...
+		if apiKey == "" {
+			apiKey = section.Key("apikey").String()
+		}
+
+		if secretKey == "" {
+			secretKey = section.Key("secretkey").String()
+		}
+	}
+
+	if apiURL == "" {
```
svanharmelen: Why do you only check the URL? We need to check the key and secret as well right?

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +88,42 @@ func Provider() terraform.ResourceProvider {
...
+			apiURL = section.Key("url").String()
+		}
+
+		if apiKey == "" {
+			apiKey = section.Key("apikey").String()
+		}
+
+		if secretKey == "" {
+			secretKey = section.Key("secretkey").String()
+		}
```
svanharmelen: Why do you only assign the config values if the variables are empty? Either all the values come from the config file, or they all come from the individual `api_url`, `api_key` and `secret_key`  parameters. To prevent unexpected behaviour they should never be mixed...

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +88,42 @@ func Provider() terraform.ResourceProvider {
 }
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	apiURL := d.Get("api_url").(string)
+	apiKey := d.Get("api_key").(string)
+	secretKey := d.Get("secret_key").(string)
+
+	if configFile := d.Get("config").(string); configFile != "" {
```
svanharmelen: I think it's cleaner to check for this like so:

```go
if configFile, ok := d.GetOk("config"); ok {
  config, err := ini.Load(configFile.(string))
<snip>
```
This would give an error if you did set the `config` param, but assigned it an empty string (the `ini.Load()` func would return an error in that case). I think that would be more clear for users.

builtin/providers/cloudstack/provider.go
```diff
@@ -1,30 +1,46 @@
...
+			"config": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  fmt.Sprintf("%s/.cloudmonkey/config", usr.HomeDir),
+			},
+
+			"profile": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "",
```
svanharmelen: If the default is an empty string, then please don't set a default...

builtin/providers/cloudstack/provider.go
```diff
@@ -1,30 +1,46 @@
...
+	"fmt"
+	"github.com/go-ini/ini"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
+	"os/user"
 )
 
 // Provider returns a terraform.ResourceProvider.
 func Provider() terraform.ResourceProvider {
+	usr, _ := user.Current()
```
mcanevet: done

builtin/providers/cloudstack/provider.go
```diff
@@ -1,30 +1,46 @@
 package cloudstack
 
 import (
+	"fmt"
+	"github.com/go-ini/ini"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
+	"os/user"
```
mcanevet: done

builtin/providers/cloudstack/provider.go
mcanevet: ok

mcanevet: I changed the logic to use the config file when profile is set so that I can set the default config file.

builtin/providers/cloudstack/provider.go
mcanevet: done

builtin/providers/cloudstack/provider.go
mcanevet: I thought the validation would fail if neither the environment variable nor the parameter was set with `nil`, but I tested and looks to be ok.

builtin/providers/cloudstack/provider.go
mcanevet: done

builtin/providers/cloudstack/provider.go
mcanevet: done

builtin/providers/cloudstack/provider.go
mcanevet: done

builtin/providers/cloudstack/provider.go
mcanevet: done

builtin/providers/cloudstack/provider.go
mcanevet: I'm trying to mimic the `aws` provider behavior where you can override the parameters in the `~/.aws/credentials` on provider instantiation.

builtin/providers/cloudstack/provider.go
mcanevet: done

mcanevet:
@svanharmelen I fixed most of your comment. I'm not sure we should mark parameters as conflicts as we should be able to override the config file we either the environment variables or hardcoded parameters, just like the `aws` provider does.

builtin/providers/cloudstack/provider.go
```diff
@@ -1,32 +1,51 @@
...
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 )
 
 // Provider returns a terraform.ResourceProvider.
 func Provider() terraform.ResourceProvider {
+	usr, err := user.Current()
+	if err != nil {
+		return nil
+	}
```
svanharmelen: This is still dangerous as the calling side of this function does not expect `nil`, but assumes a `*schema.Provider`. So when you do this and the calling side calls anything on the returned value, it will again panic.

builtin/providers/cloudstack/provider.go
```diff
@@ -1,32 +1,51 @@
...
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("CLOUDSTACK_SECRET_KEY", nil),
 			},
 
+			"config": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  fmt.Sprintf("%s/.cloudmonkey/config", usr.HomeDir),
```
svanharmelen: I more or less understand your take on this one, but I don't think we should mimic the AWS provider. This is a new feature that is not in any way something CloudStack users expect. I'm OK with adding it, but I would like to make sure the use of it is as clear as possible.

So either you use the file or the inline fields. But not a combination of both. I don't see the use case that would benefit from being able to do that. I only see the possible confusion this may cause users.

So please remove this default which will also solve the error handling thingy when calling `user.Current()` as that can then be deleted as well...

builtin/providers/cloudstack/provider.go
svanharmelen: As  explained in the comment above, I don't think we should add this option here. Let's make this as clear as possible and either use one or the other...

builtin/providers/cloudstack/provider.go
```diff
@@ -72,10 +91,42 @@ func Provider() terraform.ResourceProvider {
 }
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
+	apiURL := d.Get("api_url").(string)
+	apiKey := d.Get("api_key").(string)
+	secretKey := d.Get("secret_key").(string)
+
+	if profile, ok := d.GetOk("profile"); ok {
```
svanharmelen: I would revert this to checking the `config` field as that makes more sense (for people reading the code)

mcanevet:
@svanharmelen I added `ConflictsWith` but somehow it looks like it does not work. I can mix parameters without Terraform complaining... Do you have any idea?

mcanevet:
@svanharmelen do you have any idea for `ConflictsWith`?

svanharmelen:
I've been pretty busy today with some other PR's. I'll have a look at this one again coming Monday! From the top of my head I have no idea why this wouldn't work. So have to take a closer look...

mcanevet:
@svanharmelen did you find time to check this PR?

svanharmelen:
@mcanevet I just downloaded and compiled your branch and it works as expected:

image

mcanevet:
@svanharmelen cool, I must have do something wrong during my tests.
Could you please merge then?

svanharmelen:
I'll merge it and then add a small validation func to improve the UX just bit. Thanks!

PR #13941, opened 2017-04-25
improve SIGINT output

jbardin:
- add data-loss warning to SIGINT handler in apply
- remove redundant output when interrupting apply


jbardin:
@radeksimko, The backend message doesn't need to be the same, as both messages are printed one right after the other.

PR #13954, opened 2017-04-25
provider/google: fix panic in GKE provisioning with addons

danawillow:
If any element in `addons_config` is set, the data structure will be returned fully expanded. Check that subelements aren't empty before attempting to access their contents.


radeksimko:
@danawillow sorry for breaching the etiquette here a little and merging this for you, but I thought it would be good to get this in prior to the next release which we're going to cut very soon since it's fixing a crash. 😄

danawillow:
No worries! Thanks @radeksimko :)

PR #13965, opened 2017-04-26
provider/azurerm: Expose any dynamically-assigned private IP address

codekoala:
Currently, if you want to use a private IP address with an Azure Load Balancer, you must statically assign the address in order to reference the IP elsewhere (such as in an ``azurerm_dns_a_record``). If you have something like the following:

```hcl
resource "azurerm_lb" "demo" {
  name                = "demo-lb"
  location            = "${var.location}"
  resource_group_name = "${var.resource_group}"

  frontend_ip_configuration {
    name                          = "default"
    private_ip_address_allocation = "dynamic"
    subnet_id                     = "${var.subnet_id}"
  }
}
```

the dynamically-allocated private IP address is visible when you dig into the state using ``terraform console``, but it is inaccessible using interpolation.

This pull request exposes the dynamic private IP address (if any) as ``private_ip_address`` for use elsewhere:

```hcl
resource "azurerm_dns_a_record" "demo" {
  name                = "demo"
  zone_name           = "${var.dns_zone}"
  resource_group_name = "${var.resource_group}"
  ttl                 = "300"
  records             = ["${azurerm_lb.demo.private_ip_address}"]
}

```

### DIFF
--- a/builtin/providers/azurerm/resource_arm_loadbalancer.go
+++ b/builtin/providers/azurerm/resource_arm_loadbalancer.go
@@ -92,6 +92,11 @@ func resourceArmLoadBalancer() *schema.Resource {
 				},
 			},
 
+			"private_ip_address": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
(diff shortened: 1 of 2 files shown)


builtin/providers/azurerm/resource_arm_loadbalancer.go
```diff
@@ -172,7 +177,18 @@ func resourecArmLoadBalancerRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("resource_group_name", id.ResourceGroup)
 
 	if loadBalancer.LoadBalancerPropertiesFormat != nil && loadBalancer.LoadBalancerPropertiesFormat.FrontendIPConfigurations != nil {
-		d.Set("frontend_ip_configuration", flattenLoadBalancerFrontendIpConfiguration(loadBalancer.LoadBalancerPropertiesFormat.FrontendIPConfigurations))
+		ipconfig := flattenLoadBalancerFrontendIpConfiguration(loadBalancer.LoadBalancerPropertiesFormat.FrontendIPConfigurations)
+		d.Set("frontend_ip_configuration", ipconfig)
+
+		for _, config := range ipconfig {
+			cfg := config.(map[string]interface{})
+			if priv_ip, ok := cfg["private_ip_address"]; ok {
```
tombuildsstuff: I think it'd be better to iterate over the `loadBalancer.LoadBalancerPropertiesFormat.FrontendIPConfigurations` here, rather than the flattened list?

codekoala: That had occurred to me too. I opted to iterate over the flattened list so there would be a single piece of code accessing ``loadBalancer.LoadBalancerPropertyFormat.FrontendIPConfigurations.*.FrontendIPConfigurationPropertiesFormat.PrivateIPAddress``, just in case the API changes in the future.

I can definitely switch it up when I get back to a keyboard tomorrow if you'd like though!

codekoala:
Thank you!

PR #13982, opened 2017-04-26
provider/aws: Change running of CloudTrail tests to not be parallel

stack72:
We can only ever have 5 cloudtrails in an AWS account so we want to make
sure we run these tests serially to make sure we don't exhaust limits
and get non-deterministic failures


stack72:
screen shot 2017-04-27 at 02 39 32

Looks good as an acceptance test run

PR #13990, opened 2017-04-26
providers/heroku: use a helper for common app post-create tasks

danp:
Buildpacks were being configured post-create for non-org apps but not
for org apps. Consolidate post-create tasks into a helper used by both
paths.


catsby:
Also, "hi Dan 👋"

Issue #14003, opened 2017-04-26
aws_ebs_volume Tags not working/removed by aws_instance resource


eriksw:
After upgrading to Terraform 0.9.4, terraform removed the Name tags I had on some aws_ebs_volume resources.

### Terraform Version
0.9.4

### Affected Resource(s)
- aws_instance
- aws_ebs_volume

### Terraform Configuration Files
```hcl
resource "aws_instance" "cyanite-cassandra" {
  count         = "${var.cyanite-cassandra-count}"
  ami           = "${data.aws_ami.ubuntu-xenial.id}"
  instance_type = "t2.medium"

  # ebs_optimized          = true
  subnet_id = "${element(aws_subnet.private.*.id, count.index)}"

  vpc_security_group_ids = [
    "${aws_security_group.cyanite-cassandra.id}",
  ]

  iam_instance_profile = "${aws_iam_instance_profile.cyanite-cassandra.name}"
  user_data            = "${element(data.template_cloudinit_config.cyanite-cassandra.*.rendered, count.index)}"
  key_name             = "${aws_key_pair.default.key_name}"

  tags {
    Name = "${format("cyanite-cassandra-%d", count.index + 1)}"
    fqdn = "${format("cyanite-cassandra-%d.%s", count.index + 1, aws_route53_zone.beach.name)}"
  }

  lifecycle {
    ignore_changes = [
      "ami",
      "user_data",
    ]
  }
}

resource "aws_ebs_volume" "cyanite-cassandra-data" {
  count             = "${var.cyanite-cassandra-count}"
  availability_zone = "${element(aws_subnet.private.*.availability_zone, count.index)}"
  size              = 200
  encrypted         = true
  type              = "gp2"

  tags {
    Name = "${format("cyanite-cassandra-%d data", count.index + 1)}"
  }
}

resource "aws_volume_attachment" "cyanite-cassandra-data" {
  count       = "${var.cyanite-cassandra-count}"
  device_name = "/dev/xvdf"
  volume_id   = "${element(aws_ebs_volume.cyanite-cassandra-data.*.id, count.index)}"
  instance_id = "${element(aws_instance.cyanite-cassandra.*.id, count.index)}"

  lifecycle {
    # Needed to work around terraform weirdness. When replacing an instance, taint the particular attachment too.
    ignore_changes = [
      "instance_id",
      "volume_id",
    ]
  }
}
```

### Expected Behavior
Terraform should have left the Name tags on the ebs_volumes in place.

[...]

Issue #3531, opened 2015-10-16
feature request: allow tags for ebs_block_device entities within an aws_instance resource


FlorinAndrei:
I would like to be able to tag the additional volumes that I create attached to new instances. Like this:

```
resource "aws_instance" "pg02-prod2-useast1-aws" {
    ami = "ami-17fbbc72"
    instance_type = "c4.xlarge"
    key_name = "XXXXXXXXXXXXXXXX"
    subnet_id = "${aws_subnet.prod2-db-b.id}"
    source_dest_check = true
    security_groups = ["${aws_security_group.prod2-pg-instances.id}"]
    ebs_block_device {
      device_name = "/dev/sdf"
      volume_size = 2048
      volume_type = "gp2"
      delete_on_termination = true
      encrypted = false
      tags {
        Name = "pg02-prod2-useast1-aws:/var/lib/postgres"
      }
    }
    tags {
        Name = "pg02-prod2-useast1-aws"
    }
}
```

mmmorris1975:
My team relies on ebs volume tags heavily for identifying attachment information, and snapshot management.  This would be very helpful for us to add into TF

thebeardisred:
Yes I need this as well.

jimmystewpot:
This is very useful. There should also be the capability to automatically create tags for the volumes based on the instance tags.

jkopacze:
Good idea, its clearly a missing feature.

caarlos0:
:+1:

kovyrin:
Would be a great feature. Without it every time we drop an instance with delete_on_termination=false on its root partition (we rely heavily on EBS for root volumes) we end up with a volume that is pretty hard to associate with a particular node...

AlfonsoPantoja:
+1

tamsky:
:+1:

jkinred:
I'm keen to have a go at this but would appreciate input from @phinze or @jen20 before starting, as this enhancement requires changing the behaviour of the `aws_instance` resource when tags are specified.

The AWS API `BlockDeviceMapping` doesn't allow for tags to be set at creation time:
http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_BlockDeviceMapping.html
http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html

The most obvious way for me to implement this is to duplicate the behaviour from the `aws_ebs_volume` resource, which waits for the volume(s) to become available before setting the tags:
[...]

PR #14007, opened 2017-04-26
provider/aws: Set aws_instance volume_tags to be Computed

stack72:
Fixes: #14003

When an EBS volume was created and tags were specified on that resource
and NOT the aws_instance it was attached to, the tags would be removed
on subsequent Terraform runs.

We need to set volume_tags to be Computed to allow for changes to EBS
volumes not created as part of the instance but that are attached to the
instance

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccAWSInstance_volumeTagsComputed'
2017/04/27 07:33:36 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSInstance_volumeTagsComputed -timeout 120m
=== RUN   TestAccAWSInstance_volumeTagsComputed
--- PASS: TestAccAWSInstance_volumeTagsComputed (151.37s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	151.411s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -234,7 +234,7 @@ func resourceAwsInstance() *schema.Resource {
 
 			"tags": tagsSchema(),
 
-			"volume_tags": tagsSchema(),
+			"volume_tags": tagsSchemaComputed(),
 
 			"block_device": {
 				Type:     schema.TypeMap,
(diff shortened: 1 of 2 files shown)


hfarukuslu:
I guess no one cares about this bug.
#3531 was closed by a bot, and similar requests are being linked to that one. 
This is a very annoying bug for big solutions.

Easiest solution would be to set it as 

```
root_block_device { 
    tags { name= "Name tag on AWS" }  
} 
```

Please someone hear us :)

PR #14001, opened 2017-04-26
Fix vet errors in fastly provider

cblecker:
Fixes #13983.
Another instance of #13122 where vet errors weren't picked up by Travis, and as such, they exist in a tagged release.

PR #14010, opened 2017-04-26
build: Fix Travis vet issues

### LINKED ISSUE #13122: build: Travis not picking up vet errors
It seems that Travis isn't picking up vet errors:
- https://github.com/hashicorp/terraform/pull/12839
- https://github.com/hashicorp/terraform/pull/13121

Both of these were discovered on my local machine by just pulling down an updated master and running `make vet`. I'm not sure if this is intended, but it doesn't seem so.

cblecker:
Fixes #13122.

This runs the vet step *after* the test step, so that all dependencies are installed by the `go test -i` as mentioned here: https://github.com/hashicorp/terraform/pull/14001#issuecomment-297525938.

I'm going to submit a second PR with these changes *and* a purposeful vet error introduced to see if Travis picks it up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: trusty
 sudo: false
 language: go
 go:
-- 1.8
+- 1.8.x
 
 # add TF_CONSUL_TEST=1 to run consul tests
 # they were causing timouts in travis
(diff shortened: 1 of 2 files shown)


cblecker:
https://github.com/hashicorp/terraform/pull/14011 failed as expected when a vet error was introduced.

cblecker:
cc: @jbardin @stack72

stack72:
LGTM! Thanks for helping us out with this :)

Issue #12617, opened 2017-03-12
Enable support for Google Cloud SQL Postgres


samhamilton:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.8.8

### Affected Resource(s)
- google_sql_database_instance

### Terraform Configuration Files
```hcl
resource "google_sql_database_instance" "prod-us-east1-postgres-01" {
  name    = "prod-us-east1-postgres-01"
  region  = "us-east1"

  settings {
    database_version = "POSTGRES_9_6"
    tier = "db-f1-micro"
    disk_size = "10"
    disk_type = "PD_SSD"

  }
}
```

### Debug Output
I don't believe this is a bug, more a feature request 

### Panic Output
N/A

### Expected Behavior
Create a new Google Cloud managed postgres instance

### Actual Behavior
I got an error: `google_sql_database_instance.prod-us-east1-postgres-01: settings.0: invalid or unknown key: database_version`

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Nope

### References
Could not find any other open Google Cloud Postgres issues.

samhamilton:
Sorry - please ignore

nodesocket:
Is PostgreSQL supported using `google_sql_database_instance`?

samhamilton:
Yes just set `database_version = "POSTGRES_9_6"`

ctavan:
Just to be clear, the correct config has to look like this (notice that `database_version` is defined on the top level):

```hcl
resource "google_sql_database_instance" "prod-us-east1-postgres-01" {
  name             = "prod-us-east1-postgres-01"
  region           = "us-east1"
  database_version = "POSTGRES_9_6"

  settings {
    tier      = "db-f1-micro"
    disk_size = "10"
    disk_type = "PD_SSD"
  }
}
```

ctavan:
I proposed to add this to the official documentation in #14028

davejlong:
This no longer works. Google Cloud SQL requires a custom machine instance or shared-core for PostgreSQL. For example, the following resource:

```
resource "google_sql_database_instance" "development" {
  name = "development-instance"
  region = "${var.region}"
  database_version = "POSTGRES_9_6"

  settings {
    tier = "db-n1-standard-1"
    backup_configuration {
      enabled = true
      start_time = "03:00"
[...]

PR #14028, opened 2017-04-27
provider/google: Improve Google provider documentation

ctavan:
Add two small improvements to the google provider docs.

### DIFF
--- a/website/source/docs/providers/google/r/sql_database.html.markdown
+++ b/website/source/docs/providers/google/r/sql_database.html.markdown
@@ -24,7 +24,7 @@ resource "google_sql_database_instance" "master" {
 }
 
 resource "google_sql_database" "users" {
-  name     = "image-store-bucket"
+  name     = "users"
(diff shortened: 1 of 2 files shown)


ctavan:
As @davejlong pointed out in https://github.com/hashicorp/terraform/issues/12617#issuecomment-298395597 Google seems to have changed the API slightly removing support for machine types in Postgres. I suppose this means we cannot merge this one yet.

ctavan:
I added the findings from https://github.com/hashicorp/terraform/issues/12617#issuecomment-298617855 to the documentation. Not sure though if the terraform docs really want to document a feature which is totally undocumented on the Google-Side 😼

ctavan:
@danawillow @paddycarver sorry for bumping this, but is there any interest in merging these additions to the terraform docs? I can would assume that some terraform users would be happy to give CloudSQL Postgres a try…

ctavan:
Moved to https://github.com/terraform-providers/terraform-provider-google/pull/124

PR #14030, opened 2017-04-27
docs/digitalocean: Add image data source to sidebar

radeksimko:
Ref #13787

### DIFF
--- a/website/source/layouts/digitalocean.erb
+++ b/website/source/layouts/digitalocean.erb
@@ -10,6 +10,15 @@
         <a href="/docs/providers/do/index.html">DigitalOcean Provider</a>
                 </li>
 
+        <li<%= sidebar_current("docs-do-datasource") %>>
+        <a href="#">Data Sources</a>
+                <ul class="nav nav-visible">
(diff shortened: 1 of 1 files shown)


website/source/layouts/digitalocean.erb
```diff
@@ -11,6 +11,15 @@
                 </li>
 
         <li<%= sidebar_current("docs-do-resource") %>>
```
tombuildsstuff: Given this is a data source - I think this should be `docs-do-datasource`?

website/source/layouts/digitalocean.erb
```diff
@@ -11,6 +11,15 @@
...
+        <a href="#">Data Sources</a>
+                <ul class="nav nav-visible">
+                    <li<%= sidebar_current("docs-do-resource-image") %>>
```
tombuildsstuff: Given this is a data source, should this be `docs-do-datasource-image`?

radeksimko: ah, good point, will fix that

radeksimko:
@tombuildsstuff fixed.

PR #14035, opened 2017-04-27
providers/heroku: add heroku_app_feature resource

danp:
All credit to @bernerdschaefer, just PR'ing something he did internally.

Also, 👋 @catsby!

### DIFF
--- a/builtin/providers/heroku/provider.go
+++ b/builtin/providers/heroku/provider.go
@@ -1,7 +1,9 @@
 package heroku
 
 import (
+	"fmt"
 	"log"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


catsby:
@danp ah, I merged this before I noticed the new resource wasn't in the layout! I added that in https://github.com/hashicorp/terraform/commit/d6560b96d72f5c5a5f582a5e01c8df14b199817e and had to do a little hack on the `heroku_app` docs to get the regex to only highlight one or the other on our docs page. The regex hack is a blemish but it works for now 😄

danp:
Oh, thanks! Will be sure to update that list too next time.

PR #14041, opened 2017-04-27
Added data source for aws_efs_file_system

craigknott:
heavily based off of resource_aws_efs_file_systems read method.

### DIFF
--- a/builtin/providers/aws/data_source_aws_efs_file_system.go
+++ b/builtin/providers/aws/data_source_aws_efs_file_system.go
@@ -1,0 +1,110 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/efs"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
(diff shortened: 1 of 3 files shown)


stack72:
Hi @craigknott 

Thanks so much for the work here - I have made a few changes to the code as follows:

* renamed id to be file_system_id - ID is generally not something we use as a parameter name
* changed performance_mode to be computed only as we never use it as a parameter
* changes file_system_id and creation_token to be forcenew so that any changes forces a new data source

I also added some documentation as well. You tests passed as expected which was awesome :)

```
% make testacc TEST=./builtin/providers/aws TESTARGS='-run=TestAccDataSourceAwsEfsFileSystem'
2017/04/29 11:17:59 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccDataSourceAwsEfsFileSystem -timeout 120m
=== RUN   TestAccDataSourceAwsEfsFileSystem
--- PASS: TestAccDataSourceAwsEfsFileSystem (48.82s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	48.844s
```

Thanks for the work here :) I just merged this manually

Paul

PR #14045, opened 2017-04-27
provider/google: fix project metadata tests

paddycarver:
Update our project metadata tests to stand up their own projects, so
they don't trample all over each other anymore.

The fixes for this were more invasive than I had hoped they would be,
but the tests all pass now (when run sequentially) and there's no reason
for them not to pass when run in parallel.


paddycarver:
Both comments should be addressed :)

paddycarver:
Self-merging as I have two LGTMs and Travis passed.

Issue #14049, opened 2017-04-27
creating aws_instance with tags fails in GovCloud (0.9.4 regression)


dsw2127:
### Terraform Version
0.9.4

### Affected Resource(s)
- aws_instance

### Terraform Configuration Files
```
resource "aws_instance" "instance" {
  ami                         = "${var.ami}"
  key_name                    = "${var.key_name}"
  instance_type               = "t2.medium"
  vpc_security_group_ids      = ["${var.security_group_id}"]
  subnet_id                   = "${var.subnet_id}"
  associate_public_ip_address = "true"

  ebs_block_device {
    device_name           = "/dev/sda1"
    volume_size           = "15"
    delete_on_termination = "true"
  }

  ebs_block_device {
    device_name           = "/dev/sdz"
    volume_type           = "gp2"
    volume_size           = "15"
    delete_on_termination = "true"
  }

  tags {
    Name = "ami builder instance"
  }
}
```

### Expected Behavior
Instance launched with tags applied afterwards, as occurred in Terraform 0.9.3

### Actual Behavior
```
Error applying plan:



1 error(s) occurred:



* aws_instance.instance: 1 error(s) occurred:



* aws_instance.instance: Error launching source instance: UnsupportedOperation: Specifying tags during instance launch is not supported in this region.

	status code: 400, request id: f5aa3123-ec4e-4815-8325-d0ba76706740
```

### Steps to Reproduce
Specify an `aws_instance` with `tags` in region `us-gov-west-1` and apply.

### References
Tags were changed to be created on launch by https://github.com/hashicorp/terraform/pull/13945 . However, as noted in https://aws.amazon.com/blogs/aws/new-tag-ec2-instances-ebs-volumes-on-creation/ (at the bottom) this is not available in GovCloud or Beijing.

catsby:
Hey @dsw2127 sorry for the hassle here, we're patching this in https://github.com/hashicorp/terraform/pull/14055 and should get this straightened out soon!

rokka-n:
Looks like it's fixed in 0.9.5

stack72:
Thanks for letting me know @rokka-n

dsw2127:
Thanks for fixing @stack72! I upgraded to 0.9.5 when it came out, and it indeed resolves the issue.

stack72:
excellent! sorry for the hassles here

PR #14055, opened 2017-04-27
provider/aws: Exclude aws_instance volume tagging for China and Gov C…

stack72:
…louds

Fixes: #14049

The China and Gov regions do not support the new way of tagging
instances and volumes on creation. Therefore, we need to hack this to
make sure we don't try and set these on instance creation

### DIFF
--- a/builtin/providers/aws/config.go
+++ b/builtin/providers/aws/config.go
@@ -171,6 +171,20 @@ func (c *AWSClient) DynamoDB() *dynamodb.DynamoDB {
 	return c.dynamodbconn
 }
 
+func (c *AWSClient) IsGovCloud() bool {
+	if c.region == "us-gov-west-1" {
+		return true
+	}
+	return false
+}
+
+func (c *AWSClient) IsChinaCloud() bool {
+	if c.region == "cn-north-1" {
+		return true
+	}
+	return false
+}
+
 // Client configures and returns a fully initialized AWSClient
 func (c *Config) Client() (interface{}, error) {
 	// Get the auth and region. This can fail if keys/regions were not
--- a/builtin/providers/aws/resource_aws_instance.go
+++ b/builtin/providers/aws/resource_aws_instance.go
@@ -432,32 +432,35 @@ func resourceAwsInstanceCreate(d *schema.ResourceData, meta interface{}) error {
 		runOpts.Ipv6Addresses = ipv6Addresses
 	}
 
-	tagsSpec := make([]*ec2.TagSpecification, 0)
+	restricted := meta.(*AWSClient).IsGovCloud() || meta.(*AWSClient).IsChinaCloud()
+	if !restricted {
+		tagsSpec := make([]*ec2.TagSpecification, 0)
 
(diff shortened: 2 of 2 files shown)


stack72:
screen shot 2017-04-28 at 08 38 00

radeksimko:
I think we should just silently ignore the error if we get `UnsupportedOperation` when listing the tags. That way we won't have to ship a new release just because Amazon suddenly decides to support tagging in gov/chinese region and also the fix will work for any other regions which might not support tagging for any weird reason.

We should _not_ ignore errors coming from add/remove tags though - because that's not something user would bump into unintentionally - they'd have to have tags in the config and they might want to know that tags defined in the config were not applied because they're in a region which doesn't support it.

-----

Just to clarify - this is different to the discussion we had between me and you a while back about silently ignoring failed API calls where I was on the other side (of _not_ ignoring) - that was about insufficient permissions - which is something user can fix. This is different. 😃

catsby:
Ignoring `UnsupportedOperation` is OK in some cases, maybe? Wouldn't that cause a diff in what your config says and what your infrastructure has? OR worse, not report a diff when there is one?

The regression in question threw `UnsupportedOperation` and simply did not create `aws_instance` resources at all... we can't ignore that (not that you were suggesting we would)

radeksimko:
> The regression in question threw `UnsupportedOperation` and simply did not create `aws_instance` resources at all... we can't ignore that (not that you were suggesting we would)

Oh I see, good point! 😞  It's more difficult than I thought.

I still think we should be notifying the user about the fact that they can't specify certain fields in their region. I'm not suggesting we implement it here as it may be difficult and it's important to make things work first and unblock folks, but as a long-term solution I think we should perhaps support something like this in the schema:

```go
map[string]*Schema{
    "field_name": {
        Type: TypeString,
        UnsupportedFunc: func(d *ResourceData, meta interface{}) error {
            restricted := meta.(*AWSClient).IsGovCloud() || meta.(*AWSClient).IsChinaCloud()
            if restricted {
                return fmt.Errorf("field_name is not supported in gov and chinese regions")
            }
        },
    },
}
```

We can't avoid cutting a release after Amazon begins to support tagging, but at least we wouldn't be so silent about this and we could even return that ^ error as part of plan/refresh as we have the region available.

catsby:
`UnsupportedFunc` I like this... mostly. I like it, but I wish we didn't need it 😄

builtin/providers/aws/config.go
```diff
@@ -171,6 +171,20 @@ func (c *AWSClient) DynamoDB() *dynamodb.DynamoDB {
 	return c.dynamodbconn
 }
 
+func (c *AWSClient) IsGovCloud() bool {
+	if c.region == "us-gov-west-1" {
```
tamsky: seems like this would work better as a regex match on `^us-gov-'

https://aws.amazon.com/blogs/aws/aws-govcloud-us-heads-east-new-region-in-the-works-for-2018/

same for `^cn-`

Issue #14024, opened 2017-04-27
AWS SNS delivery policy always planned to be changed


dgolja:
Hi there,

When creating an AWS SNS whenever I run `terraform plan/apply` the delivery policy resource is planned to be modified. 

### Terraform Version
* terraform v0.9.2
* terraform v0.9.3
* terraform v0.9.4

### Affected Resource(s)
- aws_sns_topic

### Terraform Configuration Files

Content of `sns.tf`

```hcl
provider "aws" {
  profile = "omitted"
  region  = "omitted"
}

resource "aws_sns_topic" "sns" {
  name = "delivery_policy_bug"
  delivery_policy = "${data.template_file.delivery_policy.rendered}"
}

data "template_file" "delivery_policy" {
  vars {
    minimum_delay = "${var.minimum_delay}"
    maximum_delay = "${var.maximum_delay}"
  }

  template = "${file("delivery_policy.tpl.json")}"
}

variable "minimum_delay" {
  default = 20
}

variable "maximum_delay" {
  default = 20
}

```

Content of `delivery_policy.tpl.json` file

```
{
  "http": {
    "defaultHealthyRetryPolicy": {
      "minDelayTarget": ${minimum_delay},
      "maxDelayTarget": ${maximum_delay},
      "numRetries": 3,
      "numMaxDelayRetries": 0,
      "numNoDelayRetries": 0,
      "numMinDelayRetries": 0,
      "backoffFunction": "linear"
    },
    "disableSubscriptionOverrides": false
  }
}
```

### Expected Behavior
If I run `terraform plan` after `terraform apply` I expect that no changes would be detected.

### Actual Behavior
When running `terraform plan` for the second time I get this:

```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.template_file.delivery_policy: Refreshing state...
aws_sns_topic.sns: Refreshing state... (ID: arn:aws:s...licy_bug)
The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed. Cyan entries are data sources to be read.

Note: You didn't specify an "-out" parameter to save this plan, so when
"apply" is called, Terraform can't guarantee this is what will execute.

~ aws_sns_topic.sns
[...]

PR #14064, opened 2017-04-28
provider/aws: Fixes the bug where SNS delivery policy get always recreated

dgolja:
This PR should address the problem from https://github.com/hashicorp/terraform/issues/14024.

I tested it locally and it works, however I couldn't find any good example for how to test this behaviour. Some tips would be great.

### DIFF
--- a/builtin/providers/aws/resource_aws_sns_topic.go
+++ b/builtin/providers/aws/resource_aws_sns_topic.go
@@ -55,9 +55,15 @@ func resourceAwsSnsTopic() *schema.Resource {
 				},
 			},
 			"delivery_policy": &schema.Schema{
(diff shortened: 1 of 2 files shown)


grubernaut:
Hi @dgolja, thanks for the PR! This looks like a decent bugfix to me. For an acceptance test, we can add a new test here: https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_sns_topic_test.go#L69, following the other acceptance test styles above to test setting a `delivery_policy` doesn't trigger a resource diff. 
Thanks!

vicente-valls:
this is happening again on version 0.11.8

PR #14080, opened 2017-04-28
provider/aws: Add EMR Security Configuration Resource

catsby:
Adds new `emr_security_configuration` resource for Terraform to manage. 
This is a companion PR to https://github.com/hashicorp/terraform/pull/13620 and implements https://github.com/hashicorp/terraform/issues/11830

It *does not* cover [ListConfigurations](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_ListSecurityConfigurations.html); that should probably be a data source


builtin/providers/aws/resource_aws_emr_security_configuration.go
```diff
@@ -0,0 +1,134 @@
...
+			log.Printf("[WARN] EMR Security Configuraiton (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("creation_date", resp.CreationDateTime)
+	d.Set("configuration", resp.SecurityConfiguration)
+
```
grubernaut: Need to set `name` here for the import case, more than likely

website/source/docs/providers/aws/r/emr_security_configuration.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Optional) A unique name for this Security Configuration
+* `name_prefix` - (Optional) A prefix for the name of this Security Configuration. 
+  Terraform will generate a unique suffix.E.g: `tf-emr-sc-1234`
```
grubernaut: Maybe clarify that this is only when both `name` and `name_prefix` are unset

builtin/providers/aws/resource_aws_emr_security_configuration.go
```diff
@@ -0,0 +1,134 @@
...
+					}
+					return
+				},
+			},
+
+			"configuration": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateJsonString,
```
grubernaut: Should we normalize the JSON string to prevent diffs?

catsby: Sadly it seems that adding the normal `StateFunc` for this actually causes some diffs

catsby:
```
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSEmrSecurityConfiguration -timeout 120m
=== RUN   TestAccAWSEmrSecurityConfiguration_importBasic
--- PASS: TestAccAWSEmrSecurityConfiguration_importBasic (11.32s)
=== RUN   TestAccAWSEmrSecurityConfiguration_basic
--- PASS: TestAccAWSEmrSecurityConfiguration_basic (8.92s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    20.266s
```

PR #14082, opened 2017-04-28
provider/fastly: Make request_condition optional

mmwtsn:
Makes the `request_condition` setting optional for Fastly `request_settings` as seen here:

Fastly request setting example

This also updates the relevant documentation.


mmwtsn:
Fixed by #14103.

PR #14084, opened 2017-04-28
Document GOOGLE_APPLICATION_CREDENTIALS support.

### LINKED ISSUE #12626: Google provider: Use GOOGLE_APPLICATION_CREDENTIALS environment file
Most Google client code uses the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to designate a path to the credentials JSON file to open.  

Examples:

* https://developers.google.com/identity/protocols/application-default-credentials#howtheywork
* https://github.com/google/google-api-ruby-client/blob/v0.10/samples/cli/README.md
* https://github.com/google/google-auth-library-php

Terraform should be able to use it too. 

(Note that this is separate from the existing support for the `GOOGLE_CREDENTIALS` environment variable, in which the *actual* JSON must be placed, as opposed to a path to a file containing it.  I'm curious about the history behind that; it's very unusual to place JSON -- especially JSON containing secrets -- in an environment variable.)

paddycarver:
We've supported GOOGLE_APPLICATION_CREDENTIALS as an environment
variable (it comes free with our OAuth2 client) but it has never been
documented. Documenting it now to resolve #12626.

### DIFF
--- a/website/source/docs/providers/google/index.html.markdown
+++ b/website/source/docs/providers/google/index.html.markdown
@@ -46,6 +46,10 @@ The following keys can be used to configure the provider.
     * `GOOGLE_CLOUD_KEYFILE_JSON`
     * `GCLOUD_KEYFILE_JSON`
 
+    The [`GOOGLE_APPLICATION_CREDENTIALS`](https://developers.google.com/identity/protocols/application-default-credentials#howtheywork)
+    environment variable can also contain the path of a file to obtain credentials
+    from.
+
 * `project` - (Required) The ID of the project to apply any resources to.  This
   can be specified using any of the following environment variables (listed in
   order of precedence):


naoleyrashmi:
Hello Team,

Seems this issue is still underlying. I did try using environment variable as below so as to pass the value of credentials.json file. But still i am getting below error:

Error: provider.google: credentials are not valid JSON '/home/rashmin/GCTF/01-key/credentials.json': invalid character '/' looking for beginning of value

Variables Tried:

* `GOOGLE_CREDENTIALS`
* `GOOGLE_CLOUD_KEYFILE_JSON`
* `GCLOUD_KEYFILE_JSON`

Whereas if i add the direct path instead of the variable it works seamlessly . Can anyone please help. Please let me know if a new issue needs to be raised for same.

Issue #14074, opened 2017-04-28
Unable to create custom AWS Config rule


bodgit:
I'm trying to create a custom AWS Config rule, I have the Lambda part set up but when I try and create the `aws_config_config_rule` resource I get an error from the AWS API. Details below:

### Terraform Version
0.9.4

### Affected Resource(s)
- aws_config_config_rule

### Terraform Configuration Files
```hcl
provider "aws" {}

resource "aws_config_config_rule" "example" {
  name = "example"

  source {
    owner             = "CUSTOM_LAMBDA"
    source_identifier = "arn:aws:lambda:eu-west-1:123412341234:function:example"

    source_detail {
      message_type = "ConfigurationItemChangeNotification"
    }
  }
}
```

### Debug Output
I can't access gists from behind my corporate proxy, but here's the snippet of the debug from when it tries to create the rule:

```
2017/04/28 15:20:25 [DEBUG] apply: aws_config_config_rule.example: executing Apply
2017/04/28 15:20:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/04/28 15:20:25 [DEBUG] No meta timeoutkey found in Apply()
2017/04/28 15:20:25 [DEBUG] plugin: terraform: aws-provider (internal) 2017/04/28 15:20:25 [DEBUG] Creating AWSConfig config rule: {
2017/04/28 15:20:25 [DEBUG] plugin: terraform:   ConfigRule: {
2017/04/28 15:20:25 [DEBUG] plugin: terraform:     ConfigRuleName: "example",
2017/04/28 15:20:25 [DEBUG] plugin: terraform:     Source: {
2017/04/28 15:20:25 [DEBUG] plugin: terraform:       Owner: "CUSTOM_LAMBDA",
2017/04/28 15:20:25 [DEBUG] plugin: terraform:       SourceDetails: [{
2017/04/28 15:20:25 [DEBUG] plugin: terraform:           MessageType: "ConfigurationItemChangeNotification"
2017/04/28 15:20:25 [DEBUG] plugin: terraform:         }],
2017/04/28 15:20:25 [DEBUG] plugin: terraform:       SourceIdentifier: "arn:aws:lambda:eu-west-1:123412341234:function:example"
2017/04/28 15:20:25 [DEBUG] plugin: terraform:     }
2017/04/28 15:20:25 [DEBUG] plugin: terraform:   }
2017/04/28 15:20:25 [DEBUG] plugin: terraform: }
[...]

PR #14090, opened 2017-04-29
provider/aws Added default to event source

mbfrahry:
Looking at the api docs https://docs.aws.amazon.com/config/latest/APIReference/API_SourceDetail.html, EventSource is not required but it only accepts one value so we'll just default that in Terraform so #14074 doesn't happen again.

### DIFF
--- a/builtin/providers/aws/resource_aws_config_config_rule.go
+++ b/builtin/providers/aws/resource_aws_config_config_rule.go
@@ -110,6 +110,7 @@ func resourceAwsConfigConfigRule() *schema.Resource {
 									"event_source": {
 										Type:     schema.TypeString,
 										Optional: true,
+										Default:  "aws.config",
 									},
 									"maximum_execution_frequency": {
 										Type:         schema.TypeString,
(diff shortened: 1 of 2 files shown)


builtin/providers/aws/resource_aws_config_config_rule.go
```diff
@@ -110,6 +110,7 @@ func resourceAwsConfigConfigRule() *schema.Resource {
 									"event_source": {
 										Type:     schema.TypeString,
 										Optional: true,
+										Default:  "aws.config",
```
stack72: @mbfrahry I think we are going to need a migration for this - if people are not setting this value then they may get changes in their plans

Thoughts?

mbfrahry: I'm not sure we need to. If someone has a `source_detail` in their elb then they have to have `aws.config` or it'll error out like in #14074. We could do a migration just to be safe but I don't think it's needed

catsby: 👍

PR #14097, opened 2017-04-30
provider/aws: add additional custom service endpoint options

thirstydeveloper:
* Add option for custom CloudFormation, KMS, RDS, SNS, SQS endpoints
* Minor formatting to arrange internal endpoint variables alphabetically and with consistent naming
* Update CHANGELOG
* Update docs accordingly


stack72:
Hi @thirstydeveloper 

Thanks for the work here - I am going to bring this down locally and remove the update of the CHANGELOG. We generally don't expect contributors to update that document. The project moves so fast so we would get a lot of merge conflicts :)

Thanks again

Paul

Issue #13970, opened 2017-04-26
Remote-exec provisioner crashes when zero value strings are used in slice


vtorhonen:
### Terraform Version

Terraform v0.9.3

### Affected Resource(s)

Any resource using `remote-exec` provisioner.

### Terraform Configuration Files

With `inline` argument:

```hcl
resource "null_resource" "foo" {
  connection {
    host = "localhost"
    port = "2222"
    user = "someuser"
    password = "somepassword"
  }
  provisioner "remote-exec" {
    inline = [ "" ]
  }
}
```

With `scripts` argument:

```hcl
resource "null_resource" "foo" {
  connection {
    host = "localhost"
    port = "2222"
    user = "someuser"
    password = "somepassword"
  }
  provisioner "remote-exec" {
    scripts = [ "" ]
  }
}
```

### Debug Output

https://gist.github.com/vtorhonen/f35988391b043f6ae9217fefe62f3d9e

### Panic Output

Causes a panic, output in the gist above.

### Expected Behavior

Personally I think Terraform should consider a zero-value string in `inline` or `scripts` arguments as a no-op. However, it could give an error stating that a non zero-value string is required.

### Actual Behavior

Terraform crashed when `inline` parameter was set to a slice including a zero value string.

### Steps to Reproduce

1. Run `terraform apply` with the given example HCLs.

### Important Factoids

None.

### References

It might be that the logic behind handling zero value strings has changed between 0.8.8 and 0.9.3. Issue #10144 shows that in Terraform 0.8.8 an empty string causes Terraform to fail until provisioner timeout has been reached.

vtorhonen:
This crash is caused by type assertions in [resource_provisioner.go](builtin/provisioners/remote-exec/resource_provisioner.go), which currently are not checking whether or not the type assertion succeeded. For example, in `generateScripts()` we see the following:

```go
	var lines []string
	for _, l := range d.Get("inline").([]interface{}) {
		lines = append(lines, l.(string))
	}
```

Unsafe type assertions are used multiple times in the file. There are many ways to fix this, like using a wrapper function for doing the type assertion and then by using `recover()` to print out the error, but is the underlying problem in the way HCLs are parsed? In 0.8.8 things were different as a zero-value string was not parsed as `nil`.

PR #14134, opened 2017-05-01
provisioner/remote-exec: Fix panic from remote_exec provisioner

grubernaut:
Fixes panic on `nil` values of `inline` and `scripts` from improper interface casts.

Fixes: #13970


mbfrahry:
LGTM!

PR #14151, opened 2017-05-02
Add dockercfg example to Kubernetes secret

phekmat:


radeksimko:
Thanks @phekmat 
just FYI - we also have [`/examples`](https://github.com/hashicorp/terraform/tree/master/examples) directory where we keep more complex examples 😉

PR #14154, opened 2017-05-02
Add rancher provider registration labels

cloudnautique:
This PR is built off of: #13908 It is a super set, I split them up for easier review. Please let me know if you want me to do anything different here.

The enhancement here is to allow the user to add host labels to the registration command for an environment. The primary usecase for this is to support users being able to launch HA K8s clusters from Terraform using Rancher.


onlydole:
@mbfrahry - Is there an ETA or verdict on when this will be merged in?

cloudnautique:
@mbfrahry, any chance this can make it into 0.9.5? We have some folks that would love for this to be in the main line. Or is it more of a 0.10.0 type change? We would like to be able to manage expectations with our users. 

Anything I can do to help with the review process?

cloudnautique:
Closing this, and will just focus on #13908. I'll add the registration stuff to it as well.

PR #14155, opened 2017-05-02
provider/gitlab: add gitlab_project_hook resource

richardc:
Here we add a new resource type `gitlab_project_hook`.  It allows for
management of custom hooks for a gitlab project.

This is a relatively simple resource as a project hook is a simple
association between a project, and a url to hit when one of the flagged
events occurs on that project.

Hooks (called Webhooks in some user documentation, but simply Hooks
in the api documentation) are covered here for users
https://docs.gitlab.com/ce/user/project/integrations/webhooks.html and
in the API documentation at
https://docs.gitlab.com/ce/api/projects.html#hooks

### DIFF
--- a/builtin/providers/gitlab/resource_gitlab_project_hook.go
+++ b/builtin/providers/gitlab/resource_gitlab_project_hook.go
@@ -1,0 +1,185 @@
+package gitlab
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	gitlab "github.com/xanzy/go-gitlab"
+)
+
+func resourceGitlabProjectHook() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGitlabProjectHookCreate,
+		Read:   resourceGitlabProjectHookRead,
+		Update: resourceGitlabProjectHookUpdate,
+		Delete: resourceGitlabProjectHookDelete,
+
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"url": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"token": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+			},
+			"push_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"issues_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"merge_requests_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"tag_push_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"note_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"build_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"pipeline_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"wiki_page_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"enable_ssl_verification": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+		},
+	}
+}
+
+func resourceGitlabProjectHookCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*gitlab.Client)
+	project := d.Get("project").(string)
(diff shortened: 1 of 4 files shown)


builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional:  true,
+				Sensitive: true,
+			},
+			"push_events": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"issues_events": {
+				Type:     schema.TypeBool,
```
stack72: What is the default here? There is a known issue with d.GetOk around using the value false - maybe we should specify a default?

builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional: true,
+			},
+			"merge_requests_events": {
+				Type:     schema.TypeBool,
```
stack72: Same comment about default

builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional: true,
+			},
+			"tag_push_events": {
+				Type:     schema.TypeBool,
```
stack72: Same comment about default

builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional: true,
+			},
+			"note_events": {
+				Type:     schema.TypeBool,
```
stack72: Same comment about default

builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional: true,
+			},
+			"build_events": {
+				Type:     schema.TypeBool,
```
stack72: Same comment about default

builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional: true,
+			},
+			"pipeline_events": {
+				Type:     schema.TypeBool,
```
stack72: Same comment about default

builtin/providers/gitlab/resource_gitlab_project_hook.go
```diff
@@ -0,0 +1,185 @@
...
+				Optional: true,
+			},
+			"wiki_page_events": {
+				Type:     schema.TypeBool,
```
stack72: Same comment about default

stack72:
Hi @richardc 

I added some comments about default values of bools due to a bug i described

apart from that, it's looking good - thoughts on those would be good

Thanks

Paul

builtin/providers/gitlab/resource_gitlab_project_hook.go
richardc: The default for most of the bools in this type is false so this should be fine as that's the zero value for a golang bool.

Could you drop me a link to the bug report about GetOk so I can get a better understanding of the potential bug?  Thanks

richardc: That said if you think it would make the intent more clear to supply explicit false defaults rather than the implied zero-false that's a simple edit that I'll be happy to make tomorrow when I'm back at a machine with a development environment

stack72:
Hi @richardc 

ok, false sounds good here - I am going to make that change for you - it will take like 1 minute and then run the acceptance tests :)

Thanks for the work!

Paul

richardc:
@stack72 👍most kind, cheers

stack72:
```
% make testacc TEST=./builtin/providers/gitlab                                                                        130 ↵ ✹ ✭
2017/05/04 19:57:46 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/gitlab -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccGitlabProjectHook_basic
--- PASS: TestAccGitlabProjectHook_basic (48.64s)
=== RUN   TestAccGitlabProject_basic
--- PASS: TestAccGitlabProject_basic (32.75s)
=== RUN   TestGitlab_validation
--- PASS: TestGitlab_validation (0.00s)
=== RUN   TestGitlab_visbilityHelpers
--- PASS: TestGitlab_visbilityHelpers (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/gitlab	81.406s
```

PR #14170, opened 2017-05-03
aws_route53_record: More consistent unquoting of TXT/SPF records.

cakoose:
Before, 'flatten' would remove the first two quotes.  This results in confusing behavior if the value contained two quoted strings.

Now, 'flatten' we only remove the surrounding qutoes, which is more consistent with 'expand'.

Should enable stuff like hashicorp/terraform#8423.

### DIFF
--- a/builtin/providers/aws/resource_aws_route53_record_test.go
+++ b/builtin/providers/aws/resource_aws_route53_record_test.go
@@ -406,6 +406,24 @@ func TestAccAWSRoute53Record_empty(t *testing.T) {
 	})
 }
 
+// Regression test for https://github.com/hashicorp/terraform/issues/8423
+func TestAccAWSRoute53Record_longTXTrecord(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_route53_record.long_txt",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckRoute53RecordDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccRoute53RecordConfigLongTxtRecord,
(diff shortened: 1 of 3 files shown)


cakoose:
This is option 1 discussed in #11257.  Users who want to use multiple quoted strings in a single entry are expected to supply the quotes themselves.

If we'd rather do option 2, we need to make the flatten/expand functions a little more invasive.

cakoose:
I think option 2 could be the right to go, but just to restate the motivation for option 1:
- The 255-character limit isn't just a Route 53 peculiarity.  It's also a limitation of BIND.  I can't quite figure out if it's part of the DNS spec, but it's mentioned in the SPF spec ([RFC 4408, §3.1.3](https://tools.ietf.org/html/rfc4408#section-3.1.3)).
- The choice of where to split is visible to DNS clients.  When I use `dig` to query Route 53, the response has the multiple strings quotes as I specified.  (The spacing between the strings is normalized to a single space, though.)

cakoose:
I just wanted to try one final explanation for why I don't think this is a breaking change :-)

Most TXT entries only have a single quoted string.  In those cases, the behavior is unchanged.

Sometimes, a TXT entry will have multiple quoted strings.  In that case, Terraform will currently be broken(ish); it will keep saying "there are differences" even when there are not.  Because of this, I don't think anyone is able to use Terraform for TXT entries with multiple quoted strings.  Because this PR only changes the behavior in those cases, maybe it's not a breaking change?

Another way of looking at it: let's say Terraform crashed if there were multiple quoted strings.  Fixing that crash would be changing the behavior, but not really a breaking change, right?  (The current behavior isn't a crash, but it's somewhat similar.)

Then again, I'm pretty new to Terraform.  Maybe the project's definition of a breaking change is different from the definition I'm using.

cakoose:
Re: BC -- Makes sense.

Thanks for adding the test.  Would have taken me much longer.

Issue #11147, opened 2017-01-11
azurerm_virtual_machine  autorest/azure: Service returned an error. Status=400 Code="InvalidParameter" Message="Parameter 'osProfile' is not allowed."


pearcec:
We made a copy of our existing vhds. We needed to spin them up in a new resource group.  We get an error with osProfile not allowed, but terraform requires it.  If attaching a OSdisk we should probably eliminate the requirement for osProfile.  Might need 

### Terraform Version
Terraform v0.8.2

Your version of Terraform is out of date! The latest version
is 0.8.3. You can update by downloading from www.terraform.io

(Yes it is out of date, but there isn't a Windows build yet. I don't see any references in the CHANGELOG which fix this bug)

### Affected Resource(s)
- azure_virtual_machine

### Terraform Configuration Files
```hcl
resource "azurerm_network_interface" "network" {
    name = "computer-nic"
    location = "East US 2"
    resource_group_name = "${azurerm_resource_group.rg.name}"

    ip_configuration {
        name = "computer-ip"
        subnet_id = "${azurerm_subnet.nonprodnet.id}"
        private_ip_address_allocation = "dynamic"
    }
}

resource "azurerm_virtual_machine" "computer" {
    name = "computer"
    location = "East US 2"
    resource_group_name = "${azurerm_resource_group.rg.name}"
    network_interface_ids = ["${azurerm_network_interface.interface.id}"]
    vm_size = "Standard_DS5_v2"

    storage_os_disk {
        name = "osdisk"
        os_type = "linux"
        vhd_uri = "${azurerm_storage_account.storageaccount.primary_blob_endpoint}vhds/osdisk.vhd"
        caching = "ReadWrite"
        create_option = "attach"
    }

    storage_data_disk {
      name = "disk01"
      vhd_uri = "${azurerm_storage_account.storageaccount.primary_blob_endpoint}vhds/disk01.vhd"
      create_option = "attach"
      disk_size_gb = "512"
      lun = "1"
    }
    storage_data_disk {
      name = "disk02"
      vhd_uri = "${azurerm_storage_account.storargeaccount.primary_blob_endpoint}vhds/disk02.vhd"
      create_option = "attach"
      disk_size_gb = "512"
      lun = "2"
    }

    os_profile {
      computer_name = "computer"
      admin_username = "username"
      admin_password = "password"
    }

     os_profile_linux_config {
       disable_password_authentication = false
       ssh_keys = {
[...]

Issue #13660, opened 2017-04-14
azurerm_storage_blobs can't be attached to vms since v.0.9.3


juls:
Hi there,

since [this commit](https://github.com/hashicorp/terraform/commit/5d915dd4f5da3cf6d41d4e6ffa8e33750bc98c18), a existing _azurerm_storage_blob_ can't be attached to a  _azurerm_virtual_machine_ anymore. It fails because [this test](https://github.com/hashicorp/terraform/blob/6f2551ec33887e730d6dbd4fe0e4794c82ed487d/builtin/providers/azurerm/resource_arm_virtual_machine.go#L1328-L1329) only allows attaching a managed disk.

### Terraform Version
Terraform v0.9.3

### Affected Resource(s)
- azurerm_virtual_machine
- azurerm_storage_blob

### Terraform Configuration Files
Based on the [azurerm_virtual_machine example](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html):

```
resource "azurerm_storage_blob" "test" {
  name                    = "datadisk0.vhd"
  resource_group_name     = "${azurerm_resource_group.test.name}"
  storage_account_name    = "${azurerm_storage_account.test.name}"
  storage_container_name  = "${azurerm_storage_container.test.name}"
  source_uri              = "https://accsa.blob.core.cloudapi.de/images/empty-1024gb.vhd"
  type                    = "page"
  size                    = "1099511627776"
}

resource "azurerm_virtual_machine" "test" {
[...]
  storage_data_disk {
    name          = "datadisk0"
    vhd_uri       = "${azurerm_storage_account.test.primary_blob_endpoint}${azurerm_storage_container.test.name}/datadisk0.vhd"
    disk_size_gb  = "1023"
    create_option = "Attach"
    lun           = 0
  }
[...]
```

### Expected Behavior
An existing image should be attached. This worked fine until Terraform v0.9.2.

### Actual Behavior
Since [this was added](https://github.com/hashicorp/terraform/commit/5d915dd4f5da3cf6d41d4e6ffa8e33750bc98c18#diff-1ad228d05271c678d81bf043a44c12dcR1320), it fails with 
`* azurerm_virtual_machine.test: [ERROR] Must specify which disk to attach`

### Steps to Reproduce
1. Take the [example from the documentation](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html)
2. Add the above mentioned _azurerm_storage_blob_. _source_uri_ can be any existing image.
[...]

PR #14176, opened 2017-05-03
Make os_profile optional #11147

pearcec:
Fix for https://github.com/hashicorp/terraform/issues/11147


builtin/providers/azurerm/resource_arm_virtual_machine.go
```diff
@@ -553,11 +553,13 @@ func resourceArmVirtualMachineCreate(d *schema.ResourceData, meta interface{}) e
...
-	osProfile, err := expandAzureRmVirtualMachineOsProfile(d)
-	if err != nil {
-		return err
+	if _, ok := d.GetOk("os_profile"); ok {
+		osProfile, err := expandAzureRmVirtualMachineOsProfile(d)
+		if err != nil {
+			return err
+		}
+		properties.OsProfile = osProfile
 	}
```
pearcec: I thought about that.  You need an image to attach in order for it to make sense.  That seemed out of scope for testing.  Thoughts?

tombuildsstuff: So I'd suggest the simplest way to achieve this would be to use a multi-phase acceptance test - we're doing this in a few places (for [example here in the AzureRM VM tests](https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_virtual_machine_test.go#L219)).

I believe the following three phases should allow us to reproduce this test-case:

1. Spin up a VM
2. Delete the VM, but keep the Disk (using the same code, without the VM resource)
3. Attach the Disk from the original VM to a new VM, using this config (without `os_profile`)

I'd expect this test might take a little while to run - so it might be worth using a `Standard_F2` size instance, rather than an A0/A1?

pearcec: @tombuildsstuff What is the etiquette for pushing an update to my PR?  Do I push additional commits or should I squash it somehow?  Guidance appreciated.

tombuildsstuff: @pearcec you can push additional commits to the branch and it'll update the PR - don't worry about squashing commits down, we can do that when we merge the PR :)

pearcec:
Getting blocked by #13686

pearcec:
Issue #13660 is blocking.

pearcec:
Doesn't look like the PR for #13686 addresses my blocked issue.  I am simply going to fix the code in this PR as well.

builtin/providers/azurerm/resource_arm_virtual_machine.go
pearcec: Ok @tombuildsstuff I beleive I got the changes in.  Of course I had to add more code.  This is why we test.  Thanks for your guidance.

I ran Standard_F2 which shaved about 2 minutes from the Standard_D1_v2.

tombuildsstuff:
Tests pass (ignoring a subscription-specific failure):

screen shot 2017-06-05 at 09 57 09

tombuildsstuff:
@djsowa this PR was included as a part of [in Terraform v0.9.7](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#097-june-7-2017) - would you be able to upgrade to that release and take a look and see if that works for you?

Since this PR was merged over a year ago - I'm going to lock this issue for the moment; but please feel free to open another issue and we'll take a look :)

Thanks!

Issue #13943, opened 2017-04-25
refresh error on manually-deleted google_compute_instance_group_manager


evralston:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

```
$ terraform --version
Terraform v0.9.3
```

### Affected Resource(s)
Please list the resources as a list, for example:
- google_compute_instance_group_manager

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "google_compute_instance_group_manager" "placeholder_group" {
  name               = "placeholder-group"
  instance_template  = "${google_compute_instance_template.placeholder_template.self_link}"
  base_instance_name = "placeholder"
  zone               = "us-central1-b"
  target_size        = "3"
}
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
What should have happened?

Remove the missing object from the state file.

### Actual Behavior
What actually happened?

```
Error refreshing state: 1 error(s) occurred:

* google_compute_instance_group_manager.placeholder_group: google_compute_instance_group_manager.placeholder_group: googleapi: Error 404: The resource 'projects/project-name/zones/us-central1-b/instanceGroupManagers/placeholder-group' was not found, notFound
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Delete the instance group from the Google console
3. `terraform refresh`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/hashicorp/terraform/issues/496

evralston:
[...]

PR #14190, opened 2017-05-03
provider/google: Move 404 checking into a function in provider.go, call it from instance and IGM

danawillow:
Initial PR for #13943. It fixes the specific concern by adding in 404 checking for `resource_compute_instance_group_manager`, and also shows the new function being reused in `resource_compute_instance` (note that it also exposed a bug- if there is a 404, we should return `nil` instead of an error). If this looks good, I'll check it in and then open up a separate PR for the rest of the resources.

### DIFF
--- a/builtin/providers/google/provider.go
+++ b/builtin/providers/google/provider.go
@@ -252,7 +252,7 @@ func getNetworkNameFromSelfLink(network string) (string, error) {
 	return network, nil
 }
 
-func readError(err error, d *schema.ResourceData, resource string) error {
+func handleNotFoundError(err error, d *schema.ResourceData, resource string) error {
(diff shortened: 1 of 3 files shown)


builtin/providers/google/provider.go
```diff
@@ -250,3 +251,15 @@ func getNetworkNameFromSelfLink(network string) (string, error) {
 
 	return network, nil
 }
+
+func readError(err error, d *schema.ResourceData, resource string) error {
```
paddycarver: nit: maybe `handleNotFoundError`? `readError` doesn't really indicate this is for 404s, to me.

danawillow: Done.

PR #14196, opened 2017-05-04
Fix docs for openstack lbaas v2 resource type names.

philsttr:
The documentation title/sidbar incorrectly referred to resource types `openstack_lbass_*` when the actual type names are `openstack_lb_*`.

The current version of the docs are incorrect.  This doc update does not need to wait for the next release.

### DIFF
--- a/website/source/docs/providers/openstack/r/lb_listener_v2.html.markdown
+++ b/website/source/docs/providers/openstack/r/lb_listener_v2.html.markdown
@@ -1,12 +1,12 @@
 ---
 layout: "openstack"
 page_title: "OpenStack: openstack_lb_listener_v2"
-sidebar_current: "docs-openstack-resource-lbaas-listener-v2"
+sidebar_current: "docs-openstack-resource-lb-listener-v2"
 description: |-
   Manages a V2 listener resource within OpenStack.
 ---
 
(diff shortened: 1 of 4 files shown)


radeksimko:
Hi @philsttr 
thanks for the catch.

philsttr:
It looks like this was merged, but the website still shows `openstack_lbaas_*`.  Does something need to be done to update the website?  Or did I miss a change that needed to be made?

jtopjian:
@philsttr I think doc changes are published upon next releases.

philsttr:
@jtopjian , yeah, there have been two releases since my commit (03c7cfb79997a412443f2df84042da4fa77d6737) was merged:  0.9.5 and 0.9.6.  And the docs were definitely regenerated for 0.9.6, because I see `openstack_dns_recordset_v2` (new in 0.9.6) on the [openstack provider page](https://www.terraform.io/docs/providers/openstack/index.html).

Ah, looking further, looks like i missed [openstack.erb](https://github.com/hashicorp/terraform/blob/1de2cca3da0bfa30e58aae1be2dc7d708f26add1/website/source/layouts/openstack.erb#L139-L150)

philsttr:
Created #15108 to update openstack.erb

Issue #1425, opened 2015-04-08
State file shows initial public_ip and not an eip


alexclifford:
During the initial creation (for example during https://www.terraform.io/intro/getting-started/provision.html) if you have an EC2 instance with an EIP associated to it during apply `terraform show` lists the public_ip as what is presumably the IP address before the EIP is attached. Re-running `terraform refresh` fixes the state file. Should this detail be correct on first orchestration?

robertfirek:
I have similar issue with my outputs. These outputs generate different public_ip during the first apply. The second apply fix that issue and output contains the right values. Adding depends_on doesn't help. 

resource "template_file" "example_template" {
    depends_on = ["aws_eip.some_instance"]
    filename = "template-some_instance"
    vars {
        some_instance_public_ip     = "${aws_instance.some_instance.public_ip}"
    }
} 

output "public_ips" {
    value = "${template_file.example_template.rendered}"
}

phinze:
Hey folks, definitely see the issue here. Terraform is simply reflecting the upstream behavior of the AWS API, which changes the reported value of `PublicIP` after an EIP is attached.

Since the `aws_eip` is attached after the `aws_instance` resource is fully created, Terraform doesn't have any way of knowing that one of those attributes will report differently, nor do we currently have any core facilities for "revisiting" a node after a downstream change has occurred.

I'm wondering if this is a scenario that we can handle with a documentation warning on the field? Something like:

 > If you are using an `aws_eip` with an instance, you should refer to the EIP's address directly and not use public_ip, as it will change after the EIP is attached.

What do you think?

alexclifford:
Hi @phinze,
Thanks for reviewing this and your explanation of how this interacts with the AWS API.
Until Terraform can refresh its state for aws_instance, after an aws_eip action updates it, the documentation workaround seems good to me.

PR #14199, opened 2017-05-04
[WIP] Add support for Task Roles when running on ECS or CodeBuild

jekh:
This PR is intended to address #8746. It adds the `RemoteCredProvider` from the AWS SDK if the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment variable is set. The `RemoteCredProvider` uses the value of that environment variable and appends it to the hard-coded URL `http://169.254.170.2`, and uses the resulting URL to obtain Task Role credentials.

A few questions about the PR:
1. I'm not sure what `Handlers` to provide to the `RemoteCredProvider`, so I went with the SDK defaults. Is this the right pattern to follow for adding this provider?
1. Testing - This is quite difficult to unit test, as the SDK is hard-coded to use 169.254.170.2 for container credentials, making it difficult to mock. Any suggestions?
1. (Somewhat tangential) What's the rationale behind a custom provider chain, rather than allowing the AWS SDK to find credentials using its default search path (at least when credentials aren't specified directly in TF)?

### DIFF
--- a/builtin/providers/aws/auth_helpers.go
+++ b/builtin/providers/aws/auth_helpers.go
@@ -13,6 +13,7 @@ import (
 	awsCredentials "github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/defaults"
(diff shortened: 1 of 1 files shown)


blaines:
Item 3 is a good point because I expect Terraform to behave similarly to a current aws-sdk

jekh:
I'm going to close this PR as I've opened a PR against the AWS provider itself, [#1425](https://github.com/terraform-providers/terraform-provider-aws/pull/1425).

PR #14203, opened 2017-05-04
Handle `google_storage_bucket_object` not being found

dasch:
Mark the resource as no longer available.


stack72:
LGTM! Thanks for this :)

PR #14210, opened 2017-05-04
provider/openstack: Handle disassociating FloatingIP's from a server

### LINKED ISSUE #14209: provider/openstack: Unable to delete floating ip association from deleted server
Not quite sure how I got into this situation, however the result is stuck in a loop whereby I can't delete the resource, and can't apply a new association as the deletion is failing. 

PR incoming to fix. 

### Terraform Version
`Terraform v0.9.5-dev`

### Affected Resource(s)
- openstack_compute_floatingip_associate_v2

2017/05/04 14:41:12 [DEBUG] plugin: terraform:   "nova_error": {
2017/05/04 14:41:12 [DEBUG] plugin: terraform:     "message": "ERROR:Selectable VM host not found. (specified resource(server) not found.)",
2017/05/04 14:41:12 [DEBUG] plugin: terraform:     "request_id": "d0b5be41-465b-424e-b8b2-9c1f4f9aa514"
2017/05/04 14:41:12 [DEBUG] plugin: terraform:   }
2017/05/04 14:41:12 [DEBUG] plugin: terraform: }

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
`openstack_compute_floatingip_associate_v2` resource should be marked as deleted

### Actual Behavior
Resource fails to apply.

fatmcgav:
that's been deleted.

Fixes #14209

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_floatingip_associate_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_floatingip_associate_v2.go
@@ -110,7 +110,7 @@ func resourceComputeFloatingIPAssociateV2Delete(d *schema.ResourceData, meta int
 
 	err = floatingips.DisassociateInstance(computeClient, instanceId, disassociateOpts).ExtractErr()
 	if err != nil {
-		return fmt.Errorf("Error disassociating floating IP: %s", err)
+		return CheckDeleted(d, err, "floating ip association")
 	}
 
 	return nil


fatmcgav:
@jtopjian I think this is the most sensible fix for this scenario, however happy to re-work if you can think of a better option...

jtopjian:
@fatmcgav Agreed. I think you should be able to use the [`CheckDeleted`](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/util.go#L33) here, though. :)

fatmcgav:
Ah, didn't spot that function... Will re-factor :)

fatmcgav:
@jtopjian Refactored to use `CheckDeleted`. Probably worth a squash actually, so I'll do that now...

jtopjian:
Looks good to me! Tests pass, too. Thanks for this 😄

PR #14228, opened 2017-05-04
website: copy vagrant cloud docs

feministy:
**Description**

As part of the docs cleanup, this page needed to be copied over. 

@KFishner any thoughts on the left nav hierarchy for this? I was going to put it under support but it didn't have an index. My next best guess was under FAQ?


KFishner:
nice @feministy !

Moving it under FAQ sounds great rather than having it as a top-level nav

feministy:
@KFishner moved under FAQ 👍  can you approve pls?

KFishner:
@feministy LGTM!

PR #14231, opened 2017-05-05
provider/azurerm: Add example of vm from user image

anniehedgpeth:
## CI: [Build Status](https://travis-ci.org/harijayms/terraform)

This example provisions and deploys the following resources into an Azure subscription:

- Resource Group
- VNET
- Subnet
- NIC
- PIP
- Storage Account
- VM from user image

**CI process:** 
```
terraform get, validate, plan to output, apply output, show
validate with azure CLI
terraform destroy
```


examples/azure-vm-from-user-image/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+  description = "VM name referenced also in storage-related names. This is also used as the label for the Domain Name and to make up the FQDN. If a domain name label is specified, an A DNS record is created for the public IP in the Microsoft Azure DNS system."
+}
+
+variable "admin_username" {
+  description = "administrator user name"
+  default     = "vmadmin"
+}
+
+variable "admin_password" {
+  description = "administrator password (recommended to disable password auth)"
```
tombuildsstuff: > administrator password (recommended to disable password auth)

Is it worth rephrasing this to make it clearer? Something like "The Password for the account specified in the `admin_username` field. We recommend disabling Password Authentication in a Production environment"?

examples/azure-vm-from-user-image/deploy.ci.sh
```diff
@@ -0,0 +1,45 @@
...
+docker run --rm -it \
+  -e ARM_CLIENT_ID \
+  -e ARM_CLIENT_SECRET \
+  -e ARM_SUBSCRIPTION_ID \
+  -e ARM_TENANT_ID \
+  -v $(pwd):/data \
+  --workdir=/data \
+  --entrypoint "/bin/sh" \
+  hashicorp/terraform:light \
+  -c "/bin/terraform destroy -force \
```
tombuildsstuff: I believe we shouldn't need to specify the individual resources as targets here, as deleting the resource group will force the deletion of any child resources?

anniehedgpeth: Hey @tombuildsstuff ,  we did that because we are targeting specific resources for our CI process, because we depend on an image in that resource group, and this helps to ensure that we don't remove it in addition to other mechanisms in place.

stack72:
LGTM thanks for the followup work here :)

Issue #13352, opened 2017-04-05
aws_s3_bucket: replication_configuration shows changes when there are none


PeteGoo:
### Terraform Version
0.8.8
0.9.2

### Affected Resource(s)
- aws_s3_bucket

### Terraform Configuration Files
```hcl
provider "aws" {
  region = "us-east-1"
}

provider "aws" {
  region = "us-west-2"
  alias  = "us-west-2"
}

resource "aws_s3_bucket" "origin-bucket" {
  bucket = "715489234-origin-bucket"
  acl    = "private"

  versioning {
    enabled = true
  }

  replication_configuration {
    role = "${aws_iam_role.replication.arn}"

    rules {
      prefix = ""
      status = "Enabled"

      destination {
        bucket        = "arn:aws:s3:::${aws_s3_bucket.replicated-bucket.id}"
        storage_class = "STANDARD"
      }
    }
  }
}

resource "aws_s3_bucket" "replicated-bucket" {
  bucket   = "715489234-replicated-bucket-us-west-2"
  acl      = "private"
  region   = "us-west-2"
  provider = "aws.us-west-2"

  versioning {
    enabled = true
  }
}

resource "aws_iam_role" "replication" {
  name = "715489234-replication-role"

  assume_role_policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "sts:AssumeRole",
      "Principal": {
        "Service": "s3.amazonaws.com"
      },
      "Effect": "Allow",
      "Sid": ""
    }
  ]
}
POLICY
}

resource "aws_iam_role_policy" "replication" {
  role = "${aws_iam_role.replication.id}"
  name = "715489234-replication-policy"

  policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "s3:GetReplicationConfiguration",
        "s3:ListBucket"
      ],
      "Effect": "Allow",
      "Resource": [
        "arn:aws:s3:::${aws_s3_bucket.origin-bucket.id}"
      ]
    },
    {
      "Action": [
        "s3:GetObjectVersion",
        "s3:GetObjectVersionAcl"
      ],
      "Effect": "Allow",
      "Resource": [
        "arn:aws:s3:::${aws_s3_bucket.origin-bucket.id}/*"
      ]
    },
    {
      "Action": [
        "s3:ReplicateObject",
        "s3:ReplicateDelete"
      ],
      "Effect": "Allow",
[...]

PR #10552, opened 2016-12-06
provider/aws: Added S3 Bucket replication

Ninir:
#### Reasonning
This adds the S3 Bucket replication.

#### Related Issues
Closes https://github.com/hashicorp/terraform/issues/7490

#### Acceptance tests
Sometimes, tests are failing and it is said that `S3Bucket error: BucketRegionError: incorrect region, the bucket is not in 'eu-west-1' region`, whereas buckets are created. I think it is related to the providerFactories stuff, but can't tell without any investigation. 

However, I used this configuration (the one i added to the documentation) to check the compiled binary, and played with it (adding, removing, updating, etc etc.):

```hcl
provider "aws" {
  alias  = "west"
  region = "eu-west-1"
}

provider "aws" {
  alias  = "central"
  region = "eu-central-1"
}

resource "aws_iam_role" "replication" {
  name               = "tf-iam-role-replication-12345"
  assume_role_policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "sts:AssumeRole",
      "Principal": {
        "Service": "s3.amazonaws.com"
      },
      "Effect": "Allow",
      "Sid": ""
    }
  ]
}
POLICY
}

resource "aws_iam_policy" "replication" {
    name = "tf-iam-role-policy-replication-12345"
    policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "s3:GetReplicationConfiguration",
        "s3:ListBucket"
      ],
      "Effect": "Allow",
      "Resource": [
        "${aws_s3_bucket.bucket.arn}"
      ]
    },
    {
      "Action": [
        "s3:GetObjectVersion",
        "s3:GetObjectVersionAcl"
      ],
      "Effect": "Allow",
      "Resource": [
        "${aws_s3_bucket.bucket.arn}/*"
      ]
    },
    {
      "Action": [
        "s3:ReplicateObject",
        "s3:ReplicateDelete"
      ],
      "Effect": "Allow",
      "Resource": "${aws_s3_bucket.destination.arn}/*"
    }
  ]
}
POLICY
}

resource "aws_iam_policy_attachment" "replication" {
    name = "tf-iam-role-attachment-replication-12345"
    roles = ["${aws_iam_role.replication.name}"]
    policy_arn = "${aws_iam_policy.replication.arn}"
}

resource "aws_s3_bucket" "destination" {
    provider = "aws.west"
[...]

PR #14243, opened 2017-05-05
[WIP] provider/aws: better handle S3 replication rules changes

Ninir:
## Description
This is a potential solution to https://github.com/hashicorp/terraform/issues/13352.

The problem is that the ID, when not specified, is generated by AWS. The logic made in the lifecycle configuration generates a prefix

The Rules order doesn't seem to matter, but it seems more logical to me to have a list there.
However, I would add the requirement of the ID, rather than generating something... 

@stack72 @catsby Could i get your thoughts on this please? 

## Tests
Tests may not pass due to a silly bug which I am curently trying to fix in here: https://github.com/hashicorp/terraform/pull/10552#issuecomment-283789618


catsby:
Hey @Ninir I just spent some time on an alt. solution but haven't quite got it yet. Unfortunately I need to switch tasks and can't look again until next week :( 

Quick thoughts:

- `Rules order doesn't seem to matter` it probably doesn't _matter_, but this suggests we may not get them back in the same order, and without a sortable thing to sort them on, we'll get new diffs based on the order they are returned, so I don't think changing to a list here is a good idea
- `rulesHash` func doesn't seem to care about the `destination` (maybe you noticed this too). I support it's removal as you've done here. This may have backwards compatibility issues, but they may not be outside of what we already have here. We need to verify that:
  1) no problem exists _right now_ if people have a rule `id` specified and that
  2) removing the rule hash does not harm them. If it does, and we can't avoid this, we need to note it in the release notes

The part I didn't get to investigate much was making the `id` both `Optional` and `Computed`, maybe you can take that for a spin? Otherwise I'll try to pick it up next week. 

We may want to always generate an `id` if it's not supplied, via `resource.UniqueId()` if no `id` is given

Ninir:
Closing for migration. Need to be more focused for this one.

PR #14250, opened 2017-05-05
Add Provider Guide

sethvargo:
This adds the guides framework as well as an initial "authoring providers" guide.

/cc @stack72 @grubernaut @catsby @paddycarver @randomcamel @burzinpatel

### DIFF
--- a/website/source/guides/index.html.md
+++ b/website/source/guides/index.html.md
@@ -1,0 +1,16 @@
+---
+layout: "guides"
+page_title: "Guides"
+sidebar_current: "guides-home"
+description: |-
+  Welcome to the Terraform guides! The guides provide examples for common
+  Terraform workflows and actions for both beginner and advanced Terraform
+  users.
+---
+
+# Terraform Guides
+
+Welcome to the Terraform guides! If you are just getting started with Terraform,
+please start with the [Terraform introduction](/intro/index.html) instead and
+then continue on to the guides. The guides provide examples for common Terraform
+workflows and actions for both beginner and advanced Terraform users.
--- a/website/source/guides/index.html.md
+++ b/website/source/guides/index.html.md
@@ -10,7 +10,8 @@ description: |-
 
 # Terraform Guides
 
-Welcome to the Terraform guides! If you are just getting started with Terraform,
-please start with the [Terraform introduction](/intro/index.html) instead and
-then continue on to the guides. The guides provide examples for common Terraform
-workflows and actions for both beginner and advanced Terraform users.
+Welcome to the Terraform guides section! If you are just getting started with
+Terraform, please start with the [Terraform introduction](/intro/index.html)
+instead and then continue on to the guides. The guides provide examples for
+common Terraform workflows and actions for both beginner and advanced Terraform
+users.
--- a/website/source/guides/writing-custom-terraform-providers.html.md
+++ b/website/source/guides/writing-custom-terraform-providers.html.md
@@ -1,0 +1,553 @@
+---
+layout: "guides"
+page_title: "Writing Custom Providers - Guides"
+sidebar_current: "guides-writing-custom-terraform-providers"
+description: |-
+  Terraform providers are easy to create and manage. This guide demonstrates
+  authoring a Terraform provider from scratch.
+---
+
+# Writing Custom Providers
+
+~> **This is an advanced guide!** Following this guide is not required for
+regular use of Terraform and is only intended for advance users or Terraform
+contributors.
+
+In Terraform, a "provider" is the logical abstraction of an upstream API. This
+guide details how to build a custom provider for Terraform.
+
+## Why?
+
+There are a few possible reasons for authoring a custom Terraform provider, such
+as:
+
+- An internal private cloud whose functionality is either proprietary or would
+  not benefit the open source community.
+
+- A "work in progress" provider being tested locally before contributing back.
+
+- Extensions of an existing provider
+
+## Local Setup
+
+Terraform supports a plugin model, and all providers and actually plugins.
+Plugins are distributed as Go binaries. Although technically possible to write a
+plugin in another language, almost all Terraform plugins are written in
+[Go](https://golang.org). For more information on installing and configuring Go,
+please visit the [Golang installation guide](https://golang.org/doc/install).
+
+This post assumes familiarity with Golang and basic programming concepts.
+
+As a reminder, all of Terraform's core providers are open source. When stuck or
+looking for examples, please feel free to reference
+[the open source providers](https://github.com/hashicorp/terraform/tree/master/builtin/providers) for help.
+
+## The Provider Schema
+
+To start, create a file named `provider.go`. This is the root of the provider
+and should include the following boilerplate code:
+
+```go
+package main
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func Provider() *schema.Provider {
+	return &schema.Provider{
+		ResourcesMap: map[string]*schema.Resource{},
+	}
+}
+```
+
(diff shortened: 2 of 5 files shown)


website/source/guides/index.html.md
```diff
@@ -0,0 +1,16 @@
+---
+layout: "guides"
+page_title: "Guides"
+sidebar_current: "guides-home"
+description: |-
+  Welcome to the Terraform guides! The guides provide examples for common
```
grubernaut: `the Terraform guides` reads weird here and again below.

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+- An internal private cloud whose functionality is either proprietary or would
+  not benefit the open source community.
+
+- A "work in progress" provider being tested locally before contributing back.
+
+- Extensions of an existing provider
+
+## Local Setup
+
+Terraform supports a plugin model, and all providers and actually plugins.
```
grubernaut: `providers are actually`

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+To verify things are working correctly, execute the binary just created:
+
+```shell
+$ ./terraform-provider-example
+This binary is a plugin. These are not meant to be executed directly.
+Please execute the program that consumes these plugins, which will
+load any plugins automatically
+```
+
+This is the basic project structure and scaffolding for a Terraform plugin.
```
grubernaut: Would be helpful to clarify the project structure and scaffolding with a `tree` or `ls` output.

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
+	}
+}
+
+```
+
+This uses the [`schema.Resource`
```
grubernaut: odd spacing?

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+type](https://godoc.org/github.com/hashicorp/terraform/helper/schema#Resource).
+This structure defines the data schema and CRUD operations for the resource.
+Defining these properties are the only required thing to create a resource.
+
+The schema above defines one element, `"address"`, which is a required string.
+Terraform's schema automatically enforces validation and type casting.
+
+Next there are four "fields" defined - `Create`, `Read`, `Update`, and `Delete`.
+These four functions are required for a resource to be functional. There are
+other functions, but these are the only required ones. Terraform itself handles
```
grubernaut: Update isn't required, if a resource cannot be updated. If that's the case, each schema attribute requires `ForceNew: true`

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+which function to call and with what data. Based on the schema and current state
+of the resource, Terraform can determine whether it needs to create a new
+resource, update an existing one, or destroy.
+
+Each of the four struct fields point to a function. While it is technically
+possible to inline all functions in the resource schema, best practice dictates
+pulling each function into its own method. This optimizes for both testing and
```
grubernaut: Might mention to pull each function into it's own method with the exception of the schema key

website/source/guides/index.html.md
catsby: I think it only reads weird because at time of writing there is only 1. When there are multiple (assuming there will be) it reads fine I think.

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+state. For this example, partial state does not mean much in this documentation
+example. If `updateAddress` were to fail, then the address field would not be
+updated.
+
+## Implementing Destroy
+
+The `Destroy` callback is exactly what it sounds like - it is called to destroy
+the resource. This operation should never update any state on the resource. It
+is not necessary to call `d.SetId("")`, since any non-error return value assumes
+the resource was deleted successfully.
```
grubernaut: might make a note that it is not necessary, but shown here to clearly demonstrate(?)

website/source/guides/index.html.md
grubernaut: Might be better to say `The Terraform guides section!`?

website/source/guides/writing-custom-terraform-providers.html.md
sethvargo: I was going to create a sample GH repo, but voted against it later. Will add output from tree

website/source/guides/writing-custom-terraform-providers.html.md
sethvargo: Noted below

sethvargo: Well, in the comment

website/source/guides/writing-custom-terraform-providers.html.md
```diff
@@ -0,0 +1,553 @@
...
+(upstream). This is called at various points by Terraform and should be a
+read-only operation. This callback should never modify the real resource.
+
+If the ID is updated to blank, this tells Terraform the resource no longer
+exists (maybe it was destroyed out of band). Just like the destroy callback, the
+`Read` function should gracefully handle this case.
+
+```go
+// Tells Terraform the resource no longer exists
+d.SetId("")
```
catsby: we should probably include `return nil` here. AFter we set ID blank we probably don't want to do anything else

catsby: Maybe explain that we need to READ the resource from the API, and update the values, so we detect drift and Terraform will alert. Maybe not need to show code, maybe just link out to an example. One of READs most important parts is that drift detection. 

ALSO, we have an EXISTS method that I think we consider best practice, who's only job is to determine if the resource still exists and if not to remove it with `SetId("")`. I say "consider" best practice because I think we do, but it's not actually widely implemented

catsby: Sorry, to be clear, I realize the `d.SetId("")` is not a comprehensive example, but I think it's important that we add `return nil` to emphasize that in this case, we should just return and the method should do no more.

sethvargo: @catsby I added a READ function example - let me know what you think.

sethvargo:
I definitely agree on adding more advanced examples such as timeouts and waitforstate. I also want to write about statefunc for encrypting and hashing state data.

grubernaut:
@sethvargo `DiffSuppressFunc` is also fairly handy for any provider.

catsby:
> I definitely agree on adding more advanced examples such as timeouts and waitforstate. I also want to write about statefunc for encrypting and hashing state data.

Awesome, yeah those would make good additional guides, outside of this one 👍

bcornils:
how are we doing with this issue? I'm getting feedback that our users are having a hard time figuring out how to create providers and modules. More detailed documentation would help our users.

sethvargo:
It was merged over a year ago... Also, providers are not the same as modules.

bcornils:
@sethvargo  so awesome to hear from you. Yes, I am fully aware there is a difference. We are working on how to potentially make improvements. It appears I grabbed the wrong issue. Thanks for calling that to my attention. Much appreciated.

Thank you very much @burzinpatel. I'll chat with you about some thoughts on the module documents.  I'll set some time for us to chat next week?

Issue #10520, opened 2016-12-04
Behaviour of instance state with schema.TypeSet


tintoy:
Hi.

I've written a couple of providers for Terraform, and there's still one thing I don't understand about the behaviour of `schema.TypeSet`. The hash function produces a value used as the "index" for each item:

```
"set_item.#": "2",
"set_item.3452886897.prop1": "CCC",
"set_item.3452886897.prop2": "DDD",
"set_item.3452886897.prop3": "Foo",
"set_item.3605873202.prop1": "AAA",
"set_item.3605873202.prop2": "BBB",
"set_item.3605873202.prop3": "Foo"
```

So the hash uniquely identifies each element in the set - it's like a key. But the behaviour of the set seems to treat the set elements as simple values (rather than composite ones) even if the data-type if the set elements is a `schema.Resource`. This makes it very difficult to model some types of data in Terraform. For example, disks in a server have a LUN (logical unit Id) which uniquely identifies them. I'd expect to be able to use the LUN as a key to identify items in the set (and ensure their uniqueness). But the way it's all been implemented, sets don't work unless the hash is derived from _all_ properties of the element. And this means that if anything about the element changes, Terraform treats it as a completely different element. This wouldn't be so bad if it wasn't for the fact that this appears to be how Terraform matches up state for elements in the set. Change a property, and the `set_item.3605873202` prefix becomes something else entirely. So how can you calculate or display a useful diff from that?

I'm just wondering if this limitation is by design, or merely incidental. I'm also open to the idea that I've just plain wrong about how this is meant to work :)

### Terraform Version

Terraform v0.7.12

### Affected Resource(s)

Any resource with a schema.TypeSet property.

### Repro

main.go
```go
package main

import (
	"fmt"
	
	"github.com/hashicorp/terraform/helper/schema"
	"github.com/hashicorp/terraform/plugin"
	"github.com/hashicorp/terraform/terraform"
)

// The main program entry-point.
func main() {
	plugin.Serve(&plugin.ServeOpts{
		ProviderFunc: func() terraform.ResourceProvider {
			return &schema.Provider{
				Schema: map[string]*schema.Schema{},
				ResourcesMap: map[string]*schema.Resource{
					"playground_example": &schema.Resource{
						Schema: map[string]*schema.Schema{
							"name": &schema.Schema{
								Type:     schema.TypeString,
								Required: true,
							},
							"list_item": &schema.Schema{
								Type:     schema.TypeList,
								Optional: true,
[...]

PR #2336, opened 2015-06-14
Recognize non-hashed changes to set items

apparentlymart:
(This PR is currently more a conversation-starter than necessarily a correct change.)

While working on #2162 it became apparent that, when using sets of sub-resources, correct diffing is only ensured if all of the element type properties are considered by the hash returned by the ``Set`` function.

It felt to me like this was a bug, for two reasons:
* if the intent was to consider the entire element content in the hash, it wouldn't make sense to have a separate ``Set`` implementation for each set since a single central implementation that iterates (recursively) over all of the element schema properties would be sufficient
* when a set item has a unique identifier element, it seems intuitive that changes to attributes *other* than the unique identifier should appear as single-attribute diffs rather than deleting and re-adding the entire element.

In the OpsWorks Layer resource from #2162, this manifests in the implementation of the ``ebs_volume`` attribute, which looks like this:

```js
resource "aws_opsworks_custom_layer" "foo" {
    // ...various stuff irrelevant to this PR...

    ebs_volume {
        mount_point = "/mnt"
        size = 4
        number_of_disks = 1
    }
}
```

For this resource my change considers each ``ebs_volume`` to be identified by its mount point, so the ``Set`` function returns a hash of the mount point path. I expected that changes to ``mount_point`` would show as deleting one item and adding another, while changes to either ``size`` or ``number_of_disks`` would show as a diff like this:

```
~ aws_opsworks_custom_layer.bar
    ebs_volume.890030859.number_of_disks: "1" => "2"
```

Instead what happens today is that changes to anything other than the mount point are not seen as a diff at all.

Looking at other resource implementations I see that the usual technique is to try to include the whole sub-element in the hash function, but several older bugs in the issue tracker seem to boil down to missing things in such implementations, so it seems like the current situation is error-prone.

The patch associated with this PR takes the position that there should be a separation between an element's identifier (the result of ``Set``) and its entire content, and there should be a visible difference in the diff between changing the value of an element vs. creating a new element depending on which attributes have changed. An advantage of this approach is that adding entirely new attributes to an existing object needn't change the hash.

Another reasonable position would be that the entire content of the element should always be considered in the hash function, in which case I'd expect not to need to provide a ``Set`` implementation at all since it should be possible to provide a fully-general single implementation that would then apply across all sets. In this case, any change to the element schema would cause all items to show as updated in the diff.

The *current* situation doesn't feel right to me, but it's very possible that I'm just missing the motivation behind it.

PR #14261, opened 2017-05-06
[MS] Add encryption_settings to VMs and computed disks (fixes #12259)

### LINKED ISSUE #12259: azurerm support disk encryption
Hi, 

Do you plan to add disk encryption support in the next Terraform version to azurerm provider? 

Thanks, 
Raz.

StephenWeatherford:
@tombuildsstuff I've got some questions regarding this code, please see my comments.  Thanks.

### DIFF
--- a/builtin/providers/azurerm/encryption_settings.go
+++ b/builtin/providers/azurerm/encryption_settings.go
@@ -1,0 +1,163 @@
+package azurerm
+
+import (
+	"github.com/Azure/azure-sdk-for-go/arm/compute"
+	"github.com/Azure/azure-sdk-for-go/arm/disk"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func encryptionSettingsSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"enabled": {
+					Type:     schema.TypeBool,
+					Required: true,
+
+					// Azure can change enabled from false to true, but not the other way around, so
+					//   to keep idempotency, we'll conservatively set this to ForceNew=true
+					// TODO: Is this the right behavior?
+					ForceNew: true,
+				},
+
+				"disk_encryption_key": {
+					Type:     schema.TypeList,
+					Optional: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"secret_url": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+
+							"source_vault_id": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+						},
+					},
+				},
+				"key_encryption_key": {
+					Type:     schema.TypeList,
+					Optional: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"key_url": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+
+							"source_vault_id": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func flattenVmDiskEncryptionSettings(encryptionSettings *compute.DiskEncryptionSettings) map[string]interface{} {
+	return map[string]interface{}{
+		"enabled": *encryptionSettings.Enabled,
+		"disk_encryption_key": []interface{}{
+			map[string]interface{}{
+				"secret_url":      *encryptionSettings.DiskEncryptionKey.SecretURL,
+				"source_vault_id": *encryptionSettings.DiskEncryptionKey.SourceVault.ID,
+			},
+		},
+		"key_encryption_key": []interface{}{
+			map[string]interface{}{
+				"key_url":         *encryptionSettings.KeyEncryptionKey.KeyURL,
+				"source_vault_id": *encryptionSettings.KeyEncryptionKey.SourceVault.ID,
+			},
+		},
+	}
+}
+
+func flattenManagedDiskEncryptionSettings(encryptionSettings *disk.EncryptionSettings) map[string]interface{} {
+	return map[string]interface{}{
+		"enabled": *encryptionSettings.Enabled,
+		"disk_encryption_key": []interface{}{
+			map[string]interface{}{
+				"secret_url":      *encryptionSettings.DiskEncryptionKey.SecretURL,
+				"source_vault_id": *encryptionSettings.DiskEncryptionKey.SourceVault.ID,
+			},
+		},
+		"key_encryption_key": []interface{}{
+			map[string]interface{}{
+				"key_url":         *encryptionSettings.KeyEncryptionKey.KeyURL,
+				"source_vault_id": *encryptionSettings.KeyEncryptionKey.SourceVault.ID,
+			},
+		},
+	}
+}
+
+func expandVmDiskEncryptionSettings(settings map[string]interface{}) *compute.DiskEncryptionSettings {
+	enabled := settings["enabled"].(bool)
+	config := &compute.DiskEncryptionSettings{
+		Enabled: &enabled,
+	}
+
+	if v := settings["disk_encryption_key"].([]interface{}); len(v) > 0 {
+		dek := v[0].(map[string]interface{})
+
+		secretURL := dek["secret_url"].(string)
+		sourceVaultId := dek["source_vault_id"].(string)
+		config.DiskEncryptionKey = &compute.KeyVaultSecretReference{
+			SecretURL:   &secretURL,
+			SourceVault: &compute.SubResource{ID: &sourceVaultId},
+		}
+	}
+
+	if v := settings["key_encryption_key"].([]interface{}); len(v) > 0 {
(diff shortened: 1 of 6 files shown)


builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
+package azurerm
```
StephenWeatherford: Currently there are no acceptance tests, because getting an end to end to work via automation is a lot of work and would also take a long time to run due to encryption, or else would require having a pre-encrypted OS image that is hosted somewhere permanent (and still be slow to copy).

What do you guys do in this situation?

builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
...
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"enabled": {
+					Type:     schema.TypeBool,
+					Required: true,
+
+					// Azure can change enabled from false to true, but not the other way around, so
+					//   to keep idempotency, we'll conservatively set this to ForceNew=true
+					// TODO: Is this the right behavior?
+					ForceNew: true,
```
StephenWeatherford: We're seeing this in a few places.  Is there a way to force new only under some circumstances?  If not, should we be conservative and set ForceNew:true?

builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
...
+				},
+
+				"disk_encryption_key": {
+					Type:     schema.TypeList,
+					Optional: true,
```
StephenWeatherford: The current schema is:
  encryption_settings {
    enabled = true
    disk_encryption_key { ... }
    key_encryption_key { ... }
  }

If enabled=true at least DEK must be supplied, but I can't mark it required because if enabled=false, DEK and KEK are not allowed to be set by Azure.  Would it be better to not expose enabled and just infer it as true if encryption_settings is supposed?

builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
...
+					Required: true,
+
+					// Azure can change enabled from false to true, but not the other way around, so
+					//   to keep idempotency, we'll conservatively set this to ForceNew=true
+					// TODO: Is this the right behavior?
+					ForceNew: true,
+				},
+
+				"disk_encryption_key": {
+					Type:     schema.TypeList,
```
StephenWeatherford: Okay, here's where things get unnecessarily complex.  I'm mimicking Azure's ARM schema here:

```hcl
  profile_os_disk {
    encryption_settings {
      enabled = true
      disk_encryption_key { ... }
      key_encryption_key { ... }
    }
  }
```
There's only ever one encryption_settings, but I can't see a good way to model this kind of nested object schema, except to use a set or list with MaxElem=1.  Using set seems like a really bad idea because of its weird semantics and bad diff'ing, so using a list seems the only available choice.

I could flatten this out to remove the nested objects, although that seems messy, too.

Advice?

helper/schema/schema.go
```diff
@@ -145,7 +145,12 @@ type Schema struct {
 	// The following fields are only valid for a TypeSet type.
 	//
 	// Set defines a function to determine the unique ID of an item so that
-	// a proper set can be built.
```
StephenWeatherford: Please correct if these comments are wrong, but they seem necessary considering https://github.com/hashicorp/terraform/issues/10520, https://github.com/hashicorp/terraform/pull/2336 and https://groups.google.com/forum/#!topic/terraform-tool/tsnWzWfkzaQ.  

Took a long time for me to understand what was going on.

Which brings up another issue.  We've got a lot of properties in azurerm (e.g. VMs/VMSS) that get ignored because they're in sets that define a set function that includes only certain elements.  I don't see any way to fix this, because that would modify the hash and cause everybody's resources to be updated.

Part of the problem is the documentation here, which to me implies that an "ID" is *not* a hash, but rather a way for hashicorp to keep track of a particular object in a set with its cloud counterpart, and thus be able to detect when non-identity properties are changed, and make them to the appropriate cloud resource.

Instead, it appears that the set function is solely a hash code, which has quite a few disadvantages:
1) Any change in the properties acts as if the entire set has changed (and if any of the properties in the set are ForceNew=true, then the entire resource gets re-created)
2) The diff shown to the user by this is way more complex than it needs to be and is being confusing
3) There's not a way to modify just certain properties of that object
4) There's a lot of code at least in azurerm that has incorrect set functions and thus properties that get ignored, with no clear way to fix these bugs

Seems like this design may need review.

The design mentioned in

builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
+package azurerm
+
+import (
+	"github.com/Azure/azure-sdk-for-go/arm/compute"
+	"github.com/Azure/azure-sdk-for-go/arm/disk"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func encryptionSettingsSchema() *schema.Schema {
```
StephenWeatherford: It's shared for both os disk and managed disks, which is why I pulled it out into its own separate file.  (If they diverge later we can diverge the code, but IMHO there's already too much duplicated code, esp. between VM and VMSS.

builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func flattenVmDiskEncryptionSettings(encryptionSettings *compute.DiskEncryptionSettings) map[string]interface{} {
```
StephenWeatherford: I think having them in a separate file still has advantages - a) I can at least re-use the schema, b) it's much more likely that someone fixing a bug in encryption settings will notice s/he has to fix it in managed disks and VMs.  I've already seen a case where stuff was fixed in VMs but the corresponding code in VMSSs was untouched.

builtin/providers/azurerm/encryption_settings.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+func expandVmDiskEncryptionSettings(settings map[string]interface{}) *compute.DiskEncryptionSettings {
+	enabled := settings["enabled"].(bool)
+	config := &compute.DiskEncryptionSettings{
+		Enabled: &enabled,
+	}
+
+	if v := settings["disk_encryption_key"].([]interface{}); len(v) > 0 {
```
StephenWeatherford: It appears that a list is always present, even if empty.  If I change it to set, then I could make that change.

StephenWeatherford: Ditto for the other two

builtin/providers/azurerm/resource_arm_virtual_machine.go
```diff
@@ -854,6 +857,10 @@ func resourceArmVirtualMachineStorageOsDiskHash(v interface{}) int {
 	if m["vhd_uri"] != nil {
 		buf.WriteString(fmt.Sprintf("%s-", m["vhd_uri"].(string)))
 	}
+	if encryptionSettings := m["encryption_settings"]; encryptionSettings != nil {
+		var encryptionHash = schema.HashSchema(encryptionSettingsSchema())(encryptionSettings)
```
StephenWeatherford: It doesn't have a hash function, on purpose (going by the guidance of not using one for new code, to avoid the issues of when people add members but forget to update the code).

website/source/docs/providers/azurerm/r/virtual_machine.html.markdown
```diff
@@ -345,11 +345,26 @@ For more information on the different example configurations, please check out t
...
+`encryption_settings` supports the following:
+
+* `enabled` - (Required) Specifies if the encryption is enabled. If false, disk_encryption_key and key_encryption_key must not be specified.
+* `disk_encryption_key` - (Optional) Specifies the location of the disk encryption key using `secret_url` and `source_vault_id` subproperties. Must be specified if enabled is true.
+* `key_encryption_key` - (Optional) Specifies the location of the key encryption key using `key_url` and `source_vault_id` subproperties.
+
+~> **Note:** An example `disk_encryption_key` could look like:
+```hcl
+disk_encryption_key {
+  secret_url = "https://{keyvaultname}.vault.azure.net/secrets/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```
StephenWeatherford: Changed to 0's

builtin/providers/azurerm/encryption_settings.go
StephenWeatherford: @tombuildsstuff, @whiskeyjay

I'm not sure that using a set over a list is worth it for consistency reasons, given all its disadvantages (see https://github.com/hashicorp/terraform/pull/14261).  Is there any other reason not to go with a list?  Thx.

I also notice that azurerm_redis_cache is using a list for this purpose (redis_configuration).

builtin/providers/azurerm/resource_arm_virtual_machine.go
StephenWeatherford: See https://github.com/hashicorp/terraform/issues/15420

tombuildsstuff:
Closing the original PR since this has been migrated over to the new repo: https://github.com/terraform-providers/terraform-provider-azurerm/pull/157

PR #14262, opened 2017-05-06
 #9419 - Feature request: Support for ElastiCache Redis cluster mode

dpetzold:
Feature request: Support for ElastiCache Redis cluster mode #9419

Added support for provisioning a native redis cluster elasticache replication group.
A new TypeSet attribute `cluster_mode` has been added. It requires the following
fields:

  - `replicas_per_node_group` - The number of replica nodes in each node group
  - `num_node_groups` - The number of node groups for this Redis replication group

Notes:

  - `automatic_failover_enabled` must be set to true.
  - `number_cache_clusters` is now a optional or a computed field. If `cluster_mode` is set
    its value will be calculated as:
      ```num_node_groups + num_node_groups * replicas_per_node_group```

Below is a sample config:

```hcl
  resource "aws_elasticache_replication_group" "bar" {
      replication_group_id = "tf-redis-cluser"
      replication_group_description = "test description"
      node_type = "cache.t2.micro"
      port = 6379
      parameter_group_name = "default.redis3.2.cluster.on"
      automatic_failover_enabled = true
      cluster_mode {
          replicas_per_node_group = 1
          num_node_groups = 2
      }
  }
```
Caveats:

`number_cache_clusters` is now a optional or a computed field. I am not sure how out of bounds that is.

### DIFF
--- a/builtin/providers/aws/resource_aws_elasticache_replication_group.go
+++ b/builtin/providers/aws/resource_aws_elasticache_replication_group.go
@@ -44,7 +44,8 @@ func resourceAwsElasticacheReplicationGroup() *schema.Resource {
 
 	resourceSchema["number_cache_clusters"] = &schema.Schema{
 		Type:     schema.TypeInt,
-		Required: true,
+		Computed: true,
+		Optional: true,
 		ForceNew: true,
 	}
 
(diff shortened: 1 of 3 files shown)


stack72:
HI @dpetzold 

So I made a slight design change to this. Basically, I moved the else to be a first class check to make sure we have set either num_cach_nodes or cluster_mode. I was able to write a test for this as well

I just opened PR #14317 for this - I didn't think it was fair to go back to you and ask you to make more changes considering you have been waiting on this for a while

As you can see, i preserved your work in that PR as well :)

Thanks for the work here! It works very well :)

@stack72

Issue #9672, opened 2016-10-27
google_compute_target_https_proxy does not handle certs correctly


jasonkeene:
### Terraform Version

```
$ terraform version
Terraform v0.7.7
```
### Affected Resource
- google_compute_target_https_proxy
### Terraform Configuration Files

vars.tf:

``` hcl
variable "concourse-cert" {
  type = "string"
}
```

terraform.tfvars:

``` hcl
concourse-cert = "projects/cff-loggregator/global/sslCertificates/wildcard-ci"
```

concourse.tf:

``` hcl
resource "google_compute_instance_group" "concourse-1" {
  name        = "concourse-${var.zone-1}-1"
  zone        = "${var.zone-1}"
  named_port {
    name = "http"
    port = "8080"
  }
  instances = "${var.concourse-web-instance-urls}"
}

resource "google_compute_backend_service" "concourse" {
  name        = "concourse"
  port_name   = "http"
  protocol    = "HTTP"
  timeout_sec = 10
  enable_cdn  = false

  backend {
    group = "${google_compute_instance_group.concourse-1.self_link}"
  }

  health_checks = ["${google_compute_http_health_check.concourse.self_link}"]
}

resource "google_compute_http_health_check" "concourse" {
  name               = "concourse"
  request_path       = "/login"
  check_interval_sec = 1
  timeout_sec        = 1
  port               = 8080
}

resource "google_compute_target_https_proxy" "concourse" {
  name             = "concourse"
  url_map          = "${google_compute_url_map.concourse.self_link}"
  ssl_certificates = ["${var.concourse-cert}"]
}

resource "google_compute_url_map" "concourse" {
  name        = "concourse"

  default_service = "${google_compute_backend_service.concourse.self_link}"
}
```
### Debug Output

https://gist.github.com/jasonkeene/9a4c5a6f0c2032bf5471ed44742141c2
### Expected Behavior

An apply that contains a `google_compute_target_https_proxy` resource with certs referenced should be idempotent across multiple applys.
### Actual Behavior

With the first run, terraform creates the `google_compute_target_https_proxy` referencing an existing cert. The tfstate file does not contain the reference to the cert. When I run `tf apply` it attempts to add the cert to the `google_compute_target_https_proxy` again and errors out.
### Steps to Reproduce
1. `terraform apply`
2. `terraform plan # see the bad plan`
3. `terraform apply # get the error from gcp api`

cblecker:
[...]

PR #14264, opened 2017-05-06
provider/google: Improved SSL certs handling on target_https_proxy

tmshn:
### Problem
As reported in #9672, passing a partial URL of SSL cert (e.g.: `projects/foo/global/sslCertificates/bar`) to `target_https_proxy` will not be idempotent; the passed URL will not be recorded in state file and Terraform will try to add the same certs on next apply even though it's actually already set in the real resource.

### Solution
This PR fixes this problem by converting partial URL to full URL every time before SSL certs comparison and API call.


builtin/providers/google/provider.go
```diff
@@ -264,3 +265,63 @@ func handleNotFoundError(err error, d *schema.ResourceData, resource string) err
...
+	apiVersion   string
+	project      string
+	region       string
+	zone         string
+	resourceType string
+	name         string
+	url          string
+}
+
+func parseUrl(url string) (resourceInfo, error) {
```
tmshn: This function is kinda over-spec only for this purpose, but hope it will be useful on later usage.

builtin/providers/google/resource_compute_target_https_proxy_test.go
```diff
@@ -7,46 +7,73 @@ import (
...
+	var resourceName = fmt.Sprintf("httpsproxy-test-%s", acctest.RandString(10))
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckComputeTargetHttpsProxyDestroy,
 		Steps: []resource.TestStep{
 			resource.TestStep{
-				Config: testAccComputeTargetHttpsProxy_basic1,
+				Config: testAccComputeTargetHttpsProxy_basic(resourceName, 0, 0),
```
tmshn: I totally rewrote this `TestAccComputeTargetHttpsProxy_update` since the old one was insufficient: it used to create a set of resources, then destroy them and recreate a new set of resources -- this is not a test of update!

Sorry for the ugly diff.

builtin/providers/google/provider.go
paddycarver: So after discussing with @danawillow, I think we'd rather a more specific, less-encompassing function that just takes a URL (full or partial) for specifically the certs, and always returns the full URL. Something like [what we do for disk images](https://github.com/hashicorp/terraform/blob/a47d0544a996e0f4f7ec648ab68943ed242d258a/builtin/providers/google/image.go#L70), though hopefully more minimal. The reason we think it'd be better to do a more specific version is because our resources tend to have shorthand inputs (like `family/image` or just `image` for images) to make the UX more intuitive and easier to work with. I also worry that the regular expression is descriptive, not prescribed, and so it may not be followed for future resources, which would lead to some maintenance headaches. :( So for the time being, I think specific solutions are probably better than generic ones. But thank you for the hard work on this! It definitely lays the groundwork if we want to move in this direction in the future.

tmshn:
@paddycarver Sorry for the late reply, I was busy last week...

- About the too "encompassing" `parseUrl`:
    - Fixed in c24ec09.
- About utilizing `StateFunc`:
    - Ok, I'm happy to do the fix, but want some guidance. Could you please give me ?

tmshn:
@paddycarver ping

paddycarver:
Hey @tmshn, my apologies for dropping the ball on this. It looks like terraform-providers/terraform-provider-google#210 fixed this, so it should work now. If you continue to have problems, please let us know. And please do accept my sincerest apologies for this going unanswered; we're currently working to make sure this doesn't happen in the future. Thank you for your work and patience!

tmshn:
@paddycarver I understand the situation, no problem! And happy to hear the issue is fixed, thanks! :+1:

PR #14271, opened 2017-05-07
Fix data dns txt record set

sbfaulkner:
### Problem

When attempting to use the DNS provider's `data "dns_txt_record_set" …` I was not able to retrieve the TXT records.

#### Input
```
data "dns_txt_record_set" "example" {
  host = "example.com"
}

resource "local_file" "record" {
  content  = "${data.dns_txt_record_set.example.record}"
  filename = "${path.module}/record.txt"
}
```

#### Result
```
Error applying plan:

1 error(s) occurred:

* local_file.record: Resource 'data.dns_txt_record_set.example' does not have attribute 'record' for variable 'data.dns_txt_record_set.example.record'
```

Attempts to access the `id` and `records` attributes failed as well.

### Solution
1. Updated the A and CNAME record set tests to verify that the `id` attribute is correctly set for those (it was).
2. Added test for the DNS record set.
3. Updated the `testCheckAttrStringArray` to not care about order (since DNS record order is not predictable).
4. Fixed DNS record set by adding missing call to `d.SetId(host)`

### DIFF
--- a/builtin/providers/dns/data_dns_a_record_set_test.go
+++ b/builtin/providers/dns/data_dns_a_record_set_test.go
@@ -12,6 +12,7 @@ func TestAccDataDnsARecordSet_Basic(t *testing.T) {
 		DataSourceBlock string
 		DataSourceName  string
 		Expected        []string
+		Host            string
 	}{
 		{
 			`
@@ -23,6 +24,7 @@ func TestAccDataDnsARecordSet_Basic(t *testing.T) {
 			[]string{
 				"127.0.0.1",
 			},
+			"127.0.0.1.nip.io",
 		},
 		{
 			`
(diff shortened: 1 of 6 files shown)


sbfaulkner:
fwiw... the actual fix was this line... https://github.com/sbfaulkner/terraform/blob/a67d487aa559f400988794ff3d6572da6bbecad3/builtin/providers/dns/data_dns_txt_record_set.go#L49

sbfaulkner:
@grubernaut let me know if there's anything I need to do to clean this up or expedite getting it merged... cheers

builtin/providers/dns/test_check_attr_string_array.go
```diff
@@ -27,24 +27,27 @@ func testCheckAttrStringArray(name, key string, value []string) r.TestCheckFunc
...
 
-		got, _ := strconv.Atoi(count)
-		if got != len(value) {
+		gotCount, _ := strconv.Atoi(count)
+		if gotCount != len(value) {
 			return fmt.Errorf("Mismatch array count for %s: got %s, wanted %d", key, count, len(value))
 		}
 
-		for i, want := range value {
+	Next:
```
grubernaut: Should this be here?

builtin/providers/dns/data_dns_a_record_set_test.go
```diff
@@ -34,23 +36,30 @@ func TestAccDataDnsARecordSet_Basic(t *testing.T) {
...
-		ts := resource.TestStep{
-			Config: test.DataSourceBlock,
-			Check: resource.ComposeTestCheckFunc(
-				testCheckAttrStringArray(fmt.Sprintf("data.dns_a_record_set.%s", test.DataSourceName), "addrs", test.Expected),
-			),
-		}
-		steps = append(steps, ts)
+		recordName := fmt.Sprintf("data.dns_a_record_set.%s", test.DataSourceName)
+
+		resource.UnitTest(t, resource.TestCase{
```
grubernaut: Curious if/why this was changed from an acceptance test to a unit test.

builtin/providers/dns/test_check_attr_string_array.go
sbfaulkner: yes, see the `continue Next` on line 44

builtin/providers/dns/data_dns_a_record_set_test.go
sbfaulkner: 1. didn't make sense to me that the data_cname_record_set test was a unit test, and this was an acceptance test
2. I can change it back, but it didn't seem to run when I ran `make test TEST=./builtin/providers/dns` ... may be I'm just missing something for how to run tests - I'm new to this world of go

sbfaulkner: I just looked into it further, and I see now that UnitTest is just a wrapper around the Test to allow some Acceptance tests to run with Unit tests.

So, it is still an acceptance test, but by using this method, the three tests will run with the unit tests as well (which I think should be ok, unless we view them as having external dependencies due to the use of DNS).

wdyt?

grubernaut: Gotcha, yeah that's okay with me. 👍

builtin/providers/dns/test_check_attr_string_array.go
grubernaut: Aye, should have clarified, sorry! 
Functionally, this looks good to me, however I feel it's slightly over complex for a simple test function. Easy to miss (as I did above), the labeled statement on first read. That being said, however, it's not critical enough to request another change on it 😄 
Would gladly accept a future PR to refactor this test function a bit, though.

PR #14276, opened 2017-05-08
Grammar correction in alarm description

RussVanBert:
It was updated in a few other places but not all.


stack72:
Hi @RussVanBert 

Thanks for finding these errors! Looks good to me :)

Paul

PR #14280, opened 2017-05-08
providers/google: add google_container_versions data source.

paddycarver:
Add a data source for listing available versions for Container Engine
clusters or retrieving the latest available version.

This is mostly to support our tests for specifying a version for cluster
creation; the withVersion test has been updated to use the data source,
meaning it will stop failing on us as new versions get released.

### DIFF
--- a/builtin/providers/google/data_source_google_container_engine_versions.go
+++ b/builtin/providers/google/data_source_google_container_engine_versions.go
@@ -7,9 +7,9 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-func dataSourceGoogleContainerVersions() *schema.Resource {
+func dataSourceGoogleContainerEngineVersions() *schema.Resource {
 	return &schema.Resource{
-		Read: dataSourceGoogleContainerVersionsRead,
+		Read: dataSourceGoogleContainerEngineVersionsRead,
 		Schema: map[string]*schema.Schema{
 			"project": {
 				Type:     schema.TypeString,
@@ -41,7 +41,7 @@ func dataSourceGoogleContainerVersions() *schema.Resource {
 	}
 }
 
-func dataSourceGoogleContainerVersionsRead(d *schema.ResourceData, meta interface{}) error {
+func dataSourceGoogleContainerEngineVersionsRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
 
 	project, err := getProject(d, config)
--- a/website/source/docs/providers/google/d/google_container_engine_versions.html.markdown
+++ b/website/source/docs/providers/google/d/google_container_engine_versions.html.markdown
@@ -1,6 +1,6 @@
 ---
 layout: "google"
-page_title: "Google: google_container_versions"
+page_title: "Google: google_container_engine_versions"
 sidebar_current: "docs-google-datasource-container-versions"
 description: |-
   Provides lists of available Google Container versions for masters and nodes.
@@ -11,14 +11,14 @@ description: |-
 Provides access to available Google Container versions in a zone for a given project.
 
 ```hcl
-data "google_container_versions" "central1b" {
+data "google_container_engine_versions" "central1b" {
   zone = "us-central1-b"
 }
 
 resource "google_container_cluster" "foo" {
   name               = "terraform-test-cluster"
   zone               = "us-central1-b"
-  node_version       = "${data.google_container_versions.central1b.latest_node_version}"
+  node_version       = "${data.google_container_engine_versions.central1b.latest_node_version}"
   initial_node_count = 1
 
   master_auth {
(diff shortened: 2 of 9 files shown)


radeksimko:
Functionally this looks overall good, but I'd think about better name for this data source.

Firstly Google Container Versions almost sounds like I'd be pulling versions/layers of the docker container running there - or at the very least it sounds a bit confusing to me.

Secondly we might decide to expose more attributes in the future from the same API, e.g. image types: https://cloud.google.com/container-engine/reference/rest/v1/projects.zones/getServerconfig

How would you feel about calling it `google_container_server_config`?

paddycarver:
Hey @radeksimko, thanks for the review. I thought about that, but had some reservations. Interested in your feedback on which approach you prefer, and @danawillow's thoughts.

* We already have trouble with users not realising we support something because we matched the resource names to the API names, not to the feature names users will likely look for. In this case, if we wanted to support more properties from that endpoint, I'd break them out into their own data sources, probably.
* I think, pedantically, there'd be some confusion as to whether it was `google_container_serverconfig` or `google_container_server_config` based on the way the API is named.
* I was a bit worried this would be confused for the ConfigMap of a server.

But you also have solid points. With the above in mind, do you still prefer the `google_container_server_config` name?

danawillow:
My biggest concern is people realizing that the data source is for their container engine setup as a whole, rather than for an individual container cluster (even though it doesn't have cluster in the name). I wonder whether just putting `engine` in the names would make that more clear- `google_container_engine_versions` or `google_container_engine_config`.

As for whether to have versions separately or with other config elements, I don't have a strong opinion, but I do agree with @paddycarver's point that users don't always find what they're looking for within the google provider because we map to API names, so having `version` be right there in the title is nice from a Ctrl-F standpoint.

radeksimko:
I see what do you mean about the discoverability.

That's a good point, although there's a thin line between what user expects vs what's Google's intention. Theoretically having names match the schema is a benefit for folks coming from other tools which do that too (e.g. gcloud). Practically though discoverability from user's POV is _probably_ little bit more important, so 👌 

`google_container_engine_versions` still looks better than the current name.

paddycarver:
Let's stick with `google_container_engine_version` for the moment, for `Ctrl+F`-ability. We can always introduce a `google_container_engine_server_config` to match the gcloud API if we want to, later, and deprecate this one.

paddycarver:
This is ready for review and merge. :)

builtin/providers/google/data_source_google_container_engine_versions.go
```diff
@@ -0,0 +1,67 @@
...
+	zone := d.Get("zone").(string)
+
+	resp, err := config.clientContainer.Projects.Zones.GetServerconfig(project, zone).Do()
+	if err != nil {
+		return fmt.Errorf("Error retrieving available container cluster versions: %s", err.Error())
+	}
+
+	d.Set("valid_master_versions", resp.ValidMasterVersions)
+	d.Set("valid_node_versions", resp.ValidNodeVersions)
+	d.Set("latest_master_version", resp.ValidMasterVersions[0])
```
danawillow: Do we know for sure that they'll come back in order?

website/source/docs/providers/google/d/google_container_engine_versions.html.markdown
```diff
@@ -0,0 +1,46 @@
+---
+layout: "google"
+page_title: "Google: google_container_engine_versions"
+sidebar_current: "docs-google-datasource-container-versions"
+description: |-
+  Provides lists of available Google Container versions for masters and nodes.
```
danawillow: s/Container/Container Engine/

website/source/docs/providers/google/d/google_container_engine_versions.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+---
+
+# google\_container\_versions
+
+Provides access to available Google Container versions in a zone for a given project.
```
danawillow: and here

website/source/docs/providers/google/d/google_container_engine_versions.html.markdown
```diff
@@ -0,0 +1,46 @@
+---
+layout: "google"
+page_title: "Google: google_container_engine_versions"
+sidebar_current: "docs-google-datasource-container-versions"
+description: |-
+  Provides lists of available Google Container versions for masters and nodes.
+---
+
+# google\_container\_versions
```
danawillow: and here

builtin/providers/google/data_source_google_container_engine_versions.go
paddycarver: For the moment, they do. I didn't want to try to write a parser for the versions to sort them, especially since I'm not sure if a spec for the versions exists.

PR #14282, opened 2017-05-08
provider/google: fix Service Account provider .Read

q3k:
The implementation would return an error if the resource was detected as
removed - this would break Terraform during plan/apply instead of making it re-create the
missing service account.


catsby:
Thanks!

Issue #14129, opened 2017-05-01
RDS ModifyOptionGroup


vincentdasari:
Hi there,

I'm trying to add a tag to an existing OptionGroup and terraform throws out the following error- "Error modifying DB Option Group: InvalidParameterValue: At least one option must be added, modified, or removed."

I'm guessing that ModifyOptionGroup does not have the ability to add a tag for an existing option group and the same options or configuration for create option group has been added.

Affected resources-
aws_db_option_group

gazoakley:
The existing code checks if there are options to add/remove here:

https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_option_group.go#L265

`ModifyOptionGroup` is called without tag changes (it doesn't support modifying tags - that's done later presumably through AddTagsToResource/RemoveTagsFromResource):

https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_db_option_group.go#L274

The above could probably just be wrapped in a condition so we can skip to calling `setTagsRDS` if there's nothing to do

stack72:
Hi @vincentdasari 

Thanks for the bug report here - we think this may have expected an issue in Terraform core - we are looking into this right now

Thanks

Paul

PR #14283, opened 2017-05-08
helper/schema: Failing test for nested sets and HasChange

stack72:
As pointed out in #14129, I found that d.HasChange was returning a
reported change for a Set with Nested sets. This shouldn't have been the
case.

Thanks to @radeksimko, I have been able to add 2 new tests to
resource_data_test.go:

1. For nested sets - which fails as expected to prove the issue
2. Set with nested lists - which passes

@jbardin / @apparentlymart would love your thoughts on this please :)

### DIFF
--- a/helper/schema/resource_data_test.go
+++ b/helper/schema/resource_data_test.go
@@ -1307,19 +1307,207 @@ func TestResourceDataHasChange(t *testing.T) {
 
 			Change: false,
 		},
+
+		// Nested Sets
+		{
+			Schema: map[string]*Schema{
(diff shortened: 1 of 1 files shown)


radeksimko:
I think that [`Set.Equal`](https://github.com/hashicorp/terraform/blob/master/helper/schema/set.go#L147-L154) will need to be modified to support comparison of nested sets, i.e. we'll probably need to go through the whole map and turn `*Set` into `[]interface{}` so that we can compare the parent map easily via `reflect.DeepEqual`.

Issue #13423, opened 2017-04-06
remote-exec too verbose


dvianello:
### Terraform Version
Terraform v0.9.2

### Affected Resource(s)
- remote-exec

### Terraform Configuration Files
```hcl
resource "openstack_compute_instance_v2" "instance" {
  count       = "${var.instances}"
  name        = "${var.name}-instance-${count.index + 1}"
  image_id    = "${var.disk_image}"
  flavor_name = "${var.machine_type}"
  key_pair    = "${var.key_name}"

  network {
    name           = "${var.net}"
    access_network = true
  }

  security_groups = [
    "${openstack_compute_secgroup_v2.allow-ssh.name}",
  ]
}

resource "openstack_compute_floatingip_associate_v2" "floatingip_assoc" {
  floating_ip = "${openstack_networking_floatingip_v2.floatingip.address}"
  instance_id = "${openstack_compute_instance_v2.instance.id}"
}

resource "null_resource" "provision" {
  depends_on = ["openstack_compute_floatingip_associate_v2.floatingip_assoc"]

  provisioner "remote-exec" {
    connection {
      user = "core"
      host = "${openstack_networking_floatingip_v2.floatingip.address}"
    }

    inline = [
      "docker run -d ${var.docker_ports} ${var.docker_repository}",
      "echo terraform executed > /tmp/foo",
    ]
  }
}
```

### Expected Behavior
TF executes the provisioner, reporting on screen what's going on.

### Actual Behavior
TF executes the provisioner, but outputs an abnormal amount of 
`[0m[0mopenstack_compute_instance_v2.instance (remote-exec): Connecting to remote host via SSH...
openstack_compute_instance_v2.instance (remote-exec):   Host: 193.62.52.106
openstack_compute_instance_v2.instance (remote-exec):   User: core
openstack_compute_instance_v2.instance (remote-exec):   Password: false
openstack_compute_instance_v2.instance (remote-exec):   Private key: false
openstack_compute_instance_v2.instance (remote-exec):   SSH Agent: true`

while waiting on SSH to become available. Piping the logs on a file we ended up with 25MB of logs to provision a single instance. Same provisioning with TF `0.7.13` writes ~500KB. 

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

dvianello:
Nobody with a bit of love for this issue? 👍

jrandall:
[...]

PR #14291, opened 2017-05-08
add backoff to remote-exec

automaticgiant:
this works (at least anecdotally), and addresses part of #13423 
is this solution sufficient to discuss how to test it?

### DIFF
--- a/builtin/provisioners/remote-exec/resource_provisioner.go
+++ b/builtin/provisioners/remote-exec/resource_provisioner.go
@@ -12,6 +12,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/hashicorp/terraform/communicator"
 	"github.com/hashicorp/terraform/communicator/remote"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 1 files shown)


builtin/provisioners/remote-exec/resource_provisioner.go
```diff
@@ -272,7 +272,7 @@ func retryFunc(ctx context.Context, timeout time.Duration, f func() error) error
 			}
 
 			// Try the function call
-			err := f()
+			err := backoff.Retry(f, backoff.NewExponentialBackOff())
```
jbardin: `backoff.Retry` already retries the function until there's a nil error, so this needs to be pulled out of the for loop.

jbardin: We should probably update the backoff package and use the `WithContext` function too, so that the Retry would get canceled with the context, otherwise it will keep running in the background.

teamterraform:
Since there was no response to the requested changes - and the code has changed quite a bit since this was opened  we are going to close this PR. Please do not hesitate to open new PRs or issues in the future!

PR #14292, opened 2017-05-08
Update _global.scss

duplich:
Fixed broken layout on mobile devices

### DIFF
--- a/website/source/assets/stylesheets/_global.scss
+++ b/website/source/assets/stylesheets/_global.scss
@@ -34,7 +34,7 @@ h1 {
   visibility: visible;
 }
 
-@media (min-width: $screen-md) {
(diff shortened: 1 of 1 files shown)


sethvargo:
Hey @duplich 

The problem with that breakpoint is that it overflows the header: 

screen shot 2017-05-08 at 2 25 33 pm

sethvargo:
You're right that the site is currently broken on mobile though. I'm not sure the best way to fix the CSS so that it doesn't overflow the header on md screens though

sethvargo:
I think we want:

```scss
@media (min-width: $screen-sm) and (max-width: $screen-md) {
  .container {
    padding: 0;
    max-width: $screen-md;
    min-width: $screen-sm;
  }
}
```

PR #14293, opened 2017-05-08
Add google_appengine_app resource

ctavan:
This adds a resource `google_appengine_app` to enable Google App Engine on a project. My major intent for this is to be able to enable Google Datastore which requires App Engine to be enabled.

Would be used like this:

```hcl
provider "google" {
  project = "my-testing-project-123"
  region  = "europe-west1"
}

resource "google_appengine_app" "default" {
  project = "my-testing-project-123"
  region  = "europe-west"
}
```

The current implementation works for me but so far it is still very bare-bones since I would love to get your feedback on this first before continuing (e.g. with writing proper tests and more docs).

The strange thing about Google App Engine is that it can only be activated once for a gcloud project but cannot be de-activated: During activation one has to choose a region (which as a side-effect also determines the data-locality of the Google Datastore instance of that gcloud project) and this cannot be changed afterwards.

How would I best formulate these conditions in terms of terraform? E.g. terraform requires me to implement the Delete method but I don't see any sensible "delete" operation for Google App Engine.

@danawillow @paddycarver @radeksimko thoughts?

### DIFF
--- a/builtin/providers/google/config.go
+++ b/builtin/providers/google/config.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
+	"google.golang.org/api/appengine/v1"
 	"google.golang.org/api/bigquery/v2"
 	"google.golang.org/api/cloudbilling/v1"
 	"google.golang.org/api/cloudresourcemanager/v1"
@@ -33,6 +34,7 @@ type Config struct {
 	Project     string
 	Region      string
 
+	clientAppengine       *appengine.APIService
 	clientBilling         *cloudbilling.Service
 	clientCompute         *compute.Service
 	clientContainer       *container.Service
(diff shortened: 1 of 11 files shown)


paddycarver:
Hi @ctavan! I want to take a close look at this, because it's confusing at first glance. But I need to play with the API a bit to understand if there's a better way to do this, or if we're just accurately representing a confusing API. Unfortunately, I doubt I'll have time this week or next. Unless @radeksimko or @danawillow have the bandwidth to look into this, I don't see it getting attention prior to 5/23, at the earliest. I just wanted to be up front about that, because I think it's a great addition to Terraform and has been on my list for a while, but I think it also deserves a thorough review. I didn't want you to hear nothing and get discouraged or think you were being ignored. It's definitely on my radar, I just can't block off the time it deserves at the moment.

Thanks for the PR! Really appreciate all the work that went into this.

ctavan:
@paddycarver no worries and thanks for the response!

Just to let you know: I'm currently using this with my patched version of terraform to enable Datastore for a production project and it works really well for me.

That of course doesn't mean that there might not be a better way to achieve activation of Datastore so I'd be really happy about a thorough review.

ctavan:
Moved over to https://github.com/terraform-providers/terraform-provider-google/pull/125

PR #14294, opened 2017-05-08
shorten output

automaticgiant:
helps with #13423


mitchellh:
This is quite reasonable, but I think #13423 should be fixed differently. Instead of posting my comments here let me just address that issue directly. Please look for my response there in a few minutes. :)

Issue #5471, opened 2016-03-06
How to set undefined value


joshuaspence:
I am writing a module for creating EC2 instances. We have a lot of duplication between our EC2 instance configuration (tags, user data, common security groups, etc) and creating a module eliminates code duplication. I am not sure, however, how to set a value to be undefined. For example, suppose that my module looked like this:

```
variable "availability_zone" {
  type = "string"
  default = ""
}

resource "aws_instance" "instance" {
  availability_zone = "${var.availability_zone}"
}
```

My question here is what is the correct default value for the `availability_zone` variable such that it is essential an optional parameter? Empty string seems to work in some cases, but I'm not sure if this would work in all cases? Basically I am looking for the Terraform equivalent of puppet's `undef`.

tmlink:
I could really use this too.    
The vSphere provider's  "vsphere_virtual_machine" resource "network_interface" block has a ipv4_prefix_length is numberic and therefore a default of `""` gives `cannot parse '' as int: strconv.ParseInt: parsing "": invalid syntax`

So I can't write modules that allow either static or dynamic addresses.   It's one or the other.

GuitarrasDeAmor:
Also curious if there is any way to set undefined variables. I am in the process of writing modules now, and it would be nice to be able to set optional values as undefined by default so that the module could contain everything the resource has to offer by default but not pulled in if it wasn't necessary.

mitchellh:
Most resources at the moment must just be written to accept some "zero" value as unset. If a resource doesn't support this please report a bug. We hope that in time we'll have a way to unset a value but at the moment we do not.

joshuaspence:
I'm curious... if this functionality does not currently exist and you do hope to implement it one day, then why close this issue? I ask out of curiosity rather than frustration. Is the reason that Hashicorp prefer only to use GitHub issues for tracking features that can be placed somewhere on the short-term roadmap rather than the long-term road map? If so, is there any place to record longer-term feature requests? Or is there no attempt to capture these feature requests because Terraform itself is moving so rapidly?

dgolja:
Resurrecting this conversation. 

> Most resources at the moment must just be written to accept some "zero" value as unset. 

What are the "zero" - unset values for the parameters ? For example I would like to know the "zero" values for the resource `aws_db_instance`, particularly the parameter `allocated_storage` ? This parameter is optional unless a `snapshot_identifier` or `replicate_source_db` is provided. The documentation doens't define the zero/undef value.

[...]

PR #14304, opened 2017-05-09
Allows zero-value ELB and ALB names

gdavison:
The `aws_alb` and `aws_elb` resources did not allow zero-value name fields.

### DIFF
--- a/builtin/providers/aws/resource_aws_alb_test.go
+++ b/builtin/providers/aws/resource_aws_alb_test.go
@@ -98,6 +98,26 @@ func TestAccAWSALB_generatedName(t *testing.T) {
 	})
 }
 
+func TestAccAWSALB_generatesNameForZeroValue(t *testing.T) {
+	var conf elbv2.LoadBalancer
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_alb.alb_test",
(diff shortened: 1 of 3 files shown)


grubernaut:
Hey @gdavison, thanks for the contribution!

I'm not quite sure, however, what this PR aims to solve. The `name` parameter is already an `Optional` and `Computed` attribute, so if the `aws_alb` resource is created without the `name` attribute specified, one will be generated. (Same applies to the `aws_elb` resource). 

Is there a use-case for explicitly setting the `name` attribute to a nil string, instead of omitting the attribute?

Thanks!

gdavison:
@grubernaut Sometimes when the `aws_alb` resource is created a generic name is ok, but other times, we may want a specific name. Allowing the zero-value field, i.e. an empty string, lets the user defer the decision or make it conditional.

```
variable "alb_name" {
  default = ""
}

resource "aws_alb" "sample_alb" {
  name = "${var.alb_name}"
 ...
}
```

This currently fails with the message "aws_alb. sample_alb: only alphanumeric characters and hyphens allowed in "name""

gdavison:
Also, @mitchellh's comment on #5471

> Most resources at the moment must just be written to accept some "zero" value as unset. If a resource doesn't support this please report a bug.

grubernaut:
@gdavison, perfect, thanks!

PR #14306, opened 2017-05-09
Add `arn` attribute to `aws_ses_domain_identity` resource

joshuaspence:
Adds an `arn` attribute to the `aws_ses_domain_identity` resource.

### DIFF
--- a/builtin/providers/aws/resource_aws_ses_domain_identity.go
+++ b/builtin/providers/aws/resource_aws_ses_domain_identity.go
@@ -19,12 +19,16 @@ func resourceAwsSesDomainIdentity() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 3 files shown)


builtin/providers/aws/resource_aws_ses_domain_identity.go
```diff
@@ -19,12 +19,16 @@ func resourceAwsSesDomainIdentity() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
stack72: This shouldn't be optional - it's only being computed :)

joshuaspence: Thanks for the fix. I just copied this from another resource :)

Issue #12901, opened 2017-03-20
Can not create custom protocol (not tcp, udp, icmp) security group rules in openstack


splashx:
### Terraform Version
```
$ terraform -v
Terraform v0.9.0-dev (89acb5811a25e59e6a9d55cf5a2429846e3b5982+CHANGES)
```

### Affected Resource(s)

- openstack_networking_secgroup_rule_v2

### Debug Output
https://gist.github.com/splashx/ccd4725837c5d3ba660eb82fc6a64319

### Expected Behavior
This is the second run, I expected nothing would change. Also, and the most important, it was expected the security group would be created correctly (`protocol: 112`)

### Actual Behavior
The vrrp rule is deleted and created again and `protocol: null`
```
.....
| name                 | tf-sg-test-vrrp                                                    |
| security_group_rules | {                                                                  |
|                      |      "remote_group_id": null,                                      |
|                      |      "direction": "ingress",                                       |
|                      |      "protocol": null,                                             |
|                      |      "description": "",                                            |
|                      |      "ethertype": "IPv4",                                          |
|                      |      "remote_ip_prefix": null,                                     |
|                      |      "port_range_max": null,                                       |
|                      |      "security_group_id": "ac42f393-89d3-4ccb-b8f5-525f7bef87be",  |

...
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` 

```
$ cat test.tf 
resource "openstack_networking_secgroup_v2" "tf-sg-test-vrrp" {
  name = "tf-sg-test-vrrp"
  description = "LDAP master backend servers"
}

resource "openstack_networking_secgroup_rule_v2" "test-tf-vrrp-rule" {
  direction = "ingress"
  ethertype = "IPv4"
  protocol = "112"
  security_group_id = "${openstack_networking_secgroup_v2.tf-sg-test-vrrp.id}"
}
```

I've also tested with:
`protocol = "vrrp"`
`protocol = 112` (no commas)

### Important Factoids
Openstack version: Mitaka

jtopjian:
Thanks for reporting this. There's a Pull Request in the upstream Gophercloud library that will help resolve this: https://github.com/gophercloud/gophercloud/pull/294

michalmedvecky:
@jtopjian do you have any idea how to workaround this with the current version?

jtopjian:
There's not a workaround at this time.

jtopjian:
[...]

PR #14307, opened 2017-05-09
provider/openstack: Add to support protocols for resourceNetworkingSecGroupRuleV2

takaishi:
issue: https://github.com/hashicorp/terraform/issues/12901

### DIFF
--- a/builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2_test.go
+++ b/builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2_test.go
@@ -85,7 +85,7 @@ func TestAccNetworkingV2SecGroupRule_timeout(t *testing.T) {
 	})
 }
 
-func TestAccNetworkingV2SecGroupRule_vrrp(t *testing.T) {
+func TestAccNetworkingV2SecGroupRule_protocols(t *testing.T) {
 	var secgroup_1 groups.SecGroup
 	var secgroup_rule_ah rules.SecGroupRule
 	var secgroup_rule_dccp rules.SecGroupRule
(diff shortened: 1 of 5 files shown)


builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2_test.go
```diff
@@ -85,6 +85,115 @@ func TestAccNetworkingV2SecGroupRule_timeout(t *testing.T) {
 	})
 }
 
+func TestAccNetworkingV2SecGroupRule_vrrp(t *testing.T) {
```
jtopjian: I think we should change the test name to something like `TestAccNetworkingV2SecGroupRule_protocols` since it's testing more than just VRRP.

builtin/providers/openstack/resource_openstack_networking_secgroup_rule_v2_test.go
```diff
@@ -226,3 +335,154 @@ resource "openstack_networking_secgroup_rule_v2" "secgroup_rule_2" {
   }
 }
 `
+
+const testAccNetworkingV2SecGroupRule_vrrp = `
```
jtopjian: Similarly here, `testAccNetworkingV2SecGroupRule_protocols`.

takaishi:
@jtopjian Thank you for comment!
I fixed test name, and rename first commit to fetch gophercloud.

jtopjian:
@takaishi Thanks! Nice work on this, thank you very much :)

jaydub:
This fix doesn't address the case where the protocol is supplied by a protocol number (because eg you're stuck using an older version of neutron). This still results in 'null' being handed to the API backend resulting in an 'any' protocol rule being created, in spite of the API supporting numeric protocols.

jtopjian:
@jaydub Thanks. Do you mind opening a new issue to track this?

PR #14319, opened 2017-05-09
provider/aws: Use helper methods for checking partition

grubernaut:
Related: #12414


builtin/providers/aws/resource_aws_cloudwatch_log_group.go
```diff
@@ -109,7 +109,7 @@ func resourceAwsCloudWatchLogGroupRead(d *schema.ResourceData, meta interface{})
 		d.Set("retention_in_days", lg.RetentionInDays)
 	}
 
-	if meta.(*AWSClient).partition != "aws-us-gov" {
+	if meta.(*AWSClient).IsChinaCloud() || meta.(*AWSClient).IsGovCloud() {
```
stack72: should be if !IsChina || !IsGovCloud

Issue #14308, opened 2017-05-09
Terraform aws-provider ignores all tags starting with "aws"


paulka:
Terraform has a check to make sure it doesn't apply AWS specific tags. These tags start with "aws:". Terraform does a RegEx check on Tag names in the form of `^aws:*`. Matching tags are ignored during an Apply.

This RegEx used by Terraform appears to be incorrect. It will match 'aws' followed by *none* or more colons. It should be 'aws" followed by *at least one* colon.

The correct RegEx would likely be along the lines of `^aws:`

### Terraform Version
Terraform v0.9.4

### Affected Resource(s)
This seems to affect most taggable AWS resources. 

### Terraform Configuration Files
```hcl
provider "aws" {
  region = "${var.aws_region}"
}

resource "aws_vpc" "default" {
  cidr_block       = "xx.y.0.0/16"
  instance_tenancy = "default"
  enable_classiclink = "false"
  enable_dns_hostnames = "true"
  enable_dns_support = "true"

  tags {
    Name = "VPC Name String"
    aws_cost_role = "bar"
    aws_cost_allocation = "foo"
  }
}
```

### Debug Output
```
2017/05/09 10:37:42 [DEBUG] apply: aws_vpc.default: executing Apply
aws_vpc.default: Modifying... (ID: vpc-yyyy725)
  tags.%:                   "1" => "3"
  tags.aws_cost_allocation: "" => "foo"
  tags.aws_cost_role:       "" => "bar"
2017/05/09 10:37:42 [DEBUG] plugin: terraform: aws-provider (internal) 2017/05/09 10:37:42 [DEBUG] No meta timeoutkey found in Apply()
2017/05/09 10:37:42 [DEBUG] plugin: terraform: aws-provider (internal) 2017/05/09 10:37:42 [DEBUG] Matching ^aws:* with Name
2017/05/09 10:37:42 [DEBUG] plugin: terraform: aws-provider (internal) 2017/05/09 10:37:42 [DEBUG] Matching ^aws:* with Name
2017/05/09 10:37:42 [DEBUG] plugin: terraform: aws-provider (internal) 2017/05/09 10:37:42 [DEBUG] Matching ^aws:* with aws_cost_role
2017/05/09 10:37:42 [DEBUG] plugin: terraform: aws-provider (internal) 2017/05/09 10:37:42 [DEBUG] Found AWS specific tag aws_cost_role (val: bar), ignoring.
[...]

Issue #14247, opened 2017-05-05
Missing Tags in ASG and Instances Confirmed in Terraform Plan


JasonKAls:
Hi there,

I'm having an issue with my billing tags not appearing on my created instances via an ASG.  As you'll see from the HCL code, I have 4 tags.  The `aws_cost_role` and the `aws_cost_allocation` do not appear what terraform is applied.

### Terraform Version
Terraform v0.9.4

### Affected Resource(s)
- aws_autoscaling_group

* Also affects AWS Instances and possibly the `aws_instance` though I'm using an ASG instead

### Terraform Configuration Files
```hcl
resource "aws_autoscaling_group" "chef-compliance-asg" {
  depends_on           = ["aws_elb.chef_compliance_elb"]
  name                 = "${var.stack_name}-${var.env}"
  vpc_zone_identifier  = ["${lookup(var.env_subnets, var.env)}"]
  max_size             = "${lookup(var.asg_max, var.env)}"
  min_size             = "${lookup(var.asg_min, var.env)}"
  desired_capacity     = "${lookup(var.asg_desired, var.env)}"
  launch_configuration = "${aws_launch_configuration.chef-compliance-lc.name}"
  load_balancers       = ["${aws_elb.chef_compliance_elb.name}"]
  health_check_type    = "EC2"

  tag {
    key                 = "api_fqdn"
    value               = "compliance-${var.env}.posrip.com"
    propagate_at_launch = true
  }

  tag {
    key                 = "environment"
    value               = "${var.env}"
    propagate_at_launch = true
  }

  tag {
    key                 = "aws_cost_role"
    value               = "${lookup(var.aws_cost_role, var.env)}"
    propagate_at_launch = true
  }

  tag {
    key                 = "aws_cost_allocation"
    value               = "${lookup(var.aws_cost_allocation, var.env)}"
    propagate_at_launch = true
  }
}.
```
### Expected Behavior
Should see all 4 Tags listed for my instances created by my ASG resource

### Actual Behavior
Missing both AWS Cost Allocation Tags

### Steps to Reproduce
1. Create ASG resource with both regular tags and Cost Allocation Tags

Below is part of the Debug I found to be interesting.  
***https://gist.github.com/JasonKAls/5c5dcb2e451a6eedf7560ddbe3873b78***
For both missing tags it states the following: Found AWS specific tag aws_cost_allocation (val: prod), ignoring.

Any help would be appreciated!

JasonKAls:
[...]

PR #14321, opened 2017-05-09
provider/aws: Update the ignoring of AWS specific tags

stack72:
We were too greedy with the AWS specific tags ignore function - we
basically were ignoring anything starting with `aws` rather than just
using `aws:`

Fixes: #14308
Fixes: #14247


grubernaut:
Agree with @apparentlymart here

stack72:
@grubernaut we can't because of generics - we pass a specific type to the function

Let's merge and then we can try and work out a better design?

PR #14332, opened 2017-05-09
provider/aws: New aws_security_group_rules resource

deftflux:
I am fixing a long-standing issue with Terraform managing AWS security group rules.  Currently, there are two ways to configure security group rules, and they each have serious short-comings:

1. Define all rules inline within the `aws_security_group` resource.
    * AWS supports referring to other security groups within security group rules.  If rules are defined inline, then a circular dependency is possible when two security groups refer to each other within their rules.  Terraform cannot resolve this dependency issue.  The only way to do this currently is to have at least one group in the circular dependency chain use the second approach rather than define inline rules.
    * If no ingress rules and/or no egress rules are defined inline, the `aws_security_group` resource will not manage one or both types of rules.  In other words, there is no way to define that a security group should have *no* ingress rules and/or *no* egress rules.  Upon applying, Terraform will not remove any of those rules.
2. Define all rules as separate `aws_security_group_rule` resources.
    * When defined this way, Terraform does not manage the entire security group as one would expect; it only manages any rules created by Terraform that exactly match what Terraform previously thought the rules were.  So if a manual change is made to a security group outside of Terraform, such as by using the AWS console, Terraform will not correct that change properly.  This fact makes this approach unacceptable from a security standpoint for managing security groups, since it does not ensure that the security groups in AWS exactly match the Terraform configuration.
        * The reason for this is because each security group rule resource only knows about itself; it does not know about the other security group rule resources, so it does not have the complete picture of all the rules that should appear in the group.  Also, in AWS, security group rules are not entities with their own IDs, so they cannot be tracked if they are modified outside of Terraform.  Therefore, none of the rule resources have enough information to know whether an existing rule in AWS does not match any rule resource and should thus be removed or replaced.
        * On a side note, each rule does have enough information to change or remove the exact rule it created when it was last applied, and it does so.  However, only if that rule still exists exactly as it was.  If it is tampered with in any way, the tampered rule will remain.
    * This approach has the same short-coming of not being able to define that there should be no ingress and/or no egress rules.
    * A minor disadvantage of this approach is that it makes the configuration rather verbose: every security group rule must be defined as a separate resource section and must separately specify the security group ID to which it belongs.

Now for my solution:

* Define all rules within a single `aws_security_group_rules` resource.  (Note the `s`; that's "rules" plural.)  This solution takes the best of the two approaches above and then some:
    * Because rules are defined in a separate resource, like `aws_security_group_rule`, there is no circular dependency in Terraform's dependency graph when rules from two security groups refer to each other.  Terraform will simply create the security groups first, then create the rules.
    * Because *all* rules are defined within a single resource, like inline rules within `aws_security_group`, the resource has enough information to verify that the rules in AWS exactly match the rules defined in the configuration.  Any manual changes done outside of Terraform will be undone.
    * Since this rules resource is for the purpose of managing the rules and, unlike `aws_security_group`, does not have to support other resources managing the rules instead, a rules resource that defines no ingress and/or no egress rules can be understood to mean that none should exist.  This is an advantage over both previous approaches.
    * While slightly more verbose than `aws_security_group`, this solution would be much less verbose than having to define many `aws_security_group_rule` resources.  Only one separate resource section with one specification of the security group ID is required to define the rules.  Other than that, the syntax matches that of the inline rules within `aws_security_group`.

Yes, this involves adding yet another resource for defining AWS security group rules.  But all three of these can still have a distinct purpose:

* `aws_security_group` is obviously necessary for defining a security group that is managed by Terraform.  The optional ability to define security group rules inline may remain for those who never refer to other security groups within their rules.

* `aws_security_group_rule` would only be safely used for adding a rule to an existing security group while preserving any rules that already exist in that group that are somehow managed elsewhere.  There may be some out there who might want to do this, but due to the short-comings of this approach, I would strongly recommend deprecating this, or at least pointing users to the new rules resource as the standard approach.

* `aws_security_group_rules` would be the standard approach for defining the exact set of rules that should belong to a given security group.


deftflux:
At this point, the new resource is functioning as I would expect it.  Although I have no way of testing if it works for legacy non-VPC security groups, if those are even still possible in AWS.  Most of the code is shared with and/or copied from `aws_security_group` or `aws_security_group_rule`, so I'm pretty confident in it.  I have a few things left to do at this point:

* Update the documentation / website.
* Update how security groups are imported.
* Author acceptance tests.

Anything else?

deftflux:
This pull request should be ready to go.

I added acceptance tests by copying rules-related ones from `aws_security_group` and adding a few around empty `aws_security_group_rules` and destroying such resources.  In order to run them myself, though, I had to make some local modifications.  I believe they should work for others, but I have no way of knowing for sure.  Also, I had no way to test one related to "classic" non-VPC security groups, because all of our AWS accounts are VPC-only.

I mentioned before that I would update the way security groups are imported.  I decided not to do so in this pull request in order to hopefully get it merged quicker.  That way others can weigh in on whether or not `aws_security_group_rules` should be the way security groups are imported without delaying the introduction of this resource.

nwalke:
I like the approach this takes and definitely would appreciate the reduced number of terraform resources.

peimanja:
+1

hlindset:
Will probably have to migrate this over to the new aws provider repository at https://github.com/terraform-providers/terraform-provider-aws.

Hopefully this PR will get some traction soon 👍

deftflux:
I have migrated this over to the new aws provider repository: https://github.com/terraform-providers/terraform-provider-aws/pull/1824

hpAB:
example please

hpAB:
how do i get this Provider doesn't support resource: aws_security_group_rules

deftflux:
This pull request was never merged in; until it is, you would have to make your own build of Terraform that includes this code.  That's why you're getting that error.

This pull request is closed now, because the aws provider is no longer part of this repository; please put your comments and voice your support over at the new pull request at the new aws provider repo: https://github.com/terraform-providers/terraform-provider-aws/pull/1824

PR #14336, opened 2017-05-09
provider/google: add failover parameter to sql database instance

### LINKED ISSUE #11652: Need ability to create Google Cloud SQL 2nd Gen *failover* replica instance
Hi,

When using Terraform to create Google Cloud SQL 2nd Gen instances, I'm trying to create a master and a failover slave/replication instance. While the docs (https://www.terraform.io/docs/providers/google/r/sql_database_instance.html) explain how to define an instance as a replica, there doesn't seem to be any way to define the slave as a replica type "failover". As a result, I have to do this in the GCE Cloud SQL web UI which prevents automation.

### Terraform Version
Terraform v0.8.5

### Affected Resource(s)
- google_sql_database_instance

```hcl
// GCP Cloud SQL Secondary DB in Zone us-central1-c
resource "google_sql_database_instance" "archesdb-secondary" {
  provider           = "google.gcp-central"
  region             = "us-central"
  database_version   = "MYSQL_5_7"
  master_instance_name = "${google_sql_database_instance.archesdb-primary.name}"

  replica_configuration {
    connect_retry_interval = "30"
  }

  settings {
    tier             = "db-n1-standard-1"

    ip_configuration {
        require_ssl        = "true"
    }

    location_preference {
        zone               = "us-central1-c"
    }
  }
}
```

### Expected Behavior
There should be a configuration flag in the replica options to specify the instance as a failover. Google Cloud SQL considers failover replicas (of which you can only have one per master instance) differently than plain read-only replicas.

### Actual Behavior
Unable to specify replica type as failover.

### References
None found.

danawillow:
Fixes #11652.

I wasn't sure the best way to test this, since it looks like the only test involving replica configs so far isn't actually run, and has this comment: `Note - this test is not feasible to run unless we generate backups first.` (https://github.com/hashicorp/terraform/blob/master/builtin/providers/google/resource_sql_database_instance_test.go#L599)

### DIFF
--- a/builtin/providers/google/resource_sql_database_instance_test.go
+++ b/builtin/providers/google/resource_sql_database_instance_test.go
@@ -408,66 +408,73 @@ func testAccCheckGoogleSqlDatabaseInstanceEquals(n string,
 			return fmt.Errorf("Error settings.pricing_plan mismatch, (%s, %s)", server, local)
 		}
 
-		if instance.ReplicaConfiguration != nil &&
-			instance.ReplicaConfiguration.MysqlReplicaConfiguration != nil {
-			server = instance.ReplicaConfiguration.MysqlReplicaConfiguration.CaCertificate
-			local = attributes["replica_configuration.0.ca_certificate"]
+		if instance.ReplicaConfiguration != nil {
+			server = strconv.FormatBool(instance.ReplicaConfiguration.FailoverTarget)
+			local = attributes["replica_configuration.0.failover_target"]
 			if server != local && len(server) > 0 && len(local) > 0 {
-				return fmt.Errorf("Error replica_configuration.ca_certificate mismatch, (%s, %s)", server, local)
+				return fmt.Errorf("Error replica_configuration.failover_target mismatch, (%s, %s)", server, local)
 			}
 
-			server = instance.ReplicaConfiguration.MysqlReplicaConfiguration.ClientCertificate
-			local = attributes["replica_configuration.0.client_certificate"]
-			if server != local && len(server) > 0 && len(local) > 0 {
-				return fmt.Errorf("Error replica_configuration.client_certificate mismatch, (%s, %s)", server, local)
-			}
+			if instance.ReplicaConfiguration.MysqlReplicaConfiguration != nil {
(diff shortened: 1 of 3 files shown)


builtin/providers/google/resource_sql_database_instance_test.go
```diff
@@ -408,66 +408,73 @@ func testAccCheckGoogleSqlDatabaseInstanceEquals(n string,
 			return fmt.Errorf("Error settings.pricing_plan mismatch, (%s, %s)", server, local)
 		}
 
-		if instance.ReplicaConfiguration != nil &&
-			instance.ReplicaConfiguration.MysqlReplicaConfiguration != nil {
-			server = instance.ReplicaConfiguration.MysqlReplicaConfiguration.CaCertificate
-			local = attributes["replica_configuration.0.ca_certificate"]
+		if instance.ReplicaConfiguration != nil {
+			server = strconv.FormatBool(instance.ReplicaConfiguration.FailoverTarget)
+			local = attributes["replica_configuration.0.failover_target"]
```
catsby: Do we need any type assertion here to make sure the value returned by  `attributes["replica_configuration.0.failover_target"]` is a bool? `make test` passes so maybe not. Did the test run at all or are we just going to skip that for now, due to the backups thing you mention?

danawillow: attributes is a map[string]string, so we do `strconv.FormatBool` to get the returned value as a string and then do a string comparison.

There is no test that currently sets this property (the one that looks like it would set replica configuration doesn't actually run, it's just a variable that never gets called)

danawillow:
+ @paddycarver as a reviewer for the testing question

builtin/providers/google/resource_sql_database_instance_test.go
catsby: Ah my bad, I misread and thought `server` was a string that was turned into a bool, not the other way around 😄

paddycarver:
The code looks good to me. The test question I can't wrap my head around. I'll follow up with you in chat tomorrow, @danawillow, if that's ok?

PR #14362, opened 2017-05-10
core/provider-split: Split out the Oracle OPC provider to new structure

stack72:
To be merged after 0.9.5 release


jbardin:
I like the look of this approach. Though the vendoring is a little awkward during the 0.9 releases, this gives a really good way to transition the providers. :+1:

PR #14367, opened 2017-05-10
Chase API in upstream joyent/triton-go

sean-:
This PR adds `context` to all of the `triton-go` API calls.  This is a new PR but identical content with #14360 but with clean history (I hosed my local git history and didn't want to clean it up).

CC @jen20


stack72:
LGTM

Issue #9328, opened 2016-10-11
AWS RDS Timeout for big Aurora deployment


iceycake:
I notice the RDS resource operation timeout has been increased multiple times in the past month.  However, for big RDS deployment, 40 minutes is not enough.  It doesn't make sense to have another PR to increase the timeout and have it hardcoded.

I am wondering if it makes sense to have the operation timeout as a parameter in the Terraform Configuration?

kwilczynski:
@iceycake hi! I wonder if supporting a `timeout` attribute that could support a genetic ISO 8601 duration format (to reduce ambiguity). I think it might work.

iceycake:
@kwilczynski It is not hard to implement the timeout duration in ISO 8601.  I also wonder if this could be a global thing.  When I was implementing Elastic Search Service, also ran into similar problem.  

Wish Hashicorp folks agree on this idea.

kwilczynski:
@iceycake I realised that perhaps even `https://golang.org/pkg/time/#ParseDuration` would be good enough, although I would prefer ISO 8601 here anyway (maybe unnecessarily).

catsby:
Hey @iceycake / @kwilczynski – thanks for bringing this up! We do in fact support this idea, but we have not committed to an approach or implementation. Questions like "should this be global" or "per resource", "both", come to mind. We agree it's necessary in some form. I'll bring this up again internally but I can't make any promises, sorry!

kwilczynski:
@catsby thank you!

stack72:
I believe this will be solved via #12311

catsby:
> I believe this will be solved via #12311

The foundation will be laid down. `aws_rds_cluster` and `_instance` will need to be updated to support the new Timeout blocks

iceycake:
@catsby Please take a look the PR and hope we can close this issue.  Thanks.

PR #14369, opened 2017-05-10
Using the new time schema helper for RDS Instance lifecycle mgmt

iceycake:
Should resolve https://github.com/hashicorp/terraform/issues/9328


stack72:
LGTM! thanks for this @iceycake :)

Issue #13729, opened 2017-04-18
feature: triton provider - support for nic ordering


pannon:
Right now if an instance is provisioned with multiple nics there is no way to influence actual nic ordering.

Example:
```
  nic     = {
    network = "b1c47e55-27f2-4bc7-a404-d9ecdd6ffbd5"
  }
  nic     = {
    network = "8359e97d-30df-4aaf-a385-527e33b7457a"
  }
```

Always prefers UUID `8359e97d` as the primary interface no matter what the order is.

Maybe there is a way to influence this right now, and I am just unaware of it?

Could be achieved by making "primary" property writable here https://github.com/hashicorp/terraform/blob/master/builtin/providers/triton/resource_machine.go#L129

jen20:
HI @pannon! Thanks for opening an issue here. I think the correct solution to this is to make NICs of `TypeList` instead of `TypeSet` such that the ordering is consistent. I'll look at doing this today.

pannon:
Hi @jen20, this is pretty much the only outstanding issue on the triton provider - so far things are looking good!

The workaround I am using right now (not ideal by any means, but works) is to add the second interface to *.tf file after the machines are created.

PS: Also, not entirely sure whether machine affinity/placement rules could be added in the future to triton driver (just like with triton cli tool). Right now I am relying on Triton auto placement rules (which I've configured to spread as much as possible across our physical nodes).

Thanks for looking at this.

sean-:
@pannon See #14370. Also, affinity and better CNS support is going to show up very soon.  PS, going to BSDCan this year?  😉

pannon:
@sean- nice one - Thanks for looking at this!

Will make it next year - missed the boat on this one.

PR #14370, opened 2017-05-10
Use a List instead of a Set for triton machine NICs.

sean-:
At present order is semantically important when communicating with the Triton CloudAPI.

CC @jen20 

Fixes: #13729

### DIFF
--- a/builtin/providers/triton/resource_machine.go
+++ b/builtin/providers/triton/resource_machine.go
@@ -3,6 +3,7 @@ package triton
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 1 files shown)


sean-:
Err, sit tight on this for a few hours: I clipped a section out of the Update() method prematurely and need to add those bits first before this can be merged.

https://github.com/hashicorp/terraform/pull/14370/files#diff-7b75852af5fdaf82fd499b31c79fe91eL521

catsby:
Hey @sean- ; I'm going to go ahead and close this, as Triton has it's own home now. If it's applicable maybe you can migrate the code, otherwise I hope it's already addressed. Thanks!

PR #14393, opened 2017-05-11
Add a `url` attribute to `google_storage_bucket`

dasch:
This makes it easier to hook up bucket resources with resource definitions that expect URLs on the format `gs://<bucket-name>`.


stack72:
Hi @dasch 

Thanks for the work here - please can you also update the documentation to show this new exported field?

Thanks

Paul

Issue #14394, opened 2017-05-11
Add support for configurable delete timeout on aws_autoscaling_group


jringuette:
### Affected Resource
- aws_autoscaling_group

### Problem
When deleting ASG, we keep hitting the hardcoded 5 minutes timeout.

```
aws_launch_configuration.app: Refreshing state... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba)
aws_autoscaling_group.app: Refreshing state... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba)
module.alert-cw-cpu-threshold-75-app.aws_cloudwatch_metric_alarm.app_asg_cpu_threshold: Destroying... (ID: PRD130-intdev2-scaling-cpu-threshold-75-app-AppAsgCriticalCpuUtilization)
module.alert-cw-cpu-threshold-85-app.aws_cloudwatch_metric_alarm.app_asg_cpu_threshold: Destroying... (ID: PRD130-intdev2-scaling-cpu-threshold-85-app-AppAsgCriticalCpuUtilization)
module.alert-cw-cpu-threshold-75-app.aws_cloudwatch_metric_alarm.app_asg_cpu_threshold: Destruction complete
module.alert-cw-cpu-threshold-85-app.aws_cloudwatch_metric_alarm.app_asg_cpu_threshold: Destruction complete
aws_autoscaling_group.app: Destroying... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba)
aws_autoscaling_group.app: Still destroying... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba, 10s elapsed)
aws_autoscaling_group.app: Still destroying... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba, 20s elapsed)
aws_autoscaling_group.app: Still destroying... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba, 30s elapsed)
aws_autoscaling_group.app: Still destroying... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba, 40s elapsed)
aws_autoscaling_group.app: Still destroying... (ID: PRD130.intdev2.scaling.app0096243463a96bdd11ff4921ba, 50s elapsed)
[...]

PR #14396, opened 2017-05-11
Make the ASG delete timeout configurable

jringuette:
Fix issue #14394 

I've added a configurable delete timeout for the ASG.

I've modified helper/schema/resource.go to read the meta from the state when it is not present in the diff. I'm not sure why this is happening, but on every tries I made, the diff never had the meta info into. It might not be the proper way to fix that problem tough, if someone more familiar with the code base could have a look. But it does not seem to cause any regression nor problem.

Documentation is also updated to reflect the new supported timeout.

### DIFF
--- a/builtin/providers/aws/resource_aws_autoscaling_group.go
+++ b/builtin/providers/aws/resource_aws_autoscaling_group.go
@@ -27,6 +27,10 @@ func resourceAwsAutoscalingGroup() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		Timeouts: &schema.ResourceTimeout{
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
 		Schema: map[string]*schema.Schema{
 			"name": &schema.Schema{
 				Type:          schema.TypeString,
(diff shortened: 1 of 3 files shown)


iceycake:
@catsby can you take a look to see if this PR is good?  Thanks.

catsby:
Hey @jringuette can you describe when this was happening? 

> I've modified helper/schema/resource.go to read the meta from the state when it is not present in the diff.

Was that all the time? Or just certain situation(s)?

jringuette:
@catsby, It was all the time. It was always returning the default timeout (20 minutes) from here https://github.com/hashicorp/terraform/blob/2fe5976aec290f4b53f07534f4cde13f6d877a3f/helper/schema/resource_data.go#L368

When opening the state file, I was able to see the timeouts correctly saved under the meta e2bfb730-ecaa-11e6-8f88-34363bc7c4c0, but when reading, it always read from the diff which did not contained that meta.

ledor473:
Any update on this ticket? We faced the same issue recently.

catsby:
This has been migrated to https://github.com/hashicorp/terraform/pull/15374 and https://github.com/terraform-providers/terraform-provider-aws/pull/949

Issue #14344, opened 2017-05-10
terraform not setting fqdn for hostname in openstack instance


aswinvg:
### Terraform Version
Terraform v0.9.2

### Affected Resource(s)
- openstack_compute_instance_v2

resource "openstack_compute_instance_v2" "server1-region1" {
  name            = "server1.region1.npe.xyzcorp.com"
  region          = "region1"
  image_id        = "id"
  flavor_id       = "fid"
  key_pair        = "keypair"
  security_groups = ["default"]

  network {
    name = "default-net"
  }
}

### Scenario
No errors. the Openstack instance is getting created fine and the name "server1.region1.npe.service.xyzcorp.com" is shown properly in openstack.

But on the box, the hostname is set to "server1" without the complete domain name (.region1.npe.xyzcorp.com) - am i missing something to set in my tf file?

Thanks,
Aswin

jtopjian:
@aswinvg This isn't specific to Terraform or OpenStack per-say. This is most likely an image issue. You should be able to resolve this with cloud-init (if the image supports cloud init). Take a look [here](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux_OpenStack_Platform/4/html/End_User_Guide/user-data.html) and search for "fqdn" (the doc is from RH, but it will work with any distro that has cloud-init.

You can pass this into the instance via Terraform by doing:

```ccl
resource "openstack_compute_instance_v2" "server1-region1" {
  ...
  user_data = "#cloud-config\nhostname: server1.region1.npe.service.xyzcorp.com\nfqdn: server1.region1.npe.service.xyzcorp.com"
  ...
}
```

I've found I've had to use the fqdn both I the `hostname` and `fqdn` settings.

I'm going to close this issue, but do let me know if you have any further questions. Hope this helps.

aswinvg:
this is fantastic and it works - but i dont see examples for user_data examples in the terraform documentation - can you point me some?

jtopjian:
> but i dont see examples for user_data examples in the terraform documentation

I'll get that fixed.

> can you point me some?

`user_data` can take any cloud-init configuration. You can read it in from a file, use the cloud-init Terraform data source, etc. I recommend checking out the docs (https://cloudinit.readthedocs.io/en/latest/) and googling around for other existing examples.

aswinvg:
thank you - i could get the info from RedHat cloud init.

rhoml:
[...]

PR #14411, opened 2017-05-11
provider/openstack: Add documentation about user_data

jtopjian:
For #14344


jtopjian:
@stack72 Thanks, Paul. Sorry - I got pulled away at the conference and wasn't able to merge. :)

PR #14414, opened 2017-05-11
Allowed method on aggregator is `avg` ! `average`

dthagard:
While Datadog will accept the value of `average` when creating the query graph, the resultant graph will be empty. Passing the value of `avg` instead correctly renders the graph.

Graph with `average`:
image

Graph with `avg`:
image


grubernaut:
Hey @dthagard, thanks for the submission! 

Looks like the unit test just needs to be updated to fix the CI tests: 
```
--- FAIL: TestValidateAggregatorMethod (0.00s)
	resource_datadog_timeboard_test.go:235: "average" should be a valid aggregator method: ["\"request\" contains an invalid method \"average\". Valid methods are either \"avg\", \"max\", \"min\", \"sum\", or \"last\""]
```

stack72:
LGTM! Thanks for this :)

PR #14424, opened 2017-05-12
provider/google: use a mutex to prevent concurrent sql instance operations

### LINKED ISSUE #9018: google_sql_database.users: operationInProgress error on `terraform apply`
Hi there,

We're seeing `google_sql_database.users: Error, failed to insert database [name] into instance [name]: googleapi: Error 403: Operation failed because another operation was already in progress., operationInProgress` very frequently but not always. We see it only on google_sql_database.users, but terraform applying other resources reliably works.
### Terraform Version

Terraform v0.7.4
### Affected Resource(s)
- google_sql_database.users

``` hcl
resource "google_sql_database_instance" "master" {
  region = "${var.sql_region}"
  settings {
    tier = "${var.google_sql_db_tier}"
   }

  count = "${var.google_sql_instance_count}"
}

resource "google_sql_database" "users" {
  name = "${var.env_name}-db"
  instance = "${google_sql_database_instance.master.name}"

  count = "${var.google_sql_instance_count}"
}

resource "google_sql_user" "users" {
  name = "${var.google_sql_db_username}"
  password = "${var.google_sql_db_password}"
  instance = "${google_sql_database_instance.master.name}"
  host     = "${var.google_sql_db_host}"

  count = "${var.google_sql_instance_count}"
}
```
### Debug Output

https://gist.github.com/maryamklabib/5681f634691f2e09782df962cf016ce8
### Steps to Reproduce
1. `terraform apply`
### Important Factoids

We're creating First Generation GoogleSQL users

danawillow:
Fixes #9018.

A few implementation questions:
* Is `resource_sql_database_instance_test` the best test file for these changes? It's not about instance creation, it's about other stuff creation, but it's also not user or database specific so instance felt more general.
* Is locking on the instance name sufficient or should we have some sort of prefix to make sure we don't accidentally conflict with other resources? I would do the instance URL but we don't have it at that point and it seems silly to do a read just to get the URL so we can lock on it (and I also don't want to reconstruct the URL). We can get the instance URL from the operation, but at that point it's too late, I think. 

```
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccGoogleSqlDatabaseInstance_multipleOperations -timeout 120m
=== RUN   TestAccGoogleSqlDatabaseInstance_multipleOperations
--- PASS: TestAccGoogleSqlDatabaseInstance_multipleOperations (51.86s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	51.985s
```


ctavan:
Thanks a lot for attacking this issue!

While I don't feel 100% competent to comment on this I believe that in my case a mutex on the instance name would not be the ideal solution. While the default for `google_sql_database_instance` is to auto-generate a random instance name where using the instance name for the mutex should be fine, I have chosen instance names like main/default/master in practice and I have used the same name for many other resources as well since many of my google resources are grouped within terraform modules where I don't see a need for more specific naming (not sure if other google resources use the plain name for mutexes though).

I think prefixing would already improve the situation a lot even though a slightly uneasy feeling remains for me since there is still the hypothetical case where names could collide… I admit that it's probably not a problem in practice though.

danawillow:
That makes sense! I added a prefix, and included the project name in the key as well just in case the same database name is being used in multiple projects with the same tf config.

ctavan:
Great, that should be reasonably unique! Now another resource in the same project would have to choose that exact prefix to produce a clash. I still wonder if we should leave a small comment on the `instanceMutexKey` that we are aware that the way we determine the mutex key is not 100% safe but that we reason that in practice it should be very close to 100% safe… Should someone ever run into the issue that might help. But maybe it's not worth the effort.

Anyways, my concrete issue should be fixed by your patch, so thanks a lot for this!

builtin/providers/google/provider.go
```diff
@@ -266,3 +267,6 @@ func handleNotFoundError(err error, d *schema.ResourceData, resource string) err
 
 	return fmt.Errorf("Error reading %s: %s", resource, err)
 }
+
+// This is a global MutexKV for use within this plugin.
```
paddycarver: Would be cool to explain what it's locking on, and what limitation makes the lock necessary.

builtin/providers/google/resource_sql_database.go
```diff
@@ -67,6 +68,7 @@ func resourceSqlDatabaseCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	err = sqladminOperationWait(config, op, "Insert Database")
+	googleMutexKV.Unlock(instanceMutexKey(project, instance_name))
```
paddycarver: I'd rather see this after line 60 in a `defer`, to be honest, just to cover our bases.

builtin/providers/google/resource_sql_database.go
```diff
@@ -121,6 +124,7 @@ func resourceSqlDatabaseDelete(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	err = sqladminOperationWait(config, op, "Delete Database")
+	googleMutexKV.Unlock(instanceMutexKey(project, instance_name))
```
paddycarver: Same, would really like to see the `defer` approach used here, for completeness.

builtin/providers/google/provider.go
paddycarver: Also, rather than having this be plugin-wide, what about making it just for this particular concept--e.g., locking insertion/removal of SQL instances? Then we don't need to worry about the prefix, we can just use something that uniquely identifies an SQL instance amongst all SQL instances.

builtin/providers/google/resource_sql_user.go
```diff
@@ -78,6 +79,7 @@ func resourceSqlUserCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	err = sqladminOperationWait(config, op, "Insert User")
+	googleMutexKV.Unlock(instanceMutexKey(project, instance))
```
paddycarver: `defer` this?

builtin/providers/google/resource_sql_user.go
```diff
@@ -154,6 +157,7 @@ func resourceSqlUserUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		err = sqladminOperationWait(config, op, "Insert User")
+		googleMutexKV.Lock(instanceMutexKey(project, instance))
```
paddycarver: `defer` this?

builtin/providers/google/resource_sql_user.go
```diff
@@ -187,6 +192,7 @@ func resourceSqlUserDelete(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	err = sqladminOperationWait(config, op, "Delete User")
+	googleMutexKV.Unlock(instanceMutexKey(project, instance))
```
paddycarver: `defer` this?

builtin/providers/google/resource_sql_database.go
catsby: Does `defer` here actually result in the `Unlock` not being called until `resourceSqlDatabaseRead` is called? If so, then I think the unlock here is probably correct, _assuming_ the `resourceSqlDatabaseRead` is idempotent

catsby: Indeed, `defer` here will result in the `lock` persisting until `resourceSqlDatabaseRead` completes, so probably not what we want to do. 

- https://play.golang.org/p/61yvP1eLbs

That said, we probably do want to add an additional unlock in the error check block, yeah?

builtin/providers/google/provider.go
paddycarver: Fair, and good to know!

builtin/providers/google/resource_sql_database.go
paddycarver: Yeah, I knew defer would block until `resourceSqlDatabaseRead` was finished; I checked to make sure a lock wasn't needed inside it. IMO, having the lock stuff work without the danger of missing a lock (like in the error check block that you noticed) outweighs the expanded lock time, but I'm willing to be outvoted on that by you and @danawillow.

Regardless, I think we do want to make sure the lock is released after that error check block, even though (IIUC) Terraform will exit anyways, meaning the lock would never be used again regardless.

builtin/providers/google/provider.go
paddycarver: Fair enough, and good to know!

builtin/providers/google/resource_sql_database.go
paddycarver: (I stand corrected on the unlock in the error check block--this Unlock will fire before the error checking, so another Unlock in the error check block below is unnecessary.

paddycarver:
I'm fine with it, but if @danawillow is willing to update to release the lock in the error block, that'd be even better. :) Leaving to @danawillow.

danawillow:
I'm fine with the defer- releasing the lock after the read is done seems reasonable. We can always change it later. Updated to add it.

danawillow:
Yes of course, just needed to merge with master. Done.

PR #14428, opened 2017-05-12
Correct argument description for targets

PaulAtkins:
Correct `targets` argument description to show the allowed keys are `InstanceIds` or `WindowTargetIds`

Relevant for v0.9.5


stack72:
Thanks for this @PaulAtkins - will get this pushed to the site today

PR #14429, opened 2017-05-12
Allow a codepipeline action to specify a role_arn

AKoetsier:
The codepipeline resource allows an action to specify a role_arn. However that role is never passed to AWS. This PR fixes that behavior.


AKoetsier:
Close in favour of 14263

PR #14431, opened 2017-05-12
Do not fail on non-existent codepipeline

AKoetsier:
This PR allow will mark a pipeline as non-existent when it cannot be found in AWS


stack72:
Nice addition @AKoetsier :) Thanks for this

Paul

PR #14441, opened 2017-05-12
provider/google: Fix health check http/https defaults

rileykarson:
Change the http_health_check and https_health_check default values in resource_compute_health_check.go to match the default values in the documentation, and the default values of the specific health check resources.


danawillow:
Technically this is a breaking change since it removes the default. However, since the original version was setting up incorrect infrastructure, @paddycarver and I are fine with this since the config files that would break are technically already broken.

PR #14457, opened 2017-05-13
provider/google: Add import support to google_sql_user

rileykarson:
Add support for "terraform import" with google_sql_user resources.

This uses a slightly different format than normal to support specifying the database instance name. The format is documented on the resource's page, along with a note that passwords are not retrieved

### DIFF
--- a/builtin/providers/google/resource_sql_user.go
+++ b/builtin/providers/google/resource_sql_user.go
@@ -136,8 +136,6 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
-	d.SetId(fmt.Sprintf("%s.%s", instance, name))
-
 	d.Set("host", user.Host)
 	d.Set("instance", user.Instance)
 	d.Set("name", user.Name)
--- a/builtin/providers/google/resource_sql_user.go
+++ b/builtin/providers/google/resource_sql_user.go
@@ -84,7 +84,7 @@ func resourceSqlUserCreate(d *schema.ResourceData, meta interface{}) error {
 			"user %s into instance %s: %s", name, instance, err)
 	}
 
-	d.SetId(fmt.Sprintf("%s.%s", instance, name))
+	d.SetId(fmt.Sprintf("%s/%s", instance, name))
 
 	err = sqladminOperationWait(config, op, "Insert User")
 
@@ -104,10 +104,10 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	instanceAndName := strings.SplitN(d.Id(), ".", 2)
+	instanceAndName := strings.SplitN(d.Id(), "/", 2)
 	if len(instanceAndName) != 2 {
 		return fmt.Errorf(
-			"Wrong number of arguments when specifying imported id. Expected: 2.  Saw: %d. Expected Input: $INSTANCENAME.$SQLUSERNAME Input: %s",
+			"Wrong number of arguments when specifying imported id. Expected: 2.  Saw: %d. Expected Input: $INSTANCENAME/$SQLUSERNAME Input: %s",
 			len(instanceAndName),
 			d.Id())
 	}
--- a/builtin/providers/google/resource_sql_user.go
+++ b/builtin/providers/google/resource_sql_user.go
@@ -1,13 +1,11 @@
 package google
 
 import (
-	"errors"
 	"fmt"
 	"log"
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
-
 	"google.golang.org/api/sqladmin/v1beta4"
 )
 
@@ -20,6 +18,9 @@ func resourceSqlUser() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+
+		SchemaVersion: 1,
+		MigrateState:  resourceSqlUserMigrateState,
 
 		Schema: map[string]*schema.Schema{
 			"host": &schema.Schema{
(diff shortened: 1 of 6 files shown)


rileykarson:
@danawillow

builtin/providers/google/resource_sql_user.go
```diff
@@ -95,32 +103,44 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
...
-	if !found {
+	if user == nil {
 		log.Printf("[WARN] Removing SQL User %q because it's gone", d.Get("name").(string))
 		d.SetId("")
 
 		return nil
 	}
 
-	d.SetId(name)
+	d.SetId(fmt.Sprintf("%s.%s", instance, name))
```
catsby: It's generally not a good idea to call `SetId` outside of a `CREATE` method. I realize this isn't your addition, but do you know off hand why this is here? Can it be removed and have the resource work as expected?

rileykarson: You're right - this was able to be safely removed. 🎉

builtin/providers/google/resource_sql_user.go
```diff
@@ -77,6 +83,8 @@ func resourceSqlUserCreate(d *schema.ResourceData, meta interface{}) error {
 			"user %s into instance %s: %s", name, instance, err)
 	}
 
+	d.SetId(fmt.Sprintf("%s.%s", instance, name))
```
danawillow: So I take back part of what I told you in person. We shouldn't change the format of the ID, since that will break refreshing (since the ID in state will still be the old version). At this point we have three possible solutions:
1. Write a migration function to change all the IDs in state. I _think_ it would work, but migration functions are kind of a pain.
2. Do the ID parsing in the StateFn, so instead of doing a `schema.ImportStatePassthrough`, create a custom function that parses the ID and then sets the two properties in the state before it gets passed to the read fn.
3. Create a custom function that reads all available instances, and then tries each one to see if the user belongs to that instance (see [resource_pagerduty_service_integration](https://github.com/hashicorp/terraform/blob/master/builtin/providers/pagerduty/resource_pagerduty_service_integration.go) for an example). This is actually less bad than I initially made it out to be. Since Terraform is currently storing the user name as the ID, it's already not allowing multiple users with the same name in different instances. And so a naive import should fail in this case anyway. However the point could be made that Terraform _should_ allow users with the same name in different instances.

I'm not sure yet which the best solution is. I'll consult with Hashicorp people and get back to you.

builtin/providers/google/resource_sql_user.go
```diff
@@ -95,32 +103,42 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	name := d.Get("name").(string)
-	instance := d.Get("instance").(string)
+	instanceAndName := strings.SplitN(d.Id(), ".", 2)
+	if len(instanceAndName) != 2 {
+		return errors.New(
```
danawillow: fmt.Errorf :D

builtin/providers/google/resource_sql_user.go
danawillow: All right, the experts have spoken and recommended solution 1. I can explain migration fns to you once I'm in the office, or feel free to give it a try on your own (I actually don't think it'll be much of a pain, this will be a pretty straightforward one)

JDiPierro:
This is great to see! I started adding importability for `google_dns_record_set`s but stopped since they don't have a singular ID; import would have to do something similar to `{instance}.{name}` that's done here. I think I'll pick that back up soon and follow this pattern. :)

Also: I think you'll want to add `google_sql_user` to the list of importable resources at `/website/source/docs/import/importability.html.md`.

builtin/providers/google/resource_sql_user_migrate.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func resourceSqlUserMigrateState(
+	v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
+	if is.Empty() {
+		log.Println("[DEBUG] Empty SqlUserState; nothing to migrate.")
```
danawillow: Actually InstanceState is a terraform type so you do want InstanceState here instead of SqlUserState

builtin/providers/google/resource_sql_user_migrate_test.go
```diff
@@ -0,0 +1,78 @@
...
+		ID           string
+		ExpectedID   string
+	}{
+		"change id from $NAME to $INSTANCENAME.$NAME": {
+			StateVersion: 0,
+			Attributes: map[string]string{
+				"name":     "tf-user",
+				"instance": "tf-instance",
+			},
+			Expected:   map[string]string{},
```
danawillow: Why is `Expected` empty?

builtin/providers/google/resource_sql_user.go
```diff
@@ -95,32 +104,41 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	name := d.Get("name").(string)
-	instance := d.Get("instance").(string)
+	instanceAndName := strings.SplitN(d.Id(), ".", 2)
```
danawillow: I'm slightly nervous about using dots as a separator. Even though this set of properties is safe, I can imagine a resource that needs to be imported in a similar way but dots are acceptable characters in names for both types of resources, and then we would have an inconsistency between resources. I think a "/" would be safer, since I'd be really surprised if it ever came up as a separator (because of weirdness in self_links). What do you think?

rileykarson: This makes sense - "/" is probably a safer character to use than ".".

Done.

builtin/providers/google/resource_sql_user_migrate.go
rileykarson: Done.

builtin/providers/google/resource_sql_user_migrate_test.go
rileykarson: Done.

PR #14461, opened 2017-05-13
Check whether floatingIP or associated instance deleted

gmlexx:
In case if openstack instance with associated floating_ip was terminated, applying the same plan could raise an error like this:
Error applying plan:

1 error(s) occurred:

* openstack_compute_floatingip_associate_v2.test-tmp (destroy): 1 error(s) occurred:

* openstack_compute_floatingip_associate_v2.test-tmp: floating ip association: Expected HTTP response code [201 202] when accessing [POST http://10.94.13.13:8774/v2.1/servers/3293d6eb-f252-46c6-9a69-22692537557d/action], but got 409 instead
{"conflictingRequest": {"message": "Floating IP 10.95.2.30 is not associated with instance 3293d6eb-f252-46c6-9a69-22692537557d.", "code": 409}}

Added check of instance of floating_ip existence.

### DIFF
--- a/builtin/providers/openstack/resource_openstack_compute_floatingip_associate_v2.go
+++ b/builtin/providers/openstack/resource_openstack_compute_floatingip_associate_v2.go
@@ -79,6 +79,12 @@ func resourceComputeFloatingIPAssociateV2Create(d *schema.ResourceData, meta int
 }
 
 func resourceComputeFloatingIPAssociateV2Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
 	// Obtain relevant info from parsing the ID
 	floatingIP, instanceId, fixedIP, err := parseComputeFloatingIPAssociateId(d.Id())
 	if err != nil {
@@ -89,6 +95,17 @@ func resourceComputeFloatingIPAssociateV2Read(d *schema.ResourceData, meta inter
 	d.Set("instance_id", instanceId)
 	d.Set("fixed_ip", fixedIP)
 	d.Set("region", GetRegion(d))
+
+	// check whether floatingIP or associated instance deleted
+	fip, err := floatingips.Get(computeClient, floatingIP).Extract()
+	if err != nil {
+		if CheckDeleted(d, err, "floating ip") == nil {
+			return nil
+		}
+	}
+	if fip.InstanceID == "" {
+		d.SetId("")
+	}
 
 	return nil
 }


builtin/providers/openstack/resource_openstack_compute_floatingip_associate_v2.go
```diff
@@ -90,6 +90,22 @@ func resourceComputeFloatingIPAssociateV2Read(d *schema.ResourceData, meta inter
 	d.Set("fixed_ip", fixedIP)
 	d.Set("region", GetRegion(d))
 
+	// check whether floatingIP or associated instance deleted
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
```
jtopjian: Can you move lines 94-98 to the top of the function like how it is in the other functions?

gmlexx: Ok, I've rewritten commit

jtopjian:
@gmlexx Arg. I'm really sorry. I was doing another review of this and discovered a significant issue.

First, the issue you're trying to solve is absolutely valid and should be implemented. Unfortunately getting the solution so it works correctly with a wide variety of OpenStack environments isn't so simple.

I'm not sure how familiar you are with the OpenStack APIs, but in short, most of the Compute/Nova API support for Floating IPs has been deprecated, and I've seen some user reports of environments that now no longer support those calls. The only remaining non-deprecated API calls are associating and disassociating floating IPs to instances. That's why this resource is so slim to begin with.

So that poses a problem when trying to `Get` or `List` an IP through the Compute API. It might very well cause this resource to begin failing in certain OpenStack environments.

So the next best thing to do is to introduce a Network/Neutron client into this resource to query the floating IP. But this isn't perfect, either, because not everyone might have that service deployed.

I think what needs to happen (and what I've been trying to avoid, but maybe that's no longer possible) is to begin supporting both types of API calls.

For reference, [here's](https://gist.github.com/jtopjian/7f9efea8647b05278de040a6a91c8616) a fully working solution using the Compute/Nova API. This needs further modified to first check and see if creating a network client is successful. If so, use the [network](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/layer3/floatingips/requests.go) API to List the IP address and get the ID, then do a Get on the ID. If a network client fails to be created, then a secondary measure of using the Compute API (as shown in the gist) is a good fallback.

If you want to try to tackle this, I'd be happy to help. But I totally understand if this is something much more than you had anticipated, so I can also take over.

jtopjian:
@gmlexx Actually, I have a solution almost done. It'll be finished up within the next day :)

gmlexx:
@jtopjian Thanks a lot for quick and really comprehensive answer. Closing this request.

Issue #14465, opened 2017-05-14
Import of aws_route53_record with Underscore Fails


BlueDragonX:
### Terraform Version
Terraform v0.9.5

### Affected Resource(s)
aws_route53_record

### Terraform Configuration Files
```hcl
resource "aws_route53_record" "tld_mailgun_key" {
  zone_id = "${aws_route53_zone.tld.id}"
  name    = "krs._domainkey.${var.tld}"
  type    = "TXT"
  ttl     = 300

  records = [
    "\"k=rsa; p=ABCDEFG\"",
  ]
}
```

### Expected Behavior
The Route53 record is imported into Terraform successfully.

### Actual Behavior
This error is printed:
```
aws_route53_record.tld_mailgun_key: Import complete!
  Imported aws_route53_record (ID: SANITIZEDZONEID_krs._domainkey.cntr.io_TXT)
aws_route53_record.tld_mailgun_key: Refreshing state... (ID: SANITIZEDZONEID_krs._domainkey.cntr.io_TXT)
Error importing: 1 error(s) occurred:

* aws_route53_record.tld_mailgun_key (import id: SANITIZEDZONEID_krs._domainkey.cntr.io_TXT): 1 error(s) occurred:

* import aws_route53_record.tld_mailgun_key result: SANITIZEDZONEID_krs._domainkey.cntr.io_TXT: aws_route53_record.tld_mailgun_key: InvalidInput: 1 validation error detected: Value 'domainkey.cntr.io' at 'startRecordType' failed to satisfy constraint: Member must satisfy enum value set: [AAAA, A, NS, SOA, SPF, MX, PTR, CNAME, TXT, SRV, NAPTR]
	status code: 400, request id: SANITIZED
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create a Route53 record in AWS with an underscore in the name.
2. Create a Terraform config for that record.
3. Attempt to import the record with `terraform import`.

### Important Factoids
It clearly attempts to split the ID on the underscore in the domain name. I verified this by finding [the specific piece of code responsible for parsing the ID](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_route53_record.go#L461). As I suspected it simply calls `strings.Split(d.Id(), "_")`.

I may cook up a PR if someone doesn't beat me to it.

PR #14466, opened 2017-05-14
provider/aws: Support import of Route53 records with underscores.

BlueDragonX:
This is an attempt to fix https://github.com/hashicorp/terraform/issues/14465.

### DIFF
--- a/builtin/providers/aws/resource_aws_route53_record.go
+++ b/builtin/providers/aws/resource_aws_route53_record.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


BlueDragonX:
My initial attempt to fix this doesn't actually work. When I run a `terraform plan` I end up with this:
```
-/+ aws_route53_record.tld_mailgun_key
    fqdn:               "cntr.io" => "<computed>"
    name:               "" => "krs._domainkey.cntr.io" (forces new resource)
    records.#:          "1" => "1"
    records.1008113961: "v=spf1 include:mailgun.org ~all" => ""
    records.1512342320: "" => "\"k=rsa; p=ABCDEFG\""
    ttl:                "300" => "300"
    type:               "TXT" => "TXT"
    zone_id:            "SANITIZEDZONEID" => "SANITIZEDZONEID"
```
It's generating the wrong `fqdn` in the process somehow.

stack72:
Closed via #14717 

Thanks for the work here @BlueDragonX - I have just migrated another version of this across to the new Terraform AWS Provider repo

Issue #14471, opened 2017-05-14
GitLab provider enhancements


hauleth:
Currently there is only project management and hooks management it would be to provide more GitLab resources:

- [x] [Deploy keys](https://docs.gitlab.com/ce/api/deploy_keys.html#deploy-keys)
- [x] [Groups](https://docs.gitlab.com/ce/api/groups.html)
- [ ] [Build variables](https://docs.gitlab.com/ce/api/build_variables.html#create-variable)
- [ ] [Labels](https://docs.gitlab.com/ce/api/labels.html#create-a-new-label)
- [ ] [Boards](https://docs.gitlab.com/ce/api/boards.html#new-board-list)

Ref:

- #13898

jugatsu:
It would be great to have `user` resource too.

PR #14483, opened 2017-05-15
Add namespcace ID attribute

hauleth:
This commit also introduce `id` comouted value which is numeric value
used by GitLab to iteract with repository. This should simplify use of
`gitlab_project_hook` usage and would allow to introduce other resources
as described in #14471

### DIFF
--- a/builtin/providers/gitlab/resource_gitlab_project.go
+++ b/builtin/providers/gitlab/resource_gitlab_project.go
@@ -60,10 +60,6 @@ func resourceGitlabProject() *schema.Resource {
 				Default:      "private",
 			},
 
-			"id": {
-				Type:     schema.TypeInt,
-				Computed: true,
-			},
 			"ssh_url_to_repo": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -90,7 +86,6 @@ func resourceGitlabProjectSetToState(d *schema.ResourceData, project *gitlab.Pro
 	d.Set("snippets_enabled", project.SnippetsEnabled)
 	d.Set("visibility_level", visibilityLevelToString(project.VisibilityLevel))
 
-	d.Set("id", project.ID)
 	d.Set("ssh_url_to_repo", project.SSHURLToRepo)
 	d.Set("http_url_to_repo", project.HTTPURLToRepo)
 	d.Set("web_url", project.WebURL)
--- a/builtin/providers/gitlab/resource_gitlab_project.go
+++ b/builtin/providers/gitlab/resource_gitlab_project.go
@@ -95,11 +95,14 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 	client := meta.(*gitlab.Client)
 	options := &gitlab.CreateProjectOptions{
 		Name:                 gitlab.String(d.Get("name").(string)),
-		NamespaceID:          gitlab.Int(d.Get("namespace_id").(int)),
 		IssuesEnabled:        gitlab.Bool(d.Get("issues_enabled").(bool)),
 		MergeRequestsEnabled: gitlab.Bool(d.Get("merge_requests_enabled").(bool)),
 		WikiEnabled:          gitlab.Bool(d.Get("wiki_enabled").(bool)),
 		SnippetsEnabled:      gitlab.Bool(d.Get("snippets_enabled").(bool)),
+	}
+
+	if v, ok := d.GetOk("namespace_id"); ok {
+		options.NamespaceID = gitlab.Int(v.(int))
 	}
 
 	if v, ok := d.GetOk("description"); ok {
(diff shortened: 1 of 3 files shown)


hauleth:
I am not proficient in Go enough to write tests to cover new functionality though.

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -56,6 +60,10 @@ func resourceGitlabProject() *schema.Resource {
 				Default:      "private",
 			},
 
+			"id": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
```
richardc: This isn't needed.  The reason you cite for adding it is to help work with the `gitlab_project_hook` resource, but if you look at the acceptance tests for that code you'll see that the id is already exposed and used in its tests:  https://github.com/hashicorp/terraform/blob/master/builtin/providers/gitlab/resource_gitlab_project_hook_test.go#L189   Its just an automatically exported attribute that all resources get.

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -82,6 +90,7 @@ func resourceGitlabProjectSetToState(d *schema.ResourceData, project *gitlab.Pro
 	d.Set("snippets_enabled", project.SnippetsEnabled)
 	d.Set("visibility_level", visibilityLevelToString(project.VisibilityLevel))
 
+	d.Set("id", project.ID)
```
richardc: Remove this, it's already done elsewhere

hauleth: @richardc it would be worth documenting then as I cannot see that in `gitlab_project` Attributes Reference

richardc: Sure, the docs are in the repo too.   Probably also worth adding the missing example to `gitlab_project_hooks` too, so the next user doesn't think that `"${gitlab_project.foo.id}"` isn't usable already.

richardc:
Cherry-picking this over to my gitlab_group branch (resulting frankenbranch at https://github.com/richardc/terraform/tree/feature/add_gitlab_group_plus_cherry) does result in the ability to create a project in a group with as simply as this:

~~~
resource "gitlab_group" "test1" {
  name = "test1"
  path = "test1"
  visibility_level = "public"
}

  name = "test1"
  namespace_id = "${gitlab_group.test1.id}"
  description = "A test project! Modify 2"
  visibility_level = "public"
  //issues_enabled = false
}
~~~

👍   though it will need docs and tests.

builtin/providers/gitlab/resource_gitlab_project.go
```diff
@@ -91,6 +95,7 @@ func resourceGitlabProjectCreate(d *schema.ResourceData, meta interface{}) error
 	client := meta.(*gitlab.Client)
 	options := &gitlab.CreateProjectOptions{
 		Name:                 gitlab.String(d.Get("name").(string)),
+		NamespaceID:          gitlab.Int(d.Get("namespace_id").(int)),
```
richardc: One change needed here, this is optional, so it needs to move from here and go with the d.GetOk("") patterned stuff:

~~~
if v, ok := d.GetOk("namespace_id"); ok {
   options.NamespaceID = gitlab.Int(v.(int))
}
~~~

richardc:
Spoke too soon, it breaks the case where a namespace_id isn't provided as you're not handling the optional nature correctly.  Notes on how to fix it on the line.

hauleth:
Fixed
--

Łukasz Niemier
lukasz@niemier.pl

stack72:
Hi @hauleth 

Thanks for the work here and thanks @richardc for the review. I merged #14490 first and now the tests pass as expected:

```
make testacc TEST=./builtin/providers/gitlab                                                                                         1 ↵ ✭
2017/05/24 13:03:02 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/gitlab -v  -timeout 120m
=== RUN   TestProvider
=== RUN   TestProvider_impl
=== RUN   TestAccGitlabDeployKey_basic
--- PASS: TestAccGitlabDeployKey_basic (36.70s)
=== RUN   TestAccGitlabGroup_basic
--- PASS: TestAccGitlabGroup_basic (21.09s)
=== RUN   TestAccGitlabProjectHook_basic
--- PASS: TestAccGitlabProjectHook_basic (31.13s)
=== RUN   TestAccGitlabProject_basic
--- PASS: TestAccGitlabProject_basic (29.59s)
=== RUN   TestGitlab_validation
--- PASS: TestGitlab_validation (0.00s)
=== RUN   TestGitlab_visbilityHelpers
--- PASS: TestGitlab_visbilityHelpers (0.00s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/gitlab	118.526s
```

One thing to note - this new param hasn't had any documentation included. Please can you follow up with that?

Thanks

Paul

PR #14495, opened 2017-05-15
[WIP] providers/heroku: import heroku_pipeline_coupling resource

### LINKED ISSUE #14491: heroku_pipeline_coupling does not support import
Feature request to support import on the heroku_pipeline_coupling resource

### Terraform Version

Terraform 0.9.6-dev

### Affected Resource(s)

heroku_pipeline_coupling

### Expected Behavior

Should import an existing pipeline coupling.

### Actual Behavior

Generates an error that import is not supported.

### Steps to Reproduce

run a `terraform import heroku_pipeline_coupling.<resource_name> <pipeline_coupling_id>`

cmorent:
Adds support for importing pipelines with the heroku_pipeline_coupling resource.

Resolves #14491

### DIFF
--- a/builtin/providers/heroku/import_heroku_pipeline_coupling_test.go
+++ b/builtin/providers/heroku/import_heroku_pipeline_coupling_test.go
@@ -1,0 +1,43 @@
+package heroku
+
+import (
+	"fmt"
+	"testing"
+
+	heroku "github.com/cyberdelia/heroku-go/v3"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccHerokuPipelineCoupling_importBasic(t *testing.T) {
+	var coupling heroku.PipelineCouplingInfoResult
+
+	appName := fmt.Sprintf("tftest-%s", acctest.RandString(10))
+	pipelineName := fmt.Sprintf("tftest-%s", acctest.RandString(10))
+	stageName := "development"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckHerokuPipelineCouplingDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckHerokuPipelineCouplingConfig_basic(appName, pipelineName, stageName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckHerokuPipelineCouplingExists("heroku_pipeline_coupling.default", &coupling),
+					testAccCheckHerokuPipelineCouplingAttributes(
+						&coupling,
+						"heroku_pipeline.default",
+						stageName,
+					),
+				),
+			},
+			{
+				ResourceName:            "heroku_pipeline_coupling.default",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"config_vars"},
+			},
+		},
+	})
+}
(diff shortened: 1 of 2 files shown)


catsby:
Hey @cmorent is this still `WIP`? And can we simply use `schema.ImportStatePassthrough` instead of a custom import method here?

cmorent:
Hey @catsby, yep, this is still `WIP`. Thanks, I did not know about this `ImportStatePassthrough`, I'll try to use it here and let you know if it works as expected !

cmorent:
Hey again @catsby, I am testing using the `schema.ImportStatePassthrough` and it works but it only imports the `id` and the `stage` (`pipeline` and `app` are missing). Do you know if there's any workaround or if I will have to use a custom  import method? Thanks in advance !

catsby:
Hey @cmorent so I dug in here a bit and found some things:

- the `d.Set` for `app` and `pipeline`are actually failing (`d.Set` returns an error, but for simple strings it's rarely checked). `p.Pipeline` and `p.App` here are actually structs. 
- the App returned in `p.App` only contains the app `ID`, but this resource tracks on the app `name`

I tinkered this morning and came up with this patch, which I think makes everything nice. It also adds a computed value `app_id` to the schema:

```diff
diff --git a/builtin/providers/heroku/resource_heroku_pipeline_coupling.go b/builtin/providers/heroku/resource_heroku_pipeline_coupling.go
index 4d513c363..cc0668af4 100644
--- a/builtin/providers/heroku/resource_heroku_pipeline_coupling.go
+++ b/builtin/providers/heroku/resource_heroku_pipeline_coupling.go
@@ -20,6 +20,10 @@ func resourceHerokuPipelineCoupling() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"app_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"app": {
 				Type:     schema.TypeString,
 				Required: true,
@@ -85,9 +89,16 @@ func resourceHerokuPipelineCouplingRead(d *schema.ResourceData, meta interface{}
 		return fmt.Errorf("Error retrieving pipeline: %s", err)
 	}
 
-	d.Set("app", p.App)
-	d.Set("pipeline", p.Pipeline)
+	// grab App info
+	app, err := client.AppInfo(context.TODO(), p.App.ID)
+	if err != nil {
+		log.Printf("[WARN] Error looking up addional App info for pipeline coupling (%s): %s", d.Id(), err)
+	}
+
+	d.Set("app_id", p.App.ID)
+	d.Set("app", app.Name)
 	d.Set("stage", p.Stage)
+	d.Set("pipeline", p.Pipeline.ID)
 
 	return nil
 }
```

With that, `schema.ImportStatePassthrough` works as expected, and the tests still pass:

catsby:
ack, of course I just noticed that we should only set the `app` if the API call to read it succeeds, so I'd put the `d.Set("app"..)` in an `else` block of the `err` check

cmorent:
Hey @catsby ! Thanks a lot for your inputs and for the time you've spent! It works perfectly with your patch and the tests for the import feature are passing too:

```
$ TF_ACC=1 go test ./builtin/providers/heroku -v -run=TestAccHerokuPipelineCoupling -timeout 120m                                            
=== RUN   TestAccHerokuPipelineCoupling_importBasic
--- PASS: TestAccHerokuPipelineCoupling_importBasic (9.40s)
=== RUN   TestAccHerokuPipelineCoupling_Basic
--- PASS: TestAccHerokuPipelineCoupling_Basic (8.24s)

PASS
ok  	github.com/cmorent/terraform/builtin/providers/heroku	17.673s
```

Also, if I try to import a `heroku_pipeline_coupling` resource now:

```
$ terraform state show heroku_pipeline_coupling.foobar <pipeline_coupling_id>
id       = <heroku_pipeline_coupling_id>
app      = <heroku_pipeline_coupling_app>
app_id   = <heroku_app_id>
pipeline = <heroku_pipeline_coupling_pipeline>
stage    = <heroku_pipeline_coupling_stage>
```

Which is exactly what we expect it to be !

catsby:
Looks good, thanks!

PR #14514, opened 2017-05-15
Add -update flag to init command

antonosmond:
Add the `-update` flag to the `init` command so modules can be updated if necessary without the need to use `terraform get`


antonosmond:
Closing in favour of https://github.com/hashicorp/terraform/pull/13852

Issue #14175, opened 2017-05-03
Flip the Scaleway IP -> Server relationship to Server -> IP


tinoadams:
### Terraform Version
Terraform v0.9.4

### Affected Resource(s)
- scaleway_ip
- scaleway_server

### Terraform Configuration Files
```hcl
variable "gateway_count" {
  default = 2
}
resource "scaleway_ip" "public_ip" {
  count = "${var.gateway_count}"
  server = "${element(scaleway_server.gateway.*.id, count.index)}"
}
resource "scaleway_server" "gateway" {
  count = "${var.gateway_count}"
  name  = "gateway_${count.index}"
  image = "89457135-d446-41ba-a8df-d53e5bb54710"
  type  = "C2S"
  connection {
    type         = "ssh"
    user         = "root"
    agent        = true
  }
  provisioner "remote-exec" {
    inline = [
      "echo 'SUCCESS'",
    ]
  }
}
```

### Expected Behavior
The public IP should be created, then the server, then the IP attached to the server and finally the provisioner executed.

Flipping the relationship direction to Server -> IP also means I could potentially provide the server with an existing IP (not creating a resource via Terraform). Quite useful as Scaleway IP's are reserved and can be re-used.

### Actual Behavior
The relationship between the two resources scaleway_ip and scaleway_server is such that the server is created before the IP ie. the IP is only attached after the Server resource is completely created/provisioned.

The provisioner never manages to connect as the public IP is not attached yet.

### Steps to Reproduce
terraform apply

nicolai86:
@tinoadams thank you for reporting this. 

I'm interested: why not set the `dynamic_ip_required` attribute to get a public IP for provisioning, and use `scaleway_ip` too to change the IP after provisioning? 

```
variable "gateway_count" {
  default = 2
}
resource "scaleway_ip" "public_ip" {
  count = "${var.gateway_count}"
  server = "${element(scaleway_server.gateway.*.id, count.index)}"
}
resource "scaleway_server" "gateway" {
  dynamic_ip_required = true # so provisioning succeeds
  count = "${var.gateway_count}"
  name  = "gateway_${count.index}"
  image = "89457135-d446-41ba-a8df-d53e5bb54710"
  type  = "C2S"
  connection {
    type         = "ssh"
    user         = "root"
    agent        = true
  }
  provisioner "remote-exec" {
    inline = [
      "echo 'SUCCESS'",
    ]
  }
}
```

stack72:
Closed via #14515

PR #14515, opened 2017-05-15
provider/scaleway: allow public_ip to be set on server resource

nicolai86:
this allows the `scaleway_ip` <-> `scaleway_server` relationship to be inverted as requested by #14175.

You can now assign `scaleway_server` and `scaleway_ip` in two ways:

### server dependency

```
resource "scaleway_ip" "base" {}

resource "scaleway_server" "base" {
  name = "test"
  # ubuntu 14.04
  image = "5faef9cd-ea9b-4a63-9171-9e26bec03dbc"
  type = "C1"
  tags = [ "terraform-test" ]
  public_ip = "${scaleway_ip.base.ip}"
}
```

### ip dependency

```
resource "scaleway_ip" "base" {
  server = "${scaleway_server.base.id}
}

resource "scaleway_server" "base" {
  name = "test"
  # ubuntu 14.04
  image = "5faef9cd-ea9b-4a63-9171-9e26bec03dbc"
  type = "C1"
  tags = [ "terraform-test" ]
}
```

the "server dependency" is important when you want to provision a server which should receive a static ip, like a gateway (hence the `scaleway_ip` resource and not the `dynamic_ip_required` attribute).

previously this would break the provider provisioning as servers with private IPs are not accessible outside the Scaleway network.

@tinoadams would be great to get your feedback on if this solves your issue.

### DIFF
--- a/builtin/providers/scaleway/resource_scaleway_ip.go
+++ b/builtin/providers/scaleway/resource_scaleway_ip.go
@@ -21,6 +21,7 @@ func resourceScalewayIP() *schema.Resource {
 			"server": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 			"ip": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


stack72:
Nice addition @nicolai86 :) LGTM!

tinoadams:
Excellent @nicolai86, great solution. Will take it for a spin tonight.

PR #14551, opened 2017-05-16
provider/opc: add missing documentation back from opc provider

grubernaut:
Documentation needs to be kept in the main repository until we're able to split out documentation into separate webpages


stack72:
that wasn't deleted in the first place

PR #14563, opened 2017-05-17
rancher_environment: add member support

raphink:
This PR adds member support to rancher_environments, eg:

```
resource rancher_environment "test" {
  name           = "test"

  member {
    external_id = "650430"
    external_id_type = "github_user"
    role = "owner"
  }

  member {
    external_id = "1234"
    external_id_type = "github_team"
    role = "owner"
  }
}
```

### DIFF
--- a/builtin/providers/rancher/resource_rancher_environment.go
+++ b/builtin/providers/rancher/resource_rancher_environment.go
@@ -1,11 +1,13 @@
 package rancher
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"strings"
 	"time"
 
+	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
@@ -40,6 +42,28 @@ func resourceRancherEnvironment() *schema.Resource {
 			"description": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+			},
+			"member": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"external_id_type": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"external_id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"role": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+				Set: memberHash,
 			},
 		},
 	}
@@ -85,6 +109,19 @@ func resourceRancherEnvironmentCreate(d *schema.ResourceData, meta interface{}) 
 	d.SetId(newEnv.Id)
 	log.Printf("[INFO] Environment ID: %s", d.Id())
 
+	// Add members
+	envClient, err := meta.(*Config).EnvironmentClient(d.Id())
+	if err != nil {
+		return err
+	}
+	members := d.Get("members").([]interface{})
+	_, err = envClient.Project.ActionSetmembers(&newEnv, &rancherClient.SetProjectMembersInput{
+		Members: members,
+	})
+	if err != nil {
+		return err
+	}
+
 	return resourceRancherEnvironmentRead(d, meta)
 }
 
@@ -118,6 +155,14 @@ func resourceRancherEnvironmentRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("name", env.Name)
 	d.Set("orchestration", getActiveOrchestration(env))
 
+	envClient, err := meta.(*Config).EnvironmentClient(d.Id())
+	if err != nil {
+		return err
+	}
+
+	members, _ := envClient.ProjectMember.List(NewListOpts())
+
+	d.Set("member", normalizeMembers(members.Data))
 	return nil
 }
 
@@ -145,6 +190,19 @@ func resourceRancherEnvironmentUpdate(d *schema.ResourceData, meta interface{}) 
 	}
 
 	if err := client.Update("project", &env.Resource, data, &newEnv); err != nil {
+		return err
+	}
+
+	// Update members
+	envClient, err := meta.(*Config).EnvironmentClient(d.Id())
+	if err != nil {
+		return err
+	}
+	members := d.Get("member").(*schema.Set).List()
+	_, err = envClient.Project.ActionSetmembers(&newEnv, &rancherClient.SetProjectMembersInput{
+		Members: makeProjectMembers(members),
+	})
+	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


builtin/providers/rancher/resource_rancher_environment.go
```diff
@@ -85,6 +109,19 @@ func resourceRancherEnvironmentCreate(d *schema.ResourceData, meta interface{})
 	d.SetId(newEnv.Id)
 	log.Printf("[INFO] Environment ID: %s", d.Id())
 
+	// Add members
+	envClient, err := meta.(*Config).EnvironmentClient(d.Id())
+	if err != nil {
+		return err
+	}
+	members := d.Get("members").([]interface{})
```
grubernaut: This should also be `d.GetOk("member")`, there's no attribute called `members`

builtin/providers/rancher/resource_rancher_environment.go
```diff
@@ -203,6 +261,39 @@ func setOrchestrationFields(orchestration string, data map[string]interface{}) {
 	data[orch] = true
 }
 
+func normalizeMembers(in []rancherClient.ProjectMember) (out []interface{}) {
+	for _, m := range in {
```
grubernaut: should we initialize `out` here?

builtin/providers/rancher/resource_rancher_environment.go
```diff
@@ -203,6 +261,39 @@ func setOrchestrationFields(orchestration string, data map[string]interface{}) {
...
+			"external_id":      m.ExternalId,
+			"role":             m.Role,
+		}
+		out = append(out, mm)
+	}
+	return
+}
+
+func makeProjectMembers(in []interface{}) (out []interface{}) {
+	for _, m := range in {
```
grubernaut: Should we initialize `out` here?

builtin/providers/rancher/resource_rancher_environment.go
```diff
@@ -203,6 +261,39 @@ func setOrchestrationFields(orchestration string, data map[string]interface{}) {
...
+			ExternalIdType: mMap["external_id_type"].(string),
+			ExternalId:     mMap["external_id"].(string),
+			Role:           mMap["role"].(string),
+		}
+		out = append(out, mm)
+	}
+	return
+}
+
+func memberHash(member interface{}) int {
```
grubernaut: Since all the attributes of the `member` set are required, do we need a custom `Set` function, or does it work with the default `Set` function?

builtin/providers/rancher/resource_rancher_environment.go
raphink: As far as I can tell, not specifying any member works fine.

builtin/providers/rancher/resource_rancher_environment.go
raphink: `out` needs to be initialized outside the loop, lest it is not available as a return value after the loop is finished.

grubernaut: Aye, but it's not initialized :smile:

builtin/providers/rancher/resource_rancher_environment.go
grubernaut: Okay, can we add both an acceptance test for the addition of multiple `member` blocks, as well as verification that the current acceptance tests still pass with this changeset?

builtin/providers/rancher/resource_rancher_environment.go
raphink: Yes, it's quite possible that `schema.HashString` would suffice.

builtin/providers/rancher/resource_rancher_environment.go
raphink: Sure.

builtin/providers/rancher/resource_rancher_environment.go
raphink: It's initialized on line 276, to an empty `[]interface{}`.

builtin/providers/rancher/resource_rancher_environment.go
raphink: Interestingly, not specifying members seems to fail in the acceptance tests, but not when using terraform plan/apply.

builtin/providers/rancher/resource_rancher_environment.go
raphink: Actually, since members are resources and not hashes, it seems I need a Set function.

raphink:
I wrote some acceptance tests, but they're currently failing because access control is required to test this.

builtin/providers/rancher/resource_rancher_environment.go
```diff
@@ -85,6 +109,21 @@ func resourceRancherEnvironmentCreate(d *schema.ResourceData, meta interface{})
 	d.SetId(newEnv.Id)
 	log.Printf("[INFO] Environment ID: %s", d.Id())
 
+	// Add members
+	envClient, err := meta.(*Config).EnvironmentClient(d.Id())
```
grubernaut: Do you need to create the `envClient` here if you're not setting members?

builtin/providers/rancher/resource_rancher_environment.go
grubernaut: If the `Set` key in the resource struct isn't set, terraform core should be able to form the correct hash string for each `members` object. Can you elaborate on the issue you're seeing from using this default `Set` function?

raphink: Ah ok, I can try that.

builtin/providers/rancher/resource_rancher_environment.go
raphink: Good point.

raphink: Changed that now.

builtin/providers/rancher/resource_rancher_environment.go
```diff
@@ -85,6 +106,21 @@ func resourceRancherEnvironmentCreate(d *schema.ResourceData, meta interface{})
 	d.SetId(newEnv.Id)
 	log.Printf("[INFO] Environment ID: %s", d.Id())
 
+	// Add members
+	if v, ok := d.GetOk("members"); ok {
```
grubernaut: `members` is still not an attribute. This needs to be `d.GetOk("member")`

website/source/docs/providers/rancher/r/environment.html.md
```diff
@@ -28,6 +40,7 @@ The following arguments are supported:
 * `name` - (Required) The name of the environment.
 * `description` - (Optional) An environment description.
 * `orchestration` - (Optional) Must be one of **cattle**, **swarm**, **mesos** or **kubernetes**. Defaults to **cattle**.
+* `member` - (Optional) Members to add to the environment. Each member has `external_id`, `external_id_type` and `role` values.
```
grubernaut: Would be helpful to add a section before the "Attributes Reference" with more details on what each of the parameters accepted inside the `member` block is and what their type is.

builtin/providers/rancher/resource_rancher_environment.go
raphink: Fixed

website/source/docs/providers/rancher/r/environment.html.md
raphink: I added a subsection for Member Parameter Reference.

grubernaut:
Thanks for your patience and effort on this @raphink. Looks good!

raphink:
Thank you @gubernaut! Do you think you could have a look at my other PRs? Some have been stalled without news for weeks...

SerialVelocity:
@raphink Your docs say that `member` is optional but when I don't specify it, I get:
```
rancher_environment.main: Creating...
  description:         "" => "Main environment"
  member.#:            "" => "<computed>"
  name:                "" => "Main"
  orchestration:       "" => "cattle"
  project_template_id: "" => "<computed>"
Error applying plan:

1 error(s) occurred:

* rancher_environment.main: 1 error(s) occurred:

* rancher_environment.main: Bad response statusCode [422]. Status [422 status code 422]. Body: [code=MissingRequired, fieldName=members, baseType=error] from [https://X/v2-beta/projects/1a7/?action=setmembers]

```

raphink:
@SerialVelocity could you please open a a new bug on the rancher provider repo? https://github.com/terraform-providers/terraform-provider-rancher/issues

PR #14573, opened 2017-05-17
Add ability to import Google Compute persistent disks

bashtoni:
```
$ make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccComputeDisk_importBasic'
2017/05/17 14:44:27 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccComputeDisk_importBasic -timeout 120m
=== RUN   TestAccComputeDisk_importBasic
--- PASS: TestAccComputeDisk_importBasic (26.42s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	26.434s
```

### DIFF
--- a/builtin/providers/google/resource_compute_disk.go
+++ b/builtin/providers/google/resource_compute_disk.go
@@ -234,12 +234,12 @@ func resourceComputeDiskRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("disk_encryption_key_sha256", disk.DiskEncryptionKey.Sha256)
 	}
 	if disk.SourceImage != "" {
-		imageUrl := strings.Split(disk.SourceImage, "/")
-		d.Set("image", imageUrl[len(imageUrl)-1])
+		imageUrlParts := strings.Split(disk.SourceImage, "/")
+		d.Set("image", imageUrlParts[len(imageUrlParts)-1])
 	}
 	if disk.SourceSnapshot != "" {
-		snapshotUrl := strings.Split(disk.SourceSnapshot, "/")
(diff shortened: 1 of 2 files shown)


builtin/providers/google/resource_compute_disk.go
```diff
@@ -171,16 +175,54 @@ func resourceComputeDiskRead(d *schema.ResourceData, meta interface{}) error {
...
 		d.Set("disk_encryption_key_sha256", disk.DiskEncryptionKey.Sha256)
 	}
+	if disk.SourceImage != "" {
+		imageUrl := strings.Split(disk.SourceImage, "/")
+		d.Set("image", imageUrl[len(imageUrl)-1])
+	}
+	if disk.SourceSnapshot != "" {
+		snapshotUrl := strings.Split(disk.SourceSnapshot, "/")
+		d.Set("snapshot", snapshotUrl[len(snapshotUrl)-1])
+	}
```
radeksimko: It may sound like nitpick and it may be just me, but I spent good couple of minutes trying to understand what all these changes above (with `strings.Split`) mean.
So just for clarity/readability I think it would help if we renamed `zoneUrl` to `zoneUrlParts`, `typeUrl` to `typeUrlParts` etc. so that it's more obvious that these variables hold slice, not a string?

bashtoni:
@radeksimko Reworded as suggested, fixed conflicts.

Thanks for reviewing!

radeksimko:
@bashtoni Thanks for making those changes, would you mind changing the last two variables too? i.e. `imageUrl` and `snapshotUrl`. Then I'm happy to click the big green button and 🚢  it 😃

Issue #14577, opened 2017-05-17
Add support for certificates to DigitalOcean provider


caiofilipini:
The [original PR](https://github.com/hashicorp/terraform/pull/12077) that introduced support for DigitalOcean's load balancers was missing support for TLS certificates.

https://www.digitalocean.com/community/tutorials/how-to-configure-ssl-termination-on-digitalocean-load-balancers

https://developers.digitalocean.com/documentation/v2/#certificates

PR #14578, opened 2017-05-17
provider/digitalocean: Add support for certificates

caiofilipini:
This PR addresses that by introducing the `resource_digitalocean_certificate`.

Fixes https://github.com/hashicorp/terraform/issues/14577.

A few things to note:

1) I've upgraded [godo](https://github.com/digitalocean/godo) to get support for [Certificate](https://godoc.org/github.com/digitalocean/godo#Certificate). This upgrade bought a few breaking changes which I've addressed in https://github.com/caiofilipini/terraform/commit/6681a86211666fef6a4b0daf1ca741ac6b3b01b6 (i.e. every call now takes a `context.Context`, and a few things got deprecated/removed).
2) I've introduced a `RandTLSCert` function in `helper/acctest/random.go` that generates a valid, self-signed TLS certificate so that we can properly test the integration.
3) I was getting a `Step 0 error: After applying this step, the plan was not empty:` while running the new tests and I couldn't really figure it out. I think it's because the attributes that DO's API returns for a certificate are different from the ones we use fore creating one. For that reason, I've set `ExpectNonEmptyPlan` to `true`, but I'd love some guidance here.

```
$ make testacc DIGITALOCEAN_TOKEN=XXX TEST=./builtin/providers/digitalocean TESTARGS='-run=TestAccDigitalOceanCertificate'
2017/05/17 17:54:50 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/digitalocean -v -run=TestAccDigitalOceanCertificate -timeout 120m
=== RUN   TestAccDigitalOceanCertificate_Basic
--- PASS: TestAccDigitalOceanCertificate_Basic (6.84s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/digitalocean   6.867s
```

### DIFF
--- a/builtin/providers/digitalocean/resource_digitalocean_certificate.go
+++ b/builtin/providers/digitalocean/resource_digitalocean_certificate.go
@@ -44,12 +44,14 @@ func resourceDigitalOceanCertificate() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				Computed: true,
 			},
 
 			"sha1_fingerprint": {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				Computed: true,
 			},
 		},
 	}
--- a/builtin/providers/digitalocean/resource_digitalocean_certificate.go
+++ b/builtin/providers/digitalocean/resource_digitalocean_certificate.go
@@ -1,0 +1,119 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceDigitalOceanCertificate() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceDigitalOceanCertificateCreate,
+		Read:   resourceDigitalOceanCertificateRead,
+		Delete: resourceDigitalOceanCertificateDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"private_key": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"leaf_certificate": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"certificate_chain": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"not_after": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"sha1_fingerprint": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func buildCertificateRequest(d *schema.ResourceData) (*godo.CertificateRequest, error) {
+	req := &godo.CertificateRequest{
+		Name:             d.Get("name").(string),
+		PrivateKey:       d.Get("private_key").(string),
+		LeafCertificate:  d.Get("leaf_certificate").(string),
+		CertificateChain: d.Get("certificate_chain").(string),
+	}
+
+	return req, nil
+}
+
+func resourceDigitalOceanCertificateCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*godo.Client)
+
+	log.Printf("[INFO] Create a Certificate Request")
+
+	certReq, err := buildCertificateRequest(d)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Certificate Create: %#v", certReq)
+	cert, _, err := client.Certificates.Create(context.Background(), certReq)
+	if err != nil {
+		return fmt.Errorf("Error creating Certificate: %s", err)
+	}
+
+	d.SetId(cert.ID)
+
+	return resourceDigitalOceanCertificateRead(d, meta)
+}
+
+func resourceDigitalOceanCertificateRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*godo.Client)
+
+	log.Printf("[INFO] Reading the details of the Certificate %s", d.Id())
+	cert, _, err := client.Certificates.Get(context.Background(), d.Id())
+	if err != nil {
+		return fmt.Errorf("Error retrieving Certificate: %s", err)
+	}
+
+	d.Set("name", cert.Name)
+	d.Set("not_after", cert.NotAfter)
+	d.Set("sha1_fingerprint", cert.SHA1Fingerprint)
+
+	return nil
+
+}
+
+func resourceDigitalOceanCertificateDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*godo.Client)
+
+	log.Printf("[INFO] Deleting Certificate: %s", d.Id())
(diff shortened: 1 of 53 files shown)


caiofilipini:
Just realized this is missing updates to docs, so marking it as WIP.

caiofilipini:
Docs added in https://github.com/hashicorp/terraform/pull/14578/commits/a565001014438f42aa8ecdf330767f445e735c70. Removing [WIP].

builtin/providers/digitalocean/resource_digitalocean_certificate.go
```diff
@@ -0,0 +1,119 @@
...
+
+	log.Printf("[INFO] Reading the details of the Certificate %s", d.Id())
+	cert, _, err := client.Certificates.Get(context.Background(), d.Id())
+	if err != nil {
+		return fmt.Errorf("Error retrieving Certificate: %s", err)
+	}
+
+	d.Set("name", cert.Name)
+	d.Set("not_after", cert.NotAfter)
+	d.Set("sha1_fingerprint", cert.SHA1Fingerprint)
```
grubernaut: Are the other attributes listed in the schema not available in the `cert` object? These will need to be set as well so Terraform can detect any configuration drift between config/state.

builtin/providers/digitalocean/resource_digitalocean_certificate.go
```diff
@@ -0,0 +1,119 @@
...
+func resourceDigitalOceanCertificateDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*godo.Client)
+
+	log.Printf("[INFO] Deleting Certificate: %s", d.Id())
+	_, err := client.Certificates.Delete(context.Background(), d.Id())
+	if err != nil {
+		return fmt.Errorf("Error deleting Certificate: %s", err)
+	}
+
+	d.SetId("")
```
grubernaut: The `Delete` method doesn't need to set the ID to nil, as a non-error return from the Delete method already does this. However, it's not critical to remove, just noting for future knowledge :smile:

grubernaut:
Also, as a follow-up, what attributes were changing in the acceptance test that causes the need for setting `ExpectNonEmptyPlan` to be `true`? We'll likely need to fix this issue prior to merging, as this will likely cause perpetual diff issues for users.

caiofilipini:
Hey, @grubernaut, thanks a lot for the feedback! I'll address it soon and let you know when I think this is ready for a second pass.

builtin/providers/digitalocean/resource_digitalocean_certificate.go
caiofilipini: Cool, I'm gonna remove it anyway. Thanks for the tip!

builtin/providers/digitalocean/resource_digitalocean_certificate.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+}
+
+func buildCertificateRequest(d *schema.ResourceData) (*godo.CertificateRequest, error) {
+	req := &godo.CertificateRequest{
+		Name:             d.Get("name").(string),
+		PrivateKey:       d.Get("private_key").(string),
+		LeafCertificate:  d.Get("leaf_certificate").(string),
+		CertificateChain: d.Get("certificate_chain").(string),
```
caiofilipini: @grubernaut AFAIK, there are no undesired effects to `certificate_chain` being nil. But I'd rather play safe here, so I'm gonna add a check around it.

caiofilipini: Although I just checked the docs for `Get` on `ResourceData` and it seems to be safe for `TypeString` (it just returns the zero value, or `""`). I think it's safe to leave it as is.

builtin/providers/digitalocean/resource_digitalocean_certificate.go
caiofilipini: @grubernaut Correct, they're available only for creation, but not returned by the API (for security reasons, I assume).

This is the example response from the [API docs](https://developers.digitalocean.com/documentation/v2/#certificates):

```json
{
  "certificate": {
    "id": "892071a0-bb95-49bc-8021-3afd67a210bf",
    "name": "web-cert-01",
    "not_after": "2017-02-22T00:23:00Z",
    "sha1_fingerprint": "dfcc9f57d86bf58e321c2c6c31c7a971be244ac7",
    "created_at": "2017-02-08T16:02:37Z"
  }
}
```

caiofilipini:
@grubernaut I collected the log from the failing test after removing `ExpectNonEmptyPlan` from it.

https://gist.github.com/caiofilipini/f0628fffd39abb631f2f374d62e4a057

Let me know if you have any insights so we can move this forward. Thank you!

builtin/providers/digitalocean/resource_digitalocean_certificate.go
grubernaut: That's correct about `Get` returning the `nil` type for the native golang string type, however, does that have any adverse effect during the Create request in the DO API? 
IE: will setting `certificate_chain` to `""`, cause the API request to attempt to set `certificate_chain` to `""`?

caiofilipini: @grubernaut It doesn't seem to have any adverse effects.

grubernaut:
Hey @caiofilipini, if you add `Computed: true,` to both `sha1_fingerprint` and `next_after`, the test should pass without the config/state diff that you're experiencing.

website/source/docs/providers/do/r/certificate.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+
+* `name` - (Required) The name of the certificate for identification.
+* `private_key` - (Required) The contents of a PEM-formatted private-key
+corresponding to the SSL certificate.
+* `leaf_certificate` - (Required) The contents of a PEM-formatted public
+TLS certificate.
+* `certificate_chain` - (Optional) The full PEM-formatted trust chain
+between the certificate authority's certificate and your domain's TLS
+certificate.
+
```
grubernaut: Are `not_after` and `sha1_fingerprint` not allowed to be specified? In the resource schema, they're listed as `Optional`

caiofilipini: @grubernaut They're not recognized (ignored) by the API and always computed based on the input. I wasn't sure how to best represent them so they would make it to the resulting state, but without them being required in the request. Is this the way to go, or is there a better way?

caiofilipini:
@grubernaut `Computed: true` indeed fixed it, thank you! 💯

website/source/docs/providers/do/r/certificate.html.markdown
grubernaut: @caiofilipini ah okay awesome. If they aren't accepted inputs to the API request, we can remove `Optional` and `ForceNew` and the attributes will simply be `Computed` attributes that we set inside the `Read` method.

caiofilipini: @grubernaut Ah, that makes sense. Done in https://github.com/hashicorp/terraform/pull/14578/commits/c4bbb6e8a554c004841e86f63f7f0727b07ce1cd.

grubernaut: @caiofilipini Awesome, ready to review?

caiofilipini: @grubernaut Yeah! 😄

caiofilipini:
Thanks for all your help, @grubernaut!

PR #14596, opened 2017-05-17
provider/aws: Tag VPCs created in our tests to help track down leaks

catsby:
The tags must flow


catsby:
I'm going to merge this... it only touches AWS test files and only changes resources that create a VPC with no tags, so I'm assuming there are no consequences of adding them. I'll assume the responsibly of fixing any fallout from our nightly ACC test run

PR #14598, opened 2017-05-17
Add power function

pradeepbhadani:
Ad mathematics power function

### DIFF
--- a/config/interpolate_funcs_test.go
+++ b/config/interpolate_funcs_test.go
@@ -310,6 +310,54 @@ func TestInterpolateFuncMin(t *testing.T) {
 	})
 }
 
+func TestInterpolateFuncPow(t *testing.T) {
+	testFunction(t, testFunctionConfig{
+		Cases: []testFunctionCase{
+			{
+				`${pow(1, 0)}`,
+				"1",
+				false,
+			},
+			{
+				`${pow(1, 1)}`,
+				"1",
+				false,
+			},
+
+			{
+				`${pow(2, 0)}`,
+				"1",
+				false,
+			},
+			{
+				`${pow(2, 1)}`,
+				"2",
+				false,
+			},
+			{
+				`${pow(3, 2)}`,
+				"9",
+				false,
+			},
+			{
+				`${pow(-3, 2)}`,
+				"9",
+				false,
+			},
+			{
+				`${pow(2, -2)}`,
+				"0.25",
+				false,
+			},
+			{
+				`${pow(0, 2)}`,
+				"0",
+				false,
+			},
+		},
+	})
+}
+
 func TestInterpolateFuncFloor(t *testing.T) {
 	testFunction(t, testFunctionConfig{
 		Cases: []testFunctionCase{
--- a/config/interpolate_funcs_test.go
+++ b/config/interpolate_funcs_test.go
@@ -354,6 +354,21 @@ func TestInterpolateFuncPow(t *testing.T) {
 				"0",
 				false,
 			},
+			{
+				`${pow("invalid-input", 2)}`,
+				nil,
+				true,
+			},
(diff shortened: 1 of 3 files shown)


radeksimko:
Hi @pradeepbhadani 
thank you for the contribution.

Would you mind explaining a real-world scenario where this function would be useful in the context of a Terraform config?

Thanks.

pradeepbhadani:
@radeksimko this might be useful when I need to update a number with base 2. or create number od resources with base 2.

config/interpolate_funcs_test.go
```diff
@@ -310,6 +310,54 @@ func TestInterpolateFuncMin(t *testing.T) {
...
+			{
+				`${pow(2, -2)}`,
+				"0.25",
+				false,
+			},
+			{
+				`${pow(0, 2)}`,
+				"0",
+				false,
+			},
```
grubernaut: Can you add some failing test cases as well? 
- only one parameter specific in the function should fail
- a non-int parameter specified should fail

website/source/docs/configuration/interpolation.html.md
```diff
@@ -312,6 +312,12 @@ The supported built-in functions are:
   * `pathexpand(string)` - Returns a filepath string with `~` expanded to the home directory. Note:
     This will create a plan diff between two different hosts, unless the filepaths are the same.
 
+  * `pow(x, y)` - Returns the base `x` of exponential `y`.
```
grubernaut: Note here that this function accepts floats

config/interpolate_funcs_test.go
pradeepbhadani: @grubernaut  I will add some failing test and will update the function to only accept integer.

website/source/docs/configuration/interpolation.html.md
pradeepbhadani: @grubernaut from the above comment, it looks like support for float is not required here.

config/interpolate_funcs_test.go
grubernaut: The function should fail if supplied a string, for instance. 
```
{
  `${pow("foo", "bar")}`,
  nil,
  true,
},
```

website/source/docs/configuration/interpolation.html.md
grubernaut: The `math.Pow()` function requires two `float64` numbers. And inside the interpolation function you're casting the supplied interfaces to `float64`. This will also work on integers, but we should document that the function takes in floats.

config/interpolate_funcs_test.go
pradeepbhadani: @grubernaut added more test.

website/source/docs/configuration/interpolation.html.md
pradeepbhadani: added note in documentation

config/interpolate_funcs_test.go
pradeepbhadani: make output 
`$ make test TEST='./config' TESTARGS='-run=TestInterpolateFuncPow'
2017/05/18 14:28:43 Generated command/internal_plugin_list.go
go test -i ./config || exit 1
echo ./config | \
		xargs -t -n4 go test -run=TestInterpolateFuncPow -timeout=60s -parallel=4
go test -run=TestInterpolateFuncPow -timeout=60s -parallel=4 ./config 
ok  	github.com/hashicorp/terraform/config	0.007s
`

pradeepbhadani:
@grubernaut  Looks good now?

pradeepbhadani:
@grubernaut  Thanks for merging. 
I missed commit to add `note` in documentation. 
I have raised a new PR #14629 for this.

PR #14612, opened 2017-05-18
Update news section with YAML data file

ryon:
This makes the news section a little more dynamic, and updates it with the latest webinar info from Amanda McLeod.


stack72:
screen shot 2017-05-18 at 18 49 10

PR #14621, opened 2017-05-18
Update supported AWS regions

gsaslis:
According to an AWS announcement from yesterday: 
https://aws.amazon.com/about-aws/whats-new/2017/05/amazon-lightsail-now-available-in-5-additional-europe-us-aws-regions/


stack72:
LGTM! Thanks for this :)

PR #14624, opened 2017-05-18
provider/aws: Raise timeout for attaching/detaching VPN GW

radeksimko:
Apparently [raising it to 5mins](https://github.com/hashicorp/terraform/pull/13457) wasn't sufficient as we keep seeing failures like this:

```
=== RUN   TestAccAWSRouteTable_vgwRoutePropagation
--- FAIL: TestAccAWSRouteTable_vgwRoutePropagation (702.17s)
    testing.go:280: Step 0 error: Error applying: 1 error(s) occurred:
        
        * aws_vpn_gateway.foo: 1 error(s) occurred:
        
        * aws_vpn_gateway.foo: Error waiting for VPN gateway (vgw-a0c71cbe) to attach: timeout while waiting for state to become 'attached' (last state: 'detached', timeout: 5m0s)
FAIL
```

### DIFF
--- a/builtin/providers/aws/resource_aws_vpn_gateway.go
+++ b/builtin/providers/aws/resource_aws_vpn_gateway.go
@@ -205,7 +205,7 @@ func resourceAwsVpnGatewayAttach(d *schema.ResourceData, meta interface{}) error
 		Pending: []string{"detached", "attaching"},
 		Target:  []string{"attached"},
 		Refresh: vpnGatewayAttachStateRefreshFunc(conn, d.Id(), "available"),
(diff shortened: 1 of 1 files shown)


radeksimko:
I see your point, but I think that all of our resources should have high enough default timeouts that make all acceptance tests pass regularly. I see the Timeout feature as a solution for resources which may take significantly more time to create/update/delete under certain "non-conventional" conditions (like enormous number of resources and/or enormously big instance types).

In other words I'd like all users to be able to use all resources without having to specify `timeouts` if their use cases are conventional.

stack72:
agreed :)

PR #14630, opened 2017-05-18
google container_cluster master_auth should be optional

dwradcliffe:
I'm not sure why master_auth is required; it's not even an option when you create a cluster via the console today.
This should make it optional.

### DIFF
--- a/builtin/providers/google/resource_container_cluster.go
+++ b/builtin/providers/google/resource_container_cluster.go
@@ -25,7 +25,7 @@ func resourceContainerCluster() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"master_auth": &schema.Schema{
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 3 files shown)


builtin/providers/google/resource_container_cluster.go
```diff
@@ -342,19 +342,21 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
...
+		Name:             clusterName,
+		InitialNodeCount: int64(d.Get("initial_node_count").(int)),
 	}
-	masterAuth := masterAuths[0].(map[string]interface{})
 
-	cluster := &container.Cluster{
-		MasterAuth: &container.MasterAuth{
+	if v, ok := d.GetOk("master_auth"); ok {
+		masterAuths := v.([]interface{})
+		if len(masterAuths) > 1 {
```
danawillow: While you're here, would you mind removing this check and instead putting MaxItems: 1 in the schema defined above? That'll move the check from the apply phase to the plan phase.

builtin/providers/google/resource_container_cluster.go
```diff
@@ -25,8 +25,9 @@ func resourceContainerCluster() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"master_auth": &schema.Schema{
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
+				MaxItems: 1,
```
danawillow: Whoops, missed this earlier- you'll also need to set this to computed to get tests without master_auth to pass, since the server does return master_auth info.

dwradcliffe: Why do the tests pass now? 😕

danawillow: Are you running the acceptance tests or are you just looking at the travis build output? The build passes but the acceptance tests didn't when I ran them (https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#running-an-acceptance-test)

dwradcliffe: Ohh right, forgot about that.

PR #14633, opened 2017-05-18
Missing short name in the service scope (Google compute instance)

smasue:
The missing short name is for Stackdriver Trace append.


catsby:
Thanks @smasue ! Do you think we should include `https://www.googleapis.com/auth/trace.readonly` as well?

smasue:
hello @catsby ! You are right I am going to add the line.

PR #14634, opened 2017-05-18
Mark the aws & gcs secrets as sensitive

mtougeron:
This PR marks the aws `s3_access_key` & `s3_secret_key` and the gcs `secret_key` as `Sensitive` fields.


catsby:
I can't imagine why those would be exported or reused, so yeah this makes sense. hopefully I'm not way off the mark there 😄

catsby:
Thanks!

mtougeron:
Thanks!

Issue #8667, opened 2016-09-06
Terraform destroy not deleting the resource 'google_compute_disk' intermittently


pjayaraj:
Terraform v0.7.0 has issues with terraform destroy with resource type 'google_compute_disk'

Please list the resources as a list, for example:
- terraform destroy fails intermittently  if it has resource of type 'google_compute_disk'
### Terraform Configuration Files

```
resource "google_compute_disk" "envprigfsnc18-glusterfs-disk-0" {
          name  = "envprigfsnc18-disk-0-${count.index}"
          type  = "pd-ssd"
          zone = "${element(split(",", var.zones), count.index)}"
          size  = "16"
          count = "3"
          lifecycle {
         prevent_destroy = false
          }
        }

  resource "google_compute_instance" "envprigfsnc18-glusterfs_server-0" {
          name = "envprigfsnc18-glusterfs-server-0-${count.index}"
          machine_type = "n1-standard-8"
          count = "3"
          zone = "${element(split(",", var.zones), count.index)}"
          metadata_startup_script="${element(data.template_file.envprigfsnc18-glusterfs-server_user_data-0.*.rendered,count.index)}"

          disk {
            image = "glusterfs-server-3-8-2-201608311147"
            size  = "90"
          }
          depends_on = ["google_compute_disk.envprigfsnc18-glusterfs-disk-0"]
          disk {
            disk = "${element(google_compute_disk.envprigfsnc18-glusterfs-disk-0.*.name, count.index)}"
            device_name = "${element(google_compute_disk.envprigfsnc18-glusterfs-disk-0.*.name, count.index)}"
            auto_delete = false
          }
          network_interface {
            subnetwork = "${var.datazone_subnet_ids}"
          }
          metadata {
           role = "glusterfs_server"
          }
          service_account {
            scopes = ["userinfo-email", "compute-ro", "storage-full"]
          }
          lifecycle {
             ignore_changes = ["metadata_startup_script"]
          }
        }

```
### Debug Output

terraform destroy
Do you really want to destroy?
  Terraform will delete all your managed infrastructure.
  There is no undo. Only 'yes' will be accepted to confirm.

  Enter a value: yes

google_compute_disk.envprigfsnc18-glusterfs-disk-1.2: Refreshing state... (ID: envprigfsnc18-disk-1-2)
data.template_file.consul_server_bootstrap_user_data: Refreshing state...
google_compute_disk.envprigfsnc18-glusterfs-disk-1.2: Destroying...
[...]

PR #14651, opened 2017-05-19
provider/google: detach disks before deleting them.

paddycarver:
When a `google_compute_disk` is attached to a `google_compute_instance`,
deleting can be tricky. GCP doesn't allow disks that are attached to
instances to be deleted. Normally, this is fine; the instance depends on
the disk, so by the time the disk is deleted, the instance should
already be gone.

However, some reports have cropped up (#8667) that deleting disks is
failing because they're still attached to instances. Though this
shouldn't happen, it appears it can happen under some unknown
conditions.

This PR adds logic that will attempt to detach disks from any instances
they're attached to before deleting the disks, adding another safeguard
that should prevent this behaviour.


paddycarver:
Tests pass, merging. :)

PR #14681, opened 2017-05-19
Use `helpers.shema.Provisoner` in Chef provisioner V2

svanharmelen:
Done by @VladRassokhin:
1. Migrate `chef` provisioner to `schema.Provisioner`:

 * `chef.Provisioner` structure was renamed to `ProvisionerS`and  now it's decoded from `schema.ResourceData` instead of `terraform.ResourceConfig` using simple copy-paste-based solution;
 * Added simple schema without any validation yet.

2. Support `ValidateFunc` validate function : implemented in `file` and `chef` provisioners.

Done by @svanharmelen:
1. The tests did pass, but that was because they only tested part of the changes. By using the `schema.TestResourceDataRaw` function the schema and config are better tested and so they pointed out a problem with the schema of the Chef provisioner. 

2. So fixed the `Elem` fields that did not have a `*schema.Schema` but a `schema.Schema` and in an `Elem` schema only the `Type` field may (and must) be set. Any other fields like `Optional` are not allowed here.

3. Next to fixing that problem I also did a little refactoring and cleaning up. Mainly making the `ProvisionerS` private (`provisioner`) and removing the deprecated fields.

### DIFF
--- a/builtin/provisioners/file/resource_provisioner.go
+++ b/builtin/provisioners/file/resource_provisioner.go
@@ -35,7 +35,8 @@ func Provisioner() terraform.ResourceProvisioner {
 			},
 		},
 
-		ApplyFunc: applyFn,
+		ApplyFunc:    applyFn,
+		ValidateFunc: Validate,
 	}
 }
 
@@ -75,6 +76,21 @@ func applyFn(ctx context.Context) error {
 	case <-ctx.Done():
 		return fmt.Errorf("file transfer interrupted")
 	}
+}
+
+// Validate checks if the required arguments are configured
+func Validate(d *schema.ResourceData) (ws []string, es []error) {
+	numSrc := 0
+	if _, ok := d.GetOk("source"); ok == true {
+		numSrc++
+	}
+	if _, ok := d.GetOk("content"); ok == true {
+		numSrc++
+	}
+	if numSrc != 1 {
+		es = append(es, fmt.Errorf("Must provide one of 'content' or 'source'"))
+	}
+	return
 }
 
 // getSrc returns the file to use as source
--- a/builtin/provisioners/file/resource_provisioner.go
+++ b/builtin/provisioners/file/resource_provisioner.go
@@ -36,7 +36,7 @@ func Provisioner() terraform.ResourceProvisioner {
 		},
 
 		ApplyFunc:    applyFn,
-		ValidateFunc: Validate,
+		ValidateFunc: validateFn,
 	}
 }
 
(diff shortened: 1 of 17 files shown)


svanharmelen:
Thx, @grubernaut!

@mitchellh I guess @grubernaut is referring to you for the change made in `helper/schema/provisioner.go` where a `ValidateFunc` is added to the `Provisioner` struct.

I found it a very helpful addition that we should/could maybe also add to the `Provider` struct? But it's good to hear your take on that one.

The rest of the changes are mainly updates and some refactoring of existing stuff, so I don't think there's anything exciting in the other parts...

Thx!

svanharmelen:
@grubernaut @mitchellh @stack72 can we move this one forward? Even though this PR isn't open for that long, the PR that is included in this PR was open since February. So it would be nice to get this one merged before the next release... Thx!

onorua:
This actually broken file provisioning, since release 0.9.7 till 0.9.8 (up until now):
https://github.com/hashicorp/terraform/issues/15177

builtin/provisioners/file/resource_provisioner.go
```diff
@@ -77,6 +78,21 @@ func applyFn(ctx context.Context) error {
 	}
 }
 
+// Validate checks if the required arguments are configured
+func Validate(d *schema.ResourceData) (ws []string, es []error) {
+	numSrc := 0
+	if _, ok := d.GetOk("source"); ok == true {
+		numSrc++
+	}
+	if _, ok := d.GetOk("content"); ok == true {
```
onorua: it seems [GetOk](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L89) fail to return `ok` in the following part:
https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L91

The struct is:
```
r: {${data.template_file.config.rendered} <nil> true true 0xc4203c12c0}
```

Which means `r.Exists` and `r.Computed` but expecting `r.Exists && !r.Computed`
I was able to make a workaround with following code:
```
func validateFn(d *schema.ResourceData) (ws []string, es []error) {
	numSrc := 0
	content := d.Get("content").(string)
	source := d.Get("source").(string)

	if content != "" {
		numSrc++
	}
	if source != "" {
		numSrc++
	}

	if numSrc != 1 {
		es = append(es, fmt.Errorf("Must provide one of 'content' or 'source'"))
	}
	log.Printf("Numsrc is: %v", numSrc)
	return
}
```

VladRassokhin:
@onorua Sorry for causing the issue, seems functions not working the way I expected. And seems tests are not coveting such scenario. It would be nice if you could also add test case in your pull request which fixes problem.

onorua:
@VladRassokhin yeah, I was amused with behavior of `GetOk` as well, will work on test cases.

PR #14703, opened 2017-05-20
ignition: tests as unit test

mcuadros:
To be able to run the tests easily and frequently for the `ignition` provider this PR sets the `resource.TestCase.IsUnitTest` as true


stack72:
LGTM!

Issue #14697, opened 2017-05-20
terraform apply crashes if provisioned instances are shutting down


rsslldnphy:
### Terraform Version

```
terraform -v
Terraform v0.9.5
```

### Affected Resource(s)
- aws_instance

### Terraform Configuration Files
```hcl
provider "aws" {
  region   = "eu-west-1"
  assume_role {
    role_arn = "arn:aws:iam::XXXXXXXXXXXX:role/admin"
  }
}

resource "aws_instance" "web" {
  ami           = "ami-6c101b0a"
  instance_type = "t2.micro"

  tags {
    Name = "HelloWorld"
  }
}
```
### Panic Output
https://gist.github.com/rsslldnphy/e703babfd47211a5d4a345c03559569b

### Expected Behavior
When running `terraform apply` while terraform managed ec2 instances are shutting down, new instances should be created.

### Actual Behavior
When running `terraform apply` while terraform managed ec2 instances are shutting down, terraform crashes.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. run `terraform apply` to provision some ec2 instances
2. terminate those instances in the AWS web console
3. immediately run `terraform apply` while the instances are still shutting down

### Important Factoids
We run a multi-AWS account setup, with an account per environment, and connect to each account via a single "gateway" account using STS.

PR #14727, opened 2017-05-22
provider/aws: Fix panic on instance shutdown

grubernaut:
During an instance shut-down network interfaces may be detached during the `READ` method of a Terraform run.

This protects the case where a network interface was detached, and is now `nil` at the time of the Terraform run, fixing nil pointer dereferences.

Fixes: #14697


stack72:
LGTM pending a full acctest run for instances - when green - merge at will :)

stack72:
@grubernaut all `instance` tests pass :) Nice work!

screen shot 2017-05-22 at 15 34 16

Issue #13625, opened 2017-04-13
Openstack provider should have a region argument


mcanevet:
Currently we have to either use the `OS_REGION_NAME` environment variable or pass `region` to every OpenStack resource.
It would be nive to have a `region` parameter to the `openstack` provider that defaults to `OS_REGION_NAME`.

jtopjian:
@mcanevet Agreed. It can be cumbersome to tag each resource with a region. Internally, there's a good reason for doing it this way (the region is specified upon each client creation), but from a UX perspective, cumbersome.

Adding a `region` parameter to the provider could be used to globally set all resources of that provider with the specified region, but there's probably going to be some caveats by introducing a "global" setting:

1. The provider `region` won't use `OS_REGION_NAME` ever.
2. If the provider `region` was specified, each resource will use that setting and not `OS_REGION_NAME` if `OS_REGION_NAME` is set.
3. If the resource `region` was set in the Terraform config, it will override the provider.

Thoughts?

mcanevet:
@jtopjian I think we could default to `OS_REGION_NAME`, imagine this manifest:

```terraform
# The default provider will use OS_REGION_NAME environment variable as region
provider "openstack" {
  user_name   = "admin"
  tenant_name = "admin"
  password    = "pwd"
  auth_url    = "http://myauthurl:5000/v2.0"
}

# The additional  provider will use "Honolulu" as region
provider "openstack" {
  user_name   = "admin"
  tenant_name = "admin"
  password    = "pwd"
  auth_url    = "http://myauthurl:5000/v2.0"
  alias       = "honolulu"
  region      = "Honolulu"
}

# This resource will use OS_REGION_NAME as region
resource "openstack_compute_instance_v2" "test-server" {
  # ...
}

# This resource will use "Honolulu" as region
resource "openstack_compute_instance_v2" "test-server_2" {
  provider = "openstack.honolulu"
  # ...
}

# This resource will use "Kahoolawe" as region
resource "openstack_compute_instance_v2" "test-server_3" {
  region = "Kahoolawe"
  # ...
}
```

or the other way around:

```terraform
# The default provider will use "Maui" as region
provider "openstack" {
  user_name   = "admin"
  tenant_name = "admin"
  password    = "pwd"
  auth_url    = "http://myauthurl:5000/v2.0"
  region      = "Maui"
}

# The additional provider will use OS_REGION_NAME environment variable as region
provider "openstack" {
  user_name   = "admin"
[...]

PR #14738, opened 2017-05-22
[WIP] provider/openstack: Add Region to Provider

jtopjian:
This commit adds the region argument to the OpenStack provider.
This enables a user to specify the region once at the provider-level
and have all resources within that provider be built in that region.

If a resource-level region was specified, it is still recognized
and will override the provider-level region.

For #13625

### DIFF
--- a/builtin/providers/openstack/config.go
+++ b/builtin/providers/openstack/config.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 
@@ -24,6 +25,7 @@ type Config struct {
 	IdentityEndpoint string
 	Insecure         bool
 	Password         string
+	Region           string
 	Swauth           bool
 	TenantID         string
 	TenantName       string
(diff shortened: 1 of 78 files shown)


mcanevet:
@jtopjian at first glance it looks OK, but whe I try this on an existing project a plan says me that it will force a new resource:

```
-/+ openstack_networking_port_v2.port
    ...
    region:                  "RegionOne" => "" (forces new resource)
    ...
```

This is weird because when running in debug mode I get this:
```shell
2017/05/23 09:45:39 [DEBUG] plugin: terraform-provider-openstack: 2017/05/23 09:45:39 [DEBUG] OpenStack Region is: RegionOne
```

jtopjian:
@mcanevet OK, now the fun begins :)

Did you get that error when you added `region` to the provider and removed it from the resource?

jtopjian:
@mcanevet Give the latest commit a try, but be careful. It's not triggering a "forces new resource" when I think it should be triggered... but it seems to work correctly.

This latest commit is moving the `OS_REGION_NAME` to the provider `region`. As per our previous discussion, I was hesitant about doing that, but after a few different configurations, it does in fact look like the cleanest way of applying this.

I'm to try this out in a multi-region environment myself, but it might take me a few days to get to that.

mcanevet:
@jtopjian looks like it works fine now, even with a multi-region environment.
I this you can merge this.

jtopjian:
Nice! I'm glad it's working for you.

I still want to do more testing on it, plus there's a whole slew of doc changes to make. So hopefully it'll be ready soon. 

But I'm really glad you're able to use it without blowing up your environments -- thank you for taking the time to test this. :)

jtopjian:
I haven't forgotten about this. I still need to do some testing on a few environments I have access to - I've just been pressed for time.

Also, 0.10 is coming up and I think this would be a better fit there. Lots of code being changed in this patch (although it's the same changes in each file). But a new release series is best for something like this.

jtopjian:
This has been implemented in the new provider plugin.

PR #14761, opened 2017-05-23
Update bitbucket.html.md

sprutner:
You left the test domain in the docs, not sure if this is intended.


stack72:
Thanks for this :)

Issue #14699, opened 2017-05-20
[scaleway provider] duplicate volume creation when declaring system volume


polymeris:
Looking for a way to use terraform to vertically scale on scaleway (keep a volume, change the server specs).

### Terraform Version
v0.9.5

### Affected Resource(s)
- scaleway_server
- scaleway_volume

### Terraform Configuration Files
``` json
{
  "provider": {
    "scaleway": {
      "region": "par1"
    }
  },
  "resource": {
    "scaleway_volume": {
      "data": {
        "name": "test-data",
        "size_in_gb": 50,
        "type": "l_ssd"
      }
    },
    "scaleway_volume_attachment": {
      "data": {
        "server": "${scaleway_server.host.id}",
        "volume": "${scaleway_volume.data.id}"
      }
    },
    "scaleway_server": {
      "host": {
        "name": "test",
        "type": "VC1M",
        "image": "d0c50b37-f0f5-4924-97b0-1fd40c457efe",
        "volume": {
          "size_in_gb": 50,
          "type": "l_ssd"
        }
      }
    },
    "scaleway_ip": {
      "ip": {
        "server": "${scaleway_server.host.id}"
      }
    }
  }
}
```

### Expected Behavior
2 volumes should be created:
- The system volume
- An additional 50 gb volume labeled "test-data"

### Actual Behavior
Three 50 gb volumes get created:
- System volume ("x86_64-docker-latest-2016-10-24_15:48")
- "test-50"
- "test-data"

Since the max storage per VC1M instance is 100gb, this causes `terraform apply` to fail.

### Steps to Reproduce
With the above config:
1. `terraform apply`

### References
The system volume option is not documented in the docs. I am basing the config on the example in PR #9695

nicolai86:
thanks for reporting this, @polymeris .

This is actually expected behaviour as every server has a System Volume, which can not be modified like this. The `volume` attribute on a `scaleway_server` resource was meant to allow volumes bound to the live time of the server. 

To my best knowledge modifying the System Volume would require you to build your own image, which is much more involved and outside the scope of terraform. Packer would probably be a better place for this.

polymeris:
Thanks, Nicolai.

What's the proper configuration to create a 50gb system volume + one additional 50gb volume, then?

If I leave out `scaleway_server.volume` is still get an error:

```
* scaleway_server.host: StatusCode: 400, Type: invalid_request_error, APIMessage: The total volume size of VC1M instance
[...]

PR #14772, opened 2017-05-23
provider/scaleway: clarify scaleway_server volume attribute

nicolai86:
This PR aims to improve the docs so misunderstandings like #14699 don't occur more often.

**background**
 
System volumes on Scaleway can't easily be modified - instead one has to create
a new image with the desired system volume size. This is way out of scope of
terraform - see https://community.online.net/t/expanding-lssd/907/2 for steps on
how to build a new image.

the `scaleway_server` `volume` attribute should only be used if you want to
attach additional volumes to a server which will share the lifetime of the
server, e.g. they will be destroyed once the server is shut down.

To have volumes which outlive the attached server one should use
`scaleway_volume` and `scaleway_volume_attachement` instead.


stack72:
LGTM :) Thanks for this!

PR #14773, opened 2017-05-23
provider/aws: Add Sweeper setup, Sweepers for DB Option Group, Key Pair

catsby:
Terraform's acceptance testing framework is responsible for creating the infrastructure in the configuration, and then destroying all resources created. While by itself it is largely successful in destroying all resources created, configurations can be constructed that prevent successful teardown. Additionally, any test that panics or hits the Test Runner timeout can result in dangling, leaked resources.

Here we introduce "Sweepers" into helper/resource and adds example implementations for:

- `aws_db_option_group`
- `aws_key_pair`
- `aws_launch_configurations`
- `aws_autoscaling_group`
- `google_sql_database_instances`

We use the CLI flag `sweep` to provide a list of regions to run the Sweepers in. This flag will intercept the normal test flow, and instead run the Sweepers that have been registered. Provider authors are responsible for writing the Sweeper methods and the criteria they use to determine what to destroy. For example, the included `aws_key_pair` sweeper func `testSweepKeyPairs` will destroy any EC2 key pair that match the name pattern of tmp-key*. 

A Sweeper may list another Sweeper as a dependency, via the Sweepers name. For example, the `aws_launch_configuration` depends on the `aws_autoscaling_group` Sweeper to be ran first. The `sweep-run` flag is an optional argument allowing users to specifically run that sweeper. **NOTE** there is no way at time of writing to target run a given Sweeper **and not** also run it's dependencies. 

Sweepers are meant to be ran periodically, especially in a Continuous Integration environment, and should not be used anywhere outside of a test environment, where everything should be considered disposable. The majority of Terraform users should not use sweepers at all. Sweepers have no purpose other than to destroy leaked resources. 

------

Invocation examples:

    $ go test ./builtin/providers/aws -sweep=us-east-1,us-west-2
    $ go test ./builtin/providers/aws -sweep=us-east-1 -sweep-run=launch_configuration

The `-v` option to `go test` will show a list of tests ran per region:

```
$ go test ./builtin/providers/aws -v -sweep=us-east-1 -sweep-run=launch_configuration
[DEBUG] Running Sweepers for region (us-east-1):
[...]
Sweeper Tests ran:
        - aws_launch_configuration
        - aws_autoscaling_group
ok      github.com/hashicorp/terraform/builtin/providers/aws    3.443s
```

```
$ go test ./builtin/providers/aws -v -sweep=us-west-2
[DEBUG] Running Sweepers for region (us-west-2):
[...]
Sweeper Tests ran:
        - aws_key_pair
        - aws_autoscaling_group
        - aws_launch_configuration
        - aws_db_option_group
ok      github.com/hashicorp/terraform/builtin/providers/aws    5.593s
```

Included test:

```
$ go test ./helper/resource -v -run=TestTest_Main
[debug output omitted]
--- PASS: TestTest_Main (0.00s)
    --- PASS: TestTest_Main/normal (0.00s)
    --- PASS: TestTest_Main/with_dep (0.00s)
    --- PASS: TestTest_Main/with_filter (0.00s)
    --- PASS: TestTest_Main/with_two_filters (0.00s)
    --- PASS: TestTest_Main/with_dep_and_filter (0.00s)
    --- PASS: TestTest_Main/filter_and_none (0.00s)
PASS
ok      github.com/hashicorp/terraform/helper/resource  0.012s
```

### DIFF
--- a/builtin/providers/google/resource_sql_database_instance_test.go
+++ b/builtin/providers/google/resource_sql_database_instance_test.go
@@ -22,21 +22,20 @@ import (
 )
 
 func init() {
-	var fs []*resource.Sweeper
-	// add sweepers for each region
 	for _, r := range []string{"us-central1"} {
 		c, err := sharedCredsForRegion(r)
 		if err != nil {
 			log.Printf("[ERR] error getting shared config for region: %s", err)
 			continue
 		}
-		fs = append(fs, &resource.Sweeper{
-			Config: c,
-			F:      sweepDatabases,
-		})
+		name := fmt.Sprintf("gcp-sql-db-instance-%s", r)
+		resource.AddTestSweepers(name,
+			&resource.Sweeper{
+				Name:   name,
+				Config: c,
+				F:      sweepDatabases,
+			})
 	}
-
-	resource.AddTestSweepers("google_sql_database_instance", fs)
 }
 
 func sweepDatabases(c interface{}) error {
--- a/builtin/providers/google/resource_sql_database_instance_test.go
+++ b/builtin/providers/google/resource_sql_database_instance_test.go
@@ -28,8 +28,8 @@ func init() {
 	})
 }
 
-func testSweepDatabases(c interface{}) error {
-	config, err := sharedConfigForRegion(c.(string))
+func testSweepDatabases(region string) error {
+	config, err := sharedConfigForRegion(region)
 	if err != nil {
 		return fmt.Errorf("error getting shared config for region: %s", err)
 	}
--- a/builtin/providers/google/resource_sql_database_instance_test.go
+++ b/builtin/providers/google/resource_sql_database_instance_test.go
@@ -22,20 +22,6 @@ import (
 )
 
 func init() {
-	// for _, r := range []string{"us-central1"} {
-	// 	c, err := sharedCredsForRegion(r)
-	// 	if err != nil {
-	// 		log.Printf("[ERR] error getting shared config for region: %s", err)
-	// 		continue
-	// 	}
-	// 	name := fmt.Sprintf("gcp-sql-db-instance-%s", r)
-	// 	resource.AddTestSweepers(name,
-	// 		&resource.Sweeper{
-	// 			Name:   name,
-	// 			Config: c,
-	// 			F:      sweepDatabases,
-	// 		})
-	// }
 	resource.AddTestSweepers("gcp_sql_db_instance", &resource.Sweeper{
 		Name: "gcp_sql_db_instance",
 		F:    testSweepDatabases,
@@ -43,9 +29,9 @@ func init() {
 }
 
 func testSweepDatabases(c interface{}) error {
-	config, err := sharedCredsForRegion(c.(string))
+	config, err := sharedConfigForRegion(c.(string))
 	if err != nil {
-		return fmt.Errorf("[ERR] error getting shared config for region: %s", err)
+		return fmt.Errorf("error getting shared config for region: %s", err)
 	}
 
 	err = config.loadAndValidate()
@@ -87,7 +73,7 @@ func testSweepDatabases(c interface{}) error {
 			op, err := config.clientSqlAdmin.Instances.StopReplica(config.Project, replicaName).Do()
 
 			if err != nil {
-				return fmt.Errorf("Error, failed to stop replica instance (%s) for instance (%s): %s", replicaName, d.Name, err)
+				return fmt.Errorf("error, failed to stop replica instance (%s) for instance (%s): %s", replicaName, d.Name, err)
 			}
 
 			err = sqladminOperationWait(config, op, "Stop Replica")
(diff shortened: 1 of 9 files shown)


stack72:
So I really like this approach - it is specific to a resource type and also allows us to specify the places where we need to run them

the one issue we may have is running against things like gov cloud - would we have to split the sweeper test into 2 and run 1 set in gov cloud and the other in normal AWS?

catsby:
> the one issue we may have is running against things like gov cloud - would we have to split the sweeper test into 2 and run 1 set in gov cloud and the other in normal AWS?

My initial thoughts on this is to create is separate sweeper methods in the test file specifically for GOV clouds, along side any existing sweeper methods. If we add an ENV var to the test runner to denote GOV, we could toggle on that in the test file `init` method that calls `resource.AddTestSweepers`, and give it a GOV cloud config

helper/resource/testing.go
```diff
@@ -20,6 +21,68 @@ import (
...
+// tests, and exit on any errors. At time of writing, sweepers are ran
+// sequentially.
+//
+// WARNING:
+// Sweepers are designed to be destructive. You should not use the -sweep flag
+// in any environment that is not strictly a test environment. Resources will be
+// destroyed.
+
+var FlagSweep bool
+var SweeperFuncs map[string][]*Sweeper
```
mitchellh: Any reason for this to be exported? It looks like all access is via `AddSweeperFunc` so it may be better to unexport this.

helper/resource/testing.go
```diff
@@ -20,6 +21,68 @@ import (
...
+// construct a list of sweeper funcs to be called here. We iterate through the
+// tests, and exit on any errors. At time of writing, sweepers are ran
+// sequentially.
+//
+// WARNING:
+// Sweepers are designed to be destructive. You should not use the -sweep flag
+// in any environment that is not strictly a test environment. Resources will be
+// destroyed.
+
+var FlagSweep bool
```
mitchellh: This probably can be unexported.

Also, you can make this a one-liner with `var flagSweep = flag.Bool(...)` and access it with `*flagSweep`

helper/resource/testing.go
```diff
@@ -20,6 +21,68 @@ import (
...
+	// resources
+	F SweeperFunc
+}
+
+func init() {
+	flag.BoolVar(&FlagSweep, "sweep", false, "")
+	SweeperFuncs = make(map[string][]*Sweeper)
+}
+
+func AddTestSweepers(name string, sf []*Sweeper) {
```
mitchellh: Might still be WIP but please add doc strings to the top of this function.

helper/resource/testing.go
```diff
@@ -20,6 +21,68 @@ import (
...
+	if _, ok := SweeperFuncs[name]; ok {
+		log.Printf("Error adding (%s) to SweeperFuncs: function already exists in map", name)
```
mitchellh: May just want to `log.Fatalf` since you exit right away anyways.

helper/resource/testing.go
catsby: Unexpected and simplified in dc5730d0a

helper/resource/testing.go
catsby: I may have been accessing it directly elsewhere before 🤔  but not anymore, thanks :D Unexpected in dc5730d0a

helper/resource/testing.go
catsby: Documented in dc5730d0a

helper/resource/testing.go
catsby: 👍 in dc5730d0a

builtin/providers/google/resource_sql_database_instance_test.go
```diff
@@ -20,6 +21,105 @@ import (
...
+	}
+
+	if len(found.Items) == 0 {
+		log.Printf("No databases found")
+		return nil
+	}
+
+	for _, d := range found.Items {
+		var testDbInstance bool
+		for _, testName := range []string{"tf-lw-", "sqldatabasetest"} {
```
grubernaut: Minor nit: This pattern happens a lot throughout all of the test sweepers. Perhaps extracting the logic to a function inside the `helper` package? 
```
func isTestResource(resourceName string, testNames []string) bool {
  // ...
}
```

helper/resource/testing.go
```diff
@@ -20,6 +21,152 @@ import (
...
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	if *flagSweep != "" {
+		// parse flagSweep contents for regions to run
+		regions := strings.Split(*flagSweep, ",")
+
+		// get filtered list of sweepers to run based on sweep-run flag
+		sweepers := filterSweepers(*flagSweepRun, sweeperFuncs)
+		for _, region := range regions {
```
mitchellh: I would do a `region = strings.TrimSpace(region)` as well to trim any surrounding whitespace. Since this is coming in from a CLI its reasonable that that could happen.

PR #14784, opened 2017-05-24
helper/schema: fix validating nested objects

mattrobenolt:
When interpreting a nested object, we were validating against the "raw"
value, and not the interpolated value, causing incorrect errors.

This affects structures such as:

```tf
tags = "${list(map("foo", "bar"))}"
```

Prior to this, a complaint about "expected object, got string" since the
raw value is obviously a string, when the interpolated value is the
correct shape.

### DIFF
--- a/helper/schema/schema.go
+++ b/helper/schema/schema.go
@@ -1373,7 +1373,7 @@ func (m schemaMap) validateObject(
 	k string,
 	schema map[string]*Schema,
 	c *terraform.ResourceConfig) ([]string, []error) {
-	raw, _ := c.GetRaw(k)
+	raw, _ := c.Get(k)
 	if _, ok := raw.(map[string]interface{}); !ok {
 		return nil, []error{fmt.Errorf(
 			"%s: expected object, got %s",
--- a/helper/schema/schema_test.go
+++ b/helper/schema/schema_test.go
@@ -3947,6 +3947,33 @@ func TestSchemaMap_Validate(t *testing.T) {
 			Err: false,
 		},
 
+		"Good sub-resource, interpolated value": {
+			Schema: map[string]*Schema{
+				"ingress": &Schema{
+					Type:     TypeList,
+					Optional: true,
+					Elem: &Resource{
+						Schema: map[string]*Schema{
+							"from": &Schema{
+								Type:     TypeInt,
+								Required: true,
+							},
+						},
+					},
+				},
+			},
+
+			Config: map[string]interface{}{
+				"ingress": []interface{}{
+					`${map("from", "80")}`,
+				},
+			},
+
+			Vars: map[string]string{},
+
+			Err: false,
+		},
+
 		"Invalid/unknown field": {
 			Schema: map[string]*Schema{
 				"availability_zone": &Schema{


helper/schema/schema_test.go
```diff
@@ -3947,6 +3947,33 @@ func TestSchemaMap_Validate(t *testing.T) {
...
+				},
+			},
+
+			Config: map[string]interface{}{
+				"ingress": []interface{}{
+					`${map("from", "80")}`,
+				},
+			},
+
+			Vars: map[string]string{},
```
mattrobenolt: This was needed to trigger interpolation. Otherwise it's skipped and `map()` is never resolved since `tc.Vars == nil`.

mattrobenolt:
Also, to be explicit, this now fixes behavior and allows doing syntax such as:

```tf
provider "google" {
    region = "test"
}

variable "image" {
    default = "test-image"
}

variable "root_disk" {
    type = "map"
    default = {
        image = "something-awesome"
        size = 512
    }
}

variable "disk" {
    type = "list"
    default = [{
        image = "lol"
    }]
}

resource "google_compute_instance" "test" {
    name = "test"
    machine_type = "test"
    zone = "test"

    disk = "${concat(
        list(
            merge(
                map(
                    "image", var.image,
                    "size", "100",
                ),
                var.root_disk,
            ),
        ),
        var.disk
    )}"
}
```

helper/schema/schema.go
```diff
@@ -1373,7 +1373,7 @@ func (m schemaMap) validateObject(
 	k string,
 	schema map[string]*Schema,
 	c *terraform.ResourceConfig) ([]string, []error) {
-	raw, _ := c.GetRaw(k)
+	raw, _ := c.Get(k)
```
jbardin: Since Get will return the Raw value if it's computed, this should probably check for that too. I don't think that validation error helps in the unknown variable case either.

helper/schema/schema.go
```diff
@@ -1373,7 +1373,7 @@ func (m schemaMap) validateObject(
...
 	if _, ok := raw.(map[string]interface{}); !ok {
```
jbardin: I think simply adding `; !ok && !c.IsComputed(k)` would suffice here

helper/schema/schema_test.go
```diff
@@ -3947,6 +3947,33 @@ func TestSchemaMap_Validate(t *testing.T) {
 			Err: false,
 		},
 
+		"Good sub-resource, interpolated value": {
```
jbardin: Since we're already in here, I think a test with a computed value (that fails before the IsComputed check) would be useful too.

mattrobenolt: It's a bit unclear to me what `IsComputed` is actually checking for. Can you provide me with an example? Is this something like, `foo.0`?

mattrobenolt: I _think_ I got it, but I'm unable to put together a failing test.

The best case I could come up with was:

```go
		"Bad sub-resource, interpolated value": {
			Schema: map[string]*Schema{
				"ingress": &Schema{
					Type:     TypeList,
					Optional: true,
					Elem: &Resource{
						Schema: map[string]*Schema{
							"from": &Schema{
								Type:     TypeInt,
								Required: true,
							},
						},
					},
				},
			},

			Config: map[string]interface{}{
				"ingress": []interface{}{
					`${map("from", var.port)}`,
				},
			},

			Vars: map[string]string{
				"var.port": config.UnknownVariableValue,
			},

			Err: true,
		},
```

jbardin: IsComputed basically checks if the value is known during validation, i.e. it's in the config and doesn't contain any UnknownValues.   If you look at the source to `Get`, it's actually going to call `IsComputed` there too, and otherwise return `GetRaw`. 

I think the PR is still good on its own, so I can merge this and I'll send a follow up PR with the additional changes.

mattrobenolt: > IsComputed basically checks if the value is known during validation, i.e. it's in the config and doesn't contain any UnknownValues.

Yeah, I got that far. :) I was moreso struggling to construct a test case that proves that additional check is required.

> I'll send a follow up PR with the additional changes.

Mind cc'ing me on this? I'd like to see what I'm missing here for the future.

Thanks!

jbardin: Sorry, started typing before I saw your update... 
That looks right, and I swear I had it fail when I tested it. Let me see what's going on.

jbardin: I see, it's passing because `Err: true,`. 😉 

You'll just need to change `Required` to `Optional` here to pass the test, since there there is no real value here.

PR #14793, opened 2017-05-24
provider/vsphere: Exposing moid value from vm resource

sandeep-sidhu:
moid value is needed by NSX resources, like security tag, when we attached security tags to a VMs, so needed before we commit NSX provider.


stack72:
Hi @sandeep-sidhu 

Thanks for this - please can you add this to the documentation to show this is now a valid exported parameter?

Thanks

Paul

sandeep-sidhu:
@stack72 done.

stack72:
LGTM! thanks for this :)

PR #14801, opened 2017-05-24
check for IsComputed when validating a schema object

jbardin:
GH-14784 allowed nested structures to be validate, rather than relying
on the raw value. However this still returns the same validation error
if the structures contain a computed value, since Get will return the
raw string in that case.

This simply skips the validation in the IsComputed case, since there's
nothing that can be checked.


jbardin:
@mattrobenolt, thanks for the initial PR!

mattrobenolt:
Ah, I see what I did wrong. :) Thanks for the carry! ❤️

PR #14806, opened 2017-05-24
doc: Fix custom providers snippets

dgellow:
Looking at [the guide](https://www.terraform.io/guides/writing-custom-terraform-providers.html) there is some references to `meta` but I cannot see anything called like this. Though it seems to correspond to the `m` argument.


stack72:
LGTM! Thanks for this :)

PR #14811, opened 2017-05-24
[MS] Doc fixes for managed disks in VMs

StephenWeatherford:


stack72:
Nice Addition - thanks for this :)

PR #14830, opened 2017-05-25
Kubernetes persistent volume should be namespace aware

jpbelanger-mtl:
kubernetes_persistent_volume_claim allows namespace, but not kubernetes_persistent_volume.

Changed to use namespaceMetadataSchema instead


jpbelanger-mtl:
Need to double check the kube-api. Namespace is not available as a parameter on the PersistentVolumes call. closing...

PR #14831, opened 2017-05-25
[WIP] Add AWS SES dkim resource

arr-dev:
Added `aws_ses_dkim` resource, according to http://docs.aws.amazon.com/ses/latest/DeveloperGuide/easy-dkim-dns-records.html

Would like some feedback on schema, not sure if there is anything else needed here.

Haven't ran integration test since I'am having issues with my personal AWS account, 
but hope to enable it soon so I can test this, hopefully having some feedback before that.


leonsmith:
Do you think its wise to change the count = 3 in the example docs to be `"${length(aws_ses_domain_dkim.example.dkim_tokens)}"` ?

I imagine it will be copied out of the docs verbatim so thats a bit more future proof incase the the number of dkim tokens changes?

arr-dev:
I actually tried that and Terraform reported that count can't be interpolated  or something like that.
It was strange to me to.

I should probably close this PR and reopen it on https://github.com/terraform-providers/terraform-provider-aws?

arr-dev:
@handlerbot I'll submit a PR tomorrow or monday the latest. Thanks for reminding me of this

arr-dev:
Recreated at terraform-providers/terraform-provider-aws#1786

PR #14850, opened 2017-05-26
fix incorrect Lambda GovCloud regexes

jeremy-asher:
I'm putting this up for comment on the approach.

    This updates two regexes that are used to validate and parse ARNs for
    aws_lambda_permission.  The ARNs were not accurate when using GovCloud
    Lambdas.  This resulted in failures during the read call after applying
    a resource.

Without this fix:
```
2017/05/25 17:01:25 [ERROR] root.module_name: eval: *terraform.EvalRefresh, err: aws_lambda_permission.allow_cloudwatch: Invalid ARN or otherwise unable to get qualifier from ARN ("arn:aws-us-gov:lambda:us-gov-west-1:123456789012:function:function_name")
...

* module.module-name.aws_lambda_permission.allow_cloudwatch: aws_lambda_permission.allow_cloudwatch: Invalid ARN or otherwise unable to get qualifier from ARN ("arn:aws-us-gov:lambda:us-gov-west-1:123456789012:function:function_name")
```

This occurs because the region name for GovCloud contains an additional section.

Considering the duplication of the regex and the fact that validateLambdaFunctionName is only used in aws_lambda_permission, I'm thinking some cleanup is in order as well, but this patch should work on its own.  I'm open to suggestions to improve this, but wasn't sure where I should put the Regex if it needs to be reused.  Note that it's tightly coupled to this code as well:
https://github.com/jeremy-asher/terraform/blob/28a2ec1b9fa83549a2b5c45ab75f5928ce848893/builtin/providers/aws/resource_aws_lambda_permission.go#L346-L363

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_permission.go
+++ b/builtin/providers/aws/resource_aws_lambda_permission.go
@@ -15,7 +15,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-var LambdaFunctionRegexp = `^(arn:[\w-]+:lambda:)?([a-z]{2}-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\$LATEST|[a-zA-Z0-9-_]+))?$`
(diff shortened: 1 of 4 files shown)


stack72:
Hi @jeremy-asher 

Thanks for this - please can you add a unit test to show that both gov and non-gov partitions will now satisfy the regex?

Thanks

Paul

jeremy-asher:
Done.  Looks like there were a few existing tests, but they were referencing the wrong region for GovCloud.

jeremy-asher:
@stack72 need anything else on this?

jeremy-asher:
Opened terraform-providers/terraform-provider-aws#16 to replace this.

PR #14859, opened 2017-05-26
github_team data source: fix wrong set

raphink:
Do not set `members_count` and `repos_count`, which are not declared.

The current code crashes on apply.


stack72:
LGTM! thanks for this

Issue #14188, opened 2017-05-03
Many string enums in AzureRM provider think the property has changed if the casing is different than expected


StephenWeatherford:
### Terraform Version
0.9.4

### Affected Resource(s)
Multiple

### Expected Behavior
Either the properties should accept only a single case, or else the behavior should be idempotent no matter which casing the template uses.  
Example: azurerm_virtual_machine_scale_set.agentpool.sku.tier accepts either "standard" or "Standard", and it shouldn't matter which is actually used.

### Actual Behavior
Example: azurerm_virtual_machine_scale_set.agentpool.sku.tier accepts either "standard" or "Standard", but if "standard" is used, the next tf plan will think a change occurred when none has.

### Steps to Reproduce
See #11626 for an example.

We need to go through all the enum-style string properties and verify casing is handled correctly.

StephenWeatherford:
@abhijeetgaiha is planning on working on this.

abhijeetgaiha:
Which is the more appropriate approach, restricting the input or making the behavior idempotent?

tombuildsstuff:
@StephenWeatherford @abhijeetgaiha thanks for raising this!

The convention is to make this case-insensitive - the Resource Schema exposes a `DiffSuppressFunc` property which allows us to handle this by assigning a Diff function. [Here's an example](https://github.com/hashicorp/terraform/blob/master/builtin/providers/azurerm/resource_arm_loadbalancer_rule.go#L75) of how we're using an `ignoreCaseDiffSuppressFunc` to do this in the Load Balancer Rule resource :)

Hope that helps :)

StephenWeatherford:
@tombuildsstuff Thanks!  Is there a reason to also use a state func (ignoreCaseStateFunc), or does that not matter, as long as we have the ignore-case-diff function?

tombuildsstuff:
@StephenWeatherford it's optional - so I believe you should be able to get away without it :)
(it's used should you wish to modify the data stored in the `*.tfstate` file, for example base64 encoding a large string to compress space in the state etc)

abhijeetgaiha:
@tombuildsstuff Can we have DiffSuppressFunc on individual fields on map and set types? It doesn't seem to be the case.

tombuildsstuff:
[...]

PR #14861, opened 2017-05-26
Changes to behaviour of string enum comparison in AzureRM provider

abhijeetgaiha:
Fixes issue https://github.com/hashicorp/terraform/issues/14188 and also https://github.com/hashicorp/terraform/issues/11626.


stack72:
Hi @abhijeetgaiha 

Thanks so much for this - this is a great addition

Paul

PR #14862, opened 2017-05-26
Update alicloud document to add userdata

mosuke5:
Add userdata arg to alicloud docs


stack72:
LGTM thanks for this :)

PR #14865, opened 2017-05-26
docs: Fix inline comment in AWS instance resource

DavidWittman:
`//` appears to be a valid inline comment, but it affects the formatting
in the documentation.

image

Octothorps are formatted properly when used at the end of a line:

image


catsby:
Thanks!

catsby:
Deployed! Looks much better, thanks again!

PR #14871, opened 2017-05-26
provider/librato: Fixup the Librato Docs (metrics)

wchrisjohnson:
In my latest PR against the Librato provider, I just realized I neglected to add the new metric resource to the sidebar so it would show up in the docs. This fixes that up.


wchrisjohnson:
This time I verified via the Docker based local website option.

PR #14883, opened 2017-05-27
DOCS: The resource name should be consistent.

kevinprince:
The HCL declares the terraform_remote_state with a resource name of foo. But the example invocation uses network which is incorrect.


kevinprince:
@radeksimko Yes! I have updated my PR.

PR #14896, opened 2017-05-28
[docs] add parameter used to disable locking in the S3 backend

maplebed:
The 0.9 upgrading doc specifies that you can use `lock = false` to disable locking in the S3 and consul backends. The consul backend includes this parameter but the S3 docs do not.

This change adds the `lock` parameter to the list of valid backend parameters for the S3 backend.


radeksimko:
Hi @maplebed 
thanks for taking time to send the PR, but there's no such field in the [S3 backend schema](https://github.com/hashicorp/terraform/blob/master/backend/remote-state/s3/backend.go#L17-L134). The way users can disable/enable locking for that backend is via `lock_table` field.

See relevant source code: https://github.com/hashicorp/terraform/blob/master/backend/remote-state/s3/client.go#L185-L188

Issue #14852, opened 2017-05-26
DO loadbalancer state is failing if it was already removed


onorua:
### Terraform Version
```
Terraform v0.9.5
```
### Affected Resource(s)
- digitalocean_loadbalancer

### LB state
```
resource "digitalocean_loadbalancer" "test" {
  name   = "test-lb"
  region = "nyc3"

  forwarding_rule {
    entry_port     = 80
    entry_protocol = "http"

    target_port     = 80
    target_protocol = "http"
  }

  healthcheck {
    port                     = 80
    protocol                 = "http"
    path = "/healthy.html"
    response_timeout_seconds = 3
    check_interval_seconds   = 3
  }

  droplet_ids = ["${digitalocean_droplet.node.*.id}"]
}
```
### Expected Behavior
In case LB is already removed - it should mark internal state as removed already, and either continue destroying things or create new LB. 

### Actual Behavior
It is failing with error like this:
```
* module.nyc3-dev-2.module.loadbalancer.digitalocean_loadbalancer.test: digitalocean_loadbalancer.test: Error retrieving Loadbalancer: GET https://api.digitalocean.com/v2/load_balancers/bc34d081-c48c-4fc0-8a77-21428bae2cf6: 404 (request "e5a0ef8c-4a47-4cc8-86e0-c678279ac2fb") Load Balancer not found
```

### Steps to Reproduce
1. `terraform apply` in order to create DO LB
2. Delete DO LB from Web GUI or via API manually
3. run `terraform destroy` or `terraform apply`
4. terraform is rejecting to do anything  

Update:
Just to make it clear, you can fix the state with `terraform state rm`, but this is simply counter intuitive as removed LB during destroy means work  is done already, and if it during apply - we are missing LB, and should created it.

PR #14897, opened 2017-05-28
provider/digitalocean: Refresh DO loadbalancer from state if 404

stack72:
Fixes: #14852


stack72:
@radeksimko I will follow up with a PR for an acceptance test for this - it's causing me issues right now around doing things too fast (i.e. create and destroy happens to soon)

PR #14907, opened 2017-05-29
provider/openstack: Fix issue with providing OS_IMAGE_ID or OS_IMAGE_…

fatmcgav:
…NAME when running acceptance tests.


jtopjian:
LGTM - thanks :)

PR #14912, opened 2017-05-29
Added example for globally load balanced scale sets

eltimmo:
Hi, I've created a template that creates a globally load balanced web service for Azure using scale sets. There's documentation in the code as well as the readme.md. I've read the contribution guide and think this is okay. I've done as much testing. I hope this okay. Tim.


stack72:
LGTM! thanks for this :)

PR #14922, opened 2017-05-30
fixed error message

jasminSPC:
A minor fix to ignition user test.


builtin/providers/ignition/resource_ignition_user_test.go
```diff
@@ -78,7 +78,7 @@ func TestIngnitionUser(t *testing.T) {
 		}
 
 		if u.Create.NoUserGroup != true {
-			return fmt.Errorf("no_create_home, found %t", u.Create.NoCreateHome)
+			return fmt.Errorf("no_create_home, found %t", u.Create.NoUserGroup)
```
radeksimko: Thanks for noticing this! In fact I feel like all the error messages here in this file are odd and not very user-friendly. I would personally expect something like

```
Field NoUserGroup didn't match. Given: %t, Expected: %t
```
for all the fields respectively.

Would you mind fixing all of them here in this file?

jasminSPC: @radeksimko I will make those changes throughout the provider

jasminSPC:
@radeksimko I've added improved error messages only for this file.

PR #14949, opened 2017-05-30
Replace lock_table with dynamodb_table in S3 backend config

jbardin:
Since the DynamoDB table used by the S3 backend is no longer only used
for locks, rename it in the config to remove any confusion about it
being lock-specific.

Update the docs and add missing `assume_role_policy`, `external_id`, and `session_name` parameters too.

### DIFF
--- a/backend/remote-state/s3/backend.go
+++ b/backend/remote-state/s3/backend.go
@@ -81,6 +81,14 @@ func New() backend.Backend {
 				Optional:    true,
 				Description: "DynamoDB table for state locking",
 				Default:     "",
+				Deprecated:  "please use the dynamodb_table attribute",
+			},
+
+			"dynamodb_table": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "DynamoDB table for state locking and consistency",
+				Default:     "",
 			},
 
 			"profile": {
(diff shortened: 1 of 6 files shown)


josh-padnick:
@jbardin It looks like the docs now say that the DynamoDB table is used for "state locking and consistency." Can you clarify what kind of consistency operations are performed in DynamoDB?

jbardin:
Hi @josh-padnick,

The consistency check is fairly simple, when a state is stored an md5sum of the blob (md5 because that's what the pre-existing remote client interface already supported) is saved in the dynamodb table. When the state is fetched it's compared against the stored checksum, and if it doesn't match the client will try again for a short period of time before returning an error.

While the consistency check prevents against unexpected alterations to the state, the real goal is simply to avoid pulling a stale state due the eventual consistency delays in S3. Now that locks are available, users are more comfortable running terraform executions in close succession, and the propagation time within s3 has become much more apparent.

josh-padnick:
Very interesting! Thanks for the explanation.

maheshshekhar:
i get error when using the dynamodb_table for the backend s3. 
Error: Error configuring the backend "s3": No valid credential sources found for AWS Provider. 

please let us know how we can go about this.

PR #14955, opened 2017-05-31
provider/openstack: Add Terraform version to UserAgent string

fatmcgav:
Just because it's useful... :)

### DIFF
--- a/builtin/providers/openstack/config.go
+++ b/builtin/providers/openstack/config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/swauth"
 	"github.com/hashicorp/terraform/helper/pathorcontents"
+	"github.com/hashicorp/terraform/terraform"
 )
 
 type Config struct {
@@ -68,6 +69,9 @@ func (c *Config) loadAndValidate() error {
 	if err != nil {
 		return err
 	}
+
+	// Set UserAgent
+	client.UserAgent.Prepend(terraform.UserAgentString())
 
 	config := &tls.Config{}
 	if c.CACertFile != "" {
--- a/terraform/user_agent.go
+++ b/terraform/user_agent.go
@@ -1,0 +1,14 @@
+package terraform
+
+import (
+	"fmt"
+	"runtime"
+)
+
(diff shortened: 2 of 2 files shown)


fatmcgav:
@jtopjian Thought this would be useful to add :)

jtopjian:
@fatmcgav Yep, definitely useful. Looking at the other implementations of this, they all have HashiCorp in the agent, too, so how about:

```
HashiCorp-Terraform/" + terraform.VersionString()
```

fatmcgav:
@jtopjian Yep, can make that tweak... I did look at other providers and didn't see a real standard...

thomaschaaf:
Wouldn't it make sense to create one helper function to render the user agent for every provider? I don't think it makes sense to manually try to keep the user agent implementation the same?

jtopjian:
> Yep, can make that tweak... I did look at other providers and didn't see a real standard...

Yeah, it looks like a mix of agent types. I saw that most of them had "HashiCorp" in the name, which is why I suggested it.

> Wouldn't it make sense to create one helper function to render the user agent for every provider? I don't think it makes sense to manually try to keep the user agent implementation the same?

I agree. Or at least standardize on a name.

@stack72 thoughts?

fatmcgav:
There is a 'VersionHeader' const defined here: https://github.com/hashicorp/terraform/blob/master/terraform/version.go#L24

But it still needs to be used to construct a header... 
Agree that a helper function would be a useful addition..

stack72:
@fatmcgav / @jtopjian IMO, I don't think it's so important - the AWS one is more as a partner thing so they can understand when the tool is used

But it would be nice to follow a similar convention

fatmcgav:
@stack72 / @jtopjian Thoughts on https://github.com/fatmcgav/terraform/commit/76fbe83d87f9049d2005bc83f12d2caea59cc182

stack72:
That looks good to me :)

fatmcgav:
Cool, ok.. I'll see if I can stick a quick test case on it... 

Would it be better as a separate PR, or happy for it to be included in this one?

stack72:
You can include it here :)

fatmcgav:
Cool, ok... PR updated :)

jtopjian:
Nice - thanks @fatmcgav! You're on a roll :)

PR #14956, opened 2017-05-31
provider/openstack: Sort request/response headers whilst debugging

fatmcgav:
@jtopjian Another small one, makes reading debug logs a bit more consistent :)


jtopjian:
LGTM! Thanks :)

Issue #14961, opened 2017-05-31
aws_lambda_function with dead_letter_config causes a crash


brikis98:
### Terraform Version

terraform v0.9.3, terraform v0.9.6

### Affected Resource(s)

- aws_lambda_function

### Terraform Configuration Files

```hcl
resource "aws_lambda_function" "function" {
  function_name = "example"

  filename = "${data.archive_file.source_code.output_path}"
  source_code_hash = "${data.archive_file.source_code.output_base64sha256}"

  runtime = "python2.7"
  handler = "index.handler"

  dead_letter_config {
    target_arn = "${var.dead_letter_target_arn}"
  }
}

variable "dead_letter_target_arn" {
  default = ""
}
```

### Panic Output

Crash log and stderr: https://gist.github.com/brikis98/77c5c8d20f4b54db209abf5a6d4e33d6

### Steps to Reproduce

1. Set the `dead_letter_target_arn ` to an empty string
1. `terraform apply`

PR #14964, opened 2017-05-31
provider/aws: Fix panic on nil dead_letter_config

grubernaut:
Fixes a panic where specifying a nil `target_arn` for a `dead_letter_config` inside the `aws_lambda_function` resource would throw a panic.
Now, we return a nice error to the user instead of throwing a panic and stacktrace.

```
$ make testacc TEST=./builtin/providers/aws TESTARGS="-run=TestAccAWSLambdaFunction_nilDeadLetterConfig"
2017/05/31 10:22:26 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws -v -run=TestAccAWSLambdaFunction_nilDeadLetterConfig -timeout 120m
=== RUN   TestAccAWSLambdaFunction_nilDeadLetterConfig
--- PASS: TestAccAWSLambdaFunction_nilDeadLetterConfig (20.86s)
PASS
ok      github.com/hashicorp/terraform/builtin/providers/aws    20.884s
```
Fixes: #14961

### DIFF
--- a/builtin/providers/aws/resource_aws_lambda_function.go
+++ b/builtin/providers/aws/resource_aws_lambda_function.go
@@ -263,6 +263,10 @@ func resourceAwsLambdaFunctionCreate(d *schema.ResourceData, meta interface{}) e
 	if v, ok := d.GetOk("dead_letter_config"); ok {
 		dlcMaps := v.([]interface{})
 		if len(dlcMaps) == 1 { // Schema guarantees either 0 or 1
+			// Prevent panic on nil dead_letter_config. See GH-14961
+			if dlcMaps[0] == nil {
+				return fmt.Errorf("Nil dead_letter_config supplied for function: %s", functionName)
+			}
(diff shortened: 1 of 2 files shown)


brikis98:
Thx for the quick fix!

One question: instead of an error, is it possible to just disable the dead letter queue setting if an empty string is passed in? Otherwise, since the dead letter queue settings are in an inline block, there is no easy way to make this a configurable setting in a Terraform module. I brought this up before in #14037, but it sounds like a general-purpose solution for this sort of thing is some ways off, so perhaps in the short term, we can make things a bit easier?

egarbi:
@brikis98 do you know any workaround to avoid to create a different module only to add dead_letter_config support? Agree with you this is really annoying.

brikis98:
@egarbi Unfortunately, I haven't found any better solutions :-\

PR #14966, opened 2017-05-31
Update lb_monitor_v2.html.markdown

florianlechner:
added missing required parameter pool_id and updated code sample.


jtopjian:
Thanks for catching this!

PR #14973, opened 2017-05-31
Allow heroku spaces to be imported

vanstee:
Looks like this was skipped over when adding the heroku space resource. Added an importer which is basically just the read function with a modified return value.


builtin/providers/heroku/resource_heroku_space.go
```diff
@@ -128,6 +132,19 @@ func resourceHerokuSpaceDelete(d *schema.ResourceData, meta interface{}) error {
...
+
+	spaceRaw, _, err := SpaceStateRefreshFunc(client, d.Id())()
+	if err != nil {
+		return nil, err
+	}
+	space := spaceRaw.(*heroku.Space)
+
+	setSpaceAttributes(d, space)
+	return []*schema.ResourceData{d}, nil
+}
```
alindeman: @vanstee Since this function is basically identical to `resourceHerokuSpaceRead`, is it possible to use [`schema.ImportStatePassthrough`](https://github.com/hashicorp/terraform/blob/f27aa3c826b02b84c8b5d67efd15479504df08c1/builtin/providers/github/resource_github_repository.go#L18-L20) (example from the github provider linked)?

`schema.ImportStatePassthrough` basically reuses the `Read` function to do the import. As long as the `Read` function only uses `d.Id()` to do its work (and not, for example, `d.Get("something")`), it should work properly.

vanstee: I think you're right. Fixing. Thanks!

vanstee:
@catsby Should be good to go. Thanks for the review.

PR #14975, opened 2017-05-31
docs/google: Add "list" to tags description

alessfg:
The tags argument requires a list in order to work.


stack72:
LGTM - thanks for this :)

Issue #14998, opened 2017-06-01
resource google_sql_user import fails in 0.9.6 with multiple projects and providers


devth:
### Terraform Version

v0.9.6

### Affected Resource(s)

- google_sql_user

### Expected Behavior

`google_sql_user` resource type should support import on specific projects.

### Actual Behavior

When I plan, Terraform correctly shows that it should create the `google_sql_user` resource in the correct project, `foo` (this is a separate project than the one that the service account belongs to, but the service account has access to `foo` as well):

```
+ google_sql_user.mysql
    host:     "cloudsqlproxy~%"
    instance: "mysql-us-central1-m"
    name:     "mysql"
    password: "<sensitive>"
    project:  "foo"
```

When I try to run the corresponding import:

```
terraform import google_sql_user.mysql mysql-us-central1-m/mysql
```

With debug logs, it shows that the API request is actually being made against the main provider project instead of the project that the resource exists in.

Then I noticed `terraform import` has a `-provider` flag. I used this to specify the provider for the corresponding project:

```
terraform import -provider="foo.google_sql.us-central1" \
  google_sql_user.mysql mysql-us-central1-m/mysql
```

This errors with:

```
provider foo.google_sql.us-central1 couldn't be found
```

even though the current working directory contains a file `gcp.tf.json` which contains:

```
   "provider": {
      "google": {
         "alias": "foo.google_sql.us-central1",
         "project": "foo",
         "region": "us-central1"
      }
   },
```

and I have verified that my multi-provider / multi-project configuration has worked to create GCP Projects and resources under the correct projects.

danawillow:
@rileykarson want to take a look at this?

rileykarson:
Hi @devth! 

It looks like the `-provider` flag on `terraform import` expects the argument to have a value of `TYPE.ALIAS`, like the `aws.eu` alias in [their example](https://www.terraform.io/docs/commands/import.html#provider-provider).

Could you try running this command and let me know if it works?

    terraform import -provider="google.foo.google_sql.us-central1" \
      google_sql_user.mysql mysql-us-central1-m/mysql`

devth:
Hi @rileykarson, thanks for the tip. That was indeed the problem. However, after importing the resource, when I run `plan`, Terraform is unable to fetch the state for the imported sql user. It gives the error:

```bash
[...]

PR #15030, opened 2017-06-02
[docs] Updated import command provider to specify syntax

rileykarson:
Updated import command provider flag documentation to explicitly specify syntax.

The [providers documentation](https://www.terraform.io/docs/configuration/providers.html#multiple-provider-instances) explicitly specifies the format of the value for the `provider` field in a resource, but the [import provider flag](https://www.terraform.io/docs/commands/import.html#provider-provider) documentation does not.

It was unclear to me when first looking at the page that `aws.eu` meant an `aws` provider with the alias `eu`, and not an alias of `aws.eu`. This way, the formatting of the expected value should be more clear for readers in the future.

Related to #14998

### DIFF
--- a/website/source/docs/commands/import.html.md
+++ b/website/source/docs/commands/import.html.md
@@ -48,8 +48,8 @@ The command-line flags are all optional. The list of available flags are:
 
 * `-no-color` - If specified, output won't contain any color.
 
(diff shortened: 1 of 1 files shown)


stack72:
LGTM! thanks for this - it makes it more clear :)

PR #15032, opened 2017-06-02
change init args to remove source copy

### LINKED ISSUE #14066: terraform init can't deal with specified config path
We have an `ops` repo with `packer/` and `terraform/` subdirectories. We keep the terraform state in the repo root within `.terraform/` and run our terraform commands from the root, but all terraform configuration is inside the `terraform/` subdirectory. The reason is that these commands are usually run by automated tools from the source code root, and keeping the entry point in the repo root keeps it simple for new folks to on board, but we want to keep terraform configuration encapsulated.

This used to run fine in <= 0.8:

```sh
$ terraform remote config -backend=S3
  -backend-config=region=us-east-1
  -backend-config=bucket=ops.example.com
  -backend-config=key=terraform/terraform.tfstate

$ terraform plan -out .terraform/terraform.tfplan terraform/

$ terraform apply .terraform/terraform.tfplan
```

Now since 0.9+ instead of doing a remote config we do an init. We added a configuration file with the correct backend options. But the init command can't seem deal with configuration being in a sub-directory. Supplying SOURCE works fine but copies all files into current directory leaving a dirty git tree:

```
$ terraform init terraform/
Initializing configuration from: "terraform/"...
Initializing the backend...

should now work.

$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   terraform/backend.tf

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	files/ templates/ and.tf a.tf bunch.tf of.tf files.tf

$ git clean -fd

$ terraform plan terraform/

# everything back to normal...
```

Supplying SOURCE and PATH doesn't work, pretty expected:

```
$ terraform init terraform/ terraform/
Initializing configuration from: "terraform/"...
Error copying source: The destination path contains Terraform configuration files. The init command
with a SOURCE parameter can only be used on a directory without existing
Terraform files.

Please resolve this issue and try again.
exit status 1
```

We really just need to supply a PATH without a SOURCE and without requiring a fetch. Something like this might work:

```
$ terraform init - terraform/
```

Or the arguments could be flags instead of positional.

```
$ terraform init -path terraform/
```

jbardin:
Documentation changes haven't been made yet, since those are more likely to run into merge conflicts before this branch is merged.

When init was modified in 0.9 to initialize a terraform working
directory, the legacy behavior was kept to copy or fetch module sources.
This left the init command without the ability that the plan and apply
commands have to target a specific directory for the operation.

This commit removes the legacy behavior altogether, and allows init to
target a directory for initialization, bringing it into parity with plan
and apply. If one wants to copy a module to the target or current
directory, that will have to be done manually before calling init. We
can later reintroduce fetching modules with init without breaking this new 
behavior, by adding the source as an optional second argument.

The unit tests testing the copying of sources with init have been
removed, as well as some out of date (and commented out) init tests
regarding remote states.

Fixes #14066


brikis98:
Deleting this behavior is a major backwards incompatibility. It seems like at the minimum, this should have had a deprecation warning for one release, and then been removed. Better yet:

> We can later reintroduce fetching modules with init without breaking this new behavior, by adding the source as an optional second argument.

Can this be added *before* 0.10.0 comes out? Without this, this change is completely backwards incompatible and has no workaround. Re-implementing the fetch behavior of Terraform (i.e. with support for http, git, versioning, etc) is not something you can replace with a one-liner!

PR #15034, opened 2017-06-02
provider/datadog: make datadog_user verified a computed attribute

phinze:
It's read-only and generates supurious diffs for verified users.

### DIFF
--- a/builtin/providers/datadog/resource_datadog_user.go
+++ b/builtin/providers/datadog/resource_datadog_user.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"strings"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/hashicorp/terraform/helper/schema"
 	"gopkg.in/zorkian/go-datadog-api.v2"
 )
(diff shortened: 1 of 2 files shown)


builtin/providers/datadog/resource_datadog_user.go
```diff
@@ -107,7 +108,7 @@ func resourceDatadogUserRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	log.Printf("[DEBUG] user: %v", u)
+	log.Printf("[DEBUG] user: %s", spew.Sdump(u))
```
stack72: Can we remove the spew.Dump here? Think this is debug work ;)

phinze: I only left it in as the `#v` output is useless - it's just a bunch of memory addresses. Should we remove the entire line?

phinze:
> Do the docs need updated as well?

Docs already say this is an output-only attribute so this is just fixing up the code to agree with them! 🔧 

https://www.terraform.io/docs/providers/datadog/r/user.html#verified

builtin/providers/datadog/resource_datadog_user.go
stack72: I just removed the entire debug line as it's pretty much useless :) <3 for the fixing of the debug though

PR #15045, opened 2017-06-04
Url to aws-two-tier example changed.

brettski:
Provided what I believe to the the new correct address for the example url.


stack72:
Hi @brettski 

This LGTM! thanks for this :)

Paul

PR #15046, opened 2017-06-04
Make it possible to generate local copy of website documentation

dradetsky:
I like to have docs locally on my machine, so I tried to build a copy of the terraform docs to go with the other stuff I've been working on lately. As it turns out, this isn't possible because of the way y'all's middleman setup & link format works.

Basically, because all the links are absolute, they work fine off a server, but not as local html documents. The solution is to strip off all their leading slashes, and then set `<base href="/">` in the root layout. Then you arrange to set the relative url base differently for the local build. To build locally, you run `MM_ENV=local_build make build`. Unfortunately, the version of middleman used by hashicorp wipes out the environment filed during build, but I just check ENV directly.

It's also necessary to disable turbolinks locally, since it doesn't actually know it can't fetch file urls, and I didn't know any better way of doing that than making a separate local-application.js without it.

Finally, to make it look right, you have to configure build such that for local builds, an "external assets host" (the local filesystem) is used.

To review this, just look at the first commit. The second is just the result of running `find source -name *.erb -exec sed -i -e 's/href="\//href="/' {} \;`.

I tried serving the non-local build to myself with a local nginx, and it seems the same as terraform.io.


dradetsky:
forgot about the markdown links

dradetsky:
@sethvargo don't merge this yet; i think i know a better way to do it that makes the local build relocatable

dradetsky:
Leaving this open in case I realize #15054 doesn't work properly

sethvargo:
In light of the conversation in #15054, I'm going to close this out.

PR #15050, opened 2017-06-04
provider/alicloud: Fix access_key NOT FOUND bug and nat_gateway diff bug

xiaozhu36:
This PR fix the following bugs:
1. When setting access_key to "", running 'terraform apply' will raise ACCESS KEY NOT FOUND error.  And the PR fix the bug by using ENV ALICLOUD_ACCESS_KEY and ALICLOUD_SECRET_KEY.
2. When creating nat gateway, if set "zone" is nil, diff resource will appear error. Now, the PR set "zone" Compute is true.

The running result of nat gateway testcase as follows:

TF_ACC=1 go test ./alicloud -v -run=TestAccAlicloudNatGateway -timeout 120m
=== RUN   TestAccAlicloudNatGateway_basic
--- PASS: TestAccAlicloudNatGateway_basic (76.65s)
=== RUN   TestAccAlicloudNatGateway_spec
--- PASS: TestAccAlicloudNatGateway_spec (69.36s)
PASS
ok      github.com/alibaba/terraform-provider/alicloud  146.017s

### DIFF
--- a/builtin/providers/alicloud/provider.go
+++ b/builtin/providers/alicloud/provider.go
@@ -5,6 +5,7 @@ import (
 	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
+	"os"
 )
 
 // Provider returns a schema.Provider for alicloud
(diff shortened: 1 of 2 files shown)


stack72:
Hi @xiaozhu36 

This LGTM! thanks for this :)

Paul

PR #15051, opened 2017-06-04
provider/alicloud: Add new example alicloud-build-lnmp

xiaozhu36:
This PR add a new template of using alicloud provider to build and install LNMP


stack72:
Thanks for this @xiaozhu36 :) This LGTM!

Issue #14522, opened 2017-05-15
Cannot create aws_security group with both IPv4 and IPV6 open egress


ondrap:
This happens on Terraform v0.9.5.

Affected resources: aws_security_grup

```hcl
resource "aws_security_group" "test" {
  ingress {
      from_port = 22
      to_port = 22
      protocol = "6"
      cidr_blocks = ["0.0.0.0/0"]
  }
  egress {
    from_port       = 0
    to_port         = 0
    protocol        = "-1"
    cidr_blocks  = ["0.0.0.0/0"]
  }
  egress {
    from_port       = 0
    to_port         = 0
    protocol        = "-1"
    ipv6_cidr_blocks  = ["::/0"]
  }
  tags {
      Name = "test"
  }
}
```

### Debug Output
https://gist.github.com/ondrap/888e2dcc0018dcd9f99990eb5a0af1cb

Normal output:
https://gist.github.com/ondrap/660d1f8b14960f6772ce38b364819d4f

### Expected Behavior

Create a security group with open egress rules for 0.0.0.0/0 and ::/0 (both IPv4 and IPv6).

### Actual Behavior

It crashed. When I remove the IPv6 rule, it gets created correctly. When I `terraform apply` this rule again, the IPv6 egress gets removed.

It seems to me that it has something to do with the default; when no egress rule is used, an open IPv6 egress rule is created (I do have IPv6 enabled on the VPC and subnets).

### Steps to Reproduce

1. `terraform apply`

### References
 
I looked into #14382 where it is suggested to use the `ipv6_cidr_blocks` for IPv4 addresses and it doesn't work for me.

stack72:
output:

```
* aws_security_group.test: Error authorizing security group egress rules: InvalidPermission.Duplicate: the specified rule "peer: ::/0, ALL, ALLOW" already exists
	status code: 400, request id: 03df8e5a-9fbe-4bc6-b922-9f80e89b2306
```

ravenbyron:
I think this is related to default egress rules and maybe this code https://github.com/hashicorp/terraform/pull/1765 needing to be updated for ipv6...

This is just a guess as I just ran into the same issues, only happens on the initial run...

shepdelacreme:
I'm experiencing the same issue on Terraform v0.9.6. Can recreate by just creating an empty security group in a VPC with IPv6 enabled.

```
resource "aws_security_group" "sg" {
  description = "testing"
  name = "test-test"
[...]

PR #15075, opened 2017-06-05
provider/aws: Revoke default ipv6 egress rule for aws_security_group

stack72:
Fixes: #14522

To follow similar work in IPv4, we are now going to revoke the default
IPv6 egress rule from an empty AWS security group

```
% make testacc TEST=./builtin/providers/aws/ TESTARGS='-run=TestAccAWSSecurityGroup_ipv4andipv6Egress'
2017/06/05 14:01:52 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/aws/ -v -run=TestAccAWSSecurityGroup_ipv4andipv6Egress -timeout 120m
=== RUN   TestAccAWSSecurityGroup_ipv4andipv6Egress
--- PASS: TestAccAWSSecurityGroup_ipv4andipv6Egress (63.39s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/aws	63.423s
```

### DIFF
--- a/builtin/providers/aws/resource_aws_security_group.go
+++ b/builtin/providers/aws/resource_aws_security_group.go
@@ -295,6 +295,34 @@ func resourceAwsSecurityGroupCreate(d *schema.ResourceData, meta interface{}) er
 				d.Id(), err)
 		}
 
+		log.Printf("[DEBUG] Revoking default IPv6 egress rule for Security Group for %s", d.Id())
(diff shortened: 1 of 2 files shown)


zzq889:
AWS region `cn-north-1` has no IPv6 which may cause the revoke error. Is it possible to provide default property like `revoke_default_ipv6` to allow us disable this behavior?

PR #15077, opened 2017-06-05
Allow resizing of Google Cloud persistent disks

bashtoni:
```
make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccComputeDisk_updateSize'
2017/06/05 11:31:09 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccComputeDisk_updateSize -timeout 120m
=== RUN   TestAccComputeDisk_updateSize
--- PASS: TestAccComputeDisk_updateSize (28.37s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	28.377s
```

### DIFF
--- a/builtin/providers/google/resource_compute_disk.go
+++ b/builtin/providers/google/resource_compute_disk.go
@@ -23,6 +23,7 @@ func resourceComputeDisk() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceComputeDiskCreate,
 		Read:   resourceComputeDiskRead,
+		Update: resourceComputeDiskUpdate,
 		Delete: resourceComputeDiskDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
@@ -68,7 +69,6 @@ func resourceComputeDisk() *schema.Resource {
 			"size": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
-				ForceNew: true,
 			},
 
 			"self_link": &schema.Schema{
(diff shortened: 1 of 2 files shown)


builtin/providers/google/resource_compute_disk.go
```diff
@@ -68,7 +69,7 @@ func resourceComputeDisk() *schema.Resource {
 			"size": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
-				ForceNew: true,
+				ForceNew: false,
```
stack72: This can be removed completely :)

builtin/providers/google/resource_compute_disk.go
```diff
@@ -185,6 +186,28 @@ func resourceComputeDiskCreate(d *schema.ResourceData, meta interface{}) error {
...
 
+func resourceComputeDiskUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	project, err := getProject(d, config)
+	if err != nil {
+		return err
+	}
+
+	if d.HasChange("size") {
```
stack72: Does the machine have to be stopped? Do we have to wait for a specific status before we know that this has worked as expected?

bashtoni: Resizing is an on-line operation. The API docs don't specify the possible states of a disk, and I've never been able to observe another state than 'Ready'.

bashtoni:
@stack72 Thanks for reviewing Paul. I removed the unnecessary `ForceNew: false`.

stack72:
LGTM! thanks for making the change and the explanation :)

P.

PR #15086, opened 2017-06-05
provider/openstack: Optimize the printing of request/response headers…

fatmcgav:
… when debugging Openstack HTTP requests.

Background is I hit an issue with debugging a container/object creation issue, and the `request` headers weren't being output as `request.Body` was `nil`.

@jtopjian I've tested locally and confirmed the changes work, but let me know your thoughts :)

### DIFF
--- a/builtin/providers/openstack/types.go
+++ b/builtin/providers/openstack/types.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


jtopjian:
@fatmcgav Thanks! I didn't have a chance to check this today, but I'll definitely do it tomorrow.

fatmcgav:
@jtopjian No worries... I've just added an additional commit to log the response code aswell :)

jtopjian:
@fatmcgav This is _really_ nice - thanks!

I'm going to go ahead and merge this. There's a new release being cut in a few hours, so this will just make it. :)

PR #15094, opened 2017-06-05
provider/azurerm: Add diff supress func to azurerm_traffic_manager_endpoint.endpoint_location.

rrudduck:
The default locationSchema function applies the diff supression func. Since this resource cannot use the default schema it should also apply the diff suppression func to avoid normalization issues.


stack72:
Hi @rrudduck 

Thanks for the work here - please can you add an acceptance test that demonstrates this new behaviour and how it works

Thanks

Paul

stack72:
Hi @rrudduck 

Thanks for the work here - I just applied this patch on the new repo https://github.com/terraform-providers/terraform-provider-azurerm

Paul

PR #15095, opened 2017-06-05
provider/google: Update compute_disk to read after update, always set size

catsby:
A number of `compute_disk` acceptances test are failing because of mishandling of `snapshot`. We need to store the full link returned to us and not parse it, because we can't know what we got, but we'll always get the full url back. Should update docs too but I will have a follow up PR regarding waiting for size change that should address this and will come tomorrow.


danawillow:
Hey @catsby, looks like acceptance tests are still broken. I haven't investigated too much so I'm not sure if that's because or in spite of this change. Mind taking a look?

PR #15096, opened 2017-06-05
Update link to canned GCS ACL's

danielcompton:
This points to the new location for the list of canned ACL's.


stack72:
Thanks for this @danielcompton :)

Issue #15027, opened 2017-06-02
google_compute_region_backend_service requires an compute instance in the Instance group


oppegard:
### Terraform Version
0.9.6

### Affected Resource(s)
- google_compute_region_backend_service
- google_compute_instance_group

### Terraform Configuration Files
```hcl
resource "google_compute_instance_group" "instance_group" {
  name = "instance-group"
  zone = "us-central1-a"
}

resource "google_compute_region_backend_service" "backend_service" {
  name             = "backend-service"
  protocol         = "TCP"
  timeout_sec      = 30
  session_affinity = "NONE"

  backend {
    group = "${google_compute_instance_group.instance_group.self_link}"
  }

  health_checks = ["${google_compute_health_check.health_check.self_link}"]
}

resource "google_compute_health_check" "health_check" {
  name = "health-check"

  timeout_sec        = 1
  check_interval_sec = 1

  tcp_health_check {
    port = "8983"
  }
}

```

### Expected Behavior
Terraform apply should have created a google_compute_region_backend_service.

### Actual Behavior
Terraform plan output:
```
+ google_compute_health_check.health_check
    check_interval_sec:              "1"
    healthy_threshold:               "2"
    name:                            "health-check"
    project:                         "<computed>"
    self_link:                       "<computed>"
    tcp_health_check.#:              "1"
    tcp_health_check.0.port:         "8983"
    tcp_health_check.0.proxy_header: "NONE"
    timeout_sec:                     "1"
    unhealthy_threshold:             "2"

+ google_compute_instance_group.instance_group
    name:      "instance-group"
    network:   "<computed>"
    self_link: "<computed>"
    size:      "<computed>"
    zone:      "us-central1-a"

+ google_compute_region_backend_service.backend_service
    backend.#:                       "1"
    backend.~1938307039.description: ""
    backend.~1938307039.group:       "${google_compute_instance_group.instance_group.self_link}"
    fingerprint:                     "<computed>"
    health_checks.#:                 "<computed>"
    name:                            "backend-service"
    protocol:                        "TCP"
    self_link:                       "<computed>"
    session_affinity:                "NONE"
    timeout_sec:                     "30"
```

Terraform apply error:
```
    Error applying plan:

    1 error(s) occurred:

    * google_compute_region_backend_service.backend_service: 1 error(s) occurred:

[...]

PR #15097, opened 2017-06-05
provider/google: Add note about instance group restrictions in region backend service docs.

rileykarson:
Update the regional backend service documentation to reflect that instance groups cannot be empty.

Related to #15027 
@danawillow


stack72:
LGTM! Thanks for this @rileykarson :)

PR #15099, opened 2017-06-05
Local, relocatable, fast-building website copy

dradetsky:
(maybe) See previous prs #15054 and #15046 

To address @sethvargo's concerns about the previous attempt to make it possible to build the docs for use off the local filesystem, rather than actually checking in changes to all links on the website, we just use sed-in-place to change them all to link_to helpers to before executing the local build (except for a few special cases where anchors are wrapped around `inline_svg` calls).

We also revert the source directory after running a local build, to prevent people from accidentally checking in all the changes @sethvargo wanted to prevent in the first place. Since the primary use-case here is packaging released versions of the website/docs (people who want to preview docs locally should use `make website`), this shouldn't be a major problem.

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -7,6 +7,7 @@ build:
 		--rm \
 		--tty \
 		--volume "$(shell pwd):/website" \
+		-e MM_ENV=$(MM_ENV) \
 		hashicorp/middleman-hashicorp:${VERSION} \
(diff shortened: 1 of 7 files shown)


sethvargo:
In light of the conversation in #15054, I'm going to close this out.

Issue #15056, opened 2017-06-05
google_compute_autoscaler:  Error 400: Required field 'autoscaler' not specified


mattes:
### Terraform Version
Terraform v0.9.6

### Affected Resource(s)
- google_compute_autoscaler

### Terraform Configuration Files
```hcl
resource "google_compute_autoscaler" "primary" {
  name   = "primary"
  zone   = "${google_container_cluster.primary.zone}"
  target = "${google_compute_instance_group_manager.primary.self_link}"

  autoscaling_policy = {
    max_replicas    = 5
    min_replicas    = 4
    cooldown_period = 60

    cpu_utilization {
      target = 0.75
    }
  }
}
```

### Output
```
google_compute_autoscaler.primary: Error updating Autoscaler: googleapi: Error 400: Required field 'autoscaler' not specified, required
```

danielcompton:
That autoscaler looks almost identical to my working config. Can you share more of your config, I suspect there is something else that is going on here?

mattes:
```hcl
resource "google_container_cluster" "primary" {
  name               = "primary"
  zone               = "us-central1-a"
  initial_node_count = 3

  master_auth {
    username = "${var.cluster_master_auth_username}"
    password = "${var.cluster_master_auth_password}"
  }

  node_config {
    machine_type = "n1-standard-1"
    disk_size_gb = 100
  }
}


resource "google_compute_instance_group_manager" "primary" {
  name               = "${var.google_compute_instance_group_manager_primary_name}"
  zone               = "${google_container_cluster.primary.zone}"
  instance_template  = "https://www.googleapis.com/compute/v1/projects/${var.google_project_id}/global/instanceTemplates/${replace(var.google_compute_instance_group_manager_primary_name, "/-grp$/", "")}"
  base_instance_name = "${replace(var.google_compute_instance_group_manager_primary_name, "/-grp$/", "")}"

  target_size = 3
  named_port {
    name = "http-to-https-port"
    port = 31000
  }
}
```

If it's my config, maybe the `target_size` is causing problems?

mattes:
Just tested without `target_size`. Same error.

rileykarson:
Hi @mattes,
The error isn't in your config - it's because Terraform is trying to perform an update, but that functionality is broken right now. You can work around this for now by changing the name to force a new resource, or by manually running `terraform destroy` with your resource as a `-target` parameter, followed by `terraform apply`.

[...]

PR #15101, opened 2017-06-06
provider/google: Make google_compute_autoscaler use Update instead of Patch.

rileykarson:
It looks like our vendored code thinks that Patch has "autoscaler" as an optional field, but the [documentation](https://cloud.google.com/compute/docs/reference/latest/autoscalers/patch) suggests that it is required. It probably regressed when we vendored a [new copy of the client code](https://github.com/hashicorp/terraform/commit/c9640e40dfb7b282246ea18af4bfac0384d60b78#diff-981842c12447ae1d1f29443c53b63787R336) around a month ago.

We didn't catch this because the `_update` test didn't perform an `Update` - it performed a `Destroy` followed by a `Create` instead. This PR updates the tests so that an actual `Update` is performed.

Patch doesn't make much sense in this context - we pass in an entire copy of the resource, so we should be using Update anyways.

Fixes: #15056
@danawillow

### DIFF
--- a/builtin/providers/google/import_compute_autoscaler_test.go
+++ b/builtin/providers/google/import_compute_autoscaler_test.go
@@ -3,19 +3,26 @@ package google
 import (
 	"testing"
 
+	"fmt"
+	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
-func TestAccAutoscaler_importBasic(t *testing.T) {
+func TestAccComputeAutoscaler_importBasic(t *testing.T) {
 	resourceName := "google_compute_autoscaler.foobar"
+
+	var it_name = fmt.Sprintf("autoscaler-test-%s", acctest.RandString(10))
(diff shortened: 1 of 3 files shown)


stack72:
Hi @rileykarson 

thanks for this work - this is a great change

```
% make testacc TEST=./builtin/providers/google TESTARGS='-run=TestAccComputeAutoscaler'
2017/06/06 15:12:23 Generated command/internal_plugin_list.go
TF_ACC=1 go test ./builtin/providers/google -v -run=TestAccComputeAutoscaler -timeout 120m
=== RUN   TestAccComputeAutoscaler_importBasic
--- PASS: TestAccComputeAutoscaler_importBasic (136.75s)
=== RUN   TestAccComputeAutoscaler_basic
--- PASS: TestAccComputeAutoscaler_basic (122.07s)
=== RUN   TestAccComputeAutoscaler_update
--- PASS: TestAccComputeAutoscaler_update (148.29s)
PASS
ok  	github.com/hashicorp/terraform/builtin/providers/google	407.126s
```

Paul

PR #15108, opened 2017-06-06
Correct openstack_lbaas_* to openstack_lb_* sidebars to reflect correct resource type names.

philsttr:
I missed this update earlier in PR #14196


jtopjian:
@philsttr Thanks for cleaning these up. I think you've updated everything here :)

PR #15138, opened 2017-06-07
provider/aws: Fix link and add better reference to a parameter in RDS docs

endofcake:
The current link leads to the main page, which is not terribly useful.
Also provided a more verbose explanation of what `backup_window` means and how it can be used.


stack72:
LGTM! thanks for this @endofcake :)

PR #15170, opened 2017-06-07
provider/google: Add an additional delay when checking for sql operations

### LINKED ISSUE #13091: Randomly fails to create google_sql_database_instance due to timing issue
Hi there,

### Terraform Version
Terraform v0.9.1

### Affected Resource(s)
- google_sql_database_instance

```hcl

provider "google" {
  credentials = "${file("account.json")}"
  region      = "europe-west-1"
  project = "${data.terraform_remote_state.project.project_id}"
}

resource "google_sql_database_instance" "master" {
  region = "europe-west1"
  project = "${data.terraform_remote_state.project.project_id}"
  database_version = "MYSQL_5_7"

  settings {
    tier = "db-n1-standard-1"
    activation_policy = "ALWAYS"
  }
}

data "terraform_remote_state" "project" {
  backend = "gcs"
  config {
    path = "projects/dev.tfstate"
    bucket = "application-cloud-tf-state"
  }
}
```

### Debug Output
https://gist.github.com/mogronalol/c8c308e17b390023939d85ba9c5853a3

### Expected Behavior
Successfully retrieve the instance creation operation, then block until it is complete.

### Actual Behavior
A `404` because it does a `GET` too quickly, so the operation does not exist yet.
https://github.com/hashicorp/terraform/blob/bfdeae0e3357f50a248806c0ff79c85f37215f79/builtin/providers/google/resource_sql_database_instance.go#L570-L570 is the line which fails.

### Steps to Reproduce
1. `terraform apply`

This does not always fail. It will pretty much never fail if you add a sleep before https://github.com/hashicorp/terraform/blob/bfdeae0e3357f50a248806c0ff79c85f37215f79/builtin/providers/google/resource_sql_database_instance.go#L570-L570 then it doesn't fail. 

What does not make sense, is that if I change `SqlAdminOperationWater.RefreshFunction` to retry on a `404` instead of failing, it will still be a `404` after five attempts. But, a simple sleep before calling `RefreshFunc` for the first time means that there is no `404` so I'm a little stumped.

danawillow:
This should fix #13091, as well as some of our failing tests.

The failures are happening because for some reason, creating an operation and then immediately reading it back does not always succeed. This adds a bit of a delay between the create and read portions of the operation. We may need to fiddle with the numbers a bit in the future (this is hard to test since _most_ of the time, things do work as expected), but it's at least a start.

### DIFF
--- a/builtin/providers/google/sqladmin_operation.go
+++ b/builtin/providers/google/sqladmin_operation.go
@@ -66,6 +66,7 @@ func sqladminOperationWait(config *Config, op *sqladmin.Operation, activity stri
 	state := w.Conf()
 	state.Timeout = 10 * time.Minute
 	state.MinTimeout = 2 * time.Second
+	state.Delay = 5 * time.Second
 	opRaw, err := state.WaitForState()
 	if err != nil {
 		return fmt.Errorf("Error waiting for %s (op %s): %s", activity, op.Name, err)


stack72:
LGTM! Thanks for this @danawillow :)

PR #15182, opened 2017-06-08
[docs] Group/prefix HashiCorp providers.

clstokes:
This PR groups the HashiCorp providers in the documentation so that they're easier to find or discover. Also changes the main heading and description on each of the product pages to be consistent.


stack72:
Hi @clstokes 

Now that the repositories have been split out, I think we need to handle this in another way :) I think this is worth speaking about at HEX since the entire TF team will be there

Let's close this out for now as, in it's current shape, none of the erb files exist anymore - they are each in their individual repos

Paul

PR #15196, opened 2017-06-08
Adds example of subdir notation for generic git

Fitzsimmons:
The subdirectory notation for git is shown for github and bitbucket, but not for generic git. I've just repeated it again for generic git.

I suspect it's the same notation for mercurial as well, but I'm not a mercurial user and don't have a way to verify it.

### DIFF
--- a/website/source/docs/modules/sources.html.markdown
+++ b/website/source/docs/modules/sources.html.markdown
@@ -160,6 +160,16 @@ module "consul" {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
In the current version of the page, this info is separated out as generic info that applies to all protocols: https://www.terraform.io/docs/modules/sources.html#modules-in-package-sub-directories

I think that resolves this issue. Sorry this PR was left hanging!

PR #15203, opened 2017-06-08
provider/aws: Add tag support for ebs_volume_snapshot

catsby:
I was surprised to find we don't support tags on `aws_ebs_snapshot`. Here I add that. Yay


catsby:
@stack72 did so in 2113681

catsby:
Closed in favor of terraform-providers/terraform-provider-aws#3

PR #15206, opened 2017-06-08
provider/azurerm: Add missing resources to importability section

thiagocaiubi:
First of all, thanks for Terraform!

While reading the docs I noticed an inconsistency between `Import` section under each resource document and the `Resource Importability` document.

Example:

* [Import azurerm_virtual_machine](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html#import)
* [Resource Importability](https://www.terraform.io/docs/import/importability.html#azure-resource-manager-)

This PR is based on other documentation that might be outdated as well. How I get there:

```console
$ git grep "terraform import azurerm"
```
[Output](https://gist.github.com/thiagocaiubi/6a527fe658351e7db56e1e89b0afa334)

Every resource that has an `Import` section with an example was promoted to `Resource Importability` section.

Edit:

This kind of documentation can encourage more people to migrate their actual orchestration to Terraform.

### DIFF
--- a/website/source/docs/import/importability.html.md
+++ b/website/source/docs/import/importability.html.md
@@ -107,20 +107,54 @@ To make a resource importable, please see the
 * aws_vpn_connection
 * aws_vpn_gateway
 
-
 ### Azure (Resource Manager)
 
 * azurerm_availability_set
-* azurerm_express_route_circuit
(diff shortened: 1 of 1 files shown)


stack72:
LGTM! Thanks for this :)

Issue #15211, opened 2017-06-09
Add GPG and SSH key to GitHub user data source


mootpt:
Add GPG and SSH key to GitHub user data source:

```
data "github_user" "mootpt" {
  username = "mootpt"
}

output "mootpt-gpg" {
  value = "${data.github_user.mootpt.gpg_key}"
}

output "mootpt-ssh" {
  value = "${data.github_user.mootpt.ssh_key}"
}
```

mootpt:
moved https://github.com/terraform-providers/terraform-provider-github/pull/2

PR #15212, opened 2017-06-09
[WIP] Added GPG and SSH key to GitHub user data source

mootpt:
Pretty straightforward PR. Added GPG and SSH key to GitHub user data source and changed how existing attributes were being set. Setting the attributes using the existing functions better handles situations when the value is nil, something I came across in enterprise.


mootpt:
#15211

mootpt:
do not merge. gonna modify this to return the list of keys rather than using the array literal

mootpt:
moved to https://github.com/terraform-providers/terraform-provider-github/pull/2

PR #15256, opened 2017-06-12
Replace lock_table with dynamodb_table in backend description

conorgil:
Pull request #14949 updated the documentation to change
`lock_table` to `dynamodb_table`, but missed the general
description at the top of the file. This commit updates
the general description so that users know the recommendation
is to use the new `dynamodb_table` key instead of the deprecated
`lock_table` key.


stack72:
Hi @conorgil 

Apologies for not seeing this, I just merged #15299 that took care of it

Thanks

Paul

Issue #15177, opened 2017-06-07
0.9.7 breaks provisioners using template_file with "Must provide one of 'content' or 'source'"


sjwl:
`terraform plan` fails when using a template_file in a file provisioner

### Terraform Version
`0.9.7`

### Affected Resource(s)
All

### Terraform Configuration Files
```hcl
data "template_file" "test" {
  template = "${file("${path.module}/test.tpl")}"
}

resource "test_resource" "test" {
  required = "${count.index}"
  required_map {
    key1 = "value1"
    key2 = "value2"
  }

  provisioner "file" {
    content = "${data.template_file.test.rendered}"
    destination = "foo"
  }
}
```

### Expected Behavior
Plan should succeed with no error.  here is a sample output when using terraform `0.9.6`
```
$GOPATH/bin/terraform_0.9.6 plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.template_file.test: Refreshing state...
The Terraform execution plan has been generated and is shown below.
Resources are shown in alphabetical order for quick scanning. Green resources
will be created (or destroyed and then created if an existing resource
exists), yellow resources are being changed in-place, and red resources
will be destroyed. Cyan entries are data sources to be read.

Note: You didn't specify an "-out" parameter to save this plan, so when
"apply" is called, Terraform can't guarantee this is what will execute.

+ test_resource.test
    computed_from_required:       "<computed>"
    computed_list.#:              "<computed>"
    computed_map.%:               "<computed>"
    computed_read_only:           "<computed>"
    computed_read_only_force_new: "<computed>"
    computed_set.#:               "<computed>"
    optional_computed_force_new:  "<computed>"
    optional_computed_map.%:      "<computed>"
    required:                     "0"
    required_map.%:               "2"
    required_map.key1:            "value1"
    required_map.key2:            "value2"


Plan: 1 to add, 0 to change, 0 to destroy.
```

### Actual Behavior
using 0.9.7...
```
terraform plan                  
1 error(s) occurred:

* test_resource.test: Must provide one of 'content' or 'source'
```

### Steps to Reproduce
1. `terraform plan`

akaspin:
Same issue on 0.9.8.

zambien:
Note that the doc does not state any need for content or source:

https://www.terraform.io/docs/providers/template/d/file.html

Seems like a fairly serious breaking change.

[...]

PR #15257, opened 2017-06-12
Fix file provisioner #15177 Must provide one of 'content' or 'source'

onorua:
it seems [GetOk](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L89) fail to return `ok` in the following part:
https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L91

The struct is:
```
r: {${data.template_file.config.rendered} <nil> true true 0xc4203c12c0}
```

Which means `r.Exists` and `r.Computed` but expecting `r.Exists && !r.Computed`
I was able to make a workaround with the following code. 
Feel free to close or refactor if that doesn't fit your view.

### DIFF
--- a/builtin/provisioners/file/resource_provisioner.go
+++ b/builtin/provisioners/file/resource_provisioner.go
@@ -80,12 +80,16 @@ func applyFn(ctx context.Context) error {
 
 func validateFn(d *schema.ResourceData) (ws []string, es []error) {
 	numSrc := 0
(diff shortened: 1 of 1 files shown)


serialx:
Mentioning #15177 so there exist a reference to it.

serialx:
This is fixed in HEAD. See #15273

PR #15258, opened 2017-06-12
Remove providers from core

jbardin:
This removes all providers from the core repo, except for the test provider, which is only used for its internal tests.

The `core` tag and `core-dev`, `core-test`  make targets have been removed.

The aws and openstack providers are used for remote state, and are now vendored from their new locations.


jbardin:
@apparentlymart: fixed the test provider, so it's like it never left ;)

PR #5207, opened 2016-02-19
[WIP] Better display for set diffs, drop unchanged elements

Bowbaq:
@apparentlymart @phinze initial crack at the problem for #5179

PR #15265, opened 2017-06-12
Better display for set diffs, drop unchanged elements

josephholsten:
Rebase of https://github.com/hashicorp/terraform/pull/5207


jbardin:
Sorry about losing track and not updating this. This was handled in a much more thorough refactoring of the output formatting via #15884

PR #15266, opened 2017-06-12
Add an option that requires `terraform apply` to have an argument

josephholsten:
Rebased https://github.com/hashicorp/terraform/pull/7915


josephholsten:
closed per https://github.com/hashicorp/terraform/pull/7915#issuecomment-307958263

PR #15268, opened 2017-06-13
lifecycle block not in aws_autoscaling_group

dhoeric:


stack72:
Hi @dhoeric 

Thanks for the PR here - our docs may be confusing in this area, but lifecycle_block is actually a [meta parameter](https://www.terraform.io/docs/configuration/resources.html#meta-parameters) that is available in *all* resources

Therefore, I don't think this PR is actually applicable

Thanks

Paul

PR #15274, opened 2017-06-13
Verify plugins can execute

jbardin:
The `init` command should verify that plugins are executable, and have a matching protocol version.

While at some point we may want to filter all the available plugins,
this starts with a simpler UI where specifying an invalid plugin will
be called out immediately, rather than ignored with debug logging as the
only way to diagnose the issue. This also avoids excessive slowness and
UI output if there are numerous old/incompatible plugins in the search path.

### DIFF
--- a/command/init.go
+++ b/command/init.go
@@ -221,12 +221,20 @@ func (c *InitCommand) getProviders(path string, state *terraform.State) error {
 		return errs
 	}
 
+	// Re-discover to see newly-installed plugins.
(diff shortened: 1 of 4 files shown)


jbardin:
I'm also torn cost-benefit here, and this was only implemented verbatim from the RFC. 

I'm going to hold off on merging pending some more real-use experience with the split providers.

jbardin:
We can revive this later if we decide we need this feature.

PR #15289, opened 2017-06-14
init with target dir should still write .terraform in the current directory

jbardin:
Now that init can take a directory for configuration, the old behavior
of writing the .terraform data directory into the target path no longer
makes sense. 

Clean up all references to Meta.dataDir, and only use the getter method
in case we chose to dynamically override this at some point.

### DIFF
--- a/command/init.go
+++ b/command/init.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


jbardin:
Yup. The `init` command needs a big BC entry anyways, as well as doc updates, so this just adds to it.

PR #15329, opened 2017-06-18
fix(builtin/provider): OpenStack

gappaa:
In some providers, OpenStack can't found the right network uuid. In this case, the content of old values of networks are stored and throw an error at the creation. For example, if we create an instance with a private and a public network. OpenStack doesn't provide the right uuid of private network. In this case, Terraform keep the old value, the public uuid and try to assign a private address to a public network, what is not possible to happen.

### DIFF
--- a/vendor/github.com/terraform-providers/terraform-provider-openstack/openstack/resource_openstack_compute_instance_v2.go
+++ b/vendor/github.com/terraform-providers/terraform-provider-openstack/openstack/resource_openstack_compute_instance_v2.go
@@ -1075,13 +1075,16 @@ func getInstanceNetworks(computeClient *gophercloud.ServiceClient, d *schema.Res
 
 		networkID := ""
 		networkName := ""
+		foundInformations := false
 		if tenantNetworkExt {
 			for _, network := range networkList {
 				if network.Name == rawMap["name"] {
 					tenantnet = network
+					foundInformations = true
 				}
 				if network.ID == rawMap["uuid"] {
 					tenantnet = network
+					foundInformations = true
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


vendor/github.com/terraform-providers/terraform-provider-openstack/openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1075,13 +1075,16 @@ func getInstanceNetworks(computeClient *gophercloud.ServiceClient, d *schema.Res
 
 		networkID := ""
 		networkName := ""
+		foundInformations := false
```
jtopjian: Let's change this to `networkFound`.

vendor/github.com/terraform-providers/terraform-provider-openstack/openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1090,15 +1093,20 @@ func getInstanceNetworks(computeClient *gophercloud.ServiceClient, d *schema.Res
...
+		if foundInformations {
+			newNetworks = append(newNetworks, map[string]interface{}{
+				"uuid":           networkID,
+				"name":           networkName,
+				"port":           rawMap["port"].(string),
+				"fixed_ip_v4":    rawMap["fixed_ip_v4"].(string),
+				"access_network": rawMap["access_network"].(bool),
+			})
+		} else {
+			log.Printf("[DEBUG] doenst found tenant net with the name %s or the uuid %s", rawMap["name"], rawMap["uuid"])
```
jtopjian: "Could not find network with the name [%s] or uuid [%s]"

(Using "[ ]" might be a good idea here to signify when one is blank).

jtopjian:
@gappaa Actually, I'm so sorry. I just realized this was opened under `hashicorp/terraform`. The new repo is https://github.com/terraform-providers/terraform-provider-openstack :smile:

gappaa:
Okay. Thanks for the review. I will make the changes and push it in the new repo ;)

gappaa:
@jtopjian I make again the pull request in the new repo :wink:

stack72:
Thanks for opening this against the new repo @gappaa 

I am going to close this out

PR #15352, opened 2017-06-21
allow instance metrics collection in aws-ecs-alb example

wuan:
The StartTelemetrySession action has to be enabled for metrics collection by the ECS agent


stack72:
LGTM thanks for this

PR #15361, opened 2017-06-21
Update docs to include old AWS env vars

otakup0pe:
Looks like the old AWS env vars are still used. This is relevant as in order to make use of `AWS_SHARED_CREDENTIALS_FILE` / `AWS_PROFILE` you must unset _all_ other AWS vars.

Let me know if I should add more clarity to the documentation to this effect. Note that the companion PR https://github.com/terraform-providers/terraform-provider-aws/pull/937 has a bit more context.

### DIFF
--- a/website/docs/backends/types/s3.html.md
+++ b/website/docs/backends/types/s3.html.md
@@ -89,8 +89,8 @@ The following configuration options or environment variables are supported:
  * `acl` - [Canned
    ACL](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl)
    to be applied to the state file.
- * `access_key` / `AWS_ACCESS_KEY_ID` - (Optional) AWS access key.
- * `secret_key` / `AWS_SECRET_ACCESS_KEY` - (Optional) AWS secret access key.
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @otakup0pe 
closing this with full explanation in https://github.com/terraform-providers/terraform-provider-aws/pull/937#issuecomment-310641631

I don't mind you documenting the mentioned clash between `AWS_SHARED_CREDENTIALS_FILE` / `AWS_PROFILE` as long as such suggestion has _no trace_ of the deprecated env vars. 😉 I hope my explanation makes sense.

Thanks.

otakup0pe:
Just so I am clear, PR's that mention the conflict between "all AWS variables, both deprecated and current" so long as we do not mention the old variables?

radeksimko:
That's right. I'm ok with mentioning _all other `AWS_*` environment variables_ or something like that, just not the exact name of the deprecated variables.

It may look odd, but the only motivation is to prevent users from using those deprecated variables at all costs. If they do bump into a problem because they had the deprecated variables set and later discover it was the cause then hopefully they remember it well and solve it by removing all those deprecated variables from everywhere and never experience the same problem ever again.

I understand your (potential) frustration here. Deprecation is hard 😥 🤷‍♂️

Issue #14444, opened 2017-05-12
Targeted aws_db_instance destruction does not respect explicit timeout settings


mattzimmerman:
### Terraform Version
0.9.4

### Affected Resource(s)
- aws_db_instance

### Terraform Configuration Files
```hcl
resource "aws_db_instance" "anonymizer-database" {
  identifier = "db-anonymizer"
  publicly_accessible = false
  instance_class = "${var.qa-db-instance-type}"
  storage_type = "${var.db-storage-type}"
  storage_encrypted = "${var.db-storage-encrypted}"
  
  engine = "${var.db-engine}"
  engine_version = "${var.db-major-engine-version}"
  allocated_storage = "${var.db-storage-limit}"
  username = "${var.rds-username}"
  password = "${var.rds-password}"

  license_model = "${var.db-license-model}"
  multi_az = "${var.db-multi-az}"
  backup_retention_period = "${var.db-backup-rentention-period}"

  skip_final_snapshot = true
  copy_tags_to_snapshot = true
  apply_immediately = "${var.db-apply-immediately}"

  vpc_security_group_ids = ["${aws_security_group.sql-server-anonymizer.id}"]
  db_subnet_group_name = "${data.terraform_remote_state.domain.persistence-subnet-group}"
  
  option_group_name = "${aws_db_option_group.sql-server-anonymizer-options.name}"

  timeouts {
    create = "45m"
    delete = "45m"
    update = "90m"
  }
}
```

### Debug Output
```
aws_db_instance.anonymizer-database: Destroying... (ID: db-anonymizer)
aws_db_instance.anonymizer-database: Still destroying... (ID: db-anonymizer, 10s elapsed)
...
aws_db_instance.anonymizer-database: Still destroying... (ID: db-anonymizer, 20m0s elapsed)
Error applying plan:
1 error(s) occurred:
* aws_db_instance.anonymizer-database (destroy): 1 error(s) occurred:
* aws_db_instance.anonymizer-database: timeout while waiting for resource to be gone (last state: 'deleting', timeout: 20m0s)
```

### Expected Behavior
`aws_db_instance` resource destruction should have waited 45 minutes before failing the terraform destroy run due to explicit timeout settings in the `aws_db_instance`.

### Actual Behavior
`aws_db_instance`destruction failed after a 20 minute timeout.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
[...]

PR #15374, opened 2017-06-22
helper/schema: fix issue reading Timeout data on delete operations 

catsby:
Timeout information is read from config and encoded into state. On delete, the config information isn't there, so we were failing to read it and then failing to respect the `Delete` timeout setting. This was original reported in https://github.com/hashicorp/terraform/issues/14444. This PR fixes that by reading the Timeout meta data from state, if it's not first found in the config. 

I've included `TestResourceApply_Timeout_destroy`, which checks the value of `TimeoutDelete` inside the delete method, and fails on `master`. 

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/767

------

Original fix part of https://github.com/hashicorp/terraform/pull/14396, but after our repo split, ~I couldn't get the patch to apply even after butchering it locally 😦~ I kind of fixed the attribution by adding `--author=` in there

This is just the helper/schema part of that pr – the ASG timeout addition I think I can get to apply in terraform-providers/terraform-provider-aws 

cc @jringuette for being super awesome and fixing this, I'm sorry I couldn't get the attribution right 🐼

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -142,6 +142,12 @@ func (r *Resource) Apply(
 		if err := rt.DiffDecode(d); err != nil {
 			log.Printf("[ERR] Error decoding ResourceTimeout: %s", err)
 		}
+	} else if s != nil {
+		if _, ok := s.Meta[TimeoutKey]; ok {
+			if err := rt.StateDecode(s); err != nil {
+				log.Printf("[ERR] Error decoding ResourceTimeout: %s", err)
+			}
+		}
 	} else {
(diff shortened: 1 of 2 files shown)


catsby:
Alternatively @jringuette if you want to cherry-pick my regression test and re-submit this with your name I'll accept it 😄

jringuette:
@catsby As long as it gets merged, I'm happy whoever the commiter is! Thank you

catsby:
@jringuette I managed to get your name in there! Keeping attribution is important to us, we truly value the contributions from the community

PR #15403, opened 2017-06-26
Added IntAtLeast and IntAtMost validation functions.

rileykarson:
Add IntAtLeast and IntAtMost validation functions so that we can validate that an integer is above or below some number without having to specify both sides of the range if there isn't a relevant upper or lower limit.


stack72:
LGTM thanks for this @rileykarson :)

PR #15465, opened 2017-07-04
provider/alicloud: Remove 'device_name' from alicloud examples

xiaozhu36:
The PR aims to remove 'device_name' from alicloud examples because of alicloud has decided to discard it from resource alicloud_disk_attachment. And disk's device would be specified automatically by system after attaching disks.


stack72:
This has a dependency on https://github.com/terraform-providers/terraform-provider-alicloud/pull/9

stack72:
@xiaozhu36 please can this be reopened against https://github.com/terraform-providers/terraform-provider-alicloud as all examples are moving there :)

Thanks

Paul

PR #15478, opened 2017-07-05
Update resource_provisioner.go

gibonurs:
This patch allow to correctly update chef-vaults when chef node needs to be recreated. If client with old key exists in chef-vault it will try to remove it first and then will add node with new key

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -561,16 +561,16 @@ func (p *provisioner) configureVaultsFunc(gemCmd string, knifeCmd string, confDi
 		for vault, items := range p.Vaults {
 			for _, item := range items {
 				
-+				removeCmd := fmt.Sprintf("%s vault remove %s %s -C %s -M client %s",
-+					knifeCmd,
-+					vault,
-+					item,
-+					p.NodeName,
-+					options,
-+				)
-+				if err := p.runCommand(o, comm, removeCmd); err != nil {
-+					return err
-+				}
+				removeCmd := fmt.Sprintf("%s vault remove %s %s -C %s -M client %s",
+					knifeCmd,
+					vault,
+					item,
+					p.NodeName,
+					options,
+				)
+				if err := p.runCommand(o, comm, removeCmd); err != nil {
+					return err
+				}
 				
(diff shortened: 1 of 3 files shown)


krzkowalczyk:
I would change it to bug, as with chef-vault 3.1.0 version existing client keys are not updated, so recreating a node in terraform does not refresh vaults. It's needed to remove client first.

jmccann:
@svanharmelen I know you helped me to get chef vault bootstrapping as part of chef provisioner.  Just ran into this issue myself.  Was wondering your thoughts on it.  Thanks!

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -560,6 +560,18 @@ func (p *provisioner) configureVaultsFunc(gemCmd string, knifeCmd string, confDi
 
 		for vault, items := range p.Vaults {
 			for _, item := range items {
+
```
svanharmelen: _nitting_ Could you delete this empty line?

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -560,6 +560,18 @@ func (p *provisioner) configureVaultsFunc(gemCmd string, knifeCmd string, confDi
...
 			for _, item := range items {
+				
+				removeCmd := fmt.Sprintf("%s vault remove %s %s -C %s -M client %s",
+					knifeCmd,
+					vault,
+					item,
+					p.NodeName,
+					options,
+				)
+				if err := p.runCommand(o, comm, removeCmd); err != nil {
```
svanharmelen: Are you sure this doesn't give an error when the client doesn't exist?

krzkowalczyk: I confirm, there is no error when client doesn't exist. At least in chef-vault <= 3.2.0.
IMHO chef-vault gem version should be pinned to ensure compatibility.

DukeXar:
Would be nice to provide an ability to specify version of the chef-client
actually.

svanharmelen:
@DukeXar you already can specify a Chef-Client version: https://www.terraform.io/docs/provisioners/chef.html#version-string-

mildwonkey:
Hello! I am going to close this older PR due to inactivity. Please do open a new PR if this is something you are still interested in working on. Thanks!

PR #15506, opened 2017-07-07
Automatically insert providers for import acceptance tests

jbardin:
Provider import tests previously didn't have to supply a config, but
terraform now requires the provider to be declared for discovery.

testProviderConfig returns a stub config with provider blocks based
on the TestCase Providers. This allows basic import tests in providers
to remain unchanged.

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -406,6 +406,10 @@ func Test(t TestT, c TestCase) {
 				step)
 		} else {
 			if step.ImportState {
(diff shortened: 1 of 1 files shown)


jbardin:
@apparentlymart: thanks for the heads up on TestMain here, I should have read more closely.

That is ... strange, but I don't see how it could break anything, so I'll leave it be for now. It would be nice in the long run to remove "testing" imports from the release binaries altogether.

PR #15528, opened 2017-07-11
Fix swallowed errors in command package.

alrs:
I found some errors in the command package that were being ignored.

### DIFF
--- a/command/workspace_new.go
+++ b/command/workspace_new.go
@@ -75,6 +75,9 @@ func (c *WorkspaceNewCommand) Run(args []string) int {
 	}
 
 	states, err := b.States()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


command/workspace_new.go
```diff
@@ -75,6 +75,9 @@ func (c *WorkspaceNewCommand) Run(args []string) int {
 	}
 
 	states, err := b.States()
+	if err != nil {
+		c.Ui.Error(fmt.Sprintf("Failed to get configured named states: %s", err))
```
jbardin: This should `return 1` here too, which we were probably just catching at the next command that errored.

alrs: I've added the return, thanks for the review.

PR #15559, opened 2017-07-14
allow missing x-terraform-protocol-version

jbardin:
If the release site is missing the "x-terraform-protocol-version"
header, we should fetch the latest spec'ed release. Downloading the
wrong protocol version can't do any damage, and the version present is
more than likely compatible.

Because the header information is kept out of band, there is a slightly higher chance of it going missing at some point, and provider availability is more important.


jbardin:
Yes, there's going to be a lot of details to iron out around a release where the protocol version is bumped, so I think this ends up being pretty minor at that point. As long as the header still exists at the time we upgrade, everything will work as expected, so we should probably add a check for this in an acceptance tests to ensure that it's maintained properly.

PR #15577, opened 2017-07-18
[WIP] B2 Cloud Storage backend

MattMcNam:
(WIP as I still need to update the documentation, but the code is ready to review!)

This PR adds a new backend using Backblaze's [B2 Cloud Storage](https://www.backblaze.com/b2/cloud-storage.html).

State environments are supported by appending a suffix `-env:ENV_NAME` to the state file path.
Locking is not supported at the moment. I've considered locking via a `.lock` file created alongside the state, but I need to spend more time thinking through how exactly that would work, and since I have no immediate need for it I figure it can be added later.

For acceptance testing, B2_ACCOUNT_ID and B2_APPLICATION_KEY environment variables are needed.

When deleting a state, this backend will 'hide' a file rather than fully delete it.
This is the B2 equivalent to 'deleting' a file in a versioned S3 bucket, it simply adds a delete marker. All buckets in B2 are versioned like this.


glorieux:
I'd really like to see that storage backend merged.
Is there any way I can help?

teamterraform:
Hi! I am going to close this old WIP PR - if you are still interested in this feature, please open a GitHub issue using the Feature Request template so we can discuss the design.

PR #15597, opened 2017-07-19
Documenting implicit behavior of string types to boolean and decimal.

thomasbiddle:
Updating documentation per #6254 .

**Reasoning for documentation update:**
Unclear documentation on how variables such as "0" or "true" would be handled when something such as `count` would expect a specific type (integer in this case)

**Relevant Terraform version:**
Current version

### DIFF
--- a/website/docs/configuration/variables.html.md
+++ b/website/docs/configuration/variables.html.md
@@ -57,7 +57,7 @@ These are the parameters that can be set:
 - `type` (optional) - If set this defines the type of the variable. Valid values
   are `string`, `list`, and `map`. If this field is omitted, the variable type
   will be inferred based on the `default`. If no `default` is provided, the type
-  is assumed to be `string`.
+  is assumed to be `string`. Terraform will make an implicit conversation of any values to `boolean` or `decimal` types if the consumer expects that.
 
 - `default` (optional) - This sets a default value for the variable. If no
   default is provided, the variable is considered required and Terraform will


website/docs/configuration/variables.html.md
```diff
@@ -57,7 +57,7 @@ These are the parameters that can be set:
 - `type` (optional) - If set this defines the type of the variable. Valid values
   are `string`, `list`, and `map`. If this field is omitted, the variable type
   will be inferred based on the `default`. If no `default` is provided, the type
-  is assumed to be `string`.
+  is assumed to be `string`. Terraform will make an implicit conversation of any values to `boolean` or `decimal` types if the consumer expects that.
```
tompave-roo: `s/conversation/conversion`

tompave-roo:
I think it would be helpful to have some more examples, because the distinction between legal declared types and legal literals is not clear.

For example, look [this example in the docs](https://www.terraform.io/docs/providers/aws/r/db_instance.html):

```tf
resource "aws_db_instance" "default" {
  allocated_storage    = 10
  storage_type         = "gp2"
  engine               = "mysql"
  engine_version       = "5.6.17"
  instance_class       = "db.t1.micro"
  name                 = "mydb"
  username             = "foo"
  password             = "bar"
  db_subnet_group_name = "my_database_subnet_group"
}
```

There `allocated_storage` is passed as an integer, but how should it be declared as an input variable? How should it be injected in the parent module? As a string?

nfagerlund:
For the ✨new, improved✨ 0.12 language, this is documented _much_ more explicitly: 

https://www.terraform.io/docs/configuration/expressions.html#type-conversion

I think that resolves the requests in this PR!

PR #15599, opened 2017-07-20
Fix swallowed tests in terraform package tests

alrs:
I found some errors in the tests for the terraform package that were being swallowed.


jbardin:
LGTM, Thanks for these @alrs!

PR #15615, opened 2017-07-21
[WIP] Create InputGraphBuilder

jbardin:
Don't handle module variables at all during the Input operation. Because Input happens before Refresh, module variable interpolation can fail when referencing values that aren't yet in the state, but are expected after Refresh. 

This creates a new `InputGraphBuilder`, which strips out all the unneeded pieces transformations inherited form the `PlanGraphBuilder`, most importantly the `ModuleVariableTransformer`.  This also removes the state entirely from the Input phase. Because Input happens before Refresh, but possibly depended on the state that will be refreshed, it is simpler to reason about if we don't allow the Input phase to depend on the state at all.

### DIFF
--- a/terraform/context.go
+++ b/terraform/context.go
@@ -258,11 +258,11 @@ func (c *Context) Graph(typ GraphType, opts *ContextGraphOpts) (*Graph, error) {
 	case GraphTypeInput:
 		return (&InputGraphBuilder{
 			Module:    c.module,
-			State:     c.state,
 			Providers: c.components.ResourceProviders(),
 			Targets:   c.targets,
 			Validate:  opts.Validate,
 		}).Build(RootModulePath)
+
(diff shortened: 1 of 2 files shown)


jbardin:
@apparentlymart: rewrote this to use a new graph builder, rather than layers more conditionals on the already overloaded Plan graph.

catsby:
Hey @jbardin is this still `WIP`?

jbardin:
@catsby, yes, this is still a WIP, thanks!

jbardin:
OK, dropping this in favor of my previous method of parameterizing the graph builder to use an interpolation eval node that doesn't cause fatal errors.  Because we need to call Input in the providers, and the providers may be parameterized in modules via variables, we simply have to process module variables in the Graph.

PR #15652, opened 2017-07-27
state commands with remote state backends

### LINKED ISSUE #10481: Docs: Provide some clarity with "terraform state mv" usage and remote states.
Hi there,

I started refactoring a bunch of terraform configs a couple days ago and found this to be a gigantic hole in [the documentation](https://www.terraform.io/docs/commands/state/mv.html) for the command.  I don't know if my situation was too special snowflake-y to be of use to others but thought I'd open an issue in case.

My specific case revolved around breaking up a single monolithic remote state into per-environment states.  In order to make this work I needed to use the -state-out flag to target the new state but...wait what's this?  

> Path to the state file. Defaults to "terraform.tfstate". Ignored when remote state is used.

Then what's the proper thing to do here?  I finally settled on disabling remote states for both files, performing the moves and then re-enabling and while it works it sure feels kludgy.  

Even something as simple as including another usage example in the docs for moving between two states that aren't local would have been super helpful and saved me some pain.  Or if remote states don't work with 'state mv' calling that out would be great.  

Hopefully this stuff isn't too specific for general docs, and also that I just didn't miss some really basic usage and shamed myself by opening an issue.

jbardin:
Some state commands worked with the backend configuration (push, pull, list, show), while others ignored it (mv, rm), leading to confusing behavior when trying to refactor state files. The website docs were also incorrectly updated, and did not match the behavior of the command, making the overall usage of the state commands confusing and possibly dangerous. 

This makes the `rm` and `mv` commands also work with a configured backend, reflecting the behavior of the other state commands, and the logical expectations of users. 

The one piece missing here is that there is no single command to move resources or modules from a local state file to a remote state, or between remote states. There are a number of ways we could introduce the behavior, a special identifier for `-state-out` like `@backend` for example, but complex refactoring probably warrants pulling the state into local files anyway, and not all possible options need to be supported. We can put off handing remote targets and workspaces for future refactoring.

Fixes #10481

### DIFF
--- a/command/state_mv.go
+++ b/command/state_mv.go
@@ -40,7 +40,7 @@ func (c *StateMvCommand) Run(args []string) int {
 	stateFrom, err := c.State()
 	if err != nil {
 		c.Ui.Error(fmt.Sprintf(errStateLoadingState, err))
-		return cli.RunResultHelp
+		return 1
 	}
 
 	if err := stateFrom.RefreshState(); err != nil {
@@ -64,7 +64,7 @@ func (c *StateMvCommand) Run(args []string) int {
 		stateTo, err = c.State()
 		if err != nil {
 			c.Ui.Error(fmt.Sprintf(errStateLoadingState, err))
-			return cli.RunResultHelp
+			return 1
 		}
 
 		if err := stateTo.RefreshState(); err != nil {
--- a/command/state_mv.go
+++ b/command/state_mv.go
@@ -183,28 +183,30 @@ func (c *StateMvCommand) addableResult(results []*terraform.StateFilterResult) i
 
 func (c *StateMvCommand) Help() string {
 	helpText := `
-Usage: terraform state mv [options] ADDRESS ADDRESS
+Usage: terraform state mv [options] SOURCE DESTINATION
 
-  Move an item in the state to another location or to a completely different
-  state file.
+ This command will move an item matched by the address given to the
+ destination address. This command can also move to a destination address
+ in a completely different state file.
 
-  This command is useful for module refactors (moving items into a module),
-  configuration refactors (moving items to a completely different or new
-  state file), or generally renaming of resources.
+ This can be used for simple resource renaming, moving items to and from
+ a module, moving entire modules, and more. And because this command can also
+ move data to a completely new state, it can also be used for refactoring
+ one configuration into multiple separately managed Terraform configurations.
 
-  This command creates a timestamped backup of the state on every invocation.
-  This can't be disabled. Due to the destructive nature of this command,
-  the backup is ensured by Terraform for safety reasons.
+ This command will output a backup copy of the state prior to saving any
+ changes. The backup cannot be disabled. Due to the destructive nature
+ of this command, backups are required.
 
-  If you're moving from one state file to a different state file, a backup
-  will be created for each state file.
+ If you're moving an item to a different state file, a backup will be created
+ for each state file.
 
 Options:
 
   -backup=PATH        Path where Terraform should write the backup for the original
                       state. This can't be disabled. If not set, Terraform
                       will write it to the same path as the statefile with
-                      a backup extension.
+                      a ".backup" extension.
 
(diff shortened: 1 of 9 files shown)


command/state_mv.go
```diff
@@ -215,13 +215,12 @@ Options:
...
-                      up Terraform-managed resources. By default it will
-                      use the state "terraform.tfstate" if it exists.
+  -state=PATH         Path to the source state file. Defaults to the configured
+                      backend, or "terraform.tfstate"
 
-  -state-out=PATH     Path to the destination state file to move the item
-                      to. This defaults to the same statefile. This will
-                      overwrite the destination state file.
+  -state-out=PATH     Path to the destination state file to write to. If this
+					  isn't specified, the source state file will be used. This
```
jbardin: oops, got some tabs in here

marcbachmann:
do you do a release soon? I just ran into the `state mv` issue.

dnk8n:
I am aiming to rename the resource in the remote(s3) state file once an instance has passed all tests. from app_deploying to app_deployed. This is so that next time we deploy there would not be a clash in the state files.

Next time we deploy we would first destroy the old 'app_deployed' before renaming the new app_deploying to app_deployed.

Hope the above makes sense and is reasonable. But when trying the command,

`terraform state mv aws_instance.app_deploying aws_instance.app_deployed`

I get:
```
Backend reinitialization required. Please run "terraform init".
Reason: Unsetting the previously set backend "s3"

The "backend" is the interface that Terraform uses to store state,
perform operations, etc. If this message is showing up, it means that the
Terraform configuration you're using is using a custom configuration for
the Terraform backend.

Changes to backend configurations require reinitialization. This allows
Terraform to setup the new configuration, copy existing state, etc. This is
only done during "terraform init". Please run that command now then try again.

If the change reason above is incorrect, please verify your configuration
hasn't changed and try again. At this point, no changes to your existing
configuration or state have been made.

Error loading the state: Initialization required. Please see the error message above.

Please ensure that your Terraform state exists and that you've
configured it properly. You can use the "-state" flag to point
Terraform at another state file.

```

Terraform version: v0.10.4

Is the above command expected to work?

jbardin:
@dnk8n,

I doubt the workflow you describe has anything to do with that failure, but it should work as expected.

 When you run the state command, terraform is seeing the config has changed for some reason, most likely your backend config is in a subdirectory which you are providing as an argument to `plan` and `apply`. If that's the case you need to duplicate the backend config in the current directory for the state command.

It's on our roadmap to try and reconcile how terraform can operate on paths in some cases, but retains it's local config state in the working directory.

dnk8n:
Thanks @jbardin 

I realise the issue. I have to cd in to the directory of the terraform configs in order to submit this command. I initialised with the path at the end on initialization (for some reason I have got into the habit of never using cd when writing automation if I can help it).

Appending the path like I did with init is not an option with terraform state mv though. So in this case I will need to 'cd'. Is this expected behaviour?

dnk8n:
Re-reading your comment after learning what I wrote about in my last comment makes it more clear now. I think you have already answered my questions. It is expected behaviour for now, but you plan to reconcile.

Thanks

djalexd:
@mukund1989 I am able to push the resource to a new state with terraform `0.11.7`. However, this is only true if destination is a new file. 

When destination exists - source is updated (the resource gets deleted - but I assume the next terraform plan/apply would generate it again), but destination is not! I'll experiment some more and come back with results.

djalexd:
So, it seems to work with some plumbing commands:

For example, if you want to move resource from state A to state B:
```
aws s3 cp <bucket>/a.terraform.tfstate . (download state)
aws s3 cp <bucket>/b.terraform.tfstate .
terraform state mv -state=a.terraform.tfstate -state-out=b.terraform.tfstate <resource> <resource>
aws s3 cp a.terraform.tfstate ... s3:// (upload back)
```
This works, but as you can see, it's pretty rudimentary. Perhaps there is another solution :)

PR #15663, opened 2017-07-28
memoize DiffFieldReader.ReadField

jbardin:
The field reader code path is extremely inefficient, but refactoring
it all is much to invasive a change at the moment.

Have DiffFieldReader internally cache results for ReadField.

This is a reduced version of #15641. The DiffFieldReader is the hot entry point, and localizing the caching to a particular instance seems much safer.


dshafik:
I can confirm that this does indeed solve the issue raised in #15641, thanks! 🥇

PR #15680, opened 2017-08-01
[WIP] etcd v3 backend with lock support.

brunomcustodio:
As a newcomer to Terraform, and having a few `etcd` clusters under my supervision, I thought it would be a nice idea to add an `etcdv3` backend with lock support. This is my first attempt at writing one, done mainly as an exercise and by replicating the existing `consul` backend and making whatever changes are required.

However, I need a little guidance in what concerns locking, because the lock APIs of `consul` and `etcd` are somewhat different and give different guarantees. Besides, I'm not really familiar with Consul.

Could anyone please be so kind as to have a look at the attached code and point out what can/must be changed in the `Lock` / `Unlock` methods? What is the expected "behaviour" of each method?


jbardin:
Hi @brunomcustodio,

Thanks for putting this together. 
I'll take a look as soon as I can, and we can go over the semantics of the Lock methods then.

One question I have to start, since I'm not up to date on the current state of etcd, can we "update" the existing etcd remote state or is there something fundamentally different about etcdv3?

brunomcustodio:
@jbardin thank you very much for taking the time :-) I'm looking forward to it.

Yes, there are some fundamental differences between `v2` and `v3` of `etcd`. For instance, the API's underlying protocol changed from HTTP to gRPC, and the internal structure of the data changed from a file-system-like structure to a flat binary key space. IMHO the existing `etcd` can be kept, but only for compatibility with existing code.

backend/remote-state/etcdv3/backend_state.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if err := lockUnlock(nil); err != nil {
+		return nil, err
+	}
+
+	return stateMgr, nil
+}
+
+func (b *Backend) determineKey(name string) string {
```
jbardin: If the changes in etcd means that there isn't going to be a direct upgrade from etcd2, you can probably get rid of the legacy default state path, and assume everything has a "workspace" and get rid of the conditionals around `keyEnvPrefix`and such.

backend/remote-state/etcdv3/client.go
```diff
@@ -0,0 +1,195 @@
...
+	etcdv3sync "github.com/coreos/etcd/clientv3/concurrency"
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/terraform/state"
+	"github.com/hashicorp/terraform/state/remote"
+)
+
+const (
+	lockAcquireTimeout = 2 * time.Second
+	lockInfoSuffix     = ".lockinfo"
+	lockSuffix         = ".lock"
```
jbardin: `lockSuffix` was strictly a consul implementation detail, so that the terraform lock works just like the consul cli locks. This may be different for etcd, or maybe it doesn't matter at all.

backend/remote-state/etcdv3/client.go
```diff
@@ -0,0 +1,195 @@
...
+func (c *RemoteClient) putLockInfo(info *state.LockInfo) error {
+	c.info.Path = c.Key
+	c.info.Created = time.Now().UTC()
+
+	_, err := c.Client.KV.Put(context.TODO(), c.Key+lockInfoSuffix, string(c.info.Marshal()))
+	return err
+}
+
+func (c *RemoteClient) lock() (string, error) {
+	session, err := etcdv3sync.NewSession(c.Client)
```
jbardin: I'm not sure what etcd's behavior would be here, but terraform should never recursively lock, so you can check for an existing session and return an error if there is already a lock outstanding.

backend/remote-state/etcdv3/client.go
```diff
@@ -0,0 +1,195 @@
...
+	if err1 := mutex.Lock(ctx); err1 != nil {
+		lockInfo, err2 := c.getLockInfo()
+		if err2 != nil {
+			return "", &state.LockError{Err: err2}
+		}
+		return "", &state.LockError{Info: lockInfo, Err: err1}
+	}
+
+	c.etcdMutex = mutex
+	c.etcdSession = session
```
jbardin: Again I'm not sure about etcd, but consul sessions favored liveness, and were easier to lose than desired and susceptible to network timeouts. I'm not sure if we need to do the same here, but the consul backend watches the session status and reconnects when necessary.

backend/remote-state/etcdv3/client.go
```diff
@@ -0,0 +1,195 @@
...
+
+	var errs error
+
+	if err := c.etcdMutex.Unlock(context.TODO()); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if err := c.etcdSession.Close(); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+
```
jbardin: I think we may want to remove the lockInfo too.

jbardin:
Hi @brunomcustodio,

Thais looks very good after a preliminary pass through the code. Did you have any specific questions about the locking semantics?

backend/remote-state/etcdv3/backend.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/hashicorp/terraform/backend"
+	"github.com/hashicorp/terraform/helper/schema"
+	"strings"
+)
+
+func New() backend.Backend {
+	s := &schema.Backend{
+		Schema: map[string]*schema.Schema{
+			"endpoints": &schema.Schema{
+				Type:        schema.TypeString,
```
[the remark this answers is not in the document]
brunomcustodio: Right, many thanks :+1:

backend/remote-state/etcdv3/backend_state.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]string, 1, len(res.Kvs)+1)
+	result[0] = backend.DefaultStateName
+	for _, kv := range res.Kvs {
+		result = append(result, strings.TrimPrefix(string(kv.Key), prefix))
+	}
+
```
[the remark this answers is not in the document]
brunomcustodio: Sure :+1: b89634823

backend/remote-state/etcdv3/backend_state.go
brunomcustodio: > (...) you can probably get rid of the legacy default state path (...) and get rid of the conditionals around `keyEnvPrefix` and such.

@jbardin I'm not sure I understand what you mean. 😶 Could you please clarify/exemplify?

backend/remote-state/etcdv3/client.go
brunomcustodio: Yes, of course 🙂 b8f4f6d

backend/remote-state/etcdv3/client.go
brunomcustodio: Done 👍  9aa417e

backend/remote-state/etcdv3/backend_state.go
jbardin: I meant that the reason the other backends have the "default" state handled separately is solely for backwards compatibility with state files that existing before envs/workspaces.

So rather than dealing with the `keyEnvPrefix` you've adopted from the others, you could use the same hierarchy in all cases of `prefix/name`, and just include "default" in there.

backend/remote-state/etcdv3/client.go
brunomcustodio: I believe that's not the case with`etcd`, judging by a number of factors including this

https://github.com/coreos/etcd/blob/master/clientv3/concurrency/session.go#L26
https://github.com/coreos/etcd/blob/master/clientv3/concurrency/session.go#L64

and some tests I ran locally using `etcdctl`:

```
$ ETCDCTL_API=3 etcdctl --endpoints=127.0.0.1:12379 lock my-terraform-lock -- watch date
```

This will sit on a loop printing the date every two seconds. If you start a second command against a different `etcd` member, you'll obviously be left waiting to acquire the lock:

```
$ ETCDCTL_API=3 etcdctl --endpoints=127.0.0.1:22379 lock my-terraform-lock -- watch date
```

Now, if you take down `127.0.0.1:12379` (the first member), the second command will still be left waiting to acquire the lock. If you take `127.0.0.1:12379` up again the first process keeps  printing the date. Only when the first process is stopped the second one takes over the lock and starts printing the date. Here's the code for the `lock` CLI command, which of course uses the same API as this implementation:

https://github.com/coreos/etcd/blob/master/etcdctl/ctlv3/command/lock_command.go

In the end I'd say we don't need to keep our own goroutine to handle keepalive/reconnect. WDYT @jbardin?

jbardin: I agree. Let's favor the simpler solution for now under the assumption that the etc client package takes care of all the details.

backend/remote-state/etcdv3/client.go
brunomcustodio: Thanks 🙂 Removed in f60075f.

backend/remote-state/etcdv3/backend_state.go
brunomcustodio: Done @ 3c21b9c. Thank you for your guidance. 🙂

brunomcustodio:
@jbardin I believe this to be at a good point for another review. I've addressed your suggestions and in the meantime added TLS support (for secure client connections) and documentation.

I've added a comment to `.travis.yml` about testing, though I don't really know if that's the best place. I think the best way for you to run the `etcdv3` tests on your machine is to spin up a local etcd 3.x instance and run the test suite using

```
TF_ETCDV3_ENDPOINTS="localhost:2379" TF_ETCDV3_TEST=1 make test
```

I see the `consul` tests are disabled in Travis. Do we want the `etcdv3` ones to run? If so we will need to somehow spin up a local `etcd` instance (insecure) or host an external one (secure). PTAL and let me know what your thoughts are.

brunomcustodio:
@jbardin have you had the time to look into this? Do you consider merging it in the short-term?

pires:
I've ran this manually and it seemed to work smoothly. Great job, people!

brunomcustodio:
( Sorry for the closing/reopening caused by a slight mistake on my part. )

jbardin:
Hi @brunomcustodio,

Thanks for hanging in there while we've been so busy!

This is looking great! Since backend tests still need to be run manually, we usually paste the test output into the PR comments, to record the acceptance tests as passing. 
Since you have etcd already setup and running, could you run the full set of tests for me? There's still some extra log output in these tests, so something like `TFF_ACC=1 go test -v 2>/dev/null` usually gives a concise list of the tests that were run.

brunomcustodio:
@jbardin once again thank you very much for taking the time to review this. Here's the concise version:

```text
=== RUN   TestBackend_impl
=== RUN   TestBackend
--- PASS: TestBackend (2.16s)
	backend_test.go:39: Cleaned up 0 keys.
	backend_test.go:69: TestBackend: testing state locking for *etcd.Backend
	backend_test.go:39: Cleaned up 2 keys.
=== RUN   TestBackend_lockDisabled
--- PASS: TestBackend_lockDisabled (0.06s)
	backend_test.go:39: Cleaned up 0 keys.
	backend_test.go:92: TestBackend: testing state locking for *etcd.Backend
	backend_test.go:92: TestBackend: *etcd.Backend: empty string returned for lock, assuming disabled
	backend_test.go:39: Cleaned up 3 keys.
=== RUN   TestRemoteClient
--- PASS: TestRemoteClient (0.04s)
	backend_test.go:39: Cleaned up 0 keys.
	backend_test.go:39: Cleaned up 0 keys.
=== RUN   TestEtcdv3_stateLock
--- PASS: TestEtcdv3_stateLock (2.09s)
	backend_test.go:39: Cleaned up 0 keys.
	backend_test.go:39: Cleaned up 1 keys.
=== RUN   TestEtcdv3_destroyLock
--- PASS: TestEtcdv3_destroyLock (0.05s)
	backend_test.go:39: Cleaned up 0 keys.
	backend_test.go:39: Cleaned up 1 keys.
PASS
ok  	github.com/hashicorp/terraform/backend/remote-state/etcdv3	4.422s
```

I pasted the full version (a second run) in [this Gist](https://gist.github.com/brunomcustodio/091cdb99b12368a55144ef83d2f923ec) in case you want to take a look.

jbardin:
Thanks for all the work @brunomcustodio!

Let's drop this into master so people can start trying it out!

brunomcustodio:
Most welcome @jbardin 🙂 Thanks for taking the time!

PR #15693, opened 2017-08-02
add init section to getting-started/install

jbardin:


jbardin:
It turns out this is already in the build page -- was comparing to the released branch.

PR #15695, opened 2017-08-02
config: Make 'id' a reserved field name

radeksimko:
This is to prevent mistakes like https://github.com/hashicorp/terraform/pull/9366 and generally to improve dev experience.

I could add a test, but both slices are already tested and it seems unnecessary to test just a one new item in existing slice.

### DIFF
--- a/config/loader_hcl.go
+++ b/config/loader_hcl.go
@@ -21,6 +21,7 @@ var ReservedResourceFields = []string{
 	"connection",
 	"count",
 	"depends_on",
+	"id",
 	"lifecycle",
 	"provider",
 	"provisioner",
@@ -28,6 +29,7 @@ var ReservedResourceFields = []string{
 
 var ReservedProviderFields = []string{
 	"alias",
+	"id",
 	"version",
 }
 


jbardin:
This LGTM as long as we're sure that it's not breaking any provider mistakingly using id. Because this is handled in the config loader, I don't think it would be caught by the ACC tests themselves.

radeksimko:
>  it's not breaking any provider mistakingly using id.

if there's a resource with such field then that field is useless in the schema anyways AFAIK.

> I don't think it would be caught by the ACC tests themselves

it will emerge in tests once we start bumping vendored `hashicorp/terraform` in all providers, because each provider test calls `InternalValidate()` where this will be reported. Also, we're intentionally bypassing this check for `Deprecated` & `Removed` fields, so that things can disappear gracefully - but frankly in case of `id` (or most other reserved fields) there's really no need to deprecate it, because they get removed before the CRUD even gets chance to read them.

jbardin:
thanks for confirming @radeksimko ;) 
:+1:

PR #15698, opened 2017-08-02
build: Update travis and Vagrantfile go versions

cblecker:


jbardin:
thanks for the update @cblecker!

PR #15769, opened 2017-08-09
load legacy plugin paths

### LINKED ISSUE #15705: Terraform v0.10.0 3rd party plugin location incorrectly stated
### Terraform Version
v0.10.0

NA

### Debug Output
NA

### Panic Output
NA

### Expected Behavior
The previous model for specifying provider location should work (.terraformrc). 
The correct path for my system would be outputted.
Sufficient migration documentation would be available to figure this out without [searching through code](https://github.com/hashicorp/terraform/blob/master/plugins.go#L16) 

### Actual Behavior
Incorrect path was outputted for where to locate plugin

### Steps to Reproduce
1. Update terraform to v.0.10.0 without changing legacy plugin organization
2. run `terraform init`
3. Observe inaccurate message on where 3rd party plugins should live:
```
In the latter case, the plugin must be installed manually by locating and
downloading a suitable distribution package and placing the plugin's executable
file in the following directory:
    terraform.d/plugins/darwin_amd64
```
Actual path that works is ~/.terraform.d/plugins/<provider here>

### Important Factoids
.terraformrc config approach was nice for development because it could point to an arbitrary location like the provider projects /bin folder for development work. 

### References
NA

jbardin:
Add the OS_ARCH directory to `~/.terraform.d/plugins` because Discovery  no longer searches for subdirectories.

Add the override paths from `Meta.PluginOverrides` for discovery.

Fixes #15705

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -21,6 +21,7 @@ func globalPluginDirs() []string {
 		log.Printf("[ERROR] Error finding global config directory: %s", err)
 	} else {
 		machineDir := fmt.Sprintf("%s_%s", runtime.GOOS, runtime.GOARCH)
+		ret = append(ret, filepath.Join(dir, "plugins"))
 		ret = append(ret, filepath.Join(dir, "plugins", machineDir))
 	}
 
--- a/plugins.go
+++ b/plugins.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"path/filepath"
+	"runtime"
 )
 
 // globalPluginDirs returns directories that should be searched for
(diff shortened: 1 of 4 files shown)


tehmaspc:
Will this create a plugins/OS_ARCH dir if it doesn't exist? I know TF creates some dirs during initial use - if it can also just stub out a plugins dir structure to make it more obvious to the user where plugins should go - that would be great!

thanks!

plugins.go
```diff
@@ -18,7 +20,8 @@ func globalPluginDirs() []string {
 	if err != nil {
 		log.Printf("[ERROR] Error finding global config directory: %s", err)
 	} else {
-		ret = append(ret, filepath.Join(dir, "plugins"))
+		machineDir := fmt.Sprintf("%s_%s", runtime.GOOS, runtime.GOARCH)
+		ret = append(ret, filepath.Join(dir, "plugins", machineDir))
```
jbardin: yeah, I can see keeping the `~/.terraform.d/plugins` directory just to reduce churn in a point release. 

As for for machineDir, it seems there's a const in the command package that we're not using, and everywhere else is generating longer paths I can clean that up in another PR.

jbardin:
Re-added the `~/.terraform.d/plugins/` dir to the discovery paths.

@tehmaspc, I think creating the directories might be a good idea. We can review that in another PR. Thanks!

PR #15816, opened 2017-08-15
specify a logger for go-plugin

jbardin:
The go-plugin package now uses hclog. The default Logger has a level set
to Info, but all plugin output is relayed via Debug. Create a new named
logger for plugins with the level set to Trace so that all output comes
through.

### DIFF
--- a/plugin/client.go
+++ b/plugin/client.go
@@ -1,8 +1,10 @@
 package plugin
 
 import (
+	"os"
 	"os/exec"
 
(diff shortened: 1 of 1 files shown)


jbardin:
Yes, we set `Trace`because Terraform does its own filtering later and we want everything from the plugin to get logged. Once we start using hclog throughout terraform, we can simply hook this up to our global logger. Once plugins start using hclog, the go-plugin Client will then filter the structured log output and write it at the correct log level!

PR #15819, opened 2017-08-16
helper/schema: Add Set.HashEqual

vancluever:
Equality of `schema.Sets` gets tricky when dealing with nested sets -
`Set.Equal` only superficially compares the underlying maps and hence any
sets nested under the root sets cause issues.

This adds a simple method, `HashEqual`, that does a top-level hash
comparison, helping to work around this without any complex re-invention
of things like `reflect.DeepEqual`.

Of course, in order to make effective use of this function, the user
needs to make sure they are properly hashing their nested sets, however
this is trivial with things like `HashResource`.

### DIFF
--- a/helper/schema/set.go
+++ b/helper/schema/set.go
@@ -153,6 +153,31 @@ func (s *Set) Equal(raw interface{}) bool {
 	return reflect.DeepEqual(s.m, other.m)
 }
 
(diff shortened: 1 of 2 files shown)


vancluever:
Thanks for the review @apparentlymart! :+1:

PR #15832, opened 2017-08-16
Add space after bracket

sethvargo:
I understand the desire to have things aligned vertically, but adding the space inside the bracket makes grepping challenging. I moved the space outside the bracket, which will give the same behavior, but also look less weird (in my humble opinion).


sethvargo:
Oops. This is a vendor.

PR #15872, opened 2017-08-22
add changes to generated code

darkowlzz:
Building on go1.8.3 darwin/amd64 results in these changes.


darkowlzz:
@catsby thanks a lot. So I had an older version of stringer 😅
Closing this.

PR #15873, opened 2017-08-22
Provide example for terraform taint documentation

jwinter:
Provides an example (with similar formatting as
https://www.terraform.io/docs/commands/state/list.html#example-filtering-by-module)
for tainting resources within a module.


jwinter:
@catsby , I added those additional docs. Thx for the quick review.

Issue #15875, opened 2017-08-22
Add switch to disable import config check


jzbruno:
### Terraform Version
>= 0.10.0

### Actual Behavior

Hello,

We have built up a bunch of tools that generate terraform config for our existing environment. They do this by first getting a list of resource IDs and then running the import command. The config is then generated from the state file.

The latest update to disallow import without matching config will require us to re-write these tools.

Can we add a command line switch to disable the import config check, please?

### Steps to Reproduce

Use `terraform import` with a resource that does not have a config block defined.

pkolyvas:
PR was merged. Closing. Thanks @jzbruno for the contribution.

PR #15876, opened 2017-08-22
Add ignore-missing-config flag to import command

jzbruno:
This addresses #15875

Some custom tools may depend on the old behavoir of the import command that does not require matching config. To allow backward compatability, the -ignore-missing-config flag has been added.


apparentlymart:
Hi @jzbruno! Thanks for working on this.

This seems reasonable to me. Nice work!

My only minor feedback is that I think the option should be called `-allow-missing-config`, since I think that more clearly communicates what this option does: it permits something that would usually be forbidden.

jzbruno:
@apparentlymart Sounds good. I will get that updated.

apparentlymart:
Sorry for the delay in getting back to you here, @jzbruno!

This looks good to me. Thanks for the docs, tests etc too!

jzbruno:
Great! Thanks!

Jack

PR #15890, opened 2017-08-23
Resource aws_instance does not have attr private_ip_address

fgbreel:
Hello,

This is just a small fix in docs examples where the attribute used doesn't match the resource attributes.

It generates messages like this: 

``` 
Resource 'aws_instance.instances.0' does not have attribute 'private_ip_address' for variable 'aws_instance.server.0.private_ip_address'
```

- Reasoning for docs update: `consistency`
- Relevant Terraform version: `0.8.1`

Best,
Gabriel


catsby:
Thanks!

PR #15905, opened 2017-08-24
deps: Remove xmlpath from dependencies

grubernaut:


grubernaut:
It used to be a dependency of `winrm`, `winrmtest`, and `winrmcp`, but those were removed in #15862, just forgot to remove the explicit dependency we had included for `xmlpath`.

PR #15936, opened 2017-08-27
helper/validation: Add ValidateListUniqueStrings

vancluever:
Add the `ValidateListUniqueStrings` function, which is a `ValidateFunc` that
ensures a list has no duplicate items in it. It's useful for when a list
is needed over a set because order matters, yet the items still need to
be unique.


vancluever:
:tada:

PR #15945, opened 2017-08-29
website: new provider dev program content

mitchellh:
This brings back the provider dev program content with a copy change.
Reviewed by Burzin already, requesting review from Clint.

### DIFF
--- a/website/guides/terraform-provider-development-program.html.md
+++ b/website/guides/terraform-provider-development-program.html.md
@@ -13,8 +13,8 @@ listed on the official Terraform website. The program is intended to be largely
 self-serve, with links to information sources, clearly defined steps, and
 checkpoints.
 
--> **Building your own provider?** If you're building your own
-provider and aren't interested in having HashiCorp officially maintain
+-> **Building your own provider?** If you're building your own provider and
+aren't interested in having HashiCorp officially approve and regularly test
 the provider, refer to the
 [Writing Custom Providers guide](/guides/writing-custom-terraform-providers.html).
 
@@ -36,7 +36,7 @@ and tests a particular provider.
 
 ![Provider Engagement Table](/assets/images/docs/engage-table.png)
 
-This document is primarily intended for the "HashiCorp/Vendors" row in
+-> **Note:** This document is primarily intended for the "HashiCorp/Vendors" row in
 the table above. Community contributors who’re interested in contributing to
 existing providers or building new providers should refer to the
 [Writing Custom Providers guide](/guides/writing-custom-terraform-providers.html).
@@ -201,7 +201,7 @@ Below is an ordered checklist of steps that should be followed during the
 provider development process. This just reiterates the steps already documented
 in the section above.
 
-* Fill out provider development program engagement [webform](https://goo.gl/forms/iqfz6H9UK91X9LQp2) (https://goo.gl/forms/iqfz6H9UK91X9LQp2)
+* Fill out provider development program engagement [webform](https://goo.gl/forms/iqfz6H9UK91X9LQp2)
 
 * Refer to the example providers and model the new provider based on that
 
--- a/website/guides/terraform-provider-development-program.html.md
+++ b/website/guides/terraform-provider-development-program.html.md
@@ -1,0 +1,226 @@
+---
+layout: "guides"
+page_title: "Terraform Provider Development Program"
+sidebar_current: "guides-terraform-provider-development-program"
+description: This guide is intended for vendors who're interested in having their platform supported by Teraform. The guide walks vendors through the steps involved in creating a provider and applying for it to be included with Terraform.
+---
+
+# Terraform Provider Development Program
+
+The Terraform Provider Development Program allows vendors to build
+Terraform providers that are officially approved and tested by HashiCorp and
+listed on the official Terraform website. The program is intended to be largely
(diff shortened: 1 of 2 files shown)


website/guides/terraform-provider-development-program.html.md
```diff
@@ -0,0 +1,226 @@
...
+
+The Terraform Provider Development Program allows vendors to build
+Terraform providers that are officially approved and tested by HashiCorp and
+listed on the official Terraform website. The program is intended to be largely
+self-serve, with links to information sources, clearly defined steps, and
+checkpoints.
+
+-> **Building your own provider?** If you're building your own
+provider and aren't interested in having HashiCorp officially maintain
+the provider, refer to the
```
catsby: Can we change this

    in having HashiCorp officially maintain the provider 

to 

    in having HashiCorp officially approve and regularly test the provider

or similar? We're implying that this path leads to us maintaining it.

It was my understanding that we're still looking for the "Vendor" behind this new Provider to do the maintenance here, right? Below, we detail a `Support` section about our expectations of support, but this line here suggests HashiCorp will (or will be involved) in the direct support, which I don't think was the intent

catsby: May also want to mention the "Self host" part of building your own provider. The idea being that `github.com/terraform-providers` is only official, approved, and tested providers.

website/guides/terraform-provider-development-program.html.md
```diff
@@ -0,0 +1,226 @@
...
+these as resources users can code to. Terraform presently supports more than
+70 providers, a number that has more than doubled in the past 12 months.
+
+All providers integrate into and operate with Terraform exactly the same way.
+The table below is intended to help users understand who develops, maintains
+and tests a particular provider.
+
+![Provider Engagement Table](/assets/images/docs/engage-table.png)
+
+This document is primarily intended for the "HashiCorp/Vendors" row in
```
catsby: Not a blocker, but can we add some kind of emphasis here? Either make it a `-> Note` or some bolding? I believe a big part of the confusion earlier was not being clear about the audience. We're clear here, but I imagine without any call out here it may be glossed over.

website/guides/terraform-provider-development-program.html.md
```diff
@@ -0,0 +1,226 @@
...
+existing providers or building new providers should refer to the
+[Writing Custom Providers guide](/guides/writing-custom-terraform-providers.html).
+
+## Provider Development Process
+
+The provider development process is divided into six steps below. By following
+these steps, providers can be developed alongside HashiCorp to ensure new
+providers are able to be published in Terraform as quickly as possible.
+
+![Provider Development Process](/assets/images/docs/process.png)
```
catsby: This image contains a "Recruit" starting point, but that isn't mentioned in the guide

website/guides/terraform-provider-development-program.html.md
```diff
@@ -0,0 +1,226 @@
...
+### 6. Support
+
+Many vendors view the release step to be the end of the journey, while at
+HashiCorp we view it to be the start. Getting the provider built is just the
+first step in enabling users to use it against the infrastructure. Once this is
+done on-going effort is required to maintain the provider and address any
+issues in a timely manner.
+
+The expectation is to resolve all critical issues within 48 hours and all other
+issues within 5 business days. HashiCorp Terraform has as extremely wide
```
catsby: I like that we set expectations but 48 hours and 5 business days seems really hard to place on Vendors. Not a blocker though.

website/guides/terraform-provider-development-program.html.md
```diff
@@ -0,0 +1,226 @@
...
+Vendors who choose to not support their provider and prefer to make it a
+community supported provider will not be listed on terraform.io.
+
+## Checklist
+
+Below is an ordered checklist of steps that should be followed during the
+provider development process. This just reiterates the steps already documented
+in the section above.
+
+* Fill out provider development program engagement [webform](https://goo.gl/forms/iqfz6H9UK91X9LQp2) (https://goo.gl/forms/iqfz6H9UK91X9LQp2)
```
catsby: Double (htmllink) here, in case that wasn't intended

website/guides/terraform-provider-development-program.html.md
mitchellh: Done. I think the self host part is a good point in should go in the "writing your own provider" part. I'll update that

website/guides/terraform-provider-development-program.html.md
mitchellh: Good idea. done.

mitchellh: I'll let Burzin look at that. I think this doesn't have to block this.

website/guides/terraform-provider-development-program.html.md
mitchellh: Agreed, its easier to loosen this over time but for our initial set of official providers we are requesting this. 👍

website/guides/terraform-provider-development-program.html.md
catsby: Agreed, not a blocker

website/guides/terraform-provider-development-program.html.md
catsby: 👍

PR #16021, opened 2017-09-05
helper/schema: Disallow ForceNew for computed-only field

radeksimko:
This is to prevent confusions like here https://github.com/terraform-providers/terraform-provider-packet/pull/18#discussion_r136974122


radeksimko:
Interesting, so I must have been wrong in my conclusions then... 🤔 
Data source and a few resource example schemas use this... closing then.

PR #16023, opened 2017-09-05
update go-getter

jbardin:
Update to the latest go-getter to support unpacking github release
tarballs as modules.


jbardin:
This will also fix #16023

PR #16042, opened 2017-09-07
Documentation update to explain how locals can be accessed. 

mechastorm:
There was a lack of example of how locals can be accessed.

Also included an example of how locals can be used in a map.


mechastorm:
I will resubmit my PR once #16041 is fully merged with just an example of using `locals` with map

justincampbell:
@mechastorm merged!

Issue #15707, opened 2017-08-03
provisioner/chef: find -maxdepth not supported on AIX


cattywampus:
When I'm using terraform to build virtual machines running AIX, I get an error during the chef provisioning step. Details of the error are in the report below, but it looks like the linux chef provisioner has a step where it uses `find -maxdepth 1` to find all of the config files in `/etc/chef` and change the permission bits.

The problem is the `-maxdepth` option is not supported in the version of the `find` command on AIX. I ran into this on AIX 7.1, and from my basic research this would apply to AIX 6.1 too. 

I think the fix for this would be to alter the find command to return the same list of files from that directory (and no deeper) without using `-maxdepth`. I found a Stackoverflow answer that provides an [alternative technique](https://unix.stackexchange.com/questions/372994/equivalent-maxdepth-for-find-in-aix) to using `find` without the `-maxdepth` option. I modified this solution to match what the linux chef provisioner does and tested this command in AIX 7.1 and RedHat 7.1 with success. 

If this is a bug you would be willing to accept, I'd be happy to submit a pull request for it. 

### Terraform Version
Terraform v0.9.2

I understand this isn't the latest version but `-maxdepth` flag is still being used in [master](https://github.com/hashicorp/terraform/blob/master/builtin/provisioners/chef/linux_provisioner.go#L13) and [v0.10.0](https://github.com/hashicorp/terraform/blob/v0.10.0/builtin/provisioners/chef/linux_provisioner.go#L13).

### Terraform Configuration Files
I'm currently waiting on approval to post a sanitized version of the terraform plan, but until then I can say that I have a `openstack_compute_instance_v2` resource with two provisioners:
  * `remote-exec` - downloads the chef `install.sh` script from an internal server and runs it to install the chef-client. This definitely finishes successfully. The chef client is installed. 
  * `chef` - performs the steps to actually register the node with the chef server, set the run list and environment. `skip_install` is set to `true` because chef-client is installed by the previous `remote-exec` provisioner.

### Debug Output
I can't provide a complete debug output yet due to pending approval through our security process. Once I get the 👍 then I'll add the link to a Gist for the relevant section of the debug log output that fails during the chef provisioner process. 

In the interim, I'm hoping the error below in the [Actual Behavior](#actual-behavior) section is enough to kick start this discussion since the error is pretty cut and dry. 

### Panic Output
Not applicable

### Expected Behavior
[...]

PR #16091, opened 2017-09-13
provisioners/chef: Update linux provisioner to support AIX

cattywampus:
Resolves Issue #15707 which prevents the find command from running on vm resources that have AIX as it's operating system. The `find` command on this operating system does not support the `-maxdepth` flag so I modified the command used in the linux provisioner using `-path` and `-prune` which yields the same result: returning a list of only the files located in the top level of the etc chef configuration directory. 

Once that was resolved a new issue was discovered using this provisioner on AIX when setting the ownership of the linux config and ohai hints directories. In AIX, the `root` user does not exist in a `root` group. Instead the group name is called `system`. Now the group used for owning the chef files is set based on the `os_type`. 

Both fixes were tested in our on premises data center using the openstack provider with the AIX 7.1 and RedHat 7.3 operating systems. 

I would **love** to get any feedback I can on this to make it better. Not only is this my first contribution to terraform, this is my first exposure to Go, so if there are things that can be improved, please tell me.

### DIFF
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -10,9 +10,26 @@ import (
 )
 
 const (
-	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	chmodLinux = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	chmodAix   = "find %[1]s ! -path %[1]s -prune -type f -exec /bin/chmod %[2]d {} +"
 	installURL = "https://omnitruck.chef.io/install.sh"
 )
+
+func (p *provisioner) getChmodCommand() string {
+	if p.OSType == "aix" {
+		return chmodAix
+	} else {
+		return chmodLinux
+	}
+}
+
+func (p *provisioner) getRootGroup() string {
+	if p.OSType == "aix" {
+		return "system"
+	} else {
+		return "root"
+	}
+}
 
 func (p *provisioner) linuxInstallChefClient(o terraform.UIOutput, comm communicator.Communicator) error {
 	// Build up the command prefix
@@ -54,7 +71,7 @@ func (p *provisioner) linuxCreateConfigFiles(o terraform.UIOutput, comm communic
 		if err := p.runCommand(o, comm, "chmod 777 "+linuxConfDir); err != nil {
 			return err
 		}
-		if err := p.runCommand(o, comm, fmt.Sprintf(chmod, linuxConfDir, 666)); err != nil {
+		if err := p.runCommand(o, comm, fmt.Sprintf(p.getChmodCommand(), linuxConfDir, 666)); err != nil {
 			return err
 		}
 	}
@@ -75,7 +92,7 @@ func (p *provisioner) linuxCreateConfigFiles(o terraform.UIOutput, comm communic
 			if err := p.runCommand(o, comm, "chmod 777 "+hintsDir); err != nil {
 				return err
 			}
-			if err := p.runCommand(o, comm, fmt.Sprintf(chmod, hintsDir, 666)); err != nil {
+			if err := p.runCommand(o, comm, fmt.Sprintf(p.getChmodCommand(), hintsDir, 666)); err != nil {
 				return err
 			}
 		}
@@ -89,10 +106,10 @@ func (p *provisioner) linuxCreateConfigFiles(o terraform.UIOutput, comm communic
 			if err := p.runCommand(o, comm, "chmod 755 "+hintsDir); err != nil {
 				return err
 			}
-			if err := p.runCommand(o, comm, fmt.Sprintf(chmod, hintsDir, 600)); err != nil {
+			if err := p.runCommand(o, comm, fmt.Sprintf(p.getChmodCommand(), hintsDir, 600)); err != nil {
 				return err
 			}
-			if err := p.runCommand(o, comm, "chown -R root:root "+hintsDir); err != nil {
+			if err := p.runCommand(o, comm, "chown -R root:"+p.getRootGroup()+" "+hintsDir); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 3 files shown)


builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -10,10 +10,27 @@ import (
 )
 
 const (
-	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	chmodLinux = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	chmodAix   = "find %[1]s ! -path %[1]s -prune -type f -exec /bin/chmod %[2]d {} +"
```
svanharmelen: Can you explain the difference between `find %[1]s ! -path %[1]s` and `find %s`? As I'm not familiar with the form. And is there a specific reason why you added the `-prune` option and omitted the `-maxdepth 1` option?

builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -10,10 +10,27 @@ import (
...
+		return chmodLinux
+	}
+}
+
+func (p *provisioner) getRootGroup() string {
+	if p.OSType == "aix" {
+		return "system"
+	} else {
+		return "root"
+	}
```
svanharmelen: My linter gives a warning about this as the else is not needed here:

```
	if p.OSType == "aix" {
		return "system"
	}
	return "root"
```

builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -10,10 +10,27 @@ import (
 )
 
 const (
-	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	chmodLinux = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	chmodAix   = "find %[1]s ! -path %[1]s -prune -type f -exec /bin/chmod %[2]d {} +"
 	installURL = "https://omnitruck.chef.io/install.sh"
 )
 
+func (p *provisioner) getChmodCommand() string {
```
svanharmelen: In Go getters are normally not prefixed by `set`:

```suggestion
func (p *provisioner) chmodCommand() string {
```

builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -10,10 +10,27 @@ import (
...
+	if p.OSType == "aix" {
+		return chmodAix
+	} else {
+		return chmodLinux
+	}
+}
+
+func (p *provisioner) getRootGroup() string {
```
svanharmelen: In Go getters are normally not prefixed by `set`:

```suggestion
func (p *provisioner) rootGroup() string {
```

builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -89,10 +106,10 @@ func (p *provisioner) linuxCreateConfigFiles(o terraform.UIOutput, comm communic
 			if err := p.runCommand(o, comm, "chmod 755 "+hintsDir); err != nil {
 				return err
 			}
-			if err := p.runCommand(o, comm, fmt.Sprintf(chmod, hintsDir, 600)); err != nil {
+			if err := p.runCommand(o, comm, fmt.Sprintf(p.getChmodCommand(), hintsDir, 600)); err != nil {
 				return err
 			}
-			if err := p.runCommand(o, comm, "chown -R root:root "+hintsDir); err != nil {
+			if err := p.runCommand(o, comm, "chown -R root:"+p.getRootGroup()+" "+hintsDir); err != nil {
```
svanharmelen: I think it would be cleaner if we also made a `chownAix` and `chownLinux` const, just like we did for `chmodXxx`. What do you think?

cattywampus: Yeah I can get behind that. It'll help make this change more visible by placing it with its related counterpart, and it'll make the `RootGroup` parameter unnecessary. I'll fix this.

builtin/provisioners/chef/linux_provisioner.go
cattywampus: That's good to know (Go newbie here). I think with the fix below to create a `chownAIX` and `chownLinux` command constant, this function will go away.

builtin/provisioners/chef/linux_provisioner.go
cattywampus: Good to know. I'll fix this.

builtin/provisioners/chef/linux_provisioner.go
cattywampus: I do like that much better. Will fix and push up.

builtin/provisioners/chef/linux_provisioner.go
cattywampus: In my research, the `%[N]s` notation allowed for positional parameters in formatting strings to be re-used. In this case, the values for `%[1]s` are the same (`linuxConfDir`), so instead of duplicating those values for each `%s` substitution. It basically lets me do this:

```
fmt.Sprintf(p.getChmodCommand(), linuxConfDir, 600)
```

instead of this (if I used `%s`)

```
fmt.Sprintf(p.getChmodCommand(), linuxConfDir, linuxConfDir, 600)
```

The other added benefit that I see is it allows the signature of `fmt.Sprintf` be the same for both AIX and Linux since both strings use the same values, just the AIX command re-uses one of those values more than the Linux command. If I went with the latter option of using `%s` then I would have to expose the logic of checking the os type when calling `Sprintf` to switch between the two and three parameter signature usage based on if I was running on AIX or Linux. 

As for `-maxdepth`, I removed it because the find command in AIX does not support it. So using this command as is was breaking when trying to provision the chef-client on our AIX systems. It looks to me the goal of the linux command is to find all files inside `/etc/chef` and make sure their permissions are set properly without diving into any subfolders. The use of `! -path %s -prune` accomplishes the same thing, just not as obvious. It basically prunes (removes) all results that are not in the `-path %s`.

builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -10,10 +10,27 @@ import (
...
+	chmodAix   = "find %[1]s ! -path %[1]s -prune -type f -exec /bin/chmod %[2]d {} +"
 	installURL = "https://omnitruck.chef.io/install.sh"
 )
 
+func (p *provisioner) getChmodCommand() string {
+	if p.OSType == "aix" {
+		return chmodAix
+	} else {
+		return chmodLinux
+	}
```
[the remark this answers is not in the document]
cattywampus: Definitely like this better. Will fix. Thank you for the feedback.

cattywampus:
Thank you for the suggestions and feedback. I'll make these changes and push it up later this week.

builtin/provisioners/chef/linux_provisioner.go
svanharmelen: Thanks for the explanations! I already knew about `%[N]s`, but I wondered about the `! -path` and `-maxdepth` parts. But luckily you explained those as wel, so all good 👍

teamterraform:

PR #16142, opened 2017-09-21
Make salt-masterless provisioner wait for command to finish

oldmantaiter:
Previously the command would return immediately without finishing
the salt-call run and never displayed any output.

This change uses similar logic to remote-exec to pipe the output so
we can read it, and it will wait until the command finishes before
returning.

(I realize no tests were modified with this, but adding tests seems like a much bigger story than what this change entails).

### DIFF
--- a/builtin/provisioners/salt-masterless/resource_provisioner.go
+++ b/builtin/provisioners/salt-masterless/resource_provisioner.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


oldmantaiter:
FWIW I've been using this change for a couple weeks and it has been working as expected - any feedback on concerns?

oldmantaiter:
@apparentlymart - Is there anything I'm missing that helps in getting this merged?

oldmantaiter:
Closing, this was implemented fully in #16704 - thanks @subbarao !

PR #16144, opened 2017-09-21
Adding lock to GCS remote backend using pubsub topics

kavehmz:
Adding lock mechanism to GCS remote backend by using pubsub Topics.

Concept is similar to how we create/delete lock files to acquire/release lock.

We are using remote state-file in our team. We need this in our team to be able to work in a safer way.

Code changes is small.

```hcl
terraform {
  backend "gcs" {
    bucket  = "tf-state-prod"
    path    = "path/terraform.tfstate"
    project = "myproject"
    lock_topic = "remote_lock"
  }
}
```

Only addition will be __lock_topic__. As far as I understand the change is backward compatible and won't break any existing setup.

I added lock test and all is green. Lock test passes locally against my dev project in google cloud.

I ran a separate code against pubsub to check if __create__ can be used as lock mechanism. As far as I can tell it is a serial action (clear in results.txt if you look at timing), and usable for lock. You can find the code and the result at the following gist:
`git clone git@gist.github.com:08bf9c411478d42707080773b8a7b7ec.git`
or
`https://gist.github.com/kavehmzta/08bf9c411478d42707080773b8a7b7ec`

Adding this change will help me greatly and I am hoping that this method might be useful for others too.

### DIFF
--- a/state/remote/gcs.go
+++ b/state/remote/gcs.go
@@ -3,6 +3,7 @@ package remote
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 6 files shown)


sl1pm4t:
Hi @kavehmz - there is already a  PR open that adds locking support for GCS backend over in #15592.
The locking is achieved using GCS directly ( so it doesn't require PubSub etc. ).

kavehmz:
@sl1pm4t Great then I hope it gets merged soon.

PR #16213, opened 2017-09-28
Don't evaluate locals during destroy

### LINKED ISSUE #16190: terraform destroy error for a local value from random_id resource
### Terraform Version

0.10.6

```hcl
locals {
  name = "test-${random_id.id.hex}"
}
resource "random_id" "id" { byte_length = 4 }

output "name" {
  value = "${local.name}"
}
```

### Debug Output

https://gist.github.com/daveadams/88703e0511d9c8f20757ecf061a1a9f7

### Expected Behavior

Destroy should proceed without errors, caring nothing about the impact of resource existence on local values.

### Actual Behavior

Destroy complete successfully, but with a nasty error about a missing resource value in a local value interpolation. This causes sadness.

### Steps to Reproduce

Given the TF code above:

    terraform init; terraform apply; terraform destroy

jbardin:
Locals don't need to be evaluated during destroy.  Rather than simply
skipping them, remove them from the state as they are encountered. Even
though they are not persisted in the state, it keeps the state up to
date as the destroy happens, and we reduce the chance of other
inconstancies later on.

Fixes #16190

### DIFF
--- a/terraform/context_apply_test.go
+++ b/terraform/context_apply_test.go
@@ -8868,3 +8868,51 @@ func TestContext2Apply_destroyWithLocals(t *testing.T) {
 		t.Fatalf("wrong final state\ngot:\n%s\nwant:\n%s", got, want)
 	}
 }
+
+func TestContext2Apply_providerWithLocals(t *testing.T) {
+	m := testModule(t, "provider-with-locals")
+	p := testProvider("aws")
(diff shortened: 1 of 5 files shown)


jbardin:
Good idea. This seems to work as expected since destroying the provider doesn't require re-evaluation.

PR #16223, opened 2017-09-29
Changes to module documentation 

cgriggs01:
Adding a few more requirements to the Creating Modules README docs.

### DIFF
--- a/website/docs/modules/create.html.markdown
+++ b/website/docs/modules/create.html.markdown
@@ -123,7 +123,7 @@ standard structure.
   including a visual diagram depicting the infrastructure resources the
   module may create and their relationship. The README doesn't need to 
   document inputs or outputs of the module because tooling will automatically
-  generate this. If you are linking to file or embedding an image contained
+  generate this. If you are linking to a file or embedding an image contained
   in the repository itself, use a commit-specific absolute URL so the link 
   won't point to the wrong version of a resource in the future.
 
--- a/website/docs/modules/create.html.markdown
+++ b/website/docs/modules/create.html.markdown
@@ -120,8 +120,8 @@ standard structure.
   files. This file should be named `README` or `README.md`. The latter will
   be treated as markdown. There should be a description of the module and 
   what it should be used for. If you want to include an example for how this
-  module can be used in combination with other resources, put it in an examples
-  [directory like this](https://github.com/hashicorp/terraform-aws-consul/tree/master/examples).
+  module can be used in combination with other resources, put it in an [examples
+  directory like this](https://github.com/hashicorp/terraform-aws-consul/tree/master/examples).
   Consider including a visual diagram depicting the infrastructure resources 
   the module may create and their relationship. The README doesn't need to 
   document inputs or outputs of the module because tooling will automatically 
--- a/website/docs/modules/create.html.markdown
+++ b/website/docs/modules/create.html.markdown
@@ -118,8 +118,12 @@ standard structure.
 
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
-  be treated as markdown. The README doesn't need to document inputs or
-  outputs of the module because tooling will automatically generate this.
+  be treated as markdown. There should be a description of the module and 
+  what it should be used for, as well as a basic usage example.  If possible,
+  a visual diagram of the modules should be included. The README doesn't need 
+  to document inputs or outputs of the module because tooling will automatically
+  generate this. All links to other documentation or image resources they must 
+  be absolute.
 
 * **LICENSE**. The license under which this module is available. If you are
   publishing a module publicly, many organizations will not adopt a module
--- a/website/docs/modules/create.html.markdown
+++ b/website/docs/modules/create.html.markdown
@@ -119,13 +119,15 @@ standard structure.
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
   be treated as markdown. There should be a description of the module and 
-  what it can be used for. Consider including a visual diagram depicting the 
-  infrastructure resources the module may create and their relationship. 
-  The README doesn't need to document inputs or outputs of the module because 
-  tooling will automatically generate this. If you are linking to a file or 
-  embedding an image contained in the repository itself, use a commit-specific
-  absolute URL so the link won't point to the wrong version of a resource in 
-  the future.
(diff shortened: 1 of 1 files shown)


website/docs/modules/create.html.markdown
```diff
@@ -118,8 +118,12 @@ standard structure.
 
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
-  be treated as markdown. The README doesn't need to document inputs or
-  outputs of the module because tooling will automatically generate this.
+  be treated as markdown. There should be a description of the module and 
+  what it should be used for, as well as a basic usage example.  If possible,
+  a visual diagram of the modules should be included. The README doesn't need
```
banks: "Consider including a visual diagram depicting the infrastructure resources the module may create and their relationship."

Maybe?

website/docs/modules/create.html.markdown
```diff
@@ -118,8 +118,12 @@ standard structure.
 
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
-  be treated as markdown. The README doesn't need to document inputs or
-  outputs of the module because tooling will automatically generate this.
+  be treated as markdown. There should be a description of the module and 
+  what it should be used for, as well as a basic usage example.  If possible,
+  a visual diagram of the modules should be included. The README doesn't need 
+  to document inputs or outputs of the module because tooling will automatically
+  generate this. All links to other documentation or image resources must be
```
banks: That external contributor made a fair point about that. I will try to fix README rendering to support github-hosted relative links.

banks:
Kevin orginally suggested 

> Link to any relevant documentation for the created resources

You think that's worth mentioning Chris or is that kinda obvious?

website/docs/modules/create.html.markdown
```diff
@@ -118,8 +118,12 @@ standard structure.
 
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
-  be treated as markdown. The README doesn't need to document inputs or
-  outputs of the module because tooling will automatically generate this.
+  be treated as markdown. There should be a description of the module and 
+  what it should be used for, as well as a basic usage example.  If possible,
```
banks: Did we decide on encouraging Usage examples? I thought we wanted to steer people away from that since we have machine-generated ones anyway and will have complete examples too.

website/docs/modules/create.html.markdown
banks: After some effort and consideration it's not a simple problem. While we _could_ fix for Github, we don't want to commit to reverse-engineering the URL structure of every VCS provider that Public or Private TF registries might ever support.

So this recommendation should remain. I'd maybe add a sentence like: "If you are linking to file or embedding an image contained in the repository itself, use a commit-specific absolute URL so the link won't point to the wrong version of a resource in the future.

cgriggs01:
That makes sense, your changes look good to me. Feel free to merge when ready.

website/docs/modules/create.html.markdown
```diff
@@ -118,8 +118,14 @@ standard structure.
...
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
-  be treated as markdown. The README doesn't need to document inputs or
-  outputs of the module because tooling will automatically generate this.
+  be treated as markdown. There should be a description of the module and 
+  what it should be used for, as well as a basic usage example. Consider
+  including a visual diagram depicting the infrastructure resources the
+  module may create and their relationship. The README doesn't need to 
+  document inputs or outputs of the module because tooling will automatically
+  generate this. If you are linking to file or embedding an image contained
```
banks: nitpick: "**a** file"

website/docs/modules/create.html.markdown
```diff
@@ -118,8 +118,14 @@ standard structure.
 
 * **README**. The root module and any nested modules should have README
   files. This file should be named `README` or `README.md`. The latter will
-  be treated as markdown. The README doesn't need to document inputs or
-  outputs of the module because tooling will automatically generate this.
+  be treated as markdown. There should be a description of the module and 
+  what it should be used for, as well as a basic usage example. Consider
```
banks: I think I mentioned before but did we decide on encouraging usage examples? It seems this duplicates the generated usage block at the top right?

cgriggs01:
@banks - I don't have write access. Can you do the merge?

banks:
I think I can merge.

I didn't see answer to question about:

> as well as a basic usage example

Not a big deal but I thought we talked about not encouraging usage examples since we already have the generated one at the top and will surface more in the future. @KFishner I think we said we'd prefer they don't duplicate that usage thing right to make the page less confusing which one to use?

KFishner:
Yeah I agree with @banks, we shouldn't have a basic usage example in the README since we'll want to auto-gen that eventually. However, it's important to differentiate that from a more complicated usage example, which should be put in a `/examples` folder like this: https://github.com/hashicorp/terraform-aws-consul/tree/master/examples. 

I made a suggestion above in the context of where this would go in the docs

website/docs/modules/create.html.markdown
KFishner: There should be a description of the module and what it should be used for. If you want to include an example for how this module can be used in combination with other resources, put it in an [examples directory like this](https://github.com/hashicorp/terraform-aws-consul/tree/master/examples)

cgriggs01: @KFishner we already call out the examples in a section below this:

 "Examples. Examples of using the module should exist under the examples/ subdirectory. Each example may have a README to explain the goal and usage of the example." 

Should we mention the examples in the README section as well?

KFishner: yeah I think it's a good idea to call it out in the README

cgriggs01:
Made suggested updates to the change, let me know if there is anything else. We should be ready to merge if not.

banks:
@jbardin should I just merge this? Is it worth making a release of the website to get this live? I think we should if that's not a lot of work and not going to happen for month+ anyway, but if either is not true, I anticipate much documentation updating in next few weeks for 0.11 stuff anyway...

jbardin:
@banks, go right ahead!

In Terraform, we usually have the submitter merge once the PR's approved.
As for the website, I think it's a single step process in TeamCity now, so we can probably get that updated asap.

PR #16252, opened 2017-10-03
faster DAG transitive reduction

jbardin:
In the case of highly-connected graphs, the TransitiveReduction process was far
too computationally intensive. Since no operations are applied to the nodes,
and the walk order is not even user visible, we don't need to sort them n^2
times.

The `e == nil` checks aren't directly related, but short-circuit the `marshalVertex` calls in the debug path.

### DIFF
--- a/dag/dag.go
+++ b/dag/dag.go
@@ -106,7 +106,7 @@ func (g *AcyclicGraph) TransitiveReduction() {
 		uTargets := g.DownEdges(u)
 		vs := AsVertexList(g.DownEdges(u))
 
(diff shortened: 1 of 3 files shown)


jbardin:
I think we can probably do better in the targeting transformer to prevent these fully connected graphs in the first place too (I guess that gets close to re-implementing DynamicExpand during apply? I haven't looked closely). They get created when you have some interdependent splat interpolations, yet those should be resolvable to specific resources or the total count.

@sethvargo: This should help significantly with the training config.

PR #16280, opened 2017-10-06
Add habitat provisioner

nsdavidson:
This adds a Habitat (https://habitat.sh) provisioner.

### DIFF
--- a/builtin/provisioners/habitat/resource_provisioner.go
+++ b/builtin/provisioners/habitat/resource_provisioner.go
@@ -1,0 +1,598 @@
+package habitat
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform/communicator"
+	"github.com/hashicorp/terraform/communicator/remote"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+	linereader "github.com/mitchellh/go-linereader"
+)
+
+const install_url = "https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh"
+
+var serviceTypes = map[string]bool{"unmanaged": true, "systemd": true}
+var updateStrategies = map[string]bool{"at-once": true, "rolling": true, "none": true}
+var topologies = map[string]bool{"leader": true, "standalone": true}
+
+type provisionFn func(terraform.UIOutput, communicator.Communicator) error
+
+type provisioner struct {
+	Version       string    `mapstructure:"version"`
+	Services      []Service `mapstructure:"service"`
+	PermanentPeer bool      `mapstructure:"permanent_peer"`
+	ListenGossip  string    `mapstructure:"listen_gossip"`
+	ListenHTTP    string    `mapstructure:"listen_http"`
+	Peer          string    `mapstructure:"peer"`
+	RingKey       string    `mapstructure:"ring_key"`
+	SkipInstall   bool      `mapstructure:"skip_hab_install"`
+	UseSudo       bool      `mapstructure:"use_sudo"`
+	ServiceType   string    `mapstructure:"service_type"`
+}
+
+func Provisioner() terraform.ResourceProvisioner {
+	return &schema.Provisioner{
+		Schema: map[string]*schema.Schema{
+			"version": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"peer": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"service_type": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "unmanaged",
+			},
+			"use_sudo": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"permanent_peer": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+			"listen_gossip": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"listen_http": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"ring_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"service": &schema.Schema{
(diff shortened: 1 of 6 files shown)


apparentlymart:
Hi @nsdavidson! Thanks for working on this.

We've traditionally been cautious about adding vendor-specific provisioners because the provisioner abstraction in Terraform is not very strong (compared to resources, which are the _primary_ Terraform concept) and so they tend to be limiting in non-trivial environments. We added some while we were still learning, and but we've generally been advising users to either bake necessary information into their machine images (avoiding the need for provisioning at all) or to build using the `file` and `remote-exec` provisioners, which tend to compose better with features such as destroy provisioners.

With that said, I'd like to explore the following with you:
* Is installation of Habitat a _stateless_ operation? That is, does it just affect the target machine, or does it also create records on some central service? The existing `chef` provisioner is stateful, for example, because it creates a record of the instance in the Chef server and that has, in practice, proven problematic since Terraform has no means to de-register an instance when it's destroyed.
* Is this provisioner doing something that would be impossible or very difficult to achieve using `file` and `remote-exec`? I had a quick read of the steps and it didn't _seem_ so, but given that you mentioned that it works only with the `ssh` communicator I wonder if I missed something complex going on here.
* I assume there'd be no good reason for this provisioner to be used as an "on-destroy" provisioner, since it doesn't really make sense to install new software while you're in the process of destroying an instance.

bixu:
@apparentlymart, I think I can help here, in answer to your questions:

1). Yes, Habitat is _essentially_ stateless. There's some gossip that happens between Habitat services, but that conversation is designed to allow members of the gossip ring to depart unexpectedly (like during a destroy operation).

2). At my org we do use the `file` and `remote-exec` resources, and have found them to work well, although we don't gain the advantages of HCL's readability/clarity, so I can understand the desire to have a native provider.

3). See my first answer - Habitat is designed with self-departure of group/ring members in mind.

nsdavidson:
@apparentlymart Thanks for taking a look at it!

@bixu is pretty much on point with the responses.  It is stateless and no action needs to be taken upon destroying. I think most folks currently provisioning Habitat services with Terraform are using `file` and `remote-exec`, but it's not terribly clean or clear as to what is being provisioned.  My goal here was to provide a way to clearly describe the infrastructure and service configuration together.

nsdavidson:
@apparentlymart Is there anything else I can help answer here?

bixu:
My or plans to transition completely to Terraform + Habitat for our infrastructure & service management needs, so we’d love to use this.

smartb-pair:
@apparentlymart, we'd love to see this integrated. We're trying to be an all Terraform+Habitat shop at the moment.

apparentlymart:
Hi @nsdavidson! Sorry for the silence here.

To be transparent with you, we ended up stalling on this because we've found it pretty hard to support the existing service-specific provisioners already present (chef, salt-masterless) because we don't have in-house expertise with these nor infrastructure in place to test them. Thus we find ourselves pretty nervous about adding new ones that may end up in the same situation.

Given that there are far fewer provisioners than providers, we haven't developed a process for dealing with community-driven provisioner development/maintenance and so we found ourselves a bit stalled here, not really knowing how best to proceed.

Do you think that, were this to be merged, you'd be able to help field issues and PRs that might be submitted against this provisioner on an ongoing basis? With the provisioner still living inside the main repository (vs. providers, which were split out) we don't have a proper mechanism in place yet for direct community maintenance of provisioners, but if you'd be willing to monitor for new issues and test/review new PRs (which we could then merge, trusting your review/test) I think that would help assuage the concern here. A new issue/PR label `provisioner/habitat` would allow you to more easily find these, hypothetically.

nsdavidson:
@apparentlymart No worries, that all makes sense. I totally understand the reservations.

For sure, I'd be happy to help with issues and PRs.

apparentlymart:
Thanks @nsdavidson! I've merged this for inclusion in the next release.

apparentlymart:
Any issues and PRs for this provider will show up under the [`provisioner/habitat`](https://github.com/hashicorp/terraform/labels/provisioner%2Fhabitat) label, once we tag them during our triage/labelling process.

bixu:
@nsdavidson, I could be missing something because I’m on thumbs, but I don’t see anything in this PR to handle GitHub Auth Tokens, which are used to access private Builder Origins. We’d need to pass such a token in to `hab` commmand and also make it available to the systemd service in the unit file, I think.

Feel free to ping me in the community Slack channel if you have questions. Very excited to use this :)

builtin/provisioners/habitat/resource_provisioner.go
```diff
@@ -0,0 +1,745 @@
...
+	"github.com/hashicorp/terraform/terraform"
+	linereader "github.com/mitchellh/go-linereader"
+)
+
+const installURL = "https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh"
+const systemdUnit = `[Unit]
+Description=Habitat Supervisor
+
+[Service]
+ExecStart=/bin/hab sup run %s
```
smartb-pair: I think here we optionally need something like:
```
Environment="HAB_AUTH_TOKEN=<github_token>"
```
for anyone accessing private origins in Builder.

builtin/provisioners/habitat/resource_provisioner.go
```diff
@@ -0,0 +1,745 @@
...
+		command += fmt.Sprintf("hab install core/hab-sup")
+	} else {
+		command += fmt.Sprintf("hab install core/hab-sup/%s", p.Version)
+	}
+
+	if p.UseSudo {
+		command = fmt.Sprintf("sudo -E %s", command)
+	}
+
+	command = fmt.Sprintf("env HAB_NONINTERACTIVE=true %s", command)
```
smartb-pair: @nsdavidson, I think we also need to be able to pass `HAB_AUTH_TOKEN` in here.

builtin/provisioners/habitat/resource_provisioner.go
```diff
@@ -0,0 +1,745 @@
...
+	if p.UseSudo {
+		command = fmt.Sprintf("sudo %s", command)
+	}
+	return p.runCommand(o, comm, command)
+}
+
+func (p *provisioner) startHabService(o terraform.UIOutput, comm communicator.Communicator, service Service) error {
+	var command string
+	if p.UseSudo {
+		command = fmt.Sprintf("env HAB_NONINTERACTIVE=true sudo -E hab pkg install %s", service.Name)
```
smartb-pair: @nsdavidson, I think we also need to be able to pass `HAB_AUTH_TOKEN` in here.

PR #16283, opened 2017-10-07
remove old import tree code

jbardin:
The import code and concept of a configuration tree was all pre 0.1, and
no longer used. This removes the multiple layers of indirection, and
loads configuration files more directly.


jbardin:
Not sure how failure got through :/ Anyway, closing this as we'll strip out this old code after larger refactors to minimize churn.

PR #16290, opened 2017-10-08
retry on s3 state upload errors

jbardin:
While #16243 added the ability to retry getting a state from S3, Put can
return the same `InternalError` status. Use the same retry logic when
uploading state to S3.


jbardin:
Yes, there shouldn't be any additional drawback here, plus this is only checking for `InternalError`, in which case nothing was _supposed_ to have happened.

PR #16294, opened 2017-10-09
Aliasing the module to omit errors thrown at build time

doertedev:
Signed-off-by: Stefan Staudenmeyer <stefan.staudenmeyer@instana.com>

### DIFF
--- a/config/module/tree.go
+++ b/config/module/tree.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/hashicorp/go-getter"
+	getter "github.com/hashicorp/go-getter"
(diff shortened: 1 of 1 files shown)


jbardin:
Hi @doertedev,

This is OK and editing the imports would have re-written it like this anyway, but I'm wondering what errors were you seeing? The code as it is is perfectly valid as it is, and of course has been released many times in the last few years.

doertedev:
Oy.
getter was undefined.
`github.com/hashicorp/terraform/config/module/tree.go:263:20: undefined: getter.SubdirGlob`

AnikoTheLawyer:
gettor

PR #16357, opened 2017-10-14
FIX-15921 Clean clients from chef-vault on recreate_client enabled

### LINKED ISSUE #15921: Terraform should clean clients for chef vault
When provisioning a server with chef and using chef-vaults it happens that if you destroy and recreate a server the vault is not updated and the new machine can not read/decrypt the vault.

### Terraform Version
 0.10.2

### Expected Behavior
Vaults should be encrypted for new client.

### Actual Behavior
The old chef client gets deleted [1] from the server but the vault stays encrypted for the old client.
This is tricky. There are options to to delete unknown clients (--clean in [2]) and such, but should terraform potentially clean access for clients that it does not actually touch?
We have this problem here from time to time whenever we try to recreate resources for testing restores and such and it is quite annoying.

[1] https://github.com/hashicorp/terraform/blob/0c69e466496e86b23e0024889d01efa074b75035/builtin/provisioners/chef/resource_provisioner.go#L527
[2] https://github.com/hashicorp/terraform/blob/0c69e466496e86b23e0024889d01efa074b75035/builtin/provisioners/chef/resource_provisioner.go#L563

supergicko:
Fixes #15921

When terraform re-creates an existing node/client with chef provisioner,
the already existing client (which has old keys) must be removed from
the vault items. Afterwards, the chef-vault will be updated with the
newly created client (which has the new keys). Therefore, the recreated
client will be able to decrypt the vault items properly.

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -558,6 +558,25 @@ func (p *provisioner) configureVaultsFunc(gemCmd string, knifeCmd string, confDi
 			path.Join(confDir, p.UserName+".pem"),
 		)
 
+		// if client gets recreated, remove (old) client (with old keys) from vaults/items
+		// otherwise, the (new) client (with new keys) will not be able to decrypt the vault
+		if p.RecreateClient {
+			for vault, items := range p.Vaults {
+				for _, item := range items {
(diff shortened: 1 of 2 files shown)


supergicko:
@svanharmelen would love to hear your thoughts on this one.

svanharmelen:
Thanks for the PR @supergicko! LGTM!

PR #16369, opened 2017-10-17
Makefile: handle GOBIN properly if set

### LINKED ISSUE #9912: make plugin-dev should be removed from the Makefile
### Terraform Version
fails when building modules for  0.7.9 at least, did not try with older versions

##############################################################################
terraform make plugin-dev .. fails if GOBIN is set and not same as $GOPATH/bin

should 'make plugin-dev' look like this?

plugin-dev: generate
	go install github.com/hashicorp/terraform/builtin/bins/$(PLUGIN)
	mv $(GOBIN)/$(PLUGIN) $(GOBIN)/terraform-$(PLUGIN)

About GOBIN:
The bin directory holds compiled commands. Each command is named for its source directory, but only the final element, not the entire path. That is, the command with source in DIR/src/foo/quux is installed into DIR/bin/quux, not DIR/bin/foo/quux. The "foo/" prefix is stripped so that you can add DIR/bin to your PATH to get at the installed commands. If the GOBIN environment variable is set, commands are installed to the directory it names instead of DIR/bin. GOBIN must be an absolute path

abitrolly:
Fixes #9912

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 TEST?=./...
 GOFMT_FILES?=$$(find . -name '*.go' | grep -v vendor)
+ifeq ($(GOBIN),)
+    GOBIN=$(shell go env GOPATH)/bin
+endif
 
 default: test vet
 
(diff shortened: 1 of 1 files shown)


mildwonkey:
Now that the providers are in their own repositories, make plugin-dev should just be removed from Terraform's makefile. I am going to close this PR, please feel free to open a new one and thank you for submitting this!

Issue #15520, opened 2017-07-10
Initializing providers for hundreds of modules makes plan and apply slow


vvondra:
### Expected Behavior
What should have happened?

Running a plan or apply should re-use the provider configured in the root modules

### Actual Behavior

The same provider is configured/initialized for each module, slowing down plan tremendously. If the provider doesnt change from the root/parent module, it should be re-used in the module traversal.

Sample log:
```
module.foo_bar_sns_topic.aws_cloudwatch_metric_alarm.event_bus_topic_queue_alarm", waiting for: "module.foo_bar_sns_topic.provider.aws"
module.bar_baz_sns_topic.aws_sqs_queue.event_bus_topic_queue", waiting for: "module.bar_baz_sns_topic.provider.aws"
...
...
and hundreds more like this
```

### Description

We have a large generated terraform template which includes a few hundred instances of the same locally sourced module. It contains about 5 resources (sns, sqs, policy, subscription).

If we run plan, it runs very very slow. If I enabled debug output, it seems to be running ConfigureFunc of the AWS provider for each instance of the module. This generates a lot of API calls to set up the client (most slow ones are in GetAccountInfo)

Feel free to correct me that this is an AWS provider only ticket, but my feeling is it affects all providers.

jbardin:
Hi @vvondra,

The handling of provider has been overhauled in the terraform core, and should now initialize the providers only the minimum number of times. As of 0.11.0-rc1, each provider is only instantiated once globally for each configuration (though there may still be duplicate configurations), rather than once for each module that uses it. Additionally, provider passed into a module via the new module/providers mapping will use the root provider even though the module has an empty provider block.

You can see some of the examples being worked on in the upgrade guide https://github.com/hashicorp/terraform/blob/master/website/upgrade-guides/0-11.html.markdown

vvondra:
Awesome 🙌

Issue #11578, opened 2017-01-31
allow variable in provider field


dmrzzz:
I'm using multiple AWS providers to deploy resources into different regions, but I want my modules to be generic with respect to _which_ regions.

Ideally I would write a module like this:
```hcl
variable provider { default = "aws" }

resource "aws_vpc" "vpc" {
    # works:
    #provider = "aws.us-east-1"
    # does not work:
    provider = "${var.provider}"
    cidr_block = "192.168.0.0/16"
    tags { Name = "TEST - DELETE ME" }
}
```
and invoke it like this:
```hcl
provider "aws" {
  region = "us-east-2"
}

provider "aws" {
  alias = "us-east-1"
  region = "us-east-1"
}

# do other stuff using the default AWS provider

# tell this module to use aws.us-east-1
module "module1" {
  source = "./testmodule"
  provider = "aws.us-east-1"
}
```
but that fails on `terraform get`:
```
dmrz@golbez:~/tmp/tftest2$ terraform get
Get: file:///home/dmrz/tmp/tftest2/testmodule
Error loading Terraform: 1 error(s) occurred:

* module module1: provider alias must be defined by the module or a parent: ${var.provider}
```

It works fine if I instead hardcode the provider field value within the module, but then my module isn't generic.

Note: the resource I _really_ want to use this for at this moment is `aws_cloudwatch_metric_alarm`, but `aws_vpc` behaves the same way and makes a much simpler test case.

Thanks for all the great work on Terraform!

kevintsai:
I also ran into this issue. Hope it can be improved. Thanks.

apparentlymart:
This is somewhat related to #1819.

kkarimi:
This is quite necessary in order to work effectively with modules with a multi region AWS setup

ericfode:
This is causing sadness for us.

nyetsche:
Also causing sadness for us :( 

Here's the example of what I'm hoping to accomplish:

```
resource "aws_kms_key" "cluster_cmk" {
  count       = "${var.cluster_number}"
  provider = "aws.${element(values(var.regionmap), count.index)}"
}

provider "aws" {
  alias = "us-east-1"
  region = "us-east-1"
}

regionmap = {
  "clusternickname" = "us-east-1"
}
```

In other words, have multiple KMS keys, one per cluster, and each cluster could be in a different region. So do a map lookup.

Fails with this:

```
* module root: 1 error(s) occurred:

[...]

PR #16379, opened 2017-10-17
module provider inheritance

jbardin:
The primary goal of this PR is to define the new provider inheritance model, while hoisting the provider inheritance mechanisms into the config itself. The new providers field in the module block allows users to declare a provider configuration to apply within a module. Since this new inheritance mechanism is configuration based, it is much simpler to apply all the inheritance rules within the config, rather than searching for provider configurations during graph evaluation.

The only remnant of "inheritance" in core is now through the provider input cache. Though the cached Input result no longer force unintended inheritance, the fields added by a provider's Input call are added to the config of other providers by the same name. It might be advantageous to remove the input cache just as we've done with the config cache, but that would cause every Provider missing fields to prompt, which may be inconvenient for users relying on manual Inputs. 

Due to the removal of inheritance in the terraform package, orphaned module resources can no longer be destroyed without a provider config. This is not optimal, but can be handled in a later PR to keep this more manageable.


vvondra:
could this theoretically fix https://github.com/hashicorp/terraform/issues/15520?

jbardin:
Hi @vvondra, 

Unfortunately not. This still will configure the provider for each provider instance in the configuration.

I am exploring some paths that might be able to reduce the number of virtual "provider instances" in the graph, but optimizing for lots of providers is still a ways off.

jbardin:
Orphaned modules will have to wait, merging this into a dev branch.

thomasbiddle:
@jbardin Can we get an example of how this would look? I was following #11578 originally.

jbardin:
@thomasbiddle,

I'm not sure what you're looking for exactly; does the documentation help answer your question? https://www.terraform.io/docs/modules/usage.html#providers-within-modules

jbardin:
@bukzor: Removing modules is working as intended. If the provider configuration is removed, there's no way to reliably destroy the resources in a module. If the provider configuration is passed into the module, then you can retain the configuration separate from the module source.

bukzor:
I'm hearing that the issue doesn't exist. I've worked up a demo and explained the problems in [this github repo](https://github.com/bukzor/terraform-gke-k8s-demo). Should I file each of those gripes as an issue, or would you be willing to do triage? There's two misfeatures and one outright bug here, by my count.

https://github.com/bukzor/terraform-gke-k8s-demo

jbardin:
Hi @bukzor,

I haven't gone through that repo in detail, but the first thing that stands out is that there are provider configurations within the module, which is what isn't going to work. We do have some known issues around providers, like #13018, and some new things implemented like #16835.

Of course feel free to file an other issues you have that aren't covered.

PR #16393, opened 2017-10-18
Support for path style S3 bucket address

mevansam:
Added flag to allow s3 bucket address to be created with bucket name in path i.e. http(s)://s3.amazonaws.com/<bucket>/<object>. This style of addressing is required by non-AWS S3 distributions such as Minio.

### DIFF
--- a/backend/remote-state/s3/backend.go
+++ b/backend/remote-state/s3/backend.go
@@ -150,6 +150,13 @@ func New() backend.Backend {
 				Default:     false,
 			},
 
+			"s3_force_path_style": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Force path style S3 bucket address.",
+				Default:     false,
+			},
+
 			"role_arn": {
 				Type:        schema.TypeString,
 				Optional:    true,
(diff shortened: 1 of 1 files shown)


bhood4:
Nice.  I need this as well for MOS Openstack/Ceph S3, and was just about to do something similar.

Maybe terraform/website/docs/backends/types/s3.html.md could use an extra line:

 * `s3_force_path_style` - (Optional) Force path-style instead of virtual-hosted–style URLs to access a bucket.  Default is false for virtual-hosted–style URLs.

FancyCoder0:
Hi Mevan Samaratunga(@mevansam) & @bhood4,
We are working on identifying redundant development and duplicate pull requests. We have found there is a pull request: https://github.com/hashicorp/terraform/pull/17001 which might be duplicate to this one and already merged into main stream. So maybe this pull request could be closed.

We would really appreciate if you could help us to validate and give us some feedback. Thank you very much for your time!

bflad:
Hi @mevansam 👋 Thanks so much for submitting this. It appears this was accidentally missed in previous review cycles for the S3 backend code. Another similar pull request, #17001, was merged with the same functionality (albeit the configuration was named `force_path_style` instead of `s3_force_path_style`). The `force_path_style` configuration for the S3 backend has been available in Terraform since 0.11.2, although its been missing from the Terraform website documentation (as of this writing, up to 0.11.11) as noted in #19672, which will be fixed up shortly.

Apologies again for the review miss here and thanks for your contribution.

PR #16416, opened 2017-10-22
Improve performance of dev builds by ~20%...

sean-:
At the expense of ~150MB of RAM improve the performance of compiling Terraform by ~20%.  Before this change the peak memory consumption appears to be `link` coming in at ~1.2GB of RAM.  With this change both `compile` and `link` take ~1.3GB of RAM.  The extra 100MB of RAM is justifiable given the improvements to build times.  With the new parallel compiler in Go 1.9 the improvement to wall clock time is not as significant, but the user time and power consumption on laptops should go down as a result of this change, too (see the first commit for userland CPU time, vs wall clock times reported below).

```
$ gmake dev
Total time                              : 0:55.36s
```

Time with patch:

```
$ gmake dev
Total time                              : 0:46.73s
```

This PR only makes this change for the `dev` and `quickdev` targets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ script:
 - make test
 - make e2etest
 - make vet
-- GOOS=windows go build
+- GOGC=off GOOS=windows go build
 branches:
   only:
   - master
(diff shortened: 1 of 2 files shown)


sean-:
Travis is pretty variable in its times, but it there seems to be an improvement there, too (before: ~11:50s, after: ~8:13s, 7:06s, and ~4:25s).

teamterraform:
We have made many changes to our development and build tooling since this PR was opened, and will therefor close this outdated PR. If you are still interested in this feature, please open a new Github Issue using the feature request template so we can discuss the problem and design.

PR #16437, opened 2017-10-24
helper/resource: allow truncating prefix of PrefixedUniqueId

hakamadare:
Add an optional `max_length` parameter to `PrefixedUniqueId`; when
present, this parameter truncates the `prefix` parameter before
constructing the unique ID string.  This provides a potential solution
to the problem described in
terraform-providers/terraform-provider-aws#625; since AWS enforces a
variety of length limits on resource names, it's not too hard when using
the `name_prefix` resource parameter to generate invalid names (which
cannot be discovered until the plan is applied).  The `max_length`
parameter offers provider maintainers the option of capturing the length
limits and truncating them rather than generating invalid names.

### DIFF
--- a/helper/resource/id.go
+++ b/helper/resource/id.go
@@ -20,6 +20,14 @@ func UniqueId() string {
 
 // Helper for a resource to generate a unique identifier w/ given prefix
 //
+// This wraps PrefixedUniqueIdLimit to preserve the previous interface
+
+func PrefixedUniqueId(prefix string) string {
+	return PrefixedUniqueIdLimit(prefix, -1)
+}
+
+// Helper for a resource to generate a unique identifier w/ given prefix
+//
 // After the prefix, the ID consists of an incrementing 26 digit value (to match
 // previous timestamp output).  After the prefix, the ID consists of a timestamp
 // and an incrementing 8 hex digit value The timestamp means that multiple IDs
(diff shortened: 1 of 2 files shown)


hakamadare:
hello folks!  this PR isn't ready to merge yet; I get this error when running the test suite:

```
vagrant@terraform:/opt/gopath/src/github.com/hashicorp/terraform$ make test
go generate ./...
2017/10/24 14:36:46 Generated command/internal_plugin_list.go
go test -i ./... || exit 1
# github.com/hashicorp/terraform/helper/resource
helper/resource/id.go:40:48: prefix declared and not used
Makefile:33: recipe for target 'test' failed
make: *** [test] Error 1
```

I see discussion online of using `_` to quash the warning, but that doesn't quite seem like the right way to fix it; is there a more idiomatic way to accomplish what I'm trying to do?

hakamadare:
ok, i've made some changes and now no longer see the "declared and not used" error.

helper/resource/id.go
```diff
@@ -27,14 +28,35 @@ func UniqueId() string {
...
+	}
+
 	idMutex.Lock()
 	defer idMutex.Unlock()
 	idCounter++
-	return fmt.Sprintf("%s%s%08x", prefix, timestamp, idCounter)
+	return fmt.Sprintf("%s%s%08x", final_prefix, timestamp, idCounter)
+}
+
+func TruncatedPrefix(prefix string, max_length int) string {
```
apparentlymart: Given that `TruncatedPrefix` and `MaxInt` are used only once, and they are not long functions, I'd prefer to inline them in the main function just to avoid these being in the global scope where their context is not obvious. After inlining I think we can also avoid the conversion to/from floats with `math.Max`, since we know the lower bound will always be zero:

```
if maxLength > len(prefix) {
  maxLength = len(prefix)
}
```

Note that in Go the first letter being capitalized signals that a function is part of the public API of a package, so for internal helper functions you'd normally name them with a lowercase letter to mark them as private to the package.

hakamadare:
ok, this feedback is SUPER helpful @apparentlymart, thank you :)

no objection to any of your suggestions, i’ll rework this and get back to you.

-- 
http://five.sentenc.es

hakamadare:
@apparentlymart i've made some changes based on your feedback, please lmk if you have further suggestions.

ColinHebert:
As discussed in https://github.com/terraform-providers/terraform-provider-aws/issues/625 we might be better off changing the way we're generating IDs to be more efficient rather than truncating

hakamadare:
@ColinHebert i think your suggestion of moving to a more efficient encoding is a good one; however, i'd really like to get this PR merged and released for the following reasons:

1. AFAICT it should have absolutely no impact right now; i don't expect this PR to change any behavior of existing Terraform code
2. this opens the way to begin implementing a fix to this problem, which is hurting people right now, over in the AWS provider

changing to a different encoding is a breaking change.

hakamadare:
@apparentlymart ping, please take another look at this PrefixedUniqueId truncation patch when you have a moment

teamterraform:

PR #16456, opened 2017-10-26
helper/schema: Loosen validation for 'id' field

radeksimko:
As discussed in https://github.com/terraform-providers/terraform-provider-aws/pull/1626#issuecomment-328881415

### DIFF
--- a/helper/schema/resource.go
+++ b/helper/schema/resource.go
@@ -393,19 +393,43 @@ func (r *Resource) InternalValidate(topSchemaMap schemaMap, writable bool) error
 				return err
 			}
 		}
+
+		for k, f := range tsm {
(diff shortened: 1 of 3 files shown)


helper/schema/resource.go
```diff
@@ -450,7 +474,7 @@ func (r *Resource) TestResourceData() *ResourceData {
 // Returns true if the resource is "top level" i.e. not a sub-resource.
 func (r *Resource) isTopLevel() bool {
 	// TODO: This is a heuristic; replace with a definitive attribute?
-	return r.Create != nil
+	return (r.Create != nil || r.Read != nil)
```
radeksimko: I'm making the ugly worse, but this never worked for data sources 🙈 🤷‍♂️

PR #16487, opened 2017-10-27
init -from-module

jbardin:
Add GetModule for the cli to initialize from a regisry module source.

Storage.GetModule fetches a module using the same detection and
discovery as used by the normal module loading. The final copy is still
done by module.GetCopy to remove vcs files.


jbardin:
I'll the the e2e test in a separate PR.

PR #16509, opened 2017-10-31
Use pooled http client for fetching providers

jbardin:
While the TLS handshakes are a fairly small overhead compared to
downloading the providers, clients in some situation are failing to
complete the TLS handshake in a timely manner. It's unclear if this is
because of heavily constrained clients are stalling while doing the
major crypto operations, or the edge servers are throttling repeated
requests from the same IPs.

This should allow reusing the open TLS connection to the release edge
servers during init.


jbardin:
While not the root cause, this should certainly help mitigate #16448

PR #16511, opened 2017-10-31
Override expected subject of a certificate for WinRM

jd3nn1s:
Sometimes the hostname and IP of a resource is not known ahead of
time. This makes it hard to generate a valid certificate before a VM
has been created. Overriding the expected subject of a certificate
overcomes this.

Full use case - end-to-end secure deployment of an instance:
 * Generate a self-signed certificate via terraform.
 * Supply certificate securely to VM at creation-time via user_data
 * winrm provisioner configured with ca_cert set to generated cert.
   and tls_server_name set to cert subject.
 * Provisioner connects securely.

### DIFF
--- a/communicator/winrm/communicator.go
+++ b/communicator/winrm/communicator.go
@@ -35,10 +35,11 @@ func New(s *terraform.InstanceState) (*Communicator, error) {
 	}
 
 	endpoint := &winrm.Endpoint{
-		Host:     connInfo.Host,
-		Port:     connInfo.Port,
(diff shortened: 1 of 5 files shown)


teamterraform:
We are sorry that we let this PR sit for so long. Since the underlying code base has changed so much, and there are many conflicts, we are going to close this PR. 

If this is a feature you are still interested in, please open a Github issue using the Feature Request template so we can discuss the design.

Thank you!

PR #16565, opened 2017-11-05
helper/schema: Clear existing map/set/list contents before overwriting

vancluever:
There are situations where one may need to write to a set, list, or map
more than once per single TF operation (apply/refresh/etc). In these
cases, further writes using `Set` (example: `d.Set("some_set", newSet)`)
currently create unstable results in the set writer (the name of the
writer layer that holds the data set by these calls) because old keys
are not being cleared out first.

This bug is most visible when using sets. Example: First write to set
writes elements that have been hashed at 10 and 20, and the second write
writes elements that have been hashed at 30 and 40. While the set
length has been correctly set at 2, since a set is basically a map (as
is the entire map writer) and map results are non-deterministic, reads
to this set will now deliver unstable results in a random but
predictable fashion as the map results are delivered to the caller
non-deterministic - sometimes you may correctly get 30 and 40,
but sometimes you may get 10 and 20, or even 10 and 30, etc.

This problem propagates to state which is even more damaging as unstable
results are set to state where they become part of the permanent data
set going forward.

The problem also applies to lists and maps. This is probably more of an
issue with maps as a map can contain any key/value combination and hence
there is no predictable pattern where keys would be overwritten with
default or zero values. This is contrary to complex lists, which has
this problem as well, but since lists are deterministic and the length
of a list properly gets updated during the overwrite, the problem is
masked by the fact that a read will only read to the boundary of the
list, skipping any bad data that may still be available due past the
list boundary.

This update clears the child contents of any set, list, or map before
beginning a new write to address this issue. Tests are included for all
three data types.

### DIFF
--- a/helper/schema/field_writer_map.go
+++ b/helper/schema/field_writer_map.go
@@ -39,6 +39,19 @@ func (w *MapFieldWriter) unsafeWriteField(addr string, value string) {
 	w.result[addr] = value
 }
 
+// clearTree clears a field and any sub-fields of the given address out of the
+// map. This should be used to reset some kind of complex structures (namely
+// sets) before writing to make sure that any conflicting data is removed (for
+// example, if the set was previously written to the writer's layer).
+func (w *MapFieldWriter) clearTree(addr []string) {
+	prefix := strings.Join(addr, ".") + "."
+	for k := range w.result {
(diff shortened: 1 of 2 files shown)


vancluever:
Haha, thanks @jbardin - merging now

PR #16588, opened 2017-11-08
helper/schema: Opt-in panic on invalid ResourceData.Set

radeksimko:
We have seen a number of bugs in the past involving `Set("field_name", invalid data)` in the past. These are really hard to discover without explicit error checks.

It is not desired to add error checks to each `Set()` as it would make the CRUD unnecessarily verbose. Making `panic` default seems a bit too aggressive as it would affect user's experience negatively (crash).

This is a compromise which allows us to be strict where we can afford to be strict - e.g. when running acceptance tests for providers and keep the existing behaviour as is for users (i.e. avoid crashes at runtime).

I don't think it's necessary to document this - we can just make `TF_SCHEMA_PANIC_ON_ERROR=1` default in `make testacc` of all providers.

### DIFF
--- a/builtin/providers/test/resource.go
+++ b/builtin/providers/test/resource.go
@@ -49,7 +49,6 @@ func testResource() *schema.Resource {
 			"computed_read_only": {
 				Type:     schema.TypeString,
 				Computed: true,
-				ForceNew: true,
 			},
 			"computed_from_required": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


radeksimko:
> Perhaps we can do a manually-kicked test run to start and observe what happens so we can catch any cases where it's always panicking before we have our daily runs getting caught up in this.

That's exactly what I had in mind :)

PR #16663, opened 2017-11-15
Tweaks to Registry Docs for 0.11

banks:
Thanks @apparentlymart for updating the rest of these docs. I was reviewing it assuming I needed to document all the registry changes but you have already done almost all of it!

These are a few bits left a little stale/inconsistent post 0.11.

### DIFF
--- a/website/docs/registry/index.html.md
+++ b/website/docs/registry/index.html.md
@@ -25,7 +25,7 @@ module "consul" {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


banks:
FOSSA fail seems to be bogus. Possibly related to some sneaky force-push behaviour after I fluffed up making this PR via GitHub UI.

Merging anyway since diff is clearly not affecting licenses.

PR #16704, opened 2017-11-19
wait until salt-masterless commands are finished

subbarao:
current implementation doesn't wait until the salt download, install, apply commands are finished.
So current version is not working due to this. also piped output from this command to stdout.

### DIFF
--- a/builtin/provisioners/salt-masterless/resource_provisioner.go
+++ b/builtin/provisioners/salt-masterless/resource_provisioner.go
@@ -141,13 +141,16 @@ func applyFn(ctx context.Context) error {
 		if err = comm.Start(cmd); err != nil {
 			return fmt.Errorf("Unable to download Salt: %s", err)
 		}
+		cmd.Wait()
 		cmd = &remote.Cmd{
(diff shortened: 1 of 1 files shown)


subbarao:
screen shot 2017-11-19 at 4 01 37 pm

subbarao:
https://gist.github.com/subbarao/a226b4671f235690a3b1444b9749cbc4

marco-m:
Hello, any news? We currently cannot use our extensive salt masterless configuration with Terraform due to #16678, and this PR should fix it :-)

subbarao:
@apparentlymart current masterless provisioning is not working. what do we need to do get this merged to master.

subbarao:
https://github.com/hashicorp/terraform/pull/16142 is partial fix. this pull request waits for all calls to complete.

gapotts:
I can validate this pull request fixes the issue on GCP. I will be testing it as well shortly on AWS and will update once I have verified it there as well. JFYI.

apparentlymart:
Hi @subbarao! Sorry for the delay here. Things get a bit hectic around release time so we're a bit behind.

PR #16722, opened 2017-11-20
Clean up regsrc module

jbardin:
Some basic cleanup to do before refactoring the config/module source. 

The registry's requirements for FriendlyHost are slightly different that those of terraform itself. Rather than try and unify the concepts of FriendlyHost and svchost.Hostname, I'm going to leave the FriendlyHost api as-is, and add `SvcHost` method to return a svchost.Hostname. The only test checks now omitted or changes from the original set are where a punycode source was allowed even though it was not valid.


jbardin:
forgot to rewrite the imports

PR #16743, opened 2017-11-22
Output warnings flag

jbardin:
This adds an opt out env variable `TF_WARN_OUTPUT_ERRORS` which will make output errors non fatal during interpolation. The flag will also cause config validation to insert warnings about suspected sources of output errors du to variable count values.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ package config
 
 import (
 	"fmt"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -14,6 +15,9 @@ import (
 	"github.com/hashicorp/terraform/tfdiags"
 	"github.com/mitchellh/reflectwalk"
 )
+
+// TODO: Remove this in 0.12.
+var flagWarnOutputErrors = os.Getenv("TF_WARN_OUTPUT_ERRORS") != ""
 
 // NameRegexp is the regular expression that all names (modules, providers,
 // resources, etc.) must follow.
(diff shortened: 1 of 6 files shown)


jbardin:
The test fixture is failing until we decide of an explicit `count = 1` is worth warning on.
I'll add a core test to verify the flag behavior as well at that point.

nbering:
From a user's perspective, how would you explain the difference between this and `TF_LOG`? Is there a good reason for this to be a new feature in addition to the existing log levels feature?

I ask of my own interests, but also as a user that helps a number of other users with diagnostics.

jbardin:
@nbering, this is a proposed temporary opt-out flag for output errors. This is meant to be combined with #16735 to suppress errors during plan and apply, but provide warnings about configurations that may cause such errors.

nbering:
Are errors/warnings like this something the team is going to be investing a lot of time in? I'd be curious to know if the possibility of giving such internal errors an identifier and then providing a list of suppressed identifiers was discussed. Something like linting tools (ie, pylint, eslint), or C/C++ compiler tooling often provides.

I appreciate such ID's myself because as a developer it makes it easier to search for information about the error without the error message needing to include several paragraphs of explanation.

If it's just a one-off, this PR probably makes more sense, and my suggestion would be over-engineering.

jbardin:
This is all temporary for the 0.11 releases (see the linked PR and issue for background), and we don't intend to provide any general mechanisms for suppressing errors. This proposed opt-out flag is only to allow for users with configurations that cause output errors in 0.11 to have more time to update their configurations, and will all be removed in 0.12.

jbardin:
We've decided to keep the warnings on at all times. Closing this  to submit a new PR against master.

PR #16770, opened 2017-11-27
update tfe backend to recommend atlas_token

KFishner:
Recommend `ATLAS_TOKEN` instead of `access_token`

### DIFF
--- a/website/docs/backends/types/terraform-enterprise.html.md
+++ b/website/docs/backends/types/terraform-enterprise.html.md
@@ -27,7 +27,6 @@ Enterprise**. While this transition is in progress, you may see references to
 terraform {
   backend "atlas" {
     name         = "bigbang/example"
-    ATLAS_TOKEN  = "foo"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


website/docs/backends/types/terraform-enterprise.html.md
```diff
@@ -27,7 +27,7 @@ Enterprise**. While this transition is in progress, you may see references to
 terraform {
   backend "atlas" {
     name         = "bigbang/example"
-    access_token = "foo"
+    ATLAS_TOKEN  = "foo"
```
skierkowski: `ATLAS_TOKEN` is an environment variable, not a field in the backend configuration. So instead of being in this section, it needs to be set via `export` on the command line.
`export ATLAS_TOKEN="foo"`

KFishner: ahh

KFishner:
@skierkowski updated!

PR #16773, opened 2017-11-27
Registry client refactor

jbardin:
This is all just moving code around to get it out of the way for further work on the config/module package. The storage type was a bit overloaded, and the registry handling could be split out. 

This first moves the mock registry server into `github.com/hashicorp/terraform/registry/test`, so it can be used by both the registry client package and the module loading tests. 

Then we create a new `registry` package to isolate the code that interacts with the terraform registry API, and move all the registry specific config tests to the new package.


jbardin:
rebased in the change from #16846

PR #16774, opened 2017-11-28
helper/validation: Add StringMatch

vancluever:
`StringMatch` returns a validation function that can be used to match a
string against a regular expression. This can be used for simple
substring validations or more complex validation scenarios. Optionally,
an error message can be returned so that the user is returned a better
error message other than that their field did not match a regular
expression that they might not be able to understand.

### DIFF
--- a/helper/validation/validation.go
+++ b/helper/validation/validation.go
@@ -107,9 +107,8 @@ func StringLenBetween(min, max int) schema.SchemaValidateFunc {
 }
 
 // StringMatch returns a SchemaValidateFunc which tests if the provided value
-// matches a given regexp, which must compile or else the function will panic.
-// Optionally an error message can be provided to return something friendlier
-// than "must match some globby regexp".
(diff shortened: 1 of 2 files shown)


helper/validation/validation.go
```diff
@@ -106,6 +106,28 @@ func StringLenBetween(min, max int) schema.SchemaValidateFunc {
 	}
 }
 
+// StringMatch returns a SchemaValidateFunc which tests if the provided value
+// matches a given regexp, which must compile or else the function will panic.
```
jbardin: Looks like the implementation has changed, as it takes a compiled Regexp now.

vancluever: Right my bad! Will update this shortly

vancluever:
Hey @jbardin finally got around to fixing this! Comment should reflect implementation now

vancluever:
Thanks @jbardin!

PR #16806, opened 2017-12-01
Support parallel testing in terraform testing framework:

metacpp:
1. Add Parallel() in the TestT interface so that testing.T.Parallel can be invoked in each test function.
2. Introduced the TF_PARA for user to enable it while running "go test".
3. Add unit test case.

### DIFF
--- a/backend/local/counthookaction_string.go
+++ b/backend/local/counthookaction_string.go
@@ -2,7 +2,7 @@
 
 package local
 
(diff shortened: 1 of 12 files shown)


metacpp:
@tombuildsstuff

metacpp:
Please ignore this PR, it seems that go fmt introduced some extra code changes.

PR #16816, opened 2017-12-01
Make plan command cancellable 

jbardin:
This adds cancellation support to the plan command. While plans are usually fairly uneventful, it can be useful to be able to cancel the plan during operation.

We also fix up the apply test to make it not rely on a sleep for verification. 

Also fixes #16652 which demonstrated the lack of cancellation waiting for a lock during plan.

PR #16833, opened 2017-12-04
Fully enable shutdown for plan and refresh in the local backend

jbardin:
The backends need to specifically watch for and handle context cancellation because the context received for the operation is disjoint from the context used in the RunningOperation. This was only implemented in the apply operation, so the other operations couldn't be fully cancelled.


jbardin:
This as an addendum to #16816, which enabled the cancellation in the commands, but expected the backend cancellation to work.

jbardin:
I agree. Trying to DRY those though will take some work on better abstractions, singe all three operations operate on slightly different types, with slightly different UI output, and different checks after.

PR #16846, opened 2017-12-05
Update registry API docs with browse and search

banks:
This adds documentation of the public API endpoints for browsing and searching modules.

These are not actually new, we just released with a minimal set of endpoints in our "officially supported" API that were sufficient for terraform CLI to use the registry.

Now we have experience with a few different consumers, these endpoints are important for basic discovery and, for example, are used in Terraform Enterprise, so warrant standardising.

I tweaked a test here to verify that discovery code actually works regardless of whether the discovery document has a trailing slash. It does thanks to `url.ResolveReference`. we can change it back if there is no value in it - I was in two minds about whether to leave it anyway. We _could_ change the actual discovery documents to have no slash and it would still work and might be marginally less confusing, but in practice it will work with no change too.

I ended up not actually making any changes to either the registry API (to serve from `/v1/modules/` rather than 301 to slashless path) since I think these docs are clear enough like this and it.

### DIFF
--- a/config/module/module_test.go
+++ b/config/module/module_test.go
@@ -54,6 +54,8 @@ func testStorage(t *testing.T, d *disco.Disco) *Storage {
 // and example.com to the test server.
 func testDisco(s *httptest.Server) *disco.Disco {
 	services := map[string]interface{}{
+		// Note that both with and without trailing slashes are supported behaviours
+		// TODO: add specific tests to enumerate both possibilities.
 		"modules.v1": fmt.Sprintf("%s/v1/modules", s.URL),
 	}
 	d := disco.NewDisco()
--- a/config/module/module_test.go
+++ b/config/module/module_test.go
@@ -54,7 +54,7 @@ func testStorage(t *testing.T, d *disco.Disco) *Storage {
 // and example.com to the test server.
 func testDisco(s *httptest.Server) *disco.Disco {
 	services := map[string]interface{}{
-		"modules.v1": fmt.Sprintf("%s/v1/modules/", s.URL),
+		"modules.v1": fmt.Sprintf("%s/v1/modules", s.URL),
 	}
 	d := disco.NewDisco()
 
--- a/website/docs/registry/api.html.md
+++ b/website/docs/registry/api.html.md
@@ -116,7 +116,7 @@ $ curl 'https://registry.terraform.io/v1/modules&limit=2&verified=true'
 
 ## Search Modules
 
-Theis endpoint allows searching modules.
+This endpoint allows searching modules.
 
 | Method | Path                                  | Produces                   |
 | ------ | ------------------------------------- | -------------------------- |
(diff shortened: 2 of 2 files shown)


config/module/module_test.go
```diff
@@ -54,7 +54,7 @@ func testStorage(t *testing.T, d *disco.Disco) *Storage {
 // and example.com to the test server.
 func testDisco(s *httptest.Server) *disco.Disco {
 	services := map[string]interface{}{
-		"modules.v1": fmt.Sprintf("%s/v1/modules/", s.URL),
+		"modules.v1": fmt.Sprintf("%s/v1/modules", s.URL),
```
banks: Tests still pas with this change indicating the code is fully tolerant of slashless-discovery urls.

As mentioned in description I'm in two minds about leaving this in or reverting.

banks:
@apparentlymart Good call, I added a comment to the test change to at least act as a pointer to future devs.

@mitchellh @skierkowski If you have any feedback on the content let me know, I'll wait a bit to merge this.

website/docs/registry/api.html.md
```diff
@@ -36,9 +36,150 @@ For example, if discovery produces the URL `https://modules.example.com/v1/`
...
+the [discovery document](#service-discovery) _may include_ a trailing slash, we
+include a slash after the placeholder in the `Path`s below for clarity.
+
+## List Modules
+
+These endpoints list modules according to some criteria.
+
+| Method | Path                                  | Produces                   |
+| ------ | ------------------------------------- | -------------------------- |
+| `GET`  | `<base_url>` (no trailing slash)      | `application/json`         |
```
mitchellh: Is the `(no trailing slash)` necessary? It would work with both, right?

I fear this will make things more confusing than not.

banks: If you do have a slash we will 301 redirect so it does work for some value. But it feels odd to me that we would officially document a path that is a 301 redirect. This was the most succinct way I could find to not document the wrong thing without writing an essay about it and highlighting the confusion!

I considered changing the API to always have the slash but I think that just breaks conventions within the API itself for the sake of this one thing.

I don't mind removing this and people can just figure out for themselves to strip the slash if they want to avoid a pointless redirect but at least some clients wouldn't follow 301 by default (for example curl) so it looks kinda broken to document it as being the canonical endpoint?If you do have a slash we will 301 redirect so it does work for some value. But it feels odd to me that we would officially document a path that is a 301 redirect...

I considered changing the API to always have the slash but I think that just breaks conventions within the API itself for the sake of this one thing.

I don't mind removing this and people can just figure out for themselves to strip the slash if they want to avoid a pointless redirect but at least some clients wouldn't follow 301 by default (for example curl) so it looks kinda broken to document it as being the canonical endpoint?

mitchellh: Right, my suggestion was maybe just to remove the `(no trailing slash)` text but continuing to document it as `<base_url>`. If you consider this a specification, its still exactly correct, we never say a trailing slash works, but we may avoid some confusion around the phrase.

I may be sensitive/pedantic here though around seeing that phrase and thinking "but why though"

banks: Oh and the other part here is that in fact the `base_url` came from discovery document and it _may_ have a trailing slash already - our public reg one does now although it would work to remove it.

Issue #16803, opened 2017-11-30
GCS backend GOOGLE_CREDENTIALS is a path or a JSON in terraform implementation


jp:
Hello,

I'm using terraform v0.11.0 and I'm using gcs for the state backend.

When I set a JSON in GOOGLE_CREDENTIALS I end with the following error:
```
terraform plan
Failed to load backend: 
Error configuring the backend "gcs": storage.NewClient() failed: dialing: cannot read credentials file: open { "type": "service_account", "project_id": "[REDACTED]", "private_key_id": "[REDACTED]", "private_key": "-----BEGIN PRIVATE KEY-----\n[REDACTED]\n-----END PRIVATE KEY-----\n", "client_email": "[REDACTED]", "client_id": "[REDACTED]", "auth_uri": "https://accounts.google.com/o/oauth2/auth", "token_uri": "https://accounts.google.com/o/oauth2/token", "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs", "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/[REDACTED]" }: file name too long

Please update the configuration in your Terraform files to fix this error.
If you'd like to update the configuration interactively without storing
the values in your configuration, run "terraform init".
```

When I set GOOGLE_CREDENTIALS to the path of a file containing the credential JSON I have the following error:

```
terraform plan

Error: provider.google: credentials are not valid JSON '/gcp_credentials.json': invalid character '/' looking for beginning of value
```

I'm reproducing this error with the following main.tf:
```
terraform {
  backend "gcs" {
    path="state.tfstate"
  }
}

provider "google" {
  project     = "${var.gcp_project_name}"
  region      = "${var.gcp_region_name}"
  version = "= 1.1.1"
}
```

It looks like this code is loading credentials as a JSON: https://github.com/hashicorp/terraform/blob/v0.11.0/state/remote/gcs.go#L66

This one is loading credentials as a path: https://github.com/hashicorp/terraform/blob/v0.11.0/backend/remote-state/gcs/backend.go#L120

According to this file, default should be "Local path to Google Cloud Platform account credentials in JSON format." : https://github.com/hashicorp/terraform/blob/55089e472d579acd2c539e1720a96c389d546fa9/website/docs/backends/types/gcs.html.md#L53

jp:
I managed to go around the issue by:

[...]

PR #16865, opened 2017-12-06
gcs provider fixes

jbardin:
- The backend was creating the bucket named in the configuration if it didn't exist. We don't allow other backends to do this, because these are not managed resources that terraform can control.
- Have the test fixtures create their own buckets, so parallel tests don't collide, and leftover buckets don't cause future failures. This also properly removes test buckets regardless of contents.
- Add a test that fails the condition that #16852 was fixing. This is simply looking up a locked remote state, since a backend may need to take a lock to create a non-existent state. Reading an existing locked remote state should succeed.
- Match the operation of the google provider, by accepting either a file path or contents for both `credentials` and `GOOGLE_CREDENTIALS`

### DIFF
--- a/backend/remote-state/gcs/backend.go
+++ b/backend/remote-state/gcs/backend.go
@@ -128,22 +128,5 @@ func (b *gcsBackend) configure(ctx context.Context) error {
 
 	b.storageClient = client
 
-	return b.ensureBucketExists()
-}
-
-func (b *gcsBackend) ensureBucketExists() error {
-	_, err := b.storageClient.Bucket(b.bucketName).Attrs(b.storageContext)
-	if err != storage.ErrBucketNotExist {
-		return err
-	}
-
-	if b.projectID == "" {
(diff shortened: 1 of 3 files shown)


jbardin:
```
$ TF_ACC=1 go test -v 2>/dev/null
=== RUN   TestStateFile
=== RUN   TestRemoteClient
=== RUN   TestRemoteLocks
=== RUN   TestBackend
=== RUN   TestBackendWithPrefix
--- PASS: TestStateFile (0.00s)
--- PASS: TestRemoteClient (6.73s)
--- PASS: TestRemoteLocks (8.95s)
--- PASS: TestBackend (18.68s)
	backend_test.go:114: TestBackend: testing state locking for *gcs.gcsBackend
--- PASS: TestBackendWithPrefix (21.86s)
	backend_test.go:127: TestBackend: testing state locking for *gcs.gcsBackend
PASS
```

jbardin:
This also fixes #16803

PR #16882, opened 2017-12-08
[Add] Pathing and Override to DEBUG Command

Techcadia:
This is an improvement for this issue. https://github.com/hashicorp/terraform/issues/16874

 simple improvement that adds the output of overridden providers from the `.terraformrc` file as output in debug mode allowing you to see that they exist.


jbardin:
LGTM, thanks!

PR #16892, opened 2017-12-11
vendor: github.com/hashicorp/go-version

offlinehacker:
We required updated version of `go-version`, because we have deployment of mariadb that has `~` in version name: https://github.com/hashicorp/go-version/commit/43575a7136563760458ce6756b50ac0ab7245adb


jbardin:
Thanks @offlinehacker!

PR #16895, opened 2017-12-11
helper/resource: Fail tests with no error that have ExpectError set

vancluever:
Looks like while we were checking errors correctly when ExpectError was
set, we weren't checking for the *absence* of an error, which is should
be checked as well (no error is still not the error we are looking for).

Added a few more tests for ExpectError as well.


vancluever:
Thanks @apparentlymart!

PR #16903, opened 2017-12-13
Adds "build_auth_token" option to Habitat Provisioner

rwcee:
**Context:** Per @bixu's comment (https://github.com/hashicorp/terraform/pull/16280#issuecomment-350208983) on the original Habitat Provisioner PR, @nsdavidson and I moved forward with adding the ability to use private Habitat origins that require `HAB_AUTH_TOKEN`.

**This PR:** 
* Adds the `build_auth_token` option in terraform and necessary functionality through which a user's `HAB_AUTH_TOKEN` can be passed.
* Updates the docs

### DIFF
--- a/builtin/provisioners/habitat/resource_provisioner.go
+++ b/builtin/provisioners/habitat/resource_provisioner.go
@@ -1,6 +1,7 @@
 package habitat
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -9,6 +10,7 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/hashicorp/terraform/communicator"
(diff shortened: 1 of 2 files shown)


nsdavidson:
I have built this locally and verified that it works as expected. As mentioned, this was requested on original Habitat provisioner PR, and will be a must have for folks using private origins on the build service (https://bldr.habitat.sh).

jbardin:
Thanks @rwcee and @nsdavidson!

bixu:
@jbardin, how can we track that this has been released in an official Terraform build?

jbardin:
Hi @bixu,

With few exceptions, public releases are from the master branch, so the next tag and release will contain these changes. We don't have a set schedule, but it tends to be around a 2 week cycle, give or take depending on the changes made (which means we're about due for a release).

PR #16913, opened 2017-12-14
cmd/import|refresh: take the resource config into account

svanharmelen:
Some resources use `d.GetOk()` in their `Read()` function to determine if a value should be set to the state (only update/set the value when the user configured the value).

Without passing the resource config these resources will not be imported or refreshed properly.


svanharmelen:
Whoops! I was a bit too fast with this one, so will close and fix it before opening a new PR for this :wink:

PR #16931, opened 2017-12-15
docs: Clarify the use of private github repos with TFE

nfagerlund:
The current version of TFE can use centrally-managed SSH keys to access modules from git repos, which is better than embedded machine user credentials for most use cases. 

Also, as I understand it, any other CI-like systems can use the same mechanism TFE uses for accessing private Git-sourced modules, with the `GIT_SSH_COMMAND` env var, which might be preferable depending on your worker provisioning process.

### DIFF
--- a/website/docs/modules/sources.html.markdown
+++ b/website/docs/modules/sources.html.markdown
@@ -97,17 +97,19 @@ You can use the same parameters to GitHub repositories as you can generic Git re
 
 ### Private GitHub Repos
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
Oh, and PR that adds the /docs/enterprise/workspaces/ssh-keys.html page I link here has been merged to terraform-website, but it might not be deployed yet.

skierkowski:
LGTM

KFishner:
LGTM! Merge away

nfagerlund:
🙌🏼 

But alas, I don't have write permissions. Could the first person with merge power to wander by please push the button?

Issue #16383, opened 2017-10-18
workspace_key_prefix contains more than one folder


dpavlov-smartling:
Hello, 

Seems like terraform isn't able to handle **workspace_key_prefix** which contains more than one word like:
**workspace_key_prefix = "rds-test/rds-logs-download"**
Actually it is able to create workspace and put terraform state file into needed path, but doesn't show it when you execute:
terraform workspace list

Initial task is to put state files of all workspaces inside sub-folder into main bucket like /BUCKET/SUB-folder1/SUB-folder2/WORKSPACE-NAME/terraform.tfstate, so I have used following backend configuratin:
```
terraform {
  backend "s3" {
    bucket               = "XXXXXX-terraform"
    key                  = "rds-logs-download.tfstate"
    region               = "us-east-1"
    workspace_key_prefix = "rds-test/rds-logs-download" 
  }
}

```
### Expected Behavior
According to documentation at https://www.terraform.io/docs/backends/types/s3.html#key path should be following:
XXXXXX-terraform/rds-test/rds-logs-download/<WORKSPACE_NAME>/rds-logs-download.tfstate
### Actual Behavior
```
terraform workspace new test2
Created and switched to workspace "test2"!

terraform workspace list
  default

terraform workspace new test3
Created and switched to workspace "test3"!

terraform workspace list
  default
```

While it DOES create structure inside S3 backet:
aws s3 ls XXXXXXXXX-terraform/rds-test/rds-logs-download/test
                           PRE test2/
                           PRE test3/

### Steps to Reproduce
Just try to use backend as was provided above and create few workspaces.

terraform version
Terraform v0.10.7

completenovice:
This is affecting us too, it is stopping us rolling out workspaces to our teams as it is not clear on which workspace a user is on.

xatnys:
Also affecting us.

mbowiewilson:
I'm still having this problem with v0.11.3

redbaron:
@mbowiewilson , @jbardin I can confirm that with TF 0.11.3 this is still not solved. `terraform workspace select <name>` doesn't find  workspace

jbardin:
@mbowiewilson: If you can a reproducible case, please feel free to open a new about and we can look into it. Thanks!

redbaron:
@jbardin IMHO it is better to have this issue reopened as problem described here is not solved. My reproducible test case is like for like with one reported here , with slight variation to actual values used:

```
terraform {
  backend "s3" {
[...]

Issue #16387, opened 2017-10-18
Empty string in workspace_key_prefix breaks workspace functionality


hackdna:
### Terraform Version
```
$ terraform -v
Terraform v0.10.7
```

### Terraform Configuration Files
```hcl
terraform {
  required_version = "~> 0.10"

  backend "s3" {
    key                  = "storage/terraform.tfstate"
    workspace_key_prefix = ""
  }
}
```

### Expected Behavior
Non-default workspaces should be usable.

### Actual Behavior
```
Workspace "test-workspace" doesn't exist.

You can create this workspace with the "new" subcommand.
```

### Steps to Reproduce
1. `terraform init`
1. `terraform workspace new test-workspace`
1. `terraform workspace select default`
1. `terraform workspace select test-workspace`

### References
May be related to #16383

sunil592002:
Hi,

I am also facing the same issue. Is this bug will be fixed soon, to make us of workspace in PROD

jbardin:
closed via #16932

PR #16932, opened 2017-12-15
Fixing issues with workspace_key_prefix for s3 backend

rv-jmaggio:
I believe this will fix https://github.com/hashicorp/terraform/issues/16383 and https://github.com/hashicorp/terraform/issues/16387

### DIFF
--- a/backend/remote-state/s3/backend_state.go
+++ b/backend/remote-state/s3/backend_state.go
@@ -58,7 +58,7 @@ func getWorkspaceForKey(key string, b *Backend) string {
 		return ""
 	}
 
-	parts = strings.Split(parts[1], "/")
+	parts = strings.SplitN(parts[1], "/", 3)
 
 	if len(parts) < 3 {
 		return ""
--- a/backend/remote-state/s3/backend_state.go
+++ b/backend/remote-state/s3/backend_state.go
@@ -15,9 +15,15 @@ import (
 )
 
 func (b *Backend) States() ([]string, error) {
+	prefix := b.workspaceKeyPrefix + "/"
+
+	// List bucket root if there is no workspaceKeyPrefix
+	if b.workspaceKeyPrefix == "" {
+		prefix = ""
+	}
 	params := &s3.ListObjectsInput{
 		Bucket: &b.bucketName,
-		Prefix: aws.String(b.workspaceKeyPrefix + "/"),
+		Prefix: aws.String(prefix),
 	}
 
 	resp, err := b.s3Client.ListObjects(params)
@@ -27,7 +33,7 @@ func (b *Backend) States() ([]string, error) {
 
 	wss := []string{backend.DefaultStateName}
 	for _, obj := range resp.Contents {
-		ws := getWorkspaceForKey(*obj.Key, b)
+		ws := b.keyEnv(*obj.Key)
 		if ws != "" {
 			wss = append(wss, ws)
 		}
@@ -37,7 +43,7 @@ func (b *Backend) States() ([]string, error) {
 	return wss, nil
 }
 
-func getWorkspaceForKey(key string, b *Backend) string {
+func (b *Backend) keyEnv(key string) string {
 	if b.workspaceKeyPrefix == "" {
 		parts := strings.SplitN(key, "/", 2)
 		if len(parts) > 1 && parts[1] == b.keyName {
@@ -190,7 +196,12 @@ func (b *Backend) path(name string) string {
 		return b.keyName
 	}
 
-	return strings.Join([]string{b.workspaceKeyPrefix, name, b.keyName}, "/")
+	if b.workspaceKeyPrefix != "" {
+		return strings.Join([]string{b.workspaceKeyPrefix, name, b.keyName}, "/")
+	} else {
+		// Trim the leading / for no workspace prefix
+		return strings.Join([]string{b.workspaceKeyPrefix, name, b.keyName}, "/")[1:]
+	}
 }
 
 const errStateUnlock = `
--- a/backend/remote-state/s3/backend_state.go
+++ b/backend/remote-state/s3/backend_state.go
@@ -39,8 +39,8 @@ func (b *Backend) States() ([]string, error) {
 
 func getWorkspaceForKey(key string, b *Backend) string {
 	if b.workspaceKeyPrefix == "" {
-		parts := strings.Split(key, "/")
-		if len(parts) > 1 && parts[1] == key {
+		parts := strings.SplitN(key, "/", 2)
+		if len(parts) > 1 && parts[1] == b.keyName {
 			return parts[0]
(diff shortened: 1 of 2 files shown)


rv-jmaggio:
@jbardin @apparentlymart 
Ran into this bug today and I think I've got a fix here. I added some tests for this functionality and cleaned up a bit. However, I haven't manually tested this. I'm hoping this could be released in a minor provider version.

backend/remote-state/s3/backend_state.go
```diff
@@ -51,6 +58,12 @@ func (b *Backend) keyEnv(key string) string {
 		return ""
 	}
 
+	parts = strings.Split(parts[1], "/")
```
jbardin: If the key names are allowed to have `/` characters, you're going to want `SplitN` with 3 here.

backend/remote-state/s3/backend_test.go
```diff
@@ -249,6 +249,62 @@ func TestBackendExtraPaths(t *testing.T) {
...
+
+	if err := testGetWorkspaceForKey(b0, "tfstate", ""); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := testGetWorkspaceForKey(b0, "ws1/tfstate", "ws1"); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := testGetWorkspaceForKey(b1, "root/userA/ws1/tfstate", "ws1"); err != nil {
```
jbardin: If the prefix is `root/userA` and the key is `ws1/tfstate`, `ws1` can't be the workspace name.

backend/remote-state/s3/backend_test.go
```diff
@@ -249,6 +249,62 @@ func TestBackendExtraPaths(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if err := testGetWorkspaceForKey(b1, "root/userA/ws2/tfstate", "ws2"); err != nil {
```
jbardin: Same as above for this check and the one below. The tests are checking the output of `getWorkspaceForKey`, but that doesn't match how it's used by the backend.

backend/remote-state/s3/backend_test.go
```diff
@@ -249,6 +249,62 @@ func TestBackendExtraPaths(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if err := testGetWorkspaceForKey(b2, "root/userA/ws2/env1/tfstate", "ws2"); err != nil {
+		t.Fatal(err)
+	}
```
jbardin: Once of the new backend configs needs to be run through the TestBackend helper, which would have failed with either `b1` or `b2`.

rv-jmaggio:
@jbardin I updated from your feedback. I did some testing in a playground and you are definitely correct on SplitN. Evidently I was a bit confused when I wrote my test cases, so I made them a bit more clear. I think this should be good to go now. I am happy to contribute to a tool I use nearly every day 😃

backend/remote-state/s3/backend_test.go
```diff
@@ -249,6 +249,63 @@ func TestBackendExtraPaths(t *testing.T) {
 	}
 }
 
+func TestKeyEnv(t *testing.T) {
```
jbardin: Though the rest of the ACC tests pass again, this test still fails.

backend/remote-state/s3/backend_test.go
```diff
@@ -249,6 +249,63 @@ func TestBackendExtraPaths(t *testing.T) {
...
+	}
+
+	if err := testGetWorkspaceForKey(b1, "project/env:/ws2/tfstate", "ws2"); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := testGetWorkspaceForKey(b2, "env:/ws3/tfstate", "ws3"); err != nil {
+		t.Fatal(err)
+	}
+
```
jbardin: You're not creating this bucket for the backend tests, though it currently fails earlier.

backend/remote-state/s3/backend_test.go
```diff
@@ -249,6 +249,63 @@ func TestBackendExtraPaths(t *testing.T) {
 	}
 }
 
+func TestKeyEnv(t *testing.T) {
+	testACC(t)
+	bucketName := fmt.Sprintf("terraform-remote-s3-test-%x", time.Now().Unix())
+	keyName := "tfstate"
```
jbardin: These test cases are good, but let's go back to having a keyName with a `/` to make sure it's still supported after we introduce them into the prefix.

backend/remote-state/s3/backend_state.go
```diff
@@ -25,24 +25,31 @@ func (b *Backend) States() ([]string, error) {
...
-// extract the env name from the S3 key
-func (b *Backend) keyEnv(key string) string {
-	// we have 3 parts, the prefix, the env name, and the key name
-	parts := strings.SplitN(key, "/", 3)
-	if len(parts) < 3 {
-		// no env here
+func getWorkspaceForKey(key string, b *Backend) string {
+	if b.workspaceKeyPrefix == "" {
+		parts := strings.Split(key, "/")
+		if len(parts) > 1 && parts[1] == key {
```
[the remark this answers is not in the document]
rv-jmaggio: Ah yes the intention was to check b.keyName, not key. Thanks

rv-jmaggio:
@jbardin 
Ah, thanks for the pointer on `TF_ACC=1`.

So I made those changes, but I was getting an error when running `backend.TestBackend(t, b0, nil)`
```
--- FAIL: TestKeyEnv (23.36s)
	backend_test.go:339: creating S3 bucket terraform-remote-s3-test-5a39211b-0 in us-west-2
	backend_test.go:339: creating S3 bucket terraform-remote-s3-test-5a39211e-1 in us-west-2
	backend_test.go:339: creating S3 bucket terraform-remote-s3-test-5a392121-2 in us-west-2
	backend_test.go:308: bad: []string{"default"}
FAIL```

It seems like when workspace_key_prefix is set to `""` then the workspace folder `default` isn't being created? Not exactly sure what to make of this.

rv-jmaggio:
Ah actually it seems to be related to the implementation for empty string workspace_key_prefix. I think I can get that fixed.

rv-jmaggio:
I think there may be an issue with this implementation for a workspace_key_prefix set to `""`

https://github.com/hashicorp/terraform/blob/master/backend/remote-state/s3/backend_state.go#L180

rv-jmaggio:
@jbardin I had to make 2 more changes to the backend_state implementation to allow for empty string prefix. I ran the acceptance tests (actually in acceptance test mode) and they all passed.

jbardin:
Hi @rv-jmaggio,

Thanks for the update!
This is looking good. I'll do a final review after we return from the break and aim to get this into the next release.

PR #16936, opened 2017-12-18
Support 'customer supplied encryption keys' in the GCS backend

negz:
Currently all files in GCS are transparently encrypted at rest using Google's keys, and IAM/ACL can be used to restrict access to GCS buckets. Allowing state files in GCS to be encrypted with a custom key provides an additional layer of defense around sensitive Terraform state.

Objects created with a customer supplied encryption key can only be read when the key used for creation is passed in with the read request. See https://github.com/hashicorp/terraform/issues/16836 for further details.

### DIFF
--- a/backend/remote-state/gcs/backend.go
+++ b/backend/remote-state/gcs/backend.go
@@ -3,6 +3,7 @@ package gcs
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -29,6 +30,8 @@ type gcsBackend struct {
 	bucketName       string
 	prefix           string
 	defaultStateFile string
+
+	encryptionKey []byte
 
 	projectID string
 	region    string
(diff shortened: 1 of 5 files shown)


negz:
Thanks @paddycarver !

ocervell:
Does this support customer-managed keys for use with KMS ?

negz:
@ocervell I made this change almost two years ago and no longer actively use Terraform, so I'm afraid I'm not up to speed with the interplay between customer supplied GCS keys and KMS. I recall thinking that it would be nice to store customer managed keys in KMS, but we decided this first implementation would be simpler. There's some context on that decision process in https://github.com/hashicorp/terraform/issues/16836.

negz:
> Does this support customer-managed keys for use with KMS

@ocervell  Ah - I see customer _managed_ keys are distinct from customer _supplied_ keys, so no that is not supported by this change. I don't think managed keys existed when I wrote it, but it seems like it would be a great addition now.

PR #16948, opened 2017-12-18
Allow mixing and matching of password and key auth for bastions/target hosts

rylarson:
The way that Terraform currently works, if you use a key for the target host, you can't configure password auth for the bastion host because if the `bastion_private_key` is unset, it is defaulted to the `private_key` from the target host config. The ssh connection then tries to use that key to connect to the bastion, which fails.

This changes Terraform's behavior to default the `bastion_private_key` to the `private_key` only if the `bastion_password` is also being defaulted. If the `bastion_password` is explicitly set, then terraform will not default the `bastion_private_key` to the target hosts `private key`.

### DIFF
--- a/communicator/ssh/provisioner.go
+++ b/communicator/ssh/provisioner.go
@@ -108,9 +108,12 @@ func parseConnectionInfo(s *terraform.InstanceState) (*connectionInfo, error) {
 		}
 		if connInfo.BastionPassword == "" {
 			connInfo.BastionPassword = connInfo.Password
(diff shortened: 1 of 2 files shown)


CloneOfAlex (approved):

crw:
Due to this being 7 years old, CLA not signed, and the code being (understandably) out-of-date, I am going to close this PR. If you think this is in error, please leave a comment on the thread and I will re-open. Thanks!

PR #16949, opened 2017-12-19
added .bash_login and .profile which are also valid bash profile files.

costastf:
Fixes the issue where if .profile is used and not any of the other files the -install-autocomplete fails with :
Error executing CLI: Did not found any shells to install

excerpt from bash man page

    When bash is invoked as an interactive login shell, or as a non-interactive shell with the --login option, it first reads and executes commands from the file /etc/profile, if that file exists. After reading that file, it looks for ~/.bash_profile, ~/.bash_login, and ~/.profile, in that order, and reads and executes commands from the first one that exists and is readable. The --noprofile option may be used when the shell is started to inhibit this behavior.

    When an interactive shell that is not a login shell is started, bash reads and executes commands from /etc/bash.bashrc and ~/.bashrc, if these files exist. This may be inhibited by using the --norc option. The --rcfile file option will force bash to read and execute commands from file instead of /etc/bash.bashrc and ~/.bashrc.

### DIFF
--- a/vendor/github.com/posener/complete/cmd/install/install.go
+++ b/vendor/github.com/posener/complete/cmd/install/install.go
@@ -5,8 +5,6 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
-
-	"github.com/hashicorp/go-multierror"
 )
 
 type installer interface {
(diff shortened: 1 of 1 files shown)


costastf:
will fix in vendored repo instead.

PR #16957, opened 2017-12-19
Restore ability for an empty string to be considered unset in a computed value

jbardin:
Add a test case for this situation where a resource was using an empty string to indicate an unset computed value. This may have been too common in providers, so we need to evaluate reverting that behavior. 

Reverts one of the changes from 6a4f7b0 (#14887), which broke empty strings
being seen as unset for computed values (this breaks a number of newer tests).

Referencing https://github.com/terraform-providers/terraform-provider-aws/issues/2348

### DIFF
--- a/helper/schema/resource_data.go
+++ b/helper/schema/resource_data.go
@@ -445,7 +445,7 @@ func (d *ResourceData) init() {
 }
 
 func (d *ResourceData) diffChange(
-	k string) (interface{}, interface{}, bool, bool) {
+	k string) (interface{}, interface{}, bool, bool, bool) {
 	// Get the change between the state and the config.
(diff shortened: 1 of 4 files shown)


vancluever:
Added an extension to some of the internal diff methods that allow the old functionality to co-exist with the new.

This is possible in a relatively simple way as `ResourceDiff`'s internal `getChange` method already does an explicit check to see if diff customization has touched an inparticular value - we just bubble this up to `diffChange` now, and alter the signature for said method to report back if that diff value has been customized or not (`ResourceData`'s implementation just always returns `false`).

This ultimately allows us to selectively apply the diff exemption only to values that haven't been explicitly touched during diff customization.

PR #16961, opened 2017-12-20
minor race issue in mockResourceProvider

jbardin:
The interrupt tests for providers no longer check for the condition
during the diff operation. defer the lock so other test's DiffFns don't
need to be as carefull locking themselves.


jbardin:
The command package still does this apparently. I'll see if there's a better way than manually adding locks to all test DiffFns.

PR #16972, opened 2017-12-21
ssh connection `agent_identity`

### LINKED ISSUE #16909: Terraform fails to use SSH agent on Mac OS X
### Terraform Version

```
Terraform v0.11.1
+ provider.docker v0.1.1
```

```hcl
locals {
  # If using Docker for Mac
  host = "127.0.0.1"
  # If using Docker on Linux
  #host = "${docker_container.test.ip_address}"
  port = "2222"
  pubkeys = ["${file("id_rsa.pub")}", "${file("id_rsa_encrypted.pub")}"]
}

resource "docker_container" "test" {
  # Other declarations […]

  # Add SSH public keys to authorized keys
  upload {
    file = "/root/.ssh/authorized_keys"
    content = "${join("", local.pubkeys)}"
  }

  provisioner "remote-exec" {
    connection {
      type  = "ssh"
      host = "${local.host}"
      port = "${local.port}"
      user  = "root"
      agent = "true"
    }

    inline = [
      "echo 'Hello world!' > /root/readme.txt"
    ]
  }
}
```

For completeness, the output of `ssh-add -L`:
```text
$ ssh-add -L
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDDMmGshVJkhWskSnk2v2KvPRXh53Yqh7C7pcbEWywnTDz1U92a7GbV8nwTcGSXQPP918qvsb/8LRR6gYSqD4wtUaFSJTmk6FOguJiBCUh7Pt5fNVJZOPkjC3LOeO69aoB9x7IhesSxTDtCInusUlWKXoCbvL9GFC2MUKMcK2Yw3zoY/M8rJJyw5GfJ+X1OXA0Wd2StlP/Enwl+tUeb06ZNMPG37Tx6/Puh+sy7y5uy6i8cOMxkK7yAYOL0gyI61+J+ebY58pwdP3hPO2PhKjR/xnn4Ha4i8VXlRUE+kTNPCI69KxbA9H064pMcyLN5zHepGyeHZP14EptBBUjWWcBz id_rsa
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDToUTwvcvpZ3AzmmKA1SG4Gn/0kIQpWMGK6BJLv35XApTE7cSAPLgVGTZKlSi8BgfPuhfcHLaS/dpqpiZ0aJi2hMPK9gGQGghHHEi3i24ouo25gBUr5Xd88YhsPBt5HTytulvg+C13d6o5+e0P+01DRZh2x9R4h/Lbv1SQF2t2ZOhY4/YD18zwh3EBMHQ7uYYy/28zxmTDCOwb3qZ+K9DmZPpvIui4EsL/wv72cv1C0C+InqsU0Ps3zhUbmBl7/QsHiisezkCcwJ4haH98vgIUjp8sa9uTT8crBMAnYT4VE5PF6iiBEhUA2a4h/SBLhNRRxUa3wPRvktX3OP0C+bWN id_rsa_encrypted
```

### Debug Output

https://gist.github.com/joschi/b031fe7d37119bc7ddc2a719219b9717#file-terraform-log

### Expected Behavior

The `remote-exec` provisioner should be able to log into the created Docker container via SSH with the keys provided by the running (and otherwise working) SSH agent.

### Actual Behavior

The `remote-exec` provisioner fails to log into the Docker container until Terraform times out and aborts.

Manually logging into the Docker container (`ssh -l root -p 2222 -i id_rsa_encrypted 127.0.0.1`) works.
Deactivating the use of the SSH agent (`remote-exec.connection.agent = "false"`) and explicitly using a private key instead ( `remote-exec.connection.private_key = "${file("/path/to/id_rsa")}"`) also works (but doesn't support encrypted private keys and needs to know the location of the private key).

### Steps to Reproduce

1. Start Docker daemon
1. `ssh-add id_rsa id_rsa_encrypted`
1. `terraform init`
1. `terraform apply`

Basically see https://gist.github.com/joschi/b031fe7d37119bc7ddc2a719219b9717#file-readme-md

### References

* https://gist.github.com/joschi/b031fe7d37119bc7ddc2a719219b9717#file-readme-md

jbardin:
- sort ssh agent signers by requested id

It's becoming more common for users to have many ssh keys loaded into an
agent, and with the default max auth attempts of an openssh server at 6,
one often needs to specify which id to use in order to avoid a `too many
authentication failures` error.

Add a connection field called `agent_identity` which will function
similarly to the ssh_config IdentityFile when used in conjunction with
an ssh agent. This uses `agent_identity` rather than `identity_file` to
specify that the file is not used directly for authentication, rather
it's used to choose which identity returned from the agent to
authenticate with first.

This feature tries a number of different methods to match the agent
identity. First the provisioner attempts to read the id file and extract
the public key. If that isn't available, we look for a .pub authorized
key file. Either of these will result in a public key that can be
matched directly against the agent keys. Finally we fall back to
matching the comment string exactly, and the id as a suffix. The only
result of using the agent_identity is the reordering of the public keys
used for authentication, and if there is no exact match the client
will still attempt remaining keys until there is an error.

fixes #16909

### DIFF
--- a/communicator/ssh/provisioner.go
+++ b/communicator/ssh/provisioner.go
@@ -1,11 +1,15 @@
 package ssh
 
 import (
+	"bytes"
 	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net"
 	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform/communicator/shared"
@@ -50,6 +54,8 @@ type connectionInfo struct {
 	BastionPrivateKey string `mapstructure:"bastion_private_key"`
 	BastionHost       string `mapstructure:"bastion_host"`
 	BastionPort       int    `mapstructure:"bastion_port"`
+
+	AgentIdentity string `mapstructure:"agent_identity"`
 }
 
 // parseConnectionInfo is used to convert the ConnInfo of the InstanceState into
@@ -186,7 +192,8 @@ type sshClientConfigOpts struct {
 
 func buildSSHClientConfig(opts sshClientConfigOpts) (*ssh.ClientConfig, error) {
 	conf := &ssh.ClientConfig{
-		User: opts.user,
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		User:            opts.user,
 	}
 
 	if opts.privateKey != "" {
@@ -246,6 +253,7 @@ func connectToAgent(connInfo *connectionInfo) (*sshAgent, error) {
 	return &sshAgent{
 		agent: agent,
 		conn:  conn,
+		id:    connInfo.AgentIdentity,
 	}, nil
 
 }
@@ -255,6 +263,7 @@ func connectToAgent(connInfo *connectionInfo) (*sshAgent, error) {
 type sshAgent struct {
 	agent agent.Agent
 	conn  net.Conn
+	id    string
 }
 
 func (a *sshAgent) Close() error {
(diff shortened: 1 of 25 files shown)


jbardin:
Oh, I forgot to link the latest issue, which is #16909, but I have seen this situation couple times before.

josdotso:
@jbardin  This looks cool. It doesn't seem to have a bastion connection equivalent. 

Does it work with the bastion as jump host?  Should there be a `bastion_agent_identity` to select the identity for ssh to the bastion?

https://www.terraform.io/docs/provisioners/connection.html#connecting-through-a-bastion-host-with-ssh

PR #16975, opened 2017-12-22
Small typo fix in provider documentation

yaron:
We succeeded, disable partial mode. This causes Terraform to save ~~save~~ all fields again.


mitchellh:
Hah. Thanks!

PR #17020, opened 2018-01-02
Change URL Endpoint for Chef Provider to omnitruck

SchreiverJ:
The default Chef Provider endpoint has SSL issues preventing chef installs.
switch to the URL install endpoint used by knife: https://docs.chef.io/install_omnibus.html

Not sure which branch this should go to so if you need another request let me know.

### DIFF
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
-	installURL = "https://www.chef.io/chef/install.sh"
+	installURL = "https://omnitruck-direct.chef.io/chef/install.sh"
 )
 
 func (p *provisioner) linuxInstallChefClient(o terraform.UIOutput, comm communicator.Communicator) error {
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
-	installURL = "https://omnitruck.chef.io/chef/install.sh"
+	installURL = "https://omnitruck-direct.chef.io/chef/install.sh"
 )
 
 func (p *provisioner) linuxInstallChefClient(o terraform.UIOutput, comm communicator.Communicator) error {
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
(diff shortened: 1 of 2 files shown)


apparentlymart:
Thanks for working on this, @SchreiverJ!

@svanharmelen, would you mind taking a look at this?

svanharmelen:
Thanks @SchreiverJ! I can confirm that the currently used URL is giving SSL errors, but the correct URL should not be `https://www.chef.io/install.sh`, but `https://omnitruck.chef.io/install.sh` as per the docs you linked to yourself.

And (just double checked to be sure) `https://omnitruck.chef.io/install.sh`, does not give any SSL errors, so I think we should use the documented URL instead... Could you update that in your PR? Then we can merge this one right after that.

Thanks!

SchreiverJ:
I tried it out without the direct and I still got ssl issues when performing "terraform apply". I got the -direct url from watching what knife bootstrap was doing to install chef.

here is the curl output from running each on an Amazon Linux EC2:

```bash
$ sudo curl -LO https://omnitruck.chef.io/chef/install.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (60) Peer's certificate issuer has been marked as not trusted by the user.
More details here: https://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
$ sudo curl -LO https://omnitruck-direct.chef.io/chef/install.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 23432  100 23432    0     0  42277      0 --:--:-- --:--:-- --:--:-- 42219
```

svanharmelen:
@SchreiverJ I think the problem is (somehow) with your VM... If I try this on my mac it works fine:

image

If I try it on a vanilla Centos 7 machine, it also works fine:

image

Even if I try this on a pretty old Ubuntu 10.04.4 LTS machine I still have running, it still works fine:

image

Additionally when I think a little more about this one... You said that you also had SSL issues with `https://www.chef.io`. But that is their main website which definitely has a valid certificate. I do however notice that both `www.chef.io` and `omnitruck.chef.io` use a fastly service and certificate:

image

So I guess that somehow your VM doesn't trust these certs correctly. Maybe it's related to it being a Amazon Linux distro and it needs some kind of update or something...

Don't know, but I do know the URL `omnitruck.chef.io` is the correct URL to use. Using another undocumented URL sounds like a bad idea to me. As it's not documented, it means it can change at any time without notification...

svanharmelen:
Closing this one as the URL's are now correctly updated in PR #17355 (which also adds the option to select a channel). Thanks for your efforts @SchreiverJ, and it should be fixed now...

PR #17071, opened 2018-01-09
Issue 16360 - Empty -target= points to all resources in state

MarkFreebairn:


erran:
@MarkFreebairn You can use `go build -o terraform-provider-aws_v9.9.9` to create binary with your changes included and move it into your plugins directory:

```mv terraform-provider-aws_v9.9.9 ~/.terraform.d/plugins/darwin_amd64/```

As far as I know there are no acceptance tests to be run here (that's typically for providers).

Grab me when we're in the office tomorrow and we can validate the issue in a small example tf project.

PR #17076, opened 2018-01-10
Handle refresh errors.

ubschmidt2:
The condition in line 109 could never evaluate to true because the error returned from tfCtx.Refresh() was assigned to a different variable.

This is a follow-up to #16833.


jbardin:
Thanks for catching that @ubschmidt2!

Issue #16655, opened 2017-11-15
Terraform Crash [unexpected EOF] with chef provisoner, policyfile, & Terraform 0.9.8 - 0.10.8


chrisanthropic:
### Terraform Version
Terraform 0.9.8 - 0.10.8

### Terraform Configuration Files
Relevant block from my module:
```hcl
  provisioner "chef" {
    node_name = "${var.instance_name_prefix}${count.index}"
    server_url = "${var.instance_provisioner_chef_server_url}"
    policy_group = "${var.instance_provisioner_chef_policy_group}"
    policy_name = "${var.instance_provisioner_chef_policy_name}"
    attributes_json = "${var.instance_provisioner_chef_attributes_json}"
    secret_key = "${var.instance_provisioner_chef_secret_key}"
    user_name = "${var.instance_provisioner_chef_user_name}"
    user_key = "${var.instance_provisioner_chef_user_key}"
    version  = "${var.instance_provisioner_chef_version}"
    named_run_list = "${var.instance_provisioner_chef_named_run_list}"
  }
```
Resource block calling the module:
```hcl
module "buggy-thing" {
  source = "../modules/module-pgc2"
  instance_provisioner_chef_server_url        = "https://api.opscode.com/organizations/myorg"
  instance_provisioner_chef_use_policyfile    = true
  instance_provisioner_chef_policy_group      = "staging"
  instance_provisioner_chef_policy_name       = "buggy-thing"
  instance_provisioner_chef_attributes_json = <<-EOS
    {
      "tags": ["dc-${var.region}"],
      "secrets_locker": {
        "default_data_bag_item": "staging"
      }
    }
  EOS
  instance_provisioner_chef_secret_key = "${var.chef_encrypted_data_bag_secret}"
  instance_provisioner_chef_user_name = "FrankOz"
  instance_provisioner_chef_user_key = "${var.chef_validation_key}"
  instance_provisioner_chef_version = "12.14.89"
  instance_provisioner_chef_named_run_list = ""
}
```

### Debug Output
https://gist.github.com/chrisanthropic/e9f6cfb51c0e5a4c2623a0c991e8ccc8

### Crash Output
Please see above.

### Expected Behavior
Terraform plan should execute successfully.

### Actual Behavior
Terraform crash.

### Steps to Reproduce
Please list the full steps required to reproduce the issue, for example:
1. `terraform init`
2. `terraform get`
3. `terraform plan`

### Important Factoids
[...]

PR #17147, opened 2018-01-19
Fix chef provisioner validateFn

jeremiahsnapp:
Correctly validate Chef provisioner's `use_policyfile` field even if its value is a string type.

This fixes https://github.com/hashicorp/terraform/issues/16655

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"text/template"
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks for your PR @jeremiahsnapp! I see what you are trying to do, but I wonder if this isn't actually a bug in Terraform core instead.

@apparentlymart the [schema](https://github.com/jeremiahsnapp/terraform/blob/b363193261f9d1fe049167f1e4fe3e81eeadf894/builtin/provisioners/chef/resource_provisioner.go#L175-L178) clearly specifies the `use_policyfile` field as being a `schema.TypeBool`, so shouldn't TF core return an error on this instead?

jeremiahsnapp:
Thanks for looking at this @svanharmelen and @apparentlymart. I'm new to the terraform code so I'm not certain if their is a bug in the core code. I will say that the [Booleans docs](https://www.terraform.io/docs/configuration/variables.html#booleans) gives me the impression that using string values for a boolean variable is preferred and that TF should automatically convert the value internally to a boolean as needed.

The [chef provisioner's validateFn function](https://github.com/jeremiahsnapp/terraform/blob/v0.11.2/builtin/provisioners/chef/resource_provisioner.go#L372) is being passed a `terraform.ResourceConfig`. I believe TF's core [ResourceConfig.Get](https://github.com/hashicorp/terraform/blob/v0.11.2/terraform/resource.go#L232) _does not_ do the automatic conversion from string to boolean because it _does not_ take into account the data's schema.

Contrast this with how the [chef provisioner's decodeConfig function](https://github.com/jeremiahsnapp/terraform/blob/v0.11.2/builtin/provisioners/chef/resource_provisioner.go#L726) is passed a `schema.ResourceData`. I believe the [ResourceData.Get](https://github.com/hashicorp/terraform/blob/v0.11.2/helper/schema/resource_data.go#L70) _does_ do the automatic conversion from string to boolean because it _does_ take into account the data's schema.

So while the `schema.ResourceData` part of the code is auto converting the variable properly the `terraform.ResourceConfig` part is not. If that is intentional by design then I think my PR is required to fix the chef provisioner. If that is not the intended design then I agree that the bug is somewhere in TF core. Hopefully one of you can help isolate this issue better.

Thanks!

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -370,18 +371,27 @@ func applyFn(ctx context.Context) error {
...
+		switch usePolicyFileRaw := usePolicyFileRaw.(type) {
+		case string:
+			if usePolicyFileBool, err := strconv.ParseBool(usePolicyFileRaw); err == nil {
+				usePolicyFile = usePolicyFileBool
+			} else {
+				es = append(es, errors.New("\"use_policyfile\": field is not a boolean value"))
+			}
+		case bool:
+			usePolicyFile = usePolicyFileRaw
+		}
```
svanharmelen: May I suggest a small rewrite of your `switch` statement to make it just a bit more idiomatic Go?

```go
switch usePolicyFileRaw := usePolicyFileRaw.(type) {
  case bool:
    usePolicyFile = usePolicyFileRaw
  case string:
    usePolicyFileBool, err := strconv.ParseBool(usePolicyFileRaw)
    if err != nil {
      return append(es, errors.New("\"use_policyfile\" must be a boolean"))
    }
    usePolicyFile = usePolicyFileBool
  default:
    return append(es, errors.New("\"use_policyfile\" must be a boolean")) 
}
```

jeremiahsnapp:
Thanks a lot for the idiomatic Go advice @svanharmelen. I like it a lot. I only made a minor change to your return statements so they return the correct number of arguments. Thanks!

svanharmelen:
Again thanks for the PR and your inputs! LGTM! 🎉

Issue #17224, opened 2018-01-29
Habitat supervisor service is not enabled


smartb-pair:
By using habitat provisioner we observed that the `hab-supervisor` service is `running` but it’s not `enabled` which means upon reboot the supervisor will not be loaded again.

Workaround: the supervisor can be enabled by
```
provisioner "remote-exec" {
      inline = [ "sudo systemctl enable hab-supervisor" ]
    }
``` #

PR #17225, opened 2018-01-29
`enable` habitat supervisor service before `start`

smartb-pair:
Signed-off-by: Arash Zandi <arash.zandi@smartb.eu> #17224 

Enable Habitat Supervisor before starting the service. We think this should be the default behavior.


nsdavidson:
👍  This makes total sense to me, and it was an oversight on my part that this was not in the initial implementation.

rwc:
Fixes https://github.com/hashicorp/terraform/issues/17224

smartb-pair:
As this is the first time we've looked at Terraform code, we'd appreciate some guidance here on properly handling tests. @nsdavidson, perhaps you can help?

jbardin:
Hi @smartb-pair,

I'm not sure what happened to this PR, but I moved the commit over to #17781.
Can you verify if that is  all that is required?
Thanks!

PR #17309, opened 2018-02-09
website: Clarify Registry examples behavior

justincampbell:


justincampbell:
@mitchellh Agreed! Will add those soon.

PR #17349, opened 2018-02-14
docs/import: Remove out of date listing of importable resources

bflad:
Noting where to find up to date information. 😄


bflad:
Looks like a temporary issue, restarted the build and it ran fine. 👍

PR #17355, opened 2018-02-14
Updates the chef provisioner to allow specifying a channel

scotthain:
This also updates the omnitruck url to the current url.

Signed-off-by: Scott Hain <shain@chef.io>

### DIFF
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
-	installURL = "https://www.chef.io/chef/install.sh"
+	installURL = "https://omnitruck.chef.io/install.sh"
 )
 
 func (p *provisioner) linuxInstallChefClient(o terraform.UIOutput, comm communicator.Communicator) error {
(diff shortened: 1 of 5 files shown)


builtin/provisioners/chef/linux_provisioner_test.go
```diff
@@ -120,7 +120,26 @@ func TestResourceProvider_linuxInstallChefClient(t *testing.T) {
 
 			Commands: map[string]bool{
 				"curl -LO https://www.chef.io/chef/install.sh": true,
```
schisamo: We should update this URL to be `https://omnitruck.chef.io/install.sh` (along with all other spots `www.chef.io` is being used in test and no-test files).

builtin/provisioners/chef/linux_provisioner.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const (
 	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
-	installURL = "https://www.chef.io/chef/install.sh"
+	installURL = "https://omnitruck.chef.io/chef/install.sh"
```
schisamo: Drop the `chef` in this URL...we just want to use `https://omnitruck.chef.io/install.sh`, more details at https://docs.chef.io/packages.html#omnitruck-install-script-options

scotthain:
Hey folks - any chance we could get an ETA on merging this? Thanks in advance!

svanharmelen:
Thanks for this one @scotthain! LGTM and it also fixes PR #17020, so here goes...

PR #17359, opened 2018-02-15
Halt on provisioner errors

jbardin:
Previously, provisioners would always retry on error, even when they are not recoverable, e.g. invalid credentials, file not found, etc. They then need to be interrupted, or wait for the timeout which defaults to 5min. The remote-exec retryFunc was also not returning the stored error when it was finally interrupted.

First, this moves all the retryFunc implementations to the communicator package, using the most complete (remote-exec) version as the template. This new implementation also checks of the error implements the communicator.Fatal interface, as a signal to stop retrying. While in practice "retry-able" errors are more likely the exception, signaling off fatal errors is less likely to cause regressions in the providers, since not tagging an error as fatal will only continue the existing behavior. 

The ssh communicator can now tag the handshake errors as fatal, since retrying will not ever fix the situation.

We also can remote the retry logic around the remote command portion of remote-exec, since errors related to locating, uploading, and executing scripts are not likely to ever resolve themselves.

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -15,7 +15,6 @@ import (
 	"strings"
 	"sync"
 	"text/template"
-	"time"
 
 	"github.com/hashicorp/terraform/communicator"
 	"github.com/hashicorp/terraform/communicator/remote"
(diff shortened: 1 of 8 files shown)


dghubble:
> since retrying will not ever fix the situation

There was a situation where this was used (perhaps abused). On bare-metal provisions, there is typically a live image boot and then you reboot to use the disk install. There was no way for remote-exec to "know" to wait until the reboot had occurred. A common trick was to have terraform SSH as a user that wouldn't exist until the disk install had occurred. Instead of a fatal error, a single `terraform apply` run would kindly wait until auth succeeded (or timeout).

andrewsav-datacom:
How do we now reboot a VM and wait for it to come out of the reboot?

PR #17408, opened 2018-02-21
finish removing vet from the makefile

jbardin:
The test target will already cover vet, since it's run as part of the
test command now.

Also remove the `go test -i` since it's no longer needed with the new
build cache.


jbardin:
@apparentlymart, sorry, forgot a commit in that go1.10 PR :(

PR #17431, opened 2018-02-25
Use io.SeekStart instead of deprecated os.SEEK_SET

oscr:
This patch replaces the deprecated os.SEEK_SET with the instead recommended io.SeekStart.

https://github.com/golang/go/blob/39852bf4cce6927e01d0136c7843f65a801738cb/src/os/file.go#L81


jbardin:
Thanks!

PR #17438, opened 2018-02-26
helper/acctest: Fix RandIntRange function

aeriff:
RandIntRange should return integers within the range of min...max, not within the range of 0...(max - min).

I ran into this while working on https://github.com/terraform-providers/terraform-provider-aws/pull/2861. As far as I can tell all other uses within at least the AWS provider are not affected by this bug because any random integer is acceptable in those resources, however for the linked PR, it must be within a specific range.

### DIFF
--- a/helper/acctest/random.go
+++ b/helper/acctest/random.go
@@ -37,7 +37,7 @@ func RandIntRange(min int, max int) int {
 	source := rand.New(rand.NewSource(time.Now().UnixNano()))
 	rangeMax := max - min
 
-	return int(source.Int31n(int32(rangeMax)))
+	return int(source.Int31n(int32(rangeMax))) + min
 }
 
 // RandString generates a random alphanumeric string of the length specified


ewbankkit:
Strictly, this change would break backwards-compatibility.
It may be better to use `acctest.RandIntRange(min, max) + min` in code that needs this functionality.
I agree that `RandIntRange` is a confusing name for what the function does currently.

teamterraform:
Hi @aeriff
while we agree that this is a bug, fixing it would require breaking backwards-compatibility, as @ewbankkit mentioned.

The SDK, including the test framework with this function is due for some bigger changes and a window of opportunity may open in the future to introduce breaking changes in more transparent way. We will share more details about this project 🔜.

Thank you for your understanding in the meantime and sorry for any inconvenience caused by this.

We also filed an issue, so we don't forget to address this when the time comes.
https://github.com/hashicorp/terraform/issues/22394

PR #17465, opened 2018-02-28
Add support for user configured state object names in swift containers

Elethiomel:
Currently in the swift backend, the object state name is hardcoded to "tfstate.tf". This PR makes that configurable via a "state_name" option. The option is optional and  defaults to "tfstate.tf". This puts the backend more in line with the S3 backend and will allow users to have one container with multiple states.

I'm not 100% happy with the option name. I'd prefer to use "key" or something similar, but the backend already uses that internally for the Openstack key and I didn't want to have the option and internal struct members named differently. Any suggestions?

I, of course still need to update the documentation and test it a bit more, so this is very much a WIP.

### DIFF
--- a/backend/remote-state/swift/backend.go
+++ b/backend/remote-state/swift/backend.go
@@ -237,6 +237,13 @@ func New() backend.Backend {
 				Description: "Lock state access",
 				Default:     true,
 			},
+
+			"state_name": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: descriptions["state_name"],
+				Default:     "tfstate.tf",
+			},
 		},
 	}
 
@@ -308,6 +315,8 @@ func init() {
 		"archive_container": "Swift container to archive state to.",
 
 		"expire_after": "Archive object expiry duration.",
+
+		"state_name": "Name of state object in container",
 	}
 }
 
@@ -321,6 +330,7 @@ type Backend struct {
 	expireSecs       int
 	container        string
 	lock             bool
+	stateName        string
 }
 
 func (b *Backend) configure(ctx context.Context) error {
@@ -363,6 +373,9 @@ func (b *Backend) configure(ctx context.Context) error {
 	if err := config.LoadAndValidate(); err != nil {
 		return err
 	}
+
+	// Assign state name
+	b.stateName = data.Get("state_name").(string)
 
 	// Assign Container
 	b.container = data.Get("container").(string)
--- a/backend/remote-state/swift/client.go
+++ b/backend/remote-state/swift/client.go
@@ -19,8 +19,6 @@ import (
 )
 
 const (
-	TFSTATE_NAME = "tfstate.tf"
-
 	consistencyTimeout = 15
 
 	// Suffix that will be appended to state file paths
--- a/website/docs/backends/types/swift.html.md
+++ b/website/docs/backends/types/swift.html.md
@@ -26,8 +26,6 @@ terraform {
 ```
 This will create a container called `terraform-state` and an object within that container called `tfstate.tf`. It will enable versioning using the `terraform-state-archive` container to contain the older version.
 
--> Note: Currently, the object name is statically defined as 'tfstate.tf'. Therefore Swift [pseudo-folders](https://docs.openstack.org/user-guide/cli-swift-pseudo-hierarchical-folders-directories.html) are not currently supported.
-
 For the access credentials we recommend using a
 [partial configuration](/docs/backends/config.html).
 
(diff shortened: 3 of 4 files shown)


Elethiomel:
I've now updated the documentation. I think this is ready for someone else to take a look at now :)

backend/remote-state/swift/backend.go
```diff
@@ -154,6 +154,12 @@ func New() backend.Backend {
 				Description: descriptions["container"],
 			},
 
+			"state_name": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: descriptions["state_name"],
+			},
```
paultyng: Could you set a default value in here with the original value?

website/docs/backends/types/swift.html.md
```diff
@@ -19,14 +19,13 @@ Stores the state as an artifact in [Swift](http://docs.openstack.org/developer/s
 ```hcl
 terraform {
   backend "swift" {
-    path = "terraform-state"
+    path = "terraform-state",
+    state_name = "terraform.tfstate"
   }
 }
 ```
 This will create a container called `terraform-state` and an object within that container called `tfstate.tf`.
```
paultyng: Need to update this line here as well since you changed the example.

website/docs/backends/types/swift.html.md
```diff
@@ -54,6 +53,9 @@ The following configuration options are supported:
  * `path` - (Optional) DEPRECATED: Use `container` instead.
    The name of the container to create in order to store the state file.
 
+ * `state_name` - (Optional) The path to state file in the container.
+   Defaults to `terraform.tfstate`.
```
paultyng: Defaults to `tfstate.tf`

website/docs/backends/types/swift.html.md
Elethiomel: Well spotted! I've made a commit to fix the discrepancies. Basically, the old code was hardwired to use "tfstate.tf". Unfortunately, the documentation referred to "terraform.tfstate" in many places. The new commit fixes the docs to reflect the reality that the state has always been "tfstate.tf"

website/docs/backends/types/swift.html.md
Elethiomel: Fixed :)

backend/remote-state/swift/backend.go
Elethiomel: Hi Paul, I actually set a default on lines 289 to 293 in keeping with the style of previous code on lines 283 to 286. Is it OK to do it that way?

paultyng: You should set the value here, the difference is that if someone then explicitly writes `tfstate.tf` as a value I believe it will create a diff, although this is a backend so doesn't have the same implications as a provider. But we should use it consistently.

backend/remote-state/swift/client.go
```diff
@@ -13,23 +13,19 @@ import (
...
-const (
-	TFSTATE_NAME      = "tfstate.tf"
-	TFSTATE_LOCK_NAME = "tfstate.lock"
-)
-
 // RemoteClient implements the Client interface for an Openstack Swift server.
 type RemoteClient struct {
 	client           *gophercloud.ServiceClient
 	container        string
+	state_name       string
```
paultyng: please change this to `stateName`: https://golang.org/doc/effective_go.html#mixed-caps

backend/remote-state/swift/backend.go
```diff
@@ -238,6 +246,7 @@ type Backend struct {
 	archiveContainer string
 	expireSecs       int
 	container        string
+	state_name       string
```
paultyng: `stateName`: https://golang.org/doc/effective_go.html#mixed-caps

dariakts:
Hello,

Any updates for this pull request ?

Elethiomel:
I believe this was to be included in 0.12

pselle:
Hi @Elethiomel! Might I ask for an update to this PR to resolve its conflicts? Thanks so much for your patience here.

Elethiomel:
Quite a few changes have been made to the swift backend to support locking and environments, but I'll give it a go over the next week. I'm pretty swamped at work, but I would really like this PR to be accepted so I'll try me best :)

pselle:
Great @Elethiomel! We're keeping an eye on it now, happy to review (and get this in) when you've had the chance to revisit it!

Elethiomel:
@pselle Rebased and CLA signed :)

pselle:
@Elethiomel In my enthusiasm, I missed a key step 😅 for the purposes of making sure we cover our bases, would you mind running the acceptance tests for this backend and posting the results here? This is my bad for moving too soon/not realizing to say this earlier.

🙏 Thanks so much, happy to get this in the next release! 🤞

(I'm leaving the PR merged, but do need the test results within the next day for this to go in the release)

PR #17467, opened 2018-02-28
Standardize on User-Agent usage for registry, provider downloads, etc.

paultyng:


paultyng:
FWIW `"Terraform %s (%s)"` is also not a valid User-Agent string: https://tools.ietf.org/html/rfc2616#section-3.8

PR #17470, opened 2018-03-01
GCS backend: Detect and force-unlock stale locks.

ubschmidt2:
This change fixes the issue of orphaned lock files, which Terraform leaves
behind in some scenarios. Such lock files had to be removed manually, either
by means of a force-unlock or by deleting the file in GCS.

The "updated" timestamp of the lock file is used
as an indicator of staleness. This timestamp is updated once per minute as long
as the lock is held. A lock file is considered stale and is force-unlocked if
its timestamp hasn't been updated for several minutes.

### DIFF
--- a/backend/remote-state/gcs/client.go
+++ b/backend/remote-state/gcs/client.go
@@ -4,7 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"strconv"
+	"time"
 
 	"cloud.google.com/go/storage"
 	multierror "github.com/hashicorp/go-multierror"
@@ -23,7 +25,11 @@ type remoteClient struct {
 	stateFilePath  string
 	lockFilePath   string
 	encryptionKey  []byte
+	ticker         *time.Ticker
 }
+
+var tickInterval = 1 * time.Minute
+var minMissesUntilStale = 4
 
 func (c *remoteClient) Get() (payload *remote.Payload, err error) {
 	stateFileReader, err := c.stateFile().NewReader(c.storageContext)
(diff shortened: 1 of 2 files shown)


ubschmidt2:
@octo @danawillow

jbardin:
Thanks for the PR @ubschmidt2!

This seems like a reasonable feature to me, but it does require a little more care around the lock management than in the "fail locked" case that we have now, and for some cases the architecture of terraform itself prevents us from handling it very well at all. 

The main situation to consider, is that now that we have locks available, users are putting this into automation and blocking sometimes concurrent operations using the lock. If the client holding the lock loses connectivity temporarily, this can cause a waiting instance to possible incorrectly obtain the lock (granted, losing connectivity for over 4min, then regaining it and completing the run would be a very rare case).

Optimally terraform core could detect this and abort from the unsafe situation, but that isn't really possible with the current architecture. The consul backend (in which the lock always requires active connectivity) does 2 things to try and protect the user however -- it knows when the lock was lost and can report an error to the user, and it uses a CAS operation to ensure that the final state being overwritten is the the correct version. 

These things are tradeoff's of course, as if there are two instances applying at the same time, there is technically no "correct" state to write at the end, but I think at least reporting the situation to the user will alert them that there could be inconsistencies.

So, after that background from my adventures in state locking, I'll add the rest of the comments inline in the review.

backend/remote-state/gcs/client.go
```diff
@@ -77,6 +83,58 @@ func (c *remoteClient) Delete() error {
...
+		}
+	}
+
+	return false
+}
+
+// updateLockFile periodically updates the "updated" timestamp of the lock
+// file until the lock is released in Unlock().
+func (c *remoteClient) updateLockFile(lockFile *storage.ObjectHandle, generation int64) {
+	for _ = range c.ticker.C {
```
jbardin: We're going to want to add a mutex to the `remoteClient`, as it's now being mutated during execution, and there a race condition here. Offhand, I don't _think_ this would get hit during normal execution, but it does fail during tests, and should be protected regardless.

backend/remote-state/gcs/client.go
```diff
@@ -77,6 +83,58 @@ func (c *remoteClient) Delete() error {
...
+		if attrs, err := lockFile.Attrs(c.storageContext); err == nil {
+			if attrs.Generation != generation {
```
jbardin: I think this should probably signal some sort of error, as there state lock was changed unexpectedly. We don't have a very good way of surfacing this in terraform, but other clients report it as an error during Unlock, so the user can at least manually verify the situation.

backend/remote-state/gcs/client.go
```diff
@@ -77,6 +83,58 @@ func (c *remoteClient) Delete() error {
...
+				// This is no longer the lock file that we started with. Stop updating.
+				c.ticker.Stop()
```
jbardin: This goroutine should be  directly canceled by Unlock happening, and in that case `c.ticker.Stop()` can just be deferred at the top of the function.

backend/remote-state/gcs/client.go
```diff
@@ -90,19 +148,21 @@ func (c *remoteClient) Lock(info *state.LockInfo) (string, error) {
...
-		if _, err := w.Write(infoJson); err != nil {
-			return err
-		}
-		return w.Close()
-	}()
 
+	generation, err := c.createLockFile(lockFile, infoJson)
 	if err != nil {
-		return "", c.lockError(fmt.Errorf("writing %q failed: %v", c.lockFileURL(), err))
+		if c.unlockIfStale(lockFile) {
```
jbardin: I'd like to see a log warning line about unlocking an unexpected lock. (Hopefully we'll get better log handling soon, and these will be displayed to the user)

ubschmidt2:
Thanks, James, your review comments are very helpful. I'm working on the suggested improvements.

backend/remote-state/gcs/client.go
ubschmidt2: Added a mutex and using it where needed.

backend/remote-state/gcs/client.go
ubschmidt2: Good point. I think this is already covered in a sense that Unlock will also fail when attempting to remove the lock file using the wrong generation number:

```go
func (c *remoteClient) Unlock(id string) error {
...
        if err := c.lockFile().If(storage.Conditions{GenerationMatch: gen}).Delete(c.storageContext); err != nil {
                return c.lockError(err)
        }

        return nil
}
```

In any case, added some log output.

backend/remote-state/gcs/client.go
ubschmidt2: Thanks! I've modified accordingly.

backend/remote-state/gcs/client.go
ubschmidt2: I've added some log output for this case to unlockIfStale().

ubschmidt2:
I've updated the pull request.

ubschmidt2:
FWIW, the current Travis failure is unrelated to this PR.

```
# github.com/hashicorp/terraform/command
command/output_test.go:253:15: unknown field 'ContextOpts' in struct literal of type Meta
command/output_test.go:281:15: unknown field 'ContextOpts' in struct literal of type Meta
```

ubschmidt2:
PTAL, if you have a chance.

ubschmidt2:
PTAL, if you have a chance.

jbardin:
Hi @ubschmidt2,

Sorry about the long delay here, but we're nearing completion on the long 0.12 road and I'm going over PRs that were blocked by that work.

Can you rebase on master and re-run the tests?

ubschmidt2:
Ah, sorry, James, missed your comment. Yes, I'll do that. Also, still busy here with adoption of 0.12. ;-)

crw:
Do we know if this is still an issue with the current backend? I have added it to the list of PRs for the GCS team to triage. Thanks!

PR #17484, opened 2018-03-02
validation: Add ValidateRFC3339TimeString

bflad:
Helper function so providers can easily ensure an attribute parses as `time.RFC3339`. We could also introduce a `func TimeString(format string) schema.SchemaValidateFunc`. Let me know if you think that would be better or if we should have both. 😄

### DIFF
--- a/helper/validation/validation_test.go
+++ b/helper/validation/validation_test.go
@@ -144,6 +144,58 @@ func TestValidationRegexp(t *testing.T) {
 	})
 }
 
+func TestValidateRFC3339TimeString(t *testing.T) {
+	runTestCases(t, []testCase{
+		{
+			val: "2018-03-01T00:00:00Z",
+			f:   ValidateRFC3339TimeString,
+		},
+		{
+			val: "2018-03-01T00:00:00-05:00",
+			f:   ValidateRFC3339TimeString,
+		},
+		{
+			val: "2018-03-01T00:00:00+05:00",
+			f:   ValidateRFC3339TimeString,
+		},
+		{
+			val:         "03/01/2018",
+			f:           ValidateRFC3339TimeString,
+			expectedErr: regexp.MustCompile(regexp.QuoteMeta(`cannot parse "1/2018" as "2006"`)),
+		},
+		{
+			val:         "03-01-2018",
+			f:           ValidateRFC3339TimeString,
+			expectedErr: regexp.MustCompile(regexp.QuoteMeta(`cannot parse "1-2018" as "2006"`)),
+		},
+		{
+			val:         "2018-03-01",
+			f:           ValidateRFC3339TimeString,
+			expectedErr: regexp.MustCompile(regexp.QuoteMeta(`cannot parse "" as "T"`)),
+		},
+		{
+			val:         "2018-03-01T",
+			f:           ValidateRFC3339TimeString,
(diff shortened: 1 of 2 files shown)


helper/validation/validation_test.go
```diff
@@ -144,6 +144,58 @@ func TestValidationRegexp(t *testing.T) {
...
+			f:   ValidateRFC3339TimeString,
+		},
+		{
+			val: "2018-03-01T00:00:00+05:00",
+			f:   ValidateRFC3339TimeString,
+		},
+		{
+			val:         "03/01/2018",
+			f:           ValidateRFC3339TimeString,
+			expectedErr: regexp.MustCompile(regexp.QuoteMeta(`cannot parse "1/2018" as "2006"`)),
```
apparentlymart: This error message confuses me... what does "Cannot parse ... as 2006" mean? Is this just Go exposing its weird Go formatting representation?

I'd ideally like to try to do better than this and make the messages actually understandable to people who aren't Go experts, but I'm not sure how feasible that is based on what `time.Parse` returns. If this is the quality of the error messages from this function then I'd honestly rather just say "an RFC3339 timestamp is required" for all cases, since that at least offers _some_ reasonable resolution (go and read RFC3339).

Perhaps we could improve on that generic message in _some_ cases by some preprocessing:
* If it looks like the user is using wrong delimiters for either the date or the time, indicate that.
* If the timestamp is at the wrong granularity (only a date where a full time is required), indicate that.

[The `ParseError` type](https://godoc.org/time#ParseError) suggests that we _might_ be able to get some help with recognizing these situations from the `time` package itself, but the fields there are not documented so I'm not really sure what they mean and whether there's enough information here to recognize some common error cases. Maybe you can take a look at what's populated in `ParseError` for a few of these cases and see if it seems like something we could use to produce user-friendly messages for these scenarios?

paultyng:
I think taking the layout in makes the most sense, otherwise you are going to have to potentially do this for all the constants? Considering the layout is a valid example time, I think the error messaging should be pretty straight forward as you can show an example for how to format it?

apparentlymart:
Another thought on the generic error string idea: ideally we shouldn't be using a wide variety of different time formats across all of Terraform anyway, since that's not a great user experience. Perhaps we could standardize on always using RFC3339, regardless of what the underlying API uses, and convert where necessary to the underlying API?

I wouldn't do this in any case where the underlying representation has a different granularity than RFC3339, since that'd be confusing, but it seems right to me for full timestamps that can be unambiguously represented in that format.

paultyng:
Oh I like this, having a standard TF time we translate to whatever the API wants...

apparentlymart:
I like the explicitness of the current name but I don't feel that strongly about it, so I won't object if it's renamed.

paultyng:
I just worry it implies keep adding others. But I don't feel strongly
either! @bflad it's up to you!

bflad:
Thanks for the feedback, @paultyng and @apparentlymart! I've pushed this up so it now returns `invalid RFC3339 timestamp` instead of the cryptic time package errors. Can you please take another look?

Maybe in the future, we can look at the idea of a date/time being more of a first class type in Terraform schema to handle this logic a little more structurally: e.g. more descriptive error messaging, supporting short forms of the timestamp, standardizing on one format, etc. 😄

PR #17496, opened 2018-03-03
fix grammar in description

normanjoyner:
Simple update to grammar on Cloud index description. It is __not__ critical and can be released with a future version of the website.


jbardin:
Thanks!

PR #17527, opened 2018-03-08
Remove old completion file for zsh

ryosan-470:
#16176 was introduced before 6 months. This directory is not updated for 2 years. I suggest the completions file would be removed.


jbardin:
Thanks @ryosan-470!

PR #17551, opened 2018-03-10
Export a const for validation methods

paultyng:


paultyng:
There kind of [is one](https://github.com/hashicorp/terraform/blob/master/helper/resource/id_test.go#L36), but yeah I agree as its not a compile check anywhere would be useful. Also, not sure if its better to use the constants for the prefix / suffix length in the tests, or rewrite the expected values manually just for a double check. Have thoughts on that as I can kind of see both sides of it, I wasn't sure if the rewriting of the constant values in the test expected values was an explicit thing or just an oversight.

mitchellh:
Oh, that's easy enough then, maybe just swap that 26 out over for your new constant. If either goes out of sync, that test will catch both (I'd comment that though so people know what that's testing)

PR #17552, opened 2018-03-11
backend/gcs: Create the bucket if it does not exists

kavehmz:
When running `terraform init` on a new project if bucket does not exists we get error.

```
* writing "gs://my-new-bucket-name/terraform/state/default.tflock" failed: googleapi: Error 404: Not Found, notFound
* storage: object doesn't exist
```

As far as I can read the document, if I provide `project`, `init` should try to create the bucket if it does not exists.

https://www.terraform.io/docs/backends/types/gcs.html

My guess is that behavior was changed during recent refactoring which added lock to gcs.

This PR will bring back the behaviour.

### DIFF
--- a/backend/remote-state/gcs/backend.go
+++ b/backend/remote-state/gcs/backend.go
@@ -164,6 +164,15 @@ func (b *gcsBackend) configure(ctx context.Context) error {
 
 	b.storageClient = client
 
+	// If projectID is provider we will check and create the bucket if it does not exists.
+	// projectID is only used when creating a new bucket during initialization.
+	if b.projectID != "" {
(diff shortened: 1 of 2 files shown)


jbardin:
Hi @kavehmz,

Thanks for the PR, but that was a misfeature which should not have made it through review. Terraform should never create unmanaged resources, and I had to remove it in #16865.

While there has been discussion about adding some sort of interface to backends to allow them to create resource, the facts that they cannot be properly managed and that there are always other things required for bootstrapping reduced the desire for the added complexity.

kavehmz:
noted. Makes sense.

Issue #17314, opened 2018-02-10
Manta backend does not create path and lock fails


clstokes:
The [Manta backend](https://www.terraform.io/docs/backends/types/manta.html) does not create the state path as described in the [docs](https://www.terraform.io/docs/backends/types/manta.html#path) and the state lock fails every time if the lock file does not exist.

### Terraform Version
```
tf-manta-state $ terraform version -v
Terraform v0.11.3
+ provider.triton v0.4.1
```

### Terraform Configuration Files
```hcl
terraform {
  backend "manta" {
    path       = "terraform-modules/state-test"
    objectName = "terraform.tfstate"
  }
}

resource "triton_firewall_rule" "block" {
  rule        = "FROM any TO all vms ALLOW tcp PORT 22"
  enabled     = false
}
```

### Regular Output
https://gist.github.com/clstokes/1759ee75a1e0ced2243f917a9ceeeddc

### Expected Behavior
1. Expect the Manta backend to create the path in the backend definition.
1. Expect the Manta backend to successfully lock on each apply, destroy, etc.

### Actual Behavior
1. Terraform fails if the path doesn't already exist.
1. Terraform fails if the tflock file doesn't exist (which would mean it's locked?)

### Steps to Reproduce
1. terraform init
1. terraform apply
    - apply should fail with DirectoryDoesNotExist error
1. _create the path_
1. terraform apply
    - apply should fail with ResourceNotFound error

sean-:
CC @stack72

PR #17560, opened 2018-03-12
[WIP] backend/manta: Manta Backend was not dealing with a ResourceNotFound

stack72:
Fixes: #17314

We now deal correctly with the creation of the state file - we were
not dealing well with a ResourceNotFound error

Now that this has been changed around, we try and create the statefile
and if there is an error, we look for an existing statefile - previously
this was not the order of operations


stack72:
superceeded by #17561

PR #17567, opened 2018-03-13
backend/manta: Update triton-go dependency to 1.1.1

stack72:
This will support the features of RBAC in Manta Backends


stack72:
@jbardin it's not a necessity for this to be merged at the same time as #17561 but this fixes a different bug in the same area of the code - this time we fixed it in triton-go rather than needing to change the backend itself

PR #17591, opened 2018-03-15
Allow callers to append to user agent

paultyng:


httpclient/useragent.go
```diff
@@ -2,15 +2,29 @@ package httpclient
...
+	"log"
 	"net/http"
+	"os"
+	"strings"
 
 	"github.com/hashicorp/terraform/version"
 )
 
 const userAgentFormat = "Terraform/%s"
+const uaEnvVar = "APPEND_USER_AGENT"
```
jbardin: Is `APPEND_USER_AGENT` used anywhere else?
If there's not any precedent, I'd rather prepend this with `TF_`, if only so that's is more easily discoverable with other Terraform related env vars..

paultyng: Will do, I went back and forth on that prefix, as it isn't used for `CHECKPOINT_DISABLE` which was the other precedent I looked at for an env var used in tests/ci

paultyng:
@jbardin updated the prefix

PR #17618, opened 2018-03-18
website: remove an unnecessary parameter for API

outsideris:
In [Download the Latest Version of a Module](https://www.terraform.io/docs/registry/api.html#download-the-latest-version-of-a-module) API document for registry, `version` parameter isn't needed for that API.

### DIFF
--- a/website/docs/registry/api.html.md
+++ b/website/docs/registry/api.html.md
@@ -663,9 +663,6 @@ download endpoint (above) for the latest version.
 - `provider` `(string: <required>)` - The name of the provider.
   This is required and is specified as part of the URL path.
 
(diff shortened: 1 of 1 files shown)


jbardin:
Hi @outsideris,

Thanks for contributing.
Version _is_ required for the download request.
A request without the version however, does return a 302 to the path of the latest version.

outsideris:
@jbardin 
However, the endpoint is for lastest version. So, I don't think version is required.

> This endpoint downloads the latest version of a module for a single provider.

jbardin:
@outsideris, 

Sorry, you are correct. When I expanded this I thought it was under the next header up.
LGTM!

PR #17651, opened 2018-03-20
migrate legacy remote states to backends

jbardin:
Most of this is pretty mechanical -- move the client code to the backend, write a schema, move the factory function internals to the ConfigureFunc, and fixup the existing tests to initialize the backend.

There was some other dead code still in there too, like the file and gcs which get removed.


jbardin:
I'm not sure about the migration code yet, but I'll take a look.

PR #17657, opened 2018-03-21
helper/resource: Error out of test if both ExpectError and Check are set

paultyng:
Right now, the testing framework silently skips any checks you have defined if `ExpectError` is set.


paultyng:
Checks also seem to not run if `PlanOnly` is set, so can add that in to here as well. Tests seem to be failing as there are tests that don't respect this yet.

PR #17658, opened 2018-03-21
Use built-in method to get user homedir instead of eval on sh

pawelsocha:
## Why
For me, every ```eval``` in the code look like a potential security issue :-)
It's a small change to remove this.

### DIFF
--- a/config_unix.go
+++ b/config_unix.go
@@ -3,12 +3,10 @@
 package main
 
 import (
-	"bytes"
 	"errors"
 	"os"
-	"os/exec"
+	"os/user"
 	"path/filepath"
-	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


jbardin:
Hi @pawelsocha,

Thanks! While the possibility is small, the risk for arbitrary shell execution is there.

This code stems from earlier go<1.9 which required cgo for all user lookups. This however relies specifically on `/etc/passwd` and fails on systems using NSS, but I think that might be OK, since we could consider the first check for `$HOME` as the workaround in such cases.

pawelsocha:
Hi @jbardin :-)

Yes, but right now terraform requires golang with version 1.9+. So we can easily remove this eval ;-)

PR #17700, opened 2018-03-27
Fix scrub mistake

ScottWinkler:
Fixed a bug in which an example custom provider would be saved as "terraform-provider-custom-v0.0.1" instead of "terraform-provider-custom_v0.0.1". Not an issue when running the bundle on TFE, but is annoying when trying to run a terraform bundle locally.


jbardin:
Thanks for catching that @ScottWinkler!

PR #17719, opened 2018-03-28
update go-version

jbardin:
The latest go-version update properly handles pre-release versions in
constraints.


jbardin:
OMG! I've wanted that feature for ever! thanks for the tip ;)

Issue #16007, opened 2017-09-02
WinRM with domain accounts does not work


dave-pollock:
Hi,

Unless I'm missing something. Terraform is currently unable to use a WinRM connection if the account is a domain user. Local accounts work fine provided WinRM has been correctly configured on the target server. This seems to be because Terraform uses basic authentication to connect, and WinRM will not allow domain accounts using basic authentication. See: https://stackoverflow.com/questions/11153692/wsman-and-basic-authorization

I've noticed that Chef solve this by switching to Negotiate authentication when a domain account is detected. Negotiate will allow domain accounts.

Example Terraform code:
```hcl
connection {
  type     = "winrm"
  user     = "Administrator@mydomain.com"
  password = "myplaintextpassword"
  host     = "host"
  https    = "true"
  port     = "5986"
  insecure = "true"
}
```

The error in Terraform:
remote-exec-provisioner (internal) 2017/09/02 18:58:14 connection error: http error: 401 -

We can demonstrate that WinRM is configured correctly using Test-WSMan with a local account and basic authentication:
```powershell
$myLocalAccount = New-Object System.Management.Automation.PSCredential("Administrator",$secPasword)
Test-WSMan -ComputerName host -Authentication Basic -Credential $myLocalAccount -UseSSL -Port 5986


wsmid           : http://schemas.dmtf.org/wbem/wsman/identity/1/wsmanidentity.xsd
ProtocolVersion : http://schemas.dmtf.org/wbem/wsman/1/wsman.xsd
ProductVendor   : Microsoft Corporation
ProductVersion  : OS: 10.0.14393 SP: 0.0 Stack: 3.0
```

We can then change to a domain account and try again using basic authentication, which is not allowed:
```powershell
$myDomainAccount = New-Object System.Management.Automation.PSCredential("Administrator@mydomain.com",$secPasword)
Test-WSMan -ComputerName host -Authentication Basic -Credential $myDomainAccount -UseSSL -Port 5986
Test-WSMan : <f:WSManFault xmlns:f="http://schemas.microsoft.com/wbem/wsman/1/wsmanfault" Code="5" Machine="MY-COMPUTER"><f:Message>Access is 
denied. </f:Message></f:WSManFault>
```

But if we switch to negotiate authentication the domain account is now allowed:
```powershell
$myDomainAccount = New-Object System.Management.Automation.PSCredential("Administrator@mydomain.com",$secPasword)
[...]

PR #17748, opened 2018-03-31
Support NTLM for WinRM communicators.

joekhoobyar:
Some customers prohibit *any* basic or digest authentication methods over WinRM.

This is a simple patch to add NTLM (kerberos) support for WinRM connections, modeled after packer's support for the same.

When testing functionality, I ran the following on the host to disable all insecure auth:

```
winrm set winrm/config/client/auth '@{Basic="false"}'
winrm set winrm/config/client/auth '@{Digest="false"}'
winrm set winrm/config/service/auth '@{Basic="false"}'
```

Then, I tested it with the unpatched Terraform to verify that it locked me out.  (It did)
Finally, I tested it with the patched Terraform to verify that it let me in.  (It did)

### DIFF
--- a/communicator/winrm/communicator.go
+++ b/communicator/winrm/communicator.go
@@ -214,9 +214,12 @@ func (c *Communicator) newCopyClient() (*winrmcp.Winrmcp, error) {
 		},
 		Https:                 c.connInfo.HTTPS,
 		Insecure:              c.connInfo.Insecure,
-		TransportDecorator:    c.client.TransportDecorator,
 		OperationTimeout:      c.Timeout(),
 		MaxOperationsPerShell: 15, // lowest common denominator
+	}
+	
+	if c.connInfo.NTLM == true {
+		config.TransportDecorator = func() winrm.Transporter { return &winrm.ClientNTLM{} }
 	}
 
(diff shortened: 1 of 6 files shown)


joekhoobyar:
Fixes #16007

joekhoobyar:
Mitigates #15571

jbardin:
Thanks @joekhoobyar!

nicknameyu:
Hi There, I was following below link and came to this page. https://github.com/hashicorp/terraform/issues/15571  
I got the same issue connecting the new built windows VM through winrm. winrm https service is enabled with a self-signed certificate, Want to understand what configuration should be setup for terraform to connect to the VM. Configuration tried as blow:
```
  provisioner "remote-exec" {
    inline = [
      "md c:/temp",
    ]
        connection {
        type     = "winrm"
        host     = "${azurerm_network_interface.vmstamp.private_ip_address}"
        https    = true
        insecure = true
        port     = "5986"
        user     = "${var.admin_username}"
        password = "${var.admin_password}"
        use_ntlm = true
        timeout  = "1m"
  }

 }
```

Tried both use_ntlm = true and false. Can you help?

PR #17776, opened 2018-04-04
Add required DynamoDB IAM permissions for state locking to S3 backend docs

haines:
When setting up remote state locking with S3 and DynamoDB, I had to Google around to check which  IAM permissions Terraform requires for the DynamoDB table. The S3 bucket permissions are documented, so I thought it would be helpful to also document the DynamoDB table ones.

I also spotted a reference to the deprecated `lock_table` argument and updated it to `dynamodb_table`.

Relevant Terraform versions: 0.9+


jbardin:
Thanks for the PR @haines!

PR #17811, opened 2018-04-07
helper/schema: ResourceDiff ForceNew attribute correctness

vancluever:
A couple of bugs have been discovered in `ResourceDiff.ForceNew`:

* `NewRemoved` is not preserved when a diff for a key is already present.
This is because the second diff that happens after customization
performs a second `getChange` on not just state and config, but also on
the pre-existing diff. This results in `Exists == true`, meaning nil is
never returned as a new value.
* `ForceNew` was doing the work of adding the key to the list of changed
keys by doing a full `SetNew` on the existing value. This has a side
effect of fetching zero values from what were otherwise undefined values
and creating diffs for these values where there should not have been
(example: `"" => "0"`).

This update fixes these scenarios by:

* Adding a new private function to check the existing diff for
`NewRemoved` keys. This is included in the check on new values in
`diffChange`.
* Keys that have been flagged as `ForceNew` (or parent keys of lists and
sets that have been flagged as `ForceNew`) are now maintained in a
separate map. `UpdatedKeys` now returns the results of both of these maps,
but otherwise these keys are ignored by `ResourceDiff`.
* Pursuant the above, values are no longer pushed into the `newDiff`
writer by ForceNew. This prevents the zero value problem, and makes for
a cleaner implementation where the provider has to "manually" `SetNew` to
update the appropriate values in the writer. It also prevents
non-computed keys from winding up in the diff, which `ResourceDiff`
normally blocks by design.

There are also a couple of tests for cases that should never come up
right now involving `Optional`/`Computed` values and `NewRemoved`, for which
explanations are given in annotations of each test. These are here to
guard against future regressions.

### DIFF
--- a/helper/schema/resource_diff.go
+++ b/helper/schema/resource_diff.go
@@ -135,6 +135,10 @@ type ResourceDiff struct {
 	// diff does not get re-run on keys that were not touched, or diffs that were
 	// just removed (re-running on the latter would just roll back the removal).
 	updatedKeys map[string]bool
+
+	// Tracks which keys were flagged as forceNew. These keys are not saved in
+	// newWriter, but we need to track them so that they can be re-diffed later.
+	forcedNewKeys map[string]bool
 }
 
 // newResourceDiff creates a new ResourceDiff instance.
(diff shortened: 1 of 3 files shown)


vancluever:
Thanks @jbardin!

PR #17857, opened 2018-04-13
backend/local: fix incorrect set of state out path

Amygos:
Set the correct value of state out path file when the workspace
is not the default and custom state file path was set.


Amygos:
Superseded by  #19391

PR #17861, opened 2018-04-13
Fix permission for the lock.json

guessi:
the `lock.json` under .terraform/plugins/<arch>/lock.json is now with execution bit `0655`, which is unsafe and not required, better to have `0644`.


mildwonkey:
Hi guessi! Thank you for submitting this PR, and I'm sorry for the long, _long_ silence. 

This codepath is no longer used; a new plugin installer was introduced in terraform v0.13 and while the Read() function is still used, something else is responsible for writing this file - and it's now written with appropriate permissions.

Thanks, and sorry, again!

PR #17886, opened 2018-04-17
Update change.html.md: keep consistency in example file name

justinlarose:
In previous steps in getting started guide, it refers to the example file as `example.tf`.


mildwonkey:
Hi @justinlarose,

In this example, the item being referenced is a resource (`aws_instance.example`), not a file, so I believe that the current documentation is correct. 

If you still feel that the documentation is unclear, one option would be to edit the line to something like "Edit the `aws_instance.example` resource in your `example.tf` configuration file ..."

PR #17890, opened 2018-04-18
Fix small error in resource targeting docs

cabillman:


mildwonkey:
thank you for submitting this!

Issue #14325, opened 2017-05-09
terraform import appears fails to use correct provider for parts of the operation..


silvermy:
### Terraform Version
0.9.4

### Terraform Configuration Files
```hcl
provider "aws" {
  alias   = "src"
  region  = "${var.aws_src_region}"
  profile = "${var.aws_src_profile}"
}

provider "aws" {
  alias   = "dest"
  region  = "${var.aws_dest_region}"
  profile = "${var.aws_dest_profile}"
}

data "aws_caller_identity" "src" {
  provider = "aws.src"
}

resource "aws_s3_bucket" "src" {
  provider = "aws.src"
  bucket = "${var.src_bucket_name}"
  versioning {
    enabled = true
  }
  replication_configuration {
    role = "${aws_iam_role.src.arn}",
    rules = {
      prefix = ""
      status = "Enabled"
      destination {
        bucket = "${aws_s3_bucket.dest.arn}"
      }
    }
  }
}

resource "aws_s3_bucket" "dest" {
  provider = "aws.dest"
  bucket = "${var.dest_bucket_name}"
  versioning {
    enabled = true
  }
  policy = <<EOF
{
   "Version":"2008-10-17",
   "Id":"",
   "Statement":[
      {
         "Effect":"Allow",
         "Principal":{
            "AWS":"arn:aws:iam::${data.aws_caller_identity.src.account_id}:root"
         },
         "Action":["s3:ReplicateObject", "s3:ReplicateDelete"],
         "Resource":"arn:aws:s3:::${var.dest_bucket_name}/*"
      }
   ]
}
EOF
}
```

### Expected Behavior
s3 bucket should have been imported to the state file.

### Actual Behavior
the output implies that the aws s3 bucket resource was located and queried -  but that a subsequent (internal to the command) "Refreshing state" failed due to improper aws credentials.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
```
terraform import \
>     -provider aws.src \
>     -var aws_src_profile=src -var aws_src_region=us-east-1 -var src_bucket_name=src \
>     -var aws_dest_profile=dest -var aws_dest_region=us-west-1 -var dest_bucket_name=dest \
>     -var application=app \
>     aws_s3_bucket.src src
```

### Important Factoids
[...]

PR #17927, opened 2018-04-24
attach resource config during import

jbardin:
Since we require that a resource exist in the config for import, we also
need to ensure that the configuration is attached to the resource in the
graph so that the post-import refresh configures the correct provider.

This prevents the import from failing if it can't configure a default provider, which was noted in https://github.com/hashicorp/terraform/issues/14325#issuecomment-383303637

### DIFF
--- a/terraform/context_import_test.go
+++ b/terraform/context_import_test.go
@@ -1,6 +1,7 @@
 package terraform
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


samjgalbraith:
For anyone interested, our work-around for this issue is to use the AWS_PROFILE boto3/aws-cli environment variable to temporarily set the default profile:

```bash
export AWS_PROFILE=prod
terraform import ...
unset AWS_PROFILE
```

PR #18034, opened 2018-05-12
Fix syntax in doc for interpolating outputs from a module

SpencerBrown:
should be lowercase "module" not uppercase "MODULE" for interpolation of a module output


SpencerBrown:
Closing; fixed in new version of docs

PR #18093, opened 2018-05-21
Use proxy for installing chef-vault gem.

### LINKED ISSUE #18089: Chef provisioner doesn't use proxy when installing chef-vault gem
https://github.com/hashicorp/terraform/blob/e9e4ee494070950f31d17f29cae2d1facf5d564e/builtin/provisioners/chef/resource_provisioner.go#L571

Installation of chef-vault will fail when attempting to install behind a proxy. It should use the same proxy as defined by the chef configuration by default.

ikari7789:
Attempting to provision with vault_json enabled will currently fail if the server lies behind a proxy. This should fix #18089.

### DIFF
--- a/builtin/provisioners/chef/linux_provisioner.go
+++ b/builtin/provisioners/chef/linux_provisioner.go
@@ -11,10 +11,11 @@ import (
 
 const (
 	chmod      = "find %s -maxdepth 1 -type f -exec /bin/chmod %d {} +"
+	gemCmd     = "/opt/chef/embedded/bin/gem"
 	installURL = "https://omnitruck.chef.io/install.sh"
 )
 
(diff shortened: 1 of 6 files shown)


ikari7789:
Tests failed due to some random issue on before. Fixed now. Not sure if Windows solution is correct or not as I don't have experience with working with proxies on Windows.

pkolyvas:
@ikari7789 We announced provisioner deprecation in [mid-September 2020](https://discuss.hashicorp.com/t/notice-terraform-to-begin-deprecation-of-vendor-tool-specific-provisioners-starting-in-terraform-0-13-4/13997) and we added a deprecation notice for tool-specific (vendor or 3rd-party) provisioners in `0.13.4`. 

I realise this PR is long overdue for our attention. We’re working hard to clean up our community interaction so I’m tackling each one individually.

PR #18134, opened 2018-05-28
Mention `when = "create"` in provisioner documentation.

zerok:
While working on a "null_resource" I stumbled upon the `when` property and was wondering what the default value for it was. The documentation didn't mention it so I thought it might be a nice addition 🙂


pselle:
Apologies that this has gotten outdated, I'm going to close as the documentation has changed a lot since this was originally opened.

PR #18136, opened 2018-05-28
Module multiple provider example typo

hoeg:
The example for parsing multiple providers to a module had quotationmarks arount the keys. This does not seem to be the correct format.

(note: I have not inspected the code, but the example did not work for me. When I removed the quotation marks it did work)


mildwonkey:
Thank you for submitting this!

PR #18186, opened 2018-06-05
Update remote.html.markdown

Ian-T-Price:
Added https:// to URL address to correct the connection

Without this addition, the example results in:

> Error inspecting states in the "consul" backend:
>     Unexpected response code: 503

### DIFF
--- a/website/intro/getting-started/remote.html.markdown
+++ b/website/intro/getting-started/remote.html.markdown
@@ -52,7 +52,7 @@ First, configure the backend in your configuration:
 ```hcl
 terraform {
   backend "consul" {
-    address = "demo.consul.io"
+    address = "https://demo.consul.io"
     path    = "getting-started-RANDOMSTRING"
     lock    = false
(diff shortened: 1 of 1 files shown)


Ian-T-Price:
It should be noted that going to demo.consul.io from a browser does actually redirect to https://demo.consul.io.

Maybe this is a bug in Terraform?

thomashashi:
Related to #18287

FancyCoder0:
Dear Ian T Price(@Ian-T-Price ) & Thomas Kula(@thomashashi),
We are working on identifying redundant development and duplicate pull requests. We have found there is a pull request: https://github.com/hashicorp/terraform/pull/18287 which covers this one and already merged into main stream. And also another closed pull request https://github.com/hashicorp/terraform/pull/18296 which is duplicate to this one. So maybe this pull request should be closed.

We would really appreciate if you could help us to validate and give us some feedback.

PR #18187, opened 2018-06-05
Update config.html.md

Ian-T-Price:
Added 
    scheme  = "https"

Without this addition, the example results in:

> Error inspecting states in the "consul" backend:
> Unexpected response code: 503


Ian-T-Price:
It should be noted that going to demo.consul.io from a browser does actually redirect to https://demo.consul.io.

Maybe this is a bug in Terraform?

Ian-T-Price:
Superceded by another PR

PR #18233, opened 2018-06-12
Fixed method for grabbing public ip of server

Denn0:
Previous method provided a different, incorrect ip. Not sure what happens. Perhaps a temporary public ip gets assigned by AWS which is then overwritten by the elastic ip? 

In any case, this works. 
Just trying to help get these instructions to work for people new to Terraform like myself.


Denn0:
CI process seems broken...

pkolyvas:
As we've migrated Getting Started to our Learn track, I'm going to close this one out.

PR #18234, opened 2018-06-12
Updated info on file variables

Denn0:
Looks like variables structure has been reinvented a few times. This works. Just trying to get a working set of instructions for the next user.

### DIFF
--- a/website/intro/getting-started/variables.html.md
+++ b/website/intro/getting-started/variables.html.md
@@ -78,13 +78,23 @@ have to be input repeatedly as commands are executed.
 #### From a file
 
 To persist variable values, create a file and assign variables within
-this file. Create a file named `terraform.tfvars` with the following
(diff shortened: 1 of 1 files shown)


Denn0:
CI process seems broken...

mildwonkey:
@Denn0 -
Thanks for submitting this!

This part of the documentation is describing [assigning a variable](https://github.com/Denn0/terraform/blob/master/website/intro/getting-started/variables.html.md#assigning-variables), as opposed [defining a variable](https://github.com/Denn0/terraform/blob/master/website/intro/getting-started/variables.html.md#defining-variables), so the documented syntax is accurate as written.

I don't intend to discourage you - do you have any other thoughts on how this might be made more clear?

Denn0:
Suit yourself. You try to run that runbook yourself. You'll find it doesnt
work.

PR #18245, opened 2018-06-13
fix #17903: Remove confusing count variable

### LINKED ISSUE #17903: Spurious var in documentation
https://www.terraform.io/docs/configuration/interpolation.html

Either I'm missing something or the "count" variable defined at the top of the code block underneath "Using Templates with Count" is at best pointless and at worst possibly misleading.

alexis-turpin:
This solve #17903


nfagerlund:
Resolved in #20783

PR #18247, opened 2018-06-13
Fix broken links to /docs/enterprise/run/api.html

paddycarver:
We had a couple broken links pointing to the docs for the run API. This
updates them to the new location.


nfagerlund:
Cool, thank you for catching this.

PR #18311, opened 2018-06-22
Fix spelling mistake

conkiztador:


mildwonkey:
Thanks @conkiztador, good catch!

PR #18331, opened 2018-06-26
terraform state ls: add alias command

zimbatm:
Make it more consistent with the other commands:

	terraform state cp
	terraform state rm
	terraform state ls

I am sure I am not the only one always typing `terraform state ls`. This is a quick POC that works. If the idea is accepted, let me know what else needs to be done to make it merge-worthy.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -302,6 +302,13 @@ func initCommands(config *Config) {
 			}, nil
 		},
 
+		// Alias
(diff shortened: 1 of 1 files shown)


mildwonkey:
Hi! I am going to close this PR due to inactivity/failing tests. It's a reasonable request and if you are still interested we would be happy to review another PR. Thanks!

PR #18360, opened 2018-06-29
svchost/disco: make sure to close the response body

svanharmelen:
And fix a few linter warnings while we're at it…


svanharmelen:
👍

PR #18378, opened 2018-07-03
remote/backend/swift: Add support for workspaces

yanndegat:
- add support for workspaces in remote backend swift
- fix failing acc tests


yanndegat:
ping @jtopjian

backend/remote-state/swift/backend_state.go
```diff
@@ -1,22 +1,91 @@
...
+		if idx := strings.Index(object, delimiter); idx >= 0 {
+			continue
+		}
+
+		envs[object] = struct{}{}
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
jtopjian: From what I can tell, there's no new error being generated between here and line 30, so can this be removed?

PR #18431, opened 2018-07-11
Swift locking support

yanndegat:
- remote/backend/swift: Add support for locking


yanndegat:
hi @jtopjian 

i added locking support
inspired by other backend impls

backend/remote-state/swift/backend_state.go
```diff
@@ -91,27 +105,85 @@ func (b *Backend) State(name string) (state.State, error) {
...
+		stateMgr = &state.LockDisabled{Inner: stateMgr}
+	}
+
+	// Check to see if this state already exists.
+	// If we're trying to force-unlock a state, we can't take the lock before
+	// fetching the state. If the state doesn't exist, we have to assume this
+	// is a normal create operation, and take the lock at that point.
+	//
+	// If we need to force-unlock, but for some reason the state no longer
+	// exists, the user will have to use aws tools to manually fix the
```
fatmcgav: `s/aws/openstack`

backend/remote-state/swift/client.go
```diff
@@ -63,23 +87,134 @@ func (c *RemoteClient) ListObjectsNames(prefix string, delim string) ([]string,
...
 
-func (c *RemoteClient) Get() (*remote.Payload, error) {
-	log.Printf("[DEBUG] Getting object %s in container %s", c.objectName, c.container)
-	if err := c.ensureContainerExists(); err != nil {
-		return nil, err
+// GetNewestOpts add a parameter to ensure the objects returned are the most
+// recent known by swift.
+// swift is eventually consistent. Yet end users may expect to get a consistent
+// response when querying for state or lock objects
+type GetNewestOpts struct{}
```
jtopjian: @yanndegat 

I've just merged support for `Newest` to Gophercloud: https://github.com/gophercloud/gophercloud/issues/1118

You should be able to do a vendor update and then not have to create a custom type.

jbardin:
Hi @yanndegat,

This looks pretty good at first glance, but I would like to make sure it passes the acceptance tests first. This needs to call `backend.TestBackendStateLocks` and `backend.TestBackendStateForceUnlock` in the tests.

Since we don't have a way yet to run acceptance tests directly on backend PRs, we also require that the acceptance test output be pasted as a comment here to verify that they pass.

yanndegat:
hi @jbardin 

thanks a lot. 
no problem to add those tests. 
Actually, how do they relate to what i've added: here
https://github.com/hashicorp/terraform/pull/18431/commits/25100a253208eef8ad3f5e020e6c604d484b5072#diff-7e56003b6a8dd77c875fff599ade31a3R60

& right after with the testforceunlock test func ?

i added them by getting inspiration from other tests. if i add the ones you suggest, can i remove the ones i added because it seems to me that they overlaps ?

jbardin:
It looks like you copied a lot from the S3 backend, and IIRC that test was just to catch as specific failure for that implementation. I would call `TestBackendStateForceUnlock` instead to match the coverage of the other backends. 

The `TestBackendStateLocks` function should now cover everything that `remote.TestRemoteLocks` does. The old `state/remote` package will probably be removed in the near future.

yanndegat:
@jtopjian done

jtopjian:
I opened https://github.com/hashicorp/terraform/pull/18671 last week which I think should be merged before this. It does a bunch of vendor updates and gets the Swift backend aligned with the OpenStack provider as far as auth settings and such go. It would be a good foundation to apply additional features on top of.

Issue #16763, opened 2017-11-26
Support user accounts and/or SAS tokens for AzureRM backend auth


glenjamin:
### Terraform Version
```
0.11.0
```

### Overview
The azurerm backend currently supports two methods of authentication:

*Via storage account key*

These are pretty heavy-handed, quoting from the [azure docs](https://docs.microsoft.com/en-us/azure/storage/common/storage-security-guide#data-plane-security)
> Giving out your storage account key is like sharing the keys of your storage kingdom. It grants complete access.
This is especially relevant if you want to use the remote state data provider, as there's no way to grant read-only access using this method.

It would be good to support Shared Access Signatures, which are a form of authorization token derived from the storage account key that can impose restrictions like expiry time or more granular permissions.
 
*Via service principal credentials*

The other method currently supported is to use management API access credentials to grab the storage account key on demand, this is neater as it then uses the same credentials as the rest of the API. This shares most of the granularity downsides of the above approach, but there's not a lot that can be done about that due to the way azure IAM works.

The other downside about this approach is that service principals are a separate type of authentication credential that is managed separately from user accounts - so every user who needs to use terraform needs to have their own service principal. Support for using user account credentials was added to the terraform provider in https://github.com/terraform-providers/terraform-provider-azurerm/issues/42 - it would be good to get this into the backend as well.

### References
https://github.com/terraform-providers/terraform-provider-azurerm/issues/42

glenjamin:
While playing about with this I've come up with a decent way of wrapping both `terraform init` and `az` to grab a valid storage account key and pass it in as backend config

```
#!/bin/bash
set -eo pipefail

az account set --subscription "$subscription_id"
state_storage_access_key=$(
  az storage account keys list \
    --resource-group "$state_storage_resource_group" \
    --account-name "$state_storage_account" \
    --query '[0].value' -o tsv
)

terraform init \
  -backend-config="storage_account_name=$state_storage_account" \
  -backend-config="container_name=$state_storage_container" \
  -backend-config="access_key=$state_storage_access_key"
```

alphalzh:
Any plan on implementing this? If not I could try and create a PR for this issue sometime this month.

alphalzh:
For supporting this, the vendor azure-sdk-for-go must be updated, since in later versions of azure sdk they added support for using SAS tokens to access blob storage. Shall I just create a PR for updating the dependency first?

perbergland:
[...]

PR #18476, opened 2018-07-17
Update the vendor azure-sdk-for-go from 10.0.2 to 18.1.0

alphalzh:
Background: #16763 
I would like to add support for using SAS tokens for azurerm backend authentication, since it's a more secure workflow than using long-lived access keys to authenticate an azurerm remote backend. However, the current version of azure-sdk-for-go (version 10.0.2) in vendors/ does not support SAS tokens, but it is introduced in a later version. So I updated the azure-sdk-for-go to use the latest version.

Testing:
 - I confirmed that azure-sdk-for-go is only used for remote Azure tfstate backend. Thus, no existing functionalities related to deploying resource to Azure will be affected. 
 - I verified that all unit tests pass.
 - I also manually verified that Azure tfstate backend works properly. terraform init will correctly initialize the remote backend, terraform apply will correctly update/create tfstate in Azure storage account if remote backend is used.

Note: mastr/guid is a requirement for the newer version of azure-sdk-for-go and is added as a dependency.


tombuildsstuff:
hey @alphalzh 

Thanks for this PR - apologies for the delayed response here!

To give a little more information on our plans for the AzureRM Backend - currently most of the authentication logic used by the Backend is duplicated between the AzureRM Provider and the Backend - which both use a different version of the SDK.

In the near future we're intending to refactor this authentication related logic out into it's own package, which will allow us to add support for multiple authentication mechanisms within the Backend (e.g. for example, this'd allow us to support Azure CLI / CloudShell etc - and potentially other authentication types such as SAS Tokens). From a UX perspective this should mean it's also required for users to specify the name of the storage account/container (and possibly the resource group) - meaning the backend configuration becomes considerably simpler than it is today.

Since we're taking a slightly different direction here I'm going to close this PR for the moment - but I'd like to thank you for this contribution :)

Thanks!

PR #18497, opened 2018-07-19
remote/backend/http: add env var config options

tonglil:
Allow HTTP backend's config options such as addresses and user/pass info to be configured via environment variables (similar to the other remote backend options).


syphernl:
Too bad this never got merged. It would be really useful to us and makes sense to have this, since all other backends support it too.

nicholasklick:
@syphernl this PR will finally offer full support for ENV with the `http` backend https://github.com/hashicorp/terraform/pull/25439

mildwonkey:
Hi! Thank you for submitting this. I apologize for overlooking this submission; we have merged a PR that added the same functionality: https://github.com/hashicorp/terraform/pull/25439
Thanks!

PR #18516, opened 2018-07-22
Fix documentation on tainting resources within modules

xulsitatirev:
It looks like the output does not match the command, in fact.
This pull-request fixes this issue.


mildwonkey:
Good catch, thank you @xulsitatirev!

PR #18533, opened 2018-07-25
change chef linux provisioner to use user:group chown syntax

alice-sawatzky:
`chown user.group` appears to _work_ in linux but not documented (at least not in the `chown` man page). It also simply doesn't work at all in freebsd (at least on my pfsense machine based on freebsd 11). This changes the provisioner to use the more standard `chown user:group` syntax.

Please let me know if this introduces some obscure compatibility issue I'm not aware of.


Victorion:
Hi @svanharmelen 

Would it be possible to merge this PR for 0.11.9 release? thank you.

svanharmelen:
@Victorion yes, I think this one can be added to 0.11.9 👍

Issue #18547, opened 2018-07-26
Add terraform version to user-agent in http remote state


repl-david-winiarski:
### Terraform Version

```
Latest version
```

### Terraform Configuration Files

```hcl
...
```

### Expected Behavior

I would expect terraform to identify which version is attempting to access the remote state file from the http backend.
### Actual Behavior

The regular Go user agent is displayed instead of a terraform specific one.

pkolyvas:
I'm doing a clean-up of older issues that haven't gained traction but also don't coincide with some of our current focus. This issue had its associated PR closed during the 0.12 release window and so I'm just finishing the related work. 

We appreciate every idea and contribution to Terraform, even if we can't get to them all. My door is always open to talk about Terraform, the use cases, and needs of our practitioners.

PR #18548, opened 2018-07-26
set user-agent to terraform version for remote http state calls

repl-david-winiarski:
test has been updated as well to check for a correct user agent

This is a small proof of concept, and I am open to updating the user agent format.

Issue: https://github.com/hashicorp/terraform/issues/18547

### DIFF
--- a/state/remote/http.go
+++ b/state/remote/http.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 
 	"github.com/hashicorp/terraform/state"
+	"github.com/hashicorp/terraform/version"
 )
 
 func httpFactory(conf map[string]string) (Client, error) {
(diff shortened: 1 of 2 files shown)


repl-david-winiarski:
A little more information on the use case for this. We want to be able to control which version of terraform accesses our remote state files in order to prevent an unintended terraform version upgrade. Having a user agent that identifies the version would allow us to reject the initial state file GET call, and prevent the user from inadvertently updating the state files terraform version.

teamterraform:

This PR touches code that has changed quite a bit since it was originally opened. Specifically, most of the state functions have been moved to `states` while we deprecate `state`. 

We are going to close this PR, but please do not hesitate to open a new PR or issue in the future, and thank you!

PR #18550, opened 2018-07-26
grpc proto definitions

jbardin:
This add the grpc protobuf definitions, and the generated code. 

The proposed protocol is fairly stable now, so this gives us a reference point and allows tracking of any further enhancements made in the repo.


jbardin:
Yeah, it was just simpler to do that way since there was a couple shared types and the provisioner interface is quite small. They of course can always be split out and imported later if need be.

PR #18617, opened 2018-08-07
added make install to Vagrantfile

5forcegees:
Getting started using Vagrant requires the user to run `make` commands in the VM

This command fails because `make` isn't installed inside the VM by default. 

This PR adds `make` to the `apt-get install` command
 
The documentation is already written assuming that make is installed:
https://github.com/hashicorp/terraform#developing-terraform


teamterraform:
Hi folks! Now that Terraform 0.12 is released, we are undertaking a general GitHub issue/PR cleanup.  The Vagrantfile has been removed, so we will close this PR.  Please don't hesitate to open a new PR in the future, and thank you!

PR #18621, opened 2018-08-07
Support PKCS#8 keys as per https://go-review.googlesource.com/c/crypto/+/127779

frezbo:
Support PKCS#8 keys as per https://go-review.googlesource.com/c/crypto/+/127779

Signed-off-by: Noel Georgi <git@frezbo.com>


frezbo:
@jbardin @apparentlymart can this be looked at?

jbardin:
The current modules have upgraded x/crypto past this point.

PR #18648, opened 2018-08-09
pretty-print json in http requests/responses

danawillow:
log messages in provider code get passed through this function: https://github.com/hashicorp/terraform/blob/35d47201ad24d89081d55ea3057c5f7144761b2c/vendor/github.com/hashicorp/go-plugin/client.go#L754-L790

If a line contains only JSON, it looks for specific entries in the JSON map to give it instructions on how it should be printed. In this particular case, since we're just printing JSON being sent to/from a server, we don't want it to contain those special messages, we want it to get printed directly.

By pretty-printing it, we end up with no one line that contains valid JSON (so they all get printed), and we have the benefit of easier-to-read logs. Win-win!

### DIFF
--- a/helper/logging/transport.go
+++ b/helper/logging/transport.go
@@ -1,9 +1,12 @@
 package logging
 
 import (
+	"bytes"
+	"encoding/json"
 	"log"
 	"net/http"
 	"net/http/httputil"
+	"strings"
 )
 
 type transport struct {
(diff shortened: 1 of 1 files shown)


danawillow:
Thanks Paddy! I don't have merge privileges here anymore by the way, if you'd like to merge :)

And yes, `"foo"` is valid JSON, but that fn is specifically looking for valid JSON maps:
https://github.com/hashicorp/terraform/blob/35d47201ad24d89081d55ea3057c5f7144761b2c/vendor/github.com/hashicorp/go-plugin/log_entry.go#L35-L39

paddycarver:
You're right, I missed that! I talked with some of our core team to make sure I wasn't going to mess things up for them by merging, and it sounds like we're in the clear. So I'll merge this now. :D

PR #18671, opened 2018-08-14
backend/remote: Swift Authentication Update

jtopjian:
For #18569

This commit updates the authentication configuration for the Swift
remote.

Support for cross-domain authentication has been added and mapping
environment variables to the correct domain settings has been
fixed.

In addition, support for clouds.yaml files has been added.

The bulk of the files are from a mass vendor update. Sorry about that.

Upgrade Notes:

We applied this fix to the OpenStack Provider back in April/May and noticed the following:

The majority of users will not see an issue, especially those who authenticate entirely via environment variables.

Users who have modified/altered their auth settings to account for the prior behavior might have to change their auth settings to what they should have been all along.

There's always the possibility that a user will run into some difficulty or even a bug. Please ping me on any report and I can investigate.

### DIFF
--- a/backend/remote-state/swift/backend.go
+++ b/backend/remote-state/swift/backend.go
@@ -23,7 +23,7 @@ func New() backend.Backend {
 			"auth_url": &schema.Schema{
 				Type:        schema.TypeString,
 				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("OS_AUTH_URL", nil),
+				DefaultFunc: schema.EnvDefaultFunc("OS_AUTH_URL", ""),
 				Description: descriptions["auth_url"],
 			},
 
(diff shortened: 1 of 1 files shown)


dmitrifedo:
Hi @jtopjian, I appreciate your effort to fix the issues with Swift remote backend. In fact, your fix enables us to use Swift in our environment. 

I was wondering whether we can expect this PR to be merged into next `0.11` and `0.12` release or maybe you need help with something to get it through?

jtopjian:
> I was wondering whether we can expect this PR to be merged into next 0.11 and 0.12 release or maybe you need help with something to get it through?

Thanks for the poke. I'll find some time to rebase this PR and look into getting it merged.

jtopjian:
This should be good to go. The comments at the beginning of this PR are still true. 

To clarify the following:

> Users who have modified/altered their auth settings to account for the prior behavior might have to change their auth settings to what they should have been all along.

The user is not at fault here. It was entirely due to the backend OpenStack library. The changes in this PR will allow a user to authenticate as they would with any other OpenStack compatible app/tool.

```
=== RUN   TestBackend_impl
=== RUN   TestBackendConfig
--- PASS: TestBackendConfig (0.43s)
=== RUN   TestBackend
--- PASS: TestBackend (6.07s)
=== RUN   TestBackendArchive
--- PASS: TestBackendArchive (28.58s)
=== RUN   TestRemoteClient
--- PASS: TestRemoteClient (0.53s)
```

PR #18688, opened 2018-08-15
helper/resource: Add ParallelTest() function to allow opt-in acceptance testing concurrency with t.Parallel()

bflad:
Please note: this is an alternate implementation to #16807, using a separate function rather than requiring adjustment to each existing `TestCase` and submitted just for an additional reference point. The preferred solution may be either implementation or a combination of the two! I think more importantly we should continue the conversation for supporting this type of functionality as its been implemented or requested by a few of the major providers. 😄 

While this initial implementation is a very simple wrapper function, implementing this in the helper/resource package provides some downstream benefits:
* Provides a standard interface for plugin developers to enable parallel acceptance testing
* Existing plugins can simply convert `resource.Test` to `resource.ParallelTest` references (as appropriate) to enable the functionality, rather than worrying about additional line(s) to each acceptance test function or `TestCase`
* Potential enhancements to ParallelTest (e.g. adding an environment variable to skip enabling the behavior) are consistently propagated

### DIFF
--- a/helper/resource/testing.go
+++ b/helper/resource/testing.go
@@ -418,6 +418,17 @@ func LogOutput(t TestT) (logOutput io.Writer, err error) {
 	return
 }
 
+// ParallelTest performs an acceptance test on a resource, allowing concurrency
+// with other ParallelTest.
+//
+// Tests will fail if they do not properly handle conditions to allow multiple
+// tests to occur against the same resource or service (e.g. random naming).
+// All other requirements of the Test function also apply to this function.
+func ParallelTest(t TestT, c TestCase) {
+	t.Parallel()
+	Test(t, c)
+}
+
 // Test performs an acceptance test on a resource.
 //
 // Tests are not run unless an environmental variable "TF_ACC" is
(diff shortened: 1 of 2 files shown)


paddycarver:
I can't believe I didn't think of that before adding `t.Parallel()` to hundreds of tests. 😓

MalloZup:
@bflad what is the next step for this PR to get merged? i really like this feature and could help us in our libvirt provider having this to the schema helper

Thank you for your attention. Have nice day.! :call_me_hand: :sun_with_face:

MalloZup:
Thanks :love:

MalloZup:
:rocket: :guitar: :tada:

metacpp:
@bflad thanks for the fantastic work!

The output of testing result is a known issue for go test running in parallel. Please visit https://github.com/golang/go/issues/19280 for the details. The proposal is to output test result in a structural (for example, JSON) format, see https://github.com/golang/go/issues/2981. And it was released in v1.10 through `go test -json`, we need to remind provider developer for this.

PR #18700, opened 2018-08-17
Docker multi stage build.

a4t:
Hi. I think the Docker container is heavy.
How about using docker multi stage build?
If you have already verified it sorry.

Image size: 118MB -> 101MB


crw:
Thanks for this PR. I realize it is now 6 years later, and I am going to make the assumption that this PR is no longer relevant to the current build process. In general, we no longer accept PRs that impact the build process as that is maintained by the internal release management team (and we don't necessarily actually use some of the build scripts in the repository to build the artifacts). Thanks again!

PR #18711, opened 2018-08-19
Correct relative symlinks

retr0h:
When srcPath contains a relative path symlink os.Readlink fails.
This patch corrects this behavior.  Terraform doesn't run into
this issue currently.  However, I ran into this issue when using
the CopyDir function in my own project [1].  Since CopyFile and
CopyDir are not built into golang, Terraform's copy package
sufficed.

[1] https://github.com/retr0h/go-gilt/blob/golang-port/copy/copy.go


mildwonkey:
(sorry for the labeling noise, I had a moment)

PR #18757, opened 2018-08-29
backend/migrations: check all workspaces

svanharmelen:
This commit fixes a bug that (in the case of the `local` backend) would only check if the selected workspace had a state when deciding to perform a migration.

When the selected workspace didn’t have a state (but other existing workspace(s) did), the migration would not be performed and the other workspaces would be ignored.


svanharmelen:
Thx!

PR #18792, opened 2018-09-05
Fix description for state mv and push sub-commands

davewongillies:
### Reasoning for docs update

The description for the `state mv` and `state push` had descriptions for the `state rm` command. It only shows up in places like Slack which renders link previews, so low impact.


mildwonkey:
Thanks @davewongillies, these are great catches/fixes!

PR #18795, opened 2018-09-06
Allow for nested fields with checkKey on ResourceDiffs.

paddycarver:
When checking if a ResourceDiff key is valid, allow for keys that exist
on sub-blocks. This allows things like diff.Clear to be called on
sub-block fields, instead of just on top-level fields.


paddycarver:
@apparentlymart I added a few test cases to the existing tests for ResourceDiff.Clear (which calls into the checkKey helper I modified in this). Does that work as a testing strategy?

paddycarver:
New commit pushed. The executive summary is that we don't allow `Set` on individual items in complex types, so we shouldn't consider those valid keys for `SetNew` or `SetNewComputed`.

PR #18800, opened 2018-09-06
an empty object state should decode into null

jbardin:
The default Decode path would return an EOF error if the state was
empty.  Return a null cty.Value instead.


jbardin:
OK, I missed that one of the values _must_ be set, even for an empty state. I can dig further and see where the empty state is coming from.

PR #18826, opened 2018-09-10
backend/remote: add support for state locking

svanharmelen:


backend/remote/backend_test.go
```diff
@@ -16,6 +16,18 @@ func TestRemote(t *testing.T) {
...
+	backend.TestBackendStates(t, b)
+	backend.TestBackendStateLocks(t, b, b)
+	backend.TestBackendStateForceUnlock(t, b, b)
+}
+
+func TestRemote_backendNoDefault(t *testing.T) {
+	b := testBackendNoDefault(t)
+	backend.TestBackendStates(t, b)
+}
+
```
jbardin: This doesn't look like a diff, just checking if there was supposed to be something other than moving the source.

backend/remote/backend_state.go
```diff
@@ -101,3 +103,71 @@ func (r *remoteClient) Delete() error {
...
+	}
+
+	// Check if the workspace is already locked.
+	if w.Locked {
+		lockErr.Err = fmt.Errorf("remote state already locked")
+		return "", lockErr
+	}
+
+	// Lock the workspace.
+	w, err = r.client.Workspaces.Lock(ctx, w.ID, tfe.WorkspaceLockOptions{
```
jbardin: The `Lock` operation can override the LockInfo.ID` field here if you want, so there is only one form of the lock ID to present for the user. This way you're not comparing both the current ID and the possible `r.organization+"/"+r.workspace != id`. I'm not sure if there's any additional safety here just using the random UUID if an unlock can be accomplished with just the org+workspace string.

backend/remote/backend_test.go
svanharmelen: Nope, I just wanted to move these to the top of the page 👍

backend/remote/backend_state.go
svanharmelen: I think that could cause issues as then every instance would always get the same static lock ID. The difference with most other remote state endpoints that support locking is that they have a way to amend some lock info. In this case we cannot amend any info to the lock, so we can only determine if the state is locked or not.

Doing it like this ensures that if you locked the state, only you (i.e. the current command being executed) can unlock it. So if anything breaks or fails you would always have to specifically `force-unlock` it with the organization/workspace name combination, which would be a very conscious user action.

jbardin: Is the lockInfo stored anywhere outside of the client during execution? I can see that it works as a failsafe within the execution itself, so it should be OK to keep. I was thinking that it could present the user with a lock error indicating a UUID which can't be used to force-unlock, with no indication that the ID to unlock is really `r.organization+"/"+r.workspace != id` (though it's probably less likely with a backend like TFE, and I of course might have missed something here)

svanharmelen: Good point about the UUID maybe being displayed in an error message! I checked that and this is not in a generic error, but most remote state lock functions do return the ID as part of an error.

Currently I don't, as that UUID is not something that can be referenced or used anywhere outside the currently executed command. But what I can do is add the ID that you can use as info for the user (so refer to `organization/workspace`). Let me add that to this PR as that will probably help people when they want or need to use `force-unlock`.

svanharmelen: It now looks like this:

image

Which seems nice and clear to me. Agree? Any additional thoughts?

jbardin: LGTM! :+1:

svanharmelen:
Travis is actually green, but really slow and it's probably not going to update the commit status anymore. So merging as is...

PR #18835, opened 2018-09-11
Remove a leftover test log file

alexsomesan:
Just ran into this file by chance. It seems to be a piece of a test log.
I presume it was never meant to be checked in.

@svanharmelen Can you please review?


svanharmelen:
LGTM!

PR #18859, opened 2018-09-14
Chef Solo provisioner

andreykaipov:
I don't know if you guys want another builtin provisioner for Terraform, but I really liked the [Packer provisioner for Chef Solo](https://www.packer.io/docs/provisioners/chef-solo.html), so I wrote this provisioner in that style.

Overview:
- The behavior and documentation is very similar to Packer's Chef Solo provisioner, so hopefully it's familiar to users.
- In case the existing exposed settings don't satisfy someone's needs, templating the Chef config file, execute command, and install command are all supported.
- Dynamic attributes are intended to be set by using `template_file`s and passing the rendered template as JSON attributes to the Chef config.

Lemme know what you think!

### DIFF
--- a/builtin/bins/provisioner-chef-solo/main.go
+++ b/builtin/bins/provisioner-chef-solo/main.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"github.com/hashicorp/terraform/builtin/provisioners/chef-solo"
+	"github.com/hashicorp/terraform/plugin"
+)
+
+func main() {
+	plugin.Serve(&plugin.ServeOpts{
+		ProvisionerFunc: chefsolo.Provisioner,
+	})
+}
--- a/builtin/provisioners/chef-solo/apply.go
+++ b/builtin/provisioners/chef-solo/apply.go
@@ -1,0 +1,217 @@
+package chefsolo
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/hashicorp/terraform/communicator"
+	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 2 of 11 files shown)


danieldreier:
Hi @andreykaipov -

I'm closing this PR because we announced tool-specific (vendor or 3rd-party) provisioner deprecation in [mid-September 2020](https://discuss.hashicorp.com/t/notice-terraform-to-begin-deprecation-of-vendor-tool-specific-provisioners-starting-in-terraform-0-13-4/13997). Additionally, we added a deprecation notice for tool-specific provisioners in 0.13.4. On a practical level this means we will no longer be reviewing or merging PRs for these built-in plugins.

For the time being, the best option is to distribute this as a plugin like the [ansible provisioner does](https://github.com/radekg/terraform-provisioner-ansible). I'm aware of the limitations of this approach, but it's the less-bad option compared to coupling provisioner development to the Terraform Core release lifecycle. We believe the benefit to users of having provisioner development decoupled from core, exceeds the convenience of having these provisioners built in to core. We want to provide a better user experience in the future, and our hope here is that the ability to improve, fix and repair provisioners without us blocking their development, much like providers, will help make a strong case for what's next.

I think it’s also important to highlight that we have no plans to remove the generic provisioners or the pluggable functionality during Terraform's 1.0 lifecycle. 

Please feel free to reach out to me or Petros Kolyvas if you would like to talk more about this change.

andreykaipov:
No stress! It did feel strange how some tools had built-in support but not others. I agree it's more convenient to have them decoupled from core. We've been using it [as a plugin](https://github.com/andreykaipov/terraform-provisioner-chef-solo) since I opened up this PR. Another step to move it into our runner's path was no big deal.

 Glad to know the general remote-exec one will stay though! Thanks.

PR #18866, opened 2018-09-14
provisioners/chef: prevent concurrent chef-vault changes

lanej:
This is my first terraform PR!

Race conditions abound.

Every `vault remove` rotate the keys for the entire vault via https://github.com/chef/chef-vault/blob/1206d0d5c0962e0490985ceefb91255dcfce61d4/lib/chef/knife/vault_remove.rb#L82.

Because a chef-vault is the combination of two data bags, item and item_keys, there is a possibility that the item data and the item keys
will not agree when updated concurrently.  This results in all clients being denied access to item data (including admins).

*Uncertainties*:
* I'm unsure if placing the mutex in the provisioner struct is the best place for it.
* Not sure if this qualifies for a documentation update?
* I didn't see any other bug reports relating to this issue, which is unexpected.
* I don't know if this qualifies for a test, but if deemed necessary, I'm happy to write one.

Let me know I missed anything in regards to the contributing guidelines.

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -85,6 +85,8 @@ enable_reporting false
 type provisionFn func(terraform.UIOutput, communicator.Communicator) error
 
 type provisioner struct {
+	sync.Mutex
+
 	Attributes            map[string]interface{}
 	Channel               string
 	ClientOptions         []string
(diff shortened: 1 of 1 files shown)


pkolyvas:
@lanej Josh, we've been struggling to support our contributors, and first time contributors most importantly. I realise this has gone unanswered for two years; we're trying really hard to get to a place where that doesn't happen again. 

Provisioners have been a sore spot for us for many reasons and it's going to get a little more sore for a while. 

We announced tool-specific provisioner deprecation in [mid-September 2020](https://discuss.hashicorp.com/t/notice-terraform-to-begin-deprecation-of-vendor-tool-specific-provisioners-starting-in-terraform-0-13-4/13997) and we added a deprecation notice for tool-specific (vendor or 3rd-party) provisioners in `0.13.4`. 

PR #18900, opened 2018-09-19
helper/pathorcontents: Skip one test when root

### LINKED ISSUE #7707: Cannot get terraform to build using 'make'
Hi, I'm attempting to build terraform from source using make. In fact here's a handy Dockerfile to completely demonstrate the failure:

```
FROM ubuntu:latest

RUN apt-get update && apt-get install -y git curl wget golang build-essential
RUN bash -c 'mkdir -p /opt/go/src/github.com/{hashicorp,mitchellh}'

ENV GOPATH=/opt/go
ENV PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin:/opt/go/bin

RUN ls /opt/go/src/github.com/

RUN cd /opt/go/src/github.com/hashicorp \
    && git clone https://github.com/hashicorp/terraform \
    && cd terraform \
    && git checkout v0.6.16 \
    && make
```

This always fails with:

```
--- FAIL: TestRead_PathNoPermission (0.00s)
    read_test.go:82: Expected error, got none!
FAIL
FAIL    github.com/hashicorp/terraform/helper/pathorcontents    0.003s
```

Is this test broken? It also fails when checking out master instead of v0.6.16. Thanks

phinze:
One of the tests in this package doesn't work when the tests are run as
root - like inside of a docker container. The test is still useful to
specify `pathorcontents` behavior when a file is not readable, so it's
better to skip than just delete it. See linked issue for further
disussion.

Closes #7707

### DIFF
--- a/helper/pathorcontents/read_test.go
+++ b/helper/pathorcontents/read_test.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/user"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


phinze:
Thanks @apparentlymart! Yep I went down a similar path of logic about how robust to make the check. Pushed w/ expanded comment per your suggestion. Will merge when green.

PR #18906, opened 2018-09-20
Fix proper nouns of vcs services

otofune:
This change affects document only.


mildwonkey:
thanks @otofune, great catch!

PR #18921, opened 2018-09-21
Implement new "shquote" interpolation function

nixar:
I haven't seen how to do this properly with the included functions, so I added my own. It makes it easier to robustly pass parameters to Posix shells. 

It simply encloses the whole value in single quotes, escaping only single quote itself (as `'\''`). As far as I can tell, Posix shells are supposed to handle just about anything thusly quoted, including control characters, but there may be cases I'm not aware of.

I included a simple unit test and updated the relevant documentation.


teamterraform:

This PR touches code that has changed quite a bit since it was originally opened. Specifically, the `config` package is mostly deprecated and any new functions now need to be implemented in `lang/funcs`. 

We are going to close this PR, but please do not hesitate to open a new PR or issue in the future, and thank you!

PR #18973, opened 2018-10-02
website: Reconcile docs about plugin discovery and downloading

nfagerlund:
I'm attempting to keep things simple for normal users while making sure we've
got the full behavior written down somewhere for plugin developers.

This PR doesn't stand alone; it's paired with a PR in the
terraform-website repo, to deal with some related content in the "extend"
section: https://github.com/hashicorp/terraform-website/pull/497

### DIFF
--- a/website/docs/commands/init.html.markdown
+++ b/website/docs/commands/init.html.markdown
@@ -120,23 +120,24 @@ During init, Terraform searches the configuration for both direct and indirect
 references to providers and attempts to load the required plugins.
 
 For [providers distributed by HashiCorp](/docs/providers/index.html),
-init will automatically download and install plugins if necessary. Other plugins
-can be manually installed in the user plugins directory, which is located at
+init will automatically download and install plugins if necessary. Plugins
+can also be manually installed in the user plugins directory, located at
 `~/.terraform.d/plugins` on most operating systems and
 `<APPLICATION DATA>\plugins` on Windows.
 
 For more information about configuring and installing providers, see
 [Configuration: Providers](/docs/configuration/providers.html).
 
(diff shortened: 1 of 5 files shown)


website/docs/commands/init.html.markdown
```diff
@@ -116,37 +116,41 @@ initialized with its child modules.
...
+init will automatically download and install plugins if necessary. Other plugins
+can be manually installed in the user plugins directory, which is located at
+`~/.terraform.d/plugins` on most operating systems and
+`<APPLICATION DATA>\plugins` on Windows.
+
+For more information about configuring and installing providers, see
+[Configuration: Providers](/docs/configuration/providers.html).
+
+On subsequent runs, init will keep any previously installed plugins and only
+install new providers that were added to the configuration since the last init.
```
mildwonkey: I'm not sure if it's worth mentioning here, but if the provider version constraint changes, and the existing version no long meets the constraints, terraform init will pull the correct version (be that up- or down-grade). 

I don't have a good sense if that's too much info for this block or not so this shouldn't keep you from merging.

website/docs/configuration/providers.html.md
```diff
@@ -207,24 +215,67 @@ may also be used, but currently may cause errors when running `terraform destroy
 
 ## Third-party Plugins
 
-At present Terraform can automatically install only the providers distributed
-by HashiCorp. Third-party providers can be manually installed by placing
-their plugin executables in one of the following locations depending on the
-host operating system:
+Anyone can develop and distribute their own Terraform providers. (See
```
mildwonkey: ❤️  I love the changes to this section.

PR #18979, opened 2018-10-02
backend/remote: ask to cancel a pending remote operation

svanharmelen:
Except when a lock-timeout has exceeded or auto-approve is set.

### DIFF
--- a/backend/remote/backend.go
+++ b/backend/remote/backend.go
@@ -440,7 +440,7 @@ func (b *Remote) Operation(ctx context.Context, op *backend.Operation) (*backend
 		}
 
 		if r != nil && err == context.Canceled {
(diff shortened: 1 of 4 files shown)


backend/remote/backend.go
```diff
@@ -497,6 +521,15 @@ func generalError(msg string, err error) error {
 	}
 }
 
+const generalErr = `
+%s: %v
+
+The "remote" backend encountered an unexpected error while communicating
+with remote backend. In some cases this could be caused by a network
```
jbardin: This first sentence is a little awkward, can we do something to get rid of the double "remote"?

svanharmelen: I see what you mean. I'll improve that a bit before merging the PR.

PR #18992, opened 2018-10-03
check responses for nil values

jbardin:
While providers shouldn't return nil values, we need to protect against
them. Always assign valid Null cty values when the response contains a
nil.


radeksimko:
Thanks 👍

PR #18993, opened 2018-10-03
Update remote.html.md

djaboxx:
Removed:
To use this backend you need a Terraform Enterprise account on app.terraform.io. A future release will also allow use of this backend on a private instance of Terraform Enterprise.

https://github.com/hashicorp/ptfe-releases/blob/6e105e9fce7620504ca3a7f9b097015d4db56de8/releases/v201808-1.md#changelog


pselle:
Thanks for this contribution! Closing this as outdated, as this section is no longer here.

PR #19007, opened 2018-10-04
Fix backend/swift docs

### LINKED ISSUE #19005: backend/swift docs appear to be incorrect about object versioning
I believe the wording on the object versioning warning in the backend/swift section of the documentation is incorrect. The link points to `archive_path` (which is deprecated for `archive_container`) but the wording says `expire_after` which is visible when you click the link so the error isn't necessarily obvious.

Imagine you have a backend configured something like:
```
terraform {
  backend "swift" {
    container    = "terraform-state"
    expire_after = "7d"
  }
}
```

This will cause your `tfstate.tf` file to be deleted 7 days after it was created. Which is certainly not what's desired. I believe the intended file is:

```
terraform {
  backend "swift" {
    container         = "terraform-state"
    archive_container = "terraform-state-archive"
  }
}
```

This appeared in https://github.com/hashicorp/terraform/commit/af5467b9fdc2613f975bbe4a6b9e6b29935a61ed#r30776677

cardoe:
The docs contained an error telling people to use `expire_after` when they really wanted object versioning. This error is actually quite damaging because instead of not even retaining versions of the state, it is outright deleted after a given amount of time. Fixed up the examples and other bits to not use deprecated items. Fixed bad links to the OpenStack Swift project. Added more links to the OpenStack Swift project. fixes #19005

### DIFF
--- a/website/docs/backends/types/swift.html.md
+++ b/website/docs/backends/types/swift.html.md
@@ -106,4 +106,4 @@ The following configuration options are supported:
    Swift [object versioning](https://docs.openstack.org/developer/swift/latest/overview_object_versioning.html) is enabled on the container created at `path`.
 
  * `expire_after` - (Optional) How long should the `terraform.tfstate` created at `container`
-   be retained for? Supported durations: `m` - Minutes, `h` - Hours, `d` - Days.
+   be retained for? If specified, Swift [expiring object support](https://docs.openstack.org/developer/swift/latest/overview_expiring_objects.html) is enabled on the state. Supported durations: `m` - Minutes, `h` - Hours, `d` - Days.
--- a/website/docs/backends/types/swift.html.md
+++ b/website/docs/backends/types/swift.html.md
@@ -12,7 +12,7 @@ description: |-
 
 Stores the state as an artifact in [Swift](http://docs.openstack.org/developer/swift/latest/).
 
-~> Warning! It is highly recommended that you enable [Object Versioning](https://docs.openstack.org/developer/swift/latest/overview_object_versioning.html) by setting the [`expire_after`](https://www.terraform.io/docs/backends/types/swift.html#archive_container) configuration. This allows for state recovery in the case of accidental deletions and human error.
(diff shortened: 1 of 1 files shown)


cardoe:
ping @tombuildsstuff since you marked up the issue I reported. This PR fixes that.

website/docs/backends/types/swift.html.md
```diff
@@ -99,11 +101,11 @@ The following configuration options are supported:
...
+   of the Terraform state file. If specified, Swift [object versioning](https://docs.openstack.org/developer/swift/latest/overview_object_versioning.html) is enabled on the container created at `container`.
 
  * `archive_path` - (Optional) DEPRECATED: Use `archive_container` instead.
    The path to store archived copied of `terraform.tfstate`. If specified,
-   Swift [object versioning](https://docs.openstack.org/developer/swift/overview_object_versioning.html) is enabled on the container created at `path`.
+   Swift [object versioning](https://docs.openstack.org/developer/swift/latest/overview_object_versioning.html) is enabled on the container created at `path`.
 
- * `expire_after` - (Optional) How long should the `terraform.tfstate` created at `path`
-   be retained for? Supported durations: `m` - Minutes, `h` - Hours, `d` - Days.
+ * `expire_after` - (Optional) How long should the `terraform.tfstate` created at `container`
```
tombuildsstuff: I'd suggest we want a warning box on this field too (since Terraform is inherently stateful) - and ditching the Statefile could cause issues - what do you think? e.g.

```
~> **NOTE:** Since Terraform is inherently stateful - we'd strongly recommend against auto-expiring Statefiles.
```

cardoe: Totally agree. I'll add that in and rebase since another change landed recently.

cardoe: Done. Please let me know if there's any other improvements that need to be made.

tombuildsstuff: thanks for making those changes. I'll ask @radeksimko just to give this a second glance for phrasing but this otherwise LGTM 👍

cardoe:
I've rebased it to hopefully make merging this easier. @radeksimko I'd appreciate some comments. The documentation on the website will literally delete tfstate for OpenStack backend users.

cardoe:
Any special trick that needs to get done to get the website to re-render the docs?

cardoe:
I can make a PR against branch v0.11

PR #19013, opened 2018-10-05
backend/remote: add checks for all flags we currently don’t support

svanharmelen:
For Plan only:
-module-depth=n

For Plan & Apply
-parallelism=m
-refresh=false
-var “foo=bar” and -var-file=foo


svanharmelen:
Thanks for the review @mildwonkey! I'll remove the commas (had the same sentence in two error messages) before merging the PR 👍

PR #19036, opened 2018-10-09
changed bucket resource name to example2

srvreeswijk:
changed the name to prevent AWS errors when pasting this code into the example code created thus far into the tutorial.


srvreeswijk:
I was to quick with proposing this change. Will test further on which changes to make to prevent AWS errors when adding the s3 bucket code into the example.

PR #19049, opened 2018-10-10
Fix two documentation glitches

lfrancke:
As far as I understand these two things are wrong.

1) I don't think we need the double "$" in the template

2) As far as I can tell the variable "count" is not used either in this example (it is in the next example though)

### DIFF
--- a/website/docs/configuration/interpolation.html.md
+++ b/website/docs/configuration/interpolation.html.md
@@ -444,7 +444,7 @@ A template data source looks like:
 
 ```hcl
 data "template_file" "example" {
-  template = "${hello} ${world}!"
+  template = "$${hello} $${world}!"
   vars {
     hello = "goodnight"
     world = "moon"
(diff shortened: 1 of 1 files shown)


mildwonkey:
Hi @lfrancke ! Thank you for submitting this PR. 

The first change is incorrect; the double `$` is a necessary part of the inline template syntax. If you run terraform on your example, with the double `$` removed, this is the output:

```
Error: data.template_file.example: 1 error(s) occurred:

* data.template_file.example: invalid variable syntax: "hello". Did you mean 'var.hello'? If this is part of inline `template` parameter
then you must escape the interpolation with two dollar signs. For
example: ${a} becomes $${a}.
```

Can you please revert that one change? Your other fix looks good to me!

lfrancke:
Thanks for the super quick reply!

In that case I'd rather add a comment to that effect. Chances are that I'm not the only one who misunderstands that.

nfagerlund:
Yeah, this is still confusing people, cf. #20267. 

The file moved, but these changes are still relevant, so I'm going to cherry-pick/re-create them... one sec.

nfagerlund:
@lfrancke I'm closing this one in favor of https://github.com/hashicorp/terraform/pull/20783 — I dealt with the moved file, I expanded your explanation of why to use $$, and also I changed the example to use the `file()` function the way any normal configuration would, which should cut any confusion off at the pass before the explanation is even needed.

PR #19070, opened 2018-10-14
Postgres backend

mars:
Implements the Postgres relational database as a Terraform backend with support for locking and workspaces.

Motivation is to enable [running Terraform in a Heroku app](https://github.com/mars/terraforming-app) with the [Heroku Postgres database add-on](https://www.heroku.com/postgres).

I gleaned most of the design from a combination of the **inmem**, **s3**, and **consul** backends, with a hefty dose of my own reasoning.


mars:
Just rebased on hashicorp/terraform master (9143cb5) which is now version **0.12.0-dev**.

mars:
After fighting with master, I found the **v0.11** branch to base this PR on until master (0.12.0-dev) stabilizes.

mars:
I've now rebased onto master (0.12)

Running tests, it looks like there are a bunch of API/interface changes. Would someone with knowledge of the 0.12 changes, please give me some hints or docs to help figure this out?

```
$ TF_PG_TEST=true make test TEST=./backend/remote-state/pg TESTARGS='-v'
go generate ./...
2018/11/07 15:32:56 Generated command/internal_plugin_list.go
go list ./backend/remote-state/pg | xargs -t -n4 go test -v -timeout=2m -parallel=4
go test -v -timeout=2m -parallel=4 github.com/hashicorp/terraform/backend/remote-state/pg
# github.com/hashicorp/terraform/backend/remote-state/pg [github.com/hashicorp/terraform/backend/remote-state/pg.test]
backend/remote-state/pg/backend.go:48:2: cannot use result (type *Backend) as type backend.Backend in return argument:
	*Backend does not implement backend.Backend (missing DeleteWorkspace method)
backend/remote-state/pg/backend_state.go:110:52: cannot use terraform.NewState() (type *terraform.State) as type *states.State in argument to stateMgr.WriteState
backend/remote-state/pg/backend_test.go:30:6: cannot use new(Backend) (type *Backend) as type backend.Backend in assignment:
	*Backend does not implement backend.Backend (missing DeleteWorkspace method)
backend/remote-state/pg/backend_test.go:47:39: cannot use config (type map[string]interface {}) as type hcl.Body in argument to backend.TestBackendConfig:
	map[string]interface {} does not implement hcl.Body (missing Content method)
backend/remote-state/pg/backend_test.go:47:50: impossible type assertion:
	*Backend does not implement backend.Backend (missing DeleteWorkspace method)
backend/remote-state/pg/backend_test.go:92:39: cannot use config (type map[string]interface {}) as type hcl.Body in argument to backend.TestBackendConfig:
	map[string]interface {} does not implement hcl.Body (missing Content method)
backend/remote-state/pg/backend_test.go:92:50: impossible type assertion:
	*Backend does not implement backend.Backend (missing DeleteWorkspace method)
backend/remote-state/pg/backend_test.go:115:39: cannot use config (type map[string]interface {}) as type hcl.Body in argument to backend.TestBackendConfig:
	map[string]interface {} does not implement hcl.Body (missing Content method)
backend/remote-state/pg/backend_test.go:115:50: impossible type assertion:
	*Backend does not implement backend.Backend (missing DeleteWorkspace method)
backend/remote-state/pg/backend_test.go:121:40: cannot use config (type map[string]interface {}) as type hcl.Body in argument to backend.TestBackendConfig:
	map[string]interface {} does not implement hcl.Body (missing Content method)
backend/remote-state/pg/backend_test.go:121:40: too many errors
FAIL	github.com/hashicorp/terraform/backend/remote-state/pg [build failed]
make: *** [test] Error 1

```

mars:
With Postgres now setup in TravisCI and [testing limited to only this new pg backend](https://github.com/hashicorp/terraform/pull/19070/commits/6a804ec9ca3d0146dba73f56f415e0ed78ad016d), this feature's tests are passing ✅

Of course, I will revert or undo that "Test only the pg backend in CI" commit before eventually merging.

Please let me know what y'all think 😄

jbardin:
Hi @mars,

Thank you for your patience. We're nearing the completion of the 0.12 work, and I'd like to evaluate merging this into the dev build for inclusion in the 0.12 release.

Could you rebase this and verify that everything is working correctly? I'll can still start in on reviewing this in the meantime.

For these tests, you can use the `TF_ACC=1` convention to flag these off as acceptance tests, rather than the custom `TF_PG_TEST`. 

Thanks!

mars:
Hi @jbardin, great to hear from you ☺️

Rebased on master & removed the `TF_PG_TEST` flag.

The tests show everything is still good ✅

I have not yet exercised it with `terraform apply`, because our Heroku provider is not yet compatible with master/0.12. Will update here once I figure that out 💫

backend/remote-state/pg/backend.go
```diff
@@ -0,0 +1,114 @@
...
+func New() backend.Backend {
+	s := &schema.Backend{
+		Schema: map[string]*schema.Schema{
+			"conn_str": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Postgres connection string; a `postgres://` URL",
+			},
+
+			"lock": &schema.Schema{
```
jbardin: Is there any reason to make locking optional here? S3 is optional only because it requires DynamoDB too, and the S3 remote state existed before locking.

backend/remote-state/pg/client.go
```diff
@@ -0,0 +1,158 @@
...
+		lockErr.Err = errors.New("state locked")
+		return "", lockErr
+	}
+
+	query := `INSERT INTO %s.%s (name, info) VALUES ($1, $2)`
+	data, err := json.Marshal(info)
+	if err != nil {
+		return "", err
+	}
+	_, err = c.Client.Exec(fmt.Sprintf(query, c.SchemaName, locksTableName), c.Name, data)
```
jbardin: Would there be any benefit to using native postrgres locks for this?

mars: Yes, definitely. This design was definitely first draft, MVP. I think native Postgres transaction locks will be more deterministic under contention and never require force unlocking. I’ll work on the implementation and follow-up here soon.

backend/remote-state/pg/backend.go
mars: Nice, I wondered why it was optional, but couldn’t determine the reasoning. Yes, I’ll remove the optional locking.

mars: Removed optional locking

backend/remote-state/pg/client.go
mars: Switched to Postgres advisory locks

mars:
Howdy @jbardin 😄

I switched this over to native Postgres locking and removed my special CI test commit that ran just these tests with `TEST_ACC=1`. I don't see that ACC flag being set in TravisCI, so I'm not sure these tests are really running.

I'll proceed with exercising this new version of the pg backend this week and follow up with any news.

mars:
I've now exercised this backend locally with Terraform built from master/0.12-dev.

If you've used this PR/branch before with a Terraform config, the database should be recreated to allow the new schema to take effect …but that's the only hiccup.

Otherwise, this appears to be working great ✅🙌

jbardin:
Thanks @mars.
Yes, travis doesn't handle the TF_ACC tests at all. Acceptance tests have to be setup individually in another system. While this set did run OK, the travis test suite as a whole is already quite heavy, and we'd rather not continue to add to to it.

backend/remote-state/pg/backend.go
```diff
@@ -0,0 +1,88 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Postgres connection string; a `postgres://` URL",
+			},
+
+			"schema_name": &schema.Schema{
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Name of the automatically managed Postgres schema to store state",
+				Default:     "terraform_remote_backend",
```
jbardin: let's make this `terraform_remote_state`, since "remote backend" can have different meanings

backend/remote-state/pg/client.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if c.txn == nil {
+		// Most strict transaction isolation to prevent cross-talk
+		// between incomplete state transactions.
+		txn, err = c.Client.BeginTx(context.Background(), &sql.TxOptions{
+			Isolation: sql.LevelSerializable,
+		})
+		if err != nil {
+			return "", err
```
jbardin: If there is an error in this function, we want to try and wrap it in a`*state.LockError` with the lock info. The ID isn't as important here if these are canceled if a client disconnects, but the info may be useful to other clients to indicate where the lock is being taken.

backend/remote-state/pg/client.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+// RemoteClient is a remote client that stores data in a Postgres database
+type RemoteClient struct {
+	Client     *sql.DB
+	Name       string
+	SchemaName string
+
+	// In-flight database transaction. Empty unless Locked.
+	txn  *sql.Tx
```
jbardin: we're going to want a mutex in here to protect these in the case of concurrent calls.

backend/remote-state/pg/backend.go
mars: Done

backend/remote-state/pg/client.go
mars: I see what you mean and [found](https://github.com/hashicorp/terraform/blob/master/state/state.go#L39) that the `LockError` type makes it be retried.

I've revised to return the `*state.LockError` type when retrying `Lock()` makes sense.

backend/remote-state/pg/client.go
mars: Good catch. Mutex is now in-place to ensure sanity.

jbardin:
Sorry, one last thing, let's remove the travis changes altogether. 
Thanks!

mars:
🏁

jbardin:
Oops, looks like there's a test that's not flagged off with `TF_ACC`

mars:
👌 thanks for the deep review @jbardin ☺️

jbardin:
Thanks @mars! lets see if we can test this out in master! 👍

mars:
Hi again @jbardin 😄

I've been using this backend and reviewing its current implementation. Wondering if the implementation of locking using database transactions is working as Terraform expects.

Currently, after calling `Lock()` any changes to state that are `Put()` will be lost if the transaction rolls-back due to a database connection failure, Go panic, or other unexpected error. Essentially, **calling `Unlock()` is required to commit changes**.

Do you think this is a problem? If so, I can switch from transactional advisory locking to session advisory locking, so that interrupting errors leave the state's partial updates intact.

jbardin:
Hi @mars, that's a good observation. Terraform doesn't assume the state storage is transactional in any way. While it currently only writes once at the end of each phase, it could in the future make more incremental writes to the storage, and we would want the last state written preserved.

If we could do that with the session based locks, I think that would be preferable. Thanks!

mars:
Okay, I'll open a new PR with that change very soon.

PR #19079, opened 2018-10-15
backend/remote: improve console output

svanharmelen:


backend/remote/backend.go
```diff
@@ -468,6 +475,182 @@ func (b *Remote) Operation(ctx context.Context, op *backend.Operation) (*backend
...
+						"Waiting for the manually locked workspace to be unlocked..." + elapsed))
+					continue
+				}
+			}
+
+			// Skip checking the workspace queue when we are the current run.
+			if w.CurrentRun == nil || w.CurrentRun.ID != r.ID {
+				found := false
+				options := tfe.RunListOptions{}
+			runlist:
```
evanphx: This is a bit long and could probably be broken out into a separate function, but not a big deal. It also highlights that we need include queue position as a first class API data point rather than forcing the client to calculate it.

svanharmelen: It is, you are right. Will look into refactoring this one when re-creating/re-applying this logic on the v0.12 codebase.

And again yes! I already discussed that one in the TFE team, so in v0.12 this will probably be a lot easier/nicer.

Thanks!

PR #19093, opened 2018-10-17
 fail nonfunctional resource tests

jbardin:
The helper/resource unit tests will panic, because they were using the
legacy terraform.MockResourceProvider, which doesn't have the same
internals required by the new GRPC shims.

Fail these tests for now, and a new test provider will need to be made
out of a schema.Provider instance.

Close the special listener that we use for provider tests in the plugin Close method.


jbardin:
moving to master

PR #19094, opened 2018-10-17
command Apply test fixes

jbardin:
-  add backend cli options after configuration
- don't make a backup of a nil state
- minor fixes for command Apply tests


jbardin:
moving to master

PR #19097, opened 2018-10-17
Improve the method to find out go files in Makefile

PeterDaveHello:
Properly exclude files under "./vendor/" path, and find "files" only.
The original method "might" found "directories" with the same naming,
or, might exclude occasionally exclude the go files with "vendor" string
in any part of its path.


PeterDaveHello:
@radeksimko @jbardin would you guys like to help review this one? Many thanks.

PR #19098, opened 2018-10-17
Improve shell scripts

PeterDaveHello:
Multiple shell scripts improvements here to make them more robbust.

PR #19096, opened 2018-10-17
Refactor Dockerfile & Dockerfile-release

PeterDaveHello:

PR #19099, opened 2018-10-17
Format go src files with gofmt v1.11.1

PeterDaveHello:
Fix formats issues since v0.12 development progress being merged back to master branch.

(https://github.com/hashicorp/terraform/commit/86e6481cc661b2ba694d353aab6d11838c066c7c / https://travis-ci.org/hashicorp/terraform/builds/442481009)


svanharmelen:
Hi @PeterDaveHello, thanks for this PR but it seems we already merged another PR for fixing the formatting (#19103). Didn't notice that you already created a PR with the same changes, otherwise I would have merged that one instead. So sorry for that!

PeterDaveHello:
Thanks, @svanharmelen, I got it, BTW, would you please also take a look at #19096 / #19097 / #19098? Thank you.

PR #19109, opened 2018-10-18
Prerequisite of consul agent on terraform node for state locking to w…

allthingsclowd:
…ork.

Debugged state locking issue with customer - customer presumed consul agent wasn't required as this page doesn't mention that fact. Without the agent Consul backend can be used but state locking is not possible.


remilapeyre:
Hi @allthingsclowd, this is not my experience. I've run tests with Consul and Terraform without running an agent on the node running Terraform and I had no issue with state locking.

Could you elaborate on the issues you faced?

Issue #15134, opened 2017-06-06
provisioner/chef: gracefully detect a reboot request


onedumslack:
### Terraform Version
Terraform v0.9.6

### Affected Resource(s)
Please list the resources as a list, for example:
- provisioner
  - chef
  - connection

### Terraform Configuration Files
resource "null_resource" "bootstrap_chef" {
  depends_on = ["aws_instance.ec2"]
  count      = "${var.ec2_num_instances}"

  provisioner "chef" {
    environment            = "${var.environment}"
    node_name              = "${var.environment_short}-${var.project}-${var.uniq_env_id}-${count.index}-${element(aws_instance.ec2.*.id, count.index)}"
    server_url             = "${var.chef_server_url}"
    recreate_client        = true
    os_type                = "windows"
    run_list               = ["ll_aws_windows"]
    user_name              = "${var.chef_username}"
    user_key               = "${file("${var.chef_userkey}")}"
    attributes_json        = <<EOF
      {
        "host_settings": {
          "hostname": "${var.environment_short}-${var.project}-${var.uniq_env_id}-${count.index}"
        }
      }
EOF
    connection {
      type           = "winrm"
      https          = true
      port           = 5986
      insecure       = true
      host           = "${element(aws_instance.ec2.*.private_ip, count.index)}"
      user           = "${var.ec2_user}"
      password       = "${var.ec2_password}"
      timeout        = "10m"
    }
  }
}

### Expected Behavior
With the chef provisioner when a chef bootstrap reboots a node, during the bootstrap it should detect this gracefully and wait a timeout period for it to come back online and continue the chef bootstrap.

### Actual Behavior
Reboots of a node terminate the current terraform apply session immediately and don't recover.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Nothing non standard I believe.  Just a simple powershell_script resource in a recipe that requests the node to reboot after changing it's hostname.

### References
Test kitchen supports this functionality by looking for a windows exit code 35 and then it has a retry count and timeout for reconnecting back in and finishing running.

onedumslack:
Another feature associated with this that would be extremely helpful is that if you could pass an argument to the provisioner telling it that when it detects a reboot to just exit successfully but don't try to reconnect.  This will allow to run cookbooks that do things like change the local admin password and then to reconnect again after the provisioner with different credentials using a local-exec.

smastrorocco:
+1

lokiwins:
+1

[...]

PR #19155, opened 2018-10-20
Chef: Gracefully Handle RFC062 Exit Codes

bdwyertech:
Allows exit code 35 and 37 which do not imply Chef failure.

https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md

Should help with #15134 and #16551

### DIFF
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -204,7 +204,7 @@ func Provisioner() terraform.ResourceProvisioner {
 			"max_retries": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  1,
+				Default:  0,
 			},
 			"no_proxy": &schema.Schema{
 				Type:     schema.TypeList,
@@ -392,6 +392,11 @@ func applyFn(ctx context.Context) error {
 	o.Output("Starting initial Chef-Client run...")
 
 	for attempt := 0; attempt <= p.MaxRetries; attempt++ {
+		// We need a new retry context for each attempt, to make sure
+		// they all get the correct timeout.
+		retryCtx, cancel := context.WithTimeout(ctx, comm.Timeout())
+		defer cancel()
+
 		// Make sure to (re)connect before trying to run Chef-Client.
 		if err := communicator.Retry(retryCtx, func() error {
 			return comm.Connect(o)
@@ -795,7 +800,7 @@ func decodeConfig(d *schema.ResourceData) (*provisioner, error) {
 		OSType:                d.Get("os_type").(string),
 		RecreateClient:        d.Get("recreate_client").(bool),
 		PreventSudo:           d.Get("prevent_sudo").(bool),
-		RetryOnExitCode:       getIntListAsMap(d.Get("retry_on_exit_code")),
+		RetryOnExitCode:       getRetryOnExitCodes(d),
 		RunList:               getStringList(d.Get("run_list")),
 		SecretKey:             d.Get("secret_key").(string),
 		ServerURL:             d.Get("server_url").(string),
@@ -855,12 +860,19 @@ func decodeConfig(d *schema.ResourceData) (*provisioner, error) {
 	return p, nil
 }
 
-func getIntListAsMap(v interface{}) map[int]bool {
+func getRetryOnExitCodes(d *schema.ResourceData) map[int]bool {
 	result := make(map[int]bool)
 
-	switch v := v.(type) {
-	case nil:
+	v, ok := d.GetOk("retry_on_exit_code")
+	if !ok || v == nil {
+		// Use default exit codes
+		result[35] = true
+		result[37] = true
+		result[213] = true
 		return result
+	}
+
+	switch v := v.(type) {
 	case []interface{}:
 		for _, vv := range v {
 			if vv, ok := vv.(int); ok {
--- a/builtin/provisioners/chef/resource_provisioner.go
+++ b/builtin/provisioners/chef/resource_provisioner.go
@@ -98,6 +98,7 @@ type provisioner struct {
 	PolicyName            string
 	HTTPProxy             string
 	HTTPSProxy            string
+	MaxRetries            int
 	NamedRunList          string
 	NOProxy               []string
 	NodeName              string
@@ -105,6 +106,7 @@ type provisioner struct {
 	OSType                string
 	RecreateClient        bool
 	PreventSudo           bool
+	RetryOnExitCode       []int
 	RunList               []string
 	SecretKey             string
 	ServerURL             string
@@ -115,9 +117,7 @@ type provisioner struct {
 	UserKey               string
 	Vaults                map[string][]string
 	Version               string
-	RetryOnExitCode       []int
 	WaitForRetry          int
-	MaxRetries            int
 
 	cleanupUserKeyCmd     string
 	createConfigFiles     provisionFn
@@ -201,6 +201,11 @@ func Provisioner() terraform.ResourceProvisioner {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"max_retries": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  1,
+			},
 			"no_proxy": &schema.Schema{
 				Type:     schema.TypeList,
 				Elem:     &schema.Schema{Type: schema.TypeString},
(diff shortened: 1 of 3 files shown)


communicator/remote/command.go
```diff
@@ -25,7 +25,7 @@ type Cmd struct {
 	Stderr io.Writer
 
 	// Once Wait returns, his will contain the exit code of the process.
-	exitStatus int
+	ExitStatus int
```
svanharmelen, line 28: I think it would be better to keep this field private and instead add a `ExitStatus()` method that returns the `exitStatus`. Not sure if it really matters that much, but think it provides a slightly better (safer) API.

svanharmelen:
I noticed this is an old PR, are you still willing to work with me to get the PR ready for merging? In that case please have a look at my suggestion and also please rebase the PR so that the test can (hopefully) run successfully.

Lastly it would be great if you can sign the contributor license agreement. Thanks!

bdwyertech:
@svanharmelen sure, I've also gotten quite a bit better with Go since.  I'll give this another whirl.

communicator/remote/command.go
bdwyertech: After looking at the code again, I should be able to pull the exit code from the error, ExitError looks to return it.

Not sure why I put this here in the first place.

bdwyertech:
@stevenoneill-jmfamily and others, would we want to attempt a retry here or is the pending reboot notification good enough?  Putting this in a retry loop if we get one of those exit codes should be doable, maybe with a configurable sleep

stevenoneill-jmfamily:
@bdwyertech thanks for the heads up on this, appreciate it!

That sounds reasonable to me, I like that the defaults result in nothing (keep existing behavior).  My two bits....from a Chef user pov, I feel like I'd be expecting behavior and names like the current test-kitchen experience.  eg, utilizing these variable names:

`retry_on_exit_code`
`max_retries`
`wait_for_retry`

It'd be cool to have the consistency across products, if possible.  If not, i'm still good with it because this feature would be really frickin useful :-)

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -703,7 +703,16 @@ func (p *provisioner) runCommand(o terraform.UIOutput, comm communicator.Communi
...
+		// Allow RFC062 Exit Codes
+		// https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md
+		switch err.ExitStatus {
+		case 35:
+			o.Output("Reboot has been scheduled in the run state")
+		case 37:
+			o.Output("Reboot needs to be completed")
+		default:
+			return err
+		}
```
svanharmelen, line 715: Guess it needs something like this:

```suggestion
		exitErr, ok := err.(remote.ExitError)
		if !ok {
			return err
		}
		
		// Allow RFC062 Exit Codes
		// https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md
		switch exitErr.ExitStatus {
		case 35:
			o.Output("Reboot has been scheduled in the run state")
		case 37:
			o.Output("Reboot needs to be completed")
		default:
			return exitErr
		}
```

shoekstra: Minor fix needed (and var name change to keep in line with the Puppet provisioner):

```suggestion
		// Allow RFC062 Exit Codes
		// https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md
		errStruct, _ := err.(*remote.ExitError)
		switch errStruct.ExitStatus {
		case 35:
			o.Output("Reboot has been scheduled in the run state")
		case 37:
			o.Output("Reboot needs to be completed")
		case 213:
			o.Output("Chef has exited during a client upgrade")
		default:
			return err
		}
```

I thought to include 213 as that's a controlled termination of the chef-client process and shouldn't be considered a failure.

bdwyertech:
Alright, so I have nothing better to do then clean up the laundry list...

@shoekstra @svanharmelen @stevenoneill-jmfamily  what do you think of this?

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -114,6 +115,9 @@ type provisioner struct {
 	UserKey               string
 	Vaults                map[string][]string
 	Version               string
+	RetryOnExitCode       []int
+	WaitForRetry          int
+	MaxRetries            int
```
svanharmelen, line 120: This list is ordered alphabetically to make it easy to search for certain fields. So can you please add these new ones in the correct order? Thanks!

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -252,6 +256,23 @@ func Provisioner() terraform.ResourceProvisioner {
...
+			"max_retries": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  1,
+			},
+			"wait_for_retry": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  30,
+			},
```
svanharmelen, line 275: Same here, the list is ordered alphabetically so can you please add these new ones in the correct order? Thanks!

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -252,6 +256,23 @@ func Provisioner() terraform.ResourceProvisioner {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			// Same defaults as Test-Kitchen
+			// https://github.com/test-kitchen/test-kitchen/blob/e5998e0dd1aa42601c55659da78f9b112ff9f8ee/lib/kitchen/provisioner/base.rb#L36-38
```
svanharmelen, line 260: These comments should probably not be here, but instead they options should be added and described in the docs (which are also in the repo in the `website` folder).

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -334,44 +359,79 @@ func applyFn(ctx context.Context) error {
...
 
-	// Cleanup the user key before we run Chef-Client to prevent issues
-	// with rights caused by changing settings during the run.
-	once.Do(cleanupUserKey)
+		for _, code := range p.RetryOnExitCode {
+			if code == exitStatus {
+				o.Output(fmt.Sprintf("Waiting %v seconds before reconnecting & re-running Chef...", p.WaitForRetry))
+				time.Sleep(time.Duration(p.WaitForRetry) * time.Second)
+				attempt++
+				goto retry
```
svanharmelen, line 431: Interesting way to solve this... But to be honest not one that I like or which seems idiomatic. If you are OK with it, I can make a suggestion to do this is a different way. Using `goto` is not something I would like to use and merge for this.

builtin/provisioners/chef/resource_provisioner.go
```diff
@@ -745,6 +805,9 @@ func decodeConfig(d *schema.ResourceData) (*provisioner, error) {
 		UserName:              d.Get("user_name").(string),
 		UserKey:               d.Get("user_key").(string),
 		Version:               d.Get("version").(string),
+		RetryOnExitCode:       getIntList(d.Get("retry_on_exit_code")),
+		WaitForRetry:          d.Get("wait_for_retry").(int),
+		MaxRetries:            d.Get("max_retries").(int),
```
svanharmelen, line 810: Same here, the list is ordered alphabetically so these new ones should be ordered like that as well. But let me add a commit to tweak your work just a little to see if we can get to a solution that is inline with the rest of the code in this provisioner...

shoekstra:
This looks great, but it is also a considerably bigger change than your initial PR and my concern is it'll be harder to get merged in without adding tests. It would be a shame for this PR to stagnate again as a lot of Chef users would benefit from your fix.

May I suggest creating a new PR that adds the retry functionality so it doesn't hold up merging your initial change? I think would have a better chance of being merged as is because the proposed change is much smaller.

builtin/provisioners/chef/resource_provisioner.go
bdwyertech: I'm ok with whatever method is more idiomatic @svanharmelen

svanharmelen: I couldn't push to your branch, so instead I made a PR: https://github.com/bdwyertech/terraform/pull/1 Please have a look and if your OK with it merge it so it will become part of this PR. Thanks!

bdwyertech: Merged -- Thanks @svanharmelen

svanharmelen: Cool. I asked a colleague if he could do one last check, and I guess we should be good to go after that...

svanharmelen: @shoekstra found a few small issues when testing, so we created a second PR (https://github.com/bdwyertech/terraform/pull/2) to your branch with some final tweaks. We tested a scenarios we could think of, so if you merge the PR I will merge this PR over here...

svanharmelen: Your too fast! I noticed one more thing in the docs... Will be PR #3 I guess. Hold on...

PR #19175, opened 2018-10-23
backend/local: do not use backend operation variables

### LINKED ISSUE #19163: Terraform 0.11.9 does not take in account any more the .auto.tfvars
### Terraform Version

```
Terraform v0.11.9
+ provider.archive v1.1.0
+ provider.aws v1.41.0
+ provider.null v1.0.0
+ provider.random v2.0.0
+ provider.template v1.0.0
+ provider.tls v1.2.0
```

### Expected Behavior
I expected terraform to populate some of my vars with the values declared in the .auto.tfvars files.

### Actual Behavior
When running a terraform plan/apply, the values of some variables declared in a *.auto.tfvars file are not taken in account any more and terraform asks for them during runtime - breaking my automated pipeline.

```

should now work.

rerun this command to reinitialize your working directory. If you forget, other
var.allow_ssh_cidr_blocks
  Whitelist for SSH access to EC2 instances.

  Enter a value:
```

```
$ cat ip-whitelists.auto.tfvars
allow_ssh_cidr_blocks = [
  "x.x.x.x/32", 
  "x.x.x.x/32", 
  "x.x.x.x/32", 
  "x.x.x.x/32", 
  "x.x.x.x/32", 
]
```
### Steps to Reproduce
terraform apply

### Additional Context
I have double checked by downgrading. Terraform 0.11.8 works.

svanharmelen:
Previously the `local` backend didn’t merge variables but completely replaced them when any backend operation variables were set.

Before [this change](https://github.com/hashicorp/terraform/commit/67db9da000#diff-a0d5ae04114a139f2356fb16fbac7581R174) this was essentially a dead codepath (that didn't have any tests) as the backend operation would never have a value for these variables and so the complete set of variables created while getting the `ContextOpts` would be used instead: https://github.com/hashicorp/terraform/blob/v0.11/command/meta.go#L312

But the variables were added to the backend operation because the `remote` backend needs access to these command line variables so they can be set as run variables. Run variables are not yet supported in this release, but this value is still used in the current `remote` backend to detect if command line variables were set, in which case a proper error can be returned.

We could also just remove the whole code path, instead of changing the behavior to merge instead of replace the variables. Will discuss that option before merging this.

I added a few tests to guard against this issue in any future releases.

Fixes #19163

_EDIT: See the first commit for this initial solution._

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -143,7 +143,11 @@ type Operation struct {
 	ModuleDepth  int
 	Parallelism  int
 	Targets      []string
-	Variables    map[string]interface{}
+
+	// Variables should only contain any variables passed as command
+	// arguments and not any variables read from the terraform.tfvars
+	// or *.auto.tfvars files.
+	Variables map[string]interface{}
 
 	// Input/output/control options.
 	UIIn  terraform.UIInput
--- a/backend/local/backend_apply_test.go
+++ b/backend/local/backend_apply_test.go
@@ -114,6 +114,40 @@ func TestLocal_applyEmptyDirDestroy(t *testing.T) {
 	checkState(t, b.StateOutPath, `<no state>`)
 }
 
+func TestLocal_applyWithVariables(t *testing.T) {
+	b, cleanup := TestLocal(t)
+	defer cleanup()
+	p := TestLocalProvider(t, b, "test")
+
+	p.ApplyReturn = &terraform.InstanceState{ID: "yes"}
+
+	mod, modCleanup := module.TestTree(t, "./test-fixtures/apply-vars")
+	defer modCleanup()
+
+	op := testOperationApply()
+	op.Module = mod
+	op.Variables = map[string]interface{}{"cli": "var"}
+
+	// Set some variables that would have been read from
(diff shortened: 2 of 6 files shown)


svanharmelen:
In the earlier stages of the command initialization, variables are separated by `autoVariabels` and `variables`. But in the `backend.Operation` struct there is only the `Variables` field to pass any variables. So it looks like this field was intended to have either all applicable variables, or none (in which case the ones from the context would be used). In that context, the code that was changes in this PR is actually correct and should not change.

But looking very closely through the code base, I noticed that the `Variables` field was never set (until the change referenced above) and read only once in the `local` backend. So I would suggest that I update this PR and just remove the updated lines of code from the `local` backend code instead of changing its behavior.

We can then assume that the field name matches the variable type that can be read from the `Meta` struct. In order to make that distinction more clear when reading the code we could add a `AutoVariables` field to the `backend.Operation` to reflect the different types of variables, but that field would never be used. At least not in the current code base.

So I suggest we, instead, add a very explicit comment to the `Variables` field to make clear it only expects the command line provided variables and not the auto loaded variables.

PR #19178, opened 2018-10-23
command/state: update and fix the state rm command

svanharmelen:


svanharmelen:
@apparentlymart I reverted a few of the changes you started with, to preserve the original behavior of this command.

Previously this command would remove whatever resources that matched the given addresses, in the same way that those addresses could be used as filter for the `state list` command. So what would be removed should be the same as what would be filtered using the same given addresses.

I did keep the `-dry-run` as that clearly adds value for a command like this. Of course we can always discuss the behavior and if we want to change it, but for now I tried to keep it as close to the previous behavior as possible.

PR #19197, opened 2018-10-25
command/state: update and fix the state mv command

svanharmelen:


svanharmelen:
I'm going to hold off merging merging this one as I want to double check if this is still an issue in this version: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/terraform-tool/B6pN4izPS4g/kQOWvXQtBwAJ

If it is still a problem in this version of the command, I'll add some logic to update the filter so we use the best/closest match instead of the first match returned. I think that solves another possible quirk.

PR #19196, opened 2018-10-25
Fix documented custom plugin filename convention

broamski:
This updates the terraform-bundle documentation to indicate the accurate required filename format.

PR #19225, opened 2018-10-30
update provider naming convention in readme

hagurganus:
This updates the provider naming convention in the terraform-bundle readme to use an underscore rather than a hyphen before the version identifier.

This makes it so that the tool can identify the custom provider and its version properly when parsing the given directory for custom providers. (I confirmed in testing that terraform-bundle ignores files with a hyphen between the provider name and the version identifier.)

This is related to [PR 19196](https://github.com/hashicorp/terraform/pull/19196), which appears to fix the same error in a different part of the readme.


hagurganus:
It looks like someone else made the equivalent suggestion in [a PR in February](https://github.com/hashicorp/terraform/pull/20385), and it was reviewed/approved (twice) within about half an hour of being opened, and merged the next morning.

I guess I should remember to name my PR appropriately next time? 🤔 

:)

Cheers!

PR #19227, opened 2018-10-31
website: clarify behavior of terraform_remote_state data sources

nfagerlund:
I've gotten confused about how these work multiple times over the last year, so here's what I've learned about them.


nfagerlund:
Travis failure is due to 0.12 configuration language changes, but this commit doesn't add any extra links to the deleted files.

PR #19286, opened 2018-11-05
helper/schema: Fix timeout parsing during Provider.Diff

radeksimko:
I'm not sure when/where/why we decided to produce a simple map instead of slice (read as _I didn't go through all the refactoring commits_), but it makes more sense to me, so I'm assuming it was intentional.

Also I don't know why are we were not originally erroring out on invalid timeout types as I couldn't find a corresponding PR to this commit https://github.com/hashicorp/terraform/commit/aa3677cd896809718f26d066216ded14673a7961
cc @catsby ^

### DIFF
--- a/helper/schema/provider_test.go
+++ b/helper/schema/provider_test.go
@@ -3,6 +3,7 @@ package schema
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


jbardin:
Thanks @radeksimko,

This makes sense, since blocks were always decoded in hcl as `[]map[string]interface{}`, and in special cases they were hoisted up into a single `map[string]interface{}`. Since timeouts were never in the actual schema for some reason, that differentiation couldn't be done in when the config was loading.

Now that they are present in the config as a `NestingSingle` block, they will be converted correctly by the shims before making it to the `ConfigDecode` function.

PR #19339, opened 2018-11-09
Add ssh keepalives to ssh communicator

### LINKED ISSUE #18517: Remote Exec Failing Even After Successful Execution of the Script
### Terraform Version
```
Terraform v0.11.7

* provider.null: version = "~> 1.0"
* provider.template: version = "~> 1.0"
```

```
 resource "null_resource" "pr13_remote_exec_0" {
    count = "1"
    
    provisioner "file" {
        content      = "${element(data.template_file.pr13_template_0.*.rendered, count.index)}"
        destination  = "/tmp/remote-exec.sh"
        connection {
            type     = "ssh"
            user     = "povijayan"
            private_key = "${file("/Users/povijayan/.ssh/id_rsa")}"
            host     = "${element(var.hosts_0,count.index)}"
        }
    }

    provisioner "remote-exec" {
        inline = [
            "sudo mkdir -p /opt/test/remote-exec-scripts",
            "sudo cp -Rvf /tmp/remote-exec.sh /opt/test/remote-exec-scripts/",
            "sudo chmod +x /opt/test/remote-exec-scripts/remote-exec.sh",
            "sudo /opt/test/remote-exec-scripts/remote-exec.sh"
        ]
        connection {
            type     = "ssh"
            user     = "povijayan"
            private_key = "${file("/Users/povijayan/.ssh/id_rsa")}"
            host     = "${element(var.hosts_0,count.index)}"
            timeout = "30m"
        }
    }
}
```
### Debug Output

Error: Error applying plan:

1 error(s) occurred:

* null_resource.pr13_remote_exec_0: error executing "/tmp/terraform_1857918131.sh": wait: remote command exited without exit status or exit signal

### Expected Behavior
Terraform remote exec should complete remote exec without any errors since remote exec script used having proper exit codes.

### Actual Behavior
Remote exec script which we are executing using above configuration is failing with above error.
When checked the script execution log it got executed fine and ended with proper exit code, but still remote exec is failing.

We are able to reproduce this with simple script like below(but it is not consistent). Not able to understand the reason for this error. Please help here.

```
#! /bin/bash
exec > >(tee /var/log/test.log|logger -t test -s 2>/dev/console) 2>&1

sleep 10m

exit 0

```

### Steps to Reproduce
1. `terraform init`
2. `terraform apply`

spion06:
This helps resolve an issue in stigged environments where the server-side ClientAliveInterval is set to a lower value (such as 300) and ClientAliveCount is set to 0. During the ssh provisioner if there is a long pause without data the server-side will disconnect causing terraform to fail. This would resolve that issue by sending session-level keepalive requests. The overhead of the ssh keepalive is minimal and shouldn't impact performance of anything. 

This should resolve #18517 based off a cursory search of open issues.

### DIFF
--- a/communicator/ssh/communicator.go
+++ b/communicator/ssh/communicator.go
@@ -61,6 +61,10 @@ type sshConfig struct {
 	// sshAgent is a struct surrounding the agent.Agent client and the net.Conn
 	// to the SSH Agent. It is nil if no SSH agent is configured
 	sshAgent *sshAgent
+
+	// KeepAliveInterval sets how often we send a channel request to the
+	// server. A value < 0 disables.
+	KeepAliveInterval time.Duration
 }
 
 type fatalError struct {
@@ -277,6 +281,22 @@ func (c *Communicator) Start(cmd *remote.Cmd) error {
 	if err != nil {
 		return err
 	}
+
+	// Start a goroutine that executes every KeepAliveInterval to keep the session alive
+	// will return when the session no longer responds to requests
+	go func() {
+		if c.config.KeepAliveInterval < 0 {
+			return
+		}
+		t := time.NewTicker(c.config.KeepAliveInterval)
+		defer t.Stop()
+		for range t.C {
+			_, err := session.SendRequest("keepalive@terraform.io", true, nil)
+			if err != nil {
+				return
+			}
+		}
+	}()
 
 	// Start a goroutine to wait for the session to end and set the
 	// exit boolean and status.
--- a/communicator/ssh/provisioner.go
+++ b/communicator/ssh/provisioner.go
@@ -34,6 +34,9 @@ const (
 
 	// DefaultTimeout is used if there is no timeout given
 	DefaultTimeout = 5 * time.Minute
+
+	// DefaultKeepAliveInterval is used if there is no interval given
+	DefaultKeepAliveInterval = 30 * time.Second
 )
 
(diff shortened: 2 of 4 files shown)


communicator/ssh/communicator.go
```diff
@@ -278,6 +282,22 @@ func (c *Communicator) Start(cmd *remote.Cmd) error {
 		return err
 	}
 
+	// Start a goroutine that executes every KeepAliveInterval to keep the session alive
+	// will return when the session no longer responds to requests
+	go func() {
+		if c.config.KeepAliveInterval < 0 {
```
jbardin: This will panics below with a timeout of `0`, and should probably be limited to a reasonable minimum like 1 second, so we don't flood the connection with `1ns` keepalives.

communicator/ssh/provisioner.go
```diff
@@ -51,6 +54,9 @@ type connectionInfo struct {
 	ScriptPath string        `mapstructure:"script_path"`
 	TimeoutVal time.Duration `mapstructure:"-"`
 
+	KeepAliveInterval    string        `mapstructure:"keep_alive_interval"`
```
jbardin: This isn't added to config validation, so it can't actually be used from a configuration.

jbardin:
Closing in favor of #20437, which removes the config option and only starts one keepalive routine per connection.

PR #19365, opened 2018-11-14
Move some content to learn.hashicorp.com

nfagerlund:
(To be merged on @tr0njavolta's instructions after the redirects are active.)

This is a PR to transition and redirect some getting started and guides content to https://learn.hashicorp.com/terraform. 

Companion PR in terraform-website: https://github.com/hashicorp/terraform-website/pull/554


nfagerlund:
@tr0njavolta You might have forgotten a step — this PR wasn't merged when you merged the one in terraform-website.

nfagerlund:
@apparentlymart Well, we'll have to follow up with this after the break, but: the last I heard was that "writing custom providers" was going to stay in both this repo and the learn site repo, but the others ("running terraform in automation," and "getting started") were not. 

Right now any links to those two get redirected to the learn site anyway, so the copies in this repo are ghost content that will never be shown... which makes them a bit of an attractive nuisance, since people will be tempted to update them instead of the live copies. 

In the near (a quarter or two?) future, @topfunky is hoping to source content for the learn site from multiple repos, which would let the content for the Terraform learning tracks live in this repo again. 

But for the time being: the files Rachel grabbed weren't supposed to be the canonical reference documentation for any given feature; they were just supposed to be guided walkthroughs that bring together already-documented material, such that moving them wouldn't leave major holes in terraform.io.

mildwonkey:
Hey @nfagerlund ! I would _really_ like to get this merged - do you want to clean the PR up, or do you want me to just open a new one removing these files? I'm more than happy to take this on.

nfagerlund:
This is now rebased and ready to merge! I've checked the approach with the rest of the education team, and I've also filed https://github.com/hashicorp/learn/pull/1067 to resolve drift between the two copies of this content.

PR #19368, opened 2018-11-14
website: add missing article

seishun:
I think "a module" is better than "modules" in this case because it makes it clear that one module can have more than one output value.


nfagerlund:
Travis build failures are from old issues we've already fixed.

PR #19429, opened 2018-11-21
Clarify paths for plugins and custom plugins

kikitux:


mildwonkey:
Hi @kikitux , thank you for submitting this.
We are leaving this information out of the `basics` section intentionally, as we do not expect it to be necessary for the average use case. 
This information is covered in detail [here](https://www.terraform.io/docs/extend/how-terraform-works.html#discovery), which is linked to from this page, so I am going to close this PR. Thank you!

PR #19431, opened 2018-11-21
commands: make sure the correct flagset is used

svanharmelen:
A lot of commands used `c.Meta.flagSet()` to create the initial flagset for the command, while quite a few of them didn’t actually use or support the flags that are then added.

So I updated a few commands to use `flag.NewFlagSet()` instead to only add the flags that are actually needed/supported.

Additionally this prevents a few commands from using locking while they actually don’t need locking (as locking is enabled as a default in `c.Meta.flagSet()`.

### DIFF
--- a/command/012_config_upgrade.go
+++ b/command/012_config_upgrade.go
@@ -29,7 +29,7 @@ func (c *ZeroTwelveUpgradeCommand) Run(args []string) int {
 
 	var skipConfirm, force bool
 
-	flags := c.Meta.flagSet("0.12upgrade")
+	flags := c.Meta.extendedFlagSet("0.12upgrade")
 	flags.BoolVar(&skipConfirm, "yes", false, "skip confirmation prompt")
 	flags.BoolVar(&force, "force", false, "override duplicate upgrade heuristic")
 	if err := flags.Parse(args); err != nil {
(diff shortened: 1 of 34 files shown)


jbardin:
I definitely like the cleanup. The only issue is that this loses the special UI output in the case of a flag error. I'm not a big fan of colorized and decorated cli output (and our UI abstraction might need to be revamped anyway), but this does allow the flag package to write directly to stderr, which might effect panicwrap. 

There's something else going on with the cli in this PR, possibly related to the UI and panicwrap, but I don't know what it is. The cli has a 12 second delay for all commands now:

```
$ date && TF_LOG=trace tf init
Wed Nov 21 13:10:56 EST 2018
2018/11/21 13:11:08 [INFO] Terraform version: 0.12.0 dev
...
```

svanharmelen:
That’s a wierd one. I didn’t notice that myself, but will have a look at it tomorrow.

To fix your other comment and address Martin his feedback, maybe I can refactor this so we have two flagset methods on `Meta`.

The existing (renamed) one and a new very basic one that essentialy only configured the base minumin (like the UI output).

svanharmelen:
I refactored this a little to reflect the discussion we had about this. I also tried to repro the startup delay, but I failed:

image

This was for a new (clean) `tf init` configured with an unconfigured `local` backend.

PR #19448, opened 2018-11-23
backend/azurerm: removing the `arm_` prefix from keys

tombuildsstuff:
This makes the Backend consistent with the AzureRM/Azure Stack Provider such that this is less confusing for users.

Rather than making this a breaking change I've added a compatibility path and removed references to the old names from the docs.

### DIFF
--- a/backend/remote-state/azure/backend.go
+++ b/backend/remote-state/azure/backend.go
@@ -57,28 +57,28 @@ func New() backend.Backend {
 				Description: "The resource group name.",
 			},
 
-			"arm_subscription_id": {
-				Type:        schema.TypeString,
-				Optional:    true,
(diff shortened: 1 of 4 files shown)


backend/remote-state/azure/backend.go
```diff
@@ -172,3 +205,13 @@ func (b *Backend) configure(ctx context.Context) error {
 	b.armClient = armClient
 	return nil
 }
+
+func valueFromDeprecatedField(d *schema.ResourceData, key, deprecatedFieldKey string) string {
```
radeksimko: This is nice idea 👍 🚿

tombuildsstuff:
Tests pass:

screenshot 2018-11-26 at 11 19 30

PR #19465, opened 2018-11-26
backend/azurerm: Support for authenticating using the Azure CLI

### LINKED ISSUE #16763: Support user accounts and/or SAS tokens for AzureRM backend auth

tombuildsstuff:
This PR adds support for authenticating to the Azure Backend using a token obtained from the Azure CLI

Fixes #16763

### DIFF
--- a/backend/remote-state/azure/arm_client.go
+++ b/backend/remote-state/azure/arm_client.go
@@ -65,6 +65,7 @@ func buildArmClient(config BackendConfig) (*ArmClient, error) {
 		MsiEndpoint:                   config.MsiEndpoint,
 
 		// Feature Toggles
+		SupportsAzureCliToken:          true,
 		SupportsClientSecretAuth:       true,
 		SupportsManagedServiceIdentity: config.UseMsi,
 		// TODO: support for Azure CLI / Client Certificate auth
--- a/backend/remote-state/azure/arm_client.go
+++ b/backend/remote-state/azure/arm_client.go
@@ -68,7 +68,7 @@ func buildArmClient(config BackendConfig) (*ArmClient, error) {
 		SupportsAzureCliToken:          true,
 		SupportsClientSecretAuth:       true,
 		SupportsManagedServiceIdentity: config.UseMsi,
-		// TODO: support for Azure CLI / Client Certificate auth
+		// TODO: support for Client Certificate auth
 	}
 	armConfig, err := builder.Build()
 	if err != nil {
--- a/go.mod
+++ b/go.mod
@@ -32,8 +32,6 @@ require (
 	github.com/coreos/go-systemd v0.0.0-20181012123002-c6f51f82210d // indirect
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f // indirect
 	github.com/davecgh/go-spew v1.1.1
-	github.com/dgrijalva/jwt-go v0.0.0-20160617170158-f0777076321a // indirect
-	github.com/dimchansky/utfbom v1.0.0 // indirect
 	github.com/dnaeon/go-vcr v0.0.0-20180920040454-5637cf3d8a31 // indirect
 	github.com/dylanmei/iso8601 v0.1.0 // indirect
 	github.com/dylanmei/winrmtest v0.0.0-20170819153634-c2fbb09e6c08
@@ -55,7 +53,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.5.1 // indirect
 	github.com/hashicorp/consul v0.0.0-20171026175957-610f3c86a089
 	github.com/hashicorp/errwrap v1.0.0
-	github.com/hashicorp/go-azure-helpers v0.0.0-20181122151743-c51a3103be3b
+	github.com/hashicorp/go-azure-helpers v0.0.0-20181126135526-ec113df69f49
 	github.com/hashicorp/go-checkpoint v0.0.0-20171009173528-1545e56e46de
 	github.com/hashicorp/go-cleanhttp v0.5.0
 	github.com/hashicorp/go-getter v0.0.0-20180327010114-90bb99a48d86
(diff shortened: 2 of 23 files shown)


tombuildsstuff:
Tests pass in Azure Public:

screenshot 2018-12-10 at 21 22 37

Tests pass in Azure Germany:

screenshot 2018-12-10 at 21 22 16

bentterp:
I would very much like to test this, but this seems to have been merged after alpha4 was released? @tombuildsstuff Is there a smart way of getting more recent builds of the 0.12 providers? And maybe even terraform itself?

bentterp:
@tombuildsstuff Thank you very much indeed for helping me here!!    
I managed to crosscompile terrform itself using the Vagrant approach and so I have been able to use the CLI method for backend authentication - this is so much needed, not just useful.   

Unfortunately I've not been able to figure out how to crosscompile providers, " ```XC_OS=windows XC_ARCH=amd64 make build```" just produces the linux provider

Neither can I figure out how to build them natively on Windows using Go and Gnuwin32 bash. 
I can get it to produce a terraform-provider-azurerm.exe but when I replace the one from the aplha release with the newly built one, I get " ```Error: Failed to instantiate provider "azurerm" to obtain schema: Incompatible API version with plugin. Plugin version: 4, Client versions: [5]``` " whenever I try to use it, even though I've changed  ~/go/src/github.com/terraform-providers/terraform-provider-azurerm/version/version.go from "dev" to "master" - did I misunderstand what you were suggesting?

I have to run this on Windows because our Azure AD only allows login from corporate managed devices.

Object of the exercise is to show enterprise architect that we can do a much better job with Terraform than with ARM templates, so I really need the 0.12 syntax which is such a huge improvement. I can live with a slightly older version of the provider, that's more of a nice-to-have.

Thank you very much for lending me a helping hand!

Regards,
  Bent

tombuildsstuff:
@bentterp 

> I managed to crosscompile terrform itself using the Vagrant approach and so I have been able to use the CLI method for backend authentication - this is so much needed, not just useful.

👍 glad to hear this is working for you

> Unfortunately I've not been able to figure out how to crosscompile providers, " XC_OS=windows XC_ARCH=amd64 make build" just produces the linux provider

You should be able to use `gox` to do this - if you first run `make tools` and then `gox -osarch="windows/amd64"` you should get an `*.exe` you can place in the same folder as the Terraform Core binary - however you'll need to update the SDK being used first:

> Neither can I figure out how to build them natively on Windows using Go and Gnuwin32 bash.
I can get it to produce a terraform-provider-azurerm.exe but when I replace the one from the aplha release with the newly built one, I get " Error: Failed to instantiate provider "azurerm" to obtain schema: Incompatible API version with plugin. Plugin version: 4, Client versions: [5] " whenever I try to use it, even though I've changed ~/go/src/github.com/terraform-providers/terraform-provider-azurerm/version/version.go from "dev" to "master" - did I misunderstand what you were suggesting?

So the `plugin version 4` bit here means you're running a plugin from 0.11 with 0.12, which are incompatible. Instead you'll need to ensure the Azure Provider is running the same version of the SDK as the 0.12 binary you've built, providing you're using a recent master, the following command should probably do what you need:

```
$ GO111MODULE=on go get github.com/hashicorp/terraform@master
```

at that point you should be able to re-run `gox` again and move the provider binary into the same directory as the Terraform Core binary. It's worth noting that if you're running custom binaries you'll need to ensure you do not specify the `version` argument in the provider block (as otherwise you'll download an incompatible provider) - and remove any existing `.terraform/` folders in the directory

Hope that helps :)

bentterp:
Now I've crossbuilt the provider using the vagrant process on Linux and that seemed promising until I ran into the plugin version problem again. Rereading your instructions, I think I understood them this time :-)

Unfortunately I can't get it to work:
```sh
vagrant@terraform:/opt/gopath/src/github.com/terraform-providers/terraform-provider-azurerm$ GO111MODULE=on go get github.com/hashicorp/terraform@master
go: finding github.com/hashicorp/terraform master
# cloud.google.com/go/iam
/opt/gopath/pkg/mod/cloud.google.com/go@v0.34.0/iam/iam.go:59:25: cannot use func literal (type func(context.Context, gax.CallSettings) error) as type gax.APICall in argument to gax.Invoke
/opt/gopath/pkg/mod/cloud.google.com/go@v0.34.0/iam/iam.go:71:25: cannot use func literal (type func(context.Context, gax.CallSettings) error) as type gax.APICall in argument to gax.Invoke
/opt/gopath/pkg/mod/cloud.google.com/go@v0.34.0/iam/iam.go:82:25: cannot use func literal (type func(context.Context, gax.CallSettings) error) as type gax.APICall in argument to gax.Invoke
vagrant@terraform:/opt/gopath/src/github.com/terraform-providers/terraform-provider-azurerm$
```
Do I need to update the google cloud library? Or does that need to be done as part of the docker image?

TraGicCode:
Hey @tombuildsstuff ,

Is it possible to cherry-pick this into a version of terraform that doesn't depend on 0.12?

tombuildsstuff:
@TraGicCode unfortunately we have no plans to do so since the backends are significantly different between 0.11 and 0.12 (both in terms of Terraform Core and in terms of the specific implementation for Azure, since we’ve rewritten large chunks of the backend such that this wouldn’t be easily portable unfortunately)

PR #19475, opened 2018-11-26
Shim computed containers

### LINKED ISSUE #19469: planned unknown value is after apply
An unknown value from plan is showing up as unknown after apply.

```
Error: Provider produced inconsistent final plan

When expanding the plan for null_resource.b to include new values learned so
far during apply, provider "null" produced an invalid new value for .triggers:
was null, but now cty.Value{ty: cty.Map(cty.String), v: map[string]interface
{}{"a":"e7cf7267-7816-88dc-093e-0e6ed46ccd67"}}.
```

Config:
```
resource "null_resource" "b" {
  triggers = {
    "a" = "${uuid()}"
  }
}
```

This also triggers a panic where `ElementIterator` is called on  the unexpected null value.

jbardin:
When normalizing flatmapped containers, compare the attributes to the
prior state and preserve pre-existing zero-length or unknown values. A
zero-length value that was previously unknown is preserved as a
zero-length value, as that may have been computed as such by the
provider.

This also fixes a crash in  `plans/objchange` by returning early to prevent iterating on a null value.

Fixes #19469

### DIFF
--- a/builtin/providers/test/resource_diff_suppress.go
+++ b/builtin/providers/test/resource_diff_suppress.go
@@ -81,6 +81,10 @@ func testResourceDiffSuppressCreate(d *schema.ResourceData, meta interface{}) er
 	d.Set("network", "modified")
 	d.Set("subnetwork", "modified")
 
+	if _, ok := d.GetOk("node_pool"); !ok {
+		d.Set("node_pool", []string{})
+	}
+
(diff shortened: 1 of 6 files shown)


jbardin:
rebased in the `hcl2shim.UnknownVariableValue` values

